PR #6, opened 2017-12-14
Added setting about the cpu usage of blas library

kan-bayashi:
Added setting about the cpu usage of blas library.


sw005320:
I think it requires only one of them (maybe OMP_NUM_THREADS).
Can you test it, and only keep one setting?
I don't want to make `path.sh` complicated.

kan-bayashi:
Removed unnecessary lines.

PR #7, opened 2017-12-14
[WIP] use hdf5 instead of kaldi_io

kan-bayashi:
Under construction.


kan-bayashi:
I added examples.
Before running, please install h5py.
```bash
. ./path.sh
pip install h5py
```
Then, run the `run_hdf5.sh`
```
./run_hdf5.sh
```

Issue #18, opened 2017-12-18
add unit-tests and Travis CI (.travis.yml)


ShigekiKarita:
Testing python2/3 and chainer/pytorch compatibility by our hand will be hell. we need some continuous integration on this. I think chainer's setting is good for us to test and build doc automatically. https://github.com/chainer/chainer/blob/master/.travis.yml

sw005320:
I understand and I will work on it. But at least, our setup is based on virtualenv and making an isolated setup with python 2.7, which is one way to avoid this kind of issues. So I want you to follow this setup for now.

ShigekiKarita:
gotcha 👌

ShigekiKarita:
I wonder if there were tiny toy example in espnet to run end-to-end test and benchmark in a short time. Some frameworks use AN4 for that purpose. I want someone to create AN4 egs because I'm not very familiar with kaldi. 

- an4 in CNTK https://github.com/Microsoft/CNTK/tree/master/Examples/Speech/AN4
- an4 in torch deepspeech https://github.com/SeanNaren/deepspeech.torch
- an4 in pytorch deepspeech https://github.com/SeanNaren/deepspeech.pytorch

ShigekiKarita:
I'm working on it in advance of you. I just copied chainer/.travis.yml and delete and add a few lines. Although I wrote only stupid tests, we can [see a python2/3 unicode problem here](https://travis-ci.org/ShigekiKarita/espnet/builds/318193090)

Also there can be several problems and discussions.
- pytorch installation on Travis-CI might not be possible now https://github.com/uber/pyro/pull/627
- end-to-end testing with smaller but real corpus
- need to lint python coding style like chainer? (I think it is too strict)

sw005320:
Thanks for the suggestion.
I can definitely make an4 egs.
Wait for a while.

ShigekiKarita:
@sw005320 now https://github.com/espnet/espnet/blob/master/.travis.yml is in this repo, you can enable travis ci. (I do not have such a permission.)

all you need to do is step1-2 in https://docs.travis-ci.com/user/getting-started#To-get-started-with-Travis-CI

sw005320:
@ShigekiKarita I've finished the an4 example. I don't tune the performance so much, but I think it would be okay for the test purpose.

ShigekiKarita:
Thanks!

Issue #9, opened 2017-12-14
pytorch exp


ShigekiKarita:
Hi, @sw005320 

this is not an issue, just my homework for fun. I'm making pytorch backend on your script for chainer. 

https://github.com/ShigekiKarita/espnet/blob/pytorch/src/bin/asr_train_th.py

Unless nothing is wrong (well, I don't think so ... for example, the lack of CTC loss implementation), pytorch seems twice faster than chainer in my poor environment (GTX760).

sw005320:
Great!!!
I think it would be nice to have two implementations (like Keras that uses several backends).
We can discuss it tomorrow.
I'm really excited.

I would say that my chainer-based implementation is already very nice level in terms of training speed, and if it would be twice faster, it is super duper.

ShigekiKarita:
Training BLSTMP on WSJ in 4 hours by pytorch and 1080ti

```
chainer:
this epoch [#####.............................................] 10.92%
       200 iter, 0 epoch / 15 epochs
   0.50112 iters/sec. Estimated time to finish: 15:07:17.560377.

pytorch:
this epoch [##########........................................] 21.83%                                              
       400 iter, 0 epoch / 15 epochs                      
    1.8472 iters/sec. Estimated time to finish: 4:04:20.062329.
```

sw005320:
This is amazing.
We should definitely consider to have both backends.
Please continue on it, and also try to include CTC and try to make it similar to the chainer backend.

ShigekiKarita:
@sw005320 I've implemented the CTC loss function with [WarpCTC binding](https://github.com/SeanNaren/warp-ctc) (not push yet). It does not matter at all in speed.

```
this epoch [#####.............................................] 10.92%
       200 iter, 0 epoch / 15 epochs
    1.9304 iters/sec. Estimated time to finish: 3:55:31.877189.
```

ShigekiKarita:
Pytorch backend is finally merged. As @sw005320 noted in #17, we need to add pytorch backend not only in WSJ but also in all the egs. It is not difficult for me and also for anyone else.

However there is a new concern about results. Here are WSJ-BLSTMP results by chainer and pytorch.

- chainer

![acc](https://user-images.githubusercontent.com/6745326/34089579-73f5c028-e3f4-11e7-9811-774a1fdbd7df.png)

![loss](https://user-images.githubusercontent.com/6745326/34089665-48bd8228-e3f5-11e7-90b7-83c2b60b1641.png)


[...]

PR #20, opened 2017-12-18
Test and Travis-CI

ShigekiKarita:
I'm working on Travis-CI #18


ShigekiKarita:
@sw005320 I want you to merge this if it is OK.

ShigekiKarita:
just a moment, I found Travis-CI is falling. https://travis-ci.org/ShigekiKarita/espnet

ShigekiKarita:
Now, Travis is OK. Many tests are skipped but we can wait for him https://github.com/pytorch/pytorch/issues/4178#issuecomment-352910003

src/nets/e2e_asr_attctc_th.py
```diff
@@ -417,6 +508,13 @@ def forward(self, enc_hs_pad, enc_hs_len, dec_z, att_prev, scaling=2.0):
         return c, w
 
 
+def th_accuracy(y_all, pad_target, ignore_label):
+    acc = 0
+    pad_pred = y_all.data.view(pad_target.size(0), pad_target.size(1), y_all.size(1)).max(2)[1]
+    mask = pad_target.data != ignore_label
+    return torch.sum(pad_pred.masked_select(mask) == pad_target.data.masked_select(mask)) / torch.sum(mask)
```
kan-bayashi: Final calculation `torch.sum(pad_pred.masked_select(mask) == pad_target.data.masked_select(mask)) / torch.sum(mask)` is `int / int`, therefore, in the case of python2, acc become = 0. (I confirmed) 

We should cast to float. 
I will show my modified version. 
```
def th_accuracy(y_all, pad_target, ignore_label):
    pad_pred = y_all.data.view(pad_target.size(0), pad_target.size(1), y_all.size(1)).max(2)[1]
    mask = pad_target.data != ignore_label
    numerator = torch.sum(pad_pred.masked_select(mask) == pad_target.data.masked_select(mask))
    denominator = torch.sum(mask)
    return float(numerator) / float(denominator)
```

src/nets/e2e_asr_attctc_th.py
```diff
@@ -484,21 +583,15 @@ def forward(self, hpad, hlen, ys):
...
         # acc = F.accuracy(y_all, F.concat(pad_ys_out, axis=0), ignore_label=-1)
-        acc = 0
-        pred_pad = y_all.data.view(len(ys), olength, y_all.size(1)).max(2)[1]
-        for i in range(len(ys)):
-            acc += torch.sum(pred_pad[i, :ys[i].size(0)] == ys[i].data)
-        acc /= sum(map(len, ys))
+        acc = th_accuracy(y_all, pad_ys_out, ignore_label=self.ignore_id)
         logging.info('att loss:' + str(self.loss.data))
 
         # show predicted character sequence for debug
```
kan-bayashi: In the case of chainer, ignore label is `-1`.
Therefore, we should change as follows.
```
# now 
                 idx_hat = np.argmax(y_hat_[y_true_ != -1], axis=1)
                 idx_true = y_true_[y_true_ != -1]
# proposed
                 idx_hat = np.argmax(y_hat_[y_true_ != self.ignore_id], axis=1)
                 idx_true = y_true_[y_true_ != self.ignore_id]
```

kan-bayashi:
@ShigekiKarita Hi Shigeki. I added some comments.
Could you reflect them?

src/nets/e2e_asr_attctc_th.py
```diff
@@ -304,35 +394,36 @@ def reset(self):
...
-        batch = len(enc_hs)
+        batch = enc_hs_pad.size(0)
         # pre-compute all h outside the decoder loop
         if self.pre_compute_enc_h is None:
-            self.enc_h = F.pad_sequence(enc_hs)  # utt x frame x hdim
+            self.enc_h = enc_hs_pad  # utt x frame x hdim
             self.h_length = self.enc_h.shape[1]
             # utt x frame x att_dim
-            self.pre_compute_enc_h = F.tanh(linear_tensor(self.mlp_enc, self.enc_h))
+            self.pre_compute_enc_h = linear_tensor(self.mlp_enc, self.enc_h)
```
kan-bayashi: Does it need to perform `tanh`?

src/nets/e2e_asr_attctc_th.py
ShigekiKarita: Thanks!

src/nets/e2e_asr_attctc_th.py
ShigekiKarita: OK. I think you are talking about `show predicted character sequence for debug` at Decoder.forward.

kan-bayashi: Yes, that's right.

src/nets/e2e_asr_attctc_th.py
ShigekiKarita: Hmm, I'm not sure why I delete it. However I will follow the reference  impl of chainer.

ShigekiKarita: In eq 9 in https://arxiv.org/pdf/1609.06773.pdf, `pre_compute_h` corresponds to `V h_l ` that does not have tanh. Anyway it might not be big deal because we cannot see big difference in https://github.com/espnet/espnet/issues/9#issuecomment-353366442

ShigekiKarita:
Thank you for many comments.

src/nets/e2e_asr_attctc_th.py
```diff
@@ -469,7 +469,7 @@ def forward(self, enc_hs_pad, enc_hs_len, dec_z, att_prev, scaling=2.0):
             self.enc_h = enc_hs_pad  # utt x frame x hdim
             self.h_length = self.enc_h.shape[1]
             # utt x frame x att_dim
-            self.pre_compute_enc_h = linear_tensor(self.mlp_enc, self.enc_h)
+            self.pre_compute_enc_h = torch.tanh(linear_tensor(self.mlp_enc, self.enc_h))
```
kan-bayashi: `tanh` op is performed for only `AttDot` in Chainer.
I think it is not needed for `AttLoc`.

ShigekiKarita: oh that is right

ShigekiKarita:
Does everything look good? @kan-bayashi @sw005320

sw005320:
can you include pytorch and others in all at Makefile?

ShigekiKarita:
done

src/nets/e2e_asr_attctc_th.py
mpc001: Sorry, but I noticed the current version self.ignore_id=0. Is 0 for CTC?

sw005320: Are you talking about this?
https://github.com/espnet/espnet/blob/c044f7a25dad63282071a38cef5b4678f9158fc3/src/nets/e2e_asr_attctc_th.py#L1633
This may have some problems.
@ShigekiKarita, can you tell me why you set 0 here?

kan-bayashi: In current implementation, character index for decoder is 1 to odim - 1.
Therefore, 0 is not used in decoder and it might be no effect even if we use 0 as ignore id, I think.

kan-bayashi: But to make more clear, I agree to change it to -1.

mpc001: @kan-bayashi Many thanks for your reply!

PR #42, opened 2017-12-26
use pytorch0.2.0 in travis.yml and add some 0.2.0 compatiblity workaround

ShigekiKarita:
Pytorch 0.3.0 binary is larger than travis-CI limit (https://github.com/pytorch/pytorch/issues/4178). As a workaround, we can use pytorch0.2.0 like pyro https://github.com/uber/pyro/blob/dev/.travis.yml

There were two incompatible parts between 0.2.0 and 0.3.0
1. no `.shape` attribute in 0.2.0. we have to use `.size()` method (these are exactly same)
2. softmax/log_softmax has no `dim=1` argument in 0.2.0 but 0.3.0 raise warnings if it is missing

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,10 +15,18 @@ matrix:
 install:
   - pip install -U pip wheel
   - pip install -r ./tools/test_requirements.txt
-  # unable to install pytorch as https://github.com/pytorch/pytorch/issues/4178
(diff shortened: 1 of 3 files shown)


ShigekiKarita:
@kan-bayashi could you look this PR?

kan-bayashi:
LGTM

ShigekiKarita:
https://travis-ci.org/espnet/espnet/jobs/321678988#L682

You can see all the tests are running except for the kaldi-python and kaldi-io-for-python compatibility test

PR #46, opened 2017-12-27
added end detection

sw005320:
Added end detection described in Eq. (50) of S. Watanabe et al "Hybrid CTC/Attention Architecture for End-to-End Speech Recognition,"
If we set maxlenratio 0.0, this automatic detection is triggered.
With WSJ and Voxforge experiments, it is confirmed that the detection works well.

### DIFF
--- a/src/nets/e2e_asr_attctc.py
+++ b/src/nets/e2e_asr_attctc.py
@@ -309,6 +309,9 @@ class AttDot(chainer.Chain):
         # <phi (h_t), psi (s)> for all t
         e = F.sum(self.pre_compute_enc_h * F.tile(F.reshape(F.tanh(self.mlp_dec(dec_z)), (batch, 1, self.att_dim)),
                                                   (1, self.h_length, 1)), axis=2)  # utt x frame
+        # Applying a minus-large-number filter to make a probability value zero for a padded area
+        # simply degrades the performance, and I gave up this implementation
+        # Apply a scaling to make an attention sharp
         w = F.softmax(scaling * e)
         # weighted sum over flames
         # utt x hdim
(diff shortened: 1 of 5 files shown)


src/nets/e2e_asr_attctc_th.py
```diff
@@ -697,6 +697,27 @@ def recognize(self, h, recog_args):
 
         return y_seq
 
+    # end detection desribed in Eq. (50) of
+    # S. Watanabe et al "Hybrid CTC/Attention Architecture for End-to-End Speech Recognition"
+    def end_detect(self, ended_hyps, i, M=3, D_end=np.log(1 * np.exp(-10))):
```
ShigekiKarita: I think this function should be in somewhere outside e2e_asr_attctc[_th].py and be shared from them like `from e2e_common import end_detection`. Because this function seems to be static (not use self) and be free from numpy/torch operations.

src/nets/e2e_asr_attctc.py
```diff
@@ -570,8 +596,11 @@ def recognize_beam(self, h, recog_args, char_list):
 
         # preprate sos
         y = self.xp.full(1, self.sos, 'i')
-        # maxlen >= 1
-        maxlen = max(1, int(recog_args.maxlenratio * h.shape[0]))
+        if recog_args.maxlenratio == 0:
```
ShigekiKarita: please add this specialized behavior in argparse help of asr_recog[_th].py . If you think it is worth being default, set it to 0.0 (now default is 0.5)

src/nets/e2e_asr_attctc.py
```diff
@@ -546,6 +551,27 @@ def recognize(self, h, recog_args):
 
         return y_seq
 
+    # end detection desribed in Eq. (50) of
+    # S. Watanabe et al "Hybrid CTC/Attention Architecture for End-to-End Speech Recognition"
+    def end_detect(self, ended_hyps, i, M=3, D_end=np.log(1 * np.exp(-10))):
```
ShigekiKarita: ditto (see e2e_asr_attctc_th.py)

src/nets/e2e_asr_attctc.py
ShigekiKarita: and I also think argparse should be shared like `from e2e_common import asr_recog_parser` like I said about `def end_detect` because they should be consistent.

sw005320: I'll do it for end_detect, but let me make "argparse part" common later.
Actually, asr_recog.py and asr_recog_th.py are almost same, and I'm thinking of using the common asr_recog.py for both, which is more simple for me, but I'm not sure it would be applicable to asr_train.py and asr_train_th.py, and it needs some more consideration.

ShigekiKarita: OK. It may be too much to this PR.

sw005320:
@ShigekiKarita Done!

Issue #35, opened 2017-12-25
Toward a stable version


sw005320:
I think we have fixed many issues, and we can add a version 1.0 (or 0.1) as a stable version.
Toward that we need to finish

- [ ] VGG2L for pytorch by @ShigekiKarita 
- [x] AN4 recipe by me
- [x] AMI recipe
- [x] swbd recipe
- [x] fisher_swbd recipe
- [x] LM integration @sw005320
- [x] Attention/CTC joint decoding @takaaki-hori 
- [x] End detection
- [ ] Documentation by @sw005320 @kan-bayashi 
- [x] Modify L.embed to avoid the randomness @takaaki-hori 
- [x] Add WER scoring
- [x] label smoothing by @takaaki-hori 
- [x] replace `_ilens_to_index` to `np.cumsum`
- [x] refactor main training and recognition to be independent of pytorch and chainer backends.

If you have any action items, please add them in this issue. 
Then, we can move to more research-related implementation.

kan-bayashi:
At least, we should add docstring for `src/nets`.

ShigekiKarita:
I agree. Type (class) and shape are essential information for everyone.

sw005320:
Frankly, I don't have so much experience, and if @kan-bayashi initiates this, I'll follow you and add/modify/enhance the document. 
Also, we should make a webpage somewhere.
Do you have any idea (just using github website host service?)?

sw005320:
The implementation of the end detection is finished #46

sw005320:
@ShigekiKarita I'm thinking of implementing the LM integration. This is performed by modifying an existing chainer's ptb recipe to train an LSTMLM (https://github.com/chainer/chainer/blob/master/examples/ptb/train_ptb.py). Then, integrate LSTMLM with our main E2E. Can I ask you to make pytorch version of the training part later? Once you make the LSTMLM training part, I can implement the pytorch integration part. If you agree, I'll start the chainer-based implementation. If you think we should implement LSTMLM training part more seamless way for both pytorch and chainer, let's do it.

ShigekiKarita:
@sw005320 It sounds nice. I like the separate ways because I'll be little bit away from here on a few weeks around Jan 1st but I will keep on watching and discussing with you.

And you can find PTB example in pytorch here https://github.com/pytorch/examples/tree/master/word_language_model

sw005320:
Which is easier for you to port to the pytorch-backend LSTMLM training?
[...]

PR #53, opened 2018-01-03
[WIP] integrates a language model

sw005320:
added several items to ignore when git

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,6 +2,9 @@
 *.pyc
 \#*\#
 .\#*
+*DS_Store
+
+# recipe related
 egs/*/*/data*
 egs/*/*/db
 egs/*/*/downloads
@@ -9,7 +12,14 @@ egs/*/*/dump
 egs/*/*/exp
 egs/*/*/fbank
 egs/*/*/stft
-*DS_Store
+egs/*/*/enhan
 
-
+# tools related
 src/utils/kaldi_io_py.py
+tools/kaldi
+tools/kaldi-io-for-python/
+tools/kaldi_github/
+tools/miniconda.sh
+tools/nkf/
+tools/venv/
+tools/warp-ctc/


sw005320:
@takaaki-hori Please review it.

src/nets/e2e_asr_attctc.py
```diff
@@ -606,7 +619,13 @@ def recognize_beam(self, h, recog_args, char_list):
                 hyp['a_prev'] = att_w
 
                 # get nbest local scores and their ids
-                local_scores = F.log_softmax(self.output(z_list[-1])).data
+                if rnnlm:
+                    rnnlm_state, z_rnnlm = rnnlm.predictor(hyp['rnnlm_prev'], hyp['yseq'][i])
+                    hyp['rnnlm_prev'] = rnnlm_state
```
takaaki-hori: Is this necessary? hyp['rnnlm_prev'] maybe not used anymore.

src/nets/e2e_asr_attctc.py
```diff
@@ -606,7 +619,13 @@ def recognize_beam(self, h, recog_args, char_list):
...
+                    local_scores = (1 - recog_args.lm_weight) * F.log_softmax(self.output(z_list[-1])).data \
+                        + recog_args.lm_weight * F.log_softmax(z_rnnlm).data
```
takaaki-hori: I didn't apply (1 - lm_weight) to ATT score in our Interspeech paper because the score was already combined with CTC score as (1-lambda) * ATT + lambda * CTC and I thought that (1-lm_weight) would complicate the equation and parameter tuning. What do you think?

src/nets/e2e_asr_attctc.py
sw005320: Thanks! Maybe You're right. We don't need `hyp[...] = ...`. I'll test it.

src/nets/e2e_asr_attctc.py
sw005320: I agree with you.
I'll test it.

sw005320:
@takaaki-hori if this looks good. I'll merge it. Then, we can fix your joint decoding #66 given this commit.

takaaki-hori:
@sw005320 the code looks good. you can merge it.

sw005320:
#35

PR #66, opened 2018-01-16
Add joint CTC/attention decoding

takaaki-hori:
#35


src/nets/e2e_asr_attctc.py
```diff
@@ -589,6 +608,13 @@ def recognize_beam(self, h, recog_args, char_list):
         hyp = {'score': 0.0, 'yseq': [y], 'c_prev': c_list, 'z_prev': z_list, 'a_prev': a}
         hyps = [hyp]
         ended_hyps = []
+        if lpz is not None:
+            ctc_prefix_score = CTCPrefixScore(lpz, 0, self.eos, self.xp)
+            hyp['ctc_prev'] = ctc_prefix_score.initial_state()
+            score_type = 'joint_score'
```
sw005320: It looks good, but what is your motivation of keeping both score and joint_score? If this is necessary, it is fine. If we keep these scores separately, another option is to keep `att_score` (current `score`) and `ctc_score`. Then, we may also additionally keep `total_score` so that we can extract all of these three scores, which will be good for a readability, and also good for debugging (we can provide both att, ctc, and total scores). Then, we could also add another scores with a same manner (e.g.,  `rnnlm_score`)

src/nets/ctc_prefix_score.py
```diff
@@ -0,0 +1,78 @@
...
+
+        :return: CTC state
+        '''
+        r = self.xp.full((self.input_length, 2), self.logzero, dtype=np.float32)
+        r[0, 1] = self.x[0, self.blank]
+        for i in six.moves.range(1, self.input_length):
+            r[i, 1] = r[i - 1, 1] + self.x[i, self.blank]
+        return r
+
+    def __call__(self, y, cs, r_prev):
```
sw005320: This part looks different from Algorithm 2 in the paper (probably due to log-level operations and some modifications), and I cannot fully follow it... (I'm sorry)

src/nets/e2e_asr_attctc.py
takaaki-hori: We need to keep both scores since CTC scores cannot be computed as "alpha(h) = alpha(g) + log P(c|g)" I agree with the idea keeping individual scores if overhead is not a problem.
Another possible way is that we keep only ctc_score in addition to the score, and add "ctc_weight * (ctc_score - previous_ctc_score)" to the score.

sw005320: OK. Then, if we'll not have some overheads, I prefer to keep both scores, but it's up to you.

egs/wsj/asr1/run.sh
```diff
@@ -50,6 +50,7 @@ beam_size=20
 penalty=0
 maxlenratio=0.8
 minlenratio=0.3
+ctc_weight=0.0
```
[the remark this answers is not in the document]
takaaki-hori: OK, I will put the best weight (maybe 0.3) as a default.
Is it okay to set minlenratio=0.0 and maxlenratio=0.0 (i.e. end detection) ?
The result is slightly better than min=0.3 & max=0.8 and equivalent to min=0.0 & max=1.0.

src/nets/ctc_prefix_score.py
```diff
@@ -0,0 +1,78 @@
+#!/usr/bin/env python
+
+# Copyright 2018 Mitsubishi Electric Research Labs (Takaaki Hori)
+#  Apache 2.0  (http://www.apache.org/licenses/LICENSE-2.0)
+
+import numpy as np
+import six
+
+
```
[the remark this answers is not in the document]
takaaki-hori: I agree. I will do that.

src/nets/ctc_prefix_score.py
takaaki-hori: Actually it is a little different. The code computes the scores for a set of next labels simultaneously while Algorithm 2 computes the score only for one next label. I will add some comments on this difference.

src/nets/ctc_prefix_score.py
```diff
@@ -0,0 +1,78 @@
...
+        '''Compute CTC prefix scores for next labels
+
+        :param y     : prefix label sequence
+        :param cs    : array of next labels
+        :param r_prev: previous CTC state
+        :return ctc_scores, ctc_states
+        '''
+        # initialize CTC states
+        output_length = len(y) - 1  # ignore sos
+        r = self.xp.ndarray((self.input_length, 2, len(cs)), dtype=np.float32)
```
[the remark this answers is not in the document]
takaaki-hori: OK, I will do that.

src/nets/ctc_prefix_score.py
```diff
@@ -0,0 +1,78 @@
...
+            r[0, 0] = xs[0]
+            r[0, 1] = self.logzero
+        else:
+            r[output_length - 1] = self.logzero
+
+        # prepare forward probabilities for the last label
+        r_sum = self.xp.logaddexp(r_prev[:, 0], r_prev[:, 1])  # log(r_t^n(g) + r_t^b(g))
+        last = y[-1]
+        if output_length > 0 and last in cs:
+            log_phi = self.xp.ndarray((self.input_length, len(cs)), dtype=np.float32)
```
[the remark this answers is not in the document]
takaaki-hori: Thanks. I agree.

egs/wsj/asr1/run.sh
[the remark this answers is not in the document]
sw005320: This is no problem. 
In my LM integration I still use min=0.3, but use max=0.0. With CTC scores, min=0.3 may not be required.

sw005320:
Are you ready for the merge?
Can you update RESULTS with your new setup?

takaaki-hori:
@sw005320 I just resolved the conflicts and errors for the merge. But there still remain some parts I should modify according to your review. Please wait a moment.
Regarding RESULTS file, since I don't have the same baseline with yours, it is better that you do it with your model.

sw005320:
OK. Actually, the WSJ LSTLM part takes long time, and we might need to consider to prepare two recipes that includes the LSTMLM or not.

sw005320:
LGTM. Can I merge it?

takaaki-hori:
Yes!

src/nets/e2e_asr_attctc_th.py
```diff
@@ -740,13 +797,20 @@ def recognize_beam(self, h, recog_args, char_list):
...
-                local_scores = functional.log_softmax(self.output(z_list[-1]), dim=1).data
-                local_best_scores, local_best_ids = torch.topk(local_scores, beam, dim=1)
+                local_att_scores = functional.log_softmax(self.output(z_list[-1]), dim=1).data
+                if lpz is not None:
+                    local_best_scores, local_best_ids = torch.topk(
+                        local_att_scores, ctc_beam, dim=1)
+                    ctc_scores, ctc_states = ctc_prefix_score(hyp['yseq'], local_best_ids[0], hyp['ctc_state_prev'])
+                    local_scores = \
+                        (1.0 - ctc_weight) * local_att_scores[:, local_best_ids[0]] \
+                        + ctc_weight * torch.from_numpy(ctc_scores - hyp['ctc_score_prev'])
```
geniki: @takaaki-hori if I understand it correctly, the paper you reference does not have `- hyp['ctc_score_prev']` element (line 11 of Algorithm 1). Could you please let me know whether there is indeed a difference in this implementation and whether it makes any difference?

30stomercury: > @takaaki-hori if I understand it correctly, the paper you reference does not have `- hyp['ctc_score_prev']` element (line 11 of Algorithm 1). Could you please let me know whether there is indeed a difference in this implementation and whether it makes any difference?

Hi, did you got the answer? I would like to know the intuition behind `- hyp['ctc_score_prev']`. It looks like this line wants to reduce the scare of ctc_scores..

PR #74, opened 2018-01-19
modify asr_train_th.py close to the chainer version

sw005320:


sw005320:
It's duplicated with #75. Close it.

PR #102, opened 2018-02-14
Refactoring scripts

kan-bayashi:
- both backends use `asr_train.py` and `asr_recog.py`
- shared function and class are integrated into `asr_utils.py`
- backend dependent processing is written in `asr_pytorch.py` and `asr_chainer.py`


kan-bayashi:
Now almost finished.
If you have any suggestion or question, please let me know.

egs/ami/asr1/run.sh
```diff
@@ -198,20 +198,16 @@ else
...
-# switch backend
-if [[ ${backend} == chainer ]]; then
-    train_script=asr_train.py
-    decode_script=asr_recog.py
-else
-    train_script=asr_train_th.py
-    decode_script=asr_recog_th.py
-fi
+# set train and decode script
+train_script=asr_train.py
```
sw005320: We don't have to use ${train_script} and ${decode_script} any more.

kan-bayashi:
I also waver about that.
I think`src/bin` should contain only executable scripts (`asr_train.py`, `asr_recog.py`, `lm_train.py`).
The other scripts (`asr_{utils, chainer, pytorch}.py`) should be located in other directory.

kan-bayashi:
Now I moved `asr_{chainer,pytorch,utils}.py` into `src/asr`.
And also I added `src/asr` to `PYTHONPATH`.

sw005320:
@takaaki-hori, @ShigekiKarita, Can you take a look, since this is a kind of big changes (but will not cause big issues)?

takaaki-hori:
@sw005320 I don't find any big problems in the new code. It is nice that we no longer have to modify or add options for each DL framework in the top-level python scripts. Since the core part src/nets has not changed at all, I believe the changes will not cause big issues.

PR #107, opened 2018-02-21
Multi-GPU support for PyTorch

bobchennan:
Multi-GPU support for PyTorch. It doesn't affect existing examples.

### DIFF
--- a/tools/Makefile
+++ b/tools/Makefile
@@ -32,7 +32,7 @@ venv/lib/python2.7/site-packages/torch: venv/bin/activate
 warp-ctc: venv/lib/python2.7/site-packages/torch
 	git clone https://github.com/SeanNaren/warp-ctc.git
 	. venv/bin/activate; cd warp-ctc && git checkout 9e5b238f8d9337b0c39b3fd01bbaff98ba523aa5 && mkdir build && cd build && cmake .. && make -j4
-	. venv/bin/activate; pip install cffi && 
+	. venv/bin/activate; pip install cffi
 	. venv/bin/activate; cd warp-ctc/pytorch_binding && python setup.py install # maybe need to: apt-get install python-dev
 
 chainer_ctc: venv/bin/activate


sw005320:
At least, please fix to pass travis check. See https://travis-ci.org/espnet/espnet/jobs/344545414

egs/tedlium/asr1/run.sh
```diff
@@ -3,13 +3,18 @@
 # Copyright 2017 Johns Hopkins University (Shinji Watanabe)
 #  Apache 2.0  (http://www.apache.org/licenses/LICENSE-2.0)
 
+#Multi-GPU configuration
+export num_gpu = 1 # number of gpu cores to be used
+gpu=-1         # use 0 when using GPU on slurm/grid engine, use "[0, 1]" when you want to use 2 cores, otherwise -1
```
kan-bayashi: Is it possible to integrate `num_gpu` and `gpu` ?

egs/tedlium/asr1/cmd.sh
```diff
@@ -11,10 +11,10 @@
 # or search for the string 'default_config' in utils/queue.pl or utils/slurm.pl.
 
 export train_cmd="run.pl --mem 2G"
-export cuda_cmd="run.pl --mem 2G --gpu 1"
+export cuda_cmd="run.pl --mem 2G --gpu $num_gpu"
 export decode_cmd="run.pl --mem 4G"
```
kan-bayashi: How about the overwrite in run.sh with `--gpuP option for `$cuda_cmd` ?
For example, `${cuda_cmd} --gpu ${num_gpu} ${expdir}/train.log`

bobchennan: In this case we need to update all the egs examples?

sw005320: Yes, this is a good point, but just focusing on this recipe for now.

egs/tedlium/asr1/run.sh
```diff
@@ -3,13 +3,15 @@
 # Copyright 2017 Johns Hopkins University (Shinji Watanabe)
 #  Apache 2.0  (http://www.apache.org/licenses/LICENSE-2.0)
 
+#Multi-GPU configuration
+gpu=-1         # use 0 when using GPU on slurm/grid engine, use "0,1" when you want to use 2 cores, otherwise -1
+export num_gpu=`awk -F"," '{if($0=="-1")print 0;else print NF}' <<< "${gpu}"` # number of gpu cores to be used
 . ./path.sh
```
sw005320: It would be better to print the GPU information, when we execute `run.sh`.

kan-bayashi:
I checked it works.
The rest concern is that whether to modify `cmd.sh`.
In the case of keeping the original `cmd.sh`, we just modify `run.sh` as follows:
```bash
if [ ${stage} -le 3 ]; then
    echo "stage 3: Network Training"
    num_gpu=`awk -F"," '{if($0=="-1")print 0;else print NF}' <<< "${gpu}"`
    ${cuda_cmd} --gpu ${num_gpu} ${expdir}/train.log \
```
@sw005320 Which is the suitable ?

egs/tedlium/asr1/run.sh
ShigekiKarita: I agree. Maybe using only num_gpu is the simplest solution. In addition, I prefer num_gpu=0 for CPU computation and use `model.cuda(device_ids=range(num_gpu))` because the environment variable `CUDA_VISIBLE_DEVICES` can handle [the device id mappings](http://acceleware.com/blog/cudavisibledevices-masking-gpus) and some job tools (slurm, torque, or JHU /usr/local/bin/free-gpu) specify CUDA_VISIBLE_DEVICES automatically. 

I propose to use
- num_gpu for specifying the number of GPUs (0 for CPU)
- CUDA_VISIBLE_DEVICES defined at the outside of run.sh for specifying the GPU ids

ShigekiKarita:
@kan-bayashi 
>  Is it possible to integrate num_gpu and gpu ?
A little bit confusing.


I agree. Maybe using only num_gpu is the simplest solution. In addition, I prefer num_gpu=0 for CPU computation and use `if num_gpu > 0: model.cuda(device_ids=range(num_gpu))` 

For setting device ids, I'd like to use the environment variable `CUDA_VISIBLE_DEVICES` because it can handle [the device id mappings](http://acceleware.com/blog/cudavisibledevices-masking-gpus). As you know, some job tools (slurm, torque, or JHU /usr/local/bin/free-gpu) set CUDA_VISIBLE_DEVICES automatically. 

I propose to use
- num_gpu for setting the number of GPUs (0 for CPU)
- CUDA_VISIBLE_DEVICES defined at the outside of run.sh for setting the GPU ids

bobchennan:
Yes it would be easier to use num_gpu directly with CUDA_VISIBLE_DEVICES. The only problem is that we need to update all cmd.sh.

sw005320:
Most of users do not follow this PR, and this change causes some confusions, but I think this is a necessary change.
These are action items that I come up with to minimize the confusions:

- [ ] update `README.md`
- [ ] update `cmd.sh` and `run.sh` in all recipes
- [ ] for now, we add a particular error message in `run.sh` to guide users to use `--ngpu` when they use `--gpu`. Then, later, we'll delete this error message.

Can someone also work on the chainer backend?

sw005320:
1. For `cmd.sh`, I like @kan-bayashi's solution of using 
`${cuda_cmd} --gpu ${num_gpu} ${expdir}/train.log`
Then, we can avoid to change `cmd.sh`,. 
This command level control at `run.sh` is more intuitive.

2. For `path.sh`, I still like to avoid to accept `$ngpu`, as I want to keep `cmd.sh` and `path.sh` functions similar to the original Kaldi interface. Also this is only in the JHU setup, and I want to avoid to change `path.sh` interface only because of our setup. My idea is to move the following lines in `path.sh`
```
if [[ $(hostname -f) == *.clsp.jhu.edu ]] ; then
    export CUDA_VISIBLE_DEVICES=$(/usr/local/bin/free-gpu)
    export CUDA_VISIBLE_DEVICES=$(/usr/local/bin/free-gpu -n $num_gpu)
fi
```
to `run.sh`
This makes `run.sh` bit ugly, but can keep the interface of `path.sh` same. Also controlling this in `run.sh` is again more intuitive.

@bobchennan do you have any idea, especially for 2? I also want to ask @kan-bayashi and/or @ShigekiKarita about it.

kan-bayashi:
I implemented chainer version (#109). 
If the format is fixed, I can able to merge.

bobchennan:
I found an bug. When we generate train.sh for network training, it looks like this:

> #!/bin/bash
> cd /export/b14/nchen/espnet/egs/fisher_swbd/asr1
> . ./path.sh

This is generated by queue.pl. In this case CUDA_VISIBLE_DEVICES will be exported again. However in train.sh we don't have ngpu setting. So we need to change either queue.pl or path.sh.

sw005320:
@bobchennan can you make the resume function as a separate PR? If this is tightly related to the multiple GPU stuff, this is no problem, and please continue to develop with this PR.

shrutijpalaskar:
Will multi-gpu for pytorch be merged into the master branch soon?

sw005320:
Yes, @bobchennan almost finished the implementation, but needs some modifications. It will be done in the next week.

sw005320:
@bobchennan can you restart this?

bobchennan:
Yes, definitely.

PR #109, opened 2018-02-24
Chainer multi gpu

kan-bayashi:
Multi gpu implementation with chainer


kan-bayashi:
I change from `ParallelUpdater` to `MultiProcessParallelUpdater`.
It seems to be working well with high GPU utils.
But the implementation is a little bit complicated.
@sw005320 @ShigekiKarita
If you have ideas to clean up codes, please let met know.

kan-bayashi:
@sw005320 @ShigekiKarita  I implemented following functions:
- warn the use of `--gpu` option in `run.sh`, `asr_train.py`, `asr_recog.py`, `lm_train.py`
- if `--gpu` option is used, `--ngpu` value will be automatically updated
- if use gpu, check whether `CUDA_VISIBLE_DEVICES` is set in `asr_train.py`, `asr_recog.py`, `lm_train.py`
- if the number of gpus is not matched with `CUDA_VISIBLE_DEVICES`, show warning message 
- if the number of gpus > 1, `--ngpu` will be automatically 1 in LM training

src/bin/asr_train.py
```diff
@@ -134,6 +136,24 @@ def main():
...
+            args.ngpu = 1
+
+    # check CUDA_VISIBLE_DEVICES
+    if args.ngpu > 0:
+        cvd = os.environ.get("CUDA_VISIBLE_DEVICES")
+        if cvd is None:
+            logging.warn("CUDA_VISIBLE_DEVICES is not set.")
+        elif args.ngpu != len(cvd.split(",")):
+            logging.warn(cvd)
+            logging.warn(len(cvd.split(",")))
```
ShigekiKarita: maybe `logging.error` is better?

src/asr/asr_chainer.py
```diff
@@ -215,31 +389,64 @@ def train(args):
...
     # prepare Kaldi reader
     train_reader = lazy_io.read_dict_scp(args.train_feat)
     valid_reader = lazy_io.read_dict_scp(args.valid_feat)
 
+    # set up training iterator and updater
+    if ngpu <= 1:
+        # make minibatch list (variable length)
+        train = make_batchset(train_json, args.batch_size,
+                              args.maxlen_in, args.maxlen_out, args.minibatches)
+        # hack to make batchsze argument as 1
```
ShigekiKarita: batchsze -> batchsize

src/asr/asr_chainer.py
```diff
@@ -215,31 +389,64 @@ def train(args):
...
+                                            args.maxlen_in, args.maxlen_out, args.minibatches)]
+
+        # each subset must have same length for MultiprocessParallelUpdater
+        maxlen = max([len(train_subset) for train_subset in train_subsets])
+        for train_subset in train_subsets:
+            if maxlen != len(train_subset):
+                for i in six.moves.xrange(maxlen - len(train_subset)):
+                    train_subset += [train_subset[i]]
+
+        # hack to make batchsze argument as 1
```
ShigekiKarita: ditto

src/asr/asr_chainer.py
```diff
@@ -215,31 +389,64 @@ def train(args):
...
+            # make minibatch list (variable length)
+            train_subsets += [make_batchset(train_json_subset, args.batch_size,
+                                            args.maxlen_in, args.maxlen_out, args.minibatches)]
+
+        # each subset must have same length for MultiprocessParallelUpdater
+        maxlen = max([len(train_subset) for train_subset in train_subsets])
+        for train_subset in train_subsets:
+            if maxlen != len(train_subset):
+                for i in six.moves.xrange(maxlen - len(train_subset)):
+                    train_subset += [train_subset[i]]
```
ShigekiKarita: Does a shorter minibatch appear more than a longer minibatch? Is it possible to append empty list?

egs/csj/asr1/run.sh
```diff
@@ -195,9 +212,9 @@ mkdir -p ${expdir}
 
 if [ ${stage} -le 4 ]; then
     echo "stage 4: Network Training"
-    ${cuda_cmd} ${expdir}/train.log \
+    ${cuda_cmd} --gpu ${ngpu} ${expdir}/train.log \
```
sw005320: Just make sure that do we need some changes in `cmd.sh` to pass `--gpu ${ngpu}`?

src/asr/asr_chainer.py
```diff
@@ -194,12 +357,23 @@ def train(args):
         logging.info('ARGS: ' + key + ': ' + str(vars(args)[key]))
 
     # Set gpu
-    gpu_id = int(args.gpu)
-    logging.info('gpu id: ' + str(gpu_id))
-    if gpu_id >= 0:
+    ngpu = int(args.ngpu)
```
sw005320: may not have to cast it to `int()`

egs/csj/asr1/run.sh
kan-bayashi: In the case of slurm.pl, we do not have to modify `cmd.sh`.

src/asr/asr_chainer.py
kan-bayashi: I've tried but it causes error when the empty batch is used.

sw005320:
What is the progress?
CLSP people will have a problem for this, but I think @bobchennan's solution of adding the CLSP-dependent commands to `run.sh` to get the CUDA_VISBLE_DEVICE is a reasonable solution.

kan-bayashi:
I will add CLSP dependent processing about CLSP.

Are warning messages suitable for you?
And If there is no problem, I will update all of the recipe.

sw005320:
The warning part looks fine.
Please go ahead.

kan-bayashi:
@sw005320 Could you test some recipe in CLSP?

sw005320:
I'm testing, but failed due to 
```
    'NCCL is not enabled. MultiprocessParallelUpdater '
Exception: NCCL is not enabled. MultiprocessParallelUpdater requires NCCL.
Please reinstall chainer after you install NCCL.
(see https://github.com/chainer/chainer#installation).
```

sw005320:
It seems that it's working.
We may need to have some documentation about NCCL (I just got it from the nvidia site also @bobchennan told me where it is). Also we need to remove the following lines in `path.sh`
```
if [[ $(hostname -f) == *.clsp.jhu.edu ]] ; then
    export CUDA_VISIBLE_DEVICES=$(/usr/local/bin/free-gpu)
fi
```

kan-bayashi:
@sw005320 Thank you for your testing.
I updated the use of gpu option and added explanation about NCCL in README.

sw005320:
I think this PR becomes a good shape. @bobchennan can you take a look at it? If it is no problem, I'll merge it. Then, I want to ask you to commit your pytorch multiple GPU function.

bobchennan:
Looks fine for me. But to use it on clsp grid I think we need to modify queue.pl to support multi-gpu.

sw005320:
What part do you need to modify? Since I just used a default `queue.pl` and it worked, but it might be just lucky.

bobchennan:
Yeah you are right. I checked queue.pl in kaldi and it should work. No need for modification.

sw005320:
OK. @kan-bayashi, is there any other issues or things you left? If no, I can merge it.

src/asr/asr_chainer.py
```diff
@@ -215,31 +389,64 @@ def train(args):
...
+            # make minibatch list (variable length)
+            train_subsets += [make_batchset(train_json_subset, args.batch_size,
+                                            args.maxlen_in, args.maxlen_out, args.minibatches)]
+
+        # each subset must have same length for MultiprocessParallelUpdater
+        maxlen = max([len(train_subset) for train_subset in train_subsets])
+        for train_subset in train_subsets:
+            if maxlen != len(train_subset):
+                for i in six.moves.xrange(maxlen - len(train_subset)):
+                    train_subset += [train_subset[i]]
```
kan-bayashi: @sw005320 my last concern is this part. 
If It is no problem, it can be merged.

PR #114, opened 2018-03-05
Pytorch lm training script

kan-bayashi:


ShigekiKarita:
@kan-bayashi unfortunately pytorch dropout function takes `training=False` argument. use `torch.nn.Dropout` to reflect `model.train()` or `model.eval()` 

http://pytorch.org/docs/master/_modules/torch/nn/functional.html#dropout
http://pytorch.org/docs/master/nn.html#dropout

ShigekiKarita:
Honestly, I've already implemented lm_train_th.py but your code looks better :+1: . let me paste my test code. maybe no need to change
https://gist.github.com/ShigekiKarita/3b172fd8c11d21fcd01b2a260c7e42ec

kan-bayashi:
The results are as follows:
```
# WSJ 
eval = 7.4
dev = 8.5

# WSJ with ctc joint decoding 
eval = 6.4
dev = 8.2

# WSJ with ctc joint decoding + lm integration
eval = 5.5
dev = 6.8
```

Training speed is not so different from chainer.
```
chainer = 10 hours
pytorch = 8.5 hours
```

ShigekiKarita:
I think it is reasonable. I have chainer equivalent results
```
# WSJ blstmp ctc joint decoding
eval = 6.3
dev = 8.1

# WSJ blstmp ctc joint decoding + lm integration
eval = 6.1
dev = 7.1
```
it might be faster with `nn.LSTM` (w/ cudnn) instead of `nn.LSTMCell`  (w/o cudnn) but I think the current speed is enough.

ShigekiKarita:
The test can pass when I set same values (1) to weight like
``` python
def transfer_lstm(ch_lstm, th_lstm):
    ch_lstm.upward.W.data[:] = 1
    th_lstm.weight_ih.data[:] = torch.from_numpy(ch_lstm.upward.W.data)
    ch_lstm.upward.b.data[:] = 1
    th_lstm.bias_hh.data[:] = torch.from_numpy(ch_lstm.upward.b.data)
    ch_lstm.lateral.W.data[:] = 1
    th_lstm.weight_hh.data[:] = torch.from_numpy(ch_lstm.lateral.W.data)
    th_lstm.bias_ih.data.zero_()
```
it indicates that internal order of weights and bias is quite different between chainer and pytorch. I will try to implement the transfer function on an hour.

ShigekiKarita:
I almost forget that the forget gate bias placement is different at least
- `f=bias[n//4:n//2]` https://github.com/pytorch/pytorch/issues/750
- `f=bias[n//2:-n//4]` https://github.com/chainer/chainer/blob/b2f3ac457fe2139aab3c4408dc56c85b8a00c65e/chainer/links/connection/lstm.py#L30

these changes are significantly hard to solve. we should wait for onnx.

ShigekiKarita:
I think everything looks good to me unless you wanna add some lines to `run.sh`.

src/bin/lm_train_th.py
```diff
@@ -0,0 +1,437 @@
...
+        # Evaluation routine to be used for validation and test.
+        model.predictor.eval()
+        state = {
+            'c1': model.predictor.zero_state(args.batchsize),
+            'h1': model.predictor.zero_state(args.batchsize),
+            'c2': model.predictor.zero_state(args.batchsize),
+            'h2': model.predictor.zero_state(args.batchsize)
+        }
+        if args.gpu >= 0:
+            for key in state.iterkeys():
```
ShigekiKarita: use `keys` instead of `iterkeys` that is unavailable in python3 and not memory consuming in this case

src/bin/lm_train_th.py
```diff
@@ -0,0 +1,437 @@
...
+            x = Variable(torch.from_numpy(batch[:, 0]).long(), volatile=True)
+            t = Variable(torch.from_numpy(batch[:, 1]).long(), volatile=True)
+            if args.gpu >= 0:
+                x = x.cuda(args.gpu)
+                t = t.cuda(args.gpu)
+            state, loss = model(state, x, t)
+            sum_perp += loss.data
+            if data_count % bproplen == 0:
+                # detach all states
+                for key in state.iterkeys():
```
ShigekiKarita: ditto

src/bin/lm_train_th.py
```diff
@@ -0,0 +1,437 @@
...
+    epoch_now = 0
+    best_valid = 100000000
+    state = {
+        'c1': model.predictor.zero_state(args.batchsize),
+        'h1': model.predictor.zero_state(args.batchsize),
+        'c2': model.predictor.zero_state(args.batchsize),
+        'h2': model.predictor.zero_state(args.batchsize)
+    }
+    if args.gpu >= 0:
+        for key in state.iterkeys():
```
ShigekiKarita: ditto

src/bin/lm_train_th.py
```diff
@@ -0,0 +1,437 @@
...
+            # Compute the loss at this time step and accumulate it
+            state, loss_batch = model(state, x, t)
+            loss += loss_batch
+            count += 1
+
+        sum_perp += loss.data
+        model.zero_grad()  # Clear the parameter gradients
+        loss.backward()  # Backprop
+        # detach all states
+        for key in state.iterkeys():
```
ShigekiKarita: ditto

ShigekiKarita:
pytorch was little bit faster in my environment too (python3)
- chainer on 1080ti took 64 sec / 100 iter
- pytorch on 1080ti took 44 sec / 100 iter
- pytorch on v100 took 32 sec / 100 iter
(chainer could not run on V100 by strange cupy errors)

kan-bayashi:
Thank you, Shigeki.
Now I understand the reason why both outputs are different.
And I changed script structure to be the same as `asr_train.py`.
LM integration is available for both backend :)

test/test_lm.py
```diff
@@ -0,0 +1,71 @@
...
+def test_lm():
+    n_vocab = 3
+    n_units = 2
+    batchsize = 5
+    rnnlm_ch = lm_chainer.ClassifierWithState(lm_chainer.RNNLM(n_vocab, n_units))
+    rnnlm_th = lm_pytorch.ClassifierWithState(lm_pytorch.RNNLM(n_vocab, n_units))
+    transfer_lm(rnnlm_ch.predictor, rnnlm_th.predictor)
+    import numpy
+    # test transfer function
+    # numpy.testing.assert_equal(rnnlm_ch.predictor.embed.W.data, rnnlm_th.predictor.embed.weight.data.numpy())
```
sw005320: Why these are commented out?

src/nets/e2e_asr_attctc_th.py
```diff
@@ -1695,6 +1695,13 @@ def recognize(self, h, recog_args):
         for l in six.moves.range(1, self.dlayers):
             c_list.append(self.zero_state(h.unsqueeze(0)))
             z_list.append(self.zero_state(h.unsqueeze(0)))
+        if rnnlm:
+            state = {
```
sw005320: It's better to make this as a method in the rnnlm.
Then, if we change the number of layers, etc., we don't have to change this part.
(chainer part as well....)

test/test_lm.py
ShigekiKarita: We gave up on transferring weights from chainer to pytorch until onnx. It is OK to add `# TODO(karita) implement weight  transfer`.

src/nets/e2e_asr_attctc_th.py
ShigekiKarita: I agree. RNNLM.forward should support `state=None` for the initial step.

ShigekiKarita:
@kan-bayashi you need to merge #116 or you can ignore CI error until warpctc issue #115  is solved.

kan-bayashi:
Thanks!

ShigekiKarita:
please, update README.md on master to indicate that pytorch supports LM

kan-bayashi:
OK. I will add.

sw005320:
Can you also reflect this to the new CHiME-5 recipe?

sw005320:
Thanks.
If there is no issue, I'll merge it.

PR #119, opened 2018-03-15
[WIP] added augment data functionality

arendu:
example recipe in 'egs/wsj/aug_example/run.sh'


sw005320:
@arendu, do you think you can refine this PR (at least pass the travis check?)? I think this extension is important for the workshop. However, the current problem of this PR is that it is still a research code.

PR #120, opened 2018-03-16
added multichannel end-to-end setup

TsubasaOchiai:


sw005320:
Thanks a lot for your contribution!!!
Before checking the code deeply, I'd like to ask you to solve the following:
1.  Please at least fix your codes to pass travis check
2. Use symlink for chime4 data preparation scripts rather than adding copies
3. `path.sh` and `run.sh` should be changed to deal with recent LM change #114 
4. better to use `lazy_io`. `kaldi_io_py` has a lot of issues, and we decided not to use it. @ShigekiKarita will help if you need to modify `lazy_io`
5. please add installation of `scipy` `in tools/Makefile` to see this would work with a standard building process.

ShigekiKarita:
We only use `lazy_io` for reading scp/ark lazily for lower memory consumption. For writing, it does not need to be lazy.

TsubasaOchiai:
@sw005320 @ShigekiKarita 
Thank you for your comments.
I have a question about lazy_io and kaldi_io_py.

In my current implementation, I only used kaldi_io_py.readmat to read cmvn and Mel matrices.
They are read only once and registered in a model class.
As far as I understand, they are written in single kaldi matrix format (neither scp nor ark) and lazy_io do not support such file format.
In this situation, is it a problem to use kaldi_io_py.readmat?

I'd really appreciate it if you could give me your opinion.

ShigekiKarita:
> They are read only once and registered in a model class.

I think lazy reader won't improve this usage because the model stores it anyway.

> single kaldi matrix format (neither scp nor ark) and lazy_io do not support such file format.

Yes. There is no function like that.  I have never seen that a single matrix occupies memory.

> In this situation, is it a problem to use kaldi_io_py.read_mat?

No problem.

TsubasaOchiai:
@ShigekiKarita 
Thank you for your quick response.
I will keep the current implementation about kaldi_io_py.read_mat for now.

ShigekiKarita:
@TsubasaOchiai which python version are you mainly using for this PR?

src/asr/asr_mc_chainer.py
```diff
@@ -0,0 +1,516 @@
...
+
+        data[1]['feat'] = feat
+
+    return batch
+
+
+def me2e_converter_recog(name, readers, mode):
+    # noisy mode
+    if mode == 'noisy':
+        bidim = readers[name].shape[1] / 2
```
ShigekiKarita: use `//` instead of `/` because the result is `float` in python3

TsubasaOchiai:
@ShigekiKarita I mainly use python2 for my implementation and I am not sure that it works in python3. Unfortunately, because I am not familiar with python2-3 compatibility, I cannot judge by myself whether it is difficult or not. Can you give me some advices to fulfill the compatibility?

ShigekiKarita:
We do not have such a compatibility guidelines. Now we just write some tests (e.g., running with dummy args, checking expected outputs) and run them on TravisCI with python2 and 3.

egs/chime4/asr1_mc/conf/chime4.cfg
```diff
@@ -0,0 +1,50 @@
+#BeamformIt sample configuration file for AMI data (http://groups.inf.ed.ac.uk/ami/download/)
```
sw005320: Does it require beamformit?
I think you can remove this and also beamformit parts in `run.sh`.

src/nets/me2e_asr_attctc.py
```diff
@@ -0,0 +1,204 @@
+#!/usr/bin/env python
```
sw005320: better to rename it as `e2e_asr_mc_attctc.py`

egs/chime4/asr1_mc/local
```diff
@@ -0,0 +1 @@
+../asr1/local
```
sw005320: better to use own local directory (not symlink), but use a symlink for each file.
But if you don't use any local files for this recipe, this solution is OK.

egs/chime4/asr1_mc/run.sh
```diff
@@ -0,0 +1,354 @@
...
+        --opt ${opt} \
+        --epochs ${epochs} \
+        --melmat ${melmat} \
+        --cmvn ${cmvn} \
+        --mode ${trainmode}
+fi
+
+if [ ${stage} -le 4 ]; then
+    echo "stage 4: Decoding"
+#    nj=32
```
sw005320: Please remove the comment.

egs/chime4/asr1_mc/conf/chime4.cfg
TsubasaOchiai: I do not change chime4.cfg from asr1 setup.

src/nets/me2e_asr_attctc.py
TsubasaOchiai: I see.
I will change it later.

egs/chime4/asr1_mc/local
TsubasaOchiai: Currently, I do not use any local file.

egs/chime4/asr1_mc/run.sh
TsubasaOchiai: I see.
I will modify it later.

egs/chime4/asr1_mc/conf/chime4.cfg
sw005320: My point is whether you require BeamformIt in this recipe (`run.sh`).
If you need BeamformIt, you can keep this, but otherwise, you can remove this.

sw005320: I know that you call BeamformIt in `run.sh`. Again if you need the BeamformIt processing, you can keep this file and the BeamformIt processing in `run.sh`. Otherwise remove both.

TsubasaOchiai: I see. I think I can remove BeamformIt. I just called it in run.sh because it was called in asr1's run.sh.

src/asr/asr_mc_chainer.py
```diff
@@ -0,0 +1,516 @@
...
+# rnnlm
+import lm_train
+
+# numpy related
+import matplotlib
+import numpy as np
+matplotlib.use('Agg')
+
+
+class ChainerSeqEvaluaterKaldi(extensions.Evaluator):
```
ShigekiKarita: Is this class completely same to the one in `asr_chainer.py`? If it is true, import and use it.

src/asr/asr_mc_chainer.py
```diff
@@ -0,0 +1,516 @@
...
+
+            summary.add(observation)
+
+        return summary.compute_mean()
+
+
+class ChainerSeqUpdaterKaldi(training.StandardUpdater):
+    '''Custom updater with Kaldi reader for chainer'''
+
+    def __init__(self, train_iter, optimizer, reader, device, converter):
```
ShigekiKarita: instead of this redefinition, how about adding a new last argument `converter=converter_kaldi`  in `asr_chainer.ChainerSeqUpdaterKaldi`

src/asr/asr_mc_chainer.py
```diff
@@ -0,0 +1,516 @@
...
+    # enhancement mode
+    elif mode == 'enhan':
+        bidim = readers[0][name].shape[1] / 2
+        # separate real and imaginary part
+        feat_real = [reader[name][:, :bidim] for reader in readers]
+        feat_imag = [reader[name][:, bidim:] for reader in readers]
+    else:
+        logging.error(
+            "Error: need to specify an appropriate decoding mode")
+        sys.exit()
```
ShigekiKarita: sys.exit() returns 0 that means success but it is error. so you need to change this line like
- `sys.exit(1)`
- `sys.exit("Error: need to specify an appropriate decoding mode")`

BTW I found that there are similar wrong usages in ESPnet...

src/nets/me2e_asr_attctc.py
```diff
@@ -0,0 +1,204 @@
...
+            # make a utt list (1) to use the same interface for encoder
+            hs = {}
+            hs['real'] = [h_real]
+            hs['imag'] = [h_imag]
+
+            # 1. beamformer
+            hs = self.enhan(hs)
+        else:
+            logging.error(
+                "Error: need to specify an appropriate training mode")
```
ShigekiKarita: use `raise NotImplementedError("unknown training mode %s" % recog_args.mode)` not to go on.

src/bin/asr_mc_train.py
```diff
@@ -0,0 +1,200 @@
...
+    parser.add_argument('--epochs', '-e', default=30, type=int,
+                        help='Number of maximum epochs')
+    parser.add_argument('--grad-clip', default=5, type=float,
+                        help='Gradient norm threshold to clip')
+    # me2e
+    parser.add_argument('--melmat', type=str, required=True,
+                        help='Filename of Mel-filterbank matrix data (Kaldi ark)')
+    parser.add_argument('--cmvn', type=str, required=True,
+                        help='Filename of cmvn statistics data (Kaldi ark)')
+    parser.add_argument('--mode', default='noisy+enhan', type=str,
```
ShigekiKarita: `--mode` is not informative name. maybe `--utt-type` is better? @sw005320 @TsubasaOchiai

src/bin/asr_mc_recog.py
```diff
@@ -0,0 +1,105 @@
...
+                        help='Input length ratio to obtain min output length')
+    parser.add_argument('--ctc-weight', default=0.0, type=float,
+                        help='CTC weight in joint decoding')
+    # rnnlm related
+    parser.add_argument('--rnnlm', type=str, default=None,
+                        help='RNNLM model file to read')
+    parser.add_argument('--lm-weight', default=0.1, type=float,
+                        help='RNNLM weight.')
+    # me2e
+    parser.add_argument('--mode', default=None, type=str, required=True,
```
ShigekiKarita: ditto

src/asr/asr_mc_chainer.py
TsubasaOchiai: That's right.
In my implementation, in order not to change the original asr_chainer, I just redefined it.
If you can add a new argument "converter=converter_kaldi" in the original Updater, we can just import and use it.

TsubasaOchiai: In addition, we have to add "if statement" in _sum_sqnorm function.
Because the multichannel training sometimes passes the beamformer module, the gradient become None.
Therefore, we have to ignore such None gradient.

Is such modification OK for the original asr_chainer?

src/asr/asr_mc_chainer.py
TsubasaOchiai: I see.
I will modify it.

src/nets/me2e_asr_attctc.py
```diff
@@ -0,0 +1,204 @@
...
+        ys = [self.xp.array(
+            list(map(int, i[1]['tokenid'].split())), dtype=np.int32) for i in data]
+        ys = [chainer.Variable(y) for y in ys]
+
+        # 2. feature extractor
+        hs = self.logmel(hs, self.xp)
+        hs = self.cmvn(hs, self.xp)
+
+        # 3. encoder
+        hs, ilens = self.asr.enc(hs, ilens)
```
ShigekiKarita: if following lines are completely same in `E2E.__call__`, you can separate those lines in a separated method like
``` python
    def __call__(self, data):
        '''E2E forward
        :param data:
        :return:
        '''
        # utt list of frame x dim
        xs = [i[1]['feat'] for i in data]
        # remove 0-output-length utterances
        tids = [d[1]['tokenid'].split() for d in data]
        filtered_index = list(filter(lambda i: len(tids[i]) > 0, range(len(xs))))
        if len(filtered_index) != len(xs):
            logging.warning('Target sequences include empty tokenid (batch %d -> %d).' % (
                len(xs), len(filtered_index)))
        xs = [xs[i] for i in filtered_index]
        # utt list of olen
        ys = [self.xp.array(np.fromiter(map(int, tids[i]), dtype=np.int32))
              for i in filtered_index]
        ys = [chainer.Variable(y) for y in ys]

        # subsample frame
        xs = [xx[::self.subsample[0], :] for xx in xs]
        ilens = self.xp.array([xx.shape[0] for xx in xs], dtype=np.int32)
        hs = [chainer.Variable(self.xp.array(xx, dtype=np.float32))
              for xx in xs]
        return self.loss(hs, ilens, ys)

    def loss(self, hs, ilens, ys)
        # 1. encoder
        hs, ilens = self.enc(hs, ilens)

        # 3. CTC loss
        loss_ctc = self.ctc(hs, ys)

        # 4. attention loss
        loss_att, acc, att_w = self.dec(hs, ys)
        return loss_ctc, loss_att, acc
```

src/nets/me2e_asr_attctc.py
```diff
@@ -0,0 +1,204 @@
...
+
+        # utt list of olen
+        ys = [self.xp.array(
+            list(map(int, i[1]['tokenid'].split())), dtype=np.int32) for i in data]
+        ys = [chainer.Variable(y) for y in ys]
+
+        # 2. feature extractor
+        hs = self.logmel(hs, self.xp)
+        hs = self.cmvn(hs, self.xp)
+
```
ShigekiKarita: If you assume `self.asr.subsample[0] == 1` here, add `assert self.asr.subsample[0] == 1` in `__init__`.

src/nets/me2e_asr_attctc.py
```diff
@@ -0,0 +1,204 @@
...
+            hs = self.enhan(hs)
+        else:
+            logging.error(
+                "Error: need to specify an appropriate training mode")
+
+        # 2. feature extractor
+        hs = self.logmel(hs, self.xp)
+        hs = self.cmvn(hs, self.xp)
+
+        with chainer.no_backprop_mode(), chainer.using_config('train', False):
```
ShigekiKarita: ditto (see a comment in `__call__`)

src/bin/asr_mc_train.py
TsubasaOchiai: Now, I wonder whether I remove this option or not.
Because this option is made for experimental purpose, I think it is not needed for a stable version...
I would like to know your opinion.

src/asr/asr_mc_chainer.py
ShigekiKarita: I think those modifications are OK.

src/bin/asr_mc_train.py
ShigekiKarita: I always prefer simpler codes. If you think so, remove them.

src/nets/me2e_asr_attctc.py
ShigekiKarita: or a kind of `NotImplementedError`

src/nets/me2e_asr_attctc.py
TsubasaOchiai: Basecally, the differences between E2E and ME2E are data preparation and beamforming computation parts.
After those processes, the code is completely same.
However, because data preparation part is different, we cannot call E2E's __call__ function.
If E2E has just network computing function like the above loss, we can just import and use it...

src/nets/me2e_asr_attctc.py
TsubasaOchiai: I have check again, but I think we can conduct subsample in the asr's first layer.
(Although we cannot conduct subsample in beamforming network.)

src/nets/me2e_asr_attctc.py
ShigekiKarita: OK, I know. If you are not comfortable with this change, you can keep it.

src/bin/asr_mc_train.py
TsubasaOchiai: Thank you for your opinion.
I will remove the option later.

src/asr/asr_mc_chainer.py
TsubasaOchiai: I see.
Can I modify it by myself?
Or, should I ask someone to modify it?

src/nets/me2e_asr_attctc.py
TsubasaOchiai: Sorry, did you propose the modification of the original E2E function (__call__ and loss)?
If you can modify so, I can just call E2E.loss in the ME2E's __call__ function.

ShigekiKarita: Yes that is right. I want you to modify `E2E` as you know.

sw005320:
Are there any updates? I want to use it for the chime-5 challenge.

ShigekiKarita:
I am reworking this PR nowadays (It is already much better CER but need to
merge recent changes). You can close this.

2018年8月27日(月) 23:06 stale[bot] <notifications@github.com>:

sw005320:
> I am reworking this PR nowadays (It is already much better CER but need to
merge recent changes).

Sounds very good.

I want to keep this PR open until you'll create a new PR.
This PR has a lot of useful implementations.

Rpersie:
Are there any updates? I run the code but can not reach the paper results
image

PR #124, opened 2018-03-28
add resume function to pytorch implementation

bobchennan:


sw005320:
Do you think we should also add optimizer resume function in addition to the model resume function? If it is easy, we could do it (also chainer).

bobchennan:
For pytorch it is not necessary since optimizer is stored in trainer.updater so I assume it will be resumed by serialization in 
`chainer.serializers.load_npz(args.resume, trainer)`

For chainer it should be the same.

PR #127, opened 2018-04-03
[WIP] hub4_spanish

m-wiesner:


sw005320:
This seems to be good for me.
If there is no issue, I'll merge it.

m-wiesner:
Also, just note that for hub4 Spanish there is no dev set so I just took 1000 utterances randomly from train and called that a dev set. It's about 3% of the data so I don't think it should affect performance much. Also, since there was no external text corpus (although I could find something if you wanted), I omitted the RNNLM part.

sw005320:
OK. I understood. As soon as you finish the initial experiment of the whole pipline, I'll merge it. We'll have some changes in the whole recipe scripts, and better to be merged. So you don't have to tune the parameters so much. Also the 3% dev set seems to be fine.

sw005320:
Is that everything for now?

PR #132, opened 2018-04-06
Docker support

Fhrozen:
This is a general implementation for running ESPnet on docker. 

- To keep backends separated, each image will be implemented with its corresponded backend. 

- Chainer  backend implemented

- Pytorch (WIP)

### DIFF
--- a/README.md
+++ b/README.md
@@ -76,6 +76,20 @@ $ ./run.sh --gpu 0
 ```
 Default setup uses CPU (`--gpu -1`).
 
+### Docker Container
+If you use docker, execute `run.sh` located inside the docker folder. It will build the container to execute the given egs.
+The minimal execution line is detailed as following: 
+```sh
+$ cd $ESPnet_Folder/docker
+$ ./run.sh --gpu 0 --backend chainer --egs voxforge --stage 0'
+```
+For some egs which required to use external corpus, you can use the option ` --corpus_dir $MY_CORPUS_DIR` to load the folder into the container.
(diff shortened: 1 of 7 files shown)


sw005320:
I have two comments
- Better to mention this in the `README.md` in the main directory.
- `Makefile` and `requirements.txt` seem to be duplicated. Since this would be often changed, it would be better to use the original ones in `tools/`

Also, please let me know once PyTorch implementation is finished.

Fhrozen:
`requirement.txt` should be no a problem. But in case of `Makefile` not sure if this will be possible  because of the use of venv. For a docker container does not seems necessary to implement a venv rather this is already isolated. But, I will try to do the necessary changes on it.

I will be implementing the PyTorch by this weekend

sw005320:
OK.
But if it is not harmful, and we have a benefit that we don't have to maintain two `Makefile`, why not using venv for docker?
I think this is easier solution.

sw005320:
@ShigekiKarita @kan-bayashi, what do you think?
I'm not a docker person, and I'm not sure so much. It might break the docker concept, but I don't think it is big deal to use venv in docker.

Fhrozen:
Just modified the makefile, let me know if this is acceptable so I can upload the others files depending on the acceptance. BTW, because of the use of only one makefile the docker image will have both backends to avoid modifying the makefile even more.

Fhrozen:
Updates:
- Readme file includes user instructions for docker
- docker build employs makefile from tools folder and includes support for both backends (chainer & pytorch)
- deleted && after pip install cffi, (This breaks the build execution in docker)

PR #136, opened 2018-04-11
Docker support

Fhrozen:
Docker implementation with support on both backends (pytorch and chainer)


docker/espnet.devel
```diff
@@ -0,0 +1,57 @@
...
+        sox \
+        && \
+    apt-get clean && \
+    rm -rf /var/lib/apt/lists/*
+
+RUN curl -fSsL -O https://bootstrap.pypa.io/get-pip.py && \
+    python get-pip.py && \
+    rm get-pip.py
+
+RUN pip install numpy ipython \
```
sw005320: You don't need to install many of them, as their installations are included in Makefile (also they would not be used as now we use venv.

docker/espnet.devel
```diff
@@ -0,0 +1,57 @@
...
+                h5py scipy \
+                matplotlib scikit-image \
+                scikit-learn colorama virtualenv
+
+RUN mkdir -p /espnet/tools
```
sw005320: I'm not sure why you copy them.
I thought that we can still use the original directory structure.

Fhrozen: I copy to speed up the building process. If there is additional information inside one egs folder, it will take time to buffer and build. Also, to isolate the tools built from the container.

docker/espnet.devel
Fhrozen: Sure, I will delete this additionals

sw005320:
Remove https://github.com/espnet/espnet/blob/master/egs/voxforge/asr1/run_in_docker.sh

PR #147, opened 2018-04-26
fix pytorch cpu case

sw005320:


bobchennan:
Oh I think it should be:
```
gpu_id = -1
```
In one function len(gpu_id) is used to detect number of gpus.

sw005320:
This actually caused an error when `len(gpu_id)`. How do I fix it?

bobchennan:
Yeah I think it is fine to use ```[-1]``` since ```num_gpu=1``` so
```
            loss.backward()  # Backprop
```
will be executed. This is the right choice for cpu case.

PR #149, opened 2018-04-28
Fix beamsearch

sw005320:
- remove greedy search code.
- fix cupy import


sw005320:
@m-wiesner, can you review my code as @kan-bayashi may not be available due to a vacation week in Japan?

m-wiesner:
Yeah. I'll take a look.

________________________________

PR #151, opened 2018-04-30
Babel Recipe

m-wiesner:
Initial babel recipe. It seems to work. I have not yet added in the RNNLM, nor tuned any parameters much.


sw005320:
@takaaki-hori @karaf , can you check this recipe could work in MERL and BUT, respectively?

egs/babel/asr1/README.txt
```diff
@@ -0,0 +1,22 @@
...
+To run the experiment do 
+
+cd ../expname;
+
+To specify the BABEL langauges in training refer to them by their language id.
+See  conf/lang.conf for the exhaustive list of languages and corresponding
+language ids.
+
+Examples:
+./run --langs "102" --recog "102" -- trains on 102 (assamese) and tests on 102.
```
sw005320: run -> run.sh

egs/babel/asr1/cmd.sh
```diff
@@ -0,0 +1,20 @@
...
+# commands one by one: most recipes will exhaust the memory on your
+# machine).  queue.pl works with GridEngine (qsub).  slurm.pl works
+# with slurm.  Different queues are configured differently, with different
+# queue names and different ways of specifying things like memory;
+# to account for these differences you can create and edit the file
+# conf/queue.conf to match your queue's configuration.  Search for
+# conf/queue.conf in http://kaldi-asr.org/doc/queue.html for more information,
+# or search for the string 'default_config' in utils/queue.pl or utils/slurm.pl.
+
+#export train_cmd="run.pl --mem 2G"
```
sw005320: Changed to use run.pl as default.
Just comment out lines of queue.pl

egs/babel/asr1/local/prepare_data.sh
```diff
@@ -0,0 +1,107 @@
...
+# Copyright 2018 Johns Hopkins University (Matthew Wiesner)
+#  Apache 2.0  (http://www.apache.org/licenses/LICENSE-2.0)
+
+# This is not necessarily the top-level run.sh as it is in other directories.   see README.txt first.
+
+. ./conf/lang.conf
+. ./path.sh
+. ./cmd.sh
+
+FLP=true
```
sw005320: What is FLP and LLP?

egs/babel/asr1/conf/lang.conf
```diff
@@ -0,0 +1,292 @@
...
+# be changed appropriately for other users
+
+# Cantonese
+train_data_dir_101=/export/babel/data/101-cantonese/release-current/conversational/training
+train_data_list_101=./conf/lists/101-cantonese/train.LimitedLP.list
+train_data_dir_101_FLP=/export/babel/data/101-cantonese/release-current/conversational/training
+train_data_list_101_FLP=./conf/lists/101-cantonese/train.FullLP.list
+dev10h_data_dir_101=/export/babel/data/101-cantonese/release-current/conversational/dev
+dev10h_data_list_101=./conf/lists/101-cantonese/dev.list
+lexicon_file_101=/export/babel/data/101-cantonese/release-current/conversational/reference_materials/lexicon.sub-train1.txt
```
sw005320: Are lexicon related configs required?

egs/babel/asr1/local/prepare_data.sh
m-wiesner: I made it a little more flexible than it needed to be maybe. FLP is (Full language pack), the 40, 60, or 80h training set provided. LLP is limited language pack and it is always 10h.

egs/babel/asr1/conf/lang.conf
m-wiesner: They aren't needed, but I figured it wouldn't hurt to include them since they were already there, and someone might decide to use them maybe in the future.

m-wiesner: Sorry. Actually, I am currently using it in the recipe to prepare the data/train/text file. I don't use the pronunciations, but I do use the words in order to filter out word fragments.

egs/babel/asr1/local/prepare_data.sh
sw005320: OK. You can just put that information somewhere (README?)

egs/babel/asr1/conf/lang.conf
sw005320: OK. Keep it then.

sw005320:
@m-wiesner, any update?
@takaaki-hori, can you take a look at the code to see whether it is working at MERL?

sw005320:
@m-wiesner and @takaaki-hori, can you deal with this? (sorry for asking again, but this recipe is important for our workshop, in other words, this work is becoming a bottleneck in our current baseline preparation...)

m-wiesner:
I've added those parts we discussed as well as the RNNLM part to the recipe. I guess we just have to worry about the LDC distribution.

sw005320:
@m-wiesner, now travis-ci is fixed by #160.
Can you test it?

sw005320:
Merged this PR, but I want @takaaki-hori and @karaf to test the scripts in their environments as soon as possible.

sw005320:
Please carefully check everything, and fix them as soon as possible. This will affect all our baseline studies.

sw005320:
@m-wiesner, I think you intend to do that, right? Can you tell me again the reason?

m-wiesner:
Hi,


I used the same case normalization as in the standard BABEL recipe (in kaldi) so that I could compare both systems. But I agree that we should probably just uppercase everything. Otherwise we have more training targets than we need. I think this just using this command inserted as the last step of stage 0 (data prep) should make the desired change.


paste -d' ' <(cut -d' ' -f1 data/train/text) <(cut -d' ' -f2- data/train/text | LC_ALL= sed 's/./\U&/g') > data/train/text.upper

________________________________

sw005320:
Thanks. I just make sure that if we use the sclite, I thought that we ignore the upper/lower cases, and the score is the same anyway. So, we can compare it. It is wrong?

karaf:
sclite is by default case insensitive, just need to be sure that your coding is correctly. For most of the languages en_US.UTF-8 is fine. Turkish had some problem so you have to set correctly tr_TR.utf-8 to have correct conversion. Obviously to have same train/test case is the most safe.

sw005320:
> Turkish had some problem so you have to set correctly tr_TR.utf-8 to have correct conversion.
@m-wiesner, can you confirm it?

PR #152, opened 2018-04-30
Chainer Attention Multihead

Fhrozen:
Implementing attention multihead to chainer 
Test (WIP)


sw005320:
Are you implementing everything that @kan-bayashi did for the pytorch backend? (That is great. I'm just confirming)

sw005320:
OK. I think this is fine. Please let me know after you finish experiments and they're working.

PR #153, opened 2018-05-01
add support to newer gcc (>4.9)

bobchennan:


tools/requirements_cuda91.txt
```diff
@@ -0,0 +1,4 @@
+chainer>=3.0.0
+cupy-cuda91>=2.0.0
```
sw005320: In my setup, cupy>=2.0.0 is fine.

bobchennan: let me try it

bobchennan: Yes you are right

sw005320:
Can you also add about this Makefile to README?

PR #156, opened 2018-05-04
update a json file format for multiple input/output

TsubasaOchiai:


TsubasaOchiai:
I executed the following git commands "git fetch upstream" and "git merge
upstream/master".
It said that "Already up-to-date".
I think that I already pulled the latest master branch...
Is this correct understanding?
(I am not sure about git usage...)

2018-05-05 12:23 GMT+09:00 Shinji Watanabe <notifications@github.com>:

src/bin/asr_train.py
```diff
@@ -41,14 +41,10 @@ def main():
     parser.add_argument('--verbose', '-V', default=0, type=int,
                         help='Verbose option')
     # task related
-    parser.add_argument('--train-feat', type=str, required=True,
```
sw005320: For now, it's better to add error messages (ask the user to redo the "stage 2: Dictionary and Json Data Preparation") instead of removing these four options. Later, we can remove these options.

src/asr/asr_utils.py
```diff
@@ -10,19 +10,21 @@
...
 from chainer import training
 
+# io related
+import kaldi_io_py
 
 # * -------------------- training iterator related -------------------- *
 def make_batchset(data, batch_size, max_length_in, max_length_out, num_batches=0):
     # sort it by input lengths (long to short)
     sorted_data = sorted(data.items(), key=lambda data: int(
-        data[1]['ilen']), reverse=True)
```
sw005320: Same comments with the options. For now, it's better to add error messages (ask the user to redo the "stage 2: Dictionary and Json Data Preparation") instead of removing these four options. Later, we can remove these options.
You don't have to add the error message everywhere. Only this part should be fine.

src/asr/asr_chainer.py
```diff
@@ -52,12 +52,11 @@
 
 
 class ChainerSeqEvaluaterKaldi(extensions.Evaluator):
```
sw005320: Now you don't have to pass "reader" and you can use built-in `chainer.training.extensions.Evaluator` instead of making our own Evaluator.

src/asr/asr_chainer.py
```diff
@@ -96,12 +95,11 @@ def evaluate(self):
 
 
 class ChainerSeqUpdaterKaldi(training.StandardUpdater):
```
sw005320: Same comment. Now you don't have to pass "reader" and you can use built-in `chainer.training.updaters.StandardUpdater` instead of making our own Updater.

src/asr/asr_chainer.py
```diff
@@ -146,12 +144,11 @@ def _sum_sqnorm(self, arr):
 
 
 class ChainerMultiProcessParallelUpdaterKaldi(training.updaters.MultiprocessParallelUpdater):
```
sw005320: Probably same. Now you don't have to pass "reader" and you can use built-in `chainer.training.updaters.MultiprocessParallelUpdater` instead of making our own Updater. But I want ask @kan-bayashi for this help.

src/asr/asr_utils.py
TsubasaOchiai: I have a question.
add error messages to asr_utils.py?
not asr_recog.py?

sw005320: `asr_recog.py` will be called with your new data preparation, and we would not require the error message. Some confusion will be happened in the training, and I think the message in options could be sufficient, but if we have the error message, it would be further helpful for users.

sw005320: I am assuming that many people usually skip the "stage 2: Dictionary and Json Data Preparation", since it requires to be called only once. Then, if they pull this change, and perform their experiments with `--stage 3`, the code does not work and it causes some confusions. With the error message in my suggestion, we can avoid confusions for this case.

TsubasaOchiai: I agree with your idea, i.e. we should add error messages in asr_train.py.
However, I cannnot grasp your comment.
What is the options in asr_utils.py?

Do you mean that, in make_batchset, we should add error messages when data[1] does not have 'input' or 'output' keys?

src/asr/asr_chainer.py
TsubasaOchiai: Now, I am checking "chainer.training.updaters.StandardUpdater" code.
It seems almost same as "ChainerSeqUpdaterKaldi" as you said.
As far as I checked, the differences are the following: (1) check whether grad norm is nan and (2) delete_feat.
Is it OK to use "chainer.training.extensions.Evaluator"?

sw005320:
> (1) check whether grad norm is nan and (2) delete_feat.
> Is it OK to use

I see. These are actually important, and we may still need to use our original ones.

sw005320:
This PR is done by #211

Issue #154, opened 2018-05-04
unexpected keyError during decoding


chiayuli:
Hi all,
I'm the beginner for ESpnet and I followed the instructions from ESpnet/egs/wsj/asr1/run.sh
the training of language model and acoustic model looks fine, 

export CUDA_VISIBLE_DEVICES=0 ; ./run.sh --ngpu 1 --backend pytorch --etype blstmp 

but I face problem during decoding... the following message is from decode.2.log
Does anyone face the same problem ? Any suggestion will be appreciated. Thanks.

2018-05-04 15:26:04,898 (asr_recog:97) INFO: python path = /mount/arbeitsdaten/asr/licu/Espnet/egs/wsj/asr1/../../../src/lm/:/mount/arbeitsdaten/asr/licu/Espnet/egs/wsj/asr1/../../../src/asr/:/mount/arbeitsdaten/asr/licu/Espnet/egs/wsj/asr1/../../../src/nets/:/mount/arbeitsdaten/asr/licu/Espnet/egs/wsj/asr1/../../../src/utils/:/mount/arbeitsdaten/asr/licu/Espnet/egs/wsj/asr1/../../../src/bin/:/mount/arbeitsdaten/asr/licu/Espnet/egs/wsj/asr1/../../../src/lm/:/mount/arbeitsdaten/asr/licu/Espnet/egs/wsj/asr1/../../../src/asr/:/mount/arbeitsdaten/asr/licu/Espnet/egs/wsj/asr1/../../../src/nets/:/mount/arbeitsdaten/asr/licu/Espnet/egs/wsj/asr1/../../../src/utils/:/mount/arbeitsdaten/asr/licu/Espnet/egs/wsj/asr1/../../../src/bin/:/mount/arbeitsdaten/asr/licu/Espnet/egs/wsj/asr1/../../../src/lm/:/mount/arbeitsdaten/asr/licu/Espnet/egs/wsj/asr1/../../../src/asr/:/mount/arbeitsdaten/asr/licu/Espnet/egs/wsj/asr1/../../../src/nets/:/mount/arbeitsdaten/asr/licu/Espnet/egs/wsj/asr1/../../../src/utils/:/mount/arbeitsdaten/asr/licu/Espnet/egs/wsj/asr1/../../../src/bin/:
2018-05-04 15:26:04,898 (asr_recog:102) INFO: set random seed = 1
2018-05-04 15:26:04,898 (asr_recog:105) INFO: backend = pytorch
[...]

PR #157, opened 2018-05-05
fix problem of dataparallel and resume function

bobchennan:
#154


sw005320:
This commit has some issues at the travis check

bobchennan:
@sw005320 fixed

Issue #133, opened 2018-04-06
New development plan


sw005320:
I'm now managing several development plans upon requests.
Please feel free to ask us any functions that you feel it is beneficial for your research.
What I got requests so far are

- Fix VGG-BLSTM in PyTorch
- Generalize to accept multiple input/output streams
- Back compatibility
- Add unidirectional LSTM layer in the encoder
- Output attention weights 
- Output more detailed decoder information (e.g., each score at every output time step) for post processwing
- Scheduled sampling to handle exposure bias 
- Completely switch to pure CTC or seq2seq by mtlalpha = 0.0 or 1.0
- end of sentence norm and length norm
- Tensorboard X visualization for pytorch and chainer (https://github.com/lanpa/tensorboard-pytorch ???)

sw005320:
We (@arendo and @m-wiesner) are planning to add speech translation recipes.

hirofumi0810:
@sw005320 

I'm interested in the extension of espnet for speech translation tasks.
So I want to add new recipes and develop some modules.

sw005320:
Thanks. I'm also planning it. Let's do it :)

kan-bayashi:
Move on to #334

Issue #115, opened 2018-03-10
warpctc issue


sw005320:
It seems that warpctc is not correctly installed, which makes travis check failed.

ShigekiKarita:
maybe we have to fix the warpctc version by git submodule. breaking change
of warpctc happens too frequently for us. do you agree?

2018/03/10 11:55 "Shinji Watanabe" <notifications@github.com>:

sw005320:
Yes. BTW, I used to use hash to manage this kind of problems, but git submodule has more benefit?

ShigekiKarita:
Do you use hash only for detecting change? Git submodule is used for git
clone & checkout the specific commit.

2018/03/10 12:07 "Shinji Watanabe" <notifications@github.com>:

ShigekiKarita:
If you mean 'hash' as git commit id, git submodule is same basically.

2018/03/10 12:12 "Karita Shigeki" <shigekikarita@gmail.com>:

sw005320:
Yes, I'm talking about SHA1 hash

ShigekiKarita:
BTW we should follow the latest warpctc version because it fixed the length / batch normalization that we patched in espnet.

r9y9:
What errors are you seeing? It would be nice to report it to https://github.com/SeanNaren/warp-ctc/issues.

ShigekiKarita:
@r9y9 there are two errors
- python2 cannot import warpctc https://travis-ci.org/espnet/espnet/jobs/351600502#L843
- our patched normalization conflicts to the warpctc_pytorch update https://travis-ci.org/espnet/espnet/jobs/351600503

the first one might related to your PR? https://github.com/SeanNaren/warp-ctc/pull/27

r9y9:
@ShigekiKarita Thanks, I was thinking I fixed a py2 issue... I will take a look later today for the first one.

r9y9:
@ShigekiKarita Sorry, that was my fault. The import error should be fixed by https://github.com/SeanNaren/warp-ctc/pull/31

ShigekiKarita:
@r9y9 great! we will fix our normalization issue.

r9y9:
@ShigekiKarita https://github.com/SeanNaren/warp-ctc/pull/31 PR was merged. Please let me know if it doesn't work.

ShigekiKarita:
I will update warp-ctc to HEAD after merging Makefile in #125

PR #165, opened 2018-05-08
Adding pure CTC or attention mode

sw005320:
Switch to use pure CTC or attention mode upon requests at #133. Now it is switched by specifying `--mtlalpha 0.0` (pure attention) or `--mtlalpha 1.0` (pure CTC).

### DIFF
--- a/egs/chime5/asr1/run.sh
+++ b/egs/chime5/asr1/run.sh
@@ -27,7 +27,6 @@ elayers=6
 eunits=320
 eprojs=320
 subsample=1_2_2_1_1 # skip every n frame from input to nth layers
-# loss related
 ctctype=warpctc
 # decoder related
 dlayers=1
@@ -246,7 +245,7 @@ if [ ${stage} -le 3 ]; then
 fi
 
 if [ -z ${tag} ]; then
-    expdir=exp/${train_set}_${etype}_e${elayers}_subsample${subsample}_unit${eunits}_proj${eprojs}_ctc${ctctype}_d${dlayers}_unit${dunits}_${atype}_aconvc${aconv_chans}_aconvf${aconv_filts}_mtlalpha${mtlalpha}_${opt}_bs${batchsize}_mli${maxlen_in}_mlo${maxlen_out}
+    expdir=exp/${train_set}_${etype}_e${elayers}_subsample${subsample}_unit${eunits}_proj${eprojs}_d${dlayers}_unit${dunits}_${atype}_aconvc${aconv_chans}_aconvf${aconv_filts}_mtlalpha${mtlalpha}_${opt}_bs${batchsize}_mli${maxlen_in}_mlo${maxlen_out}
     if [ "${lsm_type}" != "" ]; then
         expdir=${expdir}_lsm${lsm_type}${lsm_weight}
     fi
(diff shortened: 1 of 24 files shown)


sw005320:
Also changing warpctc as default in chainer by following the discussion in #115

sw005320:
@arendu and @m-wiesner, now I'm finishing only CTC or attention mode. Please let me know if you have any comments.

src/nets/e2e_asr_attctc.py
```diff
@@ -75,7 +75,12 @@ def __call__(self, x):
...
         alpha = self.mtlalpha
-        self.loss = alpha * loss_ctc + (1 - alpha) * loss_att
+        if alpha == 0:
+            self.loss = loss_att
+        elif alpha == 1:
+            self.loss = loss_ctc
+        else:
+            self.loss = alpha * loss_ctc + (1 - alpha) * loss_att
 
         if self.loss.data < CTC_LOSS_THRESHOLD and not math.isnan(self.loss.data):
```
kan-bayashi: Is it necessary to report ctc or att loss in pure ctc or att mode?

egs/chime5/asr1/run.sh
```diff
@@ -27,7 +27,6 @@ elayers=6
 eunits=320
 eprojs=320
 subsample=1_2_2_1_1 # skip every n frame from input to nth layers
-# loss related
 ctctype=warpctc
```
kan-bayashi: may be removed?

src/asr/asr_chainer.py
```diff
@@ -344,6 +336,16 @@ def train(args):
...
 
+    # specify attention, CTC, hybrid mode
+    if args.mtlalpha == 1.0:
+        mtl_mode = 'ctc'
+        logging.info('Pure CTC mode')
+    elif args.mtlalpha == 0.0:
+        mtl_mode = 'att'
+        logging.info('Pure attention mode')
+    else:
+        mtl_mode = 'mtl'
```
kan-bayashi: It is also better to add logging.

src/nets/e2e_asr_attctc.py
sw005320: This requires a lot of changes including several if-else, and I want to keep the reporter part simple. Also, it is not bad that all modes have the same format.

kan-bayashi: OK.

sw005320:
> My rest concern is that acc.best is not available in pure ctc mode.
> Is it difficult to calculate accuracy using greedy ctc decoder?

This is not easy. Probably, one of the future work.

shrutijpalaskar:
@sw005320 what would be a good way to monitor pure CTC training? As the accuracy is not calculated, how should we decide learning rate decay and convergence etc.?

Also, to confirm, for pure CTC decoding, the only change necessary is that ctc_weight should be set to 1.0, correct?

sw005320:
> @sw005320 what would be a good way to monitor pure CTC training? As the accuracy is not calculated, how should we decide learning rate decay and convergence etc.?

So far, we're just using the conditional likelihood (CTC loss) for monitoring the CTC training and learning rate decay.

> Also, to confirm, for pure CTC decoding, the only change necessary is that ctc_weight should be set to 1.0, correct?

You need to use both `--ctc_weight 1.0` and `--recog_model loss.best`

PR #175, opened 2018-05-14
librispeech run.sh with rnnlm (run_lm.sh)

JaejinCho:


sw005320:
Thanks for your PR!

1. Can you push form a brand rather than the master branch?
2. You can add the LM function to `run.sh` rather than preparing a new shell script.
3. Did you get some numbers with LM?

JaejinCho:
For my 1% portion experiment, I also used 1% of text for rnnlm training by mistake.
I am running using all the text now to look at how many epochs I need to the point that perplexity gets saturated. Let me check the epoch and set the rnnlm epoch accordingly. Then, I will PR new run.sh from a branch, not master branch while reporting the correct number.

Issue #161, opened 2018-05-07
Installation issues


sw005320:
ESPnet has several installation problems (many people pointed out).
#125 is one possible solution, and @jtrmal also proposed a shell script as follows

>  I have had a look at the Makefile and I'd suggest converting it to a shell script (or we should write a shell install script for each of the components). The way it's designed it's bound to generate tons of issues even in trivial cases like using python3

So, I'd like to ask your opinions, and want to fix this at least in a month for the workshop activity.

kan-bayashi:
I’m not sure current installation problems, but shell script based installation is easy to understand fro me.

sw005320:
I think we can at least check whether required modules are correctly imported after the installation by checking e.g., `python -c 'import chainer'`

sw005320:
Thanks @kan-bayashi and @jnishi. @jtrmal, could you initiate to consider the shell based installation?

sw005320:
Thanks @jtrmal 

The installation problems we had so far in addition to @jnishi's points are:

1. when our environments are updated (gcc 4.X -> gcc 6.X, cuda8.0 -> cuda9.1), warp_ctc cannot be correctly installed (the test is failed), @bobchennan is that correct?
2. the installation of `conda.mk` is failed when pytorch is updated from 0.31 to 0.40, again due to warp_ctc

I'll add others once I came up with. I can easily forget many things.
@m-wiesner, you also had some installation problems, right?

bobchennan:
Yes for warp-ctc test_gpu will fail due to incompatibility. But we can ignore that error and continue the installation process.

pozhijisi:
I am trying to install ESPnet with a (full) anaconda environment and will provide the installation recipe if succeeded. I know virtualenv is used in many places in ESPnet, but the installation experience with anaconda may be much better.

jtrmal:
I don't think we should support too many ways how to install packages. In that case, I'd suggest offloading the work to the user, asking him/her to install the packages using the system-preferred installer. There is simply too much variability for us to be able to handle all of them well.
it can be either conda or venv or anything else, but just a single package manager. The packages we installed will be restricted to espnet tools anyway, so there is no danger (I think) to go for a single installation path.

pozhijisi:
[...]

PR #181, opened 2018-05-20
[WIP] Merging depending software repository.

jnishi:
related to #161 

Merge depending software by git subtree, and simplify installation Makefile.
(Sorry, this PR is still work in progress because I only checked this in voxforge, and do not still fixed  tools/Makefile.cuda91.gcc6.)
There seems to exist several pros. and cons., which are listed as follows.

Pros.
- Can prevent installation issue caused by corruptions of master versions of depending software.
- Can apply and manage modification to outside of espnet (for example, https://github.com/espnet/espnet/commit/67310e196de9fb6ee0053f4ce991d8f79d469e9d).
- Can simplify Makefile by avoiding downloading the repositories. This will also be helpful in the case of shell scripts.

Cons.
- Huge commit logs (mainly by Kaldi) also merged to this repository. (But if you want, these logs can be removed at merging other repositories.)
- It will be more complex for maintainers to update depending softwares, because of git subtree.

What do you think about managing other repositories that are depended on by espnet by git subtree.


jnishi:
close this PR because this approach is too complex.

PR #209, opened 2018-06-09
Long utterance removal in librispeech recipe 

mallidi:
Utterances with more than 3000 frames (30 seconds) is causing GPU out-of-memory err
This happened in c nodes in clsp grid


egs/librispeech/asr1/run.sh
```diff
@@ -127,10 +127,13 @@ if [ ${stage} -le 1 ]; then
     utils/combine_data.sh data/${train_set} data/train_clean_100 data/train_clean_360 data/train_other_500
```
sw005320: how about the following?
```
utils/combine_data.sh data/${train_set}_org data/train_clean_100 data/train_clean_360 data/train_other_500
utils/combine_data.sh data/${train_dev}_org data/dev_clean data/dev_other
```
Then, you don't have to use `copy_data_dir.sh` below

PR #211, opened 2018-06-12
Json

Fhrozen:
Updates from last commits.
Tests:
AN4 (Pytorch GPU, CPU)
Voxforge (Chainer MultiGPU) 

No error found when launching Single GPU in Voxforge, WSJ.


sw005320:
Did you finish everything? I think this is almost finished.

sw005320:
One concern is that we may merge `convertjson.py` and `mergejson.py`, but you can do that later when you extend this for multichannel input/output.

sw005320:
@Fhrozen, there are some redundancy and inconsistency of making a data.json in the recognition.
Can you consider the following changes?
1. perform `data2json.sh` for all recognition data in the stage 2 
2. make a new tool called `splitjson.py` in `src/utils`, and perform the data split in the recognition stage instead of using `split_data.sh` and `data2json.sh` for each split data.

PR #226, opened 2018-06-14
[WIP] Back-translation

kan-bayashi:
Current version is for pytorch v4. 
Cannot merge due to warp-ctc issue.


ramon-astudillo:
Hi @kan-bayashi 

> Cannot merge due to warp-ctc issue.

is this an error during install?. I notticed your `tools/Makefile` is different from current master HEAD and warp_ctc build does not work for me. Using the current Makefile install went fine.

Also I have been adding some missing imports and moving the `egs/ljspeech/tts1/local/text` utils to `src/utils` to solve an import problem in virtualenv. In case its useful it is here

https://github.com/ramon-astudillo/espnet/tree/bts_pytorch_v3

kan-bayashi:
Hi @ramon-astudillo ,

> is this an error during install?. I notticed your tools/Makefile is different from current master HEAD and warp_ctc build does not work for me. Using the current Makefile install went fine.

In our environment, using current Makefile will be failed. 
And the error means this issue <https://github.com/SeanNaren/warp-ctc/issues/46>.
But according to #250, this is not observed.

> Also I have been adding some missing imports and moving the egs/ljspeech/tts1/local/text utils to src/utils to solve an import problem in virtualenv. In case its useful it is here

Thank you!
In my case, this text cleaning code is only working with python3.
I'm wondering how to control python version...

kan-bayashi:
I separated PR #260, close this PR.

PR #227, opened 2018-06-14
Simplify Makefile

sw005320:
- change Makefile to unify CUDA version differences 
- set compiled Kaldi as an argument


sw005320:
@Fhrozen, can you check this Makefile would work in your environment?
@bobchennan, can you review my change?

Fhrozen:
Tried in a computer with Mint 18.04 & cuda 9.2:
I think that will be better to stop the make when CUDA_HOME is empty. the installation will continue until pytorch requires to obtain the version of cuda.
Also, there is no release of pytorch for cuda 9.2, so it will be better to check this option.

Fhrozen:
About pytorch, I installed it using the previous version without problem. Using the cu80 in cuda 9.2. But since there is no official release it crashes.

sw005320:
@Fhrozen, thanks. How about this modification?

Fhrozen:
Yes, this works. 
But, if you do not mind I may recommend something like this:

```
venv/lib/python2.7/site-packages/torch: venv/bin/activate
ifeq ($(CUDA_VERSION),92)
	$(info The current CUDA_VERSION is not supported. A downgraded version will be installed instead.)
	$(eval CUDA_VERSION := 91)
endif
	. venv/bin/activate; pip install pip --upgrade; pip install http://download.pytorch.org/whl/cu$(CUDA_VERSION)/torch-0.3.1-cp27-cp27mu-linux_x86_64.whl
```

Also in the begining of the file, something like this:
```
ifeq ($(CUDA_HOME),)
$(error CUDA_HOME is not set)
endif
```

may help.

sw005320:
I was also thinking of that, but pytorch with cu80 is actually working, and I also don't want to make our Makefile complicated. So, I like the current simple solution.

sw005320:
Thanks guys for your quick reviews.

Issue #230, opened 2018-06-15
RNNLM seems to be working in training mode when recognizing with PyTorch


takaaki-hori:
When using `asr_recog.py --backend pytorch`, the RNNLM module seems to be used in training mode, which means that the dropout function is still active even in the recognition phase. 
I think we need to put `rnnlm.eval()` somewhere. Can someone fix this and check the performance difference?

PR #231, opened 2018-06-15
fixed rmmlm to be eval mode in recognition

kan-bayashi:
#230


sw005320:
Thanks. Did you see some performance difference?

kan-bayashi:
I've just fixed. Sorry.

PR #237, opened 2018-06-18
[WIP] Batch processing of beam search

mn5k:
TODO
- combine scores of

- [x] attention decoder

- [x] RNNLM

- [x] CTC

### DIFF
--- a/src/asr/asr_pytorch.py
+++ b/src/asr/asr_pytorch.py
@@ -413,7 +413,6 @@ def recog(args):
         return new_state_dict
 
     model.load_state_dict(remove_dataparallel(torch.load(args.model, map_location=cpu_loader)))
-
     # read rnnlm
     if args.rnnlm:
         rnnlm = lm_pytorch.ClassifierWithState(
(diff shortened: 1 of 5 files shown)


egs/librispeech/asr1/run.sh
```diff
@@ -249,7 +249,7 @@ fi
 
 if [ ${stage} -le 5 ]; then
     echo "stage 5: Decoding"
-    nj=32
+    nj=1 #32
```
sw005320: I think it's better to use cpu for the original setup.
How about making a variable `rec_nj` in the beginning of the script, and we can change it in run.sh e.g., `./run.sh --stage 5 --rec_decode 1`

egs/librispeech/asr1/run.sh
```diff
@@ -268,7 +268,11 @@ if [ ${stage} -le 5 ]; then
         done
 
         #### use CPU for decoding
-        ngpu=0
```
sw005320: Similar comment to the above.
I think it's better to use cpu for the original setup.
How about making a variable `rec_ngpu` in the beginning of the script, and we can change it in run.sh e.g., `./run.sh --stage 5 --rec_ngpu 1`

egs/librispeech/asr1/run.sh
```diff
@@ -285,6 +289,7 @@ if [ ${stage} -le 5 ]; then
             --ctc-weight ${ctc_weight} \
             --rnnlm ${lmexpdir}/rnnlm.model.best \
             --lm-weight ${lm_weight} \
+	    --batchsize 64
```
sw005320: Same again.
`rec_batchsize`?

src/bin/asr_recog.py
```diff
@@ -59,6 +59,8 @@ def main():
                         help='Input length ratio to obtain min output length')
     parser.add_argument('--ctc-weight', default=0.0, type=float,
                         help='CTC weight in joint decoding')
+    parser.add_argument('--batchsize', default=1, type=int,
```
sw005320: `rec_batchsize`

src/asr/asr_pytorch.py
```diff
@@ -411,55 +411,81 @@ def remove_dataparallel(state_dict):
...
+        feats = [kaldi_io_py.read_mat(recog_json[name]['input'][0]['feat'])
+                 for name in names]
+
+        sorted_index = sorted(range(len(feats)), key=lambda i: -len(feats[i]))
+        names = [names[i] for i in sorted_index]
+        feats = [feats[i] for i in sorted_index]
+        y_true = [map(int, recog_json[name]['output'][0]['tokenid'].split())
+                  for name in names]
+
+        nbest_hyps = e2e.recognize_batch(feats, args, train_args.char_list, rnnlm=rnnlm)
```
sw005320: You should replace `e2e.recognize_batch` -> `e2e.recognize`. It is not easy to maintain two recognition code. Also make sure that the result becomes equivalent to the original when the beamsize = 1

src/nets/e2e_asr_attctc_th.py
```diff
@@ -354,6 +355,33 @@ def recognize(self, x, recog_args, char_list, rnnlm=None):
             self.train()
         return y
 
+    def recognize_batch(self, xs, recog_args, char_list, rnnlm=None):
```
sw005320: Same comment.
You should replace the original `recognize` to this `recognize_batch`

src/nets/e2e_asr_attctc_th.py
```diff
@@ -1897,6 +1967,212 @@ def recognize_beam(self, h, lpz, recog_args, char_list, rnnlm=None):
         # remove sos
         return nbest_hyps
 
+    def recognize_beam_batch(self, h, hlens, lpz, recog_args, char_list, rnnlm=None):
```
sw005320: Same comment.
You should replace the original `recognize_beam` to this `recognize_beam_batch`

src/nets/ctc_prefix_score.py
```diff
@@ -3,10 +3,142 @@
...
+        inputs = inputs.view(-1)
+        dim = 0
+    s, _ = torch.max(inputs, dim=dim, keepdim=True)
+    outputs = s + (inputs - s).exp().sum(dim=dim, keepdim=True).log()
+    if not keepdim:
+        outputs = outputs.squeeze(dim)
+    return outputs
+
+
+class BatchCTCPrefixScoreTH(object):
```
sw005320: Same.
`BatchCTCPrefixScoreTH(object)` -> `CTCPrefixScoreTH(object)`, and replace the original one with the batch one.

src/nets/e2e_asr_attctc_th.py
```diff
@@ -1897,6 +1967,212 @@ def recognize_beam(self, h, lpz, recog_args, char_list, rnnlm=None):
...
+        nbest_hyps = [[] for _ in range(batch)]
+
+        if lpz is not None:
+            ctc_prefix_score = BatchCTCPrefixScoreTH(lpz, 0, self.eos, beam)
+            ctc_states_prev = ctc_prefix_score.initial_state()
+            ctc_scores_prev = to_cuda(self, torch.Tensor(batch, n_bo).zero_())
+
+        # expansion
+        ended_hyps = [[] for _ in range(batch)]
+        exp_hlens = torch.LongTensor(hlens * beam).view(beam, batch).transpose(0, 1).contiguous()
```
ZhuoranLyu: This does not work for me. So I changed to 
`exp_hlens = torch.LongTensor(hlens).repeat(beam).view(beam, batch).transpose(0, 1).contiguous()`
and it works.
Not sure if this is the best approach.
Thx

src/nets/e2e_asr_attctc_th.py
```diff
@@ -354,6 +355,33 @@ def recognize(self, x, recog_args, char_list, rnnlm=None):
...
+        hs = [to_cuda(self, Variable(torch.from_numpy(xx), volatile=True))
+              for xx in xs]
+
+        # 1. encoder
+        # make a utt list (1) to use the same interface for encoder
+        xpad = pad_list(hs)
+        hpad, hlens = self.enc(xpad, ilens)
+
+        # calculate log P(z_t|X) for CTC scores
+        #if recog_args.ctc_weight > 0.0:
```
ZhuoranLyu: Shall I remove these comments to use hybrid CTC/Attention? @mn5k

mn5k: Combination of CTC score causes an error at some samples.
So, I recommend you to use the default beam search for hybrid CTC/Attention..

ZhuoranLyu: @mn5k But the default uses cpu to recognize, right? I'd like to use GPU to accelerate the recognize.

mn5k: Sure. I'll fix it asap.

ZhuoranLyu: BTW, this behaves differently on pytorch 0.4 and 0.3. When recognizing on pytorch 0.3, it only takes a little memory at around 500MB using batchsize = 1. While it takes 5000MB memory with pytorch 0.4. This is really strange. @mn5k

sw005320:
This PR is replaced with #318, and it is merged now.

Issue #239, opened 2018-06-18
Issue with PlotAttentionReport


ruizhilijhu:
I was trying to rerun from a snapshot. And I got the following error related to PlotAttentionReport. 

Any help is appreciated! 

Error message: 

Exception in main training loop: 'module' object has no attribute 'pyplot'
Traceback (most recent call last):
  File "/export/b11/ruizhili/espnet/tools/venv/local/lib/python2.7/site-packages/chainer/training/trainer.py", line 309, in run
    entry.extension(self)
  File "/export/b11/ruizhili/espnet/src/asr/asr_utils.py", line 176, in __call__
    self._plot_and_save_attention(att_w, filename.format(trainer))
  File "/export/b11/ruizhili/espnet/src/asr/asr_utils.py", line 186, in _plot_and_save_attention
    matplotlib.pyplot.imshow(att_w, aspect="auto")
Will finalize trainer extensions and updater before reraising the exception.
^[[JTraceback (most recent call last):
  File "/export/b11/ruizhili/espnet/egs/jsalt18e2e/asr1/../../../src/bin/asr_train.py", line 224, in <module>
    main()
  File "/export/b11/ruizhili/espnet/egs/jsalt18e2e/asr1/../../../src/bin/asr_train.py", line 218, in main
    train(args)
  File "/export/b11/ruizhili/espnet/src/asr/asr_pytorch.py", line 370, in train
    trainer.run()
  File "/export/b11/ruizhili/espnet/tools/venv/local/lib/python2.7/site-packages/chainer/training/trainer.py", line 320, in run
    six.reraise(*sys.exc_info())
  File "/export/b11/ruizhili/espnet/tools/venv/local/lib/python2.7/site-packages/chainer/training/trainer.py", line 309, in run
    entry.extension(self)
  File "/export/b11/ruizhili/espnet/src/asr/asr_utils.py", line 176, in __call__
    self._plot_and_save_attention(att_w, filename.format(trainer))
  File "/export/b11/ruizhili/espnet/src/asr/asr_utils.py", line 186, in _plot_and_save_attention
    matplotlib.pyplot.imshow(att_w, aspect="auto")
AttributeError: 'module' object has no attribute 'pyplot'

sw005320:
@kan-bayashi, do you have any idea? @ruizhilijhu said that it did not happen in the normal training. It just happened when we use a resume function.

kan-bayashi:
[...]

PR #240, opened 2018-06-19
fixed pyplot not found error in resuming

kan-bayashi:
#239


sw005320:
@ruizhilijhu can you test it?

ruizhilijhu:
@sw005320 Yes, it seems that it worked for me now.

PR #243, opened 2018-06-20
prefetch kaldi scp file.

jnishi:
Loading kaldi's scp file is too slow, but this can be done in parallel with forward and backward calculation of other data. This can be easily done by using `MultiprocessIterator`.

It becomes about 1.5x  faster with GTX 1080Ti x 2 by applying this modification. Though modification I pushed in this PR is only for multiple GPU with Chainer backend, it might be applied to Pytorch backend.

Would it be possible for you to apply this modification to pytorch backend, and single GPU case by referencing this PR? I think modifying `converter_kaldi` has a high impact on the ESPnet, but I am not familiar with the architecture of the ESPnet.

But this requires a few modification to Chainer, to avoid memory leak.
I requested to fix this problem to Chainer team.

https://github.com/chainer/chainer/issues/4946
https://github.com/chainer/chainer/pull/4972


ShigekiKarita:
I think [torch.utils.data.DataLoader](https://pytorch.org/docs/stable/data.html#torch.utils.data.DataLoader) can do the same thing with `num_workers` option. In my opinion, you do not have to care about pytorch.

jnishi:
I noticed that memory leak did't happen if it run with single GPU or CPU.

jnishi:
Loading data to GPU in multiprocessiterator seems to be difficult, however it can be done by using  multithreaditerator. It may be better to use multithreaditerator.

It may be problem that multithreaditerator does not support to prefetch multiple iterators which is supported in multiprocessiterator. I have requested to support it to Chainer community.
https://github.com/chainer/chainer/issues/5153

jnishi:
fixed to use multithreaditerator instead of multiprocessiterator. It is slower than multiprocessiterator, but still faster than serialiterator. It would be much better if you run ESPnet on the heavy I/O environment.

jnishi:
I requested the Chainer community to able to increase prefetching iterators, but it is rejected bacause of complexity of implementation.  https://github.com/chainer/chainer/issues/5153

However, memory leak issue that I noted first will be resolved because they will merge PR that adds `MultiprocessIterator` to restart process to clean up memories. 
https://github.com/chainer/chainer/pull/4972 I hope it will be released next or after the next version. So it is better to use `MultiprocessIterator` than `MultithreadIterator`.

First, I modified iterators to `MultiprocessIterator`, and then I modified to MutithreadIterator. Moreover, it is needed to be fixed to use `MultiprocessIterator` again.  I'm going to another PR and close this PR, because it is better to clean up commits.

jnishi:
This PR is refined in #323.

PR #249, opened 2018-06-21
add recipe to train 1000-speaker subset: jsalt18e2e/asr1: run_1000spk.sh

ruizhilijhu:
add recipe to train 1000-speaker subset: jsalt18e2e/asr1: run_1000spk.sh


sw005320:
Thanks.
I think it's better to include it in `run.sh` instead of separate scripts. Then, by using `run.sh --train_set tr_babel10_1000spk` is easier way for us to maintain the script. Please fix it.

PR #256, opened 2018-06-26
[WIP] Multi-I/O data.json creation

m-wiesner:
First pass stab at generating data.json with multiple inputs and outputs.

See run.sh for details.

Basically, for each type of input, we run data2json.sh, e.g. something like ...
 **data2json.sh output1 dict1 > data.output1.json
 data2json.sh output2 dict2 > data.output2.json**

 We then later combine the data.*.json using

**combine_multimodal_json.py data.json data.*.json**

We still need to modify the training scripts to properly accept the new data format.


sw005320:
@Fhrozen, please check to see whether this would not have some conflicts with your work.

Fhrozen:
@sw005320, i think that the implementations could not be used for all the recipes. It could be better to implement the combine_multimodal part in the local folder. As I showed before, the mergejson.py may be better to be implementeded for multi-I/O.

m-wiesner:
Honestly, I don't like the way the data.json file is created and I am doing a lot of redundant work the way I implemented it. But the mergejson.py doesn't do what we need it to do. There are a lot of things that are hardcoded in that script that need to be changed to properly support multiple input and output features/transcript types. The is also the case, though to a lesser degree for data2json.sh.

I tried to just add the minimal amount of code changes even if it is not the cleanest way to do this. Maybe we can talk today and you can explain how you saw imagine this working and I can work around that.

PR #260, opened 2018-06-30
TTS recipe

kan-bayashi:
Minimum codes for TTS recipe in pytorch v0.4.
Note that python3 is required for text cleaning

### DIFF
--- a/src/bin/bts_decode.py
+++ b/src/bin/bts_decode.py
@@ -1,0 +1,89 @@
+#!/usr/bin/env python
+
+# Copyright 2018 Nagoya University (Tomoki Hayashi)
+#  Apache 2.0  (http://www.apache.org/licenses/LICENSE-2.0)
+
+import argparse
+import logging
+import os
+import subprocess
+import sys
+
+
+def main():
+    parser = argparse.ArgumentParser()
+    # general configuration
+    parser.add_argument('--gpu', default=None, type=int, nargs='?',
+                        help='GPU ID (negative value indicates CPU)')
+    parser.add_argument('--ngpu', default=0, type=int,
+                        help='Number of GPUs')
+    parser.add_argument('--backend', default='pytorch', type=str,
+                        choices=['chainer', 'pytorch'],
+                        help='Backend library')
+    parser.add_argument('--out', type=str, required=True,
+                        help='Output filename')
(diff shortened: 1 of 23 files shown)


kan-bayashi:
@ramon-astudillo Cound you test this PR?

sw005320:
I have the following error
```
  File "local/clean_text.py", line 9, in <module>
    from cleaners import english_cleaners
  File "/export/a08/shinji/201707e2e/espnet_tts/egs/ljspeech/tts1/local/text/cleaners.py", line 18, in <module>
    from numbers import normalize_numbers
  File "/export/a08/shinji/201707e2e/espnet_tts/egs/ljspeech/tts1/local/text/numbers.py", line 11
SyntaxError: Non-ASCII character '\xc2' in file /export/a08/shinji/201707e2e/espnet_tts/egs/ljspeech/tts1/local/text/numbers.py on line 11, but no encoding declared; see http://python.org/dev/peps/pep-0263/ for details
```

ramon-astudillo:
Yes, the encoding declaration must be above the docstring in 

https://github.com/kan-bayashi/espnet/blob/tts/egs/ljspeech/tts1/local/text/numbers.py

kan-bayashi:
@sw005320 As @ramon-astudillo said, the problem is already fixed.
Did you use the latest version?

src/bin/bts_decode.py
```diff
@@ -0,0 +1,89 @@
...
+
+    # check gpu argument
+    if args.gpu is not None:
+        logging.warn("--gpu option will be deprecated, please use --ngpu option.")
+        if args.gpu == -1:
+            args.ngpu = 0
+        else:
+            args.ngpu = 1
+
+    # check CUDA_VISIBLE_DEVICES
```
sw005320: better to use lines from https://github.com/espnet/espnet/pull/260/files#diff-dbc60e85a2905c1803821c64a4c81cb0R147 to make it use in CLSP

sw005320:
> @sw005320 is there a policy with regard to this?

If this tool is used for other recipes, I think it should be OK. But I feel that the current solution is better for now.

sw005320:
@kan-bayashi, can you add TTS related descriptions to `README.md`? We should emphasize that now ESPnet supports TTS in the main page :)

PR #280, opened 2018-07-06
Update makefile

miguelvr:
Docker build enters in an infinite recursive make loop because Kaldi is not being correctly cloned. This PR fixes it


sw005320:
Thanks!

PR #281, opened 2018-07-07
Multi speaker TTS

kan-bayashi:
Extend Tacotron2 to multi-speaker model
https://arxiv.org/pdf/1806.04558.pdf

- [x] update taco2 to be compatible with speaker embedding
- [x] implement x-vector extraction
- [x] update json to add x-vector as second input
- [x] update training script


kan-bayashi:
Almost done.
Let me check the result.

kan-bayashi:
I found that `librosa.core.load` is fucking unstable.
Sometimes loading is randomly failed without any error message (return empty array and None), especially loading flac.
So I'm wondering to change loading part to use soundfile.

@ramon-astudillo if you observe such phenomena, let me know.

sw005320:
My preference is to use scipy wav read, but if we need to support flac, we may not have so many options.

sw005320:
My concern of soundfile is that we need the installation of libsndfile1.

ramon-astudillo:
Hi @kan-bayashi , I did not find any errors with librosa, but `sid/nnet3/xvector/extract_xvectors.sh` seems to take forever to extract. I suspect it hangs some times.

kan-bayashi:
Hi @ramon-astudillo, Thank you for your report.
Hmm... Maybe it depends on the environment.
Extracting x-vector takes a little bit long time.
In the case of train_clean_460, it takes around 4 hours with 32 parallel jobs in my environment.
If we can use more jobs, we should do.

ramon-astudillo:
Yes, in `an4` I have to limit the number of threads to 10. It takes 40min though.

egs/librispeech/tts1/local/add_input2json.py
```diff
@@ -0,0 +1,81 @@
...
+#  Apache 2.0  (http://www.apache.org/licenses/LICENSE-2.0)
+
+import argparse
+import json
+import logging
+
+
+if __name__ == '__main__':
+    parser = argparse.ArgumentParser()
+    parser.add_argument('jsons', type=str, nargs='+',
```
sw005320: Maybe we can add an option to select to add new data for either input/output.

egs/librispeech/tts1/local/compute-fbank-feats.py
```diff
@@ -0,0 +1,92 @@
+#!/usr/bin/env python
```
sw005320: We should move this one to `src/utils`

egs/librispeech/tts1/local/add_input2json.py
```diff
@@ -0,0 +1,81 @@
+#!/usr/bin/env python2
```
sw005320: We should move this one to `src/utils`

egs/librispeech/tts1/local/convert_fbank.sh
```diff
@@ -0,0 +1,71 @@
+#!/bin/bash
```
sw005320: We should move this one to `src/utils`

egs/librispeech/tts1/local/convert_fbank_to_wav.py
```diff
@@ -0,0 +1,100 @@
+#!/usr/bin/env python
```
sw005320: We should move this one to `src/utils`

tools/requirements.txt
```diff
@@ -4,5 +4,6 @@ python_speech_features>=0.6
 setuptools>=38.5.1
 joblib==0.11.0
 librosa>=0.6.1
+soundfile>=0.10.2
```
sw005320: I'm not super happy about using this library, as it requires `libsndfile`, and it is not so common. For now, it is OK, but we may need to find other libraries.

src/tts/tts_pytorch.py
```diff
@@ -208,8 +221,8 @@ def make_batchset(data, batch_size, max_length_in, max_length_out,
         logging.info('# utts: ' + str(len(sorted_data)))
         # change batchsize depending on the input and output length
         while True:
-            ilen = int(sorted_data[start][1]['input'][0]['shape'][0])
-            olen = int(sorted_data[start][1]['output'][0]['shape'][0])
+            ilen = int(sorted_data[start][1]['output'][0]['shape'][0])
```
sw005320: Is it a bug? Why did you change it?

egs/librispeech/tts1/run.sh
```diff
@@ -0,0 +1,328 @@
...
+    ${cuda_cmd} --gpu ${ngpu} ${expdir}/train.log \
+        tts_train.py \
+           --ngpu ${ngpu} \
+           --outdir ${expdir}/results \
+           --verbose ${verbose} \
+           --seed ${seed} \
+           --resume ${resume} \
+           --train-json ${tr_json} \
+           --valid-json ${dt_json} \
+           --embed_dim ${embed_dim} \
```
sw005320: another way is to add the embedding dimension information (shape) in the JSON file. It is redundant, but it is more consistent with the normal feature.

src/tts/tts_pytorch.py
kan-bayashi: Because we use same json file for tts recipe, the input and output are reversed.

kan-bayashi: In the case of ljspeech, we use shuffuled minibatch, therefore there is no problem.

kan-bayashi:
Checked `ljspeech/tts1` and `librispeech/tts1`.

kan-bayashi:
Summary:
- `spk_embed_dim` -> `use_speaker_embedding`
    - `spk_embed_dim` is loaded from json file
- `local/add_input2json.py` -> `src/utils/addjson.py`
    - select whether add to input or output via `--is-input` option 
- changed `librispeech/tts1` recipe default parameters
     - `batchsize` 32 -> 64
     - `epochs` 50 -> 30
     - `n_shift` 256 -> 512
- updated `compute_fbank_feats.py` and `convert_fbank_to_wav.py`
     - added `win_length` option
- revert `AttLoc.forward` to use larger batchsize (#305)
- fixed mask tensor cuda sending bug

kan-bayashi:
It seems that CPU memory usage is gradually increasing...
There is a bug. I've not yet found.

ramon-astudillo:
@kan-bayashi CPU or GPU memory?. GPU memory increase is normal (at some point it gets garbage collected) unless it leads to out of memory. 

I noticed ever increasing GPU memory consumption when using Tacotron for cycle consistency (in theory there are no gradients stored for this), leading to out of memory. The weird thing is this only happens on some machines. The CLSP cluster seems to finish (while having same GPU memory available).

kan-bayashi:
@ramon-astudillo CPU memory is increasing. 
I checked cpu memory usage via `top` command. 
I've never confirmed increasing GPU memory consumption.

sw005320:
The garbage collection for features through kaldi_io_py does not work correctly sometimes. You may need to explicitly delete features like https://github.com/kan-bayashi/espnet/blob/17fafba2db0edec2dd414dea8018fcf933424f90/src/asr/asr_pytorch.py#L85 https://github.com/kan-bayashi/espnet/blob/17fafba2db0edec2dd414dea8018fcf933424f90/src/asr/asr_pytorch.py#L146 or

kan-bayashi:
I will check the memory usage via `memory_profiler.profile`.

kan-bayashi:
I checked memory usage via `memory_profiler.profile`.
[memory_profile.log](https://github.com/espnet/espnet/files/2218066/memory_profile.log)
I found that `kaldi_io_py` is not related to this issue and `torch.save(model.state_dict(), path)` consumed memery sometimes (not everytime).

sw005320:
Thanks. It seems that it is not a critical issue, and I can merge it, if it is no problem.

sw005320:
#310 seems to be critical, but it does not only appear in this recipe, but the TTS recipe as well. I'm merging this PR, but we still need to work on #310.

PR #297, opened 2018-07-14
Implementation of biGRUp encoder

mallidi:


sw005320:
Thanks!
There some minor coding-convention related issues in https://travis-ci.org/espnet/espnet/jobs/403987815
Please check https://github.com/espnet/espnet/blob/master/CONTRIBUTING.md, and fix them.

mallidi:
thanks shinji 
Ran the 
$ pytest
$ flake8 src test
$ autopep8 -r src test --exclude src/utils --global-config .pep8 --diff --max-line-length 120 | tee check_autopep8
$ test ! -s check_autopep8

and it seems to succeed. 

Do I need to add bgrup versions in test/test_e2e_model.py as well?

sw005320:
> Do I need to add bgrup versions in test/test_e2e_model.py as well?

Goog point!
Yes, please.

miguelvr:
Instead of copying a bunch of code just to change the RNN cell, wouldn't it be easier to integrate it in the BLSTM class, turning it into a generic BRNN class?

It would be adding this in https://github.com/espnet/espnet/blob/974ade39bda50b5c8f88548a83748e29f005259b/src/nets/e2e_asr_attctc_th.py#L2124:

```
if cell == ''gru":
  nrnn = torch.nn.GRU(**args)
elif cell == "lstm":
  nrnn = torch.nn.LSTM(**args)
```

Just a suggestion to avoid code bloat.

Best,
Miguel

sw005320:
@miguelvr Thanks! Yes this is easier/simpler options.
@mallidi, can you do it?

mallidi:
Thanks for the suggestion.
Does the following works? 
https://github.com/mallidi/espnet/commit/2fdd869a1a9a91c1a3124b35e59de4516eb6973a

miguelvr:
Also, bear in mind that all pre-trained models won't work after this PR

mallidi:
New commit https://github.com/mallidi/espnet/commit/5ba7cd8f3a0e7049877dae1148cbacd3c7b64009
@miguelvr yes that is right. I will let Shinji make a call on this backwards oncompatibality.

mallidi:
on a second thought, I disagree changing **etype** to **cell_type**
class BLSTMP, class BGRUP or class BRRP is a class for encoder, which can have multiple recurrent layers. 
I think leaving **etype** as **etype**  is more consistent with model architecture. 

I agree with having a single BRNNP class which can define LSTMP layers, GRUP layers and other types reccurent_projection layers, based on etype.

mallidi:
@sw005320 could you comment on the backwards compatibility of this [change](https://github.com/mallidi/espnet/commit/2fdd869a1a9a91c1a3124b35e59de4516eb6973a)

src/nets/e2e_asr_attctc_th.py
```diff
@@ -2194,3 +2149,60 @@ def forward(self, xs, ilens):
...
+        '''BRNNP forward
+
+        :param xs:
+        :param ilens:
+        :return:
+        '''
+        # logging.info(self.__class__.__name__ + ' input lengths: ' + str(ilens))
+        for layer in six.moves.range(self.elayers):
+            xpack = pack_padded_sequence(xpad, ilens, batch_first=True)
+            bigru = getattr(self, 'birnn' + str(layer))
```
sw005320: To keep the backward compatibility, can you add one check whether if it has a member of 'bilstm' + str(layer) then use that member instead of 'birnn' + str(layer)? I think this would solve the backward compatibility.

miguelvr: I think that simply the fact of changing the class name will break backwards compatibility when loading pretrained weights

Issue #310, opened 2018-07-26
Memory leaking in tts_pytorch


kan-bayashi:
I found that `tts_pytorch` has memory leaking problem.
But I have not yet found the reason.

ShigekiKarita:
@kan-bayashi Do you still have this issue? Maybe I can help now because I'm trying some TTS recipes recently.

kan-bayashi:
Thank you @ShigekiKarita. I need your help.
Yes, tts recipe still has memory leaking issue.
I’ve tried several methods:
- Use `gc.colloct()` in the end of `update_core`
- removal of `batch` variable  in `update_core`
But there is no effect, memory is still leaking...

kan-bayashi:
@ShigekiKarita Did you get any idea?

ShigekiKarita:
Thanks for ping. I have done several experiments through this month with
- egs: librispeech/tts1, ljspeech/tts1
- pytorch: 0.4.1
- python: 3.6

Strangely, I have **never** seen MemoryError. If you might use Python2.7 (I remember), please try Python 3 instead.

kan-bayashi:
Thanks @ShigekiKarita.
I will try to use python3.

ShigekiKarita:
And also I found a strange line here
https://github.com/espnet/espnet/blob/f417aee2fc7b97155ae83392cfd31e8c724027cd/src/tts/tts_pytorch.py#L260-L266

I do not believe that chainer.iterators.MultiprocessIterator has maxtasksperchild in chainer 4.4.0? I'm using this without maxtasksperchild instead. What `chainer.__version__` are you using?

http://docs.chainer.org/en/stable/reference/generated/chainer.iterators.MultiprocessIterator.html

ShigekiKarita:
Oh this one https://github.com/chainer/chainer/pull/4972?

kan-bayashi:
Yes, maybe that's one.
Now we use @jnishi patched chainer.
See https://github.com/espnet/espnet/blob/master/tools/prefetch.patch

PR #313, opened 2018-07-27
[WIP] Fix memory leak in tts_pytorch

kan-bayashi:
#310 
NOT YET FIXED.


kan-bayashi:
I could found the reason of memory leaking... 
From the analysis os `memory_profiler`, converting numpy array into `torch.Tensor` consumes memory sometimes.
@ShigekiKarita Do you have any idea or how-to-find the reason?

PR #314, opened 2018-07-27
Save model after each epoch

mallidi:
Added code to save model after each epoch, using Nelson's suggestion

### DIFF
--- a/src/asr/asr_pytorch.py
+++ b/src/asr/asr_pytorch.py
@@ -332,6 +332,10 @@ def train(args):
 
     trainer.extend(extensions.snapshot_object(model, 'model.loss.best', savefun=torch_save),
                    trigger=training.triggers.MinValueTrigger('validation/main/loss'))
+    # save the model after each epoch
+    trainer.extend(extensions.snapshot_object(model, 'model_epoch{.updater.epoch}', savefun=torch_save),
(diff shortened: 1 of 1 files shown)


sw005320:
It would be better to replace this with https://github.com/mallidi/espnet/blob/4b1e006d74f28e4073885e29e6282b6aec4ea6dc/src/asr/asr_pytorch.py#L314.
Then, we'll not have to store duplicated snapshot. Make sense?

mallidi:
Does snapshot store the model as well? 
Based on the sizes of snapshot and model (), I thought snapshot doesn't include the model. 
Example: 
   snapshot_iter_1158 ->       16596
   model.acc.best       -> 62101849

Also, I was looking at a way to decode using, say, epoch 5 or epoch 10 model. 
This gives me some idea on how the model is performing.

mallidi:
CONT...
by storing the model separately, we can the asr_recog.py by pointing to model_epoch10

mallidi:
Ran a training on a subset of the librispeech data (35 hours) by removing the code at https://github.com/mallidi/espnet/blob/4b1e006d74f28e4073885e29e6282b6aec4ea6dc/src/asr/asr_pytorch.py#L314.
The CER seems to be the same, so the above code is not doing anything.

PR #318, opened 2018-07-30
Batch processing of beam search

mn5k:
| Batchsize | ATT | CTC | RNNLM | WER (test_clean) |
----|----|----|----|---- 
| 1 (orig.)  |  ✓   |   ✓  |  ✓  |  6.5    |
| 48  | ✓   |   ✓  |  ✓  |  6.5    |


src/bin/asr_train.py
```diff
@@ -105,6 +105,25 @@ def main():
                         help='Apply label smoothing with a specified distribution type')
     parser.add_argument('--lsm-weight', default=0.0, type=float,
                         help='Label smoothing weight')
+    # recognition related
```
takaaki-hori: A bit confusing. It's better to say `recognition options to compute CER/WER`

sw005320:
Can you please revisit it? This function is really important.

mn5k:
Thank you for the review.
I'll modify the code and add an option for word-based LM.

mn5k:
@takaaki-hori There is a class (MultiLevelLM, #235) that combines character-based and word-based language models in extlm_pytorch.py. Could you tell me the WER of this method?
The WER of test_eval92 was high as 102.6% in my condition.

takaaki-hori:
Maybe something wrong.
In my test, I got a bit better result than LookAheadLM in WSJ with a 20k-word LM.
But I have not tested it recently. I will revisit it in #406 .

mn5k:
Thank you for the confirmation
.   solved the problem.

mn5k:
For the computation of word-RNNLM and char/word-RNNLM, I added a buffer for sample-level processing and concatenation of samples. I'll consider a smarter way later.

mn5k:
I used editdistance in
https://github.com/aflc/editdistance
for the computation of CER/WER now.
Should I implement it without the external package?

sw005320:
I want to restart this project. It seems that we don't have serious issue, and we may merge it. @mn5k, do you have any concerns?

mn5k:
I think there's no problem.

sw005320:
we can use batch decoding as default. maybe 1 is enough. then could you add the test to check batch and non batch decodings yield the same results?

sw005320:
Thanks a lot! I merged it.

mn5k:
I added a test code to check the output of two search algorithms.

As I mentioned before, my search algorithm does not fully correspond to the original search.
I only checked the search without a fusion of other external modules.

PR #319, opened 2018-07-31
force compression-method as 2 to have CM header so that there are no issues with kaldi_io_py read_mat

sas91:
got this error in the CSJ recipe "assert(format == 'CM ') # The formats CM2, CM3 are not supported..." in "_read_compressed_mat" of kaldi_io_py. When checked some utts had CM2 header. I think copy-feats uses two CM methods by default  "kSpeechFeature if the num-rows is more than 8, and kTwoByteAuto otherwise" and kaldi_io_py doesn't have the functionality to read CM2.


sw005320:
@jnishi had the same issue. @jnishi, if it solves your case as well, I'll merge it.

Issue #334, opened 2018-08-08
Development plan for v.0.2.0


sw005320:
@kan-bayashi, @jnishi, and me discussed the following development plan, which mainly cleans and refactors codes. We'll mainly work on https://github.com/espnet/espnet/tree/dev branch, then marge the branch to the master. If you have any suggestions for this, please let us know.
- [x] fix to use PyTorch 0.4.1 (stop to support PyTorch 0.3.x) (#332)
- [x] unify the data converter with TTS one (#343)
  - perform data preprocessing and GPU copy during the converter like https://github.com/espnet/espnet/blob/master/src/asr/asr_utils.py#L59
  - stop to copy the feature to data (https://github.com/espnet/espnet/blob/master/src/asr/asr_utils.py)
  - simplify attention plot function
- [x] use data prefetch
  - we internally observed significant speed benefits
  - need to fix chainer main branch (already requested). For now, we'll use a patch
- [x] cleaning `asr_pytorch.py` and `asr_chainer.py`. There are a lot of redundant codes come from the old Kaldi reader, but now this part should be cleaned. Also we may use default functions/classes instead of custom functions/classes
  - refactor the recog part in asr_chainer.py and asr_pytorch especially after it gets nbest. #370 
  - it's better to mention as a comment which part is customized for some custom classes (e.g, grad nan check, pytorch related (optimizer.step(), pytorch evaluation mode).
- [x] rename some functions (e.g., `e2e_asr_attctc.py` -> `e2e_asr.py`, `e2e_asr_attctc_th.py` -> `e2e_asr_th.py`)
- [x] make `nets/e2e_common.py`, and move some common functions there
- [x] unify model variable arguments between TTS and ASR
- [x] change the format of model.conf from pickle to JSON
- [x] remove deprecated options including `args.train_feat`
- [x] remove deprecated options in `run.sh` including `--gpu`
- [x] avoid to use `feat-to-len`. Use `write_utt2num_frames=true`, and read utt2num instead of executing `feat-to-len`
- [x] add test code for a multiple GPU (but not with travis. Independent shell test to check it when multiple GPUs are available) #362 
- [x] make a script to check `data.json` is correctly generated
- [x] fix pytorch backend snapshot functions including the save of optimizers #362 
- [ ] fix the last epoch behavior (no validation score calculation)
- [x] update docker preparations more user friendly
- [x] add a schedule sampling #349 
- add new recipes
  - [x] add a IWSLT speech translation recipe #325 
[...]

PR #337, opened 2018-08-09
Improve model definition consistency

kan-bayashi:
unify model variable arguments between TTS and ASR (#334)

### DIFF
--- a/test/test_e2e_tts.py
+++ b/test/test_e2e_tts.py
@@ -6,6 +6,8 @@ import numpy as np
 import pytest
 import torch
 
+from argparse import Namespace
+
 from e2e_tts_th import Tacotron2
 from e2e_tts_th import Tacotron2Loss
 from tts_pytorch import pad_ndarray_list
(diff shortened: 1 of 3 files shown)


test/test_e2e_tts.py
```diff
@@ -61,7 +73,7 @@ def make_loss_args(**kwargs):
         ({"prenet_layers": 0}, {}),
         ({"postnet_layers": 0}, {}),
         ({"prenet_layers": 0, "postnet_layers": 0}, {}),
-        ({"output_activation_fn": torch.tanh}, {}),
+        ({"output_activation": "relu"}, {}),
```
sw005320: Is this change OK?

test/test_e2e_tts.py
```diff
@@ -117,7 +130,7 @@ def test_tacotron2_trainable_and_decodable(model_dict, loss_dict):
...
```
sw005320: Is this change OK?

test/test_e2e_tts.py
kan-bayashi: OK.

PR #339, opened 2018-08-09
Avoid to use feat-to-len

sw005320:
- change `run.sh` to output `utt2num_frames` to avoid to use `feat-to-len`
- TTS feature extraction does not support such functions, and it is a future work.


sw005320:
#334

PR #342, opened 2018-08-10
change model.conf from pickle to json

sw005320:
#334


src/tts/tts_pytorch.py
```diff
@@ -301,7 +301,7 @@ def train(args):
     model_conf = args.outdir + '/model.conf'
```
kan-bayashi: Why don't you use `.json` as extension?

sw005320: Will do

PR #346, opened 2018-08-11
[WIP] Docker - v0.2.0

Fhrozen:
I create this thread for the docker development base on the updates for ESPnet v0.20 #334. This is planned for larger variations.

- Storage on Docker HUB. There will be no longer necessary to build the whole container in the computer. Just only some instructions required based on variables. There are several implementations here based on the desired CUDA version (CUDNN will be only implemented with v7). It Also includes preparing CUDA only container, but I am not sure given that using CUDNN is the best solution.
- Including test for docker in Travis. Because of that will not longer be required to build the container, just download it, makes possible to implement test for docker.
-  Deletion of Virtualenv. This need to be discussed. Right now the container works with virtualenv and do not present any inconvenient. However, it may be useless given that docker already virtualize the packages so seems redundant. 

I am already preparing the container in https://hub.docker.com/r/fhrozen/dev_espnet, which is a temporal dev hub, it will be moved later to espnet.

I am not sure of using a organization because it seems to required a paid plan for building containers as organization. Some additional discussion about this may be necessary.

If there is any additional point regarding docker and future implementations, let me know.


sw005320:
I think the directory structure of docker/espnet/cpu/Dockerfile is redundant. How about simply using docker/cpu/Dockerfile? The same comment is applied to others.

Fhrozen:
Pre-built docker containers ready to use:
- [ ] runtime
  - [ ] utils
  - [ ] kaldi-tools
  - [ ] kaldi-src
- [ ] cpu
- [ ] gpu
  - [ ] cuda8.0-cudnn7
  - [ ] cuda9.0-cudnn7
  - [ ] cuda9.1-cudnn7
  - [ ] cuda9.2-cudnn7

Fhrozen:
Did some general modification:
- Moved from fhrozen/dev_espnet > espnet/dev_espnet (Repository for espnet)
- Images are built in a local PC and then pushed to the HUB. 
- Stopped Automated builts and move to repository. This is more efficient given that we can manage long time builts.
- Pushed some images ready to use. they can be found at https://hub.docker.com/r/espnet/dev_espnet/

alirezadir:
Hi @Fhrozen , I have created docker images for cuda9.0-cudnn7 and cuda9.1-cudnn7 with GPU. Please let me know how I can contribute.

Fhrozen:
@sw005320 Could you please register to https://hub.docker.com/. Also, could you ask anyone else to join the docker hub? After that please send me the users (or mail) to add them to the docker team.

alirezadir:
Thanks it's k23033.

PR #347, opened 2018-08-11
refine model.json treatment

sw005320:
- make report interval as a variable
- add model.json for RNNLM
- automatically find model.json from args.model
- remove model.conf reading options
#334

### DIFF
--- a/src/asr/asr_utils.py
+++ b/src/asr/asr_utils.py
@@ -4,6 +4,7 @@
 #  Apache 2.0  (http://www.apache.org/licenses/LICENSE-2.0)
 
 import copy
+import json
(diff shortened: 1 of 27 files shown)


src/asr/asr_utils.py
```diff
@@ -292,3 +293,13 @@ def fields(self):
 
     def keys(self):
         return self.obj.keys()
+
+
+def get_model_conf(model_path, conf_path):
```
kan-bayashi: Add docstring.

kan-bayashi: `conf_path=None` should be default  value.

Issue #322, opened 2018-08-03
Is there any already trained model?


bluemonk482:
Is there any already trained model on conversions in English that I can use off-the-shelf? Thanks.

sw005320:
The most of models I have are based on (old) pytorch 0.3.1. I'm now preparing them with pytorch 0.4.x for Librispeech. @ruizhilijhu, @jnishi, do you have the latest pytorch Librispeech models?

ruizhilijhu:
I don’t have models in librispeech. Most of my experiments are in babel Vietnamese. Harish might have Liberispeech model.@mallidi

jnishi:
@sw005320 I also don't have character model, though I have the subword model with 5000 units.

sw005320:
I just prepared the librispeech model:
- ESPnet version 0.1.5: https://github.com/espnet/espnet/releases/tag/v.0.1.5
- Models: https://github.com/espnet/espnet/releases/download/v.0.1.5/librispeech_asr1.tgz
  - RNNLM: `exp/train_rnnlm_2layer_bs256_unigram2000/rnnlm.model.best`
  - ASR models: `exp/train_960_vggblstm_e4_subsample1_2_2_1_1_unit1024_proj1024_d1_unit1024_location1024_aconvc10_aconvf100_mtlalpha0.5_adadelta_bs30_mli800_mlo150_unigram2000/results/{model.acc.best,model.conf}`
- Performance:

|           | WER (%)  |
|-----------|:----:|
| Librispeech dev_clean | 5.0 |
| Librispeech test_clean | 5.0 |

- When we use the above models, please insert the ASR model directory (`expdir`) and RNNLM model directory (`lmexpdir`) in `run.sh` as follows:
```
expdir=exp/train_960_vggblstm_e4_subsample1_2_2_1_1_unit1024_proj1024_d1_unit1024_location1024_aconvc10_aconvf100_mtlalpha0.5_adadelta_bs30_mli800_mlo150_unigram2000
lmexpdir=exp/train_rnnlm_2layer_bs256_unigram2000

        ${decode_cmd} JOB=1:${nj} ${expdir}/${decode_dir}/log/decode.JOB.log \
            asr_recog.py \
            --ngpu ${ngpu} \
            --backend ${backend} \
            --recog-json ${feat_recog_dir}/split${nj}utt/data_${bpemode}${nbpe}.JOB.json \
            --result-label ${expdir}/${decode_dir}/data.JOB.json \
[...]

PR #354, opened 2018-08-14
[WIP] improve the librispeech recipe

sw005320:
Now I'm working on the improvement of the librispeech recipe motivated by the RWTH setup (thanks to Rohit Prabhavalkar and Kazuki Irie)
- sentencepiece model as a default
- VGG-BLSTM
- shallow and wide network (3layer BLSTMS with 1024 units (encoder), 1024 dim attention, unidirectional LSTM with 1024 units (decoder)
- fast convergence (change the maximum of epochs from 15 to 10)
- significant WER improvement (from 7.2 to 5.1 in `test_clean`)

TODO
- [ ] change the decoder layer from 1 to 2
- [ ] tune some search parameters (rnnlm weight)
- [ ] upload a model (#322)


chenzhehuai:
It is consistent with my observation except that adding more layers up to 6 with 1024 units still obtains some improvement.

sw005320:
Thanks. Will follow your suggestions (when GPUs are available). Could you share your WER results? If it is better than https://github.com/espnet/espnet/blob/d51e76c0baa556e28a3e090335944478828fbc65/egs/librispeech/asr1/RESULTS, then, I may follow your network architecture, or ask you to make a PR.

sw005320:
The trained models can be provided through the release (e.g., https://github.com/espnet/espnet/releases/download/untagged-f5ccde023841a43380a9/librispeech_asr1.tgz)

chenzhehuai:
No, my observation is from train_100. I think you system is the best til now.

sw005320:
https://arxiv.org/pdf/1805.03294
This is not exactly same as what we're now using, but our set up is based on the discussion with them what would be most dominant given our current implementation.

ruizhilijhu:
From this paper, their system played with the subsampling factor 32 in the pertaining and 8 for fine-tuning and they showed some improvement. 

In our babel-10 setting, we used subsampling factor 4, and we allowed the minimum number of frames in an utterance to be 10 frames. 

Is there any specific reason use our current setting?

sw005320:
We don't have a specific reason, and we may test further subsampling like the RWTH paper, but I internally found that further subsampling (8) slightly degrades the performance in the Librispeech task.

PR #358, opened 2018-08-16
scheduled sampling for dev branch

creatorscan:
- same changes as in master branch


hirofumi0810:
Hi @creatorscan,
I think you don't have to pass through F.log_softmax() when sampling previous tokens.

creatorscan:
@hirofumi0810 Yeah you are right, but in most datasets I observed gains by including softmax.

creatorscan:
@hirofumi0810 @sw005320 @takaaki-hori Fixed the difference in performance with inclusion of softmax. This was due to usage of torch.max, using np.argmax(z_out.detach(), dim=1) solved the issue as it was similar to inference step..

src/nets/e2e_asr.py
```diff
@@ -11,6 +11,7 @@
 from argparse import Namespace
 
 import numpy as np
+import random
```
sw005320: Did you set the seed appropriately?

src/nets/e2e_asr_th.py
```diff
@@ -14,6 +14,7 @@
 
 import chainer
 import numpy as np
+import random
```
sw005320: seed?

creatorscan: Thanks for pointing it out Shinji. I did not focus on the seed before and now after ur message I ran some experiments and found the results vary with different seeding. Till now I was using default system settings for seeding.
Tested with voxforge - Italian

| seed | % CER |
|--------|-----------|
| baseline| 14.8|
| 1 | 15.0 |
| 2 | 15.0|
|3 | 15.1|
|4 |14.5 |
|5| 15.6 |

sw005320: I think this change happens unfortunately, and we may not have a way to avoid it. We can leave it as it is now.

> So, will it be better if we set seeding as a parameter ?

Some recipe has an option to accept the seed, e.g., 
https://github.com/espnet/espnet/blob/d3cf8f0c2965d8cd3e9fc06b3d84b0af4318c9b8/egs/csj/asr1/run.sh#L18
We could make this as a default, but we may not have to do it for all recipes for now.

src/nets/e2e_asr.py
```diff
@@ -582,7 +584,18 @@ def __call__(self, hs, ys):
...
         for i in six.moves.range(olength):
             att_c, att_w = self.att(hs, z_list[0], att_w)
-            ey = F.hstack((eys[i], att_c))  # utt x (zdim + hdim)
+            if random.random() < self.sampling_probability:
+                logging.info(' scheduled sampling ')
+                if i == 0:
+                    # for initializing z_list[0]
+                    ey = F.hstack((eys[i], att_c))  # utt x (zdim + hdim)
+                    c_list[0], z_list[0] = self.lstm0(c_list[0], z_list[0], ey)
+                z_out = self.output(z_list[0])
```
hirofumi0810:
```
z_out = self.output(z_all[-1])
```
is easier to understand.

src/nets/e2e_asr_th.py
```diff
@@ -1638,7 +1641,14 @@ def forward(self, hs_pad, hlens, ys_pad):
         # loop for an output sequence
         for i in six.moves.range(olength):
             att_c, att_w = self.att(hs_pad, hlens, z_list[0], att_w)
-            ey = torch.cat((eys[:, i, :], att_c), dim=1)  # utt x (zdim + hdim)
+            if random.random() < self.sampling_probability:
+                logging.info(' scheduled sampling ')
+                z_out = self.output(z_list[-1])
```
hirofumi0810:
```
z_out = self.output(z_all[-1])
```
is easier to understand.

sw005320:
I think it's almost OK. My concern is that why the schedule sampling does not work when we use the 2-layer decoder network. Do you have any reasons?

creatorscan:
> I think it's almost OK. My concern is that why the schedule sampling does not work when we use the 2-layer decoder network. Do you have any reasons?

Yeah, this scenario only happens for very low resource datasets such as Voxforge. But for wsj and swbd I observed gains.
The reason behind performance degradation with  2 decoder layers for smaller databases has to be investigated

sw005320:
I see. If this comes from small data issues, then it is no problem.

creatorscan:
> I see. If this comes from small data issues, then it is no problem.

For example: these are the results of SWBD with the config "--etype blstmp --elayers 6 --eunits 320 --eprojs 320 --subsample 1_2_2_1_1 --dlayers 2 --dunits 300 --atype location --aconv-chans 10 --aconv-filts 100 --mtlalpha 0.0 --batch-size 30 --maxlen-in 800 --maxlen-out 150 --opt adadelta"



|Model | hub5(%WER) | dev(%WER) | hub5 (%CER)| dev(%CER)|
|---------|--------|------|--------|-----|
baseline | 44 | 23.3 | 27.1 | 13.8
mt 0 | 45.9 | 23.9 | 29.2 | 14.8
mt 0 + ss 0.5 | 43.8 | 22.4 | 26.3 | 12.5

PR #362, opened 2018-08-18
New snapshot save function and multi-gpu tests

kan-bayashi:
Save trainer / model / optimizer states into one file in pytorch.

- Added `torch_snapshot` to save trainer / model / optimizer states into one file
- Added `torch_resume` to resume training from new snapshot
- Unified `torch_save` and `torch_load` functions in ASR and TTS
- Added multi-gpu tests for pytorch
- Added save / load test for pytorch
- Added more docstrings

### DIFF
--- a/src/asr/asr_utils.py
+++ b/src/asr/asr_utils.py
@@ -374,7 +374,6 @@ def torch_resume(snapshot_path, trainer):
 
     :param str snapshot_path: snapshot file path
     :param instance trainer: chainer trainer instance
-    :return: trainer instance whose states are loaded
     """
     # load snapshot
     snapshot_dict = torch.load(snapshot_path, map_location=lambda storage, loc: storage)
--- a/src/asr/asr_utils.py
+++ b/src/asr/asr_utils.py
@@ -147,43 +147,6 @@ class CompareValueTrigger(object):
         self._summary = chainer.reporter.DictSummary()
 
 
-def restore_snapshot(model, snapshot, load_fn=chainer.serializers.load_npz):
-    '''Extension to restore snapshot'''
-    @training.make_extension(trigger=(1, 'epoch'))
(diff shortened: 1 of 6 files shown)


src/asr/asr_utils.py
```diff
@@ -235,6 +204,87 @@ def _plot_and_save_attention(self, att_w, filename):
...
+    return torch_snapshot
+
+
+def _torch_snapshot_object(trainer, target, filename, savefun):
+    # make snapshot_dict dictionary
+    s = DictionarySerializer()
+    s.save(trainer)
+    if hasattr(trainer.updater.model, "model"):
+        # (for TTS)
+        if hasattr(trainer.updater.model.model, "module"):
```
sw005320: Why TTS has a bit redundant model structure (`model.model`)?
Is there a way to make it same as ASR?

kan-bayashi: If we assume that same options are used when resuming, it will be more simple.

But if you want to keep same format as model.loss.best, current format is needed.

kan-bayashi:
@sw005320 Made more simple. which is better?
Previous: may work if we change options when resuming, but complex
Now: not work if we change options but simple

sw005320:
My preference is previous one.
The code becomes complex in `asr_utils.py`, but this is OK.

kan-bayashi:
Reverted.

kan-bayashi:
Updated.
- Replaced loading part in `asr_pytorch.recog` with `torch_load`
- Update `torch_load`
    - Both model file and snapshot file to be used in decoding phase
- Added some cleaning part (remove loaded object explicitly)

kan-bayashi:
Added more test codes.

kan-bayashi:
@sw005320 I want to merge. Is it OK?

PR #383, opened 2018-08-27
Pad encoder outputs in attention scoring

hirofumi0810:
- Pad encoder outputs with -1024.0
- Add scaling=1.0 option to multi-head score functions
- Fix scheduled sampling (remove F.log_softmax)


sw005320:
Can you commit only padding part? Also please report the ASR and ST results with and without this change.

PR #340, opened 2018-08-09
Prefetch training data by multiprocessiterator

jnishi:
Replacing #323 

Loading kaldi's scp file is too slow, but this can be done in parallel with forward and backward calculation of other data. This can be easily done by using MultiprocessIterator.
This PR improves both of Chainer and Pytorch backends.

Though  this PR requires chainer/chainer#4972 to be merged, I add patch for current version of Chainer.

PR #388, opened 2018-08-28
Reduce processes 

jnishi:
Use `SerielIterator` to validation dataset and set `n_processes=1` to train dataset to reduce memory usage which is discussed in https://github.com/espnet/espnet/pull/340#issuecomment-416615426.


sw005320:
Thanks.
Did you observe some speed degradation? If it is not so significant, I'll merge it.

PR #396, opened 2018-08-30
Update to show predction result during recognition

kan-bayashi:
This PR makes it possible to check the recognition result during recognition.


sw005320:
You can only display one best results. N best would be too much.

PR #397, opened 2018-08-31
Perform explicit value filling for attention

kan-bayashi:
This PR makes it possible to perform explicit filling to make attention weight 0 for padded part, which is pointed out by @hirofumi0810.


kan-bayashi:
I noticed masking of encoder states is performed by @ShigekiKarita in old version.
But I wrongly deleted them because I misunderstood processing of the encoder.
(If we use `pad_packed_sequence`, the sequence is already padded with 0.0. But in our case there is linear layer after `pad_packed_sequence`, so there is a bias value in padded value.)
So we should revert the deletion.

kan-bayashi:
@ShigekiKarita How do you think whether to need to perform filling for energy in attention?

kan-bayashi:
I checked the performance difference with WSJ recipe.
```
# before 
eval = 3.5, dev = 4.7
# after
eval = 3.5, dev = 4.8
```
Performance is not changed.

And I confirmed first version implementation is inefficient.
So I update by following this IBM's implementation 
<https://github.com/IBM/pytorch-seq2seq/blob/f146087a9a271e9b50f46561e090324764b081fb/seq2seq/models/attention.py>
The training time of WSJ is as follows:
```
# without masking
0.90 iters / sec
# with first version masking
0.80 iters /sec
# with updated masking
0.9 iters / sec
```
The training time of LJspeech is as follows:
```
# without masking
0.15 iters / sec
# with first version masking
0.10 iters / sec
# with updated masking
TBA
```
As my conclusion, filling padded part is:
- reasonable in theoretically
- no effect in ASR setup
- no additional time in training

ShigekiKarita:
@kan-bayashi sorry for late
> How do you think whether to need to perform filling for energy in attention?
I think that everything should be reasonable. Recap: softmax attention is defined as softmax over time frames for each utterances, padded value should be -inf (exp(-inf) = 0).

https://github.com/r9y9/tacotron_pytorch/blob/5f41d9d8f70a299a49f02aa5422478e1693ebe93/tacotron_pytorch/attention.py#L48

``` python
import torch
assert torch.nn.functional.softmax(torch.tensor([1.0, -float("inf")]), dim=0) == [1.0, 0.0]
```

ShigekiKarita:
I misunderstood your comment. Now I approve this change because I read your code.

src/nets/e2e_tts_th.py
```diff
@@ -30,27 +30,25 @@ def decoder_init(m):
...
-def make_mask(lengths, dim=None):
-    """FUNCTION TO MAKE BINARY MASK
+def make_non_pad_mask(lengths):
+    """Function to make mask tensor containing indices of non-padded part
 
-    Args:
-        length (list): list of lengths
-        dim (int): # dimension
+    e.g.: lengths = [5, 3, 2]
+          mask = [[1, 1, 1, 1 ,1],
```
ShigekiKarita: this example or implementation is wrong. This function returns as same as e2e_asr_th.make_pad_mask
``` python
           mask = [[0, 0, 0, 0, 0], 
                        [0, 0, 0, 1, 1],
                        [0, 0, 1, 1, 1]] 
```
If you really need this consider `~e2e_asr_th.make_pad_mask(length)` instead of redefinition

src/nets/e2e_tts_th.py
```diff
@@ -134,7 +132,7 @@ def forward(self, xs, ilens, ys, labels, olens=None, spembs=None):
             else:
                 weights = None
             # masking padded values
-            mask = to_cuda(self, make_mask(olens, ys.size(2)))
+            mask = to_cuda(self, make_non_pad_mask(olens).unsqueeze(-1))
```
ShigekiKarita: `~make_mask(olens).unsqueeze(-1)` is right?

src/nets/e2e_tts_th.py
```diff
@@ -30,27 +30,25 @@ def decoder_init(m):
...
     """
-    batch = int(len(lengths))
+    bs = int(len(lengths))
     maxlen = int(max(lengths))
-    if dim is None:
-        mask = torch.zeros(batch, maxlen)
-    else:
-        dim = int(dim)
-        mask = torch.zeros(batch, maxlen, dim)
+    mask = torch.zeros(bs, maxlen).byte()
```
ShigekiKarita: use `torch.zeros(bs, maxlen, dtype=torch.uint8)` because it can byte-allocate once instead of twice as float-allocate ->byte-allocate

src/nets/e2e_asr_th.py
```diff
@@ -66,20 +66,25 @@ def pad_list(xs, pad_value):
...
+    :return: mask tensor contraining indices of padded part (B, Tmax)
+    :rtype: torch.Tensor
     """
-    assert xs.size(0) == len(ilens)
-    new_xs = xs.new(*xs.size()).fill_(fill_value)
-    for idx, l in enumerate(ilens):
-        new_xs[idx, :l] = xs[idx, :l]
+    bs = int(len(lengths))
+    maxlen = int(max(lengths))
+    mask = torch.zeros(bs, maxlen).byte()
```
ShigekiKarita: use torch.zeros(bs, maxlen, dtype=torch.uint8) because it can byte-allocate once instead of twice as float-allocate ->byte-allocate

src/nets/e2e_tts_th.py
ShigekiKarita: use torch.zeros(bs, maxlen, dtype=torch.uint8) because it can byte-allocate once instead of twice as float-allocate ->byte-allocate

ShigekiKarita:
LGTM except for comments

kan-bayashi:
Done!

sw005320:
It would be better to have masking in the chainer attention as well. Please consider it.

PR #403, opened 2018-09-01
Docker v0.2

Fhrozen:
Docker updates and images uploaded to Docker Hub with last espnet version.

### DIFF
--- a/docker/requirements.txt
+++ b/docker/requirements.txt
@@ -1,8 +1,0 @@
-torch==0.4.1
-chainer==4.3.1
-python_speech_features>=0.6
-setuptools>=38.5.1
-librosa>=0.6.2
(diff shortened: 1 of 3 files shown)


docker/requirements.txt
```diff
@@ -0,0 +1,8 @@
+torch==0.4.1
```
sw005320: It would be better to use https://github.com/espnet/espnet/blob/master/tools/requirements.txt, then we can reflect the change of the main requirements.txt automatically. Also, why does it exclude cupy?

Fhrozen: cupy breaks the build for the cpu image so I tried to use a different requiremient. I will modify to exclude it using another command inside the container.

PR #406, opened 2018-09-04
Improve LM training with sentence-level iterator

takaaki-hori:


sw005320:
Thanks. Is this PR ready for a review? After that, I'll also apply it to the Librispeech (instead of #407).

takaaki-hori:
Yes, it's ready. Please review. One problem is that I got 5.4%WER for WSJ eval'92 with this version, which is a bit worse than 4.9% I obtained before. Since I can get 4.9% with the old encoder decoder trained two weeks ago using the same RNNLM, this difference might come from some small changes recently applied to ESPnet, or other factors such as initialization and chainer version.
I have also modified run.sh in egs/librispeech/asr1 to use new LM tools, but not yet tested ASR. Please tune some lm-related parameters in run.sh. If it works better for librispeech, we can apply it to other recipes. Note that once we merge this version, old LMs cannot be used for decoding.

src/lm/extlm_chainer.py
```diff
@@ -104,13 +104,22 @@ def __call__(self, state, x):
             else:
                 wlm_logprob = wlm_logprobs[:, self.word_unk] + self.log_oov_penalty
             log_y[:, self.space] = wlm_logprob
-            log_y[:, self.eos] = wlm_logprob + wlm_logprobs[:, self.word_eos]
+            log_y[:, self.eos] = wlm_logprob
```
sw005320: Why did you change like that? I just don't fully understand this part.

src/lm/lm_chainer.py
```diff
@@ -115,32 +112,124 @@ def predict(self, state, x):
...
+        emb = self.embed(x)
+        c[0], h[0] = self.lstm[0](state['c'][0], state['h'][0], F.dropout(emb))
+        for n in six.moves.range(1, self.n_layers):
+            c[n], h[n] = self.lstm[n](state['c'][n], state['h'][n], F.dropout(h[n - 1]))
+        y = self.lo(F.dropout(h[-1]))
+        state = {'c': c, 'h': h}
         return state, y
 
 
+class BPTTUpdater(training.updaters.StandardUpdater):
```
sw005320: This class seems to be quite similar to the StandardUpdater. I feel that we may just use the StandardUpdater, and we don't have to use this kind of custom updater. But maybe I'm wrong.

src/lm/lm_chainer.py
```diff
@@ -115,32 +112,124 @@ def predict(self, state, x):
...
+        reporter.report({'count': count}, optimizer.target)
+        # update
+        loss /= batch_size  # normalized by batch size
+        optimizer.target.cleargrads()  # Clear the parameter gradients
+        loss.backward()  # Backprop
+        loss.unchain_backward()  # Truncate the graph
+        optimizer.update()  # Update the parameters
+
+
+class LMEvaluator(extensions.Evaluator):
```
sw005320: Same comment with the BPTTUpdater. We may use extensions.Evaluator instead of making a custom evaluator.

src/lm/lm_chainer.py
```diff
@@ -210,78 +303,47 @@ def train(args):
...
-            epoch_now = train_iter.epoch
+    updater = BPTTUpdater(train_iter, optimizer, gpu_id)
+    trainer = training.Trainer(updater, (args.epoch, 'epoch'), out=args.outdir)
+    trainer.extend(LMEvaluator(val_iter, model, device=gpu_id))
+    trainer.extend(extensions.LogReport(postprocess=compute_perplexity,
+                                        trigger=(REPORT_INTERVAL, 'iteration')))
+    trainer.extend(extensions.PrintReport(
+        ['epoch', 'iteration', 'perplexity', 'val_perplexity', 'elapsed_time']
+    ), trigger=(REPORT_INTERVAL, 'iteration'))
+    trainer.extend(extensions.ProgressBar(update_interval=10))
```
sw005320: You may use `REPORT_INTERVAL` instead of using 10

src/lm/lm_pytorch.py
```diff
@@ -132,20 +155,110 @@ def zero_state(self, batchsize):
...
+    x, t = convert.concat_examples(batch, padding=padding)
+    x = torch.from_numpy(x)
+    t = torch.from_numpy(t)
+    if device is not None and device >= 0:
+        x = x.cuda(device)
+        t = t.cuda(device)
+    return x, t
+
+
+class BPTTUpdater(training.StandardUpdater):
```
sw005320: same comment as `lm_chainer.py`

src/lm/lm_pytorch.py
```diff
@@ -132,20 +155,110 @@ def zero_state(self, batchsize):
...
+        # update
+        loss = loss / batch_size  # normalized by batch size
+        self.model.zero_grad()  # Clear the parameter gradients
+        loss.backward()  # Backprop
+        if self.gradclip is not None:
+            nn.utils.clip_grad_norm_(self.model.parameters(), self.gradclip)
+        optimizer.step()  # Update the parameters
+
+
+class LMEvaluator(extensions.Evaluator):
```
sw005320: same comment as `lm_chainer.py`

src/lm/lm_pytorch.py
```diff
@@ -132,20 +155,110 @@ def zero_state(self, batchsize):
...
+        # report validation loss
+        observation = {}
+        with reporter.report_scope(observation):
+            reporter.report({'loss': float(loss / count)}, self.model.reporter)
+        return observation
+
+
+# Routine to rewrite the result dictionary of LogReport to add perplexity
+# values
+def compute_perplexity(result):
```
sw005320: This can be moved to `lm_utils.py`

egs/librispeech/asr1/run.sh
```diff
@@ -179,30 +186,35 @@ if [ ${stage} -le 2 ]; then
...
 
 # You can skip this and remove --rnnlm option in the recognition (stage 5)
-lmexpdir=exp/train_rnnlm_${backend}_2layer_bs256_${bpemode}${nbpe}
+lmexpdir=exp/train_rnnlm_${backend}_${lm_layers}layer_unit${lm_units}_${lm_opt}_bs${lm_batchsize}_${bpemode}${nbpe}
 mkdir -p ${lmexpdir}
 if [ ${stage} -le 3 ]; then
     echo "stage 3: LM Preparation"
     lmdatadir=data/local/lm_train_${bpemode}${nbpe}
     mkdir -p ${lmdatadir}
-    spm_encode --model=${bpemodel}.model --output_format=piece < data/lang_char/input.txt | perl -pe 's/\n/ <eos> /g' \
```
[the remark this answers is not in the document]
takaaki-hori: Okay, I will do it.

src/lm/extlm_chainer.py
takaaki-hori: Actually, `<eos>` probability was incorrect here although ASR accuracy was the same before and after this change.
The LM state needed to be updated with the last word to get `<eos>` prob, but it is computationally very expensive. So I made `final()` method in the word RNNLM classes to compute `<eos>` prob, and changed the beam search code to apply `<eos>` prob only for `ended_hyps`.

src/lm/lm_chainer.py
takaaki-hori: StandardUpdater does not have a loop for tokens in a given sequence, so we need to use this custom updater like the chainer's ptb example. Maybe it's possible to use the standard updater by changing `ClassifierWithState` class to include the loop. I think I can work on that, but need a bit more time. So it's better to do it in a different PR.

src/lm/lm_chainer.py
takaaki-hori: Maybe possible to use the default evaluator. but it's better to do it in a different PR.

src/lm/lm_utils.py
```diff
@@ -93,3 +123,142 @@ def serialize(self, serializer):
                     self._previous_epoch_detail, 0.)
             else:
                 self._previous_epoch_detail = -1.
+
+
+# Dataset iterator to create a batch of sentences.
```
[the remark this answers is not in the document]
takaaki-hori: Okay, I will do so for all the parts you pointed out.

takaaki-hori:
@sw005320 , @kan-bayashi , I have modified `run.sh` in babel, chime4, chime5, csj, fisher_swbd, hkust, librispeech, reverb, tedlium, and wsj recipes to use new LMs for this PR. But I tested only wsj and librispeech since it is difficult to run all the recipes. I think it’s better to ask some people to check if there is no problem.

sw005320:
Thanks. I just merged it. The change seems to be reasonable, and if someone has some problems, they will raise some issues.

PR #418, opened 2018-09-22
Add reduction factor option

kan-bayashi:
This PR adds `reduction factor` option.
This PR should be merged after #417.


src/nets/e2e_tts_th.py
```diff
@@ -650,6 +695,10 @@ def forward(self, hs, hlens, ys, att_w_maxlen=None):
         :return: attetion weights (B, Lmax, Tmax)
         :rtype: torch.Tensor
         """
+        # thin out frames (B, Lmax, odim) ->  (B, Lmax/r, odim)
+        if self.reduction_factor > 1:
+            ys = ys[:, self.reduction_factor - 1::self.reduction_factor]
```
sw005320: Is it OK to simply skip the output frame?
I thought that it would concatenate multiple frames, e.g., (B, Lmax, odim) ->  (B, Lmax/r, r * odim)

kan-bayashi: Yes, that is alternative choice.
But in this time, I followed original tacotron paper. 

https://arxiv.org/pdf/1703.10135.pdf
> In inference, at decoder step t, the last frame of the r predictions is fed as input to the decoder at step t + 1. Note that feeding the last prediction is an ad-hoc choice here – we could use all r predictions. During training, we always feed every r-th ground truth frame to the decoder. The input frame is passed to a pre-net as is done in the encoder.

sw005320: OK. You can leave it, but you may test it in some occasions.

PR #424, opened 2018-09-26
Change default iterator

kan-bayashi:
This PR changes default iterator.


sw005320:
Thanks.
@jnishi, we changed to control the multiprocess iterator from the option.

PR #431, opened 2018-10-03
remove CTC NaN utterances, add vggsubsample 3 option

synetkim:
for chime-5 espnet baseline

1) add vggsubsample 3 option
2) remove CTC NAN utterances

### DIFF
--- a/src/bin/asr_train.py
+++ b/src/bin/asr_train.py
@@ -55,6 +55,7 @@ def main():
                         help='Number of encoder hidden units')
     parser.add_argument('--eprojs', default=320, type=int,
                         help='Number of encoder projection units')
+    parser.add_argument('--vggsubsample', default=4, type=int, help='subsample factor in VGG')
     parser.add_argument('--subsample', default=1, type=str,
                         help='Subsample input frames x_y_z means subsample every x frame at 1st layer, '
                              'every y frame at 2nd layer etc.')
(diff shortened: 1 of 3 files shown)


sw005320:
Thanks @synetkim!
- Can you fix travis CI check? You could find how to pass it by reading https://github.com/espnet/espnet/blob/master/CONTRIBUTING.md
- Can you add your best option setup to https://github.com/espnet/espnet/blob/master/egs/chime5/asr1/run.sh? Also, update the RESULTS?

sw005320:
@synetkim, can you also tell me your experience about which is more dominant whether removing irregular sentences in CTC or changing subsampling?

synetkim:
@sw005320 Sure. I will fix travis CI check, and update run.sh / RESULTS and re-run the experiments w/ and w/o these changes.  
@Fhrozen Hi. subsampling 3 is only for the VGG net and the default is 4. So it won't affect other experiments.

Fhrozen:
@synetkim Thank you very much for your support to this community. I allow us to improve.

@sw005320, I noticed that this is only implemented for chainer_ctc. I am trying to implement for warp_ctc but I got some bug about gradients and shapes. (Which does not occur in the chainer_ctc use.)
Should I migrate to chainer_ctc? or will we using warp_ctc? In second case, could you suggest me who can I work with to implement the nan in warp_ctc. I just forked and trying to solve the problems. 
https://github.com/Fhrozen/chainer_ctc

Issue #402, opened 2018-08-31
Resume LM Training


dneelagiri:
Hello,

When I try to train a 20GB training data file using pytorch RNNLM I always run into OOM issue where the training process is being killed by OS. For Network training there is option to use the snapshot and resume the training. Is there any similar option for RNNLM training where I can split 20G file and use the snapshot to continue training for other data. 

This is useful not only for training huge data but also when we want to add new corpus we can avoid training entire RNNLM again.

Thank you for your support and the latest changes development 0.2.0 really help us in faster training.

Thanks

sw005320:
Thanks for your feedback.
We're actually working on the LM training part more similar style to the main end-to-end ASR training including a resume function. So, please wait for a while.

dneelagiri:
@sw005320 : Thank you for considering the previous request. Would you kindly also consider any way to speed up RNNLM training. Currently for a 7G data and 500K vocab file I am running into OOM error even after using batch size 4. And if I used 20K vocab file training started but its taking 3-4 days for an epoch.

Thank you!!

sw005320:
@takaaki-hori is working on #406. About the speed, 500k vocab sounds very large, and I'm not sure we can manage to make some speed up for this setup.

sw005320:
@dneelagiri, #406 would fix your issue. Could you try the latest version?

dneelagiri:
@sw005320 I was able to use the resume successfully. On Libri 960 text LM training took about 3 hours.  
One other issue I faced was gpu_id assignment. This seems to have been addressed already for  chainer in https://github.com/espnet/espnet/pull/379 . Can you also please update it for pytorch?

PR #432, opened 2018-10-04
Fix LM gpu_id bug

Fhrozen:
Adding GPU id for no gpu


sw005320:
This PR is related to #402

PR #440, opened 2018-10-25
Modularize

kamo-naoyuki:
This PR consists these changes:

- Renamed `src/` to `espnet/` to modularize "espnet" as a python module.
- Changed `Makefile/conda.mk` to use `setup.py` to install espnet and modified `egs/*/*/path.sh`.
- Changed to install kaldi_io_py using pip.
- Moved `src/utils/` to `egs/wsj/asr1/utils_espnet/` to make the scripts mode clear about its directory path.
- Changed `egs/*/*/run.sh` and all scripts in `egs/*/*/local` to refer `utils_espnet/*` instead of its script name.
- Fixed test codes due to these changes.

Current structure of espnet is  a set of multiple python modules, e.g. asr_pytorch, lm_chainer, or etc. thus every python scriptss need to import directly its name, like,

```python
from e2e_asr_th import to_cuda
from asr_utils import torch_load
```
This is very unreadable because it is not clear which directory they come from. Additionally to say, IDE, e.g. PyCharm, cannot detect their module path, and this is very annoying for me.

If the module directory is put at the root directory, PyCharm can detect the path of python modules without any setting and the previous sentences changes as:

```python
from espnet.nets.e2e_asr_th import to_cuda
from espnet.asr.asr_utils import torch_load
```

Due to this change, I also changed the place of the `src/utils` to `egs/wsj/asr1/utils`.
It is very odd for me that such scripts is put with the other core modules and, also, I want to improve the readability of the scripts. Espnet is a mixture system of the multiple tools, `Kaldi`, `sclite`, `nkf`, and etc. so I think it is better to make clear about the command path.

Finally, I added `setup.py` to install espnet using `pip`. Since no longer need to set `PATH` for espnet and `PYTHONPATH` any more , I deleted such parts from `path.sh`.


ShigekiKarita:
I think deleting PYTHON_PATH might make source code handling difficult because there will be two paths. one is repo and the other is under site-packages. However maybe setup.py is usefull for who want to reuse ESPnet resources as a library. I expect that the people who need setup.py and who need PYTHON_PATH (i.e., current ESPnet users) can be different.

kamo-naoyuki:
@ShigekiKarita Do you care about the place of the packages, to say under the `site-packages` of Python,  after installation from pip? If so, the editable mode of pip may solves it since the module is installed at the same location and you can edit them directly as in the past.

In my PR, espnet is installed as editable mode by Makefile.

``` bash
pip install -e espnet
```

Plus, espnet still can works  with PYTHONPATH:

```bash
export PYTHONAPTH=$(repository_root):$PYTHONAPTH
export PATH=$(repository_root)/espnet/bin:$PATH
```

However, additional environment changes should be the smallest as possible.

kan-bayashi:
Thank you for your contribution.
I think your PR makes it clearer and easier to check the codes.

I'm not familiar with pip installation, so let me confirm some points.
- Are `requirements.txt` and `test_requirements.txt` needed?
- If we use `-e` options, can we directly modify the codes under `espnet/*` ?
- The existence of both `utils ` and `utils_espnet` is a little bit strange for me.
  Are there any smart way to avoid it?

(I don't know why travis is terminated...)

kamo-naoyuki:
@kan-bayashi 

> - Are `requirements.txt` and `test_requirements.txt` needed?

No it doesn't.  
About test_requirements.txt,  when using test command of setup.py, the requirements for tests is installed automatically.

```bash
python setup.py test  # Run pytest in this case
```

I reserved it because I'm not sure which ways you prefer. 


> - If we use `-e` options, can we directly modify the codes under `espnet/*` ?

Yes. With -e, the source codes are not copied in the site-packages unlike normal pip and the module are referred from `espnet/*`. Thus your modification directly reflects .

@sw005320 
> Putting the utility scripts under the specific recipe may be a little bit confusing

@kan-bayashi 
> - The existence of both `utils` and `utils_espnet` is a little bit strange for me.
Are there any smart way to avoid it?

I agree it seems strange and actually I'm not confident about this. Since I'm too familiar with Kaldi, so I feel all scripts should be placed under `egs`, but I think it is just a preference.

My claims are:

- Utility should be distinguished from core modules because it is ambiguous whether it is a python module or a script, and the targets for tests are also not clear now.
  - So, to make any other name directory under the root, like ${ESPNET_ROOT}/scripts , could works for me in this point.
- The commands in `PATH` should be more formal in unix system because `PATH` can overwrite any commands, even builtin unix commands, and if too many commands exist, It might happen by mistake in the future. Thus we should make the commands to be registered in `PATH` to be the fewest.
- Because espnet uses many external libraries, kaldi, sclite, etc, so if only the command name is written in scripts, I need to search where it comes from. 
  - You may think that it has same problem about `src/bin` or any other commands as it is. The difference is that utility scripts are created easier than such core commands, thus easy to increase.
- Since Espnet inherits many kaldi style,  shouldn't utility scripts be placed under the egs? I think it is natural. 

> (I don't know why travis is terminated...)

I'm stuck too... Does anyone have idea about this?

ShigekiKarita:
> I'm stuck too... Does anyone have idea about this?

You will find what is going in the raw log https://api.travis-ci.org/v3/job/446063078/log.txt
The pytorch installation bellow seems to output too long lines.

```
if [[ "$TRAVIS_PYTHON_VERSION" == "2.7" ]]; then pip install http://download.pytorch.org/whl/cpu/torch-${TH_VERSION}-cp27-cp27mu-linux_x86_64.whl ; else pip3 install http://download.pytorch.org/whl/cpu/torch-${TH_VERSION}-cp36-cp36m-linux_x86_64.whl ; fi
```

use `pip install foo --progress-bar off`

kamo-naoyuki:
@sw005320 
I have some questions.

About Makefile:

> # for some reason, including matplotlib in requirements.txt causes errors, and
> # matplotlib should be separately pip installed

Could you explain what is this error? I can't believe this causes error.

> virtualenv: requirements.txt

Why this target name isn't  "venv"? In this Makefile, venv will be recreated at the second time.

sw005320:
Yes, it should be `venv`, not `virtualenv`

sw005320:
Two comments
1. How about `${ESPNET_ROOT}/scripts` -> `${ESPNET_ROOT}/utils`?
I still want to call the directory utils
2. What is the benefit of using `XXX.done` in `Makefile`?

kamo-naoyuki:
> Two comments
> 
> 1. How about `${ESPNET_ROOT}/scripts` -> `${ESPNET_ROOT}/utils`?
>    I still want to call the directory utils

Sorry for confusing, I meant $(REPOSITORY_ROOT)/script. 
Is it $(REPOSITORY_ROOT)/espnet/utils in your comment? If so, why?

Of course, it is no problem, but very confusing because most python projects use `utils` for python modules and normally, utility scrips are placed on another level, as I known. Thus $(REPOSITORY_ROOT)/utils might be good to keep current usage.

What do you think? I'm not particular about it, but I don't know why you prefer it.

> 2. What is the benefit of using `XXX.done` in `Makefile`?

Even if the target failed in any step, e.g compiling, the target  won't be made in the next make flow, because the directory that has the target name already exists. I want to make the target name different from the directory name. I think it is the same purpose in the current Makefile that a symbolic link 
 is created for `kaldi` directory.

And, sorry, this commits should be separated from this PR. I was lazy.

utils/remove_longshortdata.sh
```diff
@@ -32,12 +32,12 @@ cat ${sdir}/utt2num_frames \
 echo "remove utterances having more than $maxchars or less than $minchars characters"
 # counting number of chars
 if [ -z ${nlsyms} ]; then
-text2token.py -s 1 -n 1 $sdir/text \
+tutils_espnet/ext2token.py -s 1 -n 1 $sdir/text \
```
sw005320: tutils_espnet
->
utils_espnet

sw005320: Actually I think we may not need `utils_espnet/` anymore, right? Can you fix this?

sw005320:
@Fhrozen, you may take a look at this change. It will affect your docker creation.

Fhrozen:
@sw005320, i will modify the docker files after merging this commit. I don't think that the docker creations will be affected, just that they will need to be updated for this commit. I will try to test later to check for any possible issue.

kamo-naoyuki:
Maybe finished, but I didn't test run.sh yet.

sw005320:
OK, please test it with several small tasks (voxforge, an4). Also, test the installation part from scratch.

kamo-naoyuki:
@sw005320 I tested with an4/asr1 and ljspeech/tts1. Maybe no problem now.

sw005320:
Thanks a lot @kamo-naoyuki!

PR #441, opened 2018-10-26
Update travis.yml to use python 3.7

ShigekiKarita:
It needs to be updated because current miniconda env in `tools/conda.mk` uses python3.7 rather than 3.6.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -26,6 +26,7 @@ matrix:
           packages:
             - g++-7
       env:
+        - TRAVIS_PYTHON_VERSION=3.7
         - TH_VERSION="0.4.1"
         - CC="gcc-7"
         - CXX="g++-7"
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,31 +4,27 @@ cache:
   - pip
   - ccache
 
-os: linux
-      
 addons:
   apt:
     sources:
       - ubuntu-toolchain-r-test
     packages:
+      - cmake
+      - python-dev
+      - python3-dev
       - g++-7
 
-matrix:
-  include:
-    - env:
-        - USE_CONDA=false
-        - ESPNET_PYTHON_VERSION=2.7
-    - env:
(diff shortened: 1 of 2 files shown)


ShigekiKarita:
I found that Ubuntu 14.04 on Travis-CI is too old to install python 3.7 via apt. There are two options to use python 3.7 on Travis CI

1. use beta Travis CI option: `xenial: true` (use Ubuntu 16.04)
2. use miniconda

At this time, I prefer the option 2 because we need to test miniconda in `conda.mk` and the option 1 is beta.

ShigekiKarita:
Conda seems to be much faster than default python in Travis-CI: **11:18 -> 6:50** 
I expect that numpy in conda is linked with MKL and faster than the default.

https://travis-ci.org/espnet/espnet/builds/446516421

.travis.yml
```diff
@@ -4,41 +4,38 @@ cache:
...
+
+env:
+  global:
+    - CC=gcc-7
+    - CXX=g++-7
+    - TH_VERSION=0.4.1
+  matrix:
+    - USE_CONDA=false ESPNET_PYTHON_VERSION=2.7
+    - USE_CONDA=true ESPNET_PYTHON_VERSION=2.7
+    - USE_CONDA=true ESPNET_PYTHON_VERSION=3.7
```
kan-bayashi: Why don't you add `USE_CONDA=false ESPNET_PYTHON_VERSION=3.7` ?

ShigekiKarita: In travis, there are no python 3.7 default binary (under 3.6) because its ubuntu 14 is too old. 
https://github.com/travis-ci/travis-ci/issues/9815

Maybe we can use `dist: xenial` to use ubuntu 16 (beta feature)

kan-bayashi: I see.

kan-bayashi:
LGTM

PR #443, opened 2018-10-30
lm script problem in wsj's run.sh

indra622:
In wsj stage 3, lm script calls undefined variable {lmdatadir}
It defined below


sw005320:
Thanks!

PR #446, opened 2018-11-05
[WIP] Writing a recipe for fisher_callhome_spanish

JaejinCho:
This PR starts from a direct copy of egs/wsj to write a recipe for fisher_callhome_spanish


sw005320:
It seems that you upload the WSJ related scripts and not for this corpus. Please fix it.

> How do people usually check non-linguistic symbols in the text of the corpus?

You can just take a look at the transcriptions, then you could find their special syntax (usually <xxx> or (xxx)). Some corpora don't have that.

egs/fisher_callhome_spanish/asr1/run.sh
```diff
@@ -0,0 +1,347 @@
...
+train_set=train_si284
+train_dev=test_dev93
+train_test=test_eval92
+recog_set="test_dev93 test_eval92"
+
+if [ ${stage} -le 0 ]; then
+    ### Task dependent. You have to make data the following preparation part by yourself.
+    ### But you can utilize Kaldi recipes in most cases
+    echo "stage 0: Data preparation"
+    local/wsj_data_prep.sh ${wsj0}/??-{?,??}.? ${wsj1}/??-{?,??}.?
```
Fhrozen: I only find data prep for WSJ, you need to add instructions for this recipe.

sw005320:
@JaejinCho, @hirofumi0810 will make a PR of the speech translation part of this task (and of course it includes ASR part). You could refer his work.

sw005320:
@JaejinCho, please upload only necessary files. There are several binary files and unnecessary files. Remove these files.

egs/fisher_callhome_spanish/asr1/run.sh
JaejinCho: Thanks @Fhrozen. I will update the above suggestions. Shinji's suggestions and some of the above are done from the recent commit ef46731.

Fhrozen:
@JaejinCho, use the .gitgnore file to avoid problems with the unnecessary files. Be careful to not include any necessary file.

JaejinCho:
@sw005320 @Fhrozen 
Okay, I will do it. 
Q: I guess those unnecessary files mean the files not used in run.sh script. The way of doing it to the best of my knowledge is to check the scripts (sub-scripts) used in run.sh and delete those not used manually. However, are there any other ways to do so?

sw005320:
> Q: I guess those unnecessary files mean the files not used in run.sh script. The way of doing it to the best of my knowledge is to check the scripts (sub-scripts) used in run.sh and delete those not used manually. However, are there any other ways to do so?

For me, this is the best way. You have to do it manually to some extent, and also make sure everything work by deleting `data` and execute the data preparation part of `run.sh` from scratch.

JaejinCho:
> Usually, the unnecessary files are created by the script. I saw some that begin with .fish~ or local/split
@Fhrozen @sw005320 Thanks for the comments.  For the local/split directory though, it was not generated by the script. run.sh uses it to split the (callhome_)train_all into train, dev and test once the features are generated. Briefly, it extracts features first from (callhome_)train_all, and then generates data subsets for train, dev and test by using the info in the local/split directory. This is not consistent with how wsj recipe does, which extract features separately for each subset. 

My follow-up question is if there is any recipe that prepares data in this way (not wsj way)?  so that I can make the recipe as consistent as possible with it.

egs/fisher_callhome_spanish/asr1/run.sh
```diff
@@ -0,0 +1,372 @@
...
+
+# feature configuration
+do_delta=false
+
+# network architecture
+# encoder related
+etype=vggblstmp     # encoder architecture type
+elayers=6
+eunits=320
+eprojs=320
```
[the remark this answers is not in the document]
JaejinCho: @Fhrozen Where can I refer to the newest default parameter setting? My forked remote/main is even with upstream/main but it is still not changed as you mentioned here.

JaejinCho:
What's left:
- [ ] need to delete some scripts not used in ./local

egs/fisher_callhome_spanish/asr1/run.sh
```diff
@@ -0,0 +1,372 @@
...
+subsample=1_2_2_1_1 # skip every n frame from input to nth layers
+# decoder related
+dlayers=1
+dunits=300
+# attention related
+atype=location
+adim=320
```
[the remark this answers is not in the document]
JaejinCho: I keep adim=320 since there was no big difference between adim=1024, adim=320 while adim=1024 takes up larger GPU memory.

egs/fisher_callhome_spanish/asr1/run.sh
```diff
@@ -0,0 +1,372 @@
...
+lm_opt=sgd          # adam for character LMs
+lm_batchsize=300    # 1024 for character LMs
+lm_epochs=20        # number of epochs
+lm_maxlen=40        # 150 for character LMs
+lm_resume=          # specify a snapshot file to resume LM training
+lmtag=              # tag for managing LMs
+
+# decoding parameter
+lm_weight=1.0
+beam_size=30
```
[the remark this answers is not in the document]
JaejinCho: changed to beam_size=20

egs/fisher_callhome_spanish/asr1/run.sh
```diff
@@ -0,0 +1,372 @@
...
+lsm_weight=0.05
+
+# minibatch related
+batchsize=30
+maxlen_in=800  # if input length  > maxlen_in, batchsize is automatically reduced
+maxlen_out=150 # if output length > maxlen_out, batchsize is automatically reduced
+
+# optimization related
+opt=adadelta
+epochs=15
```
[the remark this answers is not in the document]
JaejinCho: I would leave epochs=15 for now. I think epochs=10 is a tight saturation point

egs/fisher_callhome_spanish/asr1/path.sh
```diff
@@ -0,0 +1,16 @@
+MAIN_ROOT=$PWD/../../..
```
sw005320: Please follow `path.sh` in the other recipes. We changed it recently #464

egs/fisher_callhome_spanish/asr1/run.sh
```diff
@@ -0,0 +1,367 @@
...
+
+# rnnlm related
+use_wordlm=true     # false means to train/use a character LM
+lm_vocabsize=65000  # effective only for word LMs
```
sw005320: If you only use the transcriptions, 65k would be a bit too large, but you can test it later.

egs/fisher_callhome_spanish/asr1/run.sh
```diff
@@ -0,0 +1,367 @@
...
+# Set bash to 'debug' mode, it will exit on :
+# -e 'error', -u 'undefined variable', -o ... 'error in pipeline', -x 'print commands',
+set -e
+set -u
+set -o pipefail
+
+train_set=train
+train_dev=dev
+train_test=test
+recog_set="dev test callhome_dev callhome_test callhome_train"
```
sw005320: why do you need to recognize `callhome_train`?

egs/fisher_callhome_spanish/asr1/path.sh
JaejinCho: Done

egs/fisher_callhome_spanish/asr1/run.sh
JaejinCho: Yes. I agree with you as we also talked during the meeting. Let me test this some time.
I would like to make sure this (i.e. setting larger vocab size than vocabularies seen in the training data) doesn't affect the decoding result. Do you know about this @takaaki-hori ?

egs/fisher_callhome_spanish/asr1/run.sh
JaejinCho: I just followed the standard recipe in the current version of Kaldi in Github

PR #448, opened 2018-11-06
Docker

Fhrozen:
Updates for docker container:
- Update requirement .txt on all containers. (editdistance becomes necessary for training now)
- Update container CUDA 9.0 for CUDNN 7.3, CUDA 9.1 & 9.2 have not been updated yet on docker hub.
- Add CUDA 10.0 script. However, this is no yet implemented in docker hub. Chainer and Pytorch have not officially release the support. Pytorch seems to have no problem with 10.0. 

Possible future changes:
- From CUDA 9.2 change the OS version to Ubuntu 18.04. Only if necessary.


Fhrozen:
I finished we the updates and fixes for docker containers. All docker containers have been updated in the Docker HUB. 
@sw005320 I am wondering about the manage of the modularization. Currently, the files in bin folder are moved into the venv folder, so when these files are modified they will need to be updated in folder too. It is not better to do a link to this files inside the venv folder, so when they are modified, the modifications will be reflected in the test. Otherwise, it will be necessary to copy them when there are any modification.

kan-bayashi:
I think current installation of espnet module is performed by `pip install -e`.
Therefore, espnet module is not copied into `venv/<path>/<to>/site-packages`.
And the modification of codes in `espnet/espnet` is directly reflected.

sw005320:
Is that OK @Fhrozen?

Fhrozen:
Is fine, my concern is about the files located at the espnet/bin folder (which are moved into the venv). Those files are not reflected. Of course, is there is any new network or anything else in other folders will be no problem. But, if you do a new asr_recog for example, then this will need to be copied into the venv folder. Am I wrong?.

sw005320:
I see your point.
Let me correct my understanding.
During the installation process, https://github.com/espnet/espnet/blob/master/tools/Makefile#L27 will execute https://github.com/espnet/espnet/blob/master/setup.py. 
Then, https://github.com/espnet/espnet/blob/master/setup.py#L58 will make scripts in `espnet/bin` ready for use.

But during some developments if we create a new script (say `espnet/bin/hoge.py`), do we need to execute `pip install -e ..` again or not.
I don't fully understand `setup.py`, and not sure we need this or not. @kamo-naoyuki and @kan-bayashi , can you answer it?

sw005320:
Or we can test them easily.

kamo-naoyuki:
> But during some developments if we create a new script (say `espnet/bin/hoge.py`), do we need to execute `pip install -e ..` again or not.

Sorry, I didn't notice it. Yes, you need to execute `pip install` again every time you modify.  As far as I know, the copying for the scripts is inevitable. 

I think there are no ways other than setting `PATH` as in the past to work with a new scripts.

As far modification for existing scripts,  [console_scripts](https://setuptools.readthedocs.io/en/latest/setuptools.html#automatic-script-creation), instead of `scripts`, solves this problem.  In this case, the commands invoke a function from the python files, that are not copied.

sw005320:
OK. This does not sound like development friendly. Can you modify it based on the `PATH` solution? (Is there any other solutions?).

kamo-naoyuki:
( I can't commit any codes in weekend always)

`PATH` may be better because it is simple. 

It may depend on preference, but  how about `-m` of Python interpreter?

```bash
python -m espnet.bin.asr_recog ...  # Same as "python /somewhere/espnet/bin/asr_recog.py"
```

Or, if new scripts are rarely appended, `console_scripts` is also O.K. The modification is reflected, as I said. This seems beautiful as a python module, but, umm, there is no benefit comparing with `PATH`.

sw005320:
It sounds like `PATH` solution is the best solution.

Fhrozen:
Docker env tested in AN4, for chainer/python with new configurations (Chainer 5.0, espnet module).
Found no problem.
Containers updated at docker Hub.
Ready to merge =)

sw005320:
Thanks!

Issue #449, opened 2018-11-06
Big dataset cause mergejson.py out of memory


fanlu:
My dataset's volumn is 2900h, all data is 8k telephone audio, and do perturb to 3 times than original.
the part of mergejson.py in data2json.sh recipe will cause an out of memory error.
```
Traceback (most recent call last):
  File "/espnet/egs/kefu/asr1/../../../src/utils/mergejson.py", line 74, in <module>                                                                                  
    jsonstring = json.dumps({'utts': new_dic}, indent=4, ensure_ascii=False, sort_keys=True).encode('utf_8')                                                                                      
  File "/miniconda3/envs/py2/lib/python2.7/json/__init__.py", line 251, in dumps                                                                                      
    sort_keys=sort_keys, **kw).encode(obj)
  File "/miniconda3/envs/py2/lib/python2.7/json/encoder.py", line 210, in encode                                                                                      
    return ''.join(chunks)
MemoryError
```
So can we change the style of data.json file to one json per line? @sw005320

sw005320:
Thanks for pointing out it.
This is an expected issue.
We'll work it later, but if you solve it and make a PR, that is great.

sw005320:
#450 solved the issue.

PR #450, opened 2018-11-07
big dataset cause json.dumps MemoryError

fanlu:
#449 
change json.dumps to json.dump can solve MemoryError what I have met before. and I think this is a temporary solution.


src/utils/mergejson.py
```diff
@@ -17,6 +17,8 @@
                         help='Test the json file for multiple input/output', default=0)
     parser.add_argument('--verbose', '-V', default=0, type=int,
                         help='Verbose option')
+    parser.add_argument('--output_json', default='', type=str,
```
sw005320: `--output_json`
->
`--output-json`
If we try to follow the ESPnet convention

src/utils/mergejson.py
```diff
@@ -71,5 +74,10 @@
             unicode('utt2spk', 'utf-8'):dic[unicode('utt2spk', 'utf-8')]}
     
     # ensure "ensure_ascii=False", which is a bug
-    jsonstring = json.dumps({'utts': new_dic}, indent=4, ensure_ascii=False, sort_keys=True).encode('utf_8')
-    print(jsonstring)
+    if not args.output_json:
```
sw005320: How about the following?
Then, we can still use a pipe.
```
    if args.output_json:
	with codecs.open(args.output_json, "w", encoding='utf-8') as json_file:
            json.dump({'utts': new_dic}, json_file, indent=4, ensure_ascii=False, sort_keys=True, encoding="utf-8")
    else:
        sys.stdout = codecs.getwriter('utf8')(sys.stdout)
        json.dump({'utts': new_dic}, sys.stdout, indent=4, ensure_ascii=False, sort_keys=True, encoding="utf-8")
```

Also, make sure to import sys and codec in the header
```
import argparse
import codecs
import json
import logging
import sys
```

Issue #451, opened 2018-11-07
check_install.py : editdistance version problem


b-flo:
Hi,

 In the lastest version it seems check_install.py will fails upon editdistance version if the package was installed with pip as the \__version__ attribute is not loaded (the [setup.py](https://github.com/aflc/editdistance/blob/master/setup.py) file is missing when installing with pip)

```
INFO: --> editdistance has no version info, but version is specified.
INFO: --> maybe it is better to reinstall the latest version.
INFO: --> warpctc_pytorch version is matched.
INFO: library version check done.
INFO: 4 / 5 libraries are correct version.
INFO: please try to setup again and then re-run this script.
```

I don't know if someone experienced the same issue?
Using pkg_resources to get the version for this specific case works :

```
if name == 'editdistance' :
   assert pkg_resources.get_distribution(name).version == version             
   logging.info("--> %s version is matched." % name)
   is_correct_version_list.append(True)
elif ...
```

```
INFO: --> editdistance version is matched.
INFO: --> warpctc_pytorch version is matched.
INFO: library version check done.
INFO: 5 / 5 libraries are correct version.
```

sw005320:
@kan-bayashi, can you fix it?

kan-bayashi:
Due to the change of the installl (#440), current `check_install.py` is not compatible with current master.
I will fix it.

kan-bayashi:
Fixed in #455. But not smart way.
I will consider better way for the current installation.

b-flo:
I made a local fix by checking if a version match either through pkg_resources or importlib, passing an error message to assert containing a tag for the module used, when raising the error the correct module is selected. It works well and can handle mismatches between local/global installations but as if it's a bit... ugly.

Forgot to tell you about the missing parameter in logging.warn (l.88) glad you saw it!

Thanks anyway, marked as solved!

PR #455, opened 2018-11-08
Fix Makefile

kan-bayashi:
This PR fixes following problems:
- fixed Makefile to install torch in the case of w/o conda
- fixed .gitignore to ignore some files related to new installation. e.g. `.eggs`, `*.done`, and so on.
- fixed check_install.py (#451)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,6 +4,9 @@
 .\#*
 *DS_Store
 slurm-*.out
+.eggs/
+.pytest_cache/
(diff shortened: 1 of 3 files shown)


sw005320:
Actually, now we stopped to use `requirements.txt`, and we have to fix some documents related to it, e.g., https://github.com/espnet/espnet/search?q=requirements.txt&unscoped_q=requirements.txt

PR #457, opened 2018-11-08
[WIP] wsj_2mix added

simpleoier:
2-mix features:
1. data preparation for 2-speaker mixing speech
2. pytorch training script for 2 or 3 speakers
3. with/without speaker parallel attention are both available
4. scoring script for 2 speakers
5. _data_prep and scoring_ for 3 speakers are partly available.


sw005320:
Thanks @simpleoier. This is a good start. My major requests (I guess you're thinking) is to move/merge/modify several main python scripts to meet the ESPnet standard directory structure. For example,

- local/src/asr/asr_pytorch.py
move to `espnet/asr_mix/asr_mix_pytorch.py`
- local/src/asr/asr_utils.py
better to move to `espnet/asr_mix/asr_mix_utils.py`. Also try to minimize to use duplicated codes.
- local/src/bin/asr_train.py
move `espnet/bin/asr_mix_train.py`
- local/src/nets/e2e_asr_common.py
It may be merged to `espnet/nets/e2e_asr_common.py` or create `espnet/nets/e2e_asr_mix_common.py`, or merged to `espnet/nets/e2e_asr_mix_th.py`
- local/src/nets/e2e_asr_th.py
This one should be more organized. First, it should be in `espnet/nets`, e.g., `espnet/nets/e2e_asr_mix_th.py`. Also, it would be better clean up to remove some redundant part.

We may not have to have `local/src/utils` directory. You can just put it in `local`
- local/src/utils/compute_perm_free_error.sh -> local/compute_perm_free_error.sh
- local/src/utils/json2trn.py -> local/json2trn.py
- local/src/utils/mergeresultjson.py -> local/mergeresultjson.py
- local/src/utils/minpermwer.py -> local/minpermwer.py
- local/src/utils/result2json.py -> local/result2json.py
- local/src/utils/score_sclite.sh -> local/score_sclite.sh

Also try to remove redundant parts, and re-use existing modules.

egs/wsj_mix/asr1/path.sh
```diff
@@ -0,0 +1,16 @@
+MAIN_ROOT=$PWD/../../..
```
sw005320: You may refer `path.sh` in the other recipe.

src/asr/asr_mix_pytorch.py
```diff
@@ -0,0 +1,426 @@
...
+
+# matplotlib related
+import matplotlib
+import numpy as np
+matplotlib.use('Agg')
+
+REPORT_INTERVAL = 100
+
+
+class CustomEvaluator(extensions.Evaluator):
```
sw005320: import this class from `asr_pytorch.py`

src/asr/asr_mix_pytorch.py
```diff
@@ -0,0 +1,426 @@
...
+                    #    will be converted to chainer variable later
+                    x = self.converter(batch, self.device)
+                    self.model(*x)
+                summary.add(observation)
+        self.model.train()
+
+        return summary.compute_mean()
+
+
+class CustomUpdater(training.StandardUpdater):
```
sw005320: same comment.

sw005320:
I think it would be great if you also consider to extend utils/data2json.sh, utils/json2trn.py, utils/mergejson.py, utils/score_sclite.sh with multiple speaker cases rather than making modified versions in the local directory. But if it makes it complicated, you can leave them as they are.

sw005320:
@simpleoier, can you fix the conflicts? We have large amounts of changes for v.0.3.0 (#484). I also want to include this recipe in the v.0.3.0.

sw005320:
Can you include the script of generating the WSJ mixture? e.g., 
1. download http://www.merl.com/demos/deep-clustering/create-speaker-mixtures.zip
2. generate the mixture data
If it makes it too complicated, you don't need it, but even in this case, it's better to mention somewhere that you need to generate the mixture with MERL's script.

Issue #456, opened 2018-11-08
How to install requirements related to test or doc


kan-bayashi:
In the current master, `test_requirements.txt` and `requirements.txt` are replaced by `setup.py`.
We can use `python setup.py test` to run `pytest`, but requirements for test is not installed in virtualenv.
Therefore, we cannot use linter libraries such `flake8` or `autopep8` written in test_requires of `setup.py`.

@kamo-naoyuki How do we install test or doc requirements written in `setup.py`?

kan-bayashi:
What I want to do is to update the contribution guide.
https://github.com/espnet/espnet/blob/master/CONTRIBUTING.md
I will update it if I understand how-to-install test requirements.

kamo-naoyuki:
You can install them manually as following,

```bash
pip install -e $REPOSITORY[test]
pip install -e $REPOSITORY[doc]
```

However,  `python setup.py test` also should install the requirements in my understanding. Could you check it again?  @kan-bayashi

kamo-naoyuki:
We also need to update https://github.com/espnet/espnet/blob/master/doc/README.md.

kan-bayashi:
I will close.

PR #458, opened 2018-11-09
Fix manuals and makefile

kan-bayashi:
#456

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -27,7 +27,8 @@ TBD
 
 ESPnet's testing is located under `test/`.  You can install additional packages for testing as follows:
 ``` console
-$ pip install -r tools/test_requirements.txt
+$ cd <espnet_root>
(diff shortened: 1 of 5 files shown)


Fhrozen:
I am having the error: `ImportError: No module named kaldi_io_py`.
Not sure how the recipe show work, because I see that kaldi-io is installed but not inside the venv folder. And because there is no more PYTHONPATH, it cannot be load from another folder. 
Should it not be installed in the venv folder? or any additional step is required?

kan-bayashi:
I had the same problem.
I reinstalled `kaldi_io_py` via
```
cd tools 
source venv/bin/activate
rm kaldi-io-for-python.done
make kaldi-io-for-python.done
```

kan-bayashi:
Hi @Fhrozen.
I found the bag in `Makefile`.
In the installation of kaldi_io_py, virtualenv is not activated.
I fixed it.

PR #465, opened 2018-11-12
[not-for-merge] Transformer

Fhrozen:
I am currently working on the Transformer for ASR (https://arxiv.org/pdf/1706.03762.pdf). I am implementing it with minimum changes from the original script. If it works, I will adapt it to fit into the E2E module, meanwhile I will keep using a different script (e2e_transformer).

I am currently testing with voxforge dataset with CPU, because I am having some memory issues. It seems that the multihead attention layers consume a huge amount of memory. When the model was training for MNT (utters with a length of ~50) the model consumed 5GB of GPU. For ASR, the input length has more the 100, and so the model requires more than 20 GB of GPU for training with reduced parameters.

I am training with:
`./run.sh --stage 3 --ngpu 0 --verbose 1 --backend chainer --mtlalpha 0.0 --elayers 3 --batchsize 20 --maxlen_in 500 --epochs 2`

Some changes: 
Implemented subsampling in the encoder (/4) I will test it also with subsampling by the may cause a memory error.
Layers for Enc and dec reduced to 3. Once more, to avoid memory issues.

TODO:
- Implement recog script.


sw005320:
Did you implement warmup? https://github.com/soskek/attention_is_all_you_need/blob/master/train.py#L270-L283
Several people told me that the transformer does not work without this trick.

Fhrozen:
Just implemented. WarmUp

Fhrozen:
@sw005320 the training with GPU is completed. The command required to execute the training was:
`./run.sh --stage 3 --ngpu 1 --verbose 1 --backend chainer --mtlalpha 0.0 --elayers 2 --batchsize 6 --maxlen_in 450 --epochs 30 --opt adam`

I am also attaching the training graphs:
- ACC:
acc
- LOSS:
loss

The model seems to be overfitted after the 17 epoch.
I will try other parameters. However, the training is taking more than 24hrs for voxforge. VGG-based took around 2 hours.

sw005320:
Which parameters will you tune? @soskek told me that we should use small warmup steps when the batch size is small.

soskek:
I talked with @sw005320 and think larger (rather than smaller) warmup and decay steps can be effective. Training with small minibatches basically takes more update steps until good convergence so maybe also takes more warmup and decay steps.

sw005320:
Sorry small -> large. @soskek, thanks for the correction!

sw005320:
@Fhrozen, what is your status?

Fhrozen:
Had some problems with pc. Just finished to add some updates.
Will report as soon as get new results

Fhrozen:
Finished some additional experiments.
Implemented VGG with subsampling /4 which allowed to use a deeper transformer (enc 3L, dec 3L) and a larger bs (20). Also doubled the number of warmup (64000 > 128000). Each epoch takes around 900 iters.

Did some test with few iters for warmup also, and found that the model can improve the performance really fast comparing to longer warmup. However, the model seems to be overfitting, the valid acc only performs to 30%.

loss
acc

With larger warmup the network seems to keep the acc for both (train and valid set) for a longer training. But, it still is overfitted after several epocs. 

loss2
acc2

 I am now training a model with dropout. After this, I will train with fewer layers at the dec and more layers at the enc.

sw005320:
@Fhrozen, great progress! (but we should work more.)
Can you make sure that the problem is purely due to the overtraining?
If so, another possibility is to use larger database (Librispeech).
Also, please make sure to work on a single GPU for now.
We have to avoid some irregular issues come from multiple GPUs.
Also, I'm not so confident that the dropout solves this problem.

@soskek, did you observe such learning curve behaviors?
Could you give us some advice?

Fhrozen:
@sw005320 About the progress: Each training (even for voxforge) takes from 12hrs (30epochs) up to 24hrs (60 epochs). The training is quite slow.

> Can you make sure that the problem is purely due to the overtraining?

I will try to debug the model and test another possible architectures.
> If so, another possibility is to use larger database (Librispeech).

This may take a couple of months for transformer, in the way is implement now. >.<. I was planning to test with another corpus, such as CHiME 4 or WSJ?
> Also, please make sure to work on a single GPU for now.

I am working for single GPU for now, My other GPUs are already busy. T.T
> Also, I'm not so confident that the dropout solves this problem.

As you said. seems that dropout does not improve the result. I trained this model for +120 epocs with dropout 0.2 at the encoder with 3 layers, and 0.4 at the decoder with 1 layer, but it is not working. I will test once more with this corpus and some modifications in the encoder (also check english) and then move to another corpus.
acc
loss

I will redo the experiments, finish with the recognition part and prepare a spreadsheet to track the experiments.

sw005320:
Given https://arxiv.org/pdf/1804.00247.pdf (@soskek introduced it to me), we may need to tune the learning rate as well. I don't think we should switch to other techniques or implementations if you don't have any bugs, but just try to tune 
1. learning rate
2. batch size
3. warmup step

Also, I'm curious why it takes long time for training. As far as I understand, the transformer eliminates the for loop over time, and it will make the computation faster.

soskek:
Great!
I heard that Transformer tended to overfit small datasets rather than other models. (though I forgot the source...) Tuning the hyperparameters may be important.

ShigekiKarita:
@Fhrozen I made `ConvolutionSentence` faster a little. We can use L.Linear instead of L.Convolution2d with `ksize=1, stride=1, pad=0` (actually your setting always uses that) like this
```python
class ConvolutionSentence(chainer.Chain):
    """Position-wise Linear Layer for Sentence Block

    Position-wise linear layer for array of shape
    (batchsize, dimension, sentence_length)
    can be implemented a convolution layer.

    """

    def __init__(self, in_channels, out_channels,
                 ksize=1, stride=1, pad=0, nobias=False,
                 initialW=None, initial_bias=None):
        super(ConvolutionSentence, self).__init__()
        with self.init_scope():
            self.use_linear = ksize == 1 and stride == 1 and pad == 0
            if self.use_linear:
                self.layer = L.Linear(in_channels, out_channels, nobias, initialW, initial_bias)
            else:
                self.layer = L.Convolution2D(
                    in_channels, out_channels,
                    ksize, stride, pad, nobias,
                    initialW, initial_bias)


    def __call__(self, x):
        """Applies the linear layer.

        Args:
            x (~chainer.Variable): Batch of input vector block. Its shape is
                (batchsize, in_channels, sentence_length).

        Returns:
            ~chainer.Variable: Output of the linear layer. Its shape is
                (batchsize, out_channels, sentence_length).

        """
        if self.use_linear:
            x = x.transpose(1, -1)  # (B, H, W, C)
            y = self.layer(x, n_batch_axes=x.ndim-1)
            y = y.transpose(1, -1)  # (B, C, H, W)
        else:
            if x.ndim < 4:
                x = F.expand_dims(x, axis=3)
            y = self.layer(x)
            y = F.squeeze(y, axis=3)
        return y
```

In your voxforge recipe with gtx1080ti, 
- batchsize=6: 0.79516 -> 0.90591 iters/sec
- batchsize=60: 0.55292 -> 0.7118 iters/sec

Fhrozen:
@ShigekiKarita Thank you for your support. I will update the code with your suggestion.

Fhrozen:
Implemented modifications based on http://www.mirlab.org/conference_papers/International_Conference/ICASSP%202018/pdfs/0005884.pdf.
The model is training with the following parameters:
`./run.sh --stage 4 --ngpu 1 --verbose 1 --dunits 256 --eunits 256 --backend chainer --mtlalpha 0.0 --aconv_chans 4 --elayers 6 --dlayers 6 --opt adam`

The training is using some of the parameter and implementations proposed by the paper, currently, to achieve the base WER for WSJ.

Also added the position embeding. This seems to be necessary. 
Will report as soon as I get new results.

Fhrozen:
Just finished a training with the new implementations, but it seems to not be working.

acc

loss

Will check once more to fix any possible mistake

Fhrozen:
Just to test, I started to train a model for WSJ with adadelta and the acc of the model is better than using adam.

acc_delta
loss_delta

This is the curves for the same model (same as base from prev. paper but without positional encoder) with adam, and there is a big difference:

acc
loss

Which are your opinions? The training should be finishing tomorrow, so then I will start with the decoding part.

sw005320:
> Which are your opinions? The training should be finishing tomorrow, so then I will start with the decoding part.

I have no ideas about the difference between adadelta and adam, but I have similar experience in the early stage of the LSTM based encoder-decoder probably due to the lack of expertise of making it work. You could do it with adadelta for now. My concern is that even for the adadelta case, the validation accuracy would not look great. With ~50% in the WSJ task, it may only train the decoder network, and does not tray attention and encoder networks. Something would be wrong. The validation accuracy should be >95%.

ShigekiKarita:
When I ran WSJ with your previous transformer impl. It results 94% valid acc 
tf_acc

It seems to converged super fast while it took 15 hours overall in my environment.
here is the script https://github.com/ShigekiKarita/espnet/blob/83115718d25ad9c512fea115617f57bc4b01b55d/egs/wsj/asr1/run_transformer.sh

ShigekiKarita:
Here is our baseline with final valid acc 95%. It took only 5 hours with pytorch impl.

acc

Fhrozen:
@ShigekiKarita, thank you for your support. Currently, I am having some problems with the new updates so I will roll back. Also, I am wondering if you these results are obtained with adadelta or adam.

ShigekiKarita:
It was Adadelta. I just run `./run_transformer.sh --wsj0 ... --wsj1 ...`. Every configuration is written in https://github.com/ShigekiKarita/espnet/blob/83115718d25ad9c512fea115617f57bc4b01b55d/egs/wsj/asr1/run_transformer.sh
I hope this commit could help you to roll back.

ShigekiKarita:
Also you can see my diff from your previous commit "add vgg & increase warmuip" here https://github.com/ShigekiKarita/espnet/commit/83115718d25ad9c512fea115617f57bc4b01b55d

Fhrozen:
@ShigekiKarita , thank you, I am updating now the code.
BTW, there is something strange about warmup and is that this should no work for adadelta. In the chainer code, there is no attr call alpha in Adadelta, this only should work for Adam. Could you please verify if the use of warmup for adadelta is necessary. It may be means that the model can be trained without warmup when using adadelta.

ShigekiKarita:
Your expectation is right. In current setting, I only use Adadelta without warmup. When I did voxforge exp, I tried several warmup settings with Adam but it did not help. The mtlalpha is just copied from baseline settings but it did not enabled inside the code as you said.

Fhrozen:
@ShigekiKarita I am working with your modifications, and I have the problem with transpose in chainer. The output says that there are invalid axes values [1, -1]. I tried to with numpy also and when the array has more than 4 axes I need the four dimensions. Do I miss something?.

ShigekiKarita:
Sorry, I was confused about `transpose` in chainer and pytorch. It should be `swapaxes` in chainer. I have not pushed the fix yet.
- `torch.transpose` is `chainer.swapaxes`
- `torch.permute` is `chainer.transpose`

sw005320:
@ShigekiKarita, this is great!!! It seems that the transformer is working correctly. We may require some tuning though.

ShigekiKarita:
I found my experiment was wrong. When I checked `numpy.load("model.acc.best").keys()`, there are `espnet.asr.e2e_asr.E2E` params. Although `model.json` says `ntype=transformer`, the model was `e2e_asr.E2E`. I'm searching the bug in current implementation or script.

ShigekiKarita:
I installed different location. So it was not transformer branch! omg

Fhrozen:
Once more, I roll back the code, and finish some experiments.
According to the code of @ShigekiKarita, one of the modifications was to not use lsm. So, I tried to executed it withou lsm.
acc_no_lsm
loss_no_lsm

However, the acc seems to improve (slightly) when lsm is active.
acc
loss

Now, I trying to improve the results to obtain at least +90% in the validation.

sw005320:
I think using LSTM or not is not a matter. The problems must be in others (again tuning or bugs (especially padding?))

ShigekiKarita:
Hi, 

I start pytorch transformer implementation today. It is not exact copy of chainer version but it seems to be faster than chainer as usual (estimated time to finish: 4 hours in WSJ). Maybe this implementation is better to try several settings.
- https://github.com/ShigekiKarita/espnet/blob/pytorch-transformer/espnet/nets/e2e_asr_transformer_th.py
- https://github.com/ShigekiKarita/espnet/blob/pytorch-transformer/egs/wsj/asr1/run_transformer_th.sh

wip accuracy (I will update this later)
pytorch_acc

At this time, I confirmed that the saved model parameters are transormer's. :sob:

Fhrozen:
@ShigekiKarita, Thank you for your support, I will like to confirm something.
How much memory are you using for the training? The subsampling seems to be working with a linear function instead of poolings or strides in conv layers. https://github.com/ShigekiKarita/espnet/blob/b9fe5b70be53e0758d6f8bad25fe2d54b3affe20/espnet/nets/e2e_asr_transformer_th.py#L172. Did you implement any additional subsampling instead of the linear?. Probably I will need to replace the Conv2d x a linear.

sw005320:
@ShigekiKarita, this looks promising, but I'm not sure why the speed is so different. In the RNN case, the decoder network is based on "for loop", which makes chainer slower. But the transformer does not have such "for loop", and I thought that pytorch and chainer would not have big differences in terms of the speed.

ShigekiKarita:
@Fhrozen 

> How much memory are you using for the training? 

nvidia-smi says around 8-9GB. 

> The subsampling seems to be working with a linear function instead of poolings or strides in conv layers. Did you implement any additional subsampling instead of the linear?. Probably I will need to replace the Conv2d x a linear.

No, I do **not** conduct **any** subsampling in my implementation now.

@sw005320 

> I'm not sure why the speed is so different. In the RNN case

Actually, transformer is much faster when the model size and minibatch size is equal to RNN. However, transformer seems to require more parameters than RNN in previous literature http://www.mirlab.org/conference_papers/International_Conference/ICASSP%202018/pdfs/0005884.pdf (and my current setting). It results smaller minibatch parallelism (e.g., 16 in the run_transformer_th.sh). These are the main reason why transformer gets as fast as RNN at this time. Anyway, I will support multi GPU training to enable larger minibatch.

ShigekiKarita:
I would like to share my current expectations and results.
- Using adadelta, I could make transformer 90% acc only with pytorch initialization (maybe a.k.a. LeCun uniform). LeCun normal, Xavier, and Kaiming initialization result 20 - 60% acc.
- Using adam with transformer's scheduler, I could make it 90% only with Xavier uniform.
- Maybe adam with the lr scheduler is faster than adadelta to converge. It seems to prefer larger minibatch (batchsize 128, lr_init 2.0, warmup_steps 1000)
acc

- Even around 90% acc, source-to-target attention is still not diagonal.

epoch 1
4k0c0302 decoder decoders 0 src_attn ep 1

epoch 27
4k0c0302 decoder decoders 0 src_attn ep 27

creatorscan:
In chainer we observed, major changes in acc by varying dropout:

As @ShigekiKarita  pointed out, this is observed with LeCunNormal or GlorotNormal initalization.

Dropout : 0.3
acc_ch
Dropout:   0.5 (makes the training acc follow validation acc)
acc_bmk

And the attention weights are left unchanged during training:
Epoch 1:
4k0c0301 ep 1
Epoch 10:
4k0c0301 ep 10

Fhrozen:
@creatorscan, thank you for your support, I am currently checking for the backprop. 
Looking at the computing graphs, I could not find any problem. Also, all the classes are initialized with init_scope, so the problem should be in another part. about the dropout, the valid curve still raises more than 0.5?. Even without dropout, it can get around 0.6.
[cg.zip](https://github.com/espnet/espnet/files/2657624/cg.zip)

creatorscan:
@Fhrozen Yes Nelson, it raises above 0.5 after 7th epoch and continues till 0.6 when it reaches 10th epoch. But, I found training accuracy shoots up (>0.75) while valid acc cannot (which is not in the case of Shigeki's plot).

ShigekiKarita:
@creatorscan It looks like positional encoding instead of attention weight (?)

This is how positional encoding.
```python
import torch
import espnet.nets.e2e_asr_transformer_th as T
pe = T.PositionalEncoding(100, 0.0)
x = pe(torch.zeros(1, 150, 140))
plt.imshow(x[0].numpy().T)
```

pe

creatorscan:
@ShigekiKarita yes you are right. Thanks for pointing it out.

espnet/nets/e2e_asr_transformer.py
```diff
@@ -0,0 +1,900 @@
...
+
+        # Calculate Weighted Sum
+        batch_A, batch_V = F.broadcast(
+            batch_A[:, None], batch_V[:, :, None])
+        batch_C = F.sum(batch_A * batch_V, axis=3)
+        assert(batch_C.shape == (batch * h, n_units // h, n_querys))
+        C = F.concat(F.split_axis(batch_C, h, axis=0), axis=1)
+        assert(C.shape == (batch, n_units, n_querys))
+        C = self.finishing_linear_layer(C)
+        return C
```
ShigekiKarita: for monitoring the soft alignment, it is nice to `return C, batch_A` and use `batch_A` instead of `C` in `calculate_all_attentions`

Fhrozen:
@ShigekiKarita , that is pretty fast. Using chainer, takes me around 7hrs for 6 epochs. I will need to reduce some parts to accomplish the training time of pytorch.

ShigekiKarita:
@Fhrozen  However, something is wrong with my implementation because the attentions are still not diagonal. You are very helpful for me to search the best (fast and accurate) implementation in different way.

These attention weights are in the last encoder/decoder layers.

- 4k0c0302 encoder encoders 5 self_attn ep 50.png

4k0c0302 encoder encoders 5 self_attn ep 50

- 4k0c0302 decoder decoders 5 self_attn ep 50

4k0c0302 decoder decoders 5 self_attn ep 50

note: this decoder self attention is masked not to see future input tokens

- 4k0c0302 decoder decoders 5 src_attn ep 50
4k0c0302 decoder decoders 5 src_attn ep 50

ShigekiKarita:
In my understanding, there might be only two parts to consume lots of computation time in Transformer.

- PositionwiseFeedForward
- MultiHeadAttention

Maybe current these chainer implementations unnecessary memory overhead ops.  I think concat and stack ops makes the program very slow because it always allocates output memory unlike split, transpose, and index ops. In addition, **reshape** ops also allocates memory if transposed tensor is not contiguous. I'm always trying to reduce this kind of memory overhead ops. For example, maybe I will do these.

- Instead of (batchsize, dimension, sentence_length) tensor, use (batchsize, sentence_length, dimension) to remove non-contiguous reshape overhead in `seq_func`. Here, the (b, l, d) -> (b * l, d) does not allocate memory when the (b, l, d) tensor is contiguous unlike the (b,d,l) -> (b * l, d).

- In Linear layer, always use `n_batch_axes=2` for the sequential input to remove the reshape overhead.

- Instead of these split -> concat, use contiguous reshape as discussed above.

https://github.com/espnet/espnet/blob/cdb43ea0d9ab68af3ef4079c6004f272051ac0dc/espnet/nets/e2e_asr_transformer.py#L556-L561

Fhrozen:
@ShigekiKarita, thank you for the support, I was also thinking about that. Trying to reduce the number of ops and avoid reshaping or others. I am coding that now, I also was planning to use ConvND with 1D, instead of 2D. I will check a little further a current training an implement the above.

ShigekiKarita:
Yes, conv1d is also good for removing reshape ops. I think changing conv2d -> conv1d might not effect because cuDNN does not have any 1d specific algorithm. See p81 for the implemented algorithm table https://docs.nvidia.com/deeplearning/sdk/pdf/cuDNN-Developer-Guide.pdf

ShigekiKarita:
I found librispeech setting in tensor2tensor https://github.com/tensorflow/tensor2tensor/blob/edd7fb2cdb49b9726f2ab2d7ea655a28677c34b6/tensor2tensor/models/transformer.py#L2183

Fhrozen:
@ShigekiKarita I also notice that the Linear employs bias in the MultiheadAttention, while the implementation in https://github.com/soskek/attention_is_all_you_need/blob/master/train.py#L270-L283 does not. This could be another possible failure about the chainer implementation?.

ShigekiKarita:
maybe this https://github.com/espnet/espnet/pull/465#issuecomment-445425958 is for @soskek

Fhrozen:
I just finished with some updates. 
Now the transformer takes 2~3 hrs for 10 epochs. I tried to transcribe most of @ShigekiKarita 's code. 

Using adadelta and the recommended parameters by @ShigekiKarita  `./run.sh --ngpu 1 --verbose 1 --adim 256 --backend chainer --mtlalpha 0.0 --aheads 2 --elayers 6 --dlayers 4 --beam_size 5 --stage 4 --maxlen_in 500 --batchsize 16 --epochs 30 --dropout 0.2 `, I obtained these plots. I will wait until finish to run the decode part.

acc
loss

Fhrozen:
After training for 60 epochs, the model is showing this plots. It seems that can be improved so I will train it for 100 epochs.
Currently 60 epochs takes 14hrs. So, I will try to check trying to reduce the training time.

acc

loss

ShigekiKarita:
image

 I also observed similar result but I used adam with lr scheduer (called noam).
`./run_transformer_th.sh --batchsize 32 --epochs 50 --accum_grad 1 --opt noam --warmup_steps 8000 --lr_init 10 --ninit xavier_uniform --dropout 0.1 --input_layer conv2d --maxlen-in 1024 --lsm_weight 0.1 `
As noam has more configurations than adadelta, it can be better. I expect that larger minibatch, larger lr, and smaller warmup make converge faster. However, I'm still not sure.

ShigekiKarita:
As it seems to around 95%, I'm gonna write beam search in pytorch. It will implement new interface discussed in #489.

ShigekiKarita:
Hi, I would like to report my progress in this weekend. I summarize my current fast and stable setting.

https://github.com/ShigekiKarita/espnet/blob/b8a2a5aeacdb46c40647a2bb9ce2318863d95e50/egs/wsj/asr1/run_transformer_th.sh

After a few trial in my single GTX1080ti at home, I found these tips
- use Adadelta with lr=10. maybe chainer does not have this option but it seems to be critical at this time.
- use both dropout=0.1 and lsm_weight=0.1. these seem to be essential to stabilize training.
- (optional) use largest batchsize as possible in GPUs e.g., `--batchsize 64 --maxlen-in 512` (see https://arxiv.org/abs/1804.00247)
- use LecunUniform initialization (pytorch default)
- monitor gradient L2 norm computed in grad_clip. it should be stable around 0.2-0.5 at the first epoch.
- monitor attention weights to early stop if they are not diagnonal

image

acc

This setting for WSJ took 3.6 hours to converge valid acc 94% after 43 epochs. Actually, this is faster than LSTM models (5-6 hours). You can find more results in https://shigekikarita.github.io/espnet-browser/

However, I'm still working on beam-search implementation, the best setting with greedy decoding (beam_size=1, no LM) results
- dev CER 13.9 WER 31.3
- eval CER 10.4 WER 25.5

Fhrozen:
@ShigekiKarita thank you for your support. 
I compared both codes, and found that lr (in pytorch) is the same as alpha (in chainer). So, I've been trying to set multiples alpha init values to train it. The best until now is 1.0. When i tried higher lr, the training fails. Also, I've been training with larger warmup, but it only work for less than 4000 iters.

I will keep working on the training to check any improvement.

ShigekiKarita:
@Fhrozen maybe you misunderstood my comment. I changed my mind to use Adadelta instead of Adam and warmup schedulers. I could not see better results using Adam than the results using Adadelta at this time.

Fhrozen:
Thank for the notification. I misunderstood Adam with adadelta. T.T
I will check this option too later.

ShigekiKarita:
I changed Adaldeta in the best setting with Noam optimizer (Adam + LR scheduer). I believe that it is same to [speech transformer](http://www.mirlab.org/conference_papers/International_Conference/ICASSP%202018/pdfs/0005884.pdf) except for undocumented ones: batchsize=64, maxlen-in=512, ninit=none (LecunNormal). The result is getting much better but it seems to be able to run more epochs (> 50).
- validation acc 95.2%
- dev CER 10.3% WER24.4% (beam=1, w/o LM)
- eval CER 8.2% WER 21.0% (beam=1, w/o LM)

acc

```
./run_transformer_th.sh --batchsize 64 --epochs 50 --accum_grad 1 --opt noam --dropout 0.1 --input_layer conv2d --maxlen-in 512 --lsm_weight 0.1 --lr_init 
10  --beam_size 1 --stage 4 --warmup_steps 25000 --lm_weight 0
```

ShigekiKarita:
The result got better by twice larger minibatch (128 utts on 2GPU).  In addition, I have implemented  beam search.

- valid acc 96.0%
- dev CER 9.0, WER 21.5 (beam=30, w/o LM)
- eval CER 7.2, WER 18.1 (beam=30, w/o LM)

```
# dev
write a CER (or TER) result in ./exp/best_base_bs128_epoch100//decode_test_dev93_beam30_emodel.acc.best_p0.0_len0.0-0.0_ctcw0_rnnlm0_1layer_unit1000_sgd_bs300_word65000/result.txt
|   SPKR      |    # Snt       # Wrd    |   Corr          Sub         Del         Ins          Err       S.Err    |
|   Sum/Avg   |     503        48634    |   93.3          3.7         3.0         2.3          9.0        89.3    |
write a WER result in ./exp/best_base_bs128_epoch100//decode_test_dev93_beam30_emodel.acc.best_p0.0_len0.0-0.0_ctcw0_rnnlm0_1layer_unit1000_sgd_bs300_word65000/result.wrd.txt
|    SPKR      |    # Snt       # Wrd    |    Corr         Sub          Del          Ins          Err        S.Err    |
|    Sum/Avg   |     503         8234    |    81.4        16.1          2.5          2.9         21.5         88.5    |

# eval
write a CER (or TER) result in ./exp/best_base_bs128_epoch100//decode_test_eval92_beam30_emodel.acc.best_p0.0_len0.0-0.0_ctcw0_rnnlm0_1layer_unit1000_sgd_bs300_word65000/result.txt
|   SPKR      |    # Snt       # Wrd    |    Corr         Sub          Del         Ins          Err       S.Err    |
|   Sum/Avg   |     333        33341    |    94.7         3.0          2.3         1.9          7.2        85.6    |
write a WER result in ./exp/best_base_bs128_epoch100//decode_test_eval92_beam30_emodel.acc.best_p0.0_len0.0-0.0_ctcw0_rnnlm0_1layer_unit1000_sgd_bs300_word65000/result.wrd.txt
|    SPKR      |    # Snt       # Wrd    |    Corr          Sub          Del          Ins          Err        S.Err    |
|    Sum/Avg   |     333         5643    |    84.5         13.7          1.8          2.7         18.1         82.9    |
```

sw005320:
Can you do it with a single GPU? You can implement it by storing gradients, truncating graphs without parameter update, repeating these storing and truncating, and finally updating parameters with the accumulated gradients.

ShigekiKarita:
I have already implemented the function by `--accum_grad N` option see https://github.com/ShigekiKarita/espnet/blob/pytorch-transformer/espnet/asr/asr_pytorch.py#L142

ShigekiKarita:
Actually, you do not need to truncate the graph, just backward several times because the backwarded gradients are always added to `param.grad` before `optimizer.step()` unless you call `optimizer.zero_grad()`.

sw005320:
I see. Then, we can recreate your result only with a single GPU, right?

sw005320:
OK. Maybe it's time to think of making PR. I still want to have both chainer and pytorch backend for the transformer, but if @Fhrozen still has issues on the chainer backend, we could think of initiating it with the pytroch backend.

kobenaxie:
will you plan to join CTC/Attenion decodding in beamsearch for transformer ?
[https://www.merl.com/publications/docs/TR2017-103.pdf](url)

Fhrozen:
I apologize for the large silence, I just finish to do some adjustments to the code and this going to be a quite large update.

After transcribing almost all the code from pytorch, The training is showing this curve:
acc
 I am using the following running line:
```
./run_transformer.sh --ngpu 1 --verbose 1 --adim 256 --backend chainer --aheads 4 --elayers 6 --dlayers 6 --beam_size 1 --stage 4 --maxlen_in 512 --batchsize 16 --dropout 0.1 --opt adam --lr_init 10.0 --warmup_steps 25000 --lsm_weight 0.1 --epochs 50
```

However, I used 64 channels (ch) in the conv layers. @ShigekiKarita is currently using dims as ch, so it means around 256 ch. So, when I setup this configuration into the chainer code, I cannot use a batchsize (bs) larger than 16 and consume more than 60% of memory (64% of GTX Titan 12G).
I am testing now with this setup, but the training seems to require 18hrs for training in bs16 and ch256.
I will test will smaller bs since the [paper](http://www.mirlab.org/conference_papers/international_conference/ICASSP%202018/pdfs/0005884.pdf) specified the use of 64ch.

I also trying to reduce the number of reshapes since there is no many of this functions requires and they consume memory
screenshot from 2018-12-21 16-34-37

Also, I tried the training with LeCunn initialization, but it did not work and Chainer init work better.
One last thing is about the Attention layers.
I found that the self attention in the decoders trains fast 
Epoch 1 - Dec Layer 0:
4k8c030u decoder_l0_self_attention ep 1
and this keeps along the training
Epoch 30 - Dec Layer 0:
4k8c030u decoder_l0_self_attention ep 30
and this happens in all the layers (deeper also)

But in the case of the Encoder and the attention to the source it usually takes large time, and in some cases it does not show the same behavior as the self attention of the decoder:
Source Att - Dec Layer 0 - Epoch 1: 
4k8c030u decoder_l0_source_attention ep 1

Source Att - Dec Layer 0 - Epoch 30: 
4k8c030u decoder_l0_source_attention ep 30
Source Att - Dec Layer 5 - Epoch 1: 
4k8c030u decoder_l5_source_attention ep 1
Source Att - Dec Layer 5 - Epoch 30:
4k8c030u decoder_l5_source_attention ep 30
Self Att - Enc Layer 0 - Epoch 1: 
4k8c030u encoder_l0_self_attention ep 1
Self Att - Enc Layer 0 - Epoch 30:
4k8c030u encoder_l0_self_attention ep 30
Self Att - Enc Layer 5 - Epoch 1: 
4k8c030u encoder_l5_self_attention ep 1
Self Att - Enc Layer 5 - Epoch 30: 
4k8c030u encoder_l5_self_attention ep 30

Let me know if there is any comment about it.
I will finish with the convolutional update and the start with the decoding part.
 BTW, the current acc is 92% for the val.

ShigekiKarita:
It looks great. For memory problem, chainer always consume more memory than pytorch because it has memory pool. I wonder there were any option to disable it.

BTW I'm still watching some different default setting accross libraries. I found that LayerNorm.eps is very different, e.g., chainer=1e-5, pytorch=1e-6, and tensorflow=1e-12. I'm currently comparing these parameters but it does not seem to be matter now. And maybe you know, chainer has very different Adam implementation from the original paper.

Fhrozen:
you mean [AdamW](https://docs.chainer.org/en/stable/reference/generated/chainer.optimizers.Adam.html). `AdamW introduces the additional parameters eta and weight_decay_rate,`, they specify.

I already reduced the reshapes, but I cannot see any improvement in the memory.

A model trained in chainer with BS 32 is giving me 94% of acc in Valid, training time 9.39 hrs.
I am testing now with two gpus BS64 (32 32), in order to see if the bs size can improve the result.

ShigekiKarita:
No, it is not about the weight decay. Update rules of the moments (m, v) are different.

ShigekiKarita:
I found my new bug. There was no backprop before the positional encoding (i.e., embedding and subsampling CNN layers) because I had implemented PE as constant buffer at first (now it looks like an activation function). After I removed `no_grad`. The results getting better with the same base setting (layers=6, d_ff=1024, batchsize=64, maxlen_in=512, ngpu=1) from the previous https://github.com/espnet/espnet/pull/465#issuecomment-447653214

diff: https://github.com/ShigekiKarita/espnet/commit/21f2ae8ad7a8d3e9bd78824b1af92e8901f6a98e

previous (without subsampling CNN and embedding update)
- dev CER 10.3% WER24.4% (beam=1, w/o LM)
- eval CER 8.2% WER 21.0% (beam=1, w/o LM)

fixed (with subsampling CNN and embedding update)
- dev CER 9.7%, WER22.9% (beam=1, w/o LM)
- eval CER 7.1%, WER18.6% (beam=1, w/o LM)

I'm surprised with the previous result that it did not even learn CNN and embedding because they are just random matrix projections LOL.

ShigekiKarita:
After I tried  the LARGE model setting (elayers=12, dlayers=6, eunits/dunits=2048) in the paper with bug-fixed PE and batchsize 16 x 4GPUs for 100 epochs (15 hours). I got the best result.

- valid acc 97%
- dev CER 7.6% WER 18.6%
- eval CER 5.9% WER 15.3%

```
#dev
exp/noam/embed1e-12_train_si284_pytorch_transformer_conv2d_e12_subsample0_unit2048_d6_unit2048_aheads4_adim256_mtlalpha0.0_noam_clip5_sampprob0.0_ngpu4_bs16_lr10_warmup25000_dropout0.1_mli1024_mlo150_ninit_none_epochs100_accum1_lsmunigram0.1/decode_test_dev93_beam30_emodel.acc.best_p0.0_len0.0-0.0_ctcw0.3_rnnlm0_1layer_unit1000_sgd_bs300_word65000/result.txt
|                 SPKR                   |                 # Snt                                   # Wrd                 |                 Corr                                     Sub                                    Del                                     Ins                                    Err                                   S.Err                 |
|                 Sum/Avg                |                  503                                    48634                 |                 94.1                                     2.8                                    3.1                                     1.6                                    7.6                                    86.5                 |
exp/noam/embed1e-12_train_si284_pytorch_transformer_conv2d_e12_subsample0_unit2048_d6_unit2048_aheads4_adim256_mtlalpha0.0_noam_clip5_sampprob0.0_ngpu4_bs16_lr10_warmup25000_dropout0.1_mli1024_mlo150_ninit_none_epochs100_accum1_lsmunigram0.1/decode_test_dev93_beam30_emodel.acc.best_p0.0_len0.0-0.0_ctcw0.3_rnnlm0_1layer_unit1000_sgd_bs300_word65000/result.wrd.txt
|                 SPKR                    |                 # Snt                                  # Wrd                  |                 Corr                                     Sub                                     Del                                     Ins                                     Err                                   S.Err                  |
|                 Sum/Avg                 |                  503                                    8234                  |                 83.2                                    14.0                                     2.8                                     1.8                                    18.6                                    85.1                  |
#eval
exp/noam/embed1e-12_train_si284_pytorch_transformer_conv2d_e12_subsample0_unit2048_d6_unit2048_aheads4_adim256_mtlalpha0.0_noam_clip5_sampprob0.0_ngpu4_bs16_lr10_warmup25000_dropout0.1_mli1024_mlo150_ninit_none_epochs100_accum1_lsmunigram0.1/decode_test_eval92_beam30_emodel.acc.best_p0.0_len0.0-0.0_ctcw0.3_rnnlm0_1layer_unit1000_sgd_bs300_word65000/result.txt
|                 SPKR                    |                 # Snt                                  # Wrd                  |                 Corr                                    Sub                                     Del                                    Ins                                     Err                                  S.Err                  |
|                 Sum/Avg                 |                  333                                   33341                  |                 95.9                                    2.3                                     1.8                                    1.8                                     5.9                                   82.3                  |
exp/noam/embed1e-12_train_si284_pytorch_transformer_conv2d_e12_subsample0_unit2048_d6_unit2048_aheads4_adim256_mtlalpha0.0_noam_clip5_sampprob0.0_ngpu4_bs16_lr10_warmup25000_dropout0.1_mli1024_mlo150_ninit_none_epochs100_accum1_lsmunigram0.1/decode_test_eval92_beam30_emodel.acc.best_p0.0_len0.0-0.0_ctcw0.3_rnnlm0_1layer_unit1000_sgd_bs300_word65000/result.wrd.txt
|                 SPKR                    |                 # Snt                                  # Wrd                  |                 Corr                                      Sub                                     Del                                     Ins                                     Err                                   S.Err                  |
|                 Sum/Avg                 |                  333                                    5643                  |                 86.6                                     11.9                                     1.5                                     1.9                                    15.3                                    79.0                  |
```

ShigekiKarita:
I also tried to compare `--maxlen-in 1024` and `--maxlen-in 512` in the LARGE setting. The above result is `1024`. And here is result with 512. Maybe 512 is better.

- valid acc 97%
- dev CER 7.3% WER 17.8%
- eval CER 5.3% WER 14.1%

```
#dev
exp/noam/embed1e-12_train_si284_pytorch_transformer_conv2d_e12_subsample0_unit2048_d6_unit2048_aheads4_adim256_mtlalpha0.0_noam_clip5_sampprob0.0_ngpu4_bs16_lr10_warmup25000_dropout0.1_mli512_mlo150_ninit_none_epochs100_accum1_lsmunigram0.1/decode_test_dev93_beam30_emodel.acc.best_p0.0_len0.0-0.0_ctcw0.3_rnnlm0_1layer_unit1000_sgd_bs300_word65000/result.txt
|                 SPKR                   |                 # Snt                                   # Wrd                 |                 Corr                                    Sub                                     Del                                    Ins                                    Err                                   S.Err                 |
|                 Sum/Avg                |                  503                                    48634                 |                 94.7                                    2.9                                     2.4                                    2.0                                    7.3                                    85.9                 |
exp/noam/embed1e-12_train_si284_pytorch_transformer_conv2d_e12_subsample0_unit2048_d6_unit2048_aheads4_adim256_mtlalpha0.0_noam_clip5_sampprob0.0_ngpu4_bs16_lr10_warmup25000_dropout0.1_mli512_mlo150_ninit_none_epochs100_accum1_lsmunigram0.1/decode_test_dev93_beam30_emodel.acc.best_p0.0_len0.0-0.0_ctcw0.3_rnnlm0_1layer_unit1000_sgd_bs300_word65000/result.wrd.txt
|                 SPKR                    |                 # Snt                                  # Wrd                  |                 Corr                                     Sub                                    Del                                     Ins                                     Err                                   S.Err                  |
|                 Sum/Avg                 |                  503                                    8234                  |                 84.6                                    13.3                                    2.0                                     2.5                                    17.8                                    84.5                  |
#eval
exp/noam/embed1e-12_train_si284_pytorch_transformer_conv2d_e12_subsample0_unit2048_d6_unit2048_aheads4_adim256_mtlalpha0.0_noam_clip5_sampprob0.0_ngpu4_bs16_lr10_warmup25000_dropout0.1_mli512_mlo150_ninit_none_epochs100_accum1_lsmunigram0.1/decode_test_eval92_beam30_emodel.acc.best_p0.0_len0.0-0.0_ctcw0.3_rnnlm0_1layer_unit1000_sgd_bs300_word65000/result.txt
|                 SPKR                   |                 # Snt                                   # Wrd                 |                 Corr                                     Sub                                    Del                                     Ins                                    Err                                   S.Err                 |
|                 Sum/Avg                |                  333                                    33341                 |                 96.3                                     2.2                                    1.5                                     1.6                                    5.3                                    82.0                 |
exp/noam/embed1e-12_train_si284_pytorch_transformer_conv2d_e12_subsample0_unit2048_d6_unit2048_aheads4_adim256_mtlalpha0.0_noam_clip5_sampprob0.0_ngpu4_bs16_lr10_warmup25000_dropout0.1_mli512_mlo150_ninit_none_epochs100_accum1_lsmunigram0.1/decode_test_eval92_beam30_emodel.acc.best_p0.0_len0.0-0.0_ctcw0.3_rnnlm0_1layer_unit1000_sgd_bs300_word65000/result.wrd.txt
|                 SPKR                    |                 # Snt                                  # Wrd                  |                 Corr                                     Sub                                     Del                                     Ins                                     Err                                   S.Err                  |
|                 Sum/Avg                 |                  333                                    5643                  |                 88.1                                    10.8                                     1.1                                     2.1                                    14.1                                    78.1                  |
```

sw005320:
Can you tell me why `--maxlen-in 512` is better? If my understanding is correct, this would reduce the amount of training. Also if the transformer has some benefits of training short utterances, we may consider some curriculum learning (first only train with short utterances and then include long utterances in the training data). Or is this also related to the discussion about the minibatch size?

ShigekiKarita:
I agree with your questions. Maybe my current settings (network struct, lr, warmup_steps) are over tuned for `--maxlen-in 512`. I started `--maxlen-in 1024` since a few days ago. I believe this is not the best result for `--maxlen-in 1024`.

Fhrozen:
I finished with the main code. Training and Decoding are working now. 
My current training command is:
```
 ./run_transformer.sh --ngpu 1 --verbose 1 --adim 256 --backend chainer --aheads 4 --elayers 6 --dlayers 6 --beam_size 1 --stage 4 --maxlen_in 512 --batchsize 32 --dropout 0.1 --opt adam --lr_init 10.0 --warmup_steps 70000 --lsm_weight 0.1 --epochs 50 --weight_init guniform
```

I am using a bs of 32 bc memory limitations. I was checking some parts to reduce the memory consumption, but could not found a solution by the moment. 

Results:
- valid acc 94.57%
- dev CER 13.9% WER 30.3%
- eval CER 11.1% WER 25.1%

```
exp/train_si284_chainer_transformer_winitguniform_e6_d6_attention256_aheads4_dropout0.1_adam10.0_warmup70000_bs32_mli512_mlo150_lsmunigram0.1/decode_test_dev93_beam1_emodel.acc.best_p0.0_len0.0-0.0_ctcw0.3_rnnlm0_1layer_unit1000_sgd_bs300_word65000/result.wrd.txt:|            SPKR              |            # Snt                      # Wrd            |            Corr                          Sub                          Del                         Ins                          Err                        S.Err            |
exp/train_si284_chainer_transformer_winitguniform_e6_d6_attention256_aheads4_dropout0.1_adam10.0_warmup70000_bs32_mli512_mlo150_lsmunigram0.1/decode_test_dev93_beam1_emodel.acc.best_p0.0_len0.0-0.0_ctcw0.3_rnnlm0_1layer_unit1000_sgd_bs300_word65000/result.wrd.txt:|            Sum/Avg           |             503                        8234            |            74.9                         22.1                          3.0                         5.2                         30.3                         94.6            |
exp/train_si284_chainer_transformer_winitguniform_e6_d6_attention256_aheads4_dropout0.1_adam10.0_warmup70000_bs32_mli512_mlo150_lsmunigram0.1/decode_test_eval92_beam1_emodel.acc.best_p0.0_len0.0-0.0_ctcw0.3_rnnlm0_1layer_unit1000_sgd_bs300_word65000/result.wrd.txt:|            SPKR              |            # Snt                       # Wrd            |            Corr                         Sub                          Del                          Ins                          Err                        S.Err            |
exp/train_si284_chainer_transformer_winitguniform_e6_d6_attention256_aheads4_dropout0.1_adam10.0_warmup70000_bs32_mli512_mlo150_lsmunigram0.1/decode_test_eval92_beam1_emodel.acc.best_p0.0_len0.0-0.0_ctcw0.3_rnnlm0_1layer_unit1000_sgd_bs300_word65000/result.wrd.txt:|            Sum/Avg           |             333                         5643            |            79.7                        18.4                          1.9                          4.8                         25.1                         91.0            |
```

I will try to test with 2 gpus, but I also had the memory limitations when I was trying to execute it in a server.

Also, I implemented and test differents inizializations (LeCun, Gorot, He Uniform) and Gorot give the best result by the moment.

Tested with different warmup and 70000 is the best by the moment.

Fhrozen:
I just finished some additional tests.
I finished the tests with multigpu (at least 2 gpus).
Also, testing different number of layers and longer training (100 epochs). The training is pretty fast (10hrs per 100 epochs approx.). 
The result could be improved with larger BS (128?) or probably adding batch norm in the input layers.

- e8 d4 epoch 100 lsm 0.2 bs 64: val 95.66%   elapsed time 10.35
```
train_si284_chainer_transformer_winitguniform_e8_d4_attention256_aheads4_dropout0.1_adam10.0_warmup70000_bs32_mli512_mlo150_lsmunigram0.2/decode_test_dev93_beam1_emodel.acc.best_p0.0_len0.0-0.0_ctcw0.3_rnnlm0_1layer_unit1000_sgd_bs300_word65000/result.wrd.txt:|            SPKR              |            # Snt                      # Wrd            |            Corr                          Sub                          Del                         Ins                          Err                        S.Err            |
train_si284_chainer_transformer_winitguniform_e8_d4_attention256_aheads4_dropout0.1_adam10.0_warmup70000_bs32_mli512_mlo150_lsmunigram0.2/decode_test_dev93_beam1_emodel.acc.best_p0.0_len0.0-0.0_ctcw0.3_rnnlm0_1layer_unit1000_sgd_bs300_word65000/result.wrd.txt:|            Sum/Avg           |             503                        8234            |            80.4                         16.9                          2.6                         2.8                         22.3                         89.3            |
train_si284_chainer_transformer_winitguniform_e8_d4_attention256_aheads4_dropout0.1_adam10.0_warmup70000_bs32_mli512_mlo150_lsmunigram0.2/decode_test_eval92_beam1_emodel.acc.best_p0.0_len0.0-0.0_ctcw0.3_rnnlm0_1layer_unit1000_sgd_bs300_word65000/result.wrd.txt:|            SPKR              |            # Snt                       # Wrd            |            Corr                         Sub                          Del                          Ins                          Err                        S.Err            |
train_si284_chainer_transformer_winitguniform_e8_d4_attention256_aheads4_dropout0.1_adam10.0_warmup70000_bs32_mli512_mlo150_lsmunigram0.2/decode_test_eval92_beam1_emodel.acc.best_p0.0_len0.0-0.0_ctcw0.3_rnnlm0_1layer_unit1000_sgd_bs300_word65000/result.wrd.txt:|            Sum/Avg           |             333                         5643            |            84.4                        13.9                          1.8                          2.8                         18.5                         86.2            |
```
- e6 d6 epoch 100 lsm 0.1 bs 64: val 95.35%   elapsed time 10.80

```
train_si284_chainer_transformer_winitguniform_e6_d6_attention256_aheads4_dropout0.1_adam10.0_warmup70000_bs32_mli512_mlo150_lsmunigram0.1/decode_test_dev93_beam1_emodel.acc.best_p0.0_len0.0-0.0_ctcw0.3_rnnlm0_1layer_unit1000_sgd_bs300_word65000/result.wrd.txt:|            SPKR              |            # Snt                      # Wrd            |            Corr                          Sub                          Del                         Ins                          Err                        S.Err            |
train_si284_chainer_transformer_winitguniform_e6_d6_attention256_aheads4_dropout0.1_adam10.0_warmup70000_bs32_mli512_mlo150_lsmunigram0.1/decode_test_dev93_beam1_emodel.acc.best_p0.0_len0.0-0.0_ctcw0.3_rnnlm0_1layer_unit1000_sgd_bs300_word65000/result.wrd.txt:|            Sum/Avg           |             503                        8234            |            77.8                         19.0                          3.2                         3.3                         25.5                         89.7            |
train_si284_chainer_transformer_winitguniform_e6_d6_attention256_aheads4_dropout0.1_adam10.0_warmup70000_bs32_mli512_mlo150_lsmunigram0.1/decode_test_eval92_beam1_emodel.acc.best_p0.0_len0.0-0.0_ctcw0.3_rnnlm0_1layer_unit1000_sgd_bs300_word65000/result.wrd.txt:|            SPKR              |            # Snt                       # Wrd            |            Corr                         Sub                          Del                          Ins                          Err                        S.Err            |
train_si284_chainer_transformer_winitguniform_e6_d6_attention256_aheads4_dropout0.1_adam10.0_warmup70000_bs32_mli512_mlo150_lsmunigram0.1/decode_test_eval92_beam1_emodel.acc.best_p0.0_len0.0-0.0_ctcw0.3_rnnlm0_1layer_unit1000_sgd_bs300_word65000/result.wrd.txt:|            Sum/Avg           |             333                         5643            |            81.6                        16.3                          2.2                          2.8                         21.2                         87.1            |
```

I will finish with the beam search and submit the results.
Once approved, I will submit a PR for the v0.3.0

sw005320:
Very cool!
@Fhrozen and @ShigekiKarita, can you summarize the difference between your implementations?
I want to make this transformer implementation a similar way as much as possible.

sw005320:
Thanks. These do not look so different, and we could unify the implementations.

Fhrozen:
Follows #655. Closed

PR #467, opened 2018-11-16
AMI recipe improvement

Xiaofei-Wang:
I updated the AMI recipe using the newest tools. Several changes have been done as follows:
1. Add a language model which is trained only by AMI dataset，batchsize=64,  lm_weight=0.5;
2. Make the mdm8 recipe work by adding the beamforming related files;
3. Update the RESULTS file.

So far, we have the results as follows:
- ihm task
  - WER: Dev 35.7; Eval 38.5
  - CER: Dev 22.6; Eval 24.0


Xiaofei-Wang:
Sure.

Xiaofei

Issue #444, opened 2018-10-31
Development plan for v.0.3.0


sw005320:
If you have any suggestions for the plan, please let us know.

- [x] modularization #440, #484 
- [x] add travis CI test for `run.sh` and `utils/*` #484
- [x] add test scripts for `utils/*` #498 
- [x] add GRU architectures #496 
- [x] tensorboard support #506 
- [x] batch decoding #318 
- [x] enhance (speech) translation recipes #468 
- new ASR recipe #491 
  - [x] improve the ASR baseline by speed perturbation, small mtlalpha, dropout #480 #491 #521 #546 
  - [x] refine noisy reverberant ASR recipes, adding frontend processing (not end-to-end now) 
  - [x] speech separation recipes #457 
  - [x] including Nara WPE and beamforming (ami, chime4, chime5, reverb) #435  #467 #469 
- [x] use of hdf5 #412 #493 
- [x] use chainer v5.X (now we fix a version at 4.3.1 with @jnishi's patch) #460 
- [x] support pytorch 1.0 #553 
- [x] remove deprecated warning from PyTorch #553 
- [ ] transformer #465 
- [ ] recurrent dropout in decoder
- [ ] use cudnn CTC for chainer and pytorch #433 
- [ ] change the beam search interface #489 
- [x] bug fix etc. #481 #490 #544 
- [ ] config files in addition to the options

sw005320:
I just made a branch for this development plan https://github.com/espnet/espnet/tree/v.0.3.0, as it affects to many people if we work on the master.
Please work on the above items in this branch.

sw005320:
Hi all, HDF5 extension seems to be working and I'm thinking of merging this v.0.3.0 to the master in a month.

I just want to list the ongoing items that I want to include to v.0.3.0
- Transformer: @ShigekiKarita (and @Fhrozen) is working on it, and we could get his PR in a week, and possibly merge it in two weeks
- Beam search interface: again @ShigekiKarita is working on it. This may be only used for the transformer for v.0.3.0, but we can extend to use this for RNN based encoder-decoder in the next update.
- Speech translation: @hirofumi0810 is working on it. Now it produces comparable/superior performance to the other people (Google)'s report. We already reviewed most of parts and we could merge it in a week or so.
- Speech separation and recognition @simpleoier is working on it. The main part is finished, but it may take some time to well merge the v.0.3.0, as it extends several main functions/classes to handle multiple output streams.

[...]

PR #468, opened 2018-11-20
Fisher callhome spanish speech translation

hirofumi0810:
new recipe for speech translation.


sw005320:
@ShigekiKarita, can you test it in your environment? @hirofumi0810 reports that the bleu score is worse than the other reports, and we might have some bugs in ESPnet.

ShigekiKarita:
OK I will test this

espnet/nets/e2e_asr_th.py
```diff
@@ -379,6 +380,9 @@ def set_forget_bias_to_one(bias):
         # https://discuss.pytorch.org/t/set-forget-gate-bias-of-lstm/1745
         for l in six.moves.range(len(self.dec.decoder)):
             set_forget_bias_to_one(self.dec.decoder[l].bias_ih)
+        for n, p in self.named_parameters():
```
ShigekiKarita: We initialize this way intendedly as chainer's default initialization (NStepLSTM does not init forget bias to one). Have you observed any improvement by using your new initialization here?

ShigekiKarita: That is the reason why your `pytest -k test_lecun_init_torch` failed in Travis-CI. Actually, I found that setting forget bias to 1 does not improve the ASR result at this time https://github.com/espnet/espnet/issues/9 (not written here but I remember)

hirofumi0810: @ShigekiKarita I just did the same manner in the decoder network.
I didn't see any performance differences, so I'll remove this part.

ShigekiKarita:
By the way, what is your baseline paper? I think this one is the most simple and detailed https://www.isca-speech.org/archive/Interspeech_2017/pdfs/0503.PDF

hirofumi0810:
@ShigekiKarita this is another one implemented by chainer.
https://www.isca-speech.org/archive/Interspeech_2018/pdfs/1326.pdf
I think we can reach the same range as this paper at least.

ShigekiKarita:
OK. I will check the consistency between the paper and this implementation.

egs/fisher_callhome_spanish_st/st1/local/splits/.dropbox.attr
```diff
@@ -0,0 +1 @@
+{}
```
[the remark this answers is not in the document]
hirofumi0810: Fixed.

ShigekiKarita:
From my experience of making pytorch backend comparable to chainer one, I suspect that weight initialization really matters but its detail is not appeared in the paper. However, I found some difference in their code from chainer default (our impl)

- In CNN initialization, he used chainer.initializers.HeNormal instead of default LeCunNormal https://github.com/0xSameer/speech2text/blob/af8c365782a4e61c8b008ee96a18d1dab7e2e73a/enc_dec.py#L148
- In RNN initialization, he also used HeNormal https://github.com/0xSameer/speech2text/blob/seq2seq/enc_dec.py#L70
- Linear layer seems to be default
- Embedding layer seems to be default

hirofumi0810:
@sw005320 @ShigekiKarita I report some results here.
In my findings, the default setting led to overfitting.
So I'm trying some regularization techniques.

| Model | fisher-dev  | fisher-dev2 | fisher-test | callhm-devtest | callhm-evaltest | 
|---|---|---|---|---|---|
| Weiss et al. | 46.5 | 47.3 | 47.3 | 16.4 | 16.6 | 
| Bansal et al. | 29.4  | N/A | N/A | N/A | N/A |
| default | 20.23 | 21.35 | 20.76 | 11.65 | 10.89 | 
| samp_prob=0.2, lsm_weight=0.1, dropout=0.3 | 23.39 | 24.66 | 23.64 | 14.12 | 12.75 | 

Callhome results are in the same range in Google's paper, however I don't know why there are so differences in Fisher sets.
I think regularization is very important for speech translation contrary to ASR in espnet.

ShigekiKarita:
I found that implementations of Adam in chainer and pytorch are different. For comparison, I rewrite those formula simpler as follows.

https://github.com/chainer/chainer/blob/v4/chainer/optimizers/adam.py
``` python
# chainer
m += (1 - beta1) * (grad - m)
v += (1 - beta2) * (grad * grad - v)
param -=  lr * m / (sqrt(v) + eps) + wd * param
```

https://github.com/pytorch/pytorch/blob/v0.4.1/torch/optim/adam.py#L89
```python
# pytorch
grad += weight_decay * param
m = m * beta1 + (1 - beta1) * grad
v = v * beta2 + (1 - beta2) * grad * grad
step += 1
p = lr * sqrt(1 - beta2 ** step) / (1 - beta1 ** step) * m  / (sqrt(v) + eqs)
```
I still not get the equality of `m, v` in pytorch and chainer but the L2 penalty (weight decay) should be different among these. I think chainer refers this paper to implement the weight decay https://openreview.net/forum?id=rk6qdGgCZ

I think this difference is worth rethinking because one of the authors in [Weiss17] and [Bansal18] said that weight decay is important when I met them in Interspeech.

hirofumi0810:
@ShigekiKarita 
I'm currently using Adadelta. I'll try Adam, weight decay, and HeNormal initialization, too.

ShigekiKarita:
On attention, I think you are using `atype=add` now 

- AttAdd https://github.com/hirofumi0810/espnet/blob/a0ee39e551f5f69c4a7a85b4473e96e0853bf640/espnet/nets/e2e_asr_th.py#L723

However [Bansal18] might be different, they used *global general attention and input-feeding* defined in http://aclweb.org/anthology/D15-1166

The implementations are
- attention https://github.com/0xSameer/speech2text/blob/af8c365782a4e61c8b008ee96a18d1dab7e2e73a/enc_dec.py#L319
- decoder https://github.com/0xSameer/speech2text/blob/af8c365782a4e61c8b008ee96a18d1dab7e2e73a/enc_dec.py#L389

To make equivalent implementation, I recommend you to
- use `AttDot` instead of `AttAdd`
- use input-feeding in

https://github.com/hirofumi0810/espnet/blob/a0ee39e551f5f69c4a7a85b4473e96e0853bf640/espnet/nets/e2e_asr_th.py#L2081-L2095

like this way
```python
# in __init__
self.context = nn.Linear(...)

# in forward
# pre-computation of embedding
eys = self.dropout_emb(self.embed(ys_in_pad)) # utt x olen x zdim
# loop for an output sequence
for i in six.moves.range(olength):
    ey = torch.cat((eys[:, i, :], att_c), dim=1)  # utt x (zdim + hdim)
    z_list[0], c_list[0] = self.decoder[0](ey, (z_list[0], c_list[0]))
    # input-feeding
    att_c, att_w = self.att(hs_pad, hlens, z_list[0], att_w)
    zc = torch.cat((z_list[0], att_c), dim=1)
    z_all.append(torch.tanh(self.context(zc)))
```

espnet/nets/e2e_asr_th.py
```diff
@@ -2075,14 +2084,14 @@ def forward(self, hs_pad, hlens, ys_pad):
                 logging.info(' scheduled sampling ')
                 z_out = self.output(z_all[-1])
                 z_out = np.argmax(z_out.detach(), axis=1)
```
ShigekiKarita: `torch.argmax(z_out.detach(), dim=1)` can be faster because it can use GPU without memory re-transfer by `z_out.cuda()`

hirofumi0810:
@ShigekiKarita 
Yeah I know the difference but I saw decreasing in BLEU when using dot-product attention.
Also, though I tried input-feeding in iwslt18 recipe, it didn't give improvements (it may change in the current espnet).

espnet/nets/e2e_asr_th.py
hirofumi0810: this is related to https://github.com/espnet/espnet/pull/358#issuecomment-420815239

ShigekiKarita:
OK

sw005320:
This is related to #444

espnet/nets/e2e_asr_th.py
ShigekiKarita: What the performance means in that issue? Is torch.argmax slower than numpy.argmax?

sw005320:
> I'm currently using Adadelta. I'll try Adam, weight decay, and HeNormal initialization, too.

Thanks.
Please report the effect one by one.
I want to find the effectiveness of each method.

egs/fisher_callhome_spanish_st/st1/local/callhome_data_prep.sh
```diff
@@ -0,0 +1,166 @@
...
+
+# Basic spot checks to see if we got the data that we needed
+if [ ! -d links/LDC96S35 -o ! -d links/LDC96T17 ];
+then
+        echo "The speech and the data directories need to be named LDC96S35 and LDC96T17 respecti
+vely"
+        exit 1;
+fi
+
+if [ ! -d links/LDC96S35/CALLHOME/SPANISH/SPEECH/DEVTEST -o ! -d links/LDC96S35/CALLHOME/SPANISH/SPEECH/EVLTEST -o ! -d links/LDC96S35/CALLHOME/SPANISH/SPEECH/TRAIN ];
```
ShigekiKarita: When I downloaded LDC96S35 from LDC yesterday, the path looks like `LDC96S35/callhome/spanish/speech/devtest`

how about this?
```bash
if [ -d links/LDC96S35/CALLHOME/SPANISH/SPEECH/DEVTEST -o -d links/LDC96S35/CALLHOME/SPANISH/SPEECH/EVLTEST -o -d links/LDC96S35/CALLHOME/SPANISH/SPEECH/TRAIN ]
then
    speech_train=$dir/links/LDC96S35/CALLHOME/SPANISH/SPEECH/TRAIN
    speech_dev=$dir/links/LDC96S35/CALLHOME/SPANISH/SPEECH/DEVTEST
    speech_test=$dir/links/LDC96S35/CALLHOME/SPANISH/SPEECH/EVLTEST
    fcount_train=`find ${speech_train} -iname '*.SPH' | wc -l`
    fcount_dev=`find ${speech_dev} -iname '*.SPH' | wc -l`
    fcount_test=`find ${speech_test} -iname '*.SPH' | wc -l`
elif [ -d links/LDC96S35/callhome/spanish/speech/devtest -o -d links/LDC96S35/callhome/spanish/speech/evltest -o -d links/LDC96S35/callhome/spanish/speech/train ]
then
    speech_train=$dir/links/LDC96S35/callhome/spanish/speech/train
    speech_dev=$dir/links/LDC96S35/callhome/spanish/speech/devtest
    speech_test=$dir/links/LDC96S35/callhome/spanish/speech/evltest
    fcount_train=`find ${speech_train} -iname '*.sph' | wc -l`
    fcount_dev=`find ${speech_dev} -iname '*.sph' | wc -l`
    fcount_test=`find ${speech_test} -iname '*.sph' | wc -l`
else
        echo "Dev, Eval or Train directories missing or not properly organised within the speech data dir"
        exit 1;
fi
```

egs/fisher_callhome_spanish_st/st1/local/callhome_data_prep.sh
```diff
@@ -0,0 +1,166 @@
+#!/bin/bash
```
ShigekiKarita: Indentation looks too nasty. (There are 2, 3, 4, 8 spaces. I suggest 4 spaces)

egs/fisher_callhome_spanish_st/st1/local/callhome_data_prep.sh
```diff
@@ -0,0 +1,166 @@
...
+then
+        echo "Incorrect number of files in the data directories"
+        echo "The paritions should contain 80/20/20 files"
+        exit 1;
+fi
+
+if [ $stage -le 0 ]; then
+  #Gather all the speech files together to create a file list
+  (
+      find $speech_train -iname '*.sph';
```
ShigekiKarita: maybe `find $speech_train -iname '*.(SPH|sph)'; `? I'm not sure.

egs/fisher_callhome_spanish_st/st1/local/callhome_data_prep.sh
```diff
@@ -0,0 +1,166 @@
...
+  $utils/utt2spk_to_spk2utt.pl <$dir/callhome_train_all/callhome_utt2spk > $dir/callhome_train_all/callhome_spk2utt
+fi
+
+if [ $stage -le 3 ]; then
+  for f in `cat $tmpdir/callhome_train_sph.flist`; do
+    # convert to absolute path
+    make_absolute.sh $f
+  done > $tmpdir/callhome_train_sph_abs.flist
+
+  cat $tmpdir/callhome_train_sph_abs.flist | perl -ane 'm:/([^/]+)\.SPH$: || die "bad line $_; ";  print lc($1)," $_"; ' > $tmpdir/callhome_sph.scp
```
ShigekiKarita: `perl -ane 'm:/([^/]+)\.(SPH|sph)$: || ...`

egs/fisher_callhome_spanish_st/st1/local/callhome_make_spk2gender.py
```diff
@@ -0,0 +1,29 @@
+#!/usr/bin/env python
```
ShigekiKarita: If you want to use python2 here, rewrite this with `#!/usr/bin/env python2`. When user's python is python3, python3 should not be called here.

egs/fisher_callhome_spanish_st/st1/local/fsp_make_spk2gender.py
```diff
@@ -0,0 +1,29 @@
+#!/usr/bin/env python
```
ShigekiKarita: ditto

ShigekiKarita:
Except for the minor fix above, this recipe works fine until the stage 4:  Network Training. GJ

ShigekiKarita:
PS. You need to add symbolic links to `utils` and `steps` from `../../../tools/kaldi/egs/wsj/s5/`

egs/fisher_callhome_spanish_st/st1/local/callhome_make_spk2gender.py
hirofumi0810: @ShigekiKarita Sorry, these files are copies from kaldi repo. I didn't modify them.
@sw005320, is it OK to change these files?

hirofumi0810:
> > I'm currently using Adadelta. I'll try Adam, weight decay, and HeNormal initialization, too.
> 
> Thanks.
> Please report the effect one by one.
> I want to find the effectiveness of each method.

@sw005320 @ShigekiKarita
Interest finding: When using Adam, models never converge.

ShigekiKarita:
My exp reproduced your results (BLEU: fisher dev 21.05, test 22.43) with `exp/train.en_pytorch_vggblstm_e5_subsample1_2_2_1_1_unit1024_proj1024_d2_unit1024_add1024_adadelta_sampprob0_lsm0_drop0_bs15_mli800_mlo150_wd0/decode_fisher_dev2.en_beam5_emodel.acc.best_p0.2_len0-0.8_rnnlm0` (maybe default)

egs/fisher_callhome_spanish_st/st1/local/score_bleu.sh
```diff
@@ -0,0 +1,109 @@
...
+fi
+
+# detokenize
+detokenizer.perl -l en < ${dir}/ref.wrd.trn > ${dir}/ref.wrd.trn.detok
+detokenizer.perl -l en < ${dir}/hyp.wrd.trn > ${dir}/hyp.wrd.trn.detok
+detokenizer.perl -l en < ${dir}/src.wrd.trn > ${dir}/src.wrd.trn.detok
+
+### case-insensitive
+multi-bleu-detok.perl -lc ${dir}/ref.wrd.trn.detok < ${dir}/hyp.wrd.trn.detok > ${dir}/result.wrd.txt
+echo "write a case-insensitive word-level BLUE result in ${dir}/result.wrd.txt"
```
ShigekiKarita: typo BLUE -> BLEU

sw005320:
The rest of TODO
- [ ] Remove call home training data
- [ ] Update RESULTS

sw005320:
The recipe seems to become a good shape. Can you fix the conflict, and also add the latest results? If they're fixed, I'll merge it.

egs/iwslt18st/st1/path.sh
```diff
@@ -3,11 +3,12 @@ KALDI_ROOT=$MAIN_ROOT/tools/kaldi
 
 [ -f $KALDI_ROOT/tools/env.sh ] && . $KALDI_ROOT/tools/env.sh
```
sw005320: can you check `path.sh` in the other recipe? We should follow the style similar to others as much as possible.

hirofumi0810: @sw005320 I'm sorry. I forgot to follow this direction. I just fixed it.

sw005320:
@hirofumi0810, could you accelerate it?
This is a great recipe, and I really want to include this to v.0.3.0.
You may not need to make it perfect. 
If you follow my suggestions above and fix them for the v.0.3.0 style, which is fine.

sw005320:
Can you change `cmd.sh` to the latest one (https://github.com/espnet/espnet/blob/v.0.3.0/egs/an4/asr1/cmd.sh)?
Then, I'll merge it.

sw005320:
Thanks for your long-term efforts!
I'll merge it.
I could not test all of your recipe, and if we have some problems, I'd ask your supports...

PR #469, opened 2018-11-20
AMI recipe improvement 2

Xiaofei-Wang:
I tried smaller vocab size, results kept same. So I did several modifications:
1. update run.sh;
2. update RESULTS.

Also, I found files for the mdm8 experiments were missing. I added as well
1. ./local/beamformit.sh
2. ./conf/ami_beamformit.cfg


sw005320:
Thanks!

PR #481, opened 2018-11-27
debug for multi-head attention

Magic-Bubble:


kan-bayashi:
Hi @Magic-Bubble.
Could you explain the contents and reason of this PR?

Magic-Bubble:
Sorry for that, I open this PR because there is a bug when I run asr with librispeech in decode stage using multi-head attention. 

In this case, att_w is a list, not a tensor, so when using function recognize_beam_batch in decode stage, torch.index_select need to be used for every head attention in att_w.

Please check PR for details, the adaption just add one line code.

kan-bayashi:
I see. Thank you for your contoribution! @Magic-Bubble

PR #491, opened 2018-12-03
speed perturbation and small changes

JaejinCho:


sw005320:
Can you also put the learning curve? If it converges soon, we may reduce the number of epochs.

JaejinCho:
**Previous model running curve (before adding speech perturbation, dropout)**
image
image

**After adding speech perturbation and dropout**
image
image

The loss curve of the speed perturbation added model looks starnge. I will check it and report back.

sw005320:
OK. It seems that you only require 5 epochs to converge. So you can change `epochs` from 15 to 8 in https://github.com/espnet/espnet/blob/8abac236544bc53975f9e44fef84922688188270/egs/fisher_callhome_spanish/asr1/run.sh#L56

sw005320:
@JaejinCho, Can you revert the change of "
Merge branch 'v.0.3.0' into new_recipe_fisher_callhome_spanish
887a7c0"?
We decided not to do this style. Then, if you fix the conflict etc., I'll merge it.

egs/librispeech/asr1/run.sh
```diff
@@ -129,8 +129,13 @@ if [ "${stage}" -le 1 ]; then
             "data/${x}" "exp/make_fbank/${x}" "${fbankdir}"
     done
 
+<<<<<<< HEAD
```
sw005320: Please fix the conflict. I guess you can just remove the HEAD.

egs/hkust/asr1/local/hkust_segment.py
```diff
@@ -14,4 +14,8 @@
             continue
         for j in seg_txt(blks[i]):
             out_line += " " + j
+<<<<<<< HEAD
```
sw005320: Please fix the conflict. I guess you can just remove the HEAD.

src/utils/reduce_data_dir.sh
```diff
@@ -0,0 +1,52 @@
+#!/bin/bash
```
sw005320: `src/utils/reduce_data_dir.sh` is moved to `utils/reduce_data_dir.sh`.
(see https://github.com/espnet/espnet/blob/v.0.3.0/utils/reduce_data_dir.sh)

Please confirm it and also modify your script accordingly (but it seems that you don't use this script, and you can just remove it).

egs/fisher_callhome_spanish/asr1/run.sh
```diff
@@ -183,12 +196,12 @@ if [ "${stage}" -le 2 ]; then
     mkdir -p data/lang_1char/
 
     echo "make a non-linguistic symbol list"
-    cut -f 2- -d' ' "data/${train_set}/text" | tr " " "\n" | sort | uniq | grep "\[" | awk -F"]" '{print $1"]"}' | uniq > "${nlsyms}"
+    cut -f 2- -d' ' "data/${train_set_ori}/text" | tr " " "\n" | sort | uniq | grep "\[" | awk -F"]" '{print $1"]"}' | uniq > "${nlsyms}"
```
sw005320: We decided not to include `"` when it is not necessary. Could you check the diff from https://github.com/espnet/espnet/blob/v.0.3.0/egs/fisher_callhome_spanish/asr1/run.sh and follow the same style?

sw005320:
I close this PR because the new PR #546 is already merged.

Issue #412, opened 2018-09-12
use hdf5 at feature dumping and reading


sw005320:
@ShigekiKarita and @TsubasaOchiai pointed out the effectiveness of using the hdf5 format. This is one of our old action item, and I want to raise this issue. I found the following source https://github.com/simpdanny/caffe-kaldi/blob/master/kaldi-ark-to-hdf5.py, and it seems that it is very straightforward to call such function in `dump.sh`. Then, in the main ESPnet, we can just read hdf5 files instead of `kaldi_io_py.read_mat`.

Fhrozen:
I did something like that for another code (CHiME3)
I used python to convert the data from the mfcc and store it into h5py.
[mfcc_to_h5.txt](https://github.com/espnet/espnet/files/2380259/mfcc_to_h5.txt)

sw005320:
@ShigekiKarita, do you have comments about it?

ShigekiKarita:
It does not need heavy rewrite in `run.sh` and intermediate files (ark) but `ark,t` should be avoided by [kaldi-io-for-python](https://github.com/vesis84/kaldi-io-for-python) for faster processing.

```
_EXEC_CMMD = '{}/src/featbin/compute-mfcc-feats --config=./conf/mfcc.conf --verbose=2 scp,p:{} ark:- | '
_EXEC_CMMD += '{}/src/featbin/copy-feats ark:- ark,t:{} '
```

ShigekiKarita:
Our preliminary experiment did following steps because hdf5 is little bit complex format to do parallel distributed creation unlike kaldi's scp/ark.
1. create ark files as same as the current recipe in the distributed way
2. convert them into single HDF5 files by `kaldi_io_py.read_scp` and `h5py.create_dataset`
3. replace kaldi_io_py in  `load_inputs_and_targets` with h5py

The problem is, as seen in the first step, there are huge intermediate ark files. I found the [HDF5 file link](http://docs.h5py.org/en/latest/high/group.html#hard-links) like scp is really easy. Maybe the best steps to create HDF5 is as follows
1. create hdf5 files (e.g., data/train/JOB.h5) as same as the current recipe in the distributed way. Here we need to rewrite `kaldi_io_py.write_mat` with `h5py.create_dataset` in compute-fbank-feats.py
2. merge hdf5 files into a single hdf5 file by linking

- src/utils/create-fbank-feats.py modification 
[...]

PR #493, opened 2018-12-06
Preliminary arrangements for multi channel system to use HDF5 for feature file and on-the-fly training from wave data

kamo-naoyuki:
#412 

## Summary of this PR

- Supports both `HDF5` and `kaldi-matrix-ark` format for input feature file: `espnet.utils.io_utils:LoadInputAndTargets`
- On-the-fly preprocessing for mini-batch creation: `espnet.utils.transform:Transformation`
- Prepares fundamental `utils` handling `HDF` files as with `kaldi-matrix-ark`.
- Essential mechanism using raw `WAV` in stead of `FBANK` in training flow: See https://github.com/espnet/espnet/pull/493#issuecomment-452373673
- Change to use [kaldiio](https://github.com/nttcslab-sp/kaldiio) instead of [kaldi-io-for-python](https://github.com/vesis84/kaldi-io-for-python/blob/master/kaldi_io.py): See https://github.com/espnet/espnet/pull/493#issuecomment-450445263. 
- Create `egs/chime4/asr1_multich` for testing this new system
    - The status is still WIP and I'll complete in the next PR.

## Why HDF5?

- We need more complex structure e.g. Multi-channel STFT feats is required for the multi-channel end-end system.
- (Maybe) Less IO overhead than `kaldi-ark/scp` because `HDF5` was much developed for performance.
- Use more general format for the people who are not used to `Kaldi`. `ark` format is not well known and we need unofficial tool such  [kaldi-io-for-python](https://github.com/vesis84/kaldi-io-for-python/blob/master/kaldi_io.py). 

## Status
- [x] Implement a Python module for the new format.
- [x] Create test codes for new python modules.
- [x] Create or modify new `utils` for the new format.
- Check egs to reproduce the previous result and the speed.
  - [x] egs/an4/asr1
  - [x] egs/ljspeech/tts1
- [x] Creates test codes for the new `utils`.
- [x] Work with `MultiProcessingIterator`
- [x] Work with  `Multi GPU`
- [x] Create scripts to dump `FBANK`  in files.
- [x] Create multi channel example


kamo-naoyuki:
@sw005320 @kan-bayashi @ShigekiKarita  
This PR includes much changing from the v0.2.0 and I need your help. 

## New JSON format
Finally, I decided the new format as following, 

```json
{
    "utts": {
        "LJ050-0029": {
            "input": [
                {
                    "feat": "somefile.32.h5:LJ050-0029",
                    "filetype": "hdf5",
                    "name": "input1",
                    "shape": [
                        408, 
                        80   
                    ]
                }
            ], 
```

The difference from the previous one is the existing of  "filetype": If the key doesn't exist, the modules performing in the legacy style.

## Unified `load_inputs_and_targets` 
Considering the maintenancebility, I merged `load_inputs_and_targets` in `espnet.asr.asr_utils` and `espnet.tts.tts_utils` into `espnet.utils.io:LoadInputsAndTargets`.

## `scp` with `HDF5`
I think the `scp` system is  quite useful even if we select `HDF5`, because some Kaldi's utils doesn't assume that the file format is `mat-ark`: e.g. `utils/split_scp.pl` can apply  all `scp` regardless of its file format.

Thus, I carefully defined the following  `scp` format for `HDF5` and created new `utils` supporting `HDF5`.

```
uttid1 somefile.32.h5:uttid1
uttid2somefile.32.h5:uttid2
...
```

Note that the file also can be converted into the JSON file using `scp2json.py`

## `mergejson.py`
I modified `mergejson.py` to support more general inputs and supporting multiple inputs and outputs.

Usage:

```bash
# 1 input and 1 output
mergejson.py --input-jsons input/*.json --output-jsons output/*.json --json other/*.json
# 2 input and 1 output (--input-jsons and --output-json can be repeated)
mergejson.py --input-jsons input/*.json --input-jsons input2/*.json  --output-jsons output/*.json --json other/*.json
```

## The other new/modified utils
To use `HDF5` in the several recipes, I needed to create  many `utils` scripts.

I'd like you to keep in mind that I made the changing to be the smallest and the CLIs are kept excluding `mergejson.py`: All tools still support `mat-ark` and `scp` style and I expect all `run.sh` can work as it is.

New: 
- `compute-cmvn-stats.py`: To calculate cmvn stas with `HDF5`
- `copy-feats.py`: To convert `mat-ark` to `HDF5`
- `feat-to-shape.py``: To get the shape with  `HDF5`

Modified:

- `compute-fbank-feats.py`: To dump in   `HDF5`
- `compute-stft-feats.py`: To dump in   `HDF5`
- `make_fbank.sh`: To dump in   `HDF5`
- `make_stft.sh`: To dump in   `HDF5`
- `dump.sh``: To use  ``copy-feats.py`
- `mergejson.py`: To include `filetype` key in the output JSON
- `data2json.sh`: To use  `feat-to-shape.py`

## [Experimental feature] Pre-processing using 
I also implemented very experimental feature: Pre-processing for data at the run time, it means that you can any functions, such las  affine-transform, can apply to the mini-batch in the training and the decoding . I implemented it considering of the experiments with mutli-channel end-to-end system.

The motivations are:

- `Stft` feats are too big to be saved.  The IO cost cannot be ignored. I want to convert `PCM` to `Stft` at the run-time.
- Now `dump.sh` dumps pnly for applying `CMVN` and `Add-Delta`, but this is the waste of disk space. 
- To make more easy to connect espnet to some front-end. I implemented a process to use `nara-wpe` for an example of this.

I used in the pre-processing as an example in `ljspeech/tts1/run.sh` , this is not for merged.

You can understand the mechanism by reading `espnet.utils.io:PreProcessing` and  `espnet.utils.io:LoadInputsAndTargets`.

sw005320:
@kamo-naoyuki, many thanks for your great efforts!
I'm really excited about this extension.

There are my minor comments
1. scp format
I just want to make sure what `:uttid1` in the following format means. Kaldi uses the onset (offset?) information here, right? Your intention is to simply use a key ID instead of the onset information to keep the Kaldi-like format?
```
uttid1 somefile.32.h5:uttid1
uttid2 somefile.32.h5:uttid2
...
```

2. Bottleneck due to pre-processing?
There are several demands to use the raw WAV data directly in the training and perform feature extraction (STFT, Mel, etc.) on the fly in python instead of using Kaldi separately to include the raw waveform processing based on neural networks, and this function seems to be really great. Then, I'm curious how much we would have a bottleneck when we perform such feature extraction every run-time during training. We may use the prefetch function default, which could avoid this bottleneck to some extent. 

3. the config format of the pre-processing
Can we also support similar format to the Kaldi config file?

kamo-naoyuki:
> 1. scp format

Yes, that's right. My motivation is that I want to reuse some kaldi's utils, such as `./utils/copy_data_dir.sh`, `./utils/split_data.sh` simply and it is not necessary to imitate kaldi's scp entirely, i.e. the offset.

> 2. Bottleneck due to pre-processing?
>    There are several demands to use the raw WAV data directly in the training and perform feature 

I haven't tested it the feature on the espnet as far the feature extraction,  but I usually use a similar strategy on my DNN recognition system, and, there are no bottleneck because multi-process iterator could cover it, though the performance really depends on the machine resources.

I'll try to measure the performance after all bugs are removed in our machine environment.

> 3. the config format of the pre-processing
>    Can we also support similar format to the Kaldi config file?

What format is it exactly? You mean the following format, right?

```
--sample-frequency=22050
--num-mel-bins=80
--frame-length=46.4       
```

sw005320:
> What format is it exactly? You mean the following format, right?

Yes.

sw005320:
OK. I understand it.

kamo-naoyuki:
All self checks have been done. As far the performance of the runtime feature extraction, I'll report  in another PL.

kamo-naoyuki:
I changed the CLI of these scripts because they are not compatible with kaldi style `scp:wav.scp`, this is user unfriendly.

- compute-fbank-feats.py
- compute-stft-feats.py

https://github.com/espnet/espnet/blob/64d98a25b46798b5d734182804ae2d1a873c8099/utils/make_fbank.sh#L89-L104

https://github.com/espnet/espnet/blob/64d98a25b46798b5d734182804ae2d1a873c8099/utils/make_stft.sh#L85-L96

Additionally, `--normalize` option is appended. This name, `--normalize` , might be not good.
`soundfile.read` normalizes the PCM data to [-1,1] by default when loading wav files, but `kaldiio` , and `Kaldi` also, doesn't it.

sw005320:
> Perhaps, there are some concern, but I'd like to use kaldiio

I don't mind to use either option.
Can you tell me why kaldiio is better?

kamo-naoyuki:
@sw005320  > Can you tell me why kaldiio is better?

- `kaldi-io-for-python` requires `copy-feats` to write `scp` file, but `kaldiio` supports it.
- `kaldiio` supports all compression_methods.
- `kaldiio` supports `wav.scp`
   - It supports the scp of pipe style, such as `uttid cat a.wav |` 
   - `wav.scp` with `segments` style: This is used in several kaldi egs, `csj`, `chime5`.

At my final planning, I'll make independent from kaldi commands such as `copy-feats`, so as possible I avoid it.

sw005320:
> @sw005320 > Can you tell me why kaldiio is better?
> 
> * `kaldi-io-for-python` requires `copy-feats` to write `scp` file, but `kaldiio` supports it.
> * `kaldiio` supports all compression_methods.
> * `kaldiio` supports `wav.scp`
>   
>   * It supports the scp of pipe style, such as `uttid cat a.wav |`
>   * `wav.scp` with `segments` style: This is used in several kaldi egs, `csj`, `chime5`.
> 

I really like the `wav.scp` support as it is essential for us to include the feature extraction pipeline as a computational graph.
We should use `kaldiio`.
Thanks for developing a great tool!

LukasDrude:
Dear Kamo,

I like the way the `WriteHelper` is a context manager with a `__setitem__` method. Nevertheless, I have a few conceptual questions and a few minor issues at the end of this message.

1.
Maybe it is worth documenting somewhere what the core motivation for hdf5 support is. Do you expect less IO overhead or need more complex structures than what an SCP with wav files can provide?

2.
Could you add an example how the JSON would look like if you need more than one utterance? Examples might include a particular background noise file (for data augmentation) or a second speech source. Will this result in more keys on the same level as `"input"`? A potential issue is that the SCP is much less expressive than the JSON.

3.
I believe the hdf5 tooling is well thought through but I would recommend to keep additional pre-processing for the mini-batch out of the current pull request. My concerns are:
- User defined functions can not be called easily. Important extensions may be data augmentation, reverberation and other on-the-fly functions.
- Currently the `processings` are hidden quite deeply in the source tree in e.g. `espnet.utils.processings.add_deltas.add_deltas()`. Would it make sense to move all transformations e.g. to `espnet.transform` since the transformations can be used not only for your on-the-fly-batch-update but also to write the features in the first place?
- Introducing a specific implemenation of e.g. stft now will make it more complicated to change in the future [1].

Small remarks:
- It is confusing that `read_rspecifier()` uses two very similar sounding variables `filetype` and `ftype`. You might want to use `expected_file_type` (the one mentioned only in the json) and `actual_file_type` or even `file_extension` (the one mentioned in the actual path).
- Changing `path, h5_key = value.split(':', 1)` [2] to `path, h5_key = value.split(':')` automatically ensures, that there is just one `:`.
- Why is `FileWriterWrapper` part of `cli_utils.py`? It seems to be used without command line interface (cli), too.
- The warning in [3] should be an error. Otherwise a preprocessing mistake quickly gets hidden.
- Change class name `PreProcessing` -> `Preprocessing` since it is a single word. Otherwise `preprocessing` would be `pre_processing`.

[1] https://github.com/espnet/espnet/pull/493/files#diff-42e0d19c28d1b438100975b94df92cb6R116
[2] https://github.com/espnet/espnet/blob/acbb299d7d410d207b8132f0f6bcdd342d475292/espnet/utils/cli_utils.py#L56
[3] https://github.com/espnet/espnet/blob/acbb299d7d410d207b8132f0f6bcdd342d475292/espnet/utils/cli_utils.py#L98

kamo-naoyuki:
@LukasDrude

Thanks for your comment. This PR is on-going and including much experimental features, so I really want such opinion.

I'll tell you beforehand that I'll complete the draft for multi-channel end-to-end system in the near future, so it may be better to examine this PR after it, though I welcome pointing.

>1.
> Maybe it is worth documenting somewhere what the core motivation  for hdf5 support is. Do you expect less IO overhead or need more complex structures than what an SCP with wav files can provide?

I appended my opinion at the top.

> 2.
> Could you add an example how the JSON would look like if you need more than one utterance? Examples might include a particular background noise file (for data augmentation) or a second speech source. Will this result in more keys on the same level as "input"? A potential issue is that the SCP is much less expressive than the JSON.

I appended the example in https://github.com/espnet/espnet/pull/493#issuecomment-447731276

Though I also agree that `scp` is less expressive,  I just extended `mergejson.py` while being carefully not to make much change. Could you tell me the case that this style doesn't go well?

I think that `scp` have high affinity with `kaldi`, so this style, i.e. `scp`s->`JSON`s-> a `JSON`, seems not so bad.

>  3.
> * User defined functions can not be called easily. Important extensions may be data augmentation, reverberation and other on-the-fly functions.

I was not sure that this feature is really accepted, so I'm grad to hear your request. O.K. I'll consider it.

> * Introducing a specific implemenation of e.g. stft now will make it more complicated to change in the future

Sorry, I'm not sure what you pointed. Do you need the each explanation for these functions?

Thank you for the other suggestions, I'll refactor base on them.
It's not a problem If you don't mind, but this PR is on-going, so too small remarks in the current status might be waste of time.

kamo-naoyuki:
> * Currently the `processings` are hidden quite deeply in the source tree in e.g. `espnet.utils.processings.add_deltas.add_deltas()`. Would it make sense to move all transformations e.g. to `espnet.transform` since the transformations can be used not only for your on-the-fly-batch-update but also to write the features in the first place?

I renamed `espnet.utils.processings` -> `espnet.transform`

> * It is confusing that `read_rspecifier()` uses two very similar sounding variables `filetype` and `ftype`. You might want to use `expected_file_type` (the one mentioned only in the json) and `actual_file_type` or even `file_extension` (the one mentioned in the actual path).

Rename `filetype` -> `ark_or_scp`.  I also renamed `read_rspecifier()` -> `ReadFileWrapper` for symmetry.

> * Changing `path, h5_key = value.split(':', 1)` [2] to `path, h5_key = value.split(':')` automatically ensures, that there is just one `:`.

I gave `maxsplit=1` because the key of `HDF5` can includes `:`.  I don't think 

> * Why is `FileWriterWrapper` part of `cli_utils.py`? It seems to be used without command line interface (cli), too.
> * The warning in [3] should be an error. Otherwise a preprocessing mistake quickly gets hidden.
> * Change class name `PreProcessing` -> `Preprocessing` since it is a single word. Otherwise `preprocessing` would be `pre_processing`.
> 
> [1] https://github.com/espnet/espnet/pull/493/files#diff-42e0d19c28d1b438100975b94df92cb6R116
> [2]
> 
> [espnet/espnet/utils/cli_utils.py](https://github.com/espnet/espnet/blob/acbb299d7d410d207b8132f0f6bcdd342d475292/espnet/utils/cli_utils.py#L56)
> 
> Line 56 in [acbb299](/espnet/espnet/commit/acbb299d7d410d207b8132f0f6bcdd342d475292)
> 
>  path, h5_key = value.split(':', 1) 
> 
> [3]
> [espnet/espnet/utils/cli_utils.py](https://github.com/espnet/espnet/blob/acbb299d7d410d207b8132f0f6bcdd342d475292/espnet/utils/cli_utils.py#L98)
> 
> Line 98 in [acbb299](/espnet/espnet/commit/acbb299d7d410d207b8132f0f6bcdd342d475292)
> 
>  logging.warning('Text mode is not supported for HDF5')

kamo-naoyuki:
@LukasDrude 

> * Currently the `processings` are hidden quite deeply in the source tree in e.g. `espnet.utils.processings.add_deltas.add_deltas()`. Would it make sense to move all transformations e.g. to `espnet.transform` since the transformations can be used not only for your on-the-fly-batch-update but also to write the features in the first place?

Renamed to `espnet.transform`.

> * It is confusing that `read_rspecifier()` uses two very similar sounding variables `filetype` and `ftype`. You might want to use `expected_file_type` (the one mentioned only in the json) and `actual_file_type` or even `file_extension` (the one mentioned in the actual path).

Rename `ftype` -> `ark_or_scp`.  I also refactored `read_rspecifier()` to `class
 ReadFileWrapper` for symmetry.

> * Changing `path, h5_key = value.split(':', 1)` [2] to `path, h5_key = value.split(':')` automatically ensures, that there is just one `:`.

I left it . 
The key of h5file can include `:`, so I gave `maxsplit=1`.  Perhaps, it should be used for hierarchy . e. g. `a.hdf5:key1:key2` -> `h5py.File('a.hdf5')['key1']['key2']`

> * Why is `FileWriterWrapper` part of `cli_utils.py`? It seems to be used without command line interface (cli), too.

it is used only in `cli` now, so I left it. 
Dumping fbanks in `tts-decode` may be that case, but I want to keep the interface for now, because this PR became very big.

> * The warning in [3] should be an error. Otherwise a preprocessing mistake quickly gets hidden.

Changed.

> * Change class name `PreProcessing` -> `Preprocessing` since it is a single word. Otherwise `preprocessing` would be `pre_processing`.

Renamed to `Preprocessing`.

LukasDrude:
Thank you for addressing my concerns. The motivation for hdf5 is now clear. Nevertheless, I still thing it is better to load as little data as possible and do as much light-weight calculation (e.g. stft) on-the-fly. A typical example might be reverberation: Creating the room impulse response is slow (it should thus be stored in hdf5 in time domain) while convolution and stft is fast, thus this can be done on the fly).

> I think that scp have high affinity with kaldi, so this style, i.e. scps->JSONs-> a JSON, seems not so bad.

Yes, I think it is ok to rely on `scp` as long as it supports the use case at hand. I now understand that you simply use more than once `scp` file, when you need more than one input feature.

> I was not sure that this feature is really accepted, so I'm grad to hear your request. O.K. I'll consider it.

We internally handle this as follows: The user may create classes (e.g. similar to `LogMelSpectrogram`). The user than provides the import string (e.g. `my_packaege.transform.MyFancyTransform`) to the main code (e.g. in your `preprocess.json` including `kwargs` for that class). That way, the user can call his own code without introducing changes into Espnet.

>> Introducing a specific implemenation of e.g. stft now will make it more complicated to change in the future

> Sorry, I'm not sure what you pointed. Do you need the each explanation for these functions?

I think my wording was a bit misleading. I think the current implementation (all transforms in `espnet.transform` is nice and can be used from anywhere in the project. I already saw, that you use it in [1], too.

[1] https://github.com/espnet/espnet/blob/1227c4734df44e932bead7ebf7a26cfa7d576da7/utils/compute-stft-feats.py#L13

kamo-naoyuki:
# Essential tools for inputting `wav` when training
So far, I prepared some `utils` for handling `HDF5` with very similar style of `Kaldi-matrix-ark`, `Kaldi-scp`. In the last week, I also created additional scripts to input `wav` for feature file.

## Core motivation

-  I believe that IO cost is the most big bottleneck for DNN. Especially, this is a fatal problem for multi-channel training or audio-visual task.
- Basically, more times to invoke system-call for fd such as `open(2)`, `read(2)` , the IO cost becomes bigger and sequential reading is faster than random reading. Thus it is better to use archive format, such as `kaldi-ark` and `HDF5` rather to use many files for each utterances. 

Thus I **rejected** the following strategy for multi-channel recognition system.

- Saving feature file for each channel: We should compose into  a tensor having multi-channel .
- Using STFT feats for each channel: e.g. The dimension becomes (Time, 512, 2, Nch): We should use `wav` for inputting file format and apply some transformations to extract feature in the training time.

## Compose single channel files into multi-channel wav
- gather-wav.py
```bash
# Concatenate 2 single channel wav files into 2ch-wav
gather-wav.py sound_1ch.wav sound_2ch.wav > sound.wav
# You can use pipe in kaldi style
gather-wav.py "cat sound_1ch.wav |"  "cat sound_2ch.wav|" > sound.wav
```
- gather-wav-scp.py

```bash
% cat 1ch.scp
a foo_1ch.wav
b bar_1ch.wav

% cat 2ch.scp
a foo_2ch.wav
b bar_2ch.wav

% gather-wav-scp.py 1ch.scp 2ch.scp
a gather-wav.py "foo_1ch.wav" "foo_2ch.wav" |
b gather-wav.py "bar_1ch.wav" "bar_2ch.wav" |
```

## Dump wav files into a `HDF5` with `flac` format

```bash
dump-pcm.py --filetype hdf5 scp:wav.scp ark,scp:out.h5,out.scp
dump-pcm.py --filetype sound.hdf5 scp:wav.scp ark,scp:out.flac.h5,out.scp
```

To make the file size smaller, I decided to use compress-format for dumping `PCM`: `flac`. 

```python
from espnet.utils.io_utils import SoundHDF5File
with SoundHDF5File('out.flac.h5', 'w') as f:
    f['uttid'] = (signal, sampling_rate)  # Compressed in flac by default.
with SoundHDF5File('out.flac.h5', 'w') as f:
    (signal, sampling_rate) = f['uttid'] 
```

This module saves the raw bytes of `flac` file into `HDF5` (Actually, there might be few meanings to use `HDF5` if type and shape information is not saved) 

## `--preprocess-conf` option 
In some cases, we need the extracted feature before training: e.g. To know the shape of feats, To calculate the statistics for CMVN. For them, I appended `--preprocess-conf` option to some scripts.

```bash
feat-to-shape.py --filetype sound.hdf5 --preprocess-conf preprocess.json scp:out.scp  shape.scp
compute-cmvn-stats.py --in-filetype sound.hdf5 --preprocess-conf preprocess.json scp:out.scp stats.mat
```

kamo-naoyuki:
I almost finished. Sorry to very big PR.

I created `egs/chime4/asr1_multich` to show an example of these system. The recognition result is very bad now and, actually, it is not even multi-channel, i.e. In the training time, single channel is selected from the input multi-channel wave file by `espnet.transform.channel_selector`.

In the next PR, I'll prepare actual end-end system with some DNN-beamformer algorithm. Before it, we should support `PyTorch=1.0` because it supports batch-inverse operation for matrix. 

BTW, I also implemented `using_transform_config` to change the behavior depending whether training or eval phase This feature is inspired by `chainer.using_config`. The usage can be found in the docstring of `espnet.transform.channel_selector`.

kamo-naoyuki:
@LukasDrude 
 
Yeah, I'll measure the performance after this PR and I'd like you to do too, if possible, because the performance depends on machine environment.

I note that already sound file format are supported in module level using `soundfile`, so it is not hard 
 to use wav files for you. Though I agree that it is better to support, but I hesitate to prepare it in the recipe level, i.e. `egs` and `utils`, due to not essential problem.

- The wave file format is different among corpora: e.g. wsj use sphere format.
- In some corpora, we must extract segments from the distributed wav files. e.g Chime5 deals each session data and each people need to extract utterances from it. In this case, the duplication is inevitable.

The most annoying problem is that we may reuse Kaldi recipe for each corpora, but that `wav.scp` is not a simple format, you may know, Lukas: It can embed unix commands.

```
uttid cat a.sph | sph2wav |
```

and, they try to cut segments in pipeline when feature extraction. This is very slow if the a wav file is big, so I don't recommend you to do it on-the-fly.
Ideally `espnet` reuse almost all part of a recipe for the other recipe, but if we'll support inputting wave files in the recipe level, we must consider how to absorb the above formats. I don't have good idea for it other than to dump it.

And, about IO speed, in my workplace we shares a computer cluster among many people and we use NFS there, you know, and that IO speed is not comparable to DNN for many cases. We also need to care the IO stress for other people. Our storage may be more poor than yours, but in general the scalable storage for cluster computing is not easy. 
  
`sequence2sequence` uses bigger data than other DNN task, e.g. image classification, since it uses sequences of some data,  so that cases is not rare. `transformer` may be notable.

Although having say that I'll think how to design to support raw wave inputting.

kamo-naoyuki:
@LukasDrude 

I modified utils slightly and now you can use raw files as it is easily only in the this case:  You have the following `wav.scp` which doesn't use pipeline and already **segmented** and has single channel.

```
uttid0 /some/path/a.wav
uttid1 /some/path/b.wav
...
```

Then,  you could get the JSON file:

```
% data2json.py --feats wav.scp --filetype sound data/train dict_file > out.json
```
```json
{"utt": {"input": [{"feat": "/some/path/a.wav"}], ... }
}
```

The JSON can work with `asr_train.py`.
However, this style shouldn't be used in `egs` explicitly. This feature is supported in `utils` level only and I still recommend `HDF5` or `kaldi-ark`.

kamo-naoyuki:
Small remarks: I'm not sure but `hacking` may enables `W503` and `W504` both:

```
# W503 recommends
a = ('foo' + 
       'bar')

# W504 recommends
a = ('foo' 
       + 'bar')
```

I disabled W504. This is a bug of `hacking`, or some module.

LukasDrude:
From my point of view the PR is fine. I understand that the `scp`s which contain shell commands need to be dumped first. Other than that it is useful to allow to load `wav` directly, as you suggested.

kamo-naoyuki:
@sw005320 @kan-bayashi Almost done. Please review.

 I changed `egs/an4/asr1/run.sh` and `egs//ljspeech/tts1/run.sh`, but, as I said beforehand, I would reverted them in the final.

kamo-naoyuki:
I have reverted an4/asr1/run.sh and ljspeech/tts1/run.sh. Lastly, I want to say,

- I haven't read `tts` part much , so please examine my modification after merged.
- Keep in mind that `HDF5` is said to have a problem with palatalization:
  - https://discuss.pytorch.org/t/hdf5-multi-threaded-alternative/6189
  - I also met some error with parallel iterator, but it can't be reproduced now.
- I used too much time about this, so I have to tackle another task for a while.

PR #484, opened 2018-11-28
Modularization, cleaning, correcting

gtache:
Here is the first of several PRs for #483. This one focuses on documentation, typos, consistency and code maintainability.

- Separates CTC, BLSTM/P, VGG2L, Attention modules and utility functions in different files for better readability and maintainability
- Creates subfolders for chainer and pytorch
- Removes duplicate make_lexical_tree
- Fixes several typos
- Uses a consistent docstyle (except in deterministic_embed, I'm not really sure what to do with it)
- Adds some documentation
- Puts braces around bash variables for safety and consistency
- Adds shebang on top of bash files when missing
- Changes extension of callhome_make_spk2gender.sh and fsp_make_spk2gender.sh to .py (as they are python files) and changes script calling them accordingly
- Uses 4 spaces everywhere
- (Uses Pycharm to easily reformat files in a nicer way)

Sorry, it's a big PR with not many changes per file, but I think it's important to fix this type of things early if the codebase starts growing.    
Don't hesitate to comment!

Issue #483, opened 2018-11-28
Improvements, additions, fixes


gtache:
Hi, thanks for the framework, it does a good job given how small and easy to use it is, which also allows to easily modify it !     
I'm using it for a project and had to make quick modifications, but now that I have more time I'd want to merge them to allow everyone else to benefit from them. It may take some time to rewrite them (I want to do it in a cleaner way) but here are some things I've already implemented in my version:    
- Additions
- - [x] GRU/LSTM choice #496
- - [ ] Encoder customization #551 
- - [x] Sortagrad #550 
- - [ ] ColdFusion #566 
- - [ ] Dynamic batchsize #580
- - [ ] Incremental scheduled sampling #576 
- Fixes
- - [x] Fix for deterministic behavior #497 
- - [x] Typos #484
- Improvements
- - [x] Consistent documentation #484
- - [x] Code modularization #484
- - [ ] Various other optimizations    
- QoL
- - [x] Tensorboard logging #506 
- - [x] EarlyStop trigger #508 
- - [x] Stop stage for run.sh #539 

You can tell if you want to put priority on some or want to remove others.

Note that some of these modifications are only on pytorch for now but if it isn't too complicated I may implement them on chainer too (I have no experience with chainer though).

The first PR #484 is just for typos, docs and modularization (no functionality change)

sw005320:
Thanks!
All of them look very interesting. Yes, let's try to incorporate it one-by-one.
I'm also curious how these additions will improve the performance, e.g., Sortagrad and Dynamic batchsize.

PR #496, opened 2018-12-07
Adds GRU support to ASR and LM

gtache:
Going on with #483 

- Adds GRU to Encoder, LM and Decoder

Sorry for the bugs with #484, I hope there aren't too many.


gtache:
If there is someone more knowledgeable with chainer understanding why the chainer decoder doesn't work with GRU and how to transfer weights for a GRULM, that'd be great !

gtache:
Well, I don't have the resources to do experiments right now sadly. When I tried with the encoder it seemed that LSTM were better indeed. I've however been using a GRU for the decoder for a bit and could get reasonable results on Tedlium.     
I've never said that GRU would improve the results however, it may depend on the dataset (one with shorter samples may be more suitable perhaps). GRU have the advantage of being faster to train. Well, from what I've read on GRU vs LSTM is that we can't really know which one is better before trying most of the time.     
This is mostly to offer the choice to users, if they want to compare or to reproduce different architectures from papers or so. The ColdFusion paper for example uses GRU for the decoder and the LM.

sw005320:
@Fhrozen, I like @gtache's idea of adding GRU options anyway to have possible options in RNNs. Also, do you have any ideas of why the chainer decoder doesn't work with GRU?

espnet/nets/chainer_backend/decoders.py
```diff
@@ -43,6 +45,17 @@ def __init__(self, eprojs, odim, dlayers, dunits, sos, eos, att, verbose=0,
         self.lsm_weight = lsm_weight
         self.sampling_probability = sampling_probability
 
+    def rnn_forward(self, ey, z_list, c_list, z_prev, c_prev):
+        if self.dtype == "lstm":
+            c_list[0], z_list[0] = self.rnn0(c_prev[0], z_prev[0], ey)
+            for l in six.moves.range(1, self.dlayers):
+                c_list[l], z_list[l] = self['rnn%d' % l](c_prev[l], z_prev[l], z_list[l - 1])
+        else:
+            z_list[0] = self.rnn0(z_prev[0], ey)
```
Fhrozen: it will be better if you add something like this:
```
if z_prev[0] is None:
                xp = self.xp
                with cuda.get_device_from_id(self._device_id):
                    z_prev[0] = variable.Variable(
                        xp.zeros((ey.shape[0], self.dunits), dtype=ey.dtype))
```
because GRU does not prepare a zeros array as Stateless LSTM does.

Fhrozen:
@sw005320 
> do you have any ideas of why the chainer decoder doesn't work with GRU?
If you mean about using `None`, then it can be solve with the part of code attached in the review. Besides that, I have no problem in executing the GRU decoder in chainer.

espnet/nets/chainer_backend/decoders.py
gtache: Thanks for the fix ! That's a strange design choice by chainer to have different behaviors for StatelessLSTM and StatelessGRU though.

gtache:
Only the LM to fix now

Fhrozen:
@gtache, could you gimme some information about transfer weights for a GRULM. There is a paper where I can check it.
I am testing right now the LM with CHiME-5 because of it is the smallest one. But, I will need to know exactly if what I am doing is correct, so I need a reference =).

Fhrozen:
Ok, I see, You want to transfer weight between backends... duh.. sorry I misunderstand. XD

sw005320:
@unnonouno, did you know the StatelessLSTM and StatelessGRU behaviors in chainer? Is that intended? Can you have any advice about this?

sw005320:
I'm revising this PR. What is the conclusion for the chainer backend? We can skip it (put some appropriate error handling) if it does not work and only support the pytorch backend.

gtache:
Everything's working, the only problem is the LM test, where the results are not the same for chainer and pytorch, because I don't know how to do the weight transfers as with LSTM. But in practice everything should work, even if the final result is not exactly the same.     
So in the meantime I've disabled the LM test for GRU.

sw005320:
@gtache, this work seems to cause a lot of conflicts with #563 done by @hirofumi0810. Could you communicate each other to find a solution? I actually like #563's change on unifying the beam search of MT and ASR, and if redoing this PR after #563 is easier, we should revert it.

gtache:
@sw005320 I looked into the PR and cloned it to check, but it seems that my changes are already merged without any visible conflict, and the only reason the tests are failing is because `make_arg` in test_e2e_asr doesn't specify `sos`.    
I'd be happy to help if @hirofumi0810 tells me what is wrong of course.

hirofumi0810:
@gtache The problem is the variable name (parameter name).
You changed self.lstm (self.nblstm) to self.rnn (self.nrnn) in the encoder and RNNLM parts.
A lot of people are running various experiments in parallel, so changing names makes it impossible to load model parameters before its change.
Please take it into account.

sw005320:
@hirofumi0810, I understand your point. My suggestion is to work on your #563 without merging #496. Once you finish it, you'd merge #496 and redo the experiment to confirm it. This happens unfortunately and this is the reason that we are developing it on the v.0.3.0 branch to avoid it affect to the other people using ESPnet with the master branch. #496's name change is necessary to unify LSTM and GRU and I would ask matured developers like you working on the v.0.3.0 to be patient with it.

PR #497, opened 2018-12-07
Fix deterministic pytorch

gtache:
[Setting deterministic = True isn't sufficient](https://github.com/pytorch/pytorch/issues/6351)

### DIFF
--- a/espnet/asr/chainer_backend/asr.py
+++ b/espnet/asr/chainer_backend/asr.py
@@ -203,7 +203,7 @@ def train(args):
 
     # debug mode setting
     # 0 would be fastest, but 1 seems to be reasonable
-    # by considering reproducability
+    # considering reproducibility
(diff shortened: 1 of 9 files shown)


kamo-naoyuki:
Could anyone test on tts? Some bugs remain.

At least, I need to insert these lines to `espnet.bin.tts_decode.py` 

```python
    parser.add_argument('--debugmode', default=1, type=int,
                        help='Debugmode')
    parser.add_argument('--seed', default=1, type=int,
                        help='Random seed')
```

gtache:
@kamo-naoyuki Indeed, I hadn't seen that tts_decode doesn't have debugmode and seed arguments... that's the type of problem I want to avoid with my next PR (will be done after earlystop is merged)      
I added the quickfix in the earlystop PR

Issue #476, opened 2018-11-25
Python3 support for scripts using numpy


ShigekiKarita:
https://docs.scipy.org/doc/numpy-1.14.1/neps/dropping-python2.7-proposal.html

Since numpy will drop python2 from Jan. 2019, we should rewrite python2 scripts with numpy under `utils` to python3 compatible. Here is a list of python2-only numpy scripts under `utils`

- splitjson.py

(EDITED these files seem to be python3 compatible already)

- compute-fbank-feats.py
- compute-stft-feats.py
- convert_fbank_to_wav.py

Do you have any other python2-only script with numpy in espnet?

sas91:
@sw005320 run_wpe.py works on python3, I can modify the shebang. Are you going to change the default python installed in tools to be python3?

sw005320:
It's fine if it's working with python3.

kan-bayashi:
> - compute-fbank-feats.py
> - compute-stft-feats.py
> - convert_fbank_to_wav.py

These above function is compatible with python3.

Some data prep scripts are working on only python2.
For example:
https://github.com/espnet/espnet/blob/master/egs/hkust/asr1/local/hkust_segment.py

This function uses `mmseg` which only support python2.
https://github.com/42qu/mmseg

kan-bayashi:
Sorry, I missed `using numpy`.
Please forget about it.

ShigekiKarita:
Thanks. Anyway we need to make the scripts python3 compatible until EOL of python2, Jan 2020.

ShigekiKarita:
Maybe this PR can fix it. https://github.com/42qu/mmseg/pull/2

PR #498, opened 2018-12-08
Tests for utils by using bats-core

kamo-naoyuki:
This is a proposal PL for unit tests for CLI tools using [bats-core](https://github.com/bats-core/bats-core).

Relating to #493, I'll modify some scripts in `utils`, but there are no tests now. I feel uneasy. In addition, you are talking about python2-3 compatibility in #476, and this system may help you. 

This PR is just a proposal, so I appended 3 codes only.

I use `bats-core` for the first time, and actually I never used no test tools for CLI, so if you know better tool, please tell me, though `bats-core` seems enough.


kamo-naoyuki:
Tests failed because `utils` of v.0.3.0 branch, which is updated by #484, can't use non-ascii character now.

kamo-naoyuki:
@sw005320 seems no problem.

PR #499, opened 2018-12-08
Fix #484

gtache:
Sorry for the trouble with the last PR.

I still have to look into the unicode problem, worst case I will just revert them.


gtache:
Okay, apparently it worked when I tested before because the stdout encoding is different when piping a script. Well you learn something new everyday. So I've wrapped everything with codecs and that should take care of the problem, and after using `cmp` to check the 4 results generated by the commands :     

`cat text | python2/3 utils/text2token.py -s 1 -n 1 | awk -v maxframes=200 '{ if  (NF < 201) print }' | awk -v minchars=100 '{ if (NF > 101) print }' | awk '{print $1}' > tmp2/3`     
and      
`python2/3 utils/text2token.py -s 1 -n 1 text | awk -v maxframes=200 '{ if  (NF < 201) print }' | awk -v minchars=100 '{ if (NF > 101) print }' | awk '{print $1}' > tmp2/3`

with text being voxforge `all_it/text`, every file is the same. I've also compared against the file generated by the old text2token script on master, and it's the same result, so it should be good now.         
I've also managed to make a docker image to run espnet on cpu only, and I was able to run voxforge (although decoding doesn't work due to recognize_beam_batch not working on cpu if I remember correctly?).

The next PRs should be a lot more focused and shouldn't touch much the utils or run scripts, so I'll have a lower chance of breaking things 😅

PR #506, opened 2018-12-17
Tensorboard support

gtache:
Adds tensorboard support


sw005320:
Thanks @gtache!
I have several comments:
1. we may create the `espnet/utils` directory, and put tensorboard_logger.py there (we can also move to your bin_utils.py to there with some appropriate naming change.
2. can you add the usage of how to use tensorboard to `README.md`?

gtache:
@sw005320 Thanks for the feedback. I added the logging to the recipes so that events are automatically generated and stored in `tensorboard/${expname}` (same for LM).      
I also moved the files in the new `utils` folder, and added a small section in the README.    
Now I'm just checking that it works with chainer.

sw005320:
@gtache, Thanks!
Can you paste a snapshot? I want to see how it looks like. 

Also, I'm thinking which is better to store the tensorboard logs. 
1. The current one is fine in the sense that all tensorboard logs are stored in `tensorboard/`, and we can easily check the experimental analysis by only checking this directory.
2. I actually don't know how tensorboard is working, but we could also use `${expdir}` in `run.sh` and put the files for each experiment there (e.g., in `${expdir}/tensorboard/`). Then, we can follow all our current directory structure, that is, all experimental results and output files are under each experimental directory specified by `${expdir}$`. But if it is not a desired directory structure for tensorboard, we don't have to consider it.
3. this is alternative option of 2. We can reuse `--outdir` option and put the tensorboard logs in `${expdir}/results` or `${expdir}/results/tensorboard/`

gtache:
Here is a screenshot from two dummy runs :        
2018-12-18_19h49_48
As you can see, tensorboard uses the subfolders names to name the experiments, so I'd say the current structure is the best, you just have to run `tensorboard --logdir tensorboard` to see all the results. I don't know how tensorboard would react if the subdirectory structure did contain other things than tensorboard logs.

sw005320:
Thanks @gtache!
Now I understood the behavior, and sorry for my wrong comments based on my misunderstanding of tensorboard. Your current implementation seems to be good. My only comment is that you could move the description of the tensorboard in `README.md` right after the explanation about the logging function of ESPnet, e.g., 

The training progress (loss and accuracy for training and validation data) can be monitored with the following command
```
$ tail -f exp/${expdir}/train.log
```
With the default verbose (=0), it gives you the following information
```
epoch       iteration   main/loss   main/loss_ctc  main/loss_att  validation/main/loss  validation/main/loss_ctc  validation/main/loss_att  main/acc    validation/main/acc  elapsed_time  eps
:
:
6           89700       63.7861     83.8041        43.768                                                                                   0.731425                         136184        1e-08
6           89800       71.5186     93.9897        49.0475                                                                                  0.72843                          136320        1e-08
6           89900       72.1616     94.3773        49.9459                                                                                  0.730052                         136473        1e-08
7           90000       64.2985     84.4583        44.1386        72.506                94.9823                   50.0296                   0.740617    0.72476              137936        1e-08
7           90100       81.6931     106.74         56.6462                                                                                  0.733486                         138049        1e-08
7           90200       74.6084     97.5268        51.6901                                                                                  0.731593                         138175        1e-08
     total [#################.................................] 35.54%
this epoch [#####.............................................] 10.84%
     91300 iter, 7 epoch / 20 epochs
   0.71428 iters/sec. Estimated time to finish: 2 days, 16:23:34.613215.
```
In addition, [Tensorboard](https://www.tensorflow.org/guide/summaries_and_tensorboard) events are automatically logged in the `tensorboard/${expname}` folder. You can therefore easily compare several experiments by using 
```
tensorboard --logdir tensorboard
``` 
and connecting to the given address (default : localhost:6006). This will provide the following information:
2018-12-18_19h49_48

sw005320:
@ShigekiKarita, @kan-bayashi, do you have any comments? I'll merge this PR soon. @shrutijpalaskar, do you also have any comments, as you requested such functions earlier.

kan-bayashi:
Looks great!
If we can see the attention weight through tensorboard like Shigeki’s one, that is awesome.

gtache:
@kan-bayashi I may be able to add that, I'll see what I can do.

gtache:
Okay, it's added now, you can change find the plots in the images tab and you can change the epoch with the slider, like this : 
2018-12-19_14h28_33
2018-12-19_14h28_49

PR #508, opened 2018-12-19
Early stop trigger

gtache:
Adds an early stop trigger for all trainings


sw005320:
Thanks @gtache!
Is `patience=10` OK?
Early stop may not be triggered in the most of cases, since the validation loss started to be decreased in 5 or 10 epochs in most cases based on my experience, and our maximum number of epochs is usually set as 10 or 15 (e.g., 5 + 10 = 15 > 10).
We could set it smaller, e.g., 3?

gtache:
Indeed, in this case it makes sense to reduce it.

sw005320:
- We may also set that default value in the python bin scripts (i.e., `'--patience', default=3,`)
- If the early stoping is triggered, we should add some logging information to note that the early stoping is triggered. Then, users could understand why the training is stopped and they would control the patience parameter by themselves if needed.

gtache:
@sw005320 done     
I got a bit carried away and have done some refactoring, I got annoyed having to change the same things in 4-5 different files      
I'll do some test runs and reread the changes to make sure everything is still working good.

sw005320:
> If you want me to do another PR for the refactoring I can do it

Year, it is better to make another PR for this refactoring.
This looks really good, but it would affect several on-going work, and I want to proceed with it carefully.

sw005320:
@kan-bayashi, it this OK for TTS?
As far as I remember, the TTS training behavior is different and it may be better to diable the patience parameter.

sw005320:
@gtache, I have found the following issue:
- We usually check the validation accuracy for ASR training (e.g., epsilon decay trigger), not loss. Then, I observed some inconsistency, e.g., the training is finished even when the validation accuracy is increased. In general end-to-end ASR, we have very strong correlations between validation accuracy and the final WER/CER. 
- I was actually changed the criterion from the loss to accuracy, but the patience option is checking the minimum value and simply changing it to accuracy causes negative effect https://github.com/espnet/espnet/blob/v.0.3.0/espnet/asr/pytorch_backend/asr.py#L399-L401

Can you consider to fix it?

PR #511, opened 2018-12-25
hkust: fix word segment script

songmeixu:
indent error


kan-bayashi:
Thank you @songmeixu.
This is my mistake.

PR #518, opened 2018-12-29
Short time recipe for integration test

kamo-naoyuki:
There aren't short-time recipes for `tts`, so I created `egs/an4/tts1`.  This recipe is aimed at the integration tests for developers.

The output wav sounds ugly now though it's well for test purpose. I'm not really familiar with `tts`.

### DIFF
--- a/egs/an4/tts1/run.sh
+++ b/egs/an4/tts1/run.sh
@@ -128,20 +128,22 @@ if [ ${stage} -le 1 ]; then
 
     # Generate the fbank features; by default 80-dimensional fbanks on each frame
     fbankdir=fbank
-    make_fbank.sh --cmd "${train_cmd}" --nj ${nj} \
-        --fs ${fs} \
-        --fmax "${fmax}" \
-        --fmin "${fmin}" \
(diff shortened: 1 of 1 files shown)


kamo-naoyuki:
I appended `egs/yesno`.  This is more suitable for this purpose.

sw005320:
> There aren't short-time recipes for `tts`, so I created `egs/an4/tts1`. This recipe is aimed at the integration tests for developers.
> 
> The output wav sounds ugly now though it's

@kan-bayashi, can you check this?
I think this is a great idea to include a small TTS example.
It would reduce your efforts of testing the TTS recipes.

kan-bayashi:
I tested following recipes:
- `egs/an4/tts1`
- `egs/yesno/asr1`
- `egs/yesno/tts1`

I slightly modified as follows:
- Changed dev subset of an4 to be same as ASR
- Added missing patience parameters
- Changed default patience parameter (=10)

kan-bayashi:
Done!

kan-bayashi:
After travis check, I will merge.

PR #520, opened 2018-12-30
Chainer MultiGPU fix v.0.3.0

Fhrozen:
There are few observations:
The Loss E2E code from previous versions returned `return self.loss`
Now, the E2E model is returning `return self.loss, loss_ctc, loss_att, acc`. This generates a error in the [Worker](https://github.com/chainer/chainer/blob/6b292dee96ad7b281589632ccecdd1348eef0f5f/chainer/training/updaters/multiprocess_parallel_updater.py#L238) of the ParallelUpdater since the code is not controlling the return of the self.loss only.
Given this, I added a flag to return only Loss during training. This flag was added in order to avoid additional changes in the code, especially in the test that request the loss as a part of a tuple.
The updates were tested on a docker container. My current setup (if I do not use docker) is having a miscalculation in the grad (Extremely larges grads and some of then Nans), so this is the only way to test.

I also added a fix for tensorboard. Gitignore and when the observations is a cupy object.
Let me know if there should be another changes.

### DIFF
--- a/espnet/nets/chainer_backend/e2e_asr.py
+++ b/espnet/nets/chainer_backend/e2e_asr.py
@@ -72,7 +72,7 @@ class E2E(chainer.Chain):
         self.loss = None
         self.flag_return = flag_return
 
-    def __call__(self, xs, ilens, ys, _old=False):
+    def __call__(self, xs, ilens, ys):
         """E2E forward
 
         :param xs:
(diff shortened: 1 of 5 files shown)


espnet/nets/chainer_backend/e2e_asr.py
```diff
@@ -70,8 +70,9 @@ def __init__(self, idim, odim, args):
 
         self.acc = None
         self.loss = None
+        self.flag_return = flag_return
 
-    def __call__(self, xs, ilens, ys):
+    def __call__(self, xs, ilens, ys, _old=False):
```
sw005320: Is `_old` used?

Fhrozen: Sorry, i did not realize that I forgot to delete that variable. I was testing different ways to return the loss and acc and this was one of those. I just updated the PR without this var.

PR #522, opened 2019-01-03
Avoid utt2num_frames regeneration in LibriSpeech recipe

akreal:
First `utt2num_frames` is not combined by `utils/combine_data.sh`, and next it is not copied in `reduce_data_dir.sh` called from `remove_longshortdata.sh`.


sw005320:
Thanks a lot!

PR #529, opened 2019-01-09
Add encoding for the case if LC_ALL=C:   setup.py

kamo-naoyuki:
To avoid unicode problem for setup.py when LC_ALL=C.


kamo-naoyuki:
Changed to send PR to master branch

PR #539, opened 2019-01-11
Add stop_stage option

gtache:
Just a little QoL change to allow stopping a run.sh at any stage.     
For example, sometimes I want to train a RNNLM without training a network directly after.


gtache:
I don't know why the conda builds fail though.

sw005320:
I'm not sure either.... Anyway #493 deletes the kaldi-io-for-python related in the v.0.3.0 branch, but Travis CI still uses the `.travis.yml` in the master branch (not the one in the v.0.3.0 branch), which seems to cause the error. I just reran the check and now it also happens in the all environments. I cannot understand why it was not appeared before then.

kamo-naoyuki:
Maybe, the travis file is picked from the source branch, `gtache:stopstage`, so it still use `kaldi-io-python`, while the testing run on merged branch `espnet:v.0.3.0` + `gtache:stopstage`.

Could you merge `espnet:v.0.3.0` into `gtache:stopstage` to use new travis.yml?

gtache:
Okay that worked, thanks!

sw005320:
I think it looks good. 
I sometimes want to skip the language model training and with this `stop_stage` we can skip it like that
`./run.sh --stage 0 --stop_stage 2; ./run.sh --stage 3`
@kan-bayashi, do you have a comment?
I remember that you had some similar mechanisms for your tools.

sw005320:
My major comment is that we need to set the stage number twice in `run.sh`, e.g `3` in `if [ ${stage} -le 3 ] && [ ${stop_stage} -ge 3 ]; then`.
It might be a little bit complicated and cause some bugs.

@kan-bayashi's method is a different way and this might be another possibility.
https://github.com/kan-bayashi/PytorchWaveNetVocoder/blob/master/egs/ljspeech/sd/run.sh#L12-L25
https://github.com/kan-bayashi/PytorchWaveNetVocoder/blob/master/egs/ljspeech/sd/run.sh#L208

We may find other ways.

kamo-naoyuki:
Can @kan-bayashi 's way support more than 10th stage?
It seems bit hacky and there is no need to control the stages such finely e.g. `stage=035679` in many cases.

I think @gtache 's way is not less maintainability, but how about this?

```bash
if python -c "assert ${stage} <= 2 <= ${stop_stage}" &> /dev/null; then
    :
fi
```

sw005320:
I see. 
Then, @gtache's way would be better and quite intuitive.
Let me think one day, but if there is no other ways, I'll merge this.

kan-bayashi:
As @kamo-nayuki said, my way cannot support more than 10th stages.
In my usage in that repository, sometimes I want to skip some stage so I made it this
format.

kamo-naoyuki:
This way supports 10 or later (but hard to understand ). 

```bash
stage="1 2 3 4"
. utils/parse_option.sh
stage=( $stage )


if for i in ${stage[@]}; do [ "${i}" = 1 ] && break; done; then
    echo "1st stage"
fi

if for i in ${stage[@]}; do [ "${i}" = 2 ] && break; done; then
    echo "2nd stage"
fi
```

sw005320:
OK, I decided to adopt @gtache's way.

PR #540, opened 2019-01-11
Fix hdf5 update related

sw005320:
Follow up of #493 

- remove "kaldi-io" check in `check_install.py`.
- fix python2.7 unicode issues.

@kamo-naoyuki, I also found that `feat_to_shape.sh` is very slow and consuming a lot of memory. At least we should fix memory issues. Can you check it? I could not make the librispeech recipe run due to huge memory consumption.

### DIFF
--- a/espnet/utils/cli_utils.py
+++ b/espnet/utils/cli_utils.py
@@ -251,12 +251,12 @@ class FileWriterWrapper(object):
         if self.writer_scp is not None:
             if self.filetype in ['hdf5', 'sound.hdf5']:
                 self.writer_scp.write(
-                    '{} {}:{}\n'.format(key, self.filename, key))
+                    '{} {}:{}\n'.format(key, self.filename, key).decode('utf-8'))
             else:
                 raise NotImplementedError
 
         if self.writer_nframe is not None:
-            self.writer_nframe.write('{} {}\n'.format(key, len(value)))
+            self.writer_nframe.write('{} {}\n'.format(key, len(value)).decode('utf-8'))
 
     def __enter__(self):
         return self
--- a/tools/check_install.py
+++ b/tools/check_install.py
@@ -11,7 +11,6 @@ import sys
 # you should add the libraries which are not included in setup.py
 MANUALLY_INSTALLED_LIBRARIES = [
     ('espnet', None),
-    ('kaldi_io_py', None),
     ('matplotlib', None),
     ('torch', "0.4.1"),
     ('chainer', "5.0.0"),


kamo-naoyuki:
Let me make sure, did you not have problem with `copy-feats.py` in `dump.sh` and which `HDF5` or `ark` did you use at that time, or both has memory issues? Could you attach the error log, too?

kamo-naoyuki:
Hmm, h5py might be slow?

```python
import kaldiio
import numpy
import h5py
x = numpy.ones((1000, 80), dtype=numpy.float32)
with h5py.File('feats.h5', 'w') as f:
    for i in range(10000):
        f[str(i)] = x 
with kaldiio.WriteHelper('ark:feats2.ark') as f:
    for i in range(10000):
        f[str(i)] = x 
```
```bash
%  time feat-to-len ark:feats.ark ark,t:out
feat-to-len ark:feats.ark ark,t:out  0.08s user 0.31s system 99% cpu 0.394 total
%  time feat-to-shape.py ark:feats.ark out
feat-to-shape.py ark:feats.ark out  0.45s user 0.30s system 100% cpu 0.745 total
%  time feat-to-shape.py --filetype=hdf5 ark:feats.h5 out
feat-to-shape.py --filetype=hdf5 ark:feats.h5 out  2.18s user 0.75s system 100% cpu 2.931 total
```

sw005320:
> Hmm, h5py might be slow?

Probably. 
But the most serious part is the memory usage.
I'll test your #542 to see whether it is solved or not.

espnet/utils/cli_utils.py
```diff
@@ -251,12 +251,12 @@ def __setitem__(self, key, value):
...
             if self.filetype in ['hdf5', 'sound.hdf5']:
                 self.writer_scp.write(
-                    '{} {}:{}\n'.format(key, self.filename, key))
+                    '{} {}:{}\n'.format(key, self.filename, key).decode('utf-8'))
             else:
                 raise NotImplementedError
 
         if self.writer_nframe is not None:
-            self.writer_nframe.write('{} {}\n'.format(key, len(value)))
+            self.writer_nframe.write('{} {}\n'.format(key, len(value)).decode('utf-8'))
```
kan-bayashi: This is not compatible with python3.

kan-bayashi: Fixed #541

sw005320:
This PR is duplicated with #541 also memory issue is already fixed.

PR #541, opened 2019-01-12
Fixed compatibility for python2 

kan-bayashi:
Fixed python2/3 compatibility.
This PR covers #540.

### DIFF
--- a/tools/Makefile
+++ b/tools/Makefile
@@ -24,6 +24,7 @@ venv:
 
 espnet.done: venv
 	. venv/bin/activate; pip install pip --upgrade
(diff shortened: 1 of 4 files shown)


tools/Makefile
```diff
@@ -24,6 +24,7 @@ venv:
 
 espnet.done: venv
 	. venv/bin/activate; pip install pip --upgrade
+	. venv/bin/activate; pip install numpy
```
sw005320: I think it's better to ask people developing nara_wpe to include numpy installation.
It's OK for now, but please add comments to remove this line in future after nara_wpe includes the numpy installation.

kan-bayashi: added comment.

PR #542, opened 2019-01-13
Fix memory issue and optimize the speed for feat-to-shape.py

kamo-naoyuki:
Minor fixes for #493

- Changed `dataset[...]` -> `dataset[()]`: The former may cause memory issue. 
   - FYI: `dataset.value` is deprecated now and `dataset[()]` is recommended.
   - I don't understand why these two are different and the former causes memory issue, I doubt this is a bug of h5py. h5py is not documented about this yet.
- Implemented `return_shape` option for `FileReaderWrapper`. 
  - `feats-to-shape.py` requires the shape information only, so the whole data loading for the dataset is not efficient for speed. 
  - It doesn't make sense for `kaldi-ark` because `kaldiio` doesn't support loading the header only.
- Changed not to use `collections.Sequence` if PY3, which will be removed in `python=3.8`

### DIFF
--- a/espnet/transform/transformation.py
+++ b/espnet/transform/transformation.py
@@ -1,12 +1,20 @@
 from collections import OrderedDict
-from collections import Sequence
 import contextlib
 import copy
 import importlib
 import io
 import json
 import logging
+import sys
 import threading
+
+PY2 = sys.version_info[0] == 2
+
+if PY2:
+    from collections import Sequence
+else:
+    # The ABCs from 'collections' will stop working in 3.8
+    from collections.abc import Sequence
 
 
 def dynamic_import(import_path):
--- a/espnet/utils/cli_utils.py
+++ b/espnet/utils/cli_utils.py
@@ -30,7 +30,7 @@ def get_commandline_args():
     return sys.executable + ' ' + ' '.join(argv)
 
 
-def is_scipy_wav_stype(value):
+def is_scipy_wav_style(value):
     # If Tuple[int, numpy.ndarray] or not
     return (isinstance(value, Sequence) and len(value) == 2 and
             isinstance(value[0], int) and
@@ -38,7 +38,7 @@ def is_scipy_wav_stype(value):
 
 
 def assert_scipy_wav_style(value):
-    assert is_scipy_wav_stype(value), \
+    assert is_scipy_wav_style(value), \
         'Must be Tuple[int, numpy.ndarray], but got {}'.format(
             type(value) if not isinstance(value, Sequence)
             else '{}[{}]'.format(type(value),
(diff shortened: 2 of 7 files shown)


sw005320:
Thanks, but I still have a memory issue when executing `feat-to-shape.py` from your fork branch. I just killed the process it consumed ~10GB.
image

kamo-naoyuki:
I tested at my home and office,but I couldn't reproduce it. Can you tell me the situation more detail?

kamo-naoyuki:
I have tested all pairs of python2 or python3, and kaldi-ark or HDF5 with 30GB test data, but I didn't observe any memory leak. I checked it using both `top` and [tracemalloc](https://docs.python.org/ja/3/library/tracemalloc.html) 
Furthermore, for my environment, librispeech also works well with giving `run.pl --mem 5GB` to `data2json.sh`. 

Your code is really updated?

sw005320:
Thanks. I'm testing it by reviewing my setup.

kamo-naoyuki:
Thanks and sorry, this is my fault in `kaldiio`. It happens only when giving a large `scp` to `kaldiio.ReadHelper`  with `Python2`:  https://github.com/nttcslab-sp/kaldiio/pull/14

It was fixed with `kaldiio==2.12.1`, please test it:

```bash
pip install -U git+https://github.com/nttcslab-sp/kaldiio
```

By the way, I changed `dump.sh` and `data2json.sh` to use `copy-feats.py` and `feat-to-shape.py` respectively for HDF5 supporting, but it must be slower than the kaldi's commands potentially because Python is very very slow (But, I think kaldiio is faster than karel's).

kamo-naoyuki:
I found that `kaldiio` is much slower with loading `scp` than `ark` due to random-accessing and, as far `ark`, the performance is tolerable. Indeed, kaldi may use sequential reading both for scp and ark in `feat-to-len `. `kaldiio` can also become faster as same way. I'll send PR to `kaldiio`

`HDF5` seems not to support sequential reading, I'm not sure.  I'm considering another archive format now, e.g. `lmdb`.

sw005320:
Thanks!
It is solved with the latest `kaldiio` update from https://github.com/nttcslab-sp/kaldiio/pull/14

sw005320:
> HDF5 seems not to support sequential reading, I'm not sure. I'm considering another archive format now, e.g. lmdb.

I think this is no problem for now.
This is not a serious bottleneck (with HDF5, it becomes faster), and also random access is quite convenient when we modify our training loop.

sw005320:
@kamo-naoyuki, should I also merge this PR?

kamo-naoyuki:
Yes, please. HDF5 also has memory problem and I fixed it.

PR #544, opened 2019-01-14
Stop to change conda-config

kamo-naoyuki:
I can't understand why this part we needs:

https://github.com/espnet/espnet/blob/c6b78d73eee35c2116d19d24a8ccbb03653c6749/tools/conda.mk#L21

The both is not recommended for normal usage.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -30,7 +30,9 @@ install:
     else
         cd tools;
         make -f conda.mk PYTHON_VERSION=${ESPNET_PYTHON_VERSION} venv;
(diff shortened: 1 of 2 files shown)


ShigekiKarita:
This setup is for travis

kamo-naoyuki:
What do you mean?  If for travis,  then you should set for travis only. This setting is very annoying.

ShigekiKarita:
If the travis ran well and tests are passed, I do not mind. It was the official snippet for travis https://conda.io/docs/user-guide/tasks/use-conda-with-travis-ci.html

PR #546, opened 2019-01-15
fisher_callhome_spanish recipe: speed perturbation and small changes (updates after starting over from v.0.3.0)

JaejinCho:
Better to close this https://github.com/espnet/espnet/pull/491 and merge this instead


sw005320:
Thanks!

PR #547, opened 2019-01-16
Change 'backend' -> 'cmd_backend' 

kamo-naoyuki:
Follow up #538 

`backend` is reserved by `run.sh`, but `cmd.sh' overrides it now, so I changed  `backend` -> `cmd_backend`.


sw005320:
This is a good point (I also thought it when I reviewed it).

PR #551, opened 2019-01-17
Custom encoder layers for ASR

gtache:
Allows creating pyramidal encoder for example, with custom projection and dropout for each layer.     
Only works with BLSTMP (or VGGBLSTMP), and tts only has BLSTM      
I'll think of something similar for the LM and the decoder.


sw005320:
Some discussions:
- As you mentioned, we could consider this way for the decoder network and LM as well, but for this PR, we should focus on the encoder.
- it would be great if we include the projection layer configuration, but I'm not sure what is the best to describe it. `3x(1024,1024)`? `(1024,512), (512, 512), (1024, 1024)`. Hmm.
- we may unify the BLSTM or VGG-BLSTM with the same manner, and it only accepts `3x1024`.

gtache:
1) I agree     
2) Not sure either. `1024_512,512_512,3x512_256`...       
3) What I'm doing at the moment is if the `etype` is not projected (BLSTM or VGGBLSTM) and the `elayers` are not all the same, I use the projected equivalent and log a warning. We could also simply exit with an error.

sw005320:
> 2) Not sure either. `1024_512,512_512,3x512_256`...       

Does anyone have ideas?
@kan-bayashi, @ShigekiKarita, @kamo-naoyuki, do you have some ideas?

> 3) What I'm doing at the moment is if the `etype` is not projected (BLSTM or VGGBLSTM) and the `elayers` are not all the same, I use the projected equivalent and log a warning. We could also simply exit with an error.

Oh great. This is what I meant to say. Sorry that I did not notice that you already implemented it.

sw005320:
> 2) Not sure either. `1024_512,512_512,3x512_256`...       

We can go with this format.
Could you do it?
I think this extension makes the encoder BLSTMP architecture completely configurable.

gtache:
Okay, you said "completely configurable" so I also added per-layer and per-projection dropout configuration too. I still need to fix some things it seems.

gtache:
Oh right, I forgot about that. I will add an explicit dropout then.

gtache:
I changed the dropout separator to `-`, because it would cause problems with the expdir name.

sw005320:
@gtache, I just want to make it pending for a while as #575 could make this work more efficient. I hope you understand it.

gtache:
No problem, it makes sense to migrate to the new arguments handling system before making modifications.

sw005320:
@gtache, can you revisit this with our new config #722?
I think now we can easily reflect your work.

sw005320:
@gtache, can you come back? :)
Again, with our new config, this PR becomes more attractive.

PR #555, opened 2019-01-21
[WIP] PyTorch Transformer

ShigekiKarita:
this is the first result of #465


ShigekiKarita:
I'm rerunning the script `run.sh` again. see you tomorrow.

sw005320:
Great!

I just changed the target of the commit from the master to v.0.3.0.
Please follow the v.0.3.0 stye.
For the main source code, there are several refactoring.
Also, the style of `cmd.sh`, `path.sh`, `run.sh` are changed, and please check the other recipe in the v.0.3.0

kobenaxie:
Does Transformer support CTC-ATT in decoding ?

ShigekiKarita:
no. even beam search decoding is buggy.

espnet/bin/asr_train.py
```diff
@@ -164,6 +169,10 @@ def main(args):
                         help='Gradient norm threshold to clip')
     parser.add_argument('--num-save-attention', default=3, type=int,
                         help='Number of samples of attention to be saved')
+
+    from espnet.nets.pytorch_backend.e2e_transformer import add_transformer_arguments
```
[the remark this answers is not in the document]
ShigekiKarita: OK I think it is possible like this way
```python
def main(cmd_args):
  args = parse.parse_args(cmd_args)
  if args.ntype == "transformer":
    args = add_transformer_arguments(parser).parse_args(cmd_args)
```

espnet/nets/pytorch_backend/e2e_transformer.py
```diff
@@ -0,0 +1,707 @@
...
+    def __init__(self, dim, dropout):
+        super(Conv2dSubsampling, self).__init__()
+        self.conv = torch.nn.Sequential(
+            torch.nn.Conv2d(1, dim, 3, 2),
+            torch.nn.ReLU(),
+            torch.nn.Conv2d(dim, dim, 3, 2),
+            torch.nn.ReLU()
+        )
+        self.out = torch.nn.Sequential(
+            torch.nn.Linear(dim * (83 // 4), dim),
```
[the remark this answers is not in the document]
ShigekiKarita: Thanks for pointing out. It should be `idim=83` in `E2E(idim, odim, args)`. I'll fix that.

espnet/bin/asr_train.py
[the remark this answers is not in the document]
ShigekiKarita: I was wrong. Argument parser does not accept unregistred options like `--transformer-init`. Maybe we can avoid name conflictions by adding prefix `--transformer-xxx` for something not needed in the default RNN model.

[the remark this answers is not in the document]
ShigekiKarita: Or we can repeat every options needed in the model like fairseq. I think it can be more robust to unexpected behaviours.
- LSTM modfel https://github.com/pytorch/fairseq/blob/d259ffa98d90fb3100ffdab70c4db65c06b6b552/fairseq/models/lstm.py#L26
- Transformer model https://github.com/pytorch/fairseq/blob/d259ffa98d90fb3100ffdab70c4db65c06b6b552/fairseq/models/transformer.py#L50

ShigekiKarita:
After seeing fairseq library, I want to provide new option like `--model-module espnet.nets.pytorch_backend.e2e_transformer` or `--model-module userlib.user_model` instead of `args.ntype`. Those modules should contain the used-defined model definition class `E2E` and optionally its arguments `def add_arguments(parser):`. This is because `args.ntype` is closed and not extensible. How do you think?

egs/wsj/asr2/run.sh
```diff
@@ -0,0 +1,365 @@
...
+set -e
+set -u
+set -o pipefail
+
+train_set=train_si284
+train_dev=test_dev93
+train_test=test_eval92
+recog_set="test_dev93 test_eval92"
+
+if [ ${stage} -le 0 ]; then
```
sw005320: Can you use stop_stage implemented by #539?

espnet/bin/asr_train.py
```diff
@@ -45,7 +45,9 @@ def main(args):
                         help='Filename of train label data (json)')
     parser.add_argument('--valid-json', type=str, default=None,
                         help='Filename of validation label data (json)')
-    # network architecture
+    # network archtecture
```
sw005320: fix typo

espnet/nets/pytorch_backend/e2e_transformer.py
```diff
@@ -0,0 +1,734 @@
...
+    >>> subsequent_mask(3)
+    [[1, 0, 0],
+     [1, 1, 0],
+     [1, 1, 1]]
+    """
+    ret = torch.ones(size, size, device=device, dtype=dtype)
+    return torch.tril(ret, out=ret)
+
+
+class MultiHeadedAttention(nn.Module):
```
sw005320: One option is to put it in https://github.com/espnet/espnet/blob/v.0.3.0/espnet/nets/pytorch_backend/attentions.py?
You ma also reuse Tomoki's multihead attention.
Probably we need some discussions of leaving this style or split encoder/decoder/attention like RNN.
@gtache, do you have some comments?

espnet/nets/pytorch_backend/e2e_transformer.py
```diff
@@ -0,0 +1,734 @@
...
+        :return tgt_mask: score mask before softmax (batch, maxlen_out)
+        :rtype: torch.Tensor
+        """
+        x = self.embed(tgt)
+        x, tgt_mask, memory, memory_mask = self.decoders(x, tgt_mask, memory, memory_mask)
+        x = self.output_layer(self.output_norm(x))
+        return x, tgt_mask
+
+
+class LabelSmoothing(nn.Module):
```
sw005320: Is there any difference between @creatorscan and this label smoothing?
It would be better to unify it in future.
@creatorscan, could you also check this?

setup.py
```diff
@@ -18,7 +18,8 @@ def get_all_scripts(dirname):
         'scipy',
         # Installation from anaconda is recommended for PyTorch
         # 'torch==0.4.1',
-        'chainer==5.0.0',
+        #  need to match to Makefile cupy==4.3.0
```
sw005320: Probably I misunderstood something, but does this implementation depend on cupy? (it should not be, right?)

espnet/nets/pytorch_backend/e2e_transformer.py
kan-bayashi: I think Shigeki’s implementation is better than mine.
We should use Shigeki’s one.

ShigekiKarita:
> This is a 1st review.
> I'll have 2nd review later.
> My major discussion item is whether we should split espnet/nets/pytorch_backend/e2e_transformer.py into attention/encoder/decoder or not.
> If it looks better, we should do it, but I don't mind to keep this style if it makes it complicated.

Yeah it is problem in this PR. At least, we can split transformer layers into `espnet.nets.pytorch_backend.transformer_layers.{attention, encoder, decoder, common}` instead of existing modules to follow the previous style. I think those tiny pieces are good for tracking git log and coverage measurement on unittest.

espnet/nets/pytorch_backend/e2e_transformer.py
ShigekiKarita: This LS implementation is originally based on this paper https://arxiv.org/pdf/1512.00567v1.pdf as described in transformer paper and opennmt-py.

There are two differences from the one in ESPnet.
1. it masks padded tokens (it is not blank=0 but ignore_id=-1 in the target tensor)
2. it is uniform smoothing (1.0 - w or w) instead of unigram histgram

setup.py
ShigekiKarita: You are right. This is for unittest. I found chainer 5.0.0 does not accept cupy 4.3.0 arrays in the unittest. I can revert this change because I do not matter if the unrelated unittest fails.

espnet/nets/pytorch_backend/e2e_transformer.py
ShigekiKarita: In my opinion, it is non sense to unify this because its mechanism is completely different (but its motivation is same). Maybe I will change the name of argparse option instead of reusing it. It should be able to use at the same time e.g., `--unigram_lsm 0.1 --uniform_lsm 0.1`.

espnet/nets/pytorch_backend/e2e_transformer.py
gtache: I'd advise splitting as much as possible when it is not already merged, because if you decide to modularize later it will probably cause a lot of conflicts.    
The file is already 700+ lines long, and contains helper methods for the arguments and the transformer, classes for the encoder, decoder, attention, etc, so if they are not compatible with the RNN model, I'd say to put them in different files in `espnet.asr.pytorch_backend.transformer`.

espnet/nets/pytorch_backend/e2e_transformer.py
sw005320: I see. Year, it's better to avoid to reuse the option then.

sw005320:
> Yeah it is problem in this PR. At least, we can split transformer layers into espnet.nets.pytorch_backend.transformer_layers.{attention, encoder, decoder, common} instead of existing modules to follow the previous style. I think those tiny pieces are good for tracking git log and coverage measurement on unittest.

This is fine, but is it possible to reuse RNN's attention/encoder/decoder module? One of use cases is to combine the VGGBLSTM encoder and transformer attention/decoder by considering the nature of speech recognition (self attention may be a bit redundant model for the speech-input encoder).

sw005320:
@ShigekiKarita could you accelerate this?
We could do some modularization later.
I want to include it to v.0.3.0.

ShigekiKarita:
OK

kan-bayashi:
Welcome home!

ShigekiKarita:
This PR will be closed due to renew PR #690 @kan-bayashi  Thanks.

PR #566, opened 2019-01-30
ColdFusion support

gtache:
Adds ColdFusion for ASR    
Needs some testing, but Chainer doesn't allow setting requires_grad apparently, which is a problem (we want to train the network but not the RNNLM). For now I set creator/creator_node to `None`, which (from what I understood) should stop the computation of the backprop.


sw005320:
@JaejinCho and @hirofumi0810, could you please review this PR?

hirofumi0810:
@gtache How was the result? I cannot accept this PR unless you show results.

sw005320:
> @gtache How was the result? I cannot accept this PR unless you show results.

@gtache don't have enough GPUs for now. I'm thinking of how to verify it experimentally. For now, it would be great if you review the code part only.

Issue #558, opened 2019-01-22
Python3 incompatible codes in egs/*/local


kamo-naoyuki:
```
./egs/voxforge/asr1/local/filter_text.py:22:13: H233  Python 3.x incompatible use of print operator
./egs/voxforge/asr1/local/make_trans.py:20:5: H233  Python 3.x incompatible use of print operator
./egs/voxforge/asr1/local/make_trans.py:49:5: H233  Python 3.x incompatible use of print operator
./egs/csj/asr1/local/csj_rm_tag.py:25:9: H233  Python 3.x incompatible use of print operator
./egs/csj/asr1/local/csj_rm_tag.py:26:9: H233  Python 3.x incompatible use of print operator
 ```

are still incompatible with python3. 

I can change all of them 2/3 compatible, but this is related to the development policy.

How do we write any policy in Readme.md or plan to throw away python2, as [numpy](https://docs.scipy.org/doc/numpy-1.14.1/neps/dropping-python2.7-proposal.html) or [pandas](https://pandas.pydata.org/pandas-docs/stable/install.html)?

(In my opinion)
we have to support python3, at least, in all python scripts because python2 will be deprecated language in 2020, now is 2019, and I even think now is already the time to throw away python2 for espnet. It takes too large cost to make compatible with python2/3 and we can't use any features of python3 forever if keeping to support python2.

sw005320:
Thanks for the suggestion.
I'm still hesitating because these are simply copied from Kaldi. If there are some update in the Kaldi, we may have some conflicts due to our own changes. Also, we may need more work if we also try to maintain these local scripts, since local scripts are highly depending on their tasks and corpora, and these coding styles are very different. But maybe i'm wrong
Sorry for the negative comments.

@kan-bayashi, @ShigekiKarita, I appreciate your inputs.

sw005320:
I agree with this idea of gradually changing it.
1st season change is a minimal change, but greatly contributes to avoid confusions.
Could you work on it then?

kamo-naoyuki:
O.K, I'll work for "1st season" anyway. As far 2nd, I'll discussed at that time, perhaps it is not necessary for local/*.py. 

(Is "1st season" strange in English?  1st step?)

So, should I change the shebang of all local/*.py?
I think some scripts are python2/3 compatible with confidence, for example,` ljspeech/local/**.py` may be compatible, but I don't know which one is compatible or not.

kamo-naoyuki:
[...]

PR #567, opened 2019-01-31
Fix egs/*/*/local/*.py for py2/3 compatibility and use Python3.7 by default

kamo-naoyuki:
#558 

- **Change the default to Python3.7 in Makefile** .
- Copied lcal/*.py from kaldi
```bash
cp ./tools/kaldi/egs/an4/s5/local/data_prep.py egs/an4/asr1/local/data_prep.py
cp ./tools/kaldi/egs/an4/s5/local/data_prep.py egs/an4/tts1/local/data_prep.py
cp ./tools/kaldi/egs/chime5/s5/local/json2text.py egs/chime5/asr1/local/json2text.py
cp ./tools/kaldi/egs/hkust/s5/local/hkust_segment.py egs/hkust/asr1/local/hkust_segment.py
cp ./tools/kaldi/egs/reverb/s5/local/run_wpe.py egs/reverb/asr1/local/run_wpe.py
cp ./tools/kaldi/egs/tedlium/s5/local/join_suffix.py egs/tedlium/asr1/local/join_suffix.py
cp ./tools/kaldi/egs/voxforge/s5/local/make_trans.py egs/voxforge/asr1/local/make_trans.py
```

These are automatically copied if some differences are found from kaldi, so some are not necessary to be changed.

- Modified

```
	modified:   egs/csj/asr1/local/csj_rm_tag.py
	modified:   egs/fisher_callhome_spanish_st/st1/local/callhome_make_spk2gender.py
	modified:   egs/fisher_callhome_spanish_st/st1/local/fsp_make_spk2gender.py
	modified:   egs/reverb/asr1/local/filterjson.py
	modified:   egs/voxforge/asr1/local/filter_text.py
````

@sw005320 
`callhome_make_spk2gender.py` and `fsp_make_spk2gender.py` are named as 
`callhome_make_spk2gender.sh` and `fsp_make_spk2gender.sh` in original kaldi repo, so maybe they didn't notice that these are not compatible with python3.
For the time being, I fixed it, but it is better to be fixed in kaldi.

The other scripts are originated from espnet, right? So I modified it for py2/3 compatibility.

### CODE REVIEW

**Comment by sw005320** [comment]
We had several discussions about miniconda (https://github.com/espnet/espnet/pull/125) and we decided to keep both installations, although I don't have a strong preference. Also I saw that you changed the default python version to 3.7. I'm fine with that.
@kan-bayashi, do you have a comment (python 2 vs. 3 and conda vs. system for default installation)?

**Comment by kan-bayashi** [comment]
Personally I do not like conda, but I agree with @kamo-naoyuki’s idea. 
Let’s change default setting.

**Comment by sw005320** [comment]
OK. Please modify README.md then (we can change the order system python and conda instructions of the Step 2-A).

File: README.md [hunk 1]
```diff
@@ -64,39 +64,45 @@ export CUDA_PATH=$CUDAROOT
...
 ```
 
-### Step 2-C) installation with specified python
-
-Install Kaldi, Python libraries and other required tools using specified python and virtualenv
+Or install specific Python version with miniconda
 ```sh
 $ cd tools
-$ make -j PYTHON=/path/to/python2.7
+$ make KALDI=/path/to/kaldi PYTHON_VERSION=3.6
```
sw005320: We should not put a Kaldi path for this instruction.

sw005320:
One of the benefits of using `conda` is that we can use MKL. I'm not sure how much, but we could expect to have some speed benefits during decoding with CPU.

kamo-naoyuki:
The main reason why I use anaconda is it can install all dependencies for pytorch, including cudnn, 
MKL not depending the system environment, while PIP can't handle c/c++ dependencies well.  Anaconda can potentially avoid the installation troubles. This is also one of the reason I selected pytorch instead of chainer.

Sometimes kaldi's recipes require specific tools, e.g. flac, sox, and they can be also installed without administrator privilege. Even Kaldi itself can be installed by anaconda: https://anaconda.org/conda-forge/kaldi

The cons of anaconda is the duplication of the managing system  i.e pip and anaconda. It sometimes confuse me.
And it takes more storage memories than virtualenv.


BTW,  chainer also provides pre-build packages.

```sh
pip install cupy-cuda80
```

I'm wondering about including it in Makefile to make the installation faster, but I'm not sure how to write it.
How about this?

```
CUDA :=  # 80, 90, 91, 92, 100

# If $CUDA is empty
pip install cupy==5.0.0
# if $CUDA is specified
pip install cupy-cuda$(CUDA)==5.0.0
```

This is not related to this pull request, so I'll send in another if you need.

egs/voxforge/asr1/local/make_trans.py
```diff
@@ -13,38 +12,40 @@
...
-
+unnorm_utt = set() 
 for l in file(sys.argv[1]):
     u, trans = l.split(None, 1)
     u = u.strip().split('/')[-1]
-    # convert to upper case
-    trans = unicode(trans, 'utf_8').strip().replace("-", " ").upper()
+    trans = trans.strip().replace("-", " ")
     if not trans.isupper() or \
-       not trans.strip().replace(' ', '').replace("'", "").isalnum():
```
sw005320: Is it OK for this change? (isalnum() vs. isalpha())

kamo-naoyuki: This file is just copied from kaldi. I'm not sure.

sw005320:
Can you keep isalnum? I couldn't remember but I think I did this change intentionally.

kamo-naoyuki:
`isalpha()` may be a bug. It sends an exception if any number is included in transcription.
I'm not sure the original intention, but at least for espnet, isalnum() seems good, so I changed.

PR #563, opened 2019-01-29
[WIP] Neural Machine Translation

hirofumi0810:
Pure NMT module.

PR #570, opened 2019-02-01
merge fisher_callhome_spanish and fisher_callhome_spanish_st directories

sw005320:


sw005320:
@hirofumi0810, I want to unify the recipe directory name. Please let me know this would have some conflict with your current work of #563.

hirofumi0810:
@sw005320 This PR requires feature re-dumping because feature paths included in json files would be changed. So it's better to merger after #563.

sw005320:
ok, but if you are talking about your working files, you can just redo only the feature dumping part and replace the path.

sw005320:
I'll do it after #563 is finished.

PR #576, opened 2019-02-04
Incremental Scheduled Sampling

gtache:
Format :     
`initialss_incss_maxss_iternum`     
So start at `initialss`, increase by `incss` every `iternum` until a maximum of `maxss`    
Or simply `ss` for a static value (doesn't change current format)

I'm wondering whether to modify the evaluation to use the current ss value or keep it at 0.


sw005320:
@creatorscan, could you review this PR?

creatorscan:
@gtache Thanks for your contribution and this PR does deal with linear decay which is an important addition.
That being said, analysing the performance of this approach wrt ESPnet (atleast in AN4 database) is important and it would be great if you update the RESULTS file with Incremental SS numbers.

PR #577, opened 2019-02-06
Fixes earlystop with acc

gtache:
I thought that the `auto` mode for earlystop checked for `acc`, but it checks for `accuracy`, so we have to set it manually. The default early stop criterion is also changed to acc


gtache:
Okay I changed back to loss for TTS and LM.     
Well the reason I changed to 'epochs' instead of 'epoch' was to have a consistent name for all the modules (and therefore a cleaner way to implement methods related to all of them), but I guess it will be fixed with the new argument parsing system anyway.

sw005320:
Thanks!

PR #578, opened 2019-02-06
fix bugs for #567

sw005320:
fix bugs in #567


sw005320:
@kamo-naoyuki, I just observed small degradation for the voxforge Italian task, but it would be cased by #577, and it would be fixed soon. Others seem to work for #567.

kamo-naoyuki:
Sorry, there is no need of abspath. I put it without thinking. Please delete it.

Issue #533, opened 2019-01-10
mergejson.py out of memory


lilian0118:
I have 10,000h 16K audio data, when load all *.json in mergejson.py, it will out of memory. My computer only have 128GB memory, can we split big jsons to small piece? and run every small piece one by one?

kamo-naoyuki:
For my reference, can you tell me the number of lines of your feats.scp file?

lilian0118:
> For my reference, can you tell me the number of lines of your feats.scp file?

2019-01-10 11:36:41,138 (mergejson:41) INFO: data/train_nodev/tmp-FO1kr/feat.json: has 43873228 utterances
2019-01-10 11:41:53,678 (mergejson:41) INFO: data/train_nodev/tmp-FO1kr/idim.json: has 43873228 utterances
2019-01-10 11:49:17,102 (mergejson:41) INFO: data/train_nodev/tmp-FO1kr/ilen.json: has 43873228 utterances
2019-01-10 11:56:36,597 (mergejson:41) INFO: data/train_nodev/tmp-FO1kr/odim.json: has 43873228 utterances
2019-01-10 12:09:55,194 (mergejson:41) INFO: data/train_nodev/tmp-FO1kr/olen.json: has 43873228 utterances
2019-01-10 12:20:15,494 (mergejson:41) INFO: data/train_nodev/tmp-FO1kr/text.json: has 43873228 utterances
Killed

kan-bayashi:
I think this problem is the same as #449.
Could you try to use `--output-json` option for `mergejson.py`?

kamo-naoyuki:
@kan-bayashi This problem is not related to #449. That memory error happens when writing JSON file, but @lilian0118 met it when loading time.

sw005320:
@lilian0118, thanks for the report.
I'm curious why the number of utterances is so large (43,873,228).
For example, in the Librispeech cases (~1000 hours), the number of utterances is 281,241. If you use the spontaneous speech, the number of utterances tend to be larger than the read speech one, but I feel that yours is still too large.
There might be some problems in the data preparation.

But, regardless of this, I'm expecting this happens, as our JSON file is quite redundant and consumes a lot of memory (but I decided this format, as it is easy to debug).
We may need to compress it (we have a lot of redundant field), using binary JSON format, or integrate it with HDF5.

lilian0118:
@sw005320 , thanks for you kind reply.
[...]

PR #579, opened 2019-02-09
[Need to discuss] New style of mergejson.py

kamo-naoyuki:
The current flow of `data2json.sh`:

1. Creates text files, e.g. feats.scp, text, ...
1. Change each files to JSON files using `scp2json.py`
1. Merge JSON files into one

The second step looks redundant. So I modified it to

1. Creates text files, e.g. feats.scp, text, ...
1. Merge text files into a JSON file

For this, I created a new util: `merge_scp2json.py`

```sh
merge_scp2json.py ¥
  --input-scps feat:feats.scp shape:shape.scp:shape ¥
  --input-scps feat:feats2.scp shape:shape2.scp:shape ¥
  --output-scps text:text ¥
  --scps utt2spk 
```

The format of each arguments is `<key>:<filepath>:<type>`, where `<type>` indicated the data type registered in the output json, e.g. :

```sh
% cat foo.txt
a 3
b 10
% merge_scp2json.py --input-scps bar:foo.txt:int
{
    "utts": {
        "a": {
            "input": [
                {
                    "bar": 3,
                    "name": "input1"
                }
            ],
            "output": []
        },
        "b": {
            "input": [
                {
                    "bar": 10,
                    "name": "input1"
                }
            ],
            "output": []
        }
    }
}
```

### DIFF
--- a/utils/merge_scp2json.py
+++ b/utils/merge_scp2json.py
@@ -58,8 +58,7 @@ if __name__ == '__main__':
                         help='The json files except for the input and outputs')
     parser.add_argument('--verbose', '-V', default=1, type=int,
                         help='Verbose option')
-    parser.add_argument('--out', '-O', type=argparse.FileType('w'),
-                        default=sys.stdout,
+    parser.add_argument('--out', '-O', type=str,
                         help='The output filename. '
                              'If omitted, then output to sys.stdout')
 
@@ -139,7 +138,11 @@ if __name__ == '__main__':
     #
     # To reduce memory usage, reading the input text files for each lines
     # and writing JSON elements per samples.
-    args.out.write('{\n    "utts": {\n')
+    if args.out is None:
+        out = sys.stdout
+    else:
+        out = open(args.out, 'w', encoding='utf-8')
+    out.write('{\n    "utts": {\n')
     nutt = 0
     while True:
         nutt += 1
@@ -180,10 +183,10 @@ if __name__ == '__main__':
         # The end of file
         if first == '':
             if nutt != 1:
-                args.out.write('\n')
+                out.write('\n')
             break
         if nutt != 1:
-            args.out.write(',\n')
+            out.write(',\n')
 
         entry = {}
         for inout, _lines, _infos in [('input', input_lines, input_infos),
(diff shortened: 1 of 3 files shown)


kamo-naoyuki:
Note that `shape ` of `shape:shape.txt:shape` is a special type:

```python
>>> shape('3,4')
[3, 4]
```

The benefit of this style, `text` files to a JSON files, is memory usage. 
JSON format cannot be load per elements, so we need to read JSON text fully at the first time, and it takes much memory if there are large samples (#533).
I designed `merge_scp2json.py` as reading each text files line by line and write each elements one by one.

sw005320:
> 2. Change each files to JSON files using `scp2json.py`

I like your idea and please proceed with it. The reason I made such redundant process is that I was thinking of handling "3. Merge JSON files into one" with some command line JSON tools, e.g., `jq`. But it turns out that it cannot handle a complicated process. Also such command line tools are not so popular and writing our own merge scripts is easier. Now we also have a memory issue. We do not have to stick to my wrong design.

kamo-naoyuki:
mergejson.py creates the intersection among the input json files at the first, but my tool assumes the keys of each input files are same and sorted and if some keys are different, the processing will stop in the middle with an error message. This is inevitable to read text files line by line.

So, we need to format text files in the preprocessing. e.g. `fix_data_dir.sh`
Is it o.k?

sw005320:
> So, we need to format text files in the preprocessing. e.g. `fix_data_dir.sh`
Is it o.k?

Sure.

utils/merge_scp2json.py
```diff
@@ -0,0 +1,248 @@
...
+import codecs
+from io import open
+import json
+import logging
+import sys
+
+from espnet.utils.cli_utils import get_commandline_args
+
+PY2 = sys.version_info[0] == 2
+sys.stdin = codecs.getwriter('utf-8')(sys.stdin if PY2 else sys.stdin.buffer)
```
gtache: It should be codecs.getreader for stdin

kamo-naoyuki: @gtache Thanks, I fixed.

kamo-naoyuki:
Note that this modification solves the memory issue for data2json.sh, but originally JSON causes memory issue if the items are so many because it's not supporting lazy loading.

HDF5 is a better choice as far as memory usage, but probably it degrades the readability because it's a binary format, simply. I don't have good idea.

sw005320:
I guess it affects to the WSJ 2mix and speech translation recipes (https://github.com/espnet/espnet/search?q=mergejson&unscoped_q=mergejson). @hirofumi0810 and @simpleoier, could you comment this PR? Could you manage to reflect it to your local mergejson?

simpleoier:
I'm good with this. I've updated my part to use it.

PR #580, opened 2019-02-11
Variable batchsize

gtache:
Basically, instead of saying batchsize=16 (fixed number of samples per batch, except if maxlen_in or out is too big), we specify something like max_batch_size=1000000, where the size of a sample depends on idim, odim, ilen and olen.     
I found that it gave a really good speedup in practice, but that may be only due to having a throttled overheating CPU on my first workstation. I'll test on a normal one as soon as I have a GPU available.


sw005320:
> max_batch_size=1000000, where the size of a sample depends on idim, odim, ilen and olen.

It's an interesting idea.
My question is that which is easier to tune (# of utterances or the size of a sample).
The size of a sample may be more accurate, but the number of utterances is more intuitive.
Also, we could at least keep both ways and switch them by an option. Is the current PR designed like that?

gtache:
Yes, what I'm doing right now is if batchsize>0, I use the default make_batchset, and if batchsize<=0, I use the variable make_batchset. We could have an explicit argument, or I can add a comment in the run.sh

I've also add a `test-batch-size` argument, if the user sets it to 50 for example, the minibatches returned will be the 20 first, 20 last, and one every 50 batches. It allows to easily find a value which fits on the GPU before the real training and avoid CUDA OoM errors.

sw005320:
> Yes, what I'm doing right now is if batchsize>0, I use the default make_batchset, and if batchsize<=0, I use the variable make_batchset. We could have an explicit argument, or I can add a comment in the run.sh

Great!
Adding the comment in `run.sh` should be fine.

ShigekiKarita:
@gtache @sw005320 Sorry for the late notice. I like this idea because some novel NN frameworks (e.g., [tensor2tensor](https://github.com/tensorflow/tensor2tensor) adopt this idea.

Do you have any problem to rework this into v.0.4.0? Let me know if I can help.

sw005320:
I'm pending it since @gtache could not test this in his environment. If you or some other people check the effectiveness, I'm happy to include it to our main branch.

ShigekiKarita:
OK, I will try `make_variable_batchset` in v.0.4.0 this weekend. I will not overwrite existing recipe with `max_batch_size` at this time.

BTW, I do not think the name `--max-batch-size` is intuitive. How about `-batch-frames`? Also, I wanna make `--batch-size` and `--batch-frames` exclusive as
```python
if args.batch_size is None ^ args.batch_frames is None:
    raise ValueError("cannot set --batch-size and --batch-frames at once")
```
because I cannot remember this behavior.
```bash
max_batch_size=1000000 # Enabled when batchsize=0, creates batches depending on the size of the samples
```

ShigekiKarita:
@kan-bayashi I think this kind of config might be essential for Transformer TTS. You know how they create the minibatch https://arxiv.org/pdf/1809.08895.pdf

sw005320:
> OK, I will try make_variable_batchset in v.0.4.0 this weekend. I will not overwrite an existing recipe with max_batch_size at this time.

Sure, please only perform it to a selected recipe.

> How about -batch-frames?

Fine

> Also, I wanna make --batch-size and --batch-frames exclusive as

Fine

ShigekiKarita:
@r9y9 @kan-bayashi I think this kind of config might be essential for Transformer TTS. You know how they create the minibatch https://arxiv.org/pdf/1809.08895.pdf

kan-bayashi:
This PR included in #759. I will close.

PR #581, opened 2019-02-12
[WIP] First step for online speech recognition

takenori-y:
For online demo, I added a script for speech recognition where the input is a single wav file.


b-flo:
Hi,

 I'm wondering if the "online" term should be used here as it could lead to some ambiguities for new user?  Quoting Kaldi's documentation : 

> By "online decoding" we mean decoding where the features are coming in in real time, and you don't want to wait until all the audio is captured before starting the online decoding. (We're not using the phrase "real-time decoding" because "real-time decoding" can also be used to mean decoding whose speed is not slower than real time, even if it is applied in batch mode).

sw005320:
Hi @b-flo, thanks for the clarification. Yes, we exactly want to use "online" for this meaning. But to do that we actually have to consider to extend the beam search algorithm with an online manner in addition to the frontend online processing (audio I/O, feature extraction, and VAD parts). This PR is intended to deal with the frontend online processing first, and consider the rest of the part later.

b-flo:
Thank you for the precision @sw005320 I didn't know you started working on the online processing/decoding! That's really exciting!

sw005320:
- [ ] it's better to accept LM and E2E ASR directories as an argument option
- [ ] no need for scoring
- [ ] no need for dict and nlsyms (you may need to change `data2json.sh` as well.

With that this script becomes more portal (we only specify the model directories and WAV files).

takenori-y:
Thank you for your valuable comments.
In the new script, a dummy dict is created to use `data2json.sh` without modification and a CMVN file can also be given by a user.

sw005320:
Thanks!
I have two additional comments:
1. I think it would be better if you add an existence check for E2E model, LM model, CMVN like you did for `--wav`.
2. decoding related files (`${decode_dir}/*`) would be overwritten when we don't correctly specify `--decode_dir`. We could avoid it by asking the user to correctly specify `--decode_dir` but we may need to deal with it. One simple idea is to get the WAV file name by `basename <WAV_FILE_NAME>.wav .wav` and creating relating files on `${decode_dir}/<WAV_FILE_NAME>/`. Then, we can at least avoid that these files are overwritten when we use different audio files.

sw005320:
If you rename it with `local/recog_wav.sh`. I'll merge it.

PR #583, opened 2019-02-13
Small change of the "Makefile" under "tools"

Xiaofei-Wang:
Today when I tried to build ESPnet from scratch, I found the file structure of "kaldi-io-for-python" has been changed (https://github.com/vesis84/kaldi-io-for-python). So I made a small modification of the Makefile under "tools".

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -51,7 +51,7 @@ install:
   - cd tools
   - git clone https://github.com/vesis84/kaldi-io-for-python.git
   - cd kaldi-io-for-python
-  - ln -s kaldi_io.py kaldi_io_py.py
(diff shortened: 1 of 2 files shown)


sw005320:
Thanks!
This seems to affect the module import (https://travis-ci.org/espnet/espnet/jobs/492463513).
Could you also fix this part?

Note that we will switch to another kaldiio developed at NTT (https://github.com/nttcslab-sp/kaldiio) in v.0.3.0 (see https://github.com/espnet/espnet/pull/493).

Xiaofei-Wang:
Got it! 
I fix the .travis.yml file. Now all checks have got passed: )

sw005320:
Thanks!

PR #584, opened 2019-02-13
Disable conda=false python=2.7 torch=1.0 for travis

gtache:
Like we discussed earlier, it fails by timeout, so we can't see if a commit is working without looking on travis.


sw005320:
thanks!

PR #587, opened 2019-02-19
[WIP] Exploit information from CTC

takenori-y:
The CER between true sentences and sentences estimated by CTC without beam search is written to a log file and a png file during training. The following example is a png generated by running the WSJ recipe.

cer


sw005320:
This is great!
The number also looks reasonable.

I'm thinking that we could use the error rate even for the encoder-decoder. Then, we could easily switch to monitor the **minimum** value of the validation loss, encoder-decoder validation error or CTC validation error for the epsilon decay or early stopping. @kan-bayashi, what do you think? @gtache, I came up with this idea because it simplifies your early stopping implementation (we don't have to switch to min or max).

gtache:
If it gives a better metric for final results and removes the need to check for acc/loss, it'd be great indeed.

PR #596, opened 2019-02-22
End to end multi channels system

kamo-naoyuki:
- Implemented MVDR and WPE using pytorch.
- Created a joint training recipe of DNN-mask based MVDR with encoder-decoder based on chime4  recipe.

Here is a brief description: https://github.com/kamo-naoyuki/espnet/blob/frontend/egs/chime4/asr1_multich/README.md

I'm not an expert as far signal processing, so these implementation should be checked by someone responsible.


egs/chime4/asr1_multich/run.sh
```diff
@@ -239,17 +250,18 @@ if [ ${stage} -le 3 ]; then
     fi
     # use only 1 gpu
     if [ ${ngpu} -gt 1 ]; then
-	echo "LM training does not support multi-gpu. single gpu will be used."
+    echo "LM training does not support multi-gpu. single gpu will be used."
```
sw005320: It's better put an indent.

egs/chime4/asr1_multich/run.sh
```diff
@@ -239,17 +250,18 @@ if [ ${stage} -le 3 ]; then
...
     fi
+
     ${cuda_cmd} --gpu ${ngpu} ${lmexpdir}/train.log \
```
sw005320: Indents of lines below are not consistent.

espnet/nets/pytorch_backend/frontends/beamformer.py
```diff
@@ -0,0 +1,158 @@
...
+    """
+    # outer product: (..., C_1, T) x (..., C_2, T) -> (..., T, C, C_2)
+    psd_Y = FC.einsum('...ct,...et->...tce', [xs, xs.conj()])
+
+    # Averaging mask along C: (..., C, T) -> (..., T)
+    mask = mask.mean(dim=-2)
+
+    # Normalized mask along T: (..., T)
+    if normalization:
+        mask = mask / mask.sum(dim=-1)[..., None]
```
sw005320: Does this consider the padding well?

kamo-naoyuki: If assuming  the tensor is padded with zero, the summation along the time axis is same regardless of the padding length. 
This might be a little confusing, so I'll put a comment there.

kamo-naoyuki:
> * with the enhanced audios, we could also calculate several enhancement measures (e.g., PESQ, STOI). 

I would rather not use any matlab codes as possible. Can I use the following?

SDR: https://github.com/sigsep/sigsep-mus-eval 
STOI, ESTOI: https://github.com/mpariente/pystoi
PESQ: Same as kaldi:  https://www.itu.int/rec/dologin_pub.asp?lang=e&id=T-REC-P.862-200102-I!!SOFT-ZST-E&type=items

`museval` is used for SISEC2018, so this is reliable.
`pystoi`'s test codes succeeded to compare https://github.com/JacobD10/SoundZone_Tools, so it may be o.k.

sw005320:
> > * with the enhanced audios, we could also calculate several enhancement measures (e.g., PESQ, STOI).
> 
> I would rather not use any matlab codes as possible. Can I use the following?
> 
> SDR: https://github.com/sigsep/sigsep-mus-eval
> STOI, ESTOI: https://github.com/mpariente/pystoi
> PESQ: Same as kaldi: https://www.itu.int/rec/dologin_pub.asp?lang=e&id=T-REC-P.862-200102-I!!SOFT-ZST-E&type=items
> 
> `museval` is used for SISEC2018, so this is reliable.
> `pystoi`'s test codes succeeded to compare https://github.com/JacobD10/SoundZone_Tools, so it may be o.k.

This has a pros and cons. (pros: can avoid matlab, cons: cannot strictly compare with other references using matlab-based) but for CHiME-4, this should be fine.

kamo-naoyuki:
@sw005320 I have a question about evaluation for the enhanced speech.
Kaldi uses the clean data which is generated by https://github.com/kaldi-asr/kaldi/blob/master/egs/chime4/s5_1ch/local/CHiME3_simulate_data_patched_parallel.m, which is originated from nn-gev, as reference file. 
I haven't read the script deeply yet. Should we use it? Or, can we use the booth data as reference?

kamo-naoyuki:
> Also, several people are exciting about this PR. It would be great if we can merge it soon.

I'm tackling to dump the enhanced speech now and I can finish on the next week. It may be good to merge at that time. 

> speed perturbation could get more improvement for this kinds of data. You may have 3-times more data, but required training epochs would be smaller, and the total training time would not be 3-times.

Speed perturbation should be also applied to the current chime4/asr1 to compare the performance. I wonder if someone try to do, then my work get easy.
Anyway, as far perturbation, I'll tackle at the next PR. I'm interested in the other perturbation, e.g. bandpass perturbation,

sw005320:
@sas91, could you review this PR? This seems to be getting a better shape.

sas91:
> @sas91, could you review this PR? This seems to be getting a better shape.

@sw005320 Sure I will review it.

kamo-naoyuki:
o.k.

kamo-naoyuki:
@sas91 That may be due to a bug of kaldiio and recently it was fixed. Try to update by pip. 
I changed to use sox because `gather-wav.py` is very slow. 
I'm curious you said it worked with 2ch. It may not work with old one.

Thanks.

sw005320:
Is it finished? (I don't fully check it since it still has [WIP])

kamo-naoyuki:
@sw005320 Finished for this PR.

I'm remaining several tasks and I'll commit one by one at the next PR if I have time.

- [x] Implement frontend modules : This PR
- [x] Create recipe using multi channel system: This PR
- [x] Implement a tool for dumping enhanced audio signal and evaluation tools for it: This PR
- [ ] Test codes for frontend
- [ ] 2ch track recipe of chime4 using multi channel system
- [ ] Speed perturbation recipe for chime4
- [ ] The other tasks: Reverb, Chime5, etc.

espnet/nets/pytorch_backend/frontends/feature_transform.py
```diff
@@ -0,0 +1,235 @@
...
+
+        _mel_options = dict(sr=fs,
+                            n_fft=n_fft,
+                            n_mels=n_mels,
+                            fmin=fmin,
+                            fmax=fmax)
+        self.mel_options = _mel_options
+
+        # melmat: (D2, D1) -> (D1, D2)
+        melmat = librosa.filters.mel(**_mel_options)
```
sw005320: Is this Mel matrix same as Kaldi's one?
I'm curious.

espnet/nets/pytorch_backend/e2e_asr.py
```diff
@@ -388,6 +454,67 @@ def subsample_frames(self, x):
         return h, ilen
 
 
+def _to_torch_tensor(x):
```
sw005320: This function may be moved to `nets_utils.py`.

tools/Makefile
```diff
@@ -88,6 +88,11 @@ chainer_ctc.done: espnet.done
 	. venv/bin/activate; cd chainer_ctc && pip install .
 	touch chainer_ctc.done
 
+pytorch_complex.done: venv
+	# Require python>=3.6, pytorch>=1.0
```
sw005320: What's happened if we use python 2.X or pytorch 0.4X?
Is the installation stopped? If so, please avoid it.

espnet/nets/pytorch_backend/frontends/feature_transform.py
kamo-naoyuki: No, it's not.
`htk=True, norm=None` may get the nearest one in librosa, but still different.

https://librosa.github.io/librosa/generated/librosa.core.mel_frequencies.html

Is it a big difference? 
Once I have implemented a fully kaldi-compatible fbank extractor in pure python, https://github.com/nttcslab-sp/kaldifeats, but I think librosa is better because it's well known.

sw005320: This is no problem, but I just want to make sure. I also have some comparisons between them and sometimes Kaldi features are better and sometimes not.
Can you add some comments that this will be different from the one generated from Kaldi?

kamo-naoyuki:
>     * How it is handled when the feature extraction is not well performed (due to very short-length audios, #659)? In the current recipe, we add `utils/fix_data_dir.sh` to remove such utterances after the feature extraction.

It can't be handled now. How about limiting the minimum in-length in `make_batchset()` or `data2json.sh`? 
`data2json.sh` may be good. I think it is helpful to see all files really used.

kamo-naoyuki:
>     * This is a little bit risky to always use `with torch.no_grad(), using_transform_config({'train': False})`. People may forget to add `using_transform_config({'train': False})`. Is there any other way?

I agree.

- I deleted `using_transform_config()` and the related functions because these system is too complex, indeed.
- Instead of it, give arguments to `__call__()` directly in `Transformation`.
- `LoadInputsAndTargets` can be set the configuration, `train=False`, when instantiation. 
- Also, I deleted `CustomConver.transform()` for refactoring.

tools/Makefile
kamo-naoyuki: The pip installation can also work with python2, but it's failed when importing.
This is a little confusing, so I modified `Makefile`.

`pytorch_complex` can work with  `pytorch0.4`  if not to use `inverse()`.

egs/chime4/asr1_multich/run.sh
```diff
@@ -336,14 +373,113 @@ if [ ${stage} -le 5 ]; then
...
+    done
+    i=0; for pid in "${pids[@]}"; do wait ${pid} || ((++i)); done
+    [ ${i} -gt 0 ] && echo "$0: ${i} background jobs are failed." && false
+
+    echo "Decoding successfully finished"
+fi
+
+if [ ${stage} -le 6 ] && [ ${stop_stage} -ge 6 ]; then
+    echo "stage 6: Enhance speech"
+    nj=200
```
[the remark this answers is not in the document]
kamo-naoyuki: Changed it to 32.

egs/chime4/asr1_multich/run.sh
```diff
@@ -303,13 +338,15 @@ if [ ${stage} -le 4 ]; then
         --patience ${patience}
 fi
 
-if [ ${stage} -le 5 ]; then
+if [ ${stage} -le 5 ] && [ ${stop_stage} -ge 5 ]; then
     echo "stage 5: Decoding"
-    nj=32
+    nj=200
```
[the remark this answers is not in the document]
kamo-naoyuki: Changed it to 32.

egs/chime4/asr1_multich/run.sh
```diff
@@ -239,17 +268,18 @@ if [ ${stage} -le 3 ]; then
...
-		--valid-label ${lmdatadir}/valid.txt \
+        lm_train.py \
+        --ngpu ${ngpu} \
+        --backend ${backend} \
+        --verbose 1 \
+        --outdir ${lmexpdir} \
+        --tensorboard-dir tensorboard/${lmexpname} \
+        --train-label ${lmdatadir}/train.txt \
+        --valid-label ${lmdatadir}/valid.txt \
                 --resume ${lm_resume} \
```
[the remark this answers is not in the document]
kamo-naoyuki: Sorry, I fixed.

egs/chime4/asr1_multich/RESULTS
```diff
@@ -0,0 +1,126 @@
...
+## Evaluation for ehnaced speech
+SDR, SIR, and SAR was evaluated by using museval, i.e. BSSEval V4, with bsseval_sources version and BSSEval V3 compatible mode.
+
+### Noisy speech
+#### SDR
+                              PED    CAF    STR    BUS
+dt05_simu_isolated_6ch_track: 8.468 6.848 4.449 2.961
+                              PED    CAF    STR    BUS
+et05_simu_isolated_6ch_track: 8.326 7.745 6.56 6.754
+#### SIR
```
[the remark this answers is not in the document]
kamo-naoyuki: Deleted SIR and SAR.

espnet/bin/asr_enhance.py
```diff
@@ -0,0 +1,122 @@
...
+    # check CUDA_VISIBLE_DEVICES
+    if args.ngpu > 0:
+        cvd = os.environ.get("CUDA_VISIBLE_DEVICES")
+        if cvd is None:
+            logging.warning("CUDA_VISIBLE_DEVICES is not set.")
+        elif args.ngpu != len(cvd.split(",")):
+            logging.error("#gpus is not matched with CUDA_VISIBLE_DEVICES.")
+            sys.exit(1)
+
+        # TODO(mn5k): support of multiple GPUs
```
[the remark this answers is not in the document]
kamo-naoyuki: Changed mn5k to kamo.

espnet/nets/pytorch_backend/e2e_asr.py
kamo-naoyuki: Moved.

espnet/asr/asr_utils.py
```diff
@@ -488,3 +538,62 @@ def add_results_to_json(js, nbest_hyps, char_list):
...
     return new_js
+
+
+def plot_spectrogram(plt, spec, mode='db', fs=None, frame_shift=None,
+                     bottom=True, left=True, right=True, top=False,
+                     labelbottom=True, labelleft=True, labelright=True,
+                     labeltop=False, cmap='inferno'):
+    """Plot spectrogram using matplotlib
+
+    :param plt:
```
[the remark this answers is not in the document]
kamo-naoyuki: Add some docstrings.

egs/chime4/asr1_multich/run.sh
```diff
@@ -336,14 +373,113 @@ if [ ${stage} -le 5 ]; then
...
+                --image-dir ${enhdir}/images \
+                --num-images 20 \
+                --fs ${fs}
+    ) &
+    pids+=($!) # store background pids
+    done
+    i=0; for pid in "${pids[@]}"; do wait ${pid} || ((++i)); done
+    [ ${i} -gt 0 ] && echo "$0: ${i} background jobs are failed." && false
+
+    for rtask in ${recog_set}; do
```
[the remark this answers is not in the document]
kamo-naoyuki: This part reduces scp files into one. I inserted a comment.

sw005320:
@sas91, please review this PR now.

espnet/nets/pytorch_backend/frontends/beamformer.py
```diff
@@ -0,0 +1,165 @@
...
+        https://ieeexplore.ieee.org/document/5089420
+
+    Args:
+        psd_s (ComplexTensor): (..., F, C, C)
+        psd_n (ComplexTensor): (..., F, C, C)
+        reference_vector (torch.Tensor): (..., C)
+    Returns:
+        beamform_vector (ComplexTensor)r: (..., F, C)
+    """
+    # numerator: (..., C_1, C_2) x (..., C_2, C_3) -> (..., C_1, C_3)
```
sas91: I have noticed inverse in pytorch becomes an issue if the magnitude of the entries in the input matrix are very large (may not be really singular) but it throws an error that it is singular. Noticed when using with REVERB data. How about a fix like this:
 
try:                                                                           
        psd_n_i = psd_n.inverse()                                                  
    except:                                                                        
        try:                                                                       
            reg_coeff_tensor = ComplexTensor(torch.rand_like(psd_n.real),          
                                             torch.rand_like(psd_n.real))*1e-2  
            psd_n = psd_n/10e+4                                                    
            psd_s = psd_s/10e+4                                                    
            psd_n += reg_coeff_tensor                                              
            psd_n_i = psd_n.inverse()                                              
        except:                                                                    
            reg_coeff_tensor = ComplexTensor(torch.rand_like(psd_n.real),          
                                             torch.rand_like(psd_n.real))*1e-1  
            psd_n = psd_n/10e+10                                                   
            psd_s = psd_s/10e+10                                                   
            psd_n += reg_coeff_tensor                                              
            psd_n_i = psd_n.inverse()

numerator = FC.einsum('...ec,...cd->...ed', [psd_n_i, psd_s])

egs/chime4/asr1_multich/run.sh
```diff
@@ -9,16 +9,22 @@
...
-do_delta=false
+# configuration path
+preprocess_conf=conf/preprocess.json
 
 # network architecture
+use_beamformer=true
+blayers=3
+bunits=300
+bprojs=300
+
```
sas91: how about adding an option to pass the reference channel. something like negative values for attention based ref or else channel id [0-5] for chime4?

Also I see other beamformer implementations like "mvdr2" and "gev" but maybe they can't be used while training because of the argmax in it but maybe the intention to include it was to use while inference as they have the same learnable parameters. How about an option like "bftype" to use them?

egs/chime4/asr1_multich/conf/preprocess.json
```diff
@@ -0,0 +1 @@
+{"process": [{"type": "stft", "win_length": 400, "n_fft": 512, "n_shift": 160, "window": "han"}]}
```
sas91: Can nfft also be "400" as time resolution is only "400", we don't get any additional freq. resolution by having nfft "512". Maybe "400" will use normal DFT while "512" might use FFT algorithm. Does "512" make it faster to compute features online?

kamo-naoyuki: This is the default parameter in Kaldi with fs == 16kHz.
FFT algo can be efficient calculation with the power of 2 number of fft, so Kaldi uses the nearest number to "400", and each window are zero padded so that it will be 512. This is a well known technique.

sas91: Yes FFT works only with a power of 2 but DFT can be taken for any points. I don't think Kaldi uses mag spec without Mel filter banks as a feature so maybe it doesn't make any difference for them but for us we are increasing the input dimensions in the enhancement network from 400 to 512 without adding any frequency resolution. While running I don't see any time efficiency difference while using 400 vs 512, so maybe it's ok to use 400.

kamo-naoyuki: In general, there are big difference especially if nfft is more large, but I agree that there are very  small difference if nfft==512. Actually, 512 might be slower due to zero padding about STFT case.

However, for me, either two case, 512 or 400, is o.k. In my experiments, both values achieved comparable results.
What do you care about this?

egs/chime4/asr1_multich/run.sh
kamo-naoyuki: > how about adding an option to pass the reference channel. something like negative values for attention based ref or else channel id [0-5] for chime4?
O.K. I'll added.

`mvdr2` and `gev` can't work because there are no  implementation for eigen value decomposition. 
(`argmax` is no problem for training in this case because it's used for getting principal eigen vector and it doesn't require differentiable.)

Actually, I won't to enable these beamformer now because I believe these are no big difference about the performance.

espnet/nets/pytorch_backend/frontends/beamformer.py
kamo-naoyuki: Can you show me the minimal reproducing script for real matrix inverse? Perhaps, my complex tensor has a problem.If not so, we should launch a issue in pytorch.

egs/chime4/asr1_multich/run.sh
sw005320: Then, we may exclude them for now.

egs/chime4/asr1_multich/conf/preprocess.json
sas91: Yes even I got comparable results with both. I would prefer 400 because it has lesser parameters in the enhancement network (lesser input and output dimensions for the masking network).

espnet/nets/pytorch_backend/frontends/beamformer.py
sas91: I am not able to find a simple example for real, but look at the simple 2*2 example with your complex tensor that I encountered while running with REVERB data. 
Screen Shot 2019-04-12 at 12 50 34 PM

It gives an error in this line: o_real = (a.real + torch.matmul(torch.matmul(a.imag, a.real.inverse()), a.imag)).inverse(), where you have used the real inverse of pytorch, so I guess that has the same issue too.

sas91: I was able to take the inverse of the same matrix by just dividing by a constant.
Screen Shot 2019-04-12 at 12 56 22 PM

that is why in my fix I divide both psd_s and psd_n by a constant if the original inverse fails, as you take inverse of one and multiply both, I guess it won't make any difference in the final solution
reg_coeff_tensor = ComplexTensor(torch.rand_like(psd_n.real),
torch.rand_like(psd_n.real))*1e-2
psd_n = psd_n/10e+4
psd_s = psd_s/10e+4
psd_n += reg_coeff_tensor
I also add a small constant in case if the matrix is really singular. I have two except blocks as if division by one constant fails, I try a a constant with higher magnitude.

egs/chime4/asr1_multich/conf/preprocess.json
kamo-naoyuki: Alright. I changed 400 to the default.
I preferred the compatibility to Kaldi, but your pointing is more important.

espnet/nets/pytorch_backend/frontends/beamformer.py
kamo-naoyuki: Thanks. In your example, real singular matrix seems to happen. So this is a problem of ComlexTensor. Let me think about this meanwhile.

https://github.com/kamo-naoyuki/pytorch_complex/blob/2673fdc6bae3e74be2828283e48dd5f6b39d0c5a/torch_complex/tensor.py#L378-L379

```
tensor([[[2.6844e+08, 2.6844e+08],
         [2.6844e+08, 2.6844e+08]]])
```

kamo-naoyuki: I updated pytorch_complex to try inverse() with several `t` parameters.
(pytorch_complex doesn't invoke the inverse of blas API directly, but contaminates the torch.Tensor.invserse() to describe the inverse() of complex matirx referring https://www.math.uwaterloo.ca/~hwolkowi/matrixcookbook.pdf  )

The condition number of this matrix looks too large, so it is hard to get the inverse() accurately essentially. 

```python
import numpy
import torch
from torch_complex.tensor import ComplexTensor

real = torch.tensor([[1.43380048e+15, 1.11050124e+15],
                     [1.11050124e+15, 1.11960107e+15]])
imag = torch.tensor([[8.758782e-04, 6.099766e+14],
                     [-6.099766e+14, 9.679516e-04]])
a = ComplexTensor(real, imag)
print(a.inverse() @ a)

an = a.numpy()
print(numpy.linalg.inv(an) @ an)
print('cond =', numpy.linalg.cond(an))
```

```
ComplexTensor(
Real:
tensor([[2.0000, 0.2500],
        [0.0000, 1.2500]])
Imag:
tensor([[ 2.2500,  2.2500],
        [-3.2500, -2.4375]])
)
[[ 0.        +1.j -0.91687727+0.j]
 [ 1.3755432 +0.j  2.        -1.j]]
cond = 116339144.0
```

Perhaps we may want to skip the uptating in training in this case, but I think we can ignore it because this is very rare event. 
I'm not particular about numerical calculation so much,  so what do you think?

sas91: Yes maybe for now at least we should include code to skip updating. I think we should not ignore it as it not that rare. I encounter it quite often while running with REVERB data and also in the correlation matrix inverse part of WPE. It will be nice if we can find a permanent fix to it.

kamo-naoyuki: WPE and REVERB is not within the range of this PR, so I won't include for now. It requires several changing in the main codes. 

Maybe such unstable matrix often happens in early training steps and that is not dominant as a whole, so , at least, the training is possible in the current code.

sw005320: OK. We can just make it pending for this PR.

egs/chime4/asr1_multich/run.sh
kamo-naoyuki: I added an option for ref channel and deleted mvdr2 and gev.

espnet/nets/pytorch_backend/frontends/beamformer.py
kamo-naoyuki: @sas91 Maybe I don't have enough time  for the duration after this PR. If you could provide your base script of REVERB as another PR, it's great for me.
Thanks for your help for this.

sw005320:
Thanks a lot!!!
This is a great contribution.

nlpss94136:
Can I download your code? Thank you very much!

PR #603, opened 2019-02-25
Docker updates v0.3.0

Fhrozen:
Updates to support v0.3.0 in docker containers.
Test running.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -27,6 +27,7 @@ egs/*/*/wav*
 
 # tools related
 tools/chainer_ctc/
+tools/kaldi*
 tools/kaldi/
 tools/kaldi-io-for-python/
 tools/kaldi_github/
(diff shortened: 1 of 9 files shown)


.gitignore
```diff
@@ -27,6 +27,7 @@ egs/*/*/wav*
 
 # tools related
 tools/chainer_ctc/
+tools/kaldi*
```
sw005320: If you use `tools/kaldi*`, we may not require `tools/kaldi/`, `tools/kaldi-io-for-python/`, etc.

Fhrozen: Yes, It will be better to set this to tools/kaldi*, especially when the kaldi installation is a symlink. Still doing some test, and upload this modification

zh794390558:
Is there any images on DockerHub ?

Fhrozen:
Yes, All the images are already allocated in docker hub as espnet/espnet:${tag}. However, they only support before v0.3.0. I will be updating them in a few hours.

Fhrozen:
Tests, are completed. All containers are working without error. Ready to merge

PR #609, opened 2019-02-27
Fixed check_install.py

kan-bayashi:
- Refactored `check_install.py`
- Fixed `check_install.py` to be compatible with several verions.
- Added `check_install` in Makefile to run `check_install.py` after installation


sw005320:
Can you modify the README.md accordingly?

kan-bayashi:
I’m not sure we should change README.

PR #610, opened 2019-02-27
Wsj mix v.0.3.0

simpleoier:
Fix bugs in data preparation and problems with zip operation in Python 3.x


egs/wsj_mix/asr1/run.sh
```diff
@@ -115,17 +115,16 @@ if [ ${stage} -le 0 ]; then
     local/wsj0_2mix_data_prep.sh ${wsj_2mix_wav}/wav16k/max ${wsj_2mix_scripts} \
         ${wsj_full_wav} || exit 1;
 
-    ### Or this part is for WSJ mix
```
sw005320: I think it's better to add some clarification comments about the difference between `local/wsj_2mix_data_prep.sh` and `local/wsj0_2mix_data_prep.sh`

egs/wsj_mix/asr1/run.sh
```diff
@@ -300,7 +299,6 @@ if [ ${stage} -le 4 ]; then
         --minibatches ${N} \
         --verbose ${verbose} \
         --resume ${resume} \
-        --seed ${seed} \
```
sw005320: You do not have to delete the seed option.

espnet/asr/pytorch_backend/asr_mix.py
```diff
@@ -83,6 +83,7 @@ def __call__(self, batch, device):
         # batch should be located in list
         assert len(batch) == 1
         xs, ys = batch[0]
+        ys = list(ys)  # Convert zip object to list in python 3.x
```
sw005320: Could you make sure that it's working on both 3.x and 2.x?
(for now, we try to support both versions).

sw005320:
@simpleoier, could you respond my comments?

egs/wsj_mix/asr1/run.sh
sw005320: Could you please cancel to delete this line?
${seed} is not used anymore then.

espnet/asr/pytorch_backend/asr_mix.py
sw005320: @simpleoier, could you answer this question? This is just for the confirmation.

simpleoier: Yes. I tested on both Python 3.x and 2.x, it works well. Besides, Marc's experience also confirmed that it works on Python 3.x.

sw005320:
@simpleoier, can you also fix https://travis-ci.org/espnet/espnet/jobs/503227382#L1331-L1339? I don't know why this warning was passed.

simpleoier:
OK.

PR #614, opened 2019-02-27
Uses hubscr for swbd scoring

gtache:
Adds trn_to_ctm and trn_to_stm to create appropriate files for scoring and adapts kaldi swbd's score_sclite for espnet

By using this scoring scheme (the same as kaldi, so we have a common scoring now), I got a WER reduction from, for example, 58.3% to 44.1%, 37.7% to 26.3%, etc

It could probably be applied to other hub5 benchmarks like fisher or callhome too


sw005320:
This is great!
The current option is fine, but we could think of converting from JSON to CTM/STM directory.
Also, yes, it would be great if you also made it for the fisher_swbd recipe.

PR #615, opened 2019-02-27
Add model compatibility test

kan-bayashi:


kan-bayashi:
Travis error is caused by #616.

kan-bayashi:
OK I will list those information with file id.

sw005320:
perfect!
now we could keep compatibility easily!

Issue #616, opened 2019-02-27
Conversion of AttributeDict with vars() returns unexpected results


kan-bayashi:
I found a bug. 
In training phase, `train_args` is `argparse.Namespace`.
So `vars(train_args)` convert into dict as follows.

```python
(Pdb) train_args
[...]

PR #617, opened 2019-02-28
Fix a fatal bug around CTC when testing

kamo-naoyuki:
#616 
This is a fatal bug because now always using pytorch.CTCLoss instead of warpctc for testing.


kan-bayashi:
Thanks!

PR #625, opened 2019-03-06
update for fisher_callhome_spanish recipe

JaejinCho:
1. change sp_perturb to be optional
2. changes in local/ to be compatible w/ py3
3. Add some lines to filter out utterances which don't have any text (transcript). After "merge_scp2json.py" is added in utils/, those utterances raised an error.


egs/fisher_callhome_spanish/asr1/local/fsp_data_prep.sh
```diff
@@ -138,7 +138,8 @@ if [ $stage -le 2 ]; then
   grep -v '()' | \
   #Now go after the non-printable characters and multiple spaces
   sed -r 's:¿::g'  | sed 's/^\s\s*|\s\s*$//g' | sed 's/\s\s*/ /g' > $tmpdir/text.2
-  cp $tmpdir/text.2 $dir/train_all/text
+  awk -F' ' '{print NF" "NR}' $tmpdir/text.2 | grep -w ^1 | awk -F' ' '{print $2}' | sed 's%$%d%' | sed -f - $tmpdir/text.2 > $tmpdir/text.3
```
sw005320: better to add a comment to explain this additional process.

egs/fisher_callhome_spanish/asr1/local/fsp_make_spk2gender.sh
```diff
@@ -12,9 +12,9 @@ tmpFileLocation = 'data/local/tmp/spk2gendertmp'
 tmpFile = None
 
 try:
-     tmpFile = open(tmpFileLocation)
+    tmpFile = open(tmpFileLocation)
```
sw005320: It's better to rename this function *.py

egs/fisher_callhome_spanish/asr1/local/fsp_data_prep.sh
```diff
@@ -138,6 +138,7 @@ if [ $stage -le 2 ]; then
   grep -v '()' | \
   #Now go after the non-printable characters and multiple spaces
   sed -r 's:¿::g'  | sed 's/^\s\s*|\s\s*$//g' | sed 's/\s\s*/ /g' > $tmpdir/text.2
+  #Filter out uttreances that don't have any text transcription
   awk -F' ' '{print NF" "NR}' $tmpdir/text.2 | grep -w ^1 | awk -F' ' '{print $2}' | sed 's%$%d%' | sed -f - $tmpdir/text.2 > $tmpdir/text.3
```
sw005320: OK. Frankly, you seem to do simple things with a complicated way. Can you rewrite this part more simply? (I could not follow this).

JaejinCho: That could be true that it is a complicated way (just tried to implement it in one line) but I cannot come up with a simipler way w/ my limited knowledge in a shell programming. However, below is the alternative that might be easier to follow. The first line gets indices of the utterances w/ NO transcription (so the number of fields (NF), separated by space, equals to 1. Then, in the next line, we delete those lines by the indices and generate a new file.

awk -F' ' '{print NF}' text.2 | grep -n "\<1\>" | awk -F':' '{print $1"d"}' > del.tmp
sed -f del.tmp text.2 > text.3 && rm del.tmp

Let me know if this looks better.

JaejinCho:
Thanks @sw005320 for making it simipler

sw005320:
thanks!

PR #627, opened 2019-03-06
Modify for parallel back ground jobs

kamo-naoyuki:
This is a bash technique: `wait` command always returns 0 without arguments, and if pid is given, then it can return the return code of the background process. 

```
set -e 

fail_command &
wait  # Always 0, so cannot stop here

fail_command &
wait $! # returns non-zero, so this code will exit here.
```

I modified an4/run.sh only for example.

### DIFF
--- a/egs/an4/asr1/run.sh
+++ b/egs/an4/asr1/run.sh
@@ -253,8 +253,7 @@ if [ ${stage} -le 4 ] && [ ${stop_stage} -ge 4 ]; then
     ) &
     pids+=($!) # store background pids
     done
-    for pid in "${pids[@]}"; do wait ${pid} ; done
-
-    echo "Finished"
+    i=0; for pid in "${pids[@]}"; do wait ${pid} || ((i++)); done
+    [ ${i} -gt 0 ] && echo "$0: ${i} background jobs are failed." && false
 fi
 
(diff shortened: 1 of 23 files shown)


egs/an4/asr1/run.sh
```diff
@@ -245,15 +246,15 @@ if [ ${stage} -le 4 ] && [ ${stop_stage} -ge 4 ]; then
...
-        wait
+            --ctc-weight ${ctc_weight}
 
         score_sclite.sh ${expdir}/${decode_dir} ${dict}
 
     ) &
+    pids+=($!)
     done
-    wait
+    for pid in "${pids[@]}"; do wait ${pid} || { echo "Some jobs failed"; exit 1; }; done
```
sw005320: How about `echo "Some background jobs are failed";`
Also, do we need `|| { echo "Some jobs failed"; exit 1; }`?

kamo-naoyuki:
```bash
set -e
pids=()
{ sleep 1 ; echo Job1 ; false; } &
pids+=($!)
{ sleep 3 ; echo Job2 ; false; } &
pids+=($!)
{ sleep 5 ; echo Job3 ; false; } &
pids+=($!)
for i in "${pids[@]}"; do wait ${i}; done
```
I noticed a problem. This program exits at the first waiting, but the other background jobs are still alive, so they are detached from the parent process.

kamo-naoyuki:
This code may be the best practice.

FYI, this one is rejected.

- `jobs -p` instead of storing pids
```bash
<some-command-quickly-finish> &
for pid in $(jobs -p); do wait ${pid}; done  # "jobs -p" possibly returns nothing
```

sw005320:
Thanks! This is one of the issues that I wanted to fix.

Issue #606, opened 2019-02-26
Download failed in the installation of mecab-ipadic-NEologd


tfygg:
install-mecab-ipadic-NEologd] : Make mecab-ipadic-NEologd
[make-mecab-ipadic-NEologd] : Start..
[make-mecab-ipadic-NEologd] : Check local seed directory
[make-mecab-ipadic-NEologd] : Check local seed file
[make-mecab-ipadic-NEologd] : Check local build directory
[make-mecab-ipadic-NEologd] : create /home/luban/tongfeiyang/speech/espnet/tools/mecab/mecab-ipadic-neologd/libexec/../build
[make-mecab-ipadic-NEologd] : Download original mecab-ipadic file
Makefile:105: recipe for target 'mecab.done' failed
make: *** [mecab.done] Error 7

kan-bayashi:
Could you test the patch #608?

ZhishengWang:
same error occured on ubuntu16.04, tested on patch #622：
[make-mecab-ipadic-NEologd] : Check local seed directory
[make-mecab-ipadic-NEologd] : Check local seed file
[make-mecab-ipadic-NEologd] : Check local build directory
[make-mecab-ipadic-NEologd] : create /home/zswang/githubs/espnet/tools/mecab/mecab-ipadic-neologd/libexec/../build
[make-mecab-ipadic-NEologd] : Download original mecab-ipadic file
Makefile:105: recipe for target 'mecab.done' failed
make: *** [mecab.done] Error 7

kan-bayashi:
I could find the reason from this error.
Could you run `make` without `-j` option?

kan-bayashi:
I checked NEologd installation code, It uses google drive to download files.
I doubt that you cannot use google drive.

ZhishengWang:
Thanks. Google service is blocked in my country, I will try to figure out if this is the reason for the issue. Belowings are the logs realted to NEologd installation when run with **make mecab.done**:

cd mecab && git clone --depth 1 https://github.com/neologd/mecab-ipadic-neologd.git
Cloning into 'mecab-ipadic-neologd'...
remote: Enumerating objects: 75, done.
remote: Counting objects: 100% (75/75), done.
remote: Compressing objects: 100% (74/74), done.
remote: Total 75 (delta 5), reused 55 (delta 0), pack-reused 0
Unpacking objects: 100% (75/75), done.
Checking connectivity... done.
[...]

PR #628, opened 2019-03-07
Add minimum installation in Makefile

kan-bayashi:
Sometimes, we want to skip the installation of `extra` modules (#606).
So I added minimum installation in Makefile.
Also, I added extra module installation check in `path.sh`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -38,7 +38,7 @@ and also follows [Kaldi](http://kaldi-asr.org/) style data processing, feature e
 - NCCL 2.0+ (for the use of multi-GPUs)
 - protocol buffer (for the sentencepiece, you need to install via package manager e.g. `sudo apt-get install libprotobuf9v5 protobuf-compiler libprotobuf-dev`. See details `Installation` of https://github.com/google/sentencepiece/blob/master/README.md)
 
-- PyTorch 0.4.1, 1.0.0 
+- PyTorch 0.4.1, 1.0.0
 - gcc>=4.9 for PyTorch1.0.0
 - Chainer 5.0.0
 
@@ -124,7 +124,14 @@ If there are some problems in python libraries, you can re-setup only python env
 ```sh
 $ cd tools
 $ make clean_python
-$ make all_python
+$ make python
+```
+
+If there are some problems during the installation of extra modules (e.g. [#606](https://github.com/espnet/espnet/issues/606), you can try minimum installation via following commands
+```sh
+$ cd tools
+$ make clean
+$ make minimum
 ```
 And then check the install is succeeded again.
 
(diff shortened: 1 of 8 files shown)


sw005320:
You can use `minimum` as `all`, and exclude `extra` from `all`.

sw005320:
It looks good, but please make sure to test this modification in advance from scratch. Then, I'll merge it.

kan-bayashi:
OK. I will test `make` on CentOS 7.5 and Ubuntu 16.04.

kan-bayashi:
I confirmed that `make -j 10 KALDI=<path_to>/kaldi` and `make -j 10 extra` worked correctly on CentOS 7.5  and Ubuntu 16.04.
But recently kaldi installation does not work in CentOS 7.5 due to following error.

```bash
cd kaldi/src; ./configure --shared --use-cuda=no; make depend; make all
Configuring ...
Checking compiler g++ ...
Checking OpenFst library in /home/t_hayashi/tmp/espnet/tools/kaldi/tools/openfst ...
Doing OS specific configurations ...
On Linux: Checking for linear algebra header files ...
Using ATLAS as the linear algebra library.
Successfully configured for red hat [dynamic libraries, fat] with ATLASLIBS =/usr/lib64/atlas/libsatlas.so.3 /usr/lib64/atlas/libtatlas.so.3
Validating presence of ATLAS libs in /usr/lib64/atlas
Could not find any libraries /usr/lib64/atlas/liblapack /usr/lib64/atlas/liblapack_atlas /usr/lib64/atlas/libclapack that seem to be an ATLAS CLAPACK library.
** Failed to configure ATLAS libraries ***
**  ERROR   **
** Configure cannot proceed automatically.
**  If you know that you have ATLAS installed somewhere on your machine, you
** may be able to proceed by replacing [somewhere] in kaldi.mk with a directory.
**  If you have sudo (root) access you could install the ATLAS package on your
** machine, e.g. 'sudo apt-get install libatlas-dev libatlas-base-dev' or
** 'sudo yum install atlas.x86_64' or 'sudo zypper install libatlas3-devel',
** or on cygwin, install atlas from the installer GUI; and then run ./configure
** again.
**
**  Otherwise (or if you prefer OpenBLAS for speed), you could go the OpenBLAS
** route: cd to ../tools, type 'extras/install_openblas.sh', cd back to here,
** and type './configure  --openblas-root=../tools/OpenBLAS/install'
```
Note that `sudo yum install atlas.x86_64` have already performed.

This is a problem, but it is not related to this PR.

kan-bayashi:
I figure out the reason of the failure of the kaldi installation on CentOS 7.5.
I made PR to fix this issue, it will be fixed soon.
https://github.com/kaldi-asr/kaldi/pull/3075

PR #634, opened 2019-03-11
update makefile for chainer-ctc installation

marvinzh:
original command may not suitable for non-root user on some Linux distributions


kan-bayashi:
>  I think we may not need --user when we use venv/bin/activate

I think so too. I should not be necessary.
When using `venv/bin/activate`, we do not need sudo to install libraries with `pip`.

@marvinzh Could you tell us your linux distributions? 
We mainly use Ubuntu or CentOS. I'm not sure other linux distributions.
And we assume that we do not edit these external libraries (e.g.`chainer_ctc, warp-ctc`)
So I think `-e` option is not needed.

It looks your edited `Makefile` is too old compared with current master.
It is better to merge current master's `Makefile` changes.

PR #637, opened 2019-03-11
Disable set -x in score_sclite.sh

kamo-naoyuki:
I think this line was inserted just for debugging and he may forget to delete it.


sw005320:
Thanks!

Issue #640, opened 2019-03-13
JSONDecodeError raised in wsj_mix decoding step


vyouman:
There's something wrong with the json IO format in the decoding step. I checked the generated json in the tmp directory. The generated result_r*h*.wrd.json files are normal, but the format of all_result.wrd.json is kind of weird. Maybe there is some bug in mergeresultjson.py. Could you please check it?

> dictionary: data/lang_1char/tr_units.txt
> stage 5: Decoding
> 2019-03-12 19:58:07,609 (splitjson:35) INFO: /home/user/vyouman/espnet/tools/venv/bin/python /home/user/vyouman/espnet/egs/wsj_mix/asr1/../../../utils/splitjson.py --parts 4 dump/tt/deltatrue/data.json
> 2019-03-12 19:58:07,720 (splitjson:47) INFO: number of utterances = 3000
[...]

PR #641, opened 2019-03-13
Wsj mix v.0.3.0

simpleoier:
#640


sw005320:
It's better to add the unit test for both scripts to avoid erroes in future.

simpleoier:
OK, I will add the unit test for it.

sw005320:
@simpleoier, again thanks for your very quick bug fix. And it would be great if you add the unit test for these scripts soon.

PR #644, opened 2019-03-13
follow #628

kamo-naoyuki:
@kan-bayashi `return` should be used for sourced script instead of `exit` because it drops the shell on interactive mode.

```
% source path.sh
[Exit]
```


kan-bayashi:
Thanks!

PR #645, opened 2019-03-13
[WIP] Streaming attention decoder

pzelasko:


qmpzzpmq:
@pzelasko Do you have the paper of this Streaming attention decoder

pzelasko:
I left it in a state where the code seems to work, but the results are not correct, because it requires re-thinking how to combine the output of the two decoder heads (ctc and attention) when the whole signal is not available up front.

I guess we ultimately decided to put this on hold and revisit it later on simpler decoder architectures. Would that be a correct statement, Shinji?

sw005320:
> I guess we ultimately decided to put this on hold and revisit it later on simpler decoder architectures. Would that be a correct statement, Shinji?

Yes.

PR #646, opened 2019-03-14
Update remove_longshortdata.sh

zh794390558:
count chars without key column


sw005320:
I think we skip the first key column with `text2token.py -s 1`
See https://github.com/espnet/espnet/blob/master/utils/text2token.py#L35-L36

zh794390558:
`text2token.py -s 1 ` also output `key` column.

sw005320:
Yes, this is true.
Thanks!

PR #648, opened 2019-03-14
Fix #627

kamo-naoyuki:
Sorry sending for many times.

If the evaluated result is 0, the returned value becomes 1.
```
((100*0))  # return 1
let i=0   # return 1
expr -1 + 1  # return 1

i=0
((i++))  # return 1
i=0
((++i))  # return 0
```


sw005320:
Thanks!

PR #649, opened 2019-03-14
Fix google drive download error

kan-bayashi:
Sometlmes, download from google drive is failed due to its vrius check confirmation.
To avoid it, I add extra processing.
https://stackoverflow.com/questions/20665881/direct-download-from-google-drive-using-google-drive-api


kan-bayashi:
Done. (But maybe not smart way...)

sw005320:
Thanks!
Another option is to use GitHub large file storage.
Let's think of it.

Issue #674, opened 2019-04-05
Run decoding on local cpu


vkothapally:
We have two machines, one has GPUs built in it with only 8 cores. On the other server, we have no GPUs but many computing nodes to speed up tasks. I was able to install Espnet successfully on the GPU machine and use it to train end-to-end architectures. However, it consumes a lot of my time to decode (which makes sense - Espnet doesn't use GPUs to decode and my machine has only 8 cores). So, I was planning on installing Espnet on the second machine which can handle the decoding well. As mentioned earlier, it doesn't have GPU and so it lacks cuda and cudnn ( I have no access to install them on this machine). I'm here to see if there is any way around this. All I'm trying to do is use one machine with GPUs to train and the other with more cpu nodes to decode.

Any help is appreciated. Thanks in Advance.!

Fhrozen:
If you want to install ESPnet in nogpu pc, you just only need to remove the cupy lines from Makefile.
as example you can check the Dockerfile that remove the cupy install to generate the CPU images:
https://github.com/espnet/espnet/blob/master/docker/prebuilt/devel/Dockerfile

sw005320:
Yes, if you remove the cupy lines you can install even with such machines, but maybe we can think of supporting the cpu only installation with Makefile options. @Fhrozen, could you make such options and corresponding documents?

@vkothapally, you could also enjoy the GPU decoding by using --ngpu 1 and --batchsize 8.

Fhrozen:
@sw005320 How should be this implemented? to disable cupy installation automatically or by user selection. In case of be automatically, I can move the commands from the dockerfile to makefile that disable cupy installation if nvcc is not detected or if it does not return a response. So there is no need for user to set any additional flag.

sw005320:
A user selection is easier.

vkothapally:
Thanks a lot for the help. I was able to decode using GPU (I guess for now it is just limited to using a single GPU). Also, I edited the Makefile and checkinstall.py files to remove any cupy instances. It worked for me.

PR #677, opened 2019-04-08
No Cupy install

Fhrozen:
Adding cupy label to makefile and no-cupy flag
Referred to #674.
Did some modifications in the makefile to reuse the CUPY_VERSION flag.
in case empty, the cupy installation is avoided.
I tested with both `make` and `make CUPY_VERSION=''` without any problem.
Let me know if the modifications are acceptable to modify the Readme file also.

### DIFF
--- a/tools/Makefile
+++ b/tools/Makefile
@@ -18,7 +18,7 @@ GCC_VERSION := $(shell gcc -dumpversion)
 
 all: kaldi.done python check_install
 
-python: venv chainer_patch.done cupy.done warp-ctc.done chainer_ctc.done
+python: venv chainer_patch.done warp-ctc.done chainer_ctc.done
 
 extra: nkf.done sentencepiece.done mecab.done moses.done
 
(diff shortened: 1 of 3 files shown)


tools/check_install.py
```diff
@@ -8,110 +8,123 @@
...
-        if hasattr(lib, "__version__"):
-            is_correct = lib.__version__ in version
-            if is_correct:
-                logging.info("--> %s version is matched." % name)
-                is_correct_version_list.append(True)
+
+def main(args):
+    parser = argparse.ArgumentParser()
+    parser.add_argument('--no-cupy', action='store_true', default=False,
+                        help='Number of GPUs')
```
kan-bayashi: Fix help comment.

tools/Makefile
```diff
@@ -62,10 +62,18 @@ espnet.done: venv
...
 	touch espnet.done
 endif
 
+ifneq ($(strip $(CUPY_VERSION)),)
+cupy.done: venv
+	. venv/bin/activate && pip install cupy==$(CUPY_VERSION)
+	touch cupy.done
+else
+cupy.done: venv
+	touch cupy.done
```
kan-bayashi: I’m not sure cupy.done is needed if we do not install cupy.

tools/check_install.py
Fhrozen: resolved in new commit

tools/Makefile
Fhrozen: resolved in new commit

Fhrozen:
Pushed a new commit resolving observations from @kan-bayashi 
Moved cupy to be installed under the chainer_patch target, if CUPY_VERSION has a value.
Test both installations without problems.
Let me know if there is any possible conflict or wrong part.

kan-bayashi:
Sorry, my intension is as follows:

```
cupy.done: espnet.done
ifneq ($(strip $(CUPY_VERSION)),)
	. venv/bin/activate && pip install cupy==$(CUPY_VERSION)
        touch cupy.done
endif
```

sw005320:
I like @kan-bayashi's idea. @Fhrozen, can you change it like that?

Fhrozen:
Just finished the test for @kan-bayashi's idea. @sw005320 Let me know if there is any additional part to modify.

Issue #656, opened 2019-03-20
there is a problem in installation with compiled Kaldi


liuchenbaidu:
when i run make KALDI=mykaldipath,it gives error. my pip version is 19.0.3.the python version is 2.7.15
 
Matplotlib 3.0+ does not support Python 2.x, 3.0, 3.1, 3.2, 3.3, or 3.4.
    Beginning with Matplotlib 3.0, Python 3.5 and above is required.

    This may be due to an out of date pip.

    Make sure you have pip >= 9.0.1.

    ----------------------------------------
Command "python setup.py egg_info" failed with error code 1 in /tmp/pip-install-40ek3V/matplotlib/

kan-bayashi:
Thank you for your report!
Did you use current master?
Now we use Makefile which uses conda with Python3.7 as default, 
so if you run `make KALDI=<your_kaldi_path>`, it should be working.
(Maybe you run `make KALDI=<your_kaldi_path> PYTHON=<your_kaldi_path>` ?)

Anyway, we will fix matplotlib version for python2.

liuchenbaidu:
i use a old v.0.3.0.and i find the version of python must be 2.7.15.

kan-bayashi:
Fixed in (#679)

PR #679, opened 2019-04-10
Add version check for matplotlib installation

kan-bayashi:
Fix the issue (#656)


sw005320:
Thanks.
We may think of excluding the python 2.X support in the near future.

Issue #665, opened 2019-03-29
recognize_beam_batch throws an error AttributeError: 'list' object has no attribute 'view' because a_prev is None (fails for all coverage, location, multi_head variants)


sknadig:
In [decoders.py#L457](https://github.com/espnet/espnet/blob/master/espnet/nets/pytorch_backend/decoders.py#L457) `a_prev` is not updated, it is always `None`.
This results in the [condition for handling list](https://github.com/espnet/espnet/blob/master/espnet/nets/pytorch_backend/decoders.py#L515) `if not isinstance(a_prev, list)` to be True always.
`att_w` will be a list for thest attention mechanisms, `att_w.view(n_bb, -1)` fails.
`AttributeError: 'list' object has no attribute 'view'`

One solution is to use `att_w` instead of `a_prev` while calling `self.att` at [decoders.py#L457](https://github.com/espnet/espnet/blob/master/espnet/nets/pytorch_backend/decoders.py#L457)
`att_c, att_w = self.att[att_idx](exp_h, exp_hlens, self.dropout_dec[0](z_prev[0]), a_prev)`

I think this is related to #510

sknadig:
For location_recurrent type attention, the return values are `return c, (w, (att_h, att_c))`
This needs to be handled separately.
```
if isinstance(att_w, tuple):
    #handle the case of location_recurrent when return is a tuple
    a_prev = torch.index_select(att_w[0].view(n_bb, -1), 0, vidx)
elif not isinstance(att_w, list):
    a_prev = torch.index_select(att_w.view(n_bb, -1), 0, vidx)
else:
    # adapt for multi-head attention
    a_prev = [torch.index_select(att_w_one.view(n_bb, -1), 0, vidx) for att_w_one in att_w]
```

sw005320:
I see, thanks. @kan-bayashi, could you fix it? Probably better to add the test for this.

kan-bayashi:
Fixed (#681)

PR #681, opened 2019-04-10
Fix recognize_batch for 2d, location_reccurent, multi-head attentions

kan-bayashi:
Fix the issue (#665) and add test to perform `recognize_batch` for all of attentions.


sw005320:
Thanks!

PR #684, opened 2019-04-19
bugfix in voxforge

Masao-Someki:
* when running the stage 0, UnicodeDeodeError was occured.
	→add encoding option in local/voxforge_data_prep.sh#L127
* when running the stage 4, a problem occered:number of utterances < #splits
	→change nj=32 to nj=16 in run.sh


kan-bayashi:
Thanks!

PR #687, opened 2019-04-19
add CUDA_VERSION in Makefile

ShigekiKarita:
I want to specify CUDA version when I install pytorch binary because its default is CUDA9 but I have a different version.

### DIFF
--- a/README.md
+++ b/README.md
@@ -82,10 +82,10 @@ $ cd tools
 $ make KALDI=/path/to/kaldi
 ```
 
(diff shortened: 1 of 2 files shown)


ShigekiKarita:
I see. I've added new subsections for conda or system python because currently `PYTHON_VERSION`, `CUDA_VERSION` and `TH_VERSION` are only available for conda.

sw005320:
Thanks!!

PR #691, opened 2019-04-22
Update the tedlium recipe

sw005320:
- Use BPE to make e2e ASR and LM training fast
- Use speed perturbation to augment the training data.
- Use a wide-shallow encoder

Now the test-set WER is 12.7%, which is very close to the LF-MMI chain model performance (~12.8%)


fanlu:
@sw005320 which RESULTS did you get 12.8%?
[https://github.com/kaldi-asr/kaldi/tree/master/egs/tedlium](url)

sw005320:
Oh, I get it from tedlium v3. I should get the number from https://github.com/kaldi-asr/kaldi/blob/master/egs/tedlium/s5_r2/RESULTS. The Kaldi best is <9.0 and Kaldi is still better.

fanlu:
thanks @sw005320

sw005320:
Further correction. I get it from tedlium v1...

PR #702, opened 2019-04-24
fix choice between pytorch and pytorch-cpu

jan-schuchardt:
With 1f758b5445dc8963d0785a612fae8c886342cc10 , an additional condition was added to the makefile in order to ensure that the cpu version of pytorch would be used, unless nvidia-smi was available.

However, as of now, pytorch-cpu is downloaded if nvidia-smi is available. The normal pytorch library is downloaded if nvidia-smi isn't available.

Also fixed a syntax error.


ShigekiKarita:
Thanks! it was my mistake.

Issue #699, opened 2019-04-23
merge_json behavior when no labels.


sw005320:
@kamo-naoyuki, I've found that there is a case when we don't have any label output in some utterances for the switchboard evaluation data (`rt03`). But current `data2json.sh` cannot allow such cases 
https://github.com/espnet/espnet/blob/master/utils/merge_scp2json.py#L210-L214
Can you relax this?
This check is good for the training data, but it excludes such utterances in the evaluation set, and makes the evaluation different from other benchmarks.

PR #708, opened 2019-04-27
Fix merge_json behavior when no labels #699

kamo-naoyuki:
#699

I added `--allow-one-column` to `data.sh` and `merge_scp2json.py`.
If `text` having no labels is expected, we need to add `--allow-one-column true` in such recipes. Maybe this is not good solution.

@sw005320 Can you modify swbd recipe? I can't get the data now.


sw005320:
Thanks a lot!

PR #712, opened 2019-05-03
[WIP] Cycle consistency (ASRTTS)

creatorscan:


ShigekiKarita:
Hi, nice to see you in ICASSP2019! Thanks for your contribution.

I will review your PR. Can you remove some garbage files e.g., `.cache`

ShigekiKarita:
There are still too many files to review. Unstage these files from git
-  `egs/librispeech/asrtts1/utils/`: I believe it should be a symbolic link to `kaldi/egs/wsj/s5/utils`. If there is anything you wrote. move your scripts to `local`
- `egs/librispeech/asrtts1/rnnlm_model/`: oh ... rnnlm.model.19 is too big (30MB). you need to completely remove this from git history. see https://stackoverflow.com/questions/2100907/how-to-remove-delete-a-large-file-from-commit-history-in-git-repository
- `egs/librispeech/asrtts1/sid/`: Could you explain where this came from? maybe we should add some commands to install this in `tools/Makefile`

sw005320:
@creatorscan, #746 affects your work. please reflect it. @kan-bayashi will help you if needed.

creatorscan:
> @creatorscan, #746 affects your work. please reflect it. @kan-bayashi will help you if needed.

@sw005320 Shinji, can we just keep cycle related codes as different files. So that it might not affect the renamings.

sw005320:
I think this is a good idea. Maybe we should have `e2e_asrtts.py`?

creatorscan:
> I think this is a good idea. Maybe we should have `e2e_asrtts.py`?

Sure Shinji. In that case we can even include the basic TTS related class and functions in the e2e_asrtts. As the original e2e_tts is going through multiple changes which our model dont necessarily require. What do u think ?

sw005320:
#746 intended to include transformer TTS, which would have a great benefit for you. So, you may need some work, but calling basic TTS related class and functions from `e2e_tts.py` would be better. Then, you can easily use the transformer :)

ShigekiKarita:
Thanks for cleaning up files. It is perfect.

> So that it might not affect the renamings.

I agree. Overall TTS impl does not change a lot while its interface is changed. You will find a few errors by argument mismatch.

My advice is writing `test/test_asrtts.py` first like [the existing code using synthetic data to emulate training and decoding](https://github.com/espnet/espnet/blob/c1737105ded30ca6f8d2c5abc5472938ec6006be/test/test_e2e_transformer.py#L123-L179). Then we can find out those errors easily without trying `run.sh` over and over. Of course, it help us a lot when we maintain the test to be passed in the future, or when we add Transformer variants as @sw005320 said. Otherwise the code will be abandoned.

sw005320:
@creatorscan, your push seems to include a lot of conflicts. It would be better to fix this before pushing.

Issue #713, opened 2019-05-06
Problem is tedlium/asr1


gruly:
The issue #694 was closed but the problem is still existing in the run.sh file of the tedlium/asr1 recipe.
I suggested a fix: https://github.com/espnet/espnet/issues/694

kan-bayashi:
I will fix this issue today.

gruly:
Thanks.

Issue #694, opened 2019-04-23
tedlium asr1 problem


fanlu:
When I user docker to run tedlium asr1 recipe
```
./run.sh --docker_gpu 0,1,2,3 --docker_egs tedlium/asr1 --ngpu 4
```
I got some bugs in stage 5
```
stage 5: Decoding
2019-04-22 17:47:02,929 (splitjson:35) INFO: /espnet/tools/venv/bin/python /espnet/egs/tedlium/asr1/../../../utils/splitjson.py --parts 32 dump/dev/deltafalse/data.json
2019-04-22 17:47:02,932 (splitjson:35) INFO: /espnet/tools/venv/bin/python /espnet/egs/tedlium/asr1/../../../utils/splitjson.py --parts 32 dump/test/deltafalse/data.json
2019-04-22 17:47:02,981 (splitjson:47) INFO: number of utterances = 507
2019-04-22 17:47:03,022 (splitjson:47) INFO: number of utterances = 1155
run.pl: 32 / 32 failed, log is in exp/train_trim_pytorch_vggblstmp_e4_subsample1_2_2_1_1_unit320_proj320_d1_unit300_location_adim320_aconvc10_aconvf100_mtlalpha0.5_adadelta_sampprob0.0_bs30_mli800_mlo150/decode_test_beam20_emodel.acc.best_p0.0_len0.0-0.0_ctcw0.3_rnnlm1.0_1layer_unit1000_sgd_bs300_word65000/log/decode.*.log
run.pl: 32 / 32 failed, log is in exp/train_trim_pytorch_vggblstmp_e4_subsample1_2_2_1_1_unit320_proj320_d1_unit300_location_adim320_aconvc10_aconvf100_mtlalpha0.5_adadelta_sampprob0.0_bs30_mli800_mlo150/decode_dev_beam20_emodel.acc.best_p0.0_len0.0-0.0_ctcw0.3_rnnlm1.0_1layer_unit1000_sgd_bs300_word65000/log/decode.*.log
./run.sh: 2 background jobs are failed.
run.sh done.
```
```
[...]

PR #714, opened 2019-05-07
Fix duplicated line in tedlium recipe

kan-bayashi:
Fix the issue #713 #694


sw005320:
Additionally, can you control that in an option level of `asr_recog.py`?

sw005320:
Thanks!

Issue #575, opened 2019-02-02
[Discussion] The configuration system for experiments


kamo-naoyuki:
This is a discussion for the general configuration for experiments. This topic is also related to #551 #555

Epsnet depends on `argparase` and `parse_options.sh` for configuration. This style works well only for a simple tool, but now espnet becomes a large project and is going to be more. I will also implement a big feature about dnn-frontend for asr.
I think now is the time to reconsider the style.

## Issue
For example, considering to implement new architecture.

```python
class MyEncoder:
    def __init__(self, param_a, param_b):
        self.param_a = param_a
        self.param_b = param_b
```

In order to integrate it to the master branch, he needs to do the following three:

- Add parser options
```python
parser.add_argument('--param-a')
parser.add_argument('--param-b')
```
- Give args to the Encoder
```
MyEncoder(param_a=args.param_a, param_b=param_b)
```
- Modify `run.sh`
```
param_a=foo
param_b=bar
. parse_options.sh

(...)

expdir=exp/..._param_a${param_a}_param_b${param_b}

(...)

asr_train.py --param-a ${param_a} --param-b ${param_b} ...
```

This is a little hard works and lacks much maintainability. 
Additionally, sometimes an experimenter add options temporally, but these steps seems too many for temporal action. Ideally, he may want to do this only by modifying his model.

## Sacred
 I usually use [sacred](https://github.com/IDSIA/sacred) for DNN experiments. Briefly, you can consider it as a more suitable replacement of `argparase` for experiments.

- argpase
```python
import argparse
if __name__ == '__main__':
    paraser = argparre.ArgumentParser()
    parase.add_argument('--a', type=int, default=3)
    parase.add_argument('--b', type=float, default=4.5)
    args = parser.parse_args()
    some_func(args.a, args.b)
```
```bash
python main.py --a=5 --b=2.3
```
- sacred
```python
from sacred import Experiment
ex = Experiment('exp')
@ex.config
def cfg():
    a = 3
    b = 4.5

@ex.automain
def run(a, b):
    some_func(a, b)
```
```bash
python main.py with a=5 b=2.3
```

### The benefits of sacred
`sacred` has many good features, but I point out two:

- Supports JSON inputting way: These two are equivalent each other.

```bash
% python main.py with a=5 b=2.3 c='foo'
```
```bash
echo '{"a": 5, "b": 2.3}' > config.json
python main.py with config.json c='foo'
```
- Supports hierarchical options by using `dict`
```python
[...]

PR #722, opened 2019-05-11
Support configuration file

kamo-naoyuki:
#575

`configargparse` is used intead of `argparse`: https://github.com/bw2/ConfigArgParse

This module is almost same as `argparse` except that it supports configuration file:

```bash
# These two are same
python asr_train.py --batch-size 10

echo "batch-size: 10" > config.yaml
python asr_train.py --config config.yaml
```

`configargparse` supports multi file formats for configuration file and I selected Yaml.

I changed `an4` to use this system and I created three configuration file: `train.yaml`, `model.yaml`, `decode.yaml`

```bash
python asr_train.py --config conf/train.yaml --config2 conf/model.yaml 
python asr_recog.py --config conf/decode.yaml 
```

I split the configuration file for training options to `train.yaml`, `model.yaml` because maybe this is more useful, but I'm not sure about this.

I mentioned about `sacred` at #575, but I decided not to use it. `configargparse` is more simple and we can integrate it in the system without big changing.

What do you think?


sw005320:
I like this idea. I also feel that `configargparse` is good enough. If we change it like that, we may also use yaml instead of json for the model configuration (model.json)?

kamo-naoyuki:
Yes, we can also unify the format for the other configuration, model for saving, rnnlm and preprocessing. I'll change the system for `rnnlm` and `preprocessing` at least. 
It is not necessary to do for `model.json`, but maybe it's natural to change it to yaml. So, I'll do it also.

kamo-naoyuki:
- I decided not to change `model.json`. I want to minimize the changing in this PR.
- I merged `train.yaml` and `model.yaml` into one. I concluded it's useless.
- I  modified  `lm_train.py` and `asr_train.py` to use configuration files in `asr1`. 
- It is a little hard alone to change all recipes and also I'm not sure that this direction is really right,  so,  I wish someone would help me for `tts` and `st1` and give me some advises.
- I didn't touch `preprocess.json` in `asr1_multich`. I'll modify it at the next PR.
- I also implemented `utils/change_yaml.py`. This tool is useful for submit `run.sh` with multiple combinations of parameters.

```bash
# -a can be repeated
<in.yaml change_yaml.py -a batch-size=20 -a epochs=15 > out.yaml
./run.sh --train-config out.yaml 
```

I have written  the usage in README.md.

ShigekiKarita:
@kamo-naoyuki I almost agree with your idea. However, do you think YAML should be used?  I like JSON because espnet uses it almost everywhere. (model.json, preprocess.json)

ShigekiKarita:
Wait. I've changed my mind after reading new conf. YAML is better in terms of readability (comment and indent). Maybe we should use YAML everywhere.

espnet/bin/asr_enhance.py
```diff
@@ -12,9 +12,14 @@
...
 
 def main(args):
-    parser = argparse.ArgumentParser(
-        formatter_class=argparse.ArgumentDefaultsHelpFormatter)
+    parser = configargparse.ArgumentParser(
+        config_file_parser_class=configargparse.YAMLConfigFileParser,
+        formatter_class=configargparse.ArgumentDefaultsHelpFormatter)
     # general configuration
+    parser.add('--config', is_config_file=True, help='config file path')
+    parser.add('--config2', is_config_file=True, help='config file path')
```
ShigekiKarita: what these `config2` `config3` for?

utils/change_yaml.py
```diff
@@ -0,0 +1,61 @@
+import argparse
```
ShigekiKarita: need shebang `!/usr/bin/env python3`

README.md
```diff
@@ -219,6 +219,37 @@ Default setup uses CPU (`--ngpu 0`).
...
+Instead of giving arguments directly, we recommend you to modify the yaml file and give it to `run.sh`:
+
+```bash
+./run.sh --train-config conf/train_modified.yaml
+```
+
+We also provide a utility to modify a yaml file from command line:
+
+```bash
+<conf/train.yaml change_yaml.py -a batch-size=24 > conf/train_batchsize24.yaml && ./run.sh --train-config conf/train_batchsize24.yaml
```
ShigekiKarita: Maybe this is too much for your PR but I wanna request auto naming to make CLI shorter. For example
```
# this line saves new file named "conf/train_batch-size24.yaml"
# and print its new name to stdout
$ change_yaml.py conf/train.yaml -a batch-size=24 
conf/train_batch-size24.yaml

# therefore we can simply do
$ ./run.sh --train-config $(change_yaml.py conf/train.yaml -a batch-size=24)
```

ShigekiKarita: If there were auto naming, we do not repeat those names in hyper parameter search
```bash
# before
for lr in 0.1 0.01 0.001; do
  for bs in 8 16 32; do
    <conf/train.yaml change_yaml.py -a batch-size=$bs -a lr=$lr > conf/train_batchsize$bs_lr$lr.yaml
    ./run.sh --train-config conf/train_batchsize$bs_lr$lr.yaml
  done  
done
```
```bash
# after
for lr in 0.1 0.01 0.001; do
  for bs in 8 16 32; do
    ./run.sh --train-config $(change_yaml.py conf/train.yaml -a batch-size=$bs -a lr=$lr)
  done  
done
```

sw005320:
> Wait. I've changed my mind after reading new conf. YAML is better in terms of readability (comment and indent). Maybe we should use YAML everywhere.

This is why I thought that it's better to use YAML instead of `model.json`. But I think `data.json` should be kept.

ShigekiKarita:
Anyway, I'm considering `data.json` is too large in large-scale corpora. It should not be raw text. We should read it in a lazy way. Maybe we can use hdf5 or something compressed and compatible with JSON.

kan-bayashi:
> Wait. I've changed my mind after reading new conf. YAML is better in terms of readability (comment and indent)

I agree with the use of YAML because adding comments in json is a little bit tricky.
I think comments for hyperparameters is very important for new users.

README.md
kan-bayashi: I like @ShigekiKarita ’s idea.
But how do we define the name of `expdir`?
(Currently `basename $train_config .*`)

espnet/bin/asr_enhance.py
kamo-naoyuki: In some use case, someone would like to write only the difference from the base configuration in another configuration file.

```bash
asr-train.py --config conf/train.yaml --config2 conf/new.yaml
```
The options can be overwritten by the latter configuration file.

ShigekiKarita: gotcha. this is more expressive
```python
parser.add('--config2', is_config_file=True, help='second config file path that overwrites the settings in `--config`.')
parser.add('--config3', is_config_file=True, help='third config file path that overwrites the settings in `--config` and `--config2`.')
```

README.md
ShigekiKarita: https://github.com/espnet/espnet/blob/d1e75dbad11b22c9f1f2debb6fc623cbc9719116/egs/aishell/asr1/run.sh#L187-L196

Do you mean these lines? If `new_yaml.py` emits `conf/train_xxx.config` and print its name to the `--train_config $(new_yaml.py ...)`, then `run.sh` creates the expdir at `exp/${train_set}_${backend}_train_xxx/` by default. I do not think this is problem in my usage. Would you describe more what you concern about?

kan-bayashi: Sorry, I misunderstood that `$(new_yaml.py ...)` is standard output of new yaml file.
It is no problem.

espnet/bin/asr_enhance.py
kamo-naoyuki: I added as your comment.

README.md
kamo-naoyuki: @ShigekiKarita 's looks fine. I adopted it in change_yaml.py.

README.md
```diff
@@ -219,6 +219,38 @@ Default setup uses CPU (`--ngpu 0`).
...
+./run.sh --train-config conf/train_modified.yaml
+# e.g.
+./run.sh --train-config conf/train_modified.yaml --decode-config conf/decode_modified.yaml
+```
+
+We also provide a utility to generate a yaml file from the input yaml file:
+
+```bash
+# e.g. You can give any parameters as '-a key=value' and '-a' is repeatable.
+./run.sh --train-config $(change_yaml.py conf/train.yaml -a batch-size=24 -a epochs=10)
```
ShigekiKarita: It is good to describe auto naming here. How about this?
```bash
# e.g. You can give any parameters as '-a key=value' and '-a' is repeatable. 
#      This generates new file at 'conf/train_batch-size24_epochs10.yaml'
./run.sh --train-config $(change_yaml.py conf/train.yaml -a batch-size=24 -a epochs=10)
# e.g. '-o' option specfies the output file name instead of auto named file.
./run.sh --train-config $(change_yaml.py conf/train.yaml -o conf/train2.yaml -a batch-size=24)
```

ShigekiKarita:
LGTM except for README

sw005320:
How about adding more information for the config file name to specify the following?
1. chainer vs. pytorch
2. transformer vs. blstmp, vs. vggblstmp

For example, `train.yaml` -> `train_ch_trans.yaml`, `train_pt_vggblstmp.yaml`.

ShigekiKarita:
I am also thinking about that. We can merge asr1 and asr2 now.

2019年5月24日(金) 2:16 Shinji Watanabe <notifications@github.com>:

sw005320:
I think we can keep asr1 and asr2, but we should merge run_chainer.sh and run_pytorch.sh in asr2

ShigekiKarita:
Done at #745

PR #725, opened 2019-05-13
JNAS_Recipe

Masao-Someki:
* add jnas recipe at ./egs/jnas/asr1/


kan-bayashi:
Could you merge current v.0.4.0 branch changes?

egs/jnas/asr1/run.sh
```diff
@@ -0,0 +1,253 @@
...
+    dump.sh --cmd "$train_cmd" --nj 4 --do_delta ${do_delta} \
+        data/${train_dev}/feats.scp data/${train_set}/cmvn.ark exp/dump_feats/dev ${feat_dt_dir}
+
+    for rtask in ${recog_set}; do
+	# Generate the fbank features.
+	steps/make_fbank_pitch.sh --cmd "$train_cmd" --nj 10 --write_utt2num_frames true \
+	    data/${rtask} exp/make_fbank/${rtask} ${fbankdir}
+
+	# remove longer or shorter datas.
+	remove_longshortdata.sh data/${rtask} data/${rtask}_trim
```
kan-bayashi: For recog sets, we should not perform `remove_longshortdata.sh`.
(Maybe `${rtask}_trim` is not used?)

Masao-Someki:
okay.
yes, if we remove `remove_longshortdata.sh`, we don't need `${rtask}_trim`.

ShigekiKarita:
Great job. Can you also add `JNAS_testset_500` CER in RESULTS?

@sw005320 @kan-bayashi Does this result look good? I could find WER7.18% in http://www.sap.ist.i.kyoto-u.ac.jp/lab/bib/intl/UEN-ICASSP18.pdf  Maybe nice to add README.md.

sw005320:
> @sw005320 @kan-bayashi Does this result look good? I could find WER7.18% in http://www.sap.ist.i.kyoto-u.ac.jp/lab/bib/intl/UEN-ICASSP18.pdf Maybe nice to add README.md.

7.18% is with LM. I think this number seems to be reasonable. We could ask them to compute CER.

Masao-Someki:
Oh, sorry!!
I thought I added the result of testset_500, but it was not in the result....

kan-bayashi:
> @sw005320 @kan-bayashi Does this result look good? 

I think this value is reasonable. JNAS database is not so large and therefore DNN-HMM is much better than E2E, I think.

sw005320:
Thanks a lot!

Issue #726, opened 2019-05-13
SWBD run.sh, text preprocessing


Dreyn:
https://github.com/espnet/espnet/blob/049b8236b160d2e48cfd13dd2891363ade9a13ec/egs/swbd/asr1/run.sh#L107

Hello,

There is, in this line, a Perl command :
```bash
perl -pe 's| \(\%.*\)||g'
```
If I'm correct, this command is executed on the transcriptions files, in order to take off every parenthesis like **(%HESITATION)**. But, Perl beeing greedy by default, if a transcription line had to groups of parenthesis with text between, the two groups of parenthesis are taken as one big, with the text.

For example, when I launch the script, the following line :
"This is huh (%HESITATION) a not so (%HESITATION) inspired example"
Becomes :
"This is huh inspired example"

I solved this issue by adding a question mark '?' behind the the star in the regular expression :
```bash
perl -pe 's| \(\%.*?\)||g'
```
It makes the research no greedy anymore, so that it will look for the shortest expression instead of the longest.

This isn't the only perl command using a greedy research (there is another one just after), but I don't recall what the others do...

kan-bayashi:
Thank you for your report, @Dreyn!
I checked this issue is reproduced.
We will fix it accordingly.

Emrys365:
Hello,

I encountered a problem related to this issue.
When testing the model trained with default config in SWBD, I got the following results:

| No. | model | (eval2000) WER/S.Err | (eval2000/callhm) WER/S.Err | (eval2000/swbd) WER/S.Err |
| --- | --- | --- | --- | --- |
| 1 | blstmp (default) | 39.8/68.8 | 43.5/73.0 | 35.7/62.6 |
| 2 |   + fix issue 726 | 33.4/68.3 | 42.0/72.7 | 24.6/61.9 |
| 3 |   + no convert stm | 31.5/67.1 | 40.6/72.1 | 22.4/59.9 |

The first model was trained using the original `run.sh`, without fixing this issue.
The second model was trained after fixing this issue in stage 0.
The third model was trained in the same way as the first one, but without replacing `stm` file in the decoding stage.

> In the decoding stage, `local/score_sclite.sh` replaces the original `stm` file with a new one, where the parentheses and `(%HESITATION)` are removed. The related code is:
> 
```bash
score_dir=${dir}/scoring
ctm=${score_dir}/hyp.ctm
stm=${score_dir}/ref.stm        # replaced by the original stm in model 3
[...]

PR #729, opened 2019-05-14
Fix perl greedy substitution issue (#726)

kan-bayashi:
Fixed the issue (#726)


kan-bayashi:
I do not have these databases.
@sw005320 Could you check it?

PR #734, opened 2019-05-17
Add implementation of specaug

bobchennan:
Some initial results on WSJ with default setup:

System(#iters) | CER(dev) | WER(dev) | CER(eval) | WER(eval)
-- | -- | -- | -- | --
Baseline(15) | 4 | 8.7 | 2.6 | 5.9
SpecAug(24) | 3.3 | 7.2 | 2.3 | 4.6

In general it requires more iterations to converge and results continue to improve. So far I am not sure if we should do specaug also on validation data.


creatorscan:
> Some initial results on WSJ with default setup:
> System(#iters) 	CER(dev) 	WER(dev) 	CER(eval) 	WER(eval)
> Baseline(15) 	4 	8.7 	2.6 	5.9
> SpecAug(24) 	3.3 	7.2 	2.3 	4.6
> 
> In general it requires more iterations to converge and results continue to improve. So far I am not sure if we should do specaug also on validation data.

This is a great start. Thanks for this. I guess, specaug is not required on validation data.

bobchennan:
Current commit should work.

Some training details:
acc
cer
loss

sw005320:
@kan-bayashi please check it. and also try it to the other recipe.

ShigekiKarita:
> try it to the other recipe

I talked with people at ICASSP19. They found it results pretty good with Librispeech (and Transformer)

kan-bayashi:
COOL,
I will check it tomorrow and test it with transformer in CSJ recipe.

ShigekiKarita:
It looks nice! How about moving this impl into asr_utils.py or somewhere chainer_backend can import.

**(This comment is not for your PR, please ignore)** @kamo-naoyuki  Can we place this at [espnet.transform.perturb](https://github.com/espnet/espnet/blob/master/espnet/transform/perturb.py)? However, this module seems to be WIP. Should we create new preprocessing interface only applied in training stage?

espnet/asr/pytorch_backend/asr.py
```diff
@@ -54,6 +55,7 @@
 from tensorboardX import SummaryWriter
 
 from espnet.utils.deterministic_utils import set_deterministic_pytorch
+from espnet.utils.SparseImageWarp import sparse_image_warp
```
kan-bayashi: Missing file?

espnet/asr/pytorch_backend/asr.py
```diff
@@ -62,6 +64,71 @@
 REPORT_INTERVAL = 100
 
 
+def time_warp(spec, W=5):
```
kan-bayashi: Could you add some docstring? (especially the shape of inputs)

espnet/asr/pytorch_backend/asr.py
```diff
@@ -62,6 +64,71 @@
...
+
+    # Uniform distribution from (0,W) with chance to be up to W negative
+    dist_to_warp = random.randrange(-W, W)
+    src_pts, dest_pts = (torch.tensor([[[y, point_to_warp]]], device=device),
+                         torch.tensor([[[y, point_to_warp + dist_to_warp]]], device=device))
+    warped_spectro, dense_flows = sparse_image_warp(spec, src_pts, dest_pts)
+    return warped_spectro.squeeze(3).squeeze(0)
+
+
+def freq_mask(spec, F=30, num_masks=1, replace_with_zero=False):
```
kan-bayashi: Same comments

espnet/asr/pytorch_backend/asr.py
```diff
@@ -62,6 +64,71 @@
...
+
+        mask_end = random.randrange(f_zero, f_zero + f)
+        if (replace_with_zero):
+            cloned[0][f_zero:mask_end] = 0
+        else:
+            cloned[0][f_zero:mask_end] = cloned.mean()
+    return cloned.squeeze(0)
+
+
+def time_mask(spec, T=40, num_masks=1, replace_with_zero=False):
```
kan-bayashi: Same comments.

bobchennan:
Updated.

sw005320:
@bobchennan could you fix the conflicts as well?

sw005320:
Also, better to use a branch instead of master.

bobchennan:
The conflicts come from difference between 0.4 and master. In 0.4 complex numbers are used.

Which branch should I submit instead?

sw005320:
> Which branch should I submit instead?

0.4.0, please.
We put new functions on 0.4.0. Master is used as a stable version + bug fix.

ShigekiKarita:
If your impl is translated from existing tensorflow impl, you should add links to the original impl in module-level docstring. 

In addition, I think you can add tests to confirm your impl has the approx same results with TF impl (hard coded) except for the random op.

ShigekiKarita:
If there is no test, we cannot notice breaking changes when we upgrade pytorch to latest.

espnet/utils/SparseImageWarp.py
```diff
@@ -0,0 +1,324 @@
+import torch
```
ShigekiKarita: Can you rename this to `espnet.utils.sparse_image_warp`? (snake_case)

espnet/utils/SparseImageWarp.py
```diff
@@ -0,0 +1,324 @@
...
+        alphas.append(alpha)
+
+    flattened_grid = torch.reshape(grid, [batch_size * height * width, channels])
+    batch_offsets = torch.reshape(torch.arange(batch_size, device=grid_device) * height * width, [batch_size, 1])
+
+    # This wraps array_ops.gather. We reshape the image data such that the
+    # batch, y, and x coordinates are pulled into the first dimension.
+    # Then we gather. Finally, we reshape the output back. It's possible this
+    # code would be made simpler by using array_ops.gather_nd.
+    def gather(y_coords, x_coords, name):
```
ShigekiKarita: this name is unused. remove it.

espnet/asr/pytorch_backend/asr.py
```diff
@@ -59,6 +61,87 @@
...
+        mask_end = random.randrange(t_zero, t_zero + t)
+        if (replace_with_zero):
+            cloned[0][:, t_zero:mask_end] = 0
+        else:
+            cloned[0][:, t_zero:mask_end] = cloned.mean()
+    return cloned.squeeze(0)
+
+
+def specaug(spec):
+    """SpecAug implementation from https://github.com/zcaceres/spec_augment
```
ShigekiKarita: Move these functions to a new individual module and add https://github.com/zcaceres/spec_augment/blob/master/LICENSE to the module-level docstring

ShigekiKarita:
> In addition, I think you can add tests to confirm your impl has the approx same results with TF impl (hard coded) except for the random op.

maybe you can reuse this https://github.com/tensorflow/tensorflow/blob/r1.9/tensorflow/contrib/image/python/kernel_tests/sparse_image_warp_test.py

espnet/asr/pytorch_backend/asr.py
```diff
@@ -202,8 +287,10 @@ def __call__(self, batch, device):
             # because torch.nn.DataParellel can't handle it.
             xs_pad = {'real': xs_pad_real, 'imag': xs_pad_imag}
         else:
-            xs_pad = pad_list(
-                [torch.from_numpy(x).float() for x in xs], 0).to(device)
+            if evaluation:
+                xs_pad = pad_list([torch.from_numpy(x).float() for x in xs], 0).to(device)
+            else:
+                xs_pad = pad_list([specaug(torch.from_numpy(x).float()) for x in xs], 0).to(device)
```
ShigekiKarita: I think specaug(...) supports GPU tensor. How about this
```python
[specaug(torch.from_numpy(x).float().to(device)) for x in xs]
```
Or you can put `specaug` inside E2E (model) because E2E know it is training or evaluation and which device to use. It might be cleaner.

bobchennan: To make it simple so far it didn't support masking. That is why I put it here so it can handle each sample individually.

ShigekiKarita:
OK. We can refactor it later anyway.

2019年5月22日(水) 8:59 cnx <notifications@github.com>:

kan-bayashi:
I made PR (https://github.com/bobchennan/espnet/pull/2/files).
Please check it.

espnet/utils/spec_augment.py
```diff
@@ -0,0 +1,456 @@
...
+    spec = spec.unsqueeze(0)
+    spec_len = spec.shape[1]
+    num_rows = spec.shape[2]
+    device = spec.device
+
+    y = num_rows // 2
+    horizontal_line_at_ctr = spec[0, :, y]
+    assert len(horizontal_line_at_ctr) == spec_len
+
+    point_to_warp = horizontal_line_at_ctr[random.randrange(W, spec_len - W)]
```
bobchennan: I am not so sure about this part. It looks very strange because point_to_warp is a value randomly selected from center line but it looks like a coordinate is needed (src and dest pts). Someone actually asked in [original repo](https://github.com/zcaceres/spec_augment/issues/3).

Without understand of "warping" operation I cannot make a conclusion now.

bobchennan: According to specaug paper time warping is not that important so maybe we can skip this part right now. Any comments are welcome.

ShigekiKarita: I have not read the details of the paper and the notebook yet. If the time warping is similar to speed perturbation, some tasks (CSJ, tedlium, etc) might prefer it, and the other (librispeech?) might not.

ShigekiKarita: After I read the paper and `sparse_image_warp`, I suspect that it could be much simpler using Pillow (torchvision). It is getting faster (32ms -> 2ms) and more stable (I found the ref impl got warning sometimes).

See `time_warp_pil` in https://gist.github.com/ShigekiKarita/d6f6a77c4c9b929b3a5fdb10ed1a70a9

ShigekiKarita: I feel that the time warping is similar to [RandomResizedCrop](https://pytorch.org/docs/stable/torchvision/transforms.html#torchvision.transforms.RandomResizedCrop) rather than speed perturbation. It is popular data augumentation in ImageNet.

ShigekiKarita: That tilted spectrogram in the ref impl is already reported but I do not think this is the problem of window size. https://github.com/zcaceres/spec_augment/issues/5

sw005320:
@bobchennan, @ShigekiKarita, @kan-bayashi, thanks a lot.
I merged it into the master.
We may think of integrating it to the transform module later.

espnet/utils/spec_augment.py
```diff
@@ -0,0 +1,456 @@
...
+    cloned = spec.unsqueeze(0).clone()
+    num_mel_channels = cloned.shape[2]
+
+    for i in range(0, num_masks):
+        f = random.randrange(0, F)
+        f_zero = random.randrange(0, num_mel_channels - f)
+
+        # avoids randrange error if values are equal and range is empty
+        if (f_zero == f_zero + f):
+            return cloned.squeeze(0)
```
ShigekiKarita: I think this is wrong. instead of return this, do `continue`

espnet/utils/spec_augment.py
```diff
@@ -0,0 +1,456 @@
...
+    cloned = spec.unsqueeze(0).clone()
+    len_spectro = cloned.shape[1]
+
+    for i in range(0, num_masks):
+        t = random.randrange(0, T)
+        t_zero = random.randrange(0, len_spectro - t)
+
+        # avoids randrange error if values are equal and range is empty
+        if (t_zero == t_zero + t):
+            return cloned.squeeze(0)
```
ShigekiKarita: ditto

ShigekiKarita:
@bobchennan could you make `run.sh`?

sw005320:
Oh, I did not check it carefully. Yes, please make `run.sh` for this.

ShigekiKarita:
I think we should wait for @kan-bayashi confirmed this in WSJ or the other corpora because this paper used different specaug setting (see table 1) https://arxiv.org/pdf/1904.08779.pdf Maybe `asr_train.py` requires `--specaug-warp` (W) `--specaug-mask-freq` (F) `--specaug-mask-time` (T) instead of just `--use-specaug`

sw005320:
OK, but it's a development branch and we don't have to revert it. I agree that if we have some options to handle W, F, T, which is great. "--specaug" to accept a string of "WFT" is fine instead of three options

ShigekiKarita:
OK. Maybe someone @kan-bayashi, @bobchennan or I can send improved PR

espnet/bin/asr_train.py
```diff
@@ -159,6 +159,8 @@ def main(cmd_args):
                         help='Number of processes of iterator')
     parser.add_argument('--preprocess-conf', type=str, default=None,
                         help='The configuration file for the pre-processing')
+    parser.add_argument('--use-specaug', type=strtobool, default=True,
```
ShigekiKarita: I think default should be `False` at this time (please fix this in next PR @kan-bayashi @bobchennan  )

kan-bayashi:
> Yes, please make run.sh for this.

Should we wait for #722?
I think If we use hierarchical options in yaml config, we cam make it clean.
e.g. `specaug.use` `specaug.T` `specaug.W`

ShigekiKarita:
That is cool. I will discuss it with him today

2019年5月23日(木) 9:32 Tomoki Hayashi <notifications@github.com>:

PR #737, opened 2019-05-21
Add CSJ transformer recipe

kan-bayashi:


egs/csj/asr2/path.sh
```diff
@@ -0,0 +1,33 @@
...
+export OMP_NUM_THREADS=1
+
+# check extra module installation
+if ! which nkf > /dev/null; then
+    echo "Error: it seems that nkf is not installed." >&2
+    echo "Error: please install nkf as follows." >&2
+    echo "Error: cd ${MAIN_ROOT}/tools && make nkf.done" >&2
+    return 1
+fi
+if ! which spm_decode > /dev/null; then
```
sw005320: `spm_decode` is not used in this recipe. We can remove these lines.

sw005320:
Thanks a lot!

PR #745, opened 2019-05-23
Implement specaug as preprocessing and unify egs/wsj/asr1 and asr2

ShigekiKarita:
#734


kan-bayashi:
Thanks Shigeki.
I think now your specaug is numpy array operation.
So it can be integrated with @kamo-naoyuki’s preprocess-conf.
Why don’t you do it?

ShigekiKarita:
Cool.  My impl is still buggy and wip. Let me summarize my thoughts

- @bobchennan impl pros: differentiable w.r.t. fbank (maybe?), similar to TF in terms of sparse_image_warp reimpl (see this discussion) https://github.com/mozilla/DeepSpeech/pull/2090/files#r281919349
- @bobchennan impl cons: slow in Transformer (4.0 iter/sec -> 3.3 iter/sec), maybe RNN doesn't matter
- My impl pros: faster in Transformer (4.0 iter/sec -> 4.0 iter/sec)
- My impl cons: non differentiable w.r.t. fbank, only similar to TF in terms of 2 order spline interpolation
- Integration with preprocess (transform)? I'm not sure because currently the transform does not care it is training or evaluation. Maybe we need big change there.

kan-bayashi:
> Integration with preprocess (transform)? I'm not sure because currently the transform does not care it is training or evaluation. Maybe we need big change there.

https://github.com/espnet/espnet/blob/2f81e52d9a8d523a414289f59e3f9295f5ddbd12/espnet/asr/pytorch_backend/asr.py#L326-L333
I think there are already train flag options.

ShigekiKarita:
I have not know that great! I will work on that.

ShigekiKarita:
So I will create something like 
```python
class SpecAugment(TransformBase):
    """Spec Augument
    apply random time warping and time/freq masking
    default setting is based on LD (Librispeech double) in Table 2 https://arxiv.org/pdf/1904.08779.pdf

    :param str resize_mode: "PIL" (fast, nondifferentiable) or "sparse_image_warp" (slow, differentiable)
    :param int max_time_warp: maximum frames to warp the center frame in spectrogram (W)
    :param int max_freq_mask: maximum width of the random freq mask (F)
    :param int n_freq_mask: the number of the random freq mask (m_F)
    :param int max_time_mask: maximum width of the random time mask (T)
    :param int n_time_mask: the number of the random time mask (m_T)
    """
    def __init__(
        self, resize_mode="PIL", max_time_warp=80, 
        max_freq_mask=27, n_freq_mask=2,
        max_time_mask=100, n_time_mask=2):
```

kan-bayashi:
Nice :)

ShigekiKarita:
This is tensorboard plot. Y-axis is train or valid acc. X-axis is time (hour). Blue line is the base recipe. Red line is my impl specaug. Cyan line is prev impl specaug.
At this time, my impl seems to have less overhead. Anyway convergence and acc does not get slow a lot in any cases. Maybe I need to tweak something from RNN config?
image

image

ShigekiKarita:
Hmm, transformer + specaug results worse than baseline. Just a moment for @bobchennan impl because it is still running.

|recipe|WER dev93 |WER eval92|
|:---|--:|--:|
|Transformer (run_pytorch.sh)|8.2|4.8|
|+specaug LD setting (run_pytorch_specaug_new.sh)|8.2|5.5|
|+specaug @bobchennan  setting (run_pytorch_specaug_new.sh)|wip|wip|
|+specaug @bobchennan  setting (run_pytorch_specaug_prev.sh)|wip|wip|

ShigekiKarita:
After refactoring some `espnet.transform` modules and Transformer recipe, I'm considering the name "transform" is very annoying. In fact, sometimes I get confused to recognize "transform" or "Transformer" 🤒 

I suspect that `espnet.transform` is derived from `torchvision`. Maybe we can rename this to something like
1. espnet.preprocess (this is most relevant to `--preprocess-conf` arg)
2. espnet.signal (like scipy?)
3. espnet.feat (like kaldi?)
4. espnet.frontend

Which one is the best? Welcome to suggest new better name.

kamo-naoyuki:
I don't care much about this confusing, but how about `espnet.processings`? It's not necessary  to be `pre`.

ShigekiKarita:
In my opinion, at least `espnet.transform` and `--preprocess-conf` should be met. I mean
- If we keep `espnet.transform`, we should change the cmd arg to `--transform-conf`
- If we keep `--preprocess-conf`, we should change the module name to `espnet.preprocess`

ShigekiKarita:
> It's not necessary to be pre.

Hmm. I'm not sure what you mean. I expect this `transform` module is mainly used inside data loader/iterator. It should be data "pre"paration i.e., preprocessing. It is true?

kan-bayashi:
I agree with `espnet.preprocess`.

ShigekiKarita:
I'm now testing @kamo-naoyuki's YAML config. This is amazing. In fact, we can merge asr1 and asr2 easily because their differences are only the YAML and the model averaging stage. How do you think about it?

kamo-naoyuki:
I'm not much particular about the naming of this module, basically.

I named it as `transform` instead of `preprocessing` because these modules are just a set of transformation functions itself and they are not necessary to be applied before main processing.
For example, someone might want to apply it after `asr_enhance.py`. I didn't want to limit the naming for future development.

ShigekiKarita:
gotcha

ShigekiKarita:
I updated the Transformer result in https://github.com/espnet/espnet/pull/745#issuecomment-495188035
Maybe we need more tweaks there but I'm go back the `asr1` recipe to check that RNN+specaug can reproduce @bobchennan result.

egs/wsj/asr2/conf/preprocess.json
```diff
@@ -0,0 +1,7 @@
+{
```
sw005320: Do we need this file?

ShigekiKarita:
For differentiable specaug, we can use torch.nn.interpolate in torch 1.1.0 https://github.com/pytorch/pytorch/pull/9849

ShigekiKarita:
Hmm I cannot reproduce the result https://github.com/espnet/espnet/pull/734#issue-279949861

| | dev CER | dev WER | eval CER | eval WER |
|:--|--:|--:|--:|--:|
|RNN baseline| 3.9 |8.6 | 2.8 | 5.1 | 
|RNN specaug(sparse_image_warp @bobchennan) | 3.9 | 8.6 | 2.8 | 6.1|
|RNN specaug(PIL resize) | 3.9 | 8.6 | 2.8 | 6.1|

ShigekiKarita:
I used default values in @bobchennan's specaug. Was it tweaked from that?

egs/wsj/asr2/conf/preprocess.json
ShigekiKarita: No. I will remove this. I just wanna confirmed our JSON is compatible to YAML format.

ShigekiKarita:
When we compare Transformer and RNN results above, I made unified `run.sh` as discussed at https://github.com/espnet/espnet/pull/758 with @sw005320 @kan-bayashi @kamo-naoyuki

ShigekiKarita:
In fact, this unification is great when we search a new method over RNN and Transformer
```bash
for net in transformer rnn; do
  for pre in no_preprocess specaug; do
    # I'm using slurm. Do not try this wihout such a job system.
    ./run.sh --ngpu 1 --stage 4 \
      --train-config conf/tuning/train_${net}.yaml \
      --decode-config conf/tuning/decode_${net}.yaml \
      --preprocess-config conf/${pre}.yaml &
  done
done
wait
```
In addition, you can edit params in yaml with `utils/change_yaml.py` inside this script. COOL.

bobchennan:
Sorry for the late reply. I am in vacation these during these two week.

You should be able to reproduce my results with commits before [this](https://github.com/espnet/espnet/pull/734/commits/68c0351566eb2257796ae55ec5b07566438f63ed). In this commit I fixed a bug in previous implementation: specaug implementation we used takes 1*feature_dim*feature_len as input but we usually use 1*feature_len*feature_dim. So I tried to fix this problem in this commit however I didn't verify results of this commit.

Please change it back and test it.

ShigekiKarita:
Thanks @bobchennan! I will check that.

BTW, I have almost no time to check the detail and to work on the reproduction this week. Maybe it is better to merge this PR for unified `asr1` and draft preprocessing specaug. I rework that in the later PR.

sw005320:
OK for me. @bobchennan, do you have time to review this PR?

sw005320:
I'll merge it, but please make sure to reproduce @bobchennan's results with later PRs.

egs/wsj/asr1/run.sh
```diff
@@ -21,6 +21,8 @@ seed=1
 # feature configuration
 do_delta=false
 
+# config files
+preprocess_config=conf/no_preprocess.yaml  # use conf/specaug.yaml for data augmentation
```
jiayu2B: Is the comment means if I change this line to preprocess_config=conf/specaug.yaml, The spec_augment module would work? If not, how can I use specaug function?

bobchennan: check #830

ShigekiKarita: @jiayu2B yes

PR #690, opened 2019-04-21
Pytorch transformer (take 2)

ShigekiKarita:
this is rework of #555 with upstream chainer implementation #655 on master

- [x] split modules as previous discussion in #555 #655
- [x] add docstrings
- [x] update asr train and recog with transformer
- [x] implement dynamic module loading like preprocessing transform module (also in chainer)
- [x] CTC/LM joint decoding
- [x] constent `ASRInterface` for all the E2E implementation
- [x] add pytorch exp and RESULTS (maybe finish tomorrow)

PR #746, opened 2019-05-23
TTS modularize

kan-bayashi:
This PR modularizes TTS components.

- add `TTSInterface`
- move each modules of tacotron2 into `espnet.nets.pytorch_backend.tacotron2`
- unified interface of TTS model (now `forward` method return loss value as the same as asr)

After this PR, I will create Transformer TTS model.


kan-bayashi:
Now it become clean and unified!

kan-bayashi:
Done. I will move on TTS Transformer in the next PR.

espnet/tts/pytorch_backend/tts.py
```diff
@@ -223,23 +221,21 @@ def train(args):
         logging.info('ARGS: ' + key + ': ' + str(vars(args)[key]))
 
     # specify model architecture
-    tacotron2 = Tacotron2(idim, odim, args)
-    logging.info(tacotron2)
+    model_class = dynamic_import(args.model_module)
```
ShigekiKarita: Assert what you expect the `model_class` to implement whenever you use `dynamic_import` because pycharm or some static analysis tool support that. I'm planning to update the `CONTRIBUTING.md`.

See https://github.com/espnet/espnet/blob/a8e18ac15396986143404e4f339c818a5a57bdc2/espnet/asr/pytorch_backend/asr.py#L255

ShigekiKarita: I mean `assert isinstance(model, TTSInterface) ` here

espnet/tts/pytorch_backend/tts.py
```diff
@@ -202,7 +200,7 @@ def train(args):
     # reverse input and output dimension
     idim = int(valid_json[utts[0]]['output'][0]['shape'][1])
     odim = int(valid_json[utts[0]]['input'][0]['shape'][1])
-    if args.use_cbhg:
+    if args.use_cbhg if hasattr(args, "use_cbhg") else False:
```
ShigekiKarita: I do not like this. We can capsule these `use_cbhg` blocks in `TTSInterface`. How about this?
```python
class TTSInterface:
    ...
    def setup_dim(self, spc_dim, spk_embed_dim):
        """setup data dependent dim

        :param int spc_dim: target spectrum dim
        :param int spk_embed_dim: input speaker embedding dim
        """
        pass

    @property
    def use_second_target(self):
        """this property is called to instantiate LoadInputsAndTargets

        :rtype bool:
        """
        return False

    def register_plot_keys(self, plot_keys):
        """register keys to plot during training. keys should match what `chainer.reporter` reports

        :param list[str] plot_keys: keys to plot during training
        """
        pass
```

```python
class Tacotron2(TTSInterface, torch.nn.Module):
    ...
    def setup_dim(self, spc_dim, spk_embed_dim):
        self.spk_embed_dim = spk_embed_dim
        if self.use_cbhg:
            self.spc_dim = args.spc_dim

    @property
    def use_second_target(self):
        return False

    def register_plot_keys(self, plot_keys):
        plot_keys += ['main/cbhg_l1_loss', 'validation/main/cbhg_l1_loss',
                              'main/cbhg_mse_loss', 'validation/main/cbhg_mse_loss']
```

espnet/tts/pytorch_backend/tts.py
```diff
@@ -397,16 +395,18 @@ def decode(args):
         logging.info('args: ' + key + ': ' + str(vars(args)[key]))
 
     # define model
-    tacotron2 = Tacotron2(idim, odim, train_args)
+    model_class = dynamic_import(train_args.model_module)
```
ShigekiKarita: assert isinstance(model_class, TTSInterface)

espnet/bin/tts_train.py
```diff
@@ -164,7 +90,13 @@ def main(args):
                         help='Gradient norm threshold to clip')
     parser.add_argument('--num-save-attention', default=5, type=int,
                         help='Number of samples of attention to be saved')
-    args = parser.parse_args(args)
+
+    args, _ = parser.parse_known_args(cmd_args)
+
+    from espnet.utils.dynamic_import import dynamic_import
+    model_class = dynamic_import(args.model_module)
```
ShigekiKarita: ditto

espnet/tts/pytorch_backend/tts.py
ShigekiKarita: BTW, how about split `train` and `decode` into separated files e.g., `tts_train.py`, `tts_decode.py`. I think  the function`decode` does not depend on many of `espnet.tts.pytorch_backend_tts.py`

ShigekiKarita:
This is great. Especially, this is a fine play https://github.com/espnet/espnet/pull/746/files#diff-22acf0acda9e4fc1f05638b07a29baaaL55
However, we can be greater with little effort. See my comments.

espnet/tts/pytorch_backend/tts.py
ShigekiKarita: For `set_dim(...)`, it is better to set these args in `run.sh` or `conf/train_tacotron.yaml` if we can.

kan-bayashi: I decide to use `use_second_target` as argment because if we set it as model property, it becomes dirty due to the `DataParallel`.

espnet/tts/pytorch_backend/tts.py
kan-bayashi: Sorry, I could not catch up your meaning.
`tts_train.py` and `tts_decode.py` are already separated.

kan-bayashi: Wait, model_class is not instance, so this `assert isinstance(model_class, TTSInterface)` is failed.
We do not have to check it.

espnet/tts/pytorch_backend/tts.py
ShigekiKarita: cool

espnet/tts/pytorch_backend/tts.py
ShigekiKarita: > 
> 
> Sorry, I could not catch up your meaning.
> `tts_train.py` and `tts_decode.py` are already separated.

Sorry. I mean split `espnet/tts/pytorch_backend/tts.py` into `espnet/tts/pytorch_backend/tts_{train, decode}.py` not `espnet/bin/tts_{train, decode}.py`

ShigekiKarita: > Wait, model_class is not instance, so this `assert isinstance(model_class, TTSInterface)` is failed.
> We do not have to check it.

You are right. Please assert this very next line.
```python
model_class = dynamic_import(args.model_module)
model = model_class(idim, odim, args)
assert isinstance(model, TTSInterface)
```

espnet/tts/pytorch_backend/tts.py
```diff
@@ -223,23 +221,21 @@ def train(args):
         logging.info('ARGS: ' + key + ': ' + str(vars(args)[key]))
 
     # specify model architecture
-    tacotron2 = Tacotron2(idim, odim, args)
-    logging.info(tacotron2)
+    model_class = dynamic_import(args.model_module)
+    model = model_class(idim, odim, args)
```
ShigekiKarita: `assert isinstance(model, TTSInterface)` here

espnet/tts/pytorch_backend/tts.py
ShigekiKarita: @kamo-naoyuki told me to use `issubclass` instead of `isinstance`!

espnet/tts/pytorch_backend/tts.py
kan-bayashi: Already added.

espnet/bin/tts_train.py
ShigekiKarita: `assert issubclass(model_class, TTSInterface)`

kan-bayashi: Added.

espnet/bin/tts_train.py
```diff
@@ -164,7 +94,14 @@ def main(args):
...
+    issubclass(model_class, TTSInterface)
```
ShigekiKarita: please `assert`

espnet/nets/pytorch_backend/e2e_tts.py
```diff
@@ -235,10 +171,96 @@ class Tacotron2(torch.nn.Module):
...
+        # loss related
+        parser.add_argument('--use_masking', default=False, type=strtobool,
+                            help='Whether to use masking in calculation of loss')
+        parser.add_argument('--bce_pos_weight', default=20.0, type=float,
+                            help='Positive sample weight in BCE calculation (only for use_masking=True)')
+        return
+
     def __init__(self, idim, odim, args):
+        # initialize base classes
         super(Tacotron2, self).__init__()
```
ShigekiKarita: Use `TTSInterface.__init__(self)`  because C4 algorithm is not popular for many python user.  Namely, I am not sure the `super` is `TTSInterface.__init__` or `nn.Module.__init__`.

espnet/bin/tts_train.py
kan-bayashi: Opps, sorry easy typo :(

ShigekiKarita:
LGTM!! Thanks for the various fix!!

espnet/nets/pytorch_backend/e2e_tts.py
kan-bayashi: Actually, I’ve never use multiple inheritance so which way is correct to initialize.
I followed your manner. Thanks.

kan-bayashi:
Thank you for your review, @ShigekiKarita!

r9y9:
I'm not quite familiar with the design of espnet yet, but my first honest impression for the TTS code is that it looks like too Tacotron 2 specific. If I understand correctly, the PR tries to reorganize the structure and moves some Tacotron specific things into separate modules so that we can add different models without changing the structure, but there are still Tacotron 2 specific code; e.g. in https://github.com/kan-bayashi/espnet/blob/dd970487df7f1b26058ea01b8f602903f6247fca/espnet/nets/pytorch_backend/e2e_tts.py, that I think (ideally) it should be applicable for any end-to-end models (Transformer, DeepVoice, etc) as the name suggests, right?

kan-bayashi:
Thank you for your comments, @r9y9.
Yes, actually I wander to rename `e2e_tts` to `e2e_tts_tacotron2`.
In the next PR, I will rename to `e2e_tts_tacotron2` and create `e2e_tts_transformer`.

ShigekiKarita:
It reminds me of `e2e_asr` that I should have renamed it to `e2e_asr_rnn` in #690

sw005320:
thanks. please take care of #712. Refactering affects other projects.

Issue #748, opened 2019-05-23
Decoding is slow


takenori-y:
In my environment, decoding requires much computational time. Given an input signal whose duration was about 15 seconds, the RTFs in the encoder and the decoder were larger than 0.2 and 3.0, respectively.

By increasing the value of OMP_NUM_THREADS in `path.sh`, the RTF in the encoder was improved twice. However, any improvements could not be observed in the decoder.

I investigated the cause of the slow decoding using cProfile and line_profiler. The main bottlenecks were calculating the context vectors (45%) and the CTC prefix scores (25%) as shown below. Can we eliminate these bottlenecks?

```
Total time: 52.0387 s
/work4/yoshimura/work/00/espnet/espnet/nets/pytorch_backend/decoders.py
Function: recognize_beam at line 214

Line #      Hits         Time  Per Hit   % Time  Line Contents
==============================================================
   214                                               def recognize_beam(self, h, lpz, recog_args, char_list, rnnlm=None, strm_idx=0):
   215                                                   """beam search implementation
   216                                           
   217                                                   :param torch.Tensor h: encoder hidden state (T, eprojs)
   218                                                   :param torch.Tensor lpz: ctc log softmax output (T, odim)
   219                                                   :param Namespace recog_args: argument Namespace containing options
   220                                                   :param char_list: list of character strings
   221                                                   :param torch.nn.Module rnnlm: language module
   222                                                   :param int strm_idx: stream index for speaker parallel attention in multi-speaker case
   223                                                   :return: N-best decoding results
   224                                                   :rtype: list of dicts
   225                                                   """
   226         1        374.0    374.0      0.0          logging.info('input lengths: ' + str(h.size(0)))
   227         1         31.0     31.0      0.0          att_idx = min(strm_idx, len(self.att) - 1)
   228                                                   # initialization
   229         1         33.0     33.0      0.0          c_list = [self.zero_state(h.unsqueeze(0))]
   230         1         11.0     11.0      0.0          z_list = [self.zero_state(h.unsqueeze(0))]
   231         1          8.0      8.0      0.0          for _ in six.moves.range(1, self.dlayers):
   232                                                       c_list.append(self.zero_state(h.unsqueeze(0)))
   233                                                       z_list.append(self.zero_state(h.unsqueeze(0)))
   234         1          1.0      1.0      0.0          a = None
   235         1         82.0     82.0      0.0          self.att[att_idx].reset()  # reset pre-computation of h
   236                                           
[...]

PR #752, opened 2019-05-24
Transformer TTS

kan-bayashi:
I will implement Transformer TTS model.

Reference: https://arxiv.org/pdf/1809.08895.pdf


ShigekiKarita:
Maybe decoding (speech generation) results really very slow or out-of-memory because I was too lazy to implement [incremental forwarding with state like fairseq](https://github.com/pytorch/fairseq/blob/master/fairseq/modules/multihead_attention.py#L81). Let me know if you need serious help there. I will work on that.

kan-bayashi:
Maybe `forward` method is done.
Next, I will implement `inference`.

kan-bayashi:
Almost done. I will make recipe tomorrow.

ShigekiKarita:
check my PR https://github.com/kan-bayashi/espnet/pull/2
- implement the scaled pos enc with trainable weight in 3.2 https://arxiv.org/pdf/1809.08895.pdf
- rename `model.encoder.input_layer` to `model.encoder.embed` like `model.decoder.embed`
- rename `test/test_e2e_transformer.py` to `test/test_e2e_asr_transformer.py`

ShigekiKarita:
In addition, I suggest further tiny changes if you do not mind
- use `-` instead of `_` in the argparse options. those are mixed now e.g., `--use_mask`, `--transformer-lr` but you need to rewrite all the existing tts recipe. it might be tough
- split `test/test_e2e_tts.py` into `test/test_tts_tacotron2.py`  and `test/test_tts_transformer.py`

r9y9:
Cool! BTW I'm on vacation till next Wed. I will take a closer look later next week.

ShigekiKarita:
have a nice vacation :tropical_drink:

kan-bayashi:
Thanks, Shigeki!
I will reflect your comments.

kan-bayashi:
Ready to learn :)

Memo: I have not checked in detail following parts
- masked multi-head-attention (`self._target_mask`)
- initialization of additional components (`Prenet` or `Postnet`)

espnet/nets/pytorch_backend/e2e_tts_transformer.py
```diff
@@ -0,0 +1,443 @@
...
+            ys = ys.masked_select(mask)
+            outs = outs.masked_select(mask)
+            labels = labels.masked_select(mask[:, :, 0])
+            logits = logits.masked_select(mask[:, :, 0])
+            weights = weights.masked_select(mask[:, :, 0]) if weights is not None else None
+
+        # calculate loss
+        l1_loss = F.l1_loss(outs, ys)
+        mse_loss = F.mse_loss(outs, ys)
+        bce_loss = F.binary_cross_entropy_with_logits(logits, labels, weights)
```
ShigekiKarita:
```python
bce_loss = F.binary_cross_entropy_with_logits(logits, labels, pos_weight=self.bce_pos_weight)
```

kan-bayashi: Wow, I did not know this option.
That is nice.

kan-bayashi:
I’ve started training with LJspeech recipe.

```
# tts_train.py --backend pytorch --ngpu 2 --minibatches 0 --outdir exp/train_no_dev_pytorch_train_transformer/results --tensorboard-dir tensorboard/train_no_dev_pytorch_train_t
ransformer --verbose 0 --seed 1 --resume --train-json dump/train_no_dev/data.json --valid-json dump/dev/data.json --config conf/train_transformer.yaml
/home/t_hayashi/workspace5/work/espnet_repositories/espnet_tts_refactor/tools/venv/lib/python3.7/site-packages/torch/nn/_reduction.py:49: UserWarning: size_average and reduce a
rgs will be deprecated, please use reduction='none' instead.
  warnings.warn(warning.format(ret))
2019-05-25 14:09:57,241 (tts_train:117) WARNING: Skip DEBUG/INFO messages
epoch       iteration   elapsed_time  main/loss   validation/main/loss  main/l1_loss  validation/main/l1_loss  main/mse_loss  validation/main/mse_loss  main/bce_loss  validatio
n/main/bce_loss
0           100         28.2606       5.83897                           1.965                                  3.0186                                   0.855371                
     total [..................................................]  0.06%
this epoch [######............................................] 12.69%
       100 iter, 0 epoch / 200 epochs
       inf iters/sec. Estimated time to finish: 0:00:00.
0           200         52.6109       5.15581                           1.9476                                 2.96401                                  0.244193                
     total [..................................................]  0.13%
this epoch [############......................................] 25.38%
       200 iter, 0 epoch / 200 epochs
     4.107 iters/sec. Estimated time to finish: 10:38:44.349509.
0           300         76.66         4.79332                           1.90481                                2.83016                                  0.0583452               
     total [..................................................]  0.19%
this epoch [###################...............................] 38.07%
       300 iter, 0 epoch / 200 epochs
    4.1323 iters/sec. Estimated time to finish: 10:34:25.604551.
0           400         100.996       4.65779                           1.87469                                2.73437                                  0.0487314               
     total [..................................................]  0.25%
this epoch [#########################.........................] 50.76%
       400 iter, 0 epoch / 200 epochs
    4.1246 iters/sec. Estimated time to finish: 10:35:12.583376.
0           500         124.723       4.54185                           1.8484                                 2.64678                                  0.0466687 
```

ShigekiKarita:
https://arxiv.org/pdf/1809.08895.pdf  In this paper, it took 0.4 sec/iter = 2.5 iter/sec on 4 Tesla P100s. Our impl seems to run 4.1 iter/sec!? Amazing. What kind of GPUs are you using?

ShigekiKarita:
https://arxiv.org/pdf/1809.08895.pdf  In this paper, it took 0.4 sec/iter = 2.5 iter/sec on 4 Tesla P100s (16 * 4 samples/iter). Our impl seems to run 4.1 iter/sec (8 samples on 1GPU?).

Maybe we can make this faster. Is `8 samples` maximum value on 1GPU?

kan-bayashi:
After few epochs, the average become slow :(

```
     total [############......................................] 25.13%
this epoch [############......................................] 25.38%
     39600 iter, 50 epoch / 200 epochs
    2.6614 iters/sec. Estimated time to finish: 12:18:57.546793.
50          39700       14819.3       1.38636                           0.755031                               0.626599                                 0.0047292               
     total [############......................................] 25.19%
this epoch [###################...............................] 38.07%
     39700 iter, 50 epoch / 200 epochs
    2.6614 iters/sec. Estimated time to finish: 12:18:19.541358.
50          39800       14843.6       1.38872                           0.755943                               0.627776                                 0.00500263              
     total [############......................................] 25.25%
this epoch [#########################.........................] 50.76%
     39800 iter, 50 epoch / 200 epochs
    2.6615 iters/sec. Estimated time to finish: 12:17:40.139910.
50          39900       14867.9       1.38568                           0.754664                               0.626146                                 0.00486764              
     total [############......................................] 25.32%
this epoch [###############################...................] 63.45%
     39900 iter, 50 epoch / 200 epochs
    2.6615 iters/sec. Estimated time to finish: 12:17:03.336982.
50          40000       14892.3       1.3917                            0.757903                               0.628834                                 0.00495798              
     total [############......................................] 25.38%
this epoch [######################################............] 76.14%
     40000 iter, 50 epoch / 200 epochs
     2.735 iters/sec. Estimated time to finish: 11:56:37.543751.
```

But I found that current setting is much bigger than papers (ours elayers=12, in paper 6).
So the speed can be a little bit faster.

kan-bayashi:
I tested with the same setting as in the paper with titan V.
`elayers=6` `dlayers=6` `batch-size=16` `accum-grad=4`
(I’m not sure the other transformer parameters.)
It looks faster than paper.

```
epoch       iteration   elapsed_time  main/loss   validation/main/loss  main/l1_loss  validation/main/l1_loss  main/bce_loss  validation/main/bce_loss
0           100         30.1411       2.47006                           1.9643                                 0.505752
     total [..................................................]  0.06%
this epoch [######............................................] 12.69%
       100 iter, 0 epoch / 200 epochs
       inf iters/sec. Estimated time to finish: 0:00:00.
0           200         59.3174       1.96869                           1.90783                                0.0608525
     total [..................................................]  0.13%
this epoch [############......................................] 25.38%
       200 iter, 0 epoch / 200 epochs
    3.4275 iters/sec. Estimated time to finish: 12:45:22.243623.
0           300         88.3927       1.89828                           1.84772                                0.050566
     total [..................................................]  0.19%
this epoch [###################...............................] 38.07%
       300 iter, 0 epoch / 200 epochs
    3.4334 iters/sec. Estimated time to finish: 12:43:34.394004.
0           400         117.456       1.8109                            1.76134                                0.0495518
     total [..................................................]  0.25%
this epoch [#########################.........................] 50.76%
       400 iter, 0 epoch / 200 epochs
    3.4359 iters/sec. Estimated time to finish: 12:42:32.816664.
0           500         146.351       1.7187                            1.66965                                0.0490543
     total [..................................................]  0.32%
this epoch [###############################...................] 63.45%
       500 iter, 0 epoch / 200 epochs
     3.442 iters/sec. Estimated time to finish: 12:40:41.472178.
```

kan-bayashi:
> After few epochs, the average become slow :(

The reason of the above seems that extra processings at the end of epochs (e.g. model saving) consume much time.

```
1           800         314.46        1.40086     1.23841               1.36682       1.20751                  0.0340405      0.0308988
     total [..................................................]  0.51%
this epoch [..................................................]  1.52%
       800 iter, 1 epoch / 200 epochs
     2.462 iters/sec. Estimated time to finish: 17:41:27.310577.
1           900         343.804       1.30614                           1.27318                                0.0329583
     total [..................................................]  0.57%
this epoch [#######...........................................] 14.21%
       900 iter, 1 epoch / 200 epochs
    2.5505 iters/sec. Estimated time to finish: 17:03:58.323440.
1           1000        372.869       1.24629                           1.21456                                0.0317342
     total [..................................................]  0.63%
this epoch [#############.....................................] 26.90%
      1000 iter, 1 epoch / 200 epochs
     2.626 iters/sec. Estimated time to finish: 16:33:54.481447.
1           1100        401.882       1.21805                           1.1877                                 0.0303456
     total [..................................................]  0.70%
this epoch [###################...............................] 39.59%
      1100 iter, 1 epoch / 200 epochs
      2.69 iters/sec. Estimated time to finish: 16:09:37.408208.
1           1200        431.005       1.19919                           1.1698                                 0.0293899
     total [..................................................]  0.76%
this epoch [##########################........................] 52.28%
      1200 iter, 1 epoch / 200 epochs
    2.7441 iters/sec. Estimated time to finish: 15:49:55.782446.
1           1300        459.924       1.18546                           1.15711                                0.0283426
     total [..................................................]  0.82%
this epoch [################################..................] 64.97%
      1300 iter, 1 epoch / 200 epochs
    2.7921 iters/sec. Estimated time to finish: 15:32:59.170598.
```
Speed is increasing gradually.

kan-bayashi:
GPU-Utils keeps around 90% and requires less than 12GB gpu memory with batchsize=16 (`elayers=6` `dlayers=6`).
It looks nice.

sw005320:
I observed the same trend. Then, I actually increased the batchsize because I heard that it is better to have larger batchsize for the transformer. But not sure.

kan-bayashi:
@ShigekiKarita, Do you know any reference of how to set dropout rate?

ShigekiKarita:
When I completely removed it, Transformer got clear overfitting. When I removed dropout on attention but kept on sublayer input (a.k.a. [residual dropout](https://arxiv.org/pdf/1706.03762.pdf)), it slightly improved the ASR performance. As seen in NMT, the dropout rate is typically 0.1 sometimes 0.3 https://github.com/pytorch/fairseq/tree/master/examples/scaling_nmt

ShigekiKarita:
@sw005320 @kan-bayashi 
> I heard that it is better to have larger batchsize for the transformer. But not sure.

I made #759 to make the batchsize larger.

kan-bayashi:
Thanks, Shigeki.
I will try #759.

kan-bayashi:
I found that decoding is too slow. 
We need to improve `recognize` method to cache the previous calculation.

ShigekiKarita:
See #748 my profiling result on `recognize` in ASR. We can use `torch.nn.MultiheadAttention` in 1.1.0 that supports incremental forward with caching states. https://pytorch.org/docs/stable/nn.html#multiheadattention

kan-bayashi:
I tried `train_transformer.v1.yaml` and `train_transformer.v2.yaml` but the model could not generate utterances.
Now I'm trying to train with `train_transformer.v3.yaml`.

ShigekiKarita:
Maybe you can check that the `ScaledPositionalEncoding.alpha` in enc/dec are updated as seen in [Fig 5]( https://arxiv.org/pdf/1809.08895.pdf) by adding them  to `base_plot_keys`

ShigekiKarita:
Also you can check MSR's Transformer TTS config for LJSpeech in Sec 4.1 of this paper (in fact, it is semi-supervised ASR-TTS task) https://speechresearch.github.io/papers/almost_unsup_tts_asr_2019.pdf
They used 128 seqs for minibatch of paired data.

kan-bayashi:
> Also you can check MSR's Transformer TTS config for LJSpeech in Sec 4.1 of this paper (in fact, it is semi-supervised ASR-TTS task) https://speechresearch.github.io/papers/almost_unsup_tts_asr_2019.pdf
They used 128 seqs for minibatch of paired data.

Year, I checked it to make `v3.yaml` config.

kan-bayashi:
Some notes from https://github.com/soobinseo/Transformer-TTS
```
- The learning rate is an important parameter for training. With initial learning rate of 0.001 and exponentially decaying doesn't work.
- The gradient clipping is also an important parameter for training. I clipped the gradient with norm value 1.
- With the stop token loss, the model did not training.
- It was very important to concatenate the input and context vectors in the Attention mechanism.
```

kan-bayashi:
Loss value will bomb around 30 epochs...

warmup-steps 400, lr=10.0, grad-clip=5
l1_loss

warmup-steps 400, lr=10.0, grad-clip=1
l1_loss

warmup-steps 400, lr=10.0, grad-clip=1 + large decoder
l1_loss

ShigekiKarita:
LOL

Maybe lr is too high and warup-steps is too low.

kan-bayashi:
Small lr with many warmup-steps make the training stable.

lr = 1.0, grad-clip=1, warmup-steps=10000
l1_loss

Compared with taco2, loss value seems to be good.
l1_loss

I hope the model can learn good attention as well.

kan-bayashi:
I trained with following parameters:

```
# This is the Transformer TTS configuration

# network architecture related
model-module: espnet.nets.pytorch_backend.e2e_tts_transformer:Transformer
embed-dim: 0
eprenet-conv-layers: 0  # one more linear layer w/o non-linear will be added for 0-centor
eprenet-conv-filts: 0
eprenet-conv-chans: 0
dprenet-layers: 1  # one more linear layer w/o non-linear will be added for 0-centor
dprenet-units: 256
adim: 256
aheads: 4
elayers: 6
eunits: 1024
dlayers: 4
dunits: 1024
postnet-layers: 5
postnet-filts: 5
postnet-chans: 256
use-masking: True
bce-pos-weight: 5.0
use-batch-norm: True
use-scaled-pos-enc: True
encoder-normalize-before: False
decoder-normalize-before: False

# minibatch related
batch-size: 16
batch-sort-key: shuffle # shuffle or input or output
maxlen-in: 0   # if input length  > maxlen-in, batchsize is reduced (if use "shuffle", not effect)
maxlen-out: 0  # if output length > maxlen-out, batchsize is reduced (if use "shuffle", not effect)

# training related
transformer-init: xavier_uniform
transformer-warmup-steps: 10000
transformer-lr: 1.0
initial-encoder-alpha: 1.0
initial-decoder-alpha: 1.0
dropout-rate: 0.1
eprenet-dropout-rate: 0.1
dprenet-dropout-rate: 0.1
postnet-dropout-rate: 0.1
transformer-attn-dropout-rate: 0.1

# optimization related
opt: noam
accum-grad: 16
grad-clip: 5
weight-decay: 0.0
epochs: 400
patience: 0
```

Learning curve
スクリーンショット 2019-05-28 午後1 05 18

But generation is totally failed. Just a noise.
This is because the model cannot learn diagonal attention weights.

Some heads of decoder self attention looks diagonal.
スクリーンショット 2019-05-28 午後1 09 04

But decoder src attention is not diagonal
スクリーンショット 2019-05-28 午後1 09 16

Encoder looks overfitting...
スクリーンショット 2019-05-28 午後1 09 47

kan-bayashi:
I wander that we need to use phoneme instead of character 👀

ShigekiKarita:
In ASR Transformer, only very last encoder-to-decoder attentions (src attn) are diagonal. The first or second layer attentions result non-diagonal.

kan-bayashi:
> In ASR Transformer, only very last encoder-to-decoder attentions (src attn) are diagonal. The first or second layer attentions result non-diagonal.

Hmm. I checked all of layers but all of src attention are not diagonal.
I will try phoneme recipe with phonemizer for debugging.
https://github.com/bootphon/phonemizer

kan-bayashi:
Another possibility to get good attention is `reduction_factor`.
I will implement it.

kan-bayashi:
I updated action items.
If you have any suggestion or advice, please let me know.

kan-bayashi:
I tried phoneme text with the same setting, but the training was not succeeded.
all_loss

I tried reduction-factor=2 option with character, but the training was not succeeded.
all_loss

kan-bayashi:
I implemented many options.
I will re-check from the first config setting.

kan-bayashi:
I found that the number of iterations is much lower than paper.
It takes around 300k but my previous training 200k but due to the `accum-grad=8 or 16`,
resulting 25k or 12.5k iterations.

kan-bayashi:
This PR becomes too big.
I will separate into several PRs.

r9y9:
Nice! Looks like config changes for tacotron2 can be moved to a separate PR.

kan-bayashi:
It becomes easy to review.

r9y9:
I have tried to train a transformer for LJSpeech with the following parameter changes but got no success so far.  None of the encoder-decoder attetion weights didn't look like diaginal. I doubt there might be a bug in tts_train or somewhere. I will look into code carefully soon.

```diff
diff --git a/egs/ljspeech/tts1/conf/tuning/train_transformer.v1.yaml b/egs/ljspeech/tts1/conf/tuning/train_transformer.v1.yaml
index 425438d..787a794 100644
--- a/egs/ljspeech/tts1/conf/tuning/train_transformer.v1.yaml
+++ b/egs/ljspeech/tts1/conf/tuning/train_transformer.v1.yaml
@@ -2,7 +2,7 @@

 # network architecture related
 model-module: espnet.nets.pytorch_backend.e2e_tts_transformer:Transformer
-embed-dim: 256
+embed-dim: 128
 eprenet-conv-layers: 5  # one more linear layer w/o non-linear will be added for 0-centor
 eprenet-conv-filts: 5
 eprenet-conv-chans: 256
@@ -11,10 +11,10 @@ dprenet-units: 256
 adim: 256
 aheads: 4
 elayers: 4
-eunits: 1024
+eunits: 512
 dlayers: 4
-dunits: 1024
-postnet-layers: 5
+dunits: 512
+postnet-layers: 2
 postnet-filts: 5
 postnet-chans: 256
 use-masking: True
@@ -25,7 +25,7 @@ encoder-normalize-before: True
 decoder-normalize-before: True
 encoder-concate-after: False
 decoder-concate-after: False
-reduction-factor: 1
+reduction-factor: 4

 # minibatch related
 batch-sort-key: input # shuffle or input or output
@@ -36,7 +36,7 @@ batch-bins: 1214400   # 16 * (870 * 80 + 180 * 35)
 # training related
 transformer-init: pytorch
 transformer-warmup-steps: 4000
-transformer-lr: 1.0
+transformer-lr: 0.001
 initial-encoder-alpha: 1.0
 initial-decoder-alpha: 1.0
 dropout-rate: 0.1
@@ -47,7 +47,7 @@ transformer-attn-dropout-rate: 0.1

 # optimization related
 opt: noam
-accum-grad: 4
+accum-grad: 1
 grad-clip: 1.0
 weight-decay: 0.0
 patience: 0
```

kan-bayashi:
Hmm... 
I have checked current implementation works well with Taco2 config.
`train_pytorch_tacotron2.v1.yaml`
スクリーンショット 2019-06-01 午後6 28 00
`train_pytorch_tacotron2.v2.yaml`
スクリーンショット 2019-06-01 午後6 27 28
Only a few epochs are needed to get diagonal attention. 
Taco2 is quite nice at least in terms of easiness of training...

So maybe `tts.py` is OK, if there are some bugs, it seems that `e2e_tts_transformer.py` is doubtful.

ShigekiKarita:
@r9y9 @kan-bayashi 
When I implemented ASR Transformer, there was a bug in subsequent masking in the decoder.  The wrong mask results similar collapsed attentions in https://github.com/espnet/espnet/pull/752#issuecomment-496358523

I suggest checking your decoder masking by adding one simple test like this https://github.com/espnet/espnet/blob/94fdb49217c7d632b2ccdbd45c3e867246ada3e4/test/test_e2e_transformer.py#L100-L108
This test just checks if the NaN value in the input `[3:]` is correctly masked until the output `[:3]`.

kan-bayashi:
Thanks, @ShigekiKarita 
I will re-check the masking part.

ShigekiKarita:
However, as seen in your plot, the decoder self attn seems correctly masked in fact.

After all, I feel that the hyperparameter seriously matters in Transformer. I will double-check our hyperparameter and https://github.com/soobinseo/Transformer-TTS

kan-bayashi:
@ShigekiKarita Here, mask shape is not `(batch, time1)`
https://github.com/espnet/espnet/blob/94fdb49217c7d632b2ccdbd45c3e867246ada3e4/espnet/nets/pytorch_backend/transformer/attention.py#L38
It should be `(batch, 1, time1)` or `(batch, time1, 1)`?

ShigekiKarita:
Oh. Yes, it should be (batch, time1, 1) but it might be (batch, time1, time2). I forget the detail... :thinking:

ShigekiKarita:
https://github.com/espnet/espnet/blob/bc274f42faafdb919c9d5f70c99a2d7913782b23/espnet/nets/pytorch_backend/tacotron2/decoder.py#L84

I do not think this line is added in this PR but I believe this line should be 
```python
 x = F.dropout(self.prenet[l](x), self.dropout_rate, training=self.training) 
```
or use `torch.nn.Dropout` like somewhere else

kan-bayashi:
> or use torch.nn.Dropout like somewhere else

In the paper, they use Dropout during inference.

> In order to introduce output variation at inference time, dropout with probability 0.5 is applied only to layers in the pre-net of the autoregressive decoder.
https://arxiv.org/pdf/1712.05884.pdf

kan-bayashi:
Currently, we only apply the mask for the direction of key.
But I think it is better to apply the mask for the direction of queue.

ShigekiKarita:
> > or use torch.nn.Dropout like somewhere else
> 
> In the paper, they use Dropout during inference.
> 
> > In order to introduce output variation at inference time, dropout with probability 0.5 is applied only to layers in the pre-net of the autoregressive decoder.
> > https://arxiv.org/pdf/1712.05884.pdf

wow

kan-bayashi:
> Currently, we only apply the mask for the direction of key.
> But I think it is better to apply the mask for the direction of queue.

Current mask 
```python
(Pdb) olens = [10, 6]
(Pdb) mask1 = make_non_pad_mask(olens)
(Pdb) mask1
tensor([[1, 1, 1, 1, 1, 1, 1, 1, 1, 1],
        [1, 1, 1, 1, 1, 1, 0, 0, 0, 0]], dtype=torch.uint8)
(Pdb) mask2 = subsequent_mask(max(olens))
(Pdb) mask2
tensor([[1, 0, 0, 0, 0, 0, 0, 0, 0, 0],
        [1, 1, 0, 0, 0, 0, 0, 0, 0, 0],
        [1, 1, 1, 0, 0, 0, 0, 0, 0, 0],
        [1, 1, 1, 1, 0, 0, 0, 0, 0, 0],
        [1, 1, 1, 1, 1, 0, 0, 0, 0, 0],
        [1, 1, 1, 1, 1, 1, 0, 0, 0, 0],
        [1, 1, 1, 1, 1, 1, 1, 0, 0, 0],
        [1, 1, 1, 1, 1, 1, 1, 1, 0, 0],
        [1, 1, 1, 1, 1, 1, 1, 1, 1, 0],
        [1, 1, 1, 1, 1, 1, 1, 1, 1, 1]], dtype=torch.uint8)
(Pdb) mask = mask1.unsqueeze(-2) & mask2.unsqueeze(0)
(Pdb) mask
tensor([[[1, 0, 0, 0, 0, 0, 0, 0, 0, 0],
         [1, 1, 0, 0, 0, 0, 0, 0, 0, 0],
         [1, 1, 1, 0, 0, 0, 0, 0, 0, 0],
         [1, 1, 1, 1, 0, 0, 0, 0, 0, 0],
         [1, 1, 1, 1, 1, 0, 0, 0, 0, 0],
         [1, 1, 1, 1, 1, 1, 0, 0, 0, 0],
         [1, 1, 1, 1, 1, 1, 1, 0, 0, 0],
         [1, 1, 1, 1, 1, 1, 1, 1, 0, 0],
         [1, 1, 1, 1, 1, 1, 1, 1, 1, 0],
         [1, 1, 1, 1, 1, 1, 1, 1, 1, 1]],

        [[1, 0, 0, 0, 0, 0, 0, 0, 0, 0],
         [1, 1, 0, 0, 0, 0, 0, 0, 0, 0],
         [1, 1, 1, 0, 0, 0, 0, 0, 0, 0],
         [1, 1, 1, 1, 0, 0, 0, 0, 0, 0],
         [1, 1, 1, 1, 1, 0, 0, 0, 0, 0],
         [1, 1, 1, 1, 1, 1, 0, 0, 0, 0],
         [1, 1, 1, 1, 1, 1, 0, 0, 0, 0],
         [1, 1, 1, 1, 1, 1, 0, 0, 0, 0],
         [1, 1, 1, 1, 1, 1, 0, 0, 0, 0],
         [1, 1, 1, 1, 1, 1, 0, 0, 0, 0]]], dtype=torch.uint8)
```
And then apply this mask to score whose shape is `(#batches, #heads, T_queue, T_key)`:
```
mask = mask.unsqueeze(1)
scores = scores.masked_fill(mask == 0, MIN_VALUE)
att_ws = torch.softmax(scores, dim=-1) 
```
After that we should apply following mask to 
```
(Pdb) mask_q = mask & mask1.unsqueeze(-1)
(Pdb) mask_q
tensor([[[1, 0, 0, 0, 0, 0, 0, 0, 0, 0],
         [1, 1, 0, 0, 0, 0, 0, 0, 0, 0],
         [1, 1, 1, 0, 0, 0, 0, 0, 0, 0],
         [1, 1, 1, 1, 0, 0, 0, 0, 0, 0],
         [1, 1, 1, 1, 1, 0, 0, 0, 0, 0],
         [1, 1, 1, 1, 1, 1, 0, 0, 0, 0],
         [1, 1, 1, 1, 1, 1, 1, 0, 0, 0],
         [1, 1, 1, 1, 1, 1, 1, 1, 0, 0],
         [1, 1, 1, 1, 1, 1, 1, 1, 1, 0],
         [1, 1, 1, 1, 1, 1, 1, 1, 1, 1]],

        [[1, 0, 0, 0, 0, 0, 0, 0, 0, 0],
         [1, 1, 0, 0, 0, 0, 0, 0, 0, 0],
         [1, 1, 1, 0, 0, 0, 0, 0, 0, 0],
         [1, 1, 1, 1, 0, 0, 0, 0, 0, 0],
         [1, 1, 1, 1, 1, 0, 0, 0, 0, 0],
         [1, 1, 1, 1, 1, 1, 0, 0, 0, 0],
         [0, 0, 0, 0, 0, 0, 0, 0, 0, 0],
         [0, 0, 0, 0, 0, 0, 0, 0, 0, 0],
         [0, 0, 0, 0, 0, 0, 0, 0, 0, 0],
         [0, 0, 0, 0, 0, 0, 0, 0, 0, 0]]], dtype=torch.uint8)
```
Apply this mask to att_ws
```
mask_q = mask_q.unsqueeze(1)
self.att_ws = att_ws.masked_fill(mask_q == 0, 0)
```
I’m not sure this masking affect the training.

ShigekiKarita:
I sent you new PR that add L2 loss and fix attention plot. https://github.com/kan-bayashi/espnet/pull/3

kan-bayashi:
Thank you so much, Shigeki.
I forgot to fix the plot report.

About loss function, originally taco2 has also used l2 loss.
> We minimize the summed mean squared error (MSE) from before
and after the post-net to aid convergence.
https://arxiv.org/pdf/1712.05884.pdf

And in another paper, 
> We extend [15] by augmenting the L2 loss on the predicted spectrogram with an additional L1 loss. In practice, we found this combined loss to be more robust on noisy training data.
https://arxiv.org/pdf/1806.04558.pdf

I also confirmed that l2 + l1 loss leads faster attention learning.
So I doubt that l2 loss is needed.
(Deep Voice3 uses only l2 loss.)
> An L1 loss is computed using the output mel-spectrograms and a binary cross-entropy loss is computed using the final-frame prediction.
https://arxiv.org/pdf/1710.07654.pdf

kan-bayashi:
About the fbank, even if the model cannot learn the attention,
If we calculate the outputs with teacher-forcing, it looks good.
This is because the model became a just like auto-encoder.
In the case of mel-outputs, if we use `reduction-factor=1`, the difference between previous frame and current frame is very small, so the model do not care about the encoder inputs.
Therefore, we carefully check the attention weights are diagonal.

ShigekiKarita:
Sorry, I forgot to run tests before the PR.

kan-bayashi:
No problem :)

kan-bayashi:
@ShigekiKarita How do you think following changes?

https://github.com/espnet/espnet/blob/0213427187b25b80800949b7d0ae8f6184f0fe2c/espnet/nets/pytorch_backend/e2e_tts_transformer.py#L573-L634

https://github.com/espnet/espnet/blob/0213427187b25b80800949b7d0ae8f6184f0fe2c/espnet/nets/pytorch_backend/transformer/attention.py#L52-L57

kan-bayashi:
Still I could not train the model (no diagonal encoder-decoder attention).
I will try following two configs:

- shallow & wide 
- large reduction-factor
- strong dropout for decoder input but weak dropout for encoder output

```yaml
# This is the Transformer TTS configuration

# network architecture related
model-module: espnet.nets.pytorch_backend.e2e_tts_transformer:Transformer
embed-dim: 0
eprenet-conv-layers: 0
eprenet-conv-filts: 0
eprenet-conv-chans: 0
dprenet-layers: 2  # one more linear layer w/o non-linear will be added for 0-centor
dprenet-units: 256
adim: 512
aheads: 4
elayers: 3
eunits: 2048
dlayers: 3
dunits: 2048
postnet-layers: 5
postnet-filts: 5
postnet-chans: 256
use-masking: True
bce-pos-weight: 5.0
use-batch-norm: True
use-scaled-pos-enc: True
encoder-normalize-before: False
decoder-normalize-before: False
encoder-concate-after: False
decoder-concate-after: False
reduction-factor: 4

# minibatch related
sortagrad: -1         # Feed samples from shortest to longest ; -1: enabled for all epochs, 0: disabled, other: enabled for 'other' epochs
batch-sort-key: input # shuffle or input or output
batch-bins: 6072000   # 80 * (870 * 80 + 180 * 35)
                      # batch-size * (max_out * dim_out + max_in * dim_in)
                      # resuling in 61 ~ 273 samples in batch (127 batches per epochs) for ljspeech

# training related
transformer-init: pytorch
transformer-warmup-steps: 4000
transformer-lr: 1.0
initial-encoder-alpha: 1.0
initial-decoder-alpha: 1.0
eprenet-dropout-rate: 0.1
dprenet-dropout-rate: 0.5
postnet-dropout-rate: 0.1
transformer-enc-dropout-rate: 0.1
transformer-enc-attn-dropout-rate: 0.0
transformer-dec-dropout-rate: 0.1
transformer-dec-attn-dropout-rate: 0.3
transformer-enc-dec-attn-dropout-rate: 0.0

# optimization related
opt: noam
accum-grad: 1
grad-clip: 1.0
weight-decay: 0.0
patience: 0
epochs: 2000  # maybe need around 200k iters
              # 2,000 epochs * 127 batches / 1 accum-grad = 254,000 iters

# other
save-interval-epoch: 10
```

```yaml
# This is the Transformer TTS configuration

# network architecture related
model-module: espnet.nets.pytorch_backend.e2e_tts_transformer:Transformer
embed-dim: 0
eprenet-conv-layers: 0
eprenet-conv-filts: 0
eprenet-conv-chans: 0
dprenet-layers: 2  # one more linear layer w/o non-linear will be added for 0-centor
dprenet-units: 256
adim: 512
aheads: 4
elayers: 3
eunits: 2048
dlayers: 3
dunits: 2048
postnet-layers: 5
postnet-filts: 5
postnet-chans: 256
use-masking: True
bce-pos-weight: 5.0
use-batch-norm: True
use-scaled-pos-enc: True
encoder-normalize-before: False
decoder-normalize-before: False
encoder-concate-after: True
decoder-concate-after: True
reduction-factor: 4

# minibatch related
sortagrad: -1         # Feed samples from shortest to longest ; -1: enabled for all epochs, 0: disabled, other: enabled for 'other' epochs
batch-sort-key: input # shuffle or input or output
batch-bins: 6072000   # 80 * (870 * 80 + 180 * 35)
                      # batch-size * (max_out * dim_out + max_in * dim_in)
                      # resuling in 61 ~ 273 samples in batch (127 batches per epochs) for ljspeech

# training related
transformer-init: pytorch
transformer-warmup-steps: 4000
transformer-lr: 1.0
initial-encoder-alpha: 1.0
initial-decoder-alpha: 1.0
eprenet-dropout-rate: 0.1
dprenet-dropout-rate: 0.5
postnet-dropout-rate: 0.1
transformer-enc-dropout-rate: 0.1
transformer-enc-attn-dropout-rate: 0.0
transformer-dec-dropout-rate: 0.1
transformer-dec-attn-dropout-rate: 0.3
transformer-enc-dec-attn-dropout-rate: 0.0

# optimization related
opt: noam
accum-grad: 1
grad-clip: 1.0
weight-decay: 0.0
patience: 0
epochs: 2000  # maybe need around 200k iters
              # 2,000 epochs * 127 batches / 1 accum-grad = 254,000 iters

# other
save-interval-epoch: 10
```

kan-bayashi:
Finally, I got diagonal encoder-decoder attention! (around 40k)

スクリーンショット 2019-06-05 午前8 42 33

This is the attention pattern of first config in https://github.com/espnet/espnet/pull/752#issuecomment-498724441

kan-bayashi:
After this PR, it is worthwhile to challenge following paper.
https://arxiv.org/pdf/1905.09263.pdf

ShigekiKarita:
Nice. I also reproduced it.
image

similar attention map is also obtained with 4gpus 1acc
image

kan-bayashi:
I got diagonal attention at the layer 2 of 3 layers.
But no diagonal attention is appeared in the layer 3.

Maybe due to this observation, the model still could not generate good speech, just like a growl.

ShigekiKarita:
Maybe it is fine.
> We notice that in both the 3-layer and 6-layer model, onlyalignments  from  certain  heads  of  the  **beginning  2  layers** are  interpretable  diagonal  lines,  which  shows  the  approx-imate correspondence between input and output sequence,while those of the following layers are disorganized

4.6 https://arxiv.org/pdf/1809.08895.pdf

kan-bayashi:
I added Guided attention in following paper’s eq 8 & 9.
https://arxiv.org/pdf/1811.04076.pdf
But it is not straight forward to apply guided attention for multi-head attention.
Currently, I apply the guided attention loss for the top 2 heads of top 2 layers.

kan-bayashi:
It seems that guided attention is quite strong.

Encoder self-attention
encoder

Decoder self-attention
decoder-self

Encoder-decoder attention
decoder-src

These are attention pattern @ only 30 epochs.
I applied guided attention loss for 2x2 attention maps at the bottom left.

r9y9:
FWIW, I tried adding encoder-prenset back based on the first config in https://github.com/espnet/espnet/pull/752#issuecomment-498724441, but it doesn't seem to work and I cannot get diagnal encoder-decoder attention anymore. This is strange...
```diff
diff transformer_1.yaml transformer_3.yaml
5,8c5,8
< embed-dim: 0
< eprenet-conv-layers: 0
< eprenet-conv-filts: 0
< eprenet-conv-chans: 0
---
> embed-dim: 512
> eprenet-conv-layers: 2
> eprenet-conv-filts: 5
> eprenet-conv-chans: 256
```

kan-bayashi:
Hmm... Actually I doubt that encoder-prenet is needed.
In this paper, they use just Embedding layer.
https://arxiv.org/pdf/1905.06791.pdf

BTW, even if we get a clear diagonal attention, the model still cannot generate speech...
What should I believe in? :(

kan-bayashi:
I still could not generate speech even if I got a diagonal attention.
I lost the target which we should check :(

train config
```python
# This is the Transformer TTS configuration

# network architecture related
model-module: espnet.nets.pytorch_backend.e2e_tts_transformer:Transformer
embed-dim: 0
eprenet-conv-layers: 0
eprenet-conv-filts: 0
eprenet-conv-chans: 0
dprenet-layers: 2  # one more linear layer w/o non-linear will be added for 0-centor
dprenet-units: 256
adim: 384
aheads: 4
elayers: 3
eunits: 1536
dlayers: 3
dunits: 1536
postnet-layers: 5
postnet-filts: 5
postnet-chans: 256
use-masking: True
bce-pos-weight: 5.0
use-batch-norm: True
use-scaled-pos-enc: True
encoder-normalize-before: False
decoder-normalize-before: False
encoder-concate-after: False
decoder-concate-after: False
reduction-factor: 4

# minibatch related
batch-sort-key: input  # shuffle or input or output
batch-bins: 4554000   # 60 * (870 * 80 + 180 * 35)
                      # batch-size * (max_out * dim_out + max_in * dim_in)
                      # resuling in 39 ~ 209 samples (avg 74 samples) in batch (169 batches per epochs) for ljspeech

# training related
transformer-init: pytorch
transformer-warmup-steps: 4000
transformer-lr: 1.0
initial-encoder-alpha: 1.0
initial-decoder-alpha: 1.0
eprenet-dropout-rate: 0
dprenet-dropout-rate: 0.5
postnet-dropout-rate: 0.5
transformer-enc-dropout-rate: 0.1
transformer-enc-positional-dropout-rate: 0.1
transformer-enc-attn-dropout-rate: 0.1
transformer-dec-dropout-rate: 0.1
transformer-dec-positional-dropout-rate: 0.5
transformer-dec-attn-dropout-rate: 0.3
transformer-enc-dec-attn-dropout-rate: 0.1
use-guided-attn-loss: true
num-heads-applied-guided-attn: 2
num-layers-applied-guided-attn: 2
modules-applied-guided-attn: ["encoder", "decoder", "encoder-decoder"]

# optimization related
opt: noam
accum-grad: 1
grad-clip: 1.0
weight-decay: 0.0
patience: 0
epochs: 2000  # maybe need around 200k iters
              # 2,000 epochs * 169 batches / 1 accum-grad = 338,000 iters

# other
save-interval-epoch: 10
```

Learning curve
all_loss

Attentions
- encoder self-attention
- decoder self-attention
- encoder-decoder attention
attn

r9y9:
Have you checked if the batch forward pass and inference pass of the decoder are doing the (exactly) same things? e.g. https://github.com/r9y9/deepvoice3_pytorch/blob/5897eedb45b6c6c16b21849c8ac2592eca2a81f5/tests/test_deepvoice3.py#L215-L231

(Sorry if you have already carefully checked it, but just in case). I was going to write a test to confirm its correctness but had no time to do that so far...

kan-bayashi:
OK. I will write such kind of test.

kan-bayashi:
I found a critical serious bug.
I will fix it.
Thank you for your advice, @r9y9!

r9y9:
Ah, that reminds me of the similar bug I had before... (https://github.com/r9y9/deepvoice3_pytorch/issues/1#issuecomment-341487171)

kan-bayashi:
Not to forget this bug, I will leave memo.

I used the same inputs and targets for decoder, so the model was not auto-regressive. 
This is the critical to learn diagonal attention and therefore the model cannot generate any speech.
To fix this issue, I changed the inputs by adding zero frame at the first frame and removing the last frame.
Also, I added test to check the forward calculation and inference calculation with groundtruth have the same results.

ShigekiKarita:
OMG I should notice this earlier when I suggest checking decoder NaN masking. Well done!

espnet/tts/pytorch_backend/tts.py
```diff
@@ -268,6 +268,9 @@ def train(args):
         optimizer = torch.optim.Adam(
             model.parameters(), args.lr, eps=args.eps,
             weight_decay=args.weight_decay)
+    elif args.opt == 'noam':
```
sw005320: I think it's better to move this optimizer to the general place in the later PR.
This optimizer should be used even for RNN.

espnet/nets/pytorch_backend/transformer/decoder_layer.py
```diff
@@ -33,10 +45,33 @@ def forward(self, tgt, tgt_mask, memory, memory_mask):
...
         :param torch.Tensor memory_mask: mask for memory (batch, max_time_in)
         """
-        x = tgt
-        nx = self.norm1(x)
-        x = x + self.dropout(self.self_attn(nx, nx, nx, tgt_mask))
-        nx = self.norm2(x)
-        x = x + self.dropout(self.src_attn(nx, memory, memory, memory_mask))
-        nx = self.norm3(x)
-        return x + self.dropout(self.feed_forward(nx)), tgt_mask, memory, memory_mask
+        residual = tgt
```
sw005320: I just want to make sure.
Is that OK?
`residual ` may not be changed?

kan-bayashi: I checked.

```
[ins] In [1]: import torch

[ins] In [2]: x = torch.zeros(3)

[ins] In [3]: x
Out[3]: tensor([0., 0., 0.])

[ins] In [4]: y = x

[ins] In [5]: x = x + 1

[ins] In [6]: y
Out[6]: tensor([0., 0., 0.])

[ins] In [7]: x
Out[7]: tensor([1., 1., 1.])
```

sw005320: OK. Thanks.

kan-bayashi:
Still training is on going but I get a nice results especially when we use a large model.
I tried 4 configs (see the bottom).
You can listen to the samples from following url.
https://drive.google.com/open?id=11AjCuzK7qIcD0IVT4-eBdJi7H97pYFm2

v1 -> failed. no speech.
v2 -> can generate but speed is slow...
v3 -> can generate but speed is slow 
v4 -> very nice. it is comparable to taco2 samples.

I doubt that reduction factor > 2 has some bugs from generated speech.
Anyway, we can successfully train the TTS-transformer in the case of large model with no reduction factor.

`train_transformer.v1.yaml`
- base config
```yaml
# network architecture related
model-module: espnet.nets.pytorch_backend.e2e_tts_transformer:Transformer
embed-dim: 512
eprenet-conv-layers: 3  # one more linear layer w/o non-linear will be added for 0-centor
eprenet-conv-filts: 5
eprenet-conv-chans: 512
dprenet-layers: 2  # one more linear layer w/o non-linear will be added for 0-centor
dprenet-units: 256
adim: 384
aheads: 4
elayers: 3
eunits: 1536
dlayers: 3
dunits: 1536
postnet-layers: 5
postnet-filts: 5
postnet-chans: 256
use-masking: True
bce-pos-weight: 5.0
use-batch-norm: True
use-scaled-pos-enc: True
encoder-normalize-before: True
decoder-normalize-before: True
encoder-concate-after: False
decoder-concate-after: False
reduction-factor: 4

# minibatch related
batch-sort-key: input # shuffle or input or output
batch-bins: 4554000   # 60 * (870 * 80 + 180 * 35)
                      # batch-size * (max_out * dim_out + max_in * dim_in)
                      # resuling in 39 ~ 209 samples (avg 74 samples) in batch (169 batches per epochs) for ljspeech

# training related
transformer-init: pytorch
transformer-warmup-steps: 4000
transformer-lr: 1.0
initial-encoder-alpha: 1.0
initial-decoder-alpha: 1.0
eprenet-dropout-rate: 0.5
dprenet-dropout-rate: 0.5
postnet-dropout-rate: 0.5
transformer-enc-dropout-rate: 0.1
transformer-enc-positional-dropout-rate: 0.1
transformer-enc-attn-dropout-rate: 0.1
transformer-dec-dropout-rate: 0.1
transformer-dec-positional-dropout-rate: 0.1
transformer-dec-attn-dropout-rate: 0.1
transformer-enc-dec-attn-dropout-rate: 0.1
use-guided-attn-loss: false

# optimization related
opt: noam
accum-grad: 1
grad-clip: 1.0
weight-decay: 0.0
patience: 0
epochs: 1000  # 1,000 epochs * 169 batches / 1 accum-grad = 169,000 iters
```

`train_transformer.v2.yaml`
- no encoder prenet
- no before normalization
- use guided attention for all attentions
- small reduction factor
```yaml
# network architecture related
model-module: espnet.nets.pytorch_backend.e2e_tts_transformer:Transformer
embed-dim: 0
eprenet-conv-layers: 0  # one more linear layer w/o non-linear will be added for 0-centor
eprenet-conv-filts: 0
eprenet-conv-chans: 0
dprenet-layers: 2  # one more linear layer w/o non-linear will be added for 0-centor
dprenet-units: 256
adim: 384
aheads: 4
elayers: 3
eunits: 1536
dlayers: 3
dunits: 1536
postnet-layers: 5
postnet-filts: 5
postnet-chans: 256
use-masking: True
bce-pos-weight: 5.0
use-batch-norm: True
use-scaled-pos-enc: True
encoder-normalize-before: False
decoder-normalize-before: False
encoder-concate-after: False
decoder-concate-after: False
reduction-factor: 2

# minibatch related
batch-sort-key: input # shuffle or input or output
batch-bins: 7590000   # 100 * (870 * 80 + 180 * 35)
                      # batch-size * (max_out * dim_out + max_in * dim_in)
                      # resuling in 79 ~ 283 samples (avg 122 samples) in batch (103 batches per epochs) for ljspeech

# training related
transformer-init: pytorch
transformer-warmup-steps: 4000
transformer-lr: 1.0
initial-encoder-alpha: 1.0
initial-decoder-alpha: 1.0
eprenet-dropout-rate: 0.0
dprenet-dropout-rate: 0.5
postnet-dropout-rate: 0.5
transformer-enc-dropout-rate: 0.1
transformer-enc-positional-dropout-rate: 0.1
transformer-enc-attn-dropout-rate: 0.1
transformer-dec-dropout-rate: 0.1
transformer-dec-positional-dropout-rate: 0.1
transformer-dec-attn-dropout-rate: 0.1
transformer-enc-dec-attn-dropout-rate: 0.1
use-guided-attn-loss: true
num-heads-applied-guided-attn: 2
num-layers-applied-guided-attn: 2
modules-applied-guided-attn: ["decoder", "encoder", "encoder-decoder"]

# optimization related
opt: noam
accum-grad: 1
grad-clip: 1.0
weight-decay: 0.0
patience: 0
epochs: 1000  # 1,000 epochs * 103 batches / 1 accum-grad = 103,000 iters
```

`train_transformer.v3.yaml`
- no encoder prenet
- no before normalization
- use guided attention for only encoder-decoder attention
- small reduction factor
```yaml
# network architecture related
model-module: espnet.nets.pytorch_backend.e2e_tts_transformer:Transformer
embed-dim: 0
eprenet-conv-layers: 0  # one more linear layer w/o non-linear will be added for 0-centor
eprenet-conv-filts: 0
eprenet-conv-chans: 0
dprenet-layers: 2  # one more linear layer w/o non-linear will be added for 0-centor
dprenet-units: 256
adim: 384
aheads: 4
elayers: 3
eunits: 1536
dlayers: 3
dunits: 1536
postnet-layers: 5
postnet-filts: 5
postnet-chans: 256
use-masking: True
bce-pos-weight: 5.0
use-batch-norm: True
use-scaled-pos-enc: True
encoder-normalize-before: False
decoder-normalize-before: False
encoder-concate-after: False
decoder-concate-after: False
reduction-factor: 2

# minibatch related
batch-sort-key: input # shuffle or input or output
batch-bins: 3795000   # 50 * (870 * 80 + 180 * 35)
                      # batch-size * (max_out * dim_out + max_in * dim_in)
                      # resuling in 50 ~ 173 samples (avg 62 samples) in batch (201 batches per epochs) for ljspeech

# training related
transformer-init: pytorch
transformer-warmup-steps: 4000
transformer-lr: 1.0
initial-encoder-alpha: 1.0
initial-decoder-alpha: 1.0
eprenet-dropout-rate: 0.0
dprenet-dropout-rate: 0.5
postnet-dropout-rate: 0.5
transformer-enc-dropout-rate: 0.1
transformer-enc-positional-dropout-rate: 0.1
transformer-enc-attn-dropout-rate: 0.1
transformer-dec-dropout-rate: 0.1
transformer-dec-positional-dropout-rate: 0.1
transformer-dec-attn-dropout-rate: 0.1
transformer-enc-dec-attn-dropout-rate: 0.1
use-guided-attn-loss: true
num-heads-applied-guided-attn: 2
num-layers-applied-guided-attn: 2
modules-applied-guided-attn: ["encoder-decoder"]

# optimization related
opt: noam
accum-grad: 2
grad-clip: 1.0
weight-decay: 0.0
patience: 0
epochs: 1000  # 1,000 epochs * 201 batches / 2 accum-grad = 100,500 iters
```

`train_transformer.v4.yaml`
- no encoder prenet
- no before normalization
- no reduction factor
- use guided attention for only encoder-decoder
- ※ requires 3 gpus with this settings
```yaml
# network architecture related
model-module: espnet.nets.pytorch_backend.e2e_tts_transformer:Transformer
embed-dim: 0
eprenet-conv-layers: 0  # one more linear layer w/o non-linear will be added for 0-centor
eprenet-conv-filts: 0
eprenet-conv-chans: 0
dprenet-layers: 2  # one more linear layer w/o non-linear will be added for 0-centor
dprenet-units: 256
adim: 384
aheads: 4
elayers: 6
eunits: 1536
dlayers: 6
dunits: 1536
postnet-layers: 5
postnet-filts: 5
postnet-chans: 256
use-masking: True
bce-pos-weight: 5.0
use-batch-norm: True
use-scaled-pos-enc: True
encoder-normalize-before: False
decoder-normalize-before: False
encoder-concate-after: False
decoder-concate-after: False
reduction-factor: 1

# minibatch related
batch-sort-key: input # shuffle or input or output
batch-bins: 2277000   # 30 * (870 * 80 + 180 * 35)
                      # batch-size * (max_out * dim_out + max_in * dim_in)
                      # resuling in 38 ~ 127 samples (avg 38 samples) in batch (330 batches per epochs) for ljspeech

# training related
transformer-init: pytorch
transformer-warmup-steps: 4000
transformer-lr: 1.0
initial-encoder-alpha: 1.0
initial-decoder-alpha: 1.0
eprenet-dropout-rate: 0.0
dprenet-dropout-rate: 0.5
postnet-dropout-rate: 0.5
transformer-enc-dropout-rate: 0.1
transformer-enc-positional-dropout-rate: 0.1
transformer-enc-attn-dropout-rate: 0.1
transformer-dec-dropout-rate: 0.1
transformer-dec-positional-dropout-rate: 0.1
transformer-dec-attn-dropout-rate: 0.1
transformer-enc-dec-attn-dropout-rate: 0.1
use-guided-attn-loss: true
num-heads-applied-guided-attn: 2
num-layers-applied-guided-attn: 2
modules-applied-guided-attn: ["encoder-decoder"]

# optimization related
opt: noam
accum-grad: 2
grad-clip: 1.0
weight-decay: 0.0
patience: 0
epochs: 1000  # 1,000 epochs * 330 batches / 2 accum-grad = 165,000 iters
```

espnet/tts/pytorch_backend/tts.py
ShigekiKarita: maybe we should create OptimizerInterface like fairseq that wraps pytorch/chainer optimizers and has ability to register cmd args.

ShigekiKarita:
I do not like this name `xxx-concate-after` because `concate` is not familiar to me. It can be `concat` or `concatenerate`. Or you can remove it if this option is useless.

kan-bayashi:
I’ve not yet confirm the effectiveness of this option. So I do not want to remove within this PR.
I will rename concate -> concat.

kan-bayashi:
I found a bug related to reduction_factor.
I will fix it.

kan-bayashi:
Training of the large model will be finished by the end of this week.

kan-bayashi:
I think now it is ready to merge.
If you want to wait the update of `RESULTS.md`, please wait by the end of this week.

kan-bayashi:
I confirmed that both v1 and v2 config models can generate reasonable speech.
Now reduction-factor option works correctly.

ShigekiKarita:
OK. LGTM already but keep on checking until RESULTS

ShigekiKarita:
https://github.com/kan-bayashi/espnet/blob/54eba80eb76f4048035ec6a633e1637f858d1c84/espnet/utils/training/batchfy.py#L300-L306

In your config, I believe `batch_sort_key=output` (it is speech side in fact because of `swap_io=True` in TTS) can allow more parallelism because speech dominates more memory than text.

The keys `input` and `output` are really confusing. Maybe we should use simply `speech` and `text` or something more expressive in the future release.

kan-bayashi:
I see. 
I’m not fully understand batch-bins algorithm, so I did not care about that so much.
I will try it.

> The keys input and output are really confusing. 
> Maybe we should use simply speech and text or something more expressive in the future release.

I think so too. But maybe we will support NMT, then `text` might be more confusing lol

ShigekiKarita:
> NMT

I think `en_text` and `ja_text` are fine for NMT, rather than just `text`.  Furthermore, in ASRTTS or any back-translation tasks, the key input or output is non-sense. However it results big breaking changes, I think the expressive (i.e., task-specific or user-defined) key is important. Maybe I will create a new ISSUE for discussion if I find a good solution.

ShigekiKarita:
Batching is conducted in two stages
1. sorting an array of training pairs by `batch_sort_key`
2. pick up contiguous ranges of the array not to exceed `--batch-xxx`

In the first stage, we can choose which matrix (input or output) has less padded elements. For example,
```python
# training pair info
ilens = [1, 4, 2, 3]
olens = [100, 200, 500, 400]
batch_size = 2

# make_batch result
if batch_sort_key="input":
  batch_io_lens = [
    {"ilens": [1, 2], "olens": [100, 500]},   # 1 + 400 frames wasted by padding
    {"ilens": [3, 4], "olens": [400, 200]}]   # 1 + 200 frames wasted by padding
else:
  batch_io_lens = [
    {"ilens": [1, 4], "olens": [100, 200]},   # 3 + 100 frames wasted by padding
    {"ilens": [3, 2], "olens": [400, 500]}]   # 1 + 100 frames wasted by padding
```
This is the reason why we should set longer side for `batch_sort_key`

ShigekiKarita:
I made tiny PR to support no postnet config like Tacotron2 https://github.com/kan-bayashi/espnet/pull/5

sw005320:
Done! Good job @kan-bayashi!

PR #753, opened 2019-05-24
Update Chainer to 6.0.0

jnishi:
Update Chainer to 6.0.0.  The patch of multiprocessiterator can be remove because it is merged to upstream Chainer from that version.


sw005320:
Thanks! Now we don't need our own patch, and also try chainerX!

PR #754, opened 2019-05-24
Make a specaugment default option False

sw005320:


sw005320:
@ShigekiKarita, @kan-bayashi. I made @bobchennan's spec augment default option false. Please be careful about it.

PR #758, opened 2019-05-24
Update voxforge recipe

sw005320:
- Added a transformer recipe for the well-know Adi set (@arendu).
- Training is still ongoing, but already (at 100 epochs) got the better result from RNN

|           | RNN (CER%) | Transformer (CER%) |
|-----------|:----:|:----:|
| dev | 12.9| 12.2|
| test |12.6 |12.5 |

- Will add the result after the training is finished.

### DIFF
--- a/egs/voxforge/asr2/RESULTS
+++ b/egs/voxforge/asr2/RESULTS
@@ -1,0 +1,24 @@
+# RNN default
+# change several update including ctc/attention decoding, label smoothing, and fixed search parameters
+write a CER (or TER) result in exp/tr_it_debug_alpha0.5/decode_dt_it_beam20_eacc.best_p0_len0.0-0.0_ctcw0.5/result.txt
+| SPKR                  | # Snt   # Wrd | Corr     Sub    Del     Ins    Err   S.Err |
+| Sum/Avg               | 1082    79133 | 89.6     5.5    5.0     2.5   12.9    98.2 |
+write a CER (or TER) result in exp/tr_it_debug_alpha0.5/decode_et_it_beam20_eacc.best_p0_len0.0-0.0_ctcw0.5/result.txt
+| SPKR                  | # Snt   # Wrd | Corr     Sub    Del     Ins    Err   S.Err |
+| Sum/Avg               | 1055    77966 | 89.7     5.5    4.8     2.3   12.6    98.4 |
+
+# Transformer 100 epochs
+shinji@b14:/export/a08/shinji/201707e2e/espnet_dev6/egs/voxforge/asr2$ grep -e Avg -e SPKR -m 2 exp/tr_it_pytorch_nopatience/decode_dt_it_decode/result.txt
+| SPKR                  | # Snt  # Wrd | Corr    Sub    Del    Ins    Err  S.Err |
+| Sum/Avg               | 1082   79133 | 90.1    4.2    5.7    2.3   12.2   98.6 |
+shinji@b14:/export/a08/shinji/201707e2e/espnet_dev6/egs/voxforge/asr2$ grep -e Avg -e SPKR -m 2 exp/tr_it_pytorch_nopatience/decode_et_it_decode/result.txt
+| SPKR                  | # Snt  # Wrd | Corr    Sub    Del    Ins    Err  S.Err |
+| Sum/Avg               | 1055   77966 | 89.7    4.4    6.0    2.2   12.5   99.1 |
+
+# Transformer 300 epochs
+write a CER (or TER) result in exp/tr_it_pytorch_ep300pa10/decode_dt_it_decode/result.txt
+| SPKR                  | # Snt  # Wrd | Corr    Sub    Del    Ins    Err  S.Err |
+| Sum/Avg               | 1082   79133 | 92.0    3.9    4.1    1.8    9.8   96.2 |
+write a CER (or TER) result in exp/tr_it_pytorch_ep300pa10/decode_et_it_decode/result.txt
+| SPKR                  | # Snt  # Wrd | Corr    Sub    Del    Ins    Err  S.Err |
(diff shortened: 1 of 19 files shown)


ShigekiKarita:
I do not have any request for this PR.

However, I have a big suggestion for `asr1` and `asr2` usage.

- At least, we can use a symbolic link from `asr1/local` to `asr2/local` because we have less effort to keep them update.
- In v.0.4.0, I want to merge `asr1` and `asr2` because RNN and Transformer does not require different data preparation at this time. We can divide them later easily but we cannot merge them later. These is no difference in `run.sh` but this guy
```bash
    if [ ${n_average} -gt 1 ]; then
	recog_model=model.last${n_average}.avg.best
	average_checkpoints.py --backend ${backend} \
			       --snapshots ${expdir}/results/snapshot.ep.* \
			       --out ${expdir}/results/${recog_model} \
			       --num ${n_average}
```

cc: how do you think @kamo-naoyuki?

sw005320:
If so, I think we should have a clear distinction for RNN and transformer configs (ruled naming or different directories). Symlink for the `local` directory seems to be fine, but I like the very clear distinction between RNN and transformer, and the idea of using `asr1` and `asr2` directories is more clear for me. Also, I personally don't like to have too many configs. I want to have other people's comments for this.

ShigekiKarita:
> Also, I personally don't like to have too many configs. 

Maybe we can hide configs into `conf/tuning/train_xxx.yaml` and link the best one to `conf/train.yaml` like kaldi. Another reason  is Tensorboard. When we build a new recipe, we can easily monitor and compare them if they share the root.

sw005320:
> Another reason is Tensorboard. When we build a new recipe, we can easily monitor and compare them if they share the root.

This is a clear reason to merge the directories. We can just use `asr2` (emphasize that if a recipe only has the `asr1` directory, which is an old style).

ShigekiKarita:
@kan-bayashi Now you are working on Transformer TTS (maybe tts2). Is it possible to unify? I think there are three options;
1. unify `asr1` and `asr2` to `asr1`
2. unify RNN and Transformer to `asr2` and leave `asr1` as legacy
3. keep `asr1` and `asr2` separate as they are

I still feel the 1st is the best because `asr1` and `asr2` are very similar at this time.  We can introduce `asr2` when we commit large breaking changes, for example, switching `data.json` to hdf5, kaldi-free recipe, etc.

kan-bayashi:
In #752, I updated all of TTS recipes to use config files.
(That is amazing useful! Thank you @kamo-naoyuki)

option1 or option2 is ok for me.
In my preference, I like option 1 because what to we need is just preparing two config files.
e.g. `conf/train_rnn.yaml` `conf/train_transformer.yaml`
or `conf/train_tacotron2.ymal` `conf/train_transformer.yaml`

In the case of option2, do we have to revert old recipes? (which do not use config files)

ShigekiKarita:
If we use the config filename as @kan-bayashi suggested, we can branch model averaging easily 
```bash
# cmd args
net=transformer # or rnn
...

# data prep
...

# training stage
${train_cmd} ... asr_train.py --config train_${net}.yaml

# decoding stage
if [ ${net} = transformer ]; then
    recog_model=model.last${n_average}.avg.best
    average_checkpoints.py ...
fi
...
${decode_cmd} ... asr_recog.py --config decode_${net}.yaml
...
```

sw005320:
I just finished running 300 epochs.
300 epochs might be too many for CTC, but the validation accuracy is still improving.

loss
acc

sw005320:
This is amazing. I did not tune anything, but I could get super significant gains for such a small task.

|           | RNN (CER%) | Transformer 100 epoch (CER%) | Transformer 300 epoch (CER%) |
|-----------|:----:|:----:|:----:|
| dev | 12.9| 12.2| 9.8 |
| test |12.6 |12.5 | 9.6 |

sw005320:
OK, we can go with option 1. I still feel that we need to define a strict naming for it (@kan-bayashi's idea may not be enough I think. We may need to add pytorch or chainer info). Also, we need to merge `RESULTS`.  I'm waiting for @kamo-naoyuki's comments.

kamo-naoyuki:
I dont have no big opposition, especially.

- I prefer the option1 plan - unify asr2 into asr1. It may be a large maintenance cost to keep the recipe both asr2 and legacy asr1.
- @kan-bayashi's naming is also good. 
- In the future, the number of the Yaml files will likely to increase and it may be a little annoying. So, as @ShigekiKarita said,  it is better to put them like `conf/tuning/train_xxx.yaml` and to create a symbolic link as `conf/train.yaml` to the default path. 
- If symbolic link is given, the option name can't be included in the expname, so it might be better to use `readlink`

```bash
train_config=$(readlink ${train_config})
expname=${train_set}_${backend}_$(basename ${train_config%.*})
```

I'm against the following @ShigekiKarita 's idea because it is inconvenient to restrict the naming style as `train_*.yaml`, we need always to create both of `train_*.py` and `decode_*.py` even if it can be reused about `decode.yaml`.
 
```bash
# cmd args
net=transformer # or rnn
```

How about getting the network name from the config file? Like,

```bash
if echo $(grep_yaml.py model-module ${train_config} ) | grep transformer &> /dev/null; then
    recog_model=model.last${n_average}.avg.best
    average_checkpoints.py ...
fi
```

kan-bayashi:
Why don’t you use symlink from `asr1/local` to `asr2/local`?

ShigekiKarita:
@kamo-naoyuki cool. this is simpler 
```bash
if [[ $(get_yaml,py ${train_config} model-module) = *transformer* ]]; then
```

ShigekiKarita:
@sw005320 why don't you add this CER to README.md? it seems very nice.

sw005320:
> @sw005320 why don't you add this CER to README.md? it seems very nice.

OK, I'll do it for the follow-up PR (merging `asr1` and `asr2`).

PR #759, opened 2019-05-25
Implement new make_batchset strategies

ShigekiKarita:
This PR includes the work of @gtache #580 

I implemented three make_batch strategies `seq`, `bin` and `frame`. You can choose one of them by `--batch-count seq`, (or `make_batchset` automatically detect it by finding enabled args).
- `seq`: Put `--batch-size` seqs in a minibatch. The batch size can be reduced by `--maxlen-in` and `--maxlen-out`. This is what we have been using.
- `bin`: Put `--batch-bins` in a minibatch. The "bin" means the number of `frames x dim`. @gtache implemented in #580 
- `frame`: Put `--batch-frames-in` (max input frames limit), `--batch-frames-out` (max output frames limit) and `--batch-frames-inout` (max input+output frames limit) in a minibatch. Each limit can be disabled when its value is 0 (default). This strategy is used in NMT frameworks e.g., tensor2tensor, fairseq.

Now I'm running some exp to see how it works...


ShigekiKarita:
I tested this PR with `egs/wsj/asr1/run.sh`. WSJ Train-SI284 has 782 (x 83 dim) input and 102 (x 52 class) output frames per utt on average (approx 800 and 100 frames). My 1080ti used to accept only max 32 seqs per minibatch in the current `--batch-size 32` config. Therefore, I tested new strategies for minibatching that also expects 32 seqs per minibatch.

| opt | Iters / epoch | GPU Memory Usage | Estimated time to finish (ETF) for 100 epochs after the 1st epoch |
|:--|--:|--:|--:|
| seq (--batch-size 32) | 2400 | 9.2GB | 17:07 |
| bin (--batch-bins 32 * (800 * 83 + 100 * 52) - 100) | 1100 | 9.9GB | 12:31 |
| frame (--batch-frames-inout 32 * (800 + 100)) | 1100 | 10.0GB | 14:49 |

(NOTE: `bin` needed some tweaks e.g., -100 to fit the memory on my GTX1080ti.)

Obviously, effective batchsize (see iters/epoch) in `frame` is **2.2 times larger** than the current strategy `seq`. I think `bin` or `frame` can exploit larger GPU memory and utilization by adaptively tiling seqs as many as possible. 

However, be careful on GPU overrun. I found my GTX1080ti went 309W/250W with `bin` and `frame`  :boom:

ShigekiKarita:
@kan-bayashi As this paper https://arxiv.org/pdf/1809.08895.pdf suggested, this might be a critical tip to train Transformer. You should try this later.

ShigekiKarita:
This is `egs/wsj/asr2` result. `--batch-bin` was faster but we need more investigation on hyperparametes to get the best the eval result. However, I think this PR is worth merging v.0.4.0 as new options. @sw005320 @gtache 

|opt| dev CER | dev WER | eval CER | eval WER|  training time sec | early stopped epoch | sec/epoch |
|:--|--:|--:|--:|--:|--:|--:|--:|
|--batch-size 32 (default)| 3.7 | 8.2 | 2.3 | 4.8 | 17893 | 29 | 617 |
|--batch-bin 32 * (800 * 83 + 100 * 52) - 100 | 3.2 | 7.1 | 2.2 | 5.9 | 20088 | 39 | 515 |

kan-bayashi:
Hi Shigeki, could you unify the your great function in ASR and TTS?

ShigekiKarita:
Do you mean existing RNN based recipe?

ShigekiKarita:
I think that this PR have not support `make_batchset` in TTS yet but I will take a look today.

kan-bayashi:
I think what we have to consider is that only the reverse of input and output.
https://github.com/espnet/espnet/blob/c1737105ded30ca6f8d2c5abc5472938ec6006be/espnet/tts/tts_utils.py#L12

ShigekiKarita:
gotcha

ShigekiKarita:
I confirmed new `tts_train.py --batch-count bin --batch-bins 780800` is working on `egs/an4/tts1`. I decided 780800 = 32 * (170 * 80 + 20 * 18 * 30), where mean(data[output][shape[0]) = 170 and mean(data[input][shape[0]) = 18. Note that `--batch_sort_key shuffle` will only work with `--batch-size`, while it raises ValueError with `--batch-bins` or `--batch-frames`

kan-bayashi:
Cool! Thanks Shigeki.

kan-bayashi:
> 32 * (170 * 80 + 20 * 18 * 30)

What is the 20?

ShigekiKarita:
It was a mistake. It should be `32 * (170 * 80 + 18 * 30)` i.e., `${batch_size} * (mean(data[:].input.shape[0]) * data[0].input.shape[1] * mean(data[:].output.shape[0]) * data[0].output.shape[1])`. I found this calc is useful to replace `--batch-size` with `--batch-bins` and `--batch-frames-xxx`. In many cases, we can raise the value a little by watching GPU memory usage.

kan-bayashi:
Now I’m tesing with following setting in tts-transformer.
It looks working good.

```
# minibatch related
batch-sort-key: input # shuffle or input or output
batch-bins: 1214400 # 16 * (870 * 80 + 180 * 35)
```

espnet/bin/asr_train.py
```diff
@@ -156,12 +157,22 @@ def main(cmd_args):
...
+                        help='Maximum seqs in a minibatch (0 to disable)')
+    parser.add_argument('--batch-bins', default=0, type=int,
+                        help='Maximum bins in a minibatch (0 to disable)')
+    parser.add_argument('--batch-frames-in', default=0, type=int,
+                        help='Maximum input frames in a minibatch (0 to disable)')
+    parser.add_argument('--batch-frames-out', default=0, type=int,
+                        help='Maximum output frames in a minibatch (0 to disable)')
+    parser.add_argument('--batch-frames-inout', default=0, type=int,
+                        help='Maximum input+output frames in a minibatch (0 to disable)')
     parser.add_argument('--maxlen-in', default=800, type=int, metavar='ML',
```
sw005320: We can unify the option names `--maxlen-in` --> `--batch-maxlen-in`, `--maxlen-out` --> `--batch-maxlen-out`
We can completely switch to `--batch-maxlen-xxx` and fix all `run.sh` or keep both `--batch-maxlen-xxx` and `--maxlen-xxx`

sw005320:
I'm a bit conservative to make this bin-based as default. It's faster but less intuitive. Let's try to test it with several recipes and decide it.

espnet/bin/asr_train.py
ShigekiKarita: I would like to keep both `--batch-frames-xxx` and  `--maxlen-xxx` because they play different roles.

ShigekiKarita: >     * it's better to move `batchfy.py` to `espnet/utils/batchfy.py` as it's used for both TTS and ASR.

agree

sw005320: I meant to say that we can just rename `--maxlen-xxx` to `--batch-maxlen-xxx` to follow the other naming conventions. You don't have to touch anything for `--maxlen-xxx`. Also, to clarify the role, we can rename them further as `--batch-seq-maxlen-xxx`. Again, by keeping the original `--maxlen-xxx` options, we'd not have confusions.

ShigekiKarita:
> I'm a bit conservative to make this bin-based as default. It's faster but less intuitive. Let's try to test it with several recipes and decide it.

In fact, [fairseq](https://github.com/pytorch/fairseq/tree/master/examples/translation#prepare-iwslt14sh), [T2T](https://github.com/tensorflow/tensor2tensor/issues/85) and [opennmt](https://github.com/OpenNMT/OpenNMT-py/blob/1b3cc33601a7d22c340044ddd7fc29f96267fae8/config/config-transformer-base-1GPU.yml#L27-L28) use this kind of batching as default. Therefore we might be unique. It is joke. I know `--batch-size` used to be the number of sentence before Transformer. The Transformer prefers `--batch-bins` because it has the 50:50 memory consumption in encoder and decoder, while RNN might not.

Anyway, I agree with your opinion. We can switch that later.

espnet/bin/asr_train.py
ShigekiKarita: I misunderstood. OK you mean
```python
parser.add_argument("--maxlen-in", "--batch-seq-maxlen-in", default=800, type=int, metavar='ML', ...)
```

sw005320: Correct.
With this, we can clarify the meaning of the option.

ShigekiKarita:
todo
- [ ] resolve conflicts
- [ ] update README
- [ ] add `--batch-seq-xxx` to argparse

ShigekiKarita:
done

espnet/bin/tts_train.py
```diff
@@ -62,13 +63,23 @@ def main(cmd_args):
...
+                        help='Maximum seqs in a minibatch (0 to disable)')
+    parser.add_argument('--batch-bins', default=0, type=int,
+                        help='Maximum bins in a minibatch (0 to disable)')
+    parser.add_argument('--batch-frames-in', default=0, type=int,
+                        help='Maximum input frames in a minibatch (0 to disable)')
+    parser.add_argument('--batch-frames-out', default=0, type=int,
+                        help='Maximum output frames in a minibatch (0 to disable)')
+    parser.add_argument('--batch-frames-inout', default=0, type=int,
+                        help='Maximum input+output frames in a minibatch (0 to disable)')
+    parser.add_argument('--maxlen-in', default=800, type=int, metavar='ML',
```
sw005320: Can you change these options similar to the ASR ones?

ShigekiKarita: ok

PR #769, opened 2019-05-28
unify rnn and transformer recipes for the voxforge task

sw005320:


ShigekiKarita:
LGTM but `get_yaml.py` is missing. Maybe it is better to merge unified WSJ asr #745 first

sw005320:
Oh yes, I found it now.

sw005320:
If #745 takes time, can you just make a PR for `get_yaml.py` related?

ShigekiKarita:
PR #745 is done except for reproducing @bobchennan previous commit result. I think we can rework it in the later PR.

PR #771, opened 2019-05-29
Fix a bug in the TEDLIUM recipe

takenori-y:
Prevent a file existence error caused by spm_encode at stage 2.

### DIFF
--- a/egs/tedlium/asr1/run.sh
+++ b/egs/tedlium/asr1/run.sh
@@ -114,7 +114,7 @@ if [ ${stage} -le 1 ] && [ ${stop_stage} -ge 1 ]; then
 fi
 
 dict=data/lang_char/${train_set}_${bpemode}${nbpe}_units.txt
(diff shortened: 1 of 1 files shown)


sw005320:
Can you tell me a little bit more detail? Which part causes the error?

takenori-y:
`spm_encode` causes an error since `$bpemodel` does not exist.

https://github.com/espnet/espnet/blob/8f9467f6784dbe2f83b3d83f705e21a2aa5cbcb3/egs/tedlium/asr1/run.sh#L127

sw005320:
Oh, I see. I'd like to use the same way as other recipes e.g., https://github.com/espnet/espnet/blob/master/egs/librispeech/asr1/run.sh (or https://github.com/espnet/espnet/search?q=bpemodel&unscoped_q=bpemodel).
Is it possible by just modifying `${bpemodel}` to `${bpemodel}.model`?

takenori-y:
Sure, no problem.

sw005320:
Thanks!

PR #761, opened 2019-05-25
revert Chainer 6.0.0 to 5.0.0

jnishi:
Revert https://github.com/espnet/espnet/pull/753.

See #760.

Issue #760, opened 2019-05-25
Chainer 6.0.0 caused some errors.


jnishi:
There are several internal modification in version 6.0.0 and they cause errors when running with pytorch backends.

For example, `Evaluator.device` returns `GpuDevice` instead of `int` (in Chainer version 5.0.0).
This causes following error in `lm/pytorch_backend/lm.py`.
```
Exception in main training loop: '>=' not supported between instances of 'GpuDevice' and 'int'
Traceback (most recent call last):
  File "/home/christopher/gitrepos/espnet-1/tools/venv/lib/python3.6/site-packages/chainer/training/trainer.py", line 319, in run
    entry.extension(self)
  File "/home/christopher/gitrepos/espnet-1/tools/venv/lib/python3.6/site-packages/chainer/training/extensions/evaluator.py", line 161, in __call__
    result = self.evaluate()
  File "/home/christopher/gitrepos/espnet-1/espnet/lm/pytorch_backend/lm.py", line 309, in evaluate
    x, t = concat_examples(batch, device=self.device, padding=(0, -100))
  File "/home/christopher/gitrepos/espnet-1/espnet/lm/pytorch_backend/lm.py", line 229, in concat_examples
    if device is not None and device >= 0:
Will finalize trainer extensions and updater before reraising the exception.
Traceback (most recent call last):
  File "/home/christopher/gitrepos/espnet-1/egs/an4/asr1/../../../espnet/bin/lm_train.py", line 145, in <module>
    main(sys.argv[1:])
  File "/home/christopher/gitrepos/espnet-1/egs/an4/asr1/../../../espnet/bin/lm_train.py", line 139, in main
    train(args)
  File "/home/christopher/gitrepos/espnet-1/espnet/lm/pytorch_backend/lm.py", line 421, in train
    trainer.run()
  File "/home/christopher/gitrepos/espnet-1/tools/venv/lib/python3.6/site-packages/chainer/training/trainer.py", line 349, in run
    six.reraise(*exc_info)
  File "/home/christopher/gitrepos/espnet-1/tools/venv/lib/python3.6/site-packages/six.py", line 693, in reraise
    raise value
  File "/home/christopher/gitrepos/espnet-1/tools/venv/lib/python3.6/site-packages/chainer/training/trainer.py", line 319, in run
    entry.extension(self)
[...]

PR #772, opened 2019-05-29
Fix pytorch for chainer v6.0.0

Fhrozen:
The corrections for pytorch with chainer v6.0.0 are included in this PR, to solve #760. 
- Modified LMTest; this returned gpuid as a GPUId class, set self.device inside __init__
- Replaced extensions.snapshot_object for (custom) snapshot_object. V6.0.0 call first [serialize](https://github.com/chainer/chainer/blob/17796d7cf17f667eb4e7b7cd41778ad9ec56c749/chainer/training/extensions/_snapshot.py#L228) and the use the savefun method (not sure why it does this), so the training will raise error due to the lack of serialize routine.  While v5.0.0 is [straight](https://github.com/chainer/chainer/blob/e4a0b8fca2fdd893c491de79f0b66129d0d15824/chainer/training/extensions/_snapshot.py#L81). Just copied the required code into the 'asr_utils' and replace the line in 'pytorch/asr/train' and 'pytorch/tts/train'.

I did not find any additional problem for pytorch with chainer v6.0.0 and lm_train, asr_train, asr_decode, and tts_train worked well.
I suppose #761 will no longer be required to merge.

### DIFF
--- a/espnet/asr/asr_utils.py
+++ b/espnet/asr/asr_utils.py
@@ -425,6 +425,28 @@ def torch_save(path, model):
         torch.save(model.state_dict(), path)
 
 
+def snapshot_object(target, filename):
+    """Returns a trainer extension to take snapshots of a given object.
+
+    Args:
+        target: Object to serialize.
+        filename (str): Name of the file into which the object is serialized.
+            It can be a format string, where the trainer object is passed to
+            the :meth:`str.format` method. For example,
+            ``'snapshot_{.updater.iteration}'`` is converted to
+            ``'snapshot_10000'`` at the 10,000th iteration.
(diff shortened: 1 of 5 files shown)


sw005320:
I found that it is appeared in the other parts (see https://github.com/espnet/espnet/search?q=extensions.snapshot_object&unscoped_q=extensions.snapshot_object).

Fhrozen:
I just modified the last one (asr_mix) for pytorch. In case of chainer, this does not have any problem.

sw005320:
Thanks, @Fhrozen!

jnishi:
Thanks. LGTM!

PR #773, opened 2019-05-29
Docker updates

Fhrozen:
Corrected some dockerfiles to support the new updates in kaldi and espnet.
Runfile and GPU containers are built without problem.

Still building and testing the containers before upload.
Once container are successfully uploaded, the WIP flag will be removed.


ShigekiKarita:
Ignore circle ci error because we do not have circle ci config in master (I'm searching how to disable circle CI build on master).

Fhrozen:
@ShigekiKarita It seems that at least it is required to define a config.yml inside the .cicleci in master to avoid building master https://circleci.com/docs/2.0/configuration-reference/#branches

@sw005320 the containers has been successfully tested and uploaded to the hub. This PR is ready to merge. Also, I've been thinking about stop the support for ubuntu 16.04 and build the containers with 18.04 and cuda 9.0+ only. The old containers will be not be deleted but they will not be updated.
Let me know about this.

sw005320:
> Also, I've been thinking about stop the support for ubuntu 16.04 and build the containers with 18.04 and cuda 9.0+ only. 

Sounds good for me.

PR #790, opened 2019-05-31
Add RESULTS.md for voxforge and chime5

sw005320:


sw005320:
I think https://github.com/koalaman/shellcheck/wiki/SC2001 is too strict...

PR #795, opened 2019-06-03
LibriTTS recipe

takenori-y:
I added the LibriTTS recipe. In my experiment, `atype=location` was clearly better than `atype=forward_ta` in terms of speech quality.

You can listen samples [here](https://drive.google.com/drive/folders/1_fKnxuFlLBFCATCsacxKzIy6UBbUPzd0?usp=sharing).


kan-bayashi:
Great. I will review it.

> I added the LibriTTS recipe. In my experiment, atype=location was clearly better than atype=forward_ta in terms of speech quality.
Did you compare with the same `reduction-factor`?

takenori-y:
Thank you for your review. Yes, `reduction-factor` was two in the both cases.

egs/libritts/tts1/local/data_prep.sh
```diff
@@ -0,0 +1,83 @@
...
+
+    find -L $chapter_dir/ -iname "*.wav" | sort | while read -r wav_file; do
+       id=$(basename $wav_file .wav)
+       echo "$id $wav_file" >>$wav_scp
+
+       txt=$(cat $(echo $wav_file | sed -e "s/\.wav$/.normalized.txt/"))
+       echo "$id $txt" >>$trans
+
+       # NOTE: For now we are using per-chapter utt2spk. That is each chapter is considered
+       #       to be a different speaker. This is done for simplicity and because we want
```
kan-bayashi: Let me confirm this point.
In this case, are speakers in `dev` and `eval` unknown or known ? (included or not in training data)

kan-bayashi: Sorry, I misunderstood.
`dev` and `test` are already separated so we do not care about that.

egs/libritts/tts1/run.sh
```diff
@@ -0,0 +1,250 @@
...
+fi
+
+if [ ${stage} -le 3 ] && [ ${stop_stage} -ge 3 ]; then
+    echo "stage 3: x-vector extraction"
+    # Make MFCCs and compute the energy-based VAD for each dataset
+    mfccdir=mfcc
+    vaddir=mfcc
+    for name in ${train_set} ${dev_set} ${eval_set}; do
+        utils/copy_data_dir.sh data/${name} data/${name}_mfcc
+        utils/data/resample_data_dir.sh 16000 data/${name}_mfcc
```
kan-bayashi: Why don't you make it clear by adding postfix `16k`? (e.g. `data/${name}_mfcc_16k`)

takenori-y: I added the postfix.

takenori-y:
OK. I modify `info.txt` if the script that summarizes experimental information is provided.

egs/libritts/tts1/run.sh
```diff
@@ -0,0 +1,250 @@
...
+resume=""    # the snapshot path to resume (if set empty, no effect)
+
+# feature extraction related
+fs=24000      # sampling frequency
+fmax=""       # maximum frequency
+fmin=""       # minimum frequency
+n_mels=80     # number of mel basis
+n_fft=1024    # number of fft points
+n_shift=256   # number of shift points
+win_length="" # window length
```
r9y9: How about `n_fft=2048`, `n_shift=300` and `win_length=1200` by following the Tacotron2 paper? If the sampling frequency is the same as experiments in Tacotron 2 (i.e. 24kHz), it would be nice to follow the acoustic feature extraction parameters in the paper unless we find other parameters that work better. Also increasing the n_shift reduces the decoder target lengths so I think it would make training seq2seq models slightly easier.

> https://arxiv.org/pdf/1712.05884.pdf
> As in Tacotron, mel spectrograms are computed through a short-time Fourier transform (STFT) using a 50 ms frame size, 12.5 ms frame hop, and a Hann window function.

egs/libritts/tts1/run.sh
```diff
@@ -0,0 +1,250 @@
...
+
+    utils/combine_data.sh data/${train_set}_org data/train_clean_100 data/train_clean_360
+    utils/combine_data.sh data/${dev_set}_org data/dev_clean
+
+    # remove utt having more than 3000 frames
+    # remove utt having more than 400 characters
+    remove_longshortdata.sh --maxframes 3000 --maxchars 400 data/${train_set}_org data/${train_set}
+    remove_longshortdata.sh --maxframes 3000 --maxchars 400 data/${dev_set}_org data/${dev_set}
+
+    # compute global CMVN
```
r9y9: I know this is a copy-and-paste, but I'd suggest rephrasing. When I first read the comment (in #752) I was very confused because
- mel-spectrogram is used for decoder target in Tacotron. It's not cepstrum.
- "compute global CMVN (cesptrum mean-variance normalization) " doesn't really say what is done here, though I can understand. "compute statistics for CMVN" is more correct and easy to understand for me.

So what about `compute statistics for global CMVN`?  or more accurately, `compute statistics for global mean-variance normalization`?

kan-bayashi: We are too familiar with kaldi...

egs/libritts/tts1/run.sh
takenori-y: Thank you for your suggestion. I will try that.

egs/libritts/tts1/run.sh
sw005320: I agree with @r9y9.
It's not easy to change the file name, but just try to change the comment part.

sw005320:
Many thanks!

> OK. I modify info.txt if the script that summarizes experimental information is provided.

You could make `RESULTS` the markdown format and follow my style a little bit (https://github.com/espnet/espnet/blob/9e63c8f85d04f2bed7dd22c6cd37e0c5acd7517e/egs/tedlium/asr1/RESULTS.md).
This is obtained by https://github.com/espnet/espnet/blob/v.0.4.0/utils/get_sys_info.sh and https://github.com/espnet/espnet/blob/v.0.4.0/utils/pack_model.sh
I believe you can still use `get_sys_info.sh`, but you could not use `pack_model.sh` as it's customized to the ASR models.
I think it's good to upload the model files, and if you also think of making the TTS version of `pack_model.sh` (another PR), which is great.

sw005320:
@takenori-y, could you fix the comment and/or restructure RESULTS?
I want to merge this PR.

takenori-y:
Could you wait until the experiment under the condition suggested by @r9y9 is finished? It requires a few more days.

sw005320:
OK

egs/libritts/tts1/run.sh
takenori-y: As you said, training speed became faster in the condition. The synthesized speech generated by the model trained on the condition was clear. This may be because `n_fft=2048`. However, the prosody was slightly worse and the power was quite low (I'm not sure why). I will continue to find good settings.

egs/libritts/tts1/run.sh
takenori-y: Changed to `compute statistics for global mean-variance normalization`.

takenori-y:
The experiment have finished due to early stopping. I updated my repository.

> I think it's good to upload the model files, and if you also think of making the TTS version of pack_model.sh (another PR), which is great.

OK, I will do it.

sw005320:
Thanks a lot. This is a great addition!

egs/libritts/tts1/run.sh
r9y9: Sorry to comment on the merged PR, but let me confirm the conclusion. Do you think we should stick to use n_fft=1024 and so on? If you think it's difficult to say one is clearly better than the other or vise versa, we may consider using ASR-based objective evaluation (#1190) to investigate differences in a measurable way.

takenori-y: Good idea. Let me try to objectively evaluate synthesized speech when the PR is merged.

PR #798, opened 2019-06-04
Add google drive downloader

kan-bayashi:


sw005320:
Only zip file? How about tgz files?

kan-bayashi:
.tgz is supported, but I don’t how to get file extension from share link so we need to specify the file extension.

PR #801, opened 2019-06-04
Fix recipes

takenori-y:
This is a trivial pull request.


sw005320:
Thanks for cleaning up those!

Issue #792, opened 2019-06-01
- or _


kamo-naoyuki:
Very trivial matter. 
The name of scripts in utils sometimes has `-` and the other has `_`.

- `make_stft.sh`
-  `compute-fbank-feats.py`

I want to unify them.

kan-bayashi:
When I implemented this function, I wanted to replace kaldi commands, so I used "-".
It is OK to replace all of "-" with "_" in file name.

r9y9:
A bit different topic though, we may also need to be consistent with the use of `A-to-B` or `A2B`. https://github.com/espnet/espnet/tree/master/utils

PR #805, opened 2019-06-05
change the scoring in multi-speaker case from shell to python compute…

simpleoier:
…-perm-free-error, add test script for it


utils/compute-perm-free-error.py
```diff
@@ -0,0 +1,147 @@
+#!/usr/bin/env python
+# encoding: utf-8
+
+# Copyright 2017 Johns Hopkins University (Shinji Watanabe)
```
sw005320: change it to your name.

utils/compute-perm-free-error.py
```diff
@@ -0,0 +1,147 @@
...
+        dic[k] = v
+
+    return dic
+
+
+def main():
+    parser = argparse.ArgumentParser()
+    parser.add_argument('--num-spkrs', type=int, default=2,
+                        help='number of mixed speakers.')
+    parser.add_argument('result_r1h1', type=str,
```
sw005320: Maybe you use `nargs='+'` and accept any numbers of files to generalize this scoring script for any speakers. Then, we can check that the number of files is consistent with the number of speakers later.

utils/minpermwer.py
```diff
@@ -4,18 +4,47 @@
...
 import sys
 
-is_python2 = sys.version_info[0] == 2
+
+# pre-set the permutation scheme (ref_idx, hyp_idx)
+def permutation_schemes(num_spkrs):
+    perm = [(i / num_spkrs + 1, i % num_spkrs + 1) for i in range(num_spkrs ** 2)]
+
+    # [r1h1, r1h2, r2h1, r2h2]
+    if num_spkrs == 2:
```
sw005320: Can we generalized this part given the number of speakers instead of writing the actual possible permutations? Then, you could generalize this script for any number of speakers.

simpleoier:
OK, I'll improve it.

sw005320:
quick comment
better to change the name of `utils/minpermwer.py`, which is consistent with others.
also see the discussion in #792 .

simpleoier:
What about the compute-perm-free-error.py? Should I also change the - with _?

sw005320:
`compute` is used for feature extraction. I think `min_perm_wer.py` is fine (let's use `_`).

sw005320:
- `utils/compute-perm-free-error.py` -> `utils/eval_perm_free_error.py`
- `test_utils/test_compute_perm_free_error_py.bats` -> `test_utils/test_eval_perm_free_error_py.bats`

sw005320:
Thanks a lot!

PR #816, opened 2019-06-07
Add guided attention loss for Tacotron2

kan-bayashi:
Should be merged after #815


kan-bayashi:
Ready to merge.

espnet/nets/pytorch_backend/e2e_tts_tacotron2.py
```diff
@@ -43,6 +43,108 @@ def make_non_pad_mask(lengths):
     return seq_range_expand < seq_length_expand
 
 
+class GuidedAttentionLoss(torch.nn.Module):
+    """Guided attention loss function
```
r9y9: Better to add a paper reference?

sw005320:
- can we use this for ASR? if so, it's better to design so that we can use this for both ASR and TTS
- can you add a specific test for this guided attention?

sw005320:
Correction. You seem to add some test. So please ignore my suggestion for the test part.

kan-bayashi:
> can we use this for ASR? if so, it's better to design so that we can use this for both ASR and TTS

It can be used for ASR.
What we need to calculate GuidedAttentionLoss are ilens, olens, att_ws.
https://github.com/espnet/espnet/blob/3f79e79639d09885cb1678e57dcd7d41d4399b86/espnet/nets/pytorch_backend/e2e_tts_tacotron2.py#L68-L75

But I do not want to support this loss for ASR in this PR.

sw005320:
OK

espnet/nets/pytorch_backend/e2e_tts_tacotron2.py
kan-bayashi: Added! Thank you for your review.

PR #819, opened 2019-06-07
add AMI multichannel recipe

Jzmo:
add ami multichannel recipe

@sw005320 added several action items
- [x] add RESULTS.md once you get it.
- [x] may need some tuning of the RNN architecture. The setup in this recipe seems to be old. I recommend Aurora4's configuration (https://github.com/espnet/espnet/blob/v.0.4.0/egs/aurora4/asr1/conf/train.yaml) for backend ASR and compare them.

These would be done with another PR, but I'm just listing them.
- [x] speed perturbation based data augmentation. The amount of training data of this recipe is too small, and such data augmentation would greatly improve the performance. See https://github.com/espnet/espnet/blob/v.0.4.0/egs/tedlium/asr1/run.sh#L91-L99
- [ ] add a transformer setup


sw005320:
Thanks. Quick comments.
- I guess you don't have to edit `utils/data2json.sh`. Please do not include this change to this PR.
- You seem to change the permission of the following files, but I believe it is not necessary. Please use the original ones.

image

Jzmo:
Thanks, I will modify it right now

sw005320:
@Xiaofei-Wang, please start a review once it is fixed.

PR #821, opened 2019-06-08
Update README.md

kamo-naoyuki:
`How to set minibatch` is unreadable. I modified it. I wish some figures were drawn.


sw005320:
thanks!

PR #823, opened 2019-06-10
Support n_average option

kan-bayashi:
This PR supports `n_average` option in TTS recipe.
In contrast to ASR, over-fit model does not always result in bad generated speech.
Therefore, I changed to use `n_average` option even if `n_average=1`.
In this case we just use the last model for decoding.


kan-bayashi:
Also, I added background parallel processing as the same as ASR.

PR #824, opened 2019-06-10
Transformer for AURORA4

27jiangziyan:
This PR adds transformer for AURORA4. I run 100 epochs right now and the final result is a little bit worse than the previous.(WER + ~1) I think there may have a little bit overfitting with 100 epochs.  I may take a try with removing model averaging and just use model.acc.best. I will report the result of this later.


ShigekiKarita:
great! 

> overfitting

you can try these tips
- increase dropout to 0.2
- increase label smoothing to 0.2
- smaller model: units=1024, elayers=6
- use subword instead of chars https://github.com/espnet/espnet/tree/v.0.4.0/egs/tedlium/asr1

sw005320:
> use subword instead of chars https://github.com/espnet/espnet/tree/v.0.4.0/egs/tedlium/asr1

I haven't had the success of using subword for the WSJ task and its variants.

@27jiangziyan, could you paste the training and validation accuracy curves?

27jiangziyan:
Thanks for your advice. Sorry since I think I use a wrong word in my last comment. I may not call it overfitting. The curve is very wield.
image
The training curve as above have a big decrease at the last epoch. 
The validation curve seems normal as below.
image
So I'm thinking to replace the model averaing with model.acc.best so that I could skip the last epoch.

sw005320:
Please tune the decoding parameter. At least you should use the same parameters with the WSJ transformer setup.

sw005320:
```
ctc-weight: 0.5
lm-weight: 0.7
```
should be
```
ctc-weight: 0.3
lm-weight: 1.0
```

We should not compare the result with different configurations....

27jiangziyan:
Follow your advice. Now the WER is better than the RNN based.(WER - 1.2). But I'm still little curious about the decrease at the last epoch of  the training curve. Do I need to try different training config?(May follow @ShigekiKarita advice to try a smaller model?)

sw005320:
Yes, you can, but I think the current result is enough for this PR. I'll merge it once we finish the test. BTW, could you also think of using the markdown files instead of the plain text RESULTS? See https://github.com/espnet/espnet/blob/v.0.4.0/egs/tedlium/asr1/RESULTS.md or https://github.com/espnet/espnet/blob/v.0.4.0/egs/aishell/asr1/RESULTS.md. To make `RESULTS.md`, you can use https://github.com/espnet/espnet/blob/v.0.4.0/utils/get_sys_info.sh and https://github.com/espnet/espnet/blob/v.0.4.0/utils/pack_model.sh

If you want to upload your models (I really encourage you to do), please let me know. I'll give you permission for uploading the models.

27jiangziyan:
Yes, I will try several different training configs. The new markdown RESULTS file looks cool. I'll use this style for my results.

sw005320:
Thanks, @27jiangziyan for your great contributions. Also, it would be great if you update the results and also make `RESULTS.md` in the later PR.

27jiangziyan:
Sure, I'm waiting for the results of two small models. I'll update the result in another PR by today.

PR #826, opened 2019-06-10
Bug fix: egs/chime4/asr1_multich/conf/train.yaml

kamo-naoyuki:
I forgot to write `use-frontend`


sw005320:
Thanks! @Jzmo, please reflect it in your AMI recipe #819.

Jzmo:
Okay, AMI recipe has included `use-frontend` in `run.sh`

sw005320:
Oh, I see. Then, you just move this option to the config file like this PR.

Jzmo:
Got it. I would update this option together with the RESULT.md after I got it.

PR #809, opened 2019-06-06
Demo of streaming ASR

takenori-y:
I modified `recog_wav.sh` in the tedlium recipe to demonstrate streaming ASR using publicly available models. This relates to #787.

PR #830, opened 2019-06-11
fix specaug performance issue

bobchennan:
fixed specaug performance issue caused by commit 68c0351. 
However I still need to fix shape problem and check performance.

Current performance for 0.4.0 with specaug:

System(#iters) | CER(dev) | WER(dev) | CER(eval) | WER(eval)
-- | -- | -- | -- | --
rnn(22) | 3.4 | 7.4 | 2.4 | 5.0
transformer(45) | 3.8 | 8.4 | 2.5 | 5.1

One way to improve it is to set `patience=0` and train with more iterations.
Currently I resumed training of transformer.

Transformer training curve before early-stopping:
image

### DIFF
--- a/egs/wsj/asr1/conf/tuning/train_rnn.yaml
+++ b/egs/wsj/asr1/conf/tuning/train_rnn.yaml
@@ -36,6 +36,3 @@ patience: 3
 
 # scheduled sampling option
 sampling-probability: 0.0
-
-# SpecAug
-use-specaug: False
--- a/egs/wsj/asr1/conf/tuning/train_rnn.yaml
+++ b/egs/wsj/asr1/conf/tuning/train_rnn.yaml
@@ -36,3 +36,6 @@ patience: 3
 
 # scheduled sampling option
 sampling-probability: 0.0
+
+# SpecAug
+use-specaug: True
--- a/egs/wsj/asr1/conf/tuning/train_rnn.yaml
+++ b/egs/wsj/asr1/conf/tuning/train_rnn.yaml
@@ -38,4 +38,4 @@ patience: 3
 sampling-probability: 0.0
 
 # SpecAug
-use-specaug: True
+use-specaug: False
--- a/espnet/asr/pytorch_backend/asr.py
+++ b/espnet/asr/pytorch_backend/asr.py
@@ -41,6 +41,7 @@ from espnet.utils.cli_utils import FileWriterWrapper
 from espnet.utils.deterministic_utils import set_deterministic_pytorch
 from espnet.utils.dynamic_import import dynamic_import
 from espnet.utils.io_utils import LoadInputsAndTargets
+from espnet.utils.spec_augment import specaug
 from espnet.utils.training.batchfy import make_batchset
 from espnet.utils.training.iterators import ShufflingEnabler
 from espnet.utils.training.iterators import ToggleableShufflingMultiprocessIterator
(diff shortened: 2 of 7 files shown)


espnet/asr/pytorch_backend/asr.py
```diff
@@ -560,7 +565,6 @@ def recog(args):
                         if hyps is not None:
                             text = ''.join([train_args.char_list[int(x)]
                                             for x in hyps[0]['yseq'][1:-1] if int(x) != -1])
-                            text = text.replace('\u2581', ' ').strip()  # for SentencePiece
```
sw005320: This was introduced in #809. Please recover it.

ShigekiKarita:
Can you confirm results w/o spectaug in your env?

bobchennan:
some updates:

System(#iters) | WER dev93 | WER eval92
-- | -- | --
Transformer | 8.2 | 4.8
Transformer+specaug | 7.3 | 4.6
RNN | 8.6 | 5.1
RNN+specaug | |

egs/wsj/asr1/conf/tuning/train_rnn.yaml
```diff
@@ -36,3 +36,6 @@ patience: 3
 
 # scheduled sampling option
 sampling-probability: 0.0
+
+# SpecAug
+use-specaug: True
```
ShigekiKarita: Can you update RESULTS.md if you wanna make this default

bobchennan: Sure I will update code and RESULTS.md later.

bobchennan:
I am training last exp which disables label smoothing after warmup.

Unless it gives different results, my conclusion is that time warping is really helpful for RNN model since it helps to learn certain gates by disturbing speed rate; while this is not helpful for transformer since it always consider all frames simultaneously. For transformer time and frequency masking improves a little bit. Conclusion might be different for larger corpus since maybe we overfitted too much on dev.

bobchennan:
System(#iters) | WER dev93 | WER eval92
-- | -- | --
Transformer | 8.2 | 4.8
Transformer+FM+TM | 7.3 | 4.6
Transformer+FM+TM+TW | 7.0 | 5.0
Transformer+FM+TM+TW w.o. label smoothing | 7.3 | 4.7
RNN | 8.6 | 5.1
RNN+FM+TM | 7.8 | 5.2
RNN+FM+TM+TW | 7.3 | 4.6

ShigekiKarita:
@bobchennan as you mentioned before, #745 implements specaug as preprocessing. And it has already utilized your `sparse_image_warp` like this
```yaml
process:
  # these three processes are a.k.a. SpecAugument
  - type: "time_warp"
    max_time_warp: 5
    inplace: true
    mode: "sparse_image_warp"  # "PIL" is faster
  - type: "freq_mask"
    F: 30
    n_mask: 2
    inplace: true
    replace_with_zero: false
  - type: "time_mask"
    T: 40
    n_mask: 2
    inplace: true
    replace_with_zero: false
```
see also https://github.com/espnet/espnet/blob/d988a9298479c1550743fbd50334c735e0ab9c65/espnet/transform/spec_augment.py#L41-L44

Therefore I do not think `--use-specaug` option is needed because it is not configurable like `--preprocess-conf`. If you agree, I would like to request these changes 
- remove `use_specaug` related codes in `asr_train.py` and `asr.py`
- add new config `egs/wsj/asr1/conf/specaug_sparse.yaml` with your new setting?

bobchennan:
Sure. #745 is too long so I didn't notice that.

bobchennan:
@ShigekiKarita 
I changed it. It may still be different with results I reported since current implementation of time masking and freq masking is actually different with what I used. Pls check espnet/espnet/transform/spec_augment.py and espnet/utils/spec_augment.py for details.

espnet/asr/pytorch_backend/asr.py
sw005320: @bobchennan, can you fix this?

Issue #733, opened 2019-05-17
GPU issues with new devices [RTX 2080Ti/V100] inside docker containers


Fhrozen:
I've been having a problem to execute the training in a server with V100, for chainer with mtlalpha > 0. I think that the code of WarpCTC is not supporting them. 
I just got a new gpu (RTX2080Ti) and is showing the same issue:
```
Traceback (most recent call last):
  File "/espnet/egs/an4/asr1/../../../espnet/bin/asr_train.py", line 274, in <module>
    main(sys.argv[1:])
  File "/espnet/egs/an4/asr1/../../../espnet/bin/asr_train.py", line 259, in main
    train(args)
  File "/espnet/espnet/asr/chainer_backend/asr.py", line 470, in train
    trainer.run()
  File "/espnet/tools/venv/lib/python3.7/site-packages/chainer/training/trainer.py", line 329, in run
    six.reraise(*sys.exc_info())
  File "/espnet/tools/venv/lib/python3.7/site-packages/six.py", line 693, in reraise
    raise value
  File "/espnet/tools/venv/lib/python3.7/site-packages/chainer/training/trainer.py", line 315, in run
    update()
  File "/espnet/tools/venv/lib/python3.7/site-packages/chainer/training/updaters/standard_updater.py", line 165, in update
    self.update_core()
  File "/espnet/espnet/asr/chainer_backend/asr.py", line 100, in update_core
    loss = optimizer.target(*x) / self.accum_grad
  File "/espnet/espnet/nets/chainer_backend/e2e_asr.py", line 90, in __call__
    loss_ctc = self.ctc(hs, ys)
  File "/espnet/espnet/nets/chainer_backend/ctc.py", line 95, in __call__
    self.loss = warp_ctc(y_hat, ilens, [cuda.to_cpu(l.data) for l in ys])[0]
  File "/espnet/tools/venv/lib/python3.7/site-packages/chainer_ctc/warpctc.py", line 125, in ctc
    return CTC(seq_lengths, labels)(x)
  File "/espnet/tools/venv/lib/python3.7/site-packages/chainer/function.py", line 235, in __call__
    ret = node.apply(inputs)
  File "/espnet/tools/venv/lib/python3.7/site-packages/chainer/function_node.py", line 263, in apply
    outputs = self.forward(in_data)
  File "/espnet/tools/venv/lib/python3.7/site-packages/chainer/function.py", line 135, in forward
    return self._function.forward(inputs)
[...]

PR #836, opened 2019-06-13
[WIP] Support new version in Docker containers

Fhrozen:
I just finish to build the containers with the new fixes in warp-ctc and chainer-ctc (Details at #733)
The containers are been uploaded to the hub and then test on a PC with RTX 2080Ti and CUDA 10.1.
The server with V100 is been used so not sure when will be available.
Once the tests on RTX are done the [WIP] will be remove and this little PR will be ready to merge.


Fhrozen:
Tests finished. RTX with CUDA 10.1 now is working with and observation that the pytorch backend is still not supported. I tested a docker container with pytorch 1.0.1 and it has not problem. So suppose that this error will be resolve when v.0.4.0 is merged.

Fhrozen:
@sw005320 I just finish to build and test a docker container with ubuntu 18.04 and cuda 10.0 without problem.
Some of the library from runtime has been remove given that they are no longer required.
Also, I tried to keep the same building route as NVIDIA and for ubuntu 18.04, the containers are built for CUDA versions 9.2, 10.0, and 10.1. So the number of container will be reduced to only 4 (3 by the moment) cpu and the versions detailed previously.

Let me know if there is any possible trouble with this. 
The recipes are ready so once you give the confirmation I will build them and push in to the Hub, and this PR will be ready to merge.

PR #840, opened 2019-06-13
Update hkust recipe with transformer

sw005320:
- almost comparable to Kaldi (Kaldi 23.7 vs. ESPnet transformer 23.5)
- [x] update symlink
- [x] will redo recognition experiments around 20 epochs
- [x] update main README.md


sw005320:
This recipe seems to be over-trained after 20 epochs. We may need some tuning.

PR #843, opened 2019-06-14
Changed to use pip-install for pytorch_wpe

kamo-naoyuki:
Removed the wpe implementation and changed to install from https://github.com/nttcslab-sp/dnn_wpe


sw005320:
Could you tell me the intention of this PR? Personally, I think it is better to include the WPE part for ESPnet to keep an end-to-end concept, which combines all speech processing computational graphs as one graph toward global optimization. For that purpose, it's better to include it. I'm expecting there remain interesting studies of how we design the dereverberation subnetwork by modifying it and it would be good to include it under the ESPnet.

kamo-naoyuki:
First of all, there is an internal reason in our organization to remove WPE implementation, I expected you know it.  I put the code by mistake.
This PR just changes the module path of WPE to the external module. The DNN connection to WPE is still existing.  Maybe there are no problem for studying.

sw005320:
OK, I understood.

PR #854, opened 2019-06-15
test: avoid using grep and sed in subprocess and use python stdlib instead

r9y9:
This makes the code more readable and platform independent. With this change, I can run all unit tests successfully on osx.


sw005320:
Thanks!

PR #774, opened 2019-05-29
[WIP] Transformer Chainer

Fhrozen:
This is the second part of the updates for the transformer with Chainer:
- [x] Reflect pytorch code
- [x] Joint CTC/att training
- [x] Support Beam search
- [x] Support LM integration
- [x] Add CER for Transformer Chainer (#755)
- [x] Fix CER for Transformer pytorch (#755)
- [x] Fix accum grad #777 
- [x] Mismatch dim in Conv2D sub (Pytorch) #822 
- [x] Common CER computation for Transformer - Pytorch/Chainer
- ~~Common CER computation for RNN - Pytorch/Chainer~~ (Leave for another PR)
- ~~Format __call__ inside chainer transformer~~ (Leave for another PR)

PR #857, opened 2019-06-15
CER/WER & CER_CTC in Transformer pytorch

Fhrozen:
Currently, testing on cpu.
This implements a common function for obtain CER & WER  in transformer
It is only implemented on pytorch backend. 
Chainer will be supported in #774.

- [x] Common Function CER CTC+ CER WER


Fhrozen:
@sw005320, Just finished this PR. Please check it and let me know if there is a problem for merge.
For the first epochs, the CER is displayed as following:
cer.
I compared with the master training and did not observe any bottleneck or increasing on the training time, at least for the first epochs.

espnet/nets/pytorch_backend/e2e_asr_transformer.py
```diff
@@ -181,14 +193,48 @@ def forward(self, xs_pad, ilens, ys_pad):
...
-        if self.ctc is None:
+        if self.mtlalpha == 0.0:
             loss_ctc = None
+            cer_ctc = None
         else:
             batch_size = xs_pad.size(0)
             hs_len = hs_mask.view(batch_size, -1).sum(1)
             loss_ctc = self.ctc(hs_pad.view(batch_size, -1, self.adim), hs_len, ys_pad)
 
+            cers, char_ref_lens = [], []
```
sw005320: Can you make these entire lines (204 -- 237) as one function (class member function or function in some common directory)?
I want to make the main loop simple.

espnet/nets/pytorch_backend/e2e_asr_transformer.py
```diff
@@ -201,39 +198,15 @@ def forward(self, xs_pad, ilens, ys_pad):
...
-                    char_ref_lens.append(len(ref_chars))
-
-            cer_ctc = float(sum(cers)) / sum(char_ref_lens) if cers else None
+            ys_hat = self.ctc.argmax(hs_pad.view(batch_size, -1, self.adim)).data
+            cer_ctc = self.error_calculator(ys_hat, ys_pad, is_ctc=True)
 
         # 5. compute cer/wer
-        if self.training or not (self.report_cer or self.report_wer):
+        if self.training or self.error_calculator is None:
             cer, wer = 0.0, 0.0
```
sw005320: This may confuse people.
`None` is better?
At least `0.0` is wrong.

Fhrozen:
@sw005320 Just finished the common functions. Let me know if this was what you are looking for.
By the moment, it increases the elapsed time by 5.5 secs per 100 iters. So in a training of 20 hrs it will become 24hrs aprox.
I will try to see if it is possible to reduce this elapsed time.

espnet/nets/pytorch_backend/e2e_asr_transformer.py
Fhrozen: Ok.
I was planning to apply it but in the other PR. #774. Should I apply this here? or keep it as part of #774.

sw005320:
> By the moment, it increases the elapsed time by 5.5 secs per 100 iters. So in a training of 20 hrs it will become 24hrs approx.

Hmm. It's a bit more than I expected...

espnet/nets/pytorch_backend/e2e_asr_transformer.py
sw005320: Oh, that's better. Let's do it there.

Fhrozen:
After testing several coding combinations i found that this is the best. Currently, it takes 2 ~ 3 secs additional to the one without CER/CTC.
I tested master and each 100 iters took around 24 secs, while the previous code took 31 secs (almost 7 additional secs per 100 iters). 
The main problem was that cer/ctc was passing the data in gpu format and then each values were converted to cpu int. Passing the whole data at once is faster. 

@sw005320 let me know if this is acceptable, and please check it to merge so I can update this in #774  to finish it.

espnet/nets/pytorch_backend/e2e_asr_transformer.py
```diff
@@ -110,6 +110,15 @@ def __init__(self, idim, odim, args, ignore_id=-1):
         else:
             self.ctc = None
 
+        if 'report_cer' in vars(args) or args.mtlalpha > 0.0:
```
ShigekiKarita: why not `if args.report_cer`?

espnet/nets/pytorch_backend/e2e_asr_transformer.py
```diff
@@ -181,14 +190,24 @@ def forward(self, xs_pad, ilens, ys_pad):
...
 
+            ys_hat = self.ctc.argmax(hs_pad.view(batch_size, -1, self.adim)).data
+            cer_ctc = self.error_calculator(ys_hat.cpu(), ys_pad.cpu(), is_ctc=True)
+
+        # 5. compute cer/wer
+        if self.training or self.error_calculator is None:
+            cer, wer = None, None
+        else:
+            ys_hat = pred_pad.argmax(dim=-1)
+            cer, wer = self.error_calculator(ys_hat.cpu(), ys_pad.cpu())
```
ShigekiKarita: I think these `cer, wer` can be obtained even if `args.mtlalpha == 0`. If so, remove `if args.mtlapha > 0` in `__init__`.

ShigekiKarita:
I think the performance overhead is OK because we can disable it by `report-cer: False` `report-wer: False`. Are they default False?

espnet/nets/pytorch_backend/e2e_asr_transformer.py
ShigekiKarita: What happens if `report_cer == False && report_wer == True`?

Fhrozen:
@ yes. they already set to `False`
https://github.com/espnet/espnet/blob/fcf4c8ceeb413cea333ed98dc6c9fb5339bf1150/espnet/bin/asr_train.py#L125-L128

espnet/nets/pytorch_backend/e2e_asr_transformer.py
Fhrozen: I think that this `if` detects that `report_cer` has been declared in `args` regardless its value ( in case of `None`)

espnet/nets/pytorch_backend/e2e_asr_transformer.py
Fhrozen: The `if` in main allows to avoid adding the `error_calculator` if there is no `report_cer` in `args` and the `mtlalpha` is equal to zero. I will change the way to add `error_calculator` for `False` cases and `mtlalpha=0` then.

ShigekiKarita: Why can `args` have no `eport_cer`? It is statically added in `asr_train.py`.

espnet/nets/pytorch_backend/e2e_asr_transformer.py
ShigekiKarita: I think this is better because `report_cer` and `report_wer` should not be None as defined in `asr_train.py`.
```python
if args.report_cer or args.report_wer:
    self.error_calculator(...)
```

espnet/nets/pytorch_backend/e2e_asr_transformer.py
Fhrozen: Yes, but when I wrote the code for transformer I migrated it from the `e2e_asr` where is almost the same `if report_cer is not in vars(args)`:
https://github.com/espnet/espnet/blob/fcf4c8ceeb413cea333ed98dc6c9fb5339bf1150/espnet/nets/pytorch_backend/e2e_asr.py#L121-L134

I will modify this directly in #774

ShigekiKarita: Please do not spread this... You can do the right thing here because it does not have a breaking change. I think you can also delete `self.recog_args` in Transformer.

Fhrozen: got it.

Fhrozen:
@ShigekiKarita I just added the modifications. Please let me know if there is something missing or still wrong in the code. BTW, i think the reason why is there was the `if report_cer is not in vars(args)` is because of the tests. They were not being included in the Namespace. I just added to fix this.

espnet/nets/e2e_asr_common.py
```diff
@@ -80,3 +83,104 @@ def get_vgg2l_odim(idim, in_channel=3, out_channel=128):
     idim = np.ceil(np.array(idim, dtype=np.float32) / 2)  # 1st max pooling
     idim = np.ceil(np.array(idim, dtype=np.float32) / 2)  # 2nd max pooling
     return int(idim) * out_channel  # numer of channels
+
+
+class ER_Calculator(object):
```
sw005320: We don't use `_` in the class name. How about `ErrorCalculator`?

PR #858, opened 2019-06-15
Move the dependency links to github from Makefile to setup.py

kamo-naoyuki:
I moved python modules originated from github repos from Makefile to setup.py
This setup.py uses `module@URL` syntax https://github.com/pypa/pip/pull/4175, and it was enabled on `pip==19`, which was released on Jan 2019. 

I also added python version check.

If `warp-ctc` and `chainer-ctc` can be installed by such way, then we can freeze all python dependencies in setup.py. It may be useful for google colab environment.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -1,15 +1,20 @@
 #!/usr/bin/env python
-import io
+from distutils.version import LooseVersion
 import os
+import pip
 from setuptools import find_packages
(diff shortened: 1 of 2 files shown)


ShigekiKarita:
Wonderful. In fact, this can also accelerate CI in terms of pip cache.

PR #866, opened 2019-06-16
Fix: espnet/transform/transformation.py

kamo-naoyuki:
This line doesn't work with function having variable keyword arguments.


kamo-naoyuki:
I don't know why test failed. Maybe randomness.

ShigekiKarita:
Thanks. LGTM

Issue #764, opened 2019-05-26
Why the multichannel data is randomly processed by the " chime4/asr1_multich " in v0.4.0


Rpersie:
I check the procedures in the  chime4/asr1_multich of v0.4.0 and have some questions about the code in line 90 of espnet/espnet/nets/pytorch_backend/frontends/frontend.py.
![image](https://user-images.githubusercontent.com/12673020/58379540-73435700-7fd7-11e9-9bcb-062a2760244f.png)

It means that in the training "use_beamformer"  is randomly set true or false. So not all the data is processed by  the beamformer.   
I just change the code and find the loss is worse than the original code. Can anyone tell me why the  data is processed like this. In my opinion, the beamformer is not trained by all the data. The results can not perform better than using all the data 

Thank you

kamo-naoyuki:
It have made better result experimentally, just following previous research. 

It is said that if always the beamformer part is trained, the backend asr system cannot be trained with noisy data, without any enhancement, then it  weakens the base recognition performance for noisy data.

And, there is no need to care about not using all data, in an epoch, because almost all data are given over all epoch.

Rpersie:
How about the gap? It will behind the custom delay-sum beamformer ?  You do this just to improve the ASR performance just like the MCT.

kamo-naoyuki:
I forgot the precise value, maybe 3~5% worse in CER.

Rpersie:
It will behind the custom delay-sum beamformer ?

kamo-naoyuki:
I compared it with only beamformit, which is the baseline system for chime4, and got worse result than it.

Rpersie:
I have run the baseline many times, but it is different. Sometimes, the results vary greatly.
![image](https://user-images.githubusercontent.com/12673020/59001294-932d1300-8840-11e9-9157-3c1fa9c03a22.png)
Do you think it may cause by this random?
If I just split the multichannel data to single channel and  dump them to json file, it will perform better and stably than the random training methods ?
Thank you

sw005320:
> Do you think it may cause by this random?

Hmm.
We try to make things deterministic (specify a fixed seed for python, numpy, pytorch, chainer and use the deterministic mode).
We still observe some slight difference in the same machine and environment, but this difference seems to be too large.
@kamo-naoyuki, do you have comments?

kamo-naoyuki:
[...]

PR #875, opened 2019-06-18
Fix bugs in frontend

kamo-naoyuki:
- Changed to use wideband PESQ(2005) instead of PESQ(2001) for evaluation 
- Fix zero padding bug (zero padding after sigmoid instead of sigmoid after zero padding) #764
- Update to README.md. anc write wideband PESQ result.

This zero padding bug didn't affect  the results for my environment.


ShigekiKarita:
ignore codecov failure at this patch

sw005320:
Maybe better to add some test scripts for this?

kamo-naoyuki:
What test?

sw005320:
`utils/eval-source-separation.py`, but I think this is difficult because we need to install other packages.

sw005320:
Thanks a lot!

PR #877, opened 2019-06-18
Script for docker builds from the local repo

lumaku:
When building the docker container on a local machine, the espnet source is downloaded from the github espnet master branch. But I prefer to use the local espnet version for two reasons:

1. I can write my own modifications on the espnet code, the build environment, etc, and test it in the docker container. The current script `build_and_push.sh` does not import these.

2. reproducability: I can go back to an old espnet version and test the neural network with an older version of a library.

That is why I modified `build_and_push.sh` to make local builds. As `docker build` does not have direct read access to the espnet main folder, I implemented a workaround: The espnet source code is imported through a git archive based on git HEAD (the previous commit), and copied over within a file.


sw005320:
@lumaku, thanks!
I like your idea, but I simply don't have a docker environment and cannot test it.
@Fhrozen, I'd like to ask you to check this PR.

Fhrozen:
@lumaku, thank you very much for your support.
It will be better if you can keep a uniform way for using variables:`${foo}`, some of them are called as `$foo`
I will test it tomorrow to check if there is any problem with the current build. 

@sw005320 It will be better if you check #836 first to merge and then joint it with this PR, so it can keep the updates.

sw005320:
> @sw005320 It will be better if you check #836 first to merge and then joint it with this PR, so it can keep the updates.

Done.

lumaku:
Happy to contribute! :smile: For testing, I encountered following issue, but it might work on your system, which depends on your docker/nvidia-docker configuration.

When I build the docker container on my system, executing the commands in `build_and_push.sh`, the container build for the GPU container is aborted in the *last stage* by an error in `check_install.py`.

By inspection of this issue, importing Cupy throws an error. Cupy was compiled correctly, and is installed; CUDA was correctly installed and env variables were correctly set. The required dependencies on my machine are fulfilled (nvidia-docker 2.0.3-4, nvidia-driver 430.26, docker 18.09.6).

I find that the Cupy error occurs because it can not find my nvidia gpu (listed using `ls -la /dev | grep nvidia`), as it was not passed through the container at the initialization of the docker container.

My workaround for this issue was to modify `check_install.py` to print a warning instead of `sys.exit(1)` during docker build using the script in this pull request. Afterwards, I started the gpu container with `nvidia-docker` (that mounts the gpu device) and the `check_install.py` could successfully import cupy. I did not include the changes to `check_install.py` in the current pull request, but made a ephemeral testing branch.

Fhrozen:
@lukuma the reporting error has been observer for a while, and that is why the code has implemented:

https://github.com/espnet/espnet/blob/e2273884e98e7dee0bb535501c35ee2fdff1b070/docker/prebuilt/devel/Dockerfile#L18-L25

lumaku:
I merged the `build_and_push.sh` from master, the local docker build for me was OK. @Fhrozen, I saw that you added "-u18" to the tags, I assumed only for naming test images without overwriting the docker hub images, but now testing is possible without pushing, so I did not adopt this change.

Fhrozen:
@lumaku , No, it is not because a pushing problem. It is because the from the current version the containers will be only built for Ubuntu 18.04 and the containers based in ubuntu 16 will not receive any longer support. In case people want to use the previous version containers, I am letting them in the Hub and the current containers will have a -u18 tag.

lumaku:
The tags for the Docker Hub images are as you specified them. For locally built images, I did not add `-u18` at the end.

Please note that in this configuration, `build_and_push.py local $ver` will use the Docker Hub *runtime* image with Ubuntu 16, whereas `build_and_push.py fully_local $ver` will build the *runtime* image locally based on the Ubuntu 18 image.

sw005320:
> Please note that in this configuration, build_and_push.py local $ver will use the Docker Hub runtime image with Ubuntu 16, whereas build_and_push.py fully_local $ver will build the runtime image locally based on the Ubuntu 18 image.

Thanks. Can you add these comments to `docker/build_and_push.sh` and https://github.com/espnet/espnet/blob/master/docker/README.md? Also, could you add some descriptions about your local repo functions to this `README.md`?

lumaku:
Updated the description. Also renamed `build_and_push.sh` to `build.sh`, as it does no longer automatically push. I think that there might be one rough edge left:

How the `run.sh` handles the docker containers is still a little bit convoluted. To start a local container with `run.sh`, one has to clear `--docker_os` and write the correct configuration that fits to the tag which was outputted at the end of the local docker build.

Maybe one way to workaround/solve this would be to add a "local" tag in the build and run script. What is your suggestion?

On a side note, I personally would prefer a docker configuration file that I pass over as first argument to the `run.sh`, where my variables and folder mounts are specified. The parameter list for `run.sh` is already very long for one terminal line and such configurations would be easier to modify in a text file than on command line.

sw005320:
Thanks! @Fhrozen, can you check it again?
If it's OK, I'll merge it.

Fhrozen:
@sw005330. Sorry I could not test due to the Docker pcs are busy with other tasks. I will try to test this once the tasks are done

sw005320:
no problem. It's not urgent.

lumaku:
I appended "local" to the tags. There should be no changes to `run.sh` necessary. You can now start a local build with:
```
$ ./run.sh --docker_os local --docker_cuda 10.0 ...
```

docker/build.sh
```diff
@@ -0,0 +1,169 @@
...
+
+# 2019, Nelson Yalta
+# 2019, Ludwig Kürzinger, Technische Universität München
+
+
+SCRIPTPATH="$( cd "$(dirname "$0")" ; pwd -P )"
+
+tags="cpu
+      gpu-cuda9.2-cudnn7
+      gpu-cuda10.0-cudnn7"
```
Fhrozen: please,  replace tags with:
```
tags="cpu-u18
      gpu-cuda10.0-cudnn7-u18"
```

sw005320:
Thanks a lot!!!

PR #881, opened 2019-06-19
gradient noise injection from std normal dis

creatorscan:
Read this post on gradient noise http://deliprao.com/archives/153 and found it to be effective to handle overfitting especially for librispeech.

### DIFF
--- a/espnet/asr/asr_utils.py
+++ b/espnet/asr/asr_utils.py
@@ -237,6 +237,20 @@ def _torch_snapshot_object(trainer, target, filename, savefun):
         shutil.rmtree(tmpdir)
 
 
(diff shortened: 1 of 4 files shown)


sw005320:
Very cool!
Is it possible to add a simple test for it?
Also, can you make `0.55` as a variable with default argument or constant variable and put somewhere (do not have to make it as an option)?

sw005320:
Thanks. I'll use it. The default value is fine, right?

creatorscan:
> Thanks. I'll use it. The default value is fine, right?

Yes Shinji. It worked for me. As I preferred to use high std dev.

PR #892, opened 2019-06-20
Fix tgz typo

naxingyu:
This might cause error if download is interrupted, because wget will rename the file if old one is not deleted.


sw005320:
thanks!

PR #906, opened 2019-06-21
grad_noise fixes

creatorscan:
changes to eta and itr count to improve val accuracy in early epochs

### DIFF
--- a/espnet/asr/pytorch_backend/asr.py
+++ b/espnet/asr/pytorch_backend/asr.py
@@ -150,8 +150,8 @@ class CustomUpdater(training.StandardUpdater):
         # gradient noise injection
         if self.grad_noise:
             from espnet.asr.asr_utils import add_gradient_noise
-            eta = 0.5  # {0.01,0.3,1.0}
-            duration = 1000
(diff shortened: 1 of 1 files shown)


espnet/asr/pytorch_backend/asr.py
```diff
@@ -150,8 +150,8 @@ def update_core(self):
         # gradient noise injection
         if self.grad_noise:
             from espnet.asr.asr_utils import add_gradient_noise
-            eta = 0.5  # {0.01,0.3,1.0}
-            duration = 1000
+            eta = 1.0  # {0.01,0.3,1.0}
+            duration = 100
             itr = (self.iteration // duration) + 1
             add_gradient_noise(self.model, itr, eta)
```
sw005320: How about change the interface like that?
`add_gradient_noise(self.model, self.iteration, duration=100, eta=0.5, scale_factor=0.55)`

Then, you compute the sigma in `def add_gradient_noise(model, iteration, duration=100, eta=0.5, scale_factor=0.55)`
You can add the explanation of `duration`, `eta`, and `scale_factor` at docstring.

sw005320:
This change is included in #912

Issue #902, opened 2019-06-21
test error for test_eval_perm_free_error_py.bats


sw005320:
@simpleoier, it's not always, but I observed that `test_eval_perm_free_error_py.bats` will be failed sometimes.
Do you know what happned?
Can you fix it?
```
not ok 11 eval_perm_free_error.sh
# (in test file test_utils/test_eval_perm_free_error_py.bats, line 102)
#   `diff ${tmpdir}/min_perm_result.wrd.txt ${tmpdir}/result_wer.txt' failed
# 1,2c1,2
# < Total Scores: (#C #S #D #I) 30 46 46 27
# < Error Rate:   97.54
# ---
# > Total Scores: (#C #S #D #I) 30 54 60 33
# > Error Rate:   102.08
```

simpleoier:
I will take a look at it.

ShigekiKarita:
Maybe we should set seed
```python
np.random.seed(0)
```

simpleoier:
> Maybe we should set seed
> 
> ```python
> np.random.seed(0)
> ```

Thanks, this could solve the problem.

I guess the reason is because:
in `utils/min_perm_wer.py`, the action I used to choose the best permutation is:

> ```python
> min_idx, min_v = min(enumerate(error_rate), key=lambda x: x[1])
> ```

While in `test_utils/test_eval_perm_free_error_py.bats`, the action I used to choose the best permutation was:

> ```python
> if wers[0] < wers[1]:
>     ret = wers[0]
> else:
>     ret = wers[0]
> ```

The former method uses the leftmost smallest "wer", however, the later one uses the rightmost smallest "wer".
When there are multiple utterances, the results may differ in the case where ```wers[0] == wers[1]```, but the numbers of errors and seq-length are different. I will change the if conditions or use the same way as the former methods.

simpleoier:
This difference does not affect the real ASR performance, because the seq-length of every reference utterance is fixed. In this test script, I randomly generated the `(C S D I)` numbers. This leads to the fact that even though the wers are the same, the `(C+S+D)` and `(S+D+I)` may be different.

sw005320:
OK, thanks.
Please make a PR to fix this test.

sw005320:
#908 fixed it.

PR #908, opened 2019-06-21
Modified if conditions

simpleoier:
fixed the bug in test_eval_perm_free_error.py, related to #902


sw005320:
Thanks for your quick action!

PR #912, opened 2019-06-21
Fix grad noise v2

sw005320:
- follow the new default parameters by #906 
- modify the interface of the `add_gradient_noise` function
- add a new style documentation
- fix to use `.to(param.device)` istead of `.cuda()`


sw005320:
@creatorscan, `eta / interval ** scale_factor` and `(eta / interval) ** scale_factor`, which is correct? Can you tell me the reference about this?

sw005320:
> @creatorscan, eta / interval ** scale_factor and (eta / interval) ** scale_factor, which is correct? Can you tell me the reference about this?

I got the original reference. It seems that `eta / interval ** scale_factor ` is correct.
The second question is that we may put this noise injection right before `optimizer.step()` because now we use the gradient accumulation and if we use it in the current line, the noise injection is performed #gradient_accum times, which yields too large noises.

PR #913, opened 2019-06-22
Fix plot fail

### LINKED ISSUE #904: matplotlib.use('Agg') fail
It result in plot fail 
```
_tkinter.TclError: no display name and no $DISPLAY environment variable
```
I fixed this by applying a patch on `espnet/nets/pytorch_backend/transformer/plot.py`
```
@@ -1,5 +1,6 @@
 import logging
-
+import matplotlib
+matplotlib.use('Agg')
 import matplotlib.pyplot as plt

 from espnet.asr import asr_utils
```

naxingyu:
This patch fixes #904

### DIFF
--- a/espnet/nets/pytorch_backend/transformer/plot.py
+++ b/espnet/nets/pytorch_backend/transformer/plot.py
@@ -1,5 +1,7 @@
 import logging
 
+import matplotlib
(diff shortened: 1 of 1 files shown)


sw005320:
Thanks!

Issue #822, opened 2019-06-08
Output dimension of Conv2dSubsampling


kan-bayashi:
`Conv2dSubsampling` layer cause dimension mismatch when idim is multiple of 4.
https://github.com/espnet/espnet/blob/577baff5c5d229c860ca9936daba3cb51b9857cb/espnet/nets/pytorch_backend/transformer/subsampling.py#L14-L25

```python
[ins] In [8]: conv = torch.nn.Conv2d(1, 1, 3, 2)

[ins] In [9]: input = torch.zeros((1, 1, 20, 20))

[ins] In [10]: out1 = conv(input)

[ins] In [11]: out1.shape
Out[11]: torch.Size([1, 1, 9, 9])

[ins] In [12]: out2 = conv(out1)

[ins] In [13]: out2.shape
Out[13]: torch.Size([1, 1, 4, 4])
```

Fhrozen:
let me fix this in #774

Fhrozen:
Fixed in #774

kan-bayashi:
Fixed #774 #822 #921

PR #921, opened 2019-06-25
Fix: Output dimension of Conv2dSubsampling #822

kamo-naoyuki:
@Fhrozen This bug shoould be fixed in master.

```
convodim(x) := floor((x - 2 * pad - dilation * (kernel - 1) -1) / stride + 1)
                     = floor((x - 3) / 2 + 1)
                     = floor((x - 1) / 2)
odim = convodim(convodim(idim))
          = floor((floor((idim - 1) / 2)) - 1) / 2)
```

The first equation is referred from https://pytorch.org/docs/stable/nn.html#convolution-layers


Fhrozen:
Thank you for your support. I was planning to add it from #774 to v.0.5.0 but it is better to fix it fast.

kan-bayashi:
Thanks!

PR #927, opened 2019-06-26
Change default verbose value in example

butsugiri:
I changed the default value of verbose, so that it is consistent to what is written in [README](https://github.com/espnet/espnet/blob/master/README.md#execution-of-example-scripts);
> With the default verbose (=0), it gives you the following information

Otherwise the expected outputs are buried in the verbose logs.


ShigekiKarita:
AN4 is mainly used for debugging because it is a toy task. Maybe we should update README instead of egs/an4/run.sh.

sw005320:
Thanks for pointing out the inconsistency. Let me think which is better.

sw005320:
I decided to change `README.md` description in #930 instead of changing the default option in the an4 recipe.

butsugiri:
OK. I will discard this PR.

Issue #928, opened 2019-06-26
Number of .sph in Tedlium3 


ShigekiKarita:
The number of .sph files is different between ESPnet and Kaldi
- 1514 in https://github.com/espnet/espnet/blob/master/egs/tedlium3/asr1/local/download_data.sh
- 2351 in https://github.com/kaldi-asr/kaldi/blob/master/egs/tedlium/s5_r3/local/download_data.sh

In addition, I found it is 4683 when I downloaded it yesterday. Is it growing?

cc: @Masao-Someki

ShigekiKarita:
In the project page https://lium.univ-lemans.fr/ted-lium3/
it says 
> 2351 audio talks in NIST sphere format (SPH), including talks from TED-LIUM 2: be careful, same talks but not same audio files (only these audio file must be used with the TED-LIUM 3 STM files)

maybe kaldi is correct but the actual number is different...

ShigekiKarita:
I found TEDLIUM_release-3 contains additional sph 
- `find db/TEDLIUM_release-3/ -name "*.sph" | wc -l` 4683
- `find db/TEDLIUM_release-3/data -name "*.sph" | wc -l` 2351

I will create PRs for ESPnet and Kaldi

PR #929, opened 2019-06-26
update tedlium3 download script

ShigekiKarita:
#928


sw005320:
Thanks! Now I fully understood why the tedlium3 recipe does not reach reasonable numbers.

PR #940, opened 2019-06-29
Fix plot attention for chainer transformer

Fhrozen:
It is not possible to run the chainer backend in master because of a bug on the plot attention part. Since #774 solves this i did not consider it, but it seems that #774 will take a longer time to be merge (ASRU deadline i suppose) So this will temporary solve the problem.

@sw005320  If this is accepted, I will need to PR another one to improve the performance of transformer in master.


sw005320:
Thanks @Fhrozen!
I like your ways of splitting a PR, which makes it easier to manage PRs.

PR #941, opened 2019-06-29
Implement add_arguments method of E2E for rnn.

kamo-naoyuki:
Current E2E rnn class is incompleted because it doesn't have add_arguments method, and  instead  of it, bin/asr_train.py has arguments for it.

I moved them to E2E and also modified E2E for mix.


espnet/nets/pytorch_backend/e2e_asr_transformer.py
```diff
@@ -57,6 +58,26 @@ def add_arguments(parser):
                            help='optimizer warmup steps')
         group.add_argument('--transformer-length-normalized-loss', default=True, type=strtobool,
                            help='normalize loss by length')
+
+        E2E_rnn.loss_add_arguments(parser)
```
ShigekiKarita: This args contains an unused option `lsm-type` because of my lazy implementation (sorry).
Maybe we can simply redefine them here like the other options.
```
        group.add_argument('--ctc_type', default='warpctc', type=str,
                           choices=['builtin', 'warpctc'],
                           help='Type of CTC implementation to calculate loss.')
        group.add_argument('--mtlalpha', default=0.5, type=float,
                           help='Multitask learning coefficient, alpha: alpha*ctc_loss + (1-alpha)*att_loss ')
        group.add_argument('--lsm-weight', default=0.0, type=float,
                           help='Label smoothing weight')
```

espnet/nets/pytorch_backend/e2e_asr_transformer.py
```diff
@@ -57,6 +58,26 @@ def add_arguments(parser):
...
+        E2E_rnn.recognition_add_arguments(parser)
```
ShigekiKarita: I think this `add_arguments` might not be defined in `E2E(_rnn)` because its option can be applied to any models.
How about move this to `ASRInterface`?

kamo-naoyuki: I'm not sure which is better in the current status. 
Maybe, if always it has these option, It's better to put it back to espnet/bin/asr_train.py,
because Interface is not allowed to have any concrete methods, strictly speaking.

Is is True? These option assumes the beam search based decoding.
Hmm, I'll follow you. Please decide.

ShigekiKarita: I got it. Let's put it back to espnet/bin/asr_train.py.

sw005320:
Do you think we should do such styles for all `bin/*.py`? This requires a lot of work although it's worth to do.

espnet/nets/pytorch_backend/e2e_asr_transformer.py
kamo-naoyuki: o.k.

kamo-naoyuki:
> Do you think we should do such styles for all `bin/*.py`? This requires a lot of work although it's worth to do.

I feel this is not so hard. Main changing can be done in an hour.
However, I feel that it's enough to refactor only asr in the current state, maybe.
Basically, abstraction should be done only when needs to add new drastic features.  I expect asr's option will increase from now. For example, rnnt or slightly different structure based on transformer could be thought, or semisupervised, and I also expect it will be important to pull such immature structure quickly from some view points. Just my perspective.

espnet/bin/asr_train.py
```diff
@@ -359,7 +269,7 @@ def main(cmd_args):
 
     # train
     logging.info('backend = ' + args.backend)
-    if args.num_spkrs == 1:
+    if not hasattr(args, 'num_spkrs'):
```
sw005320: Is this working?
We may need to change  https://github.com/espnet/espnet/blob/master/egs/wsj_mix/asr1/run.sh accordingly, right?
Can you do it? Or ask @simpleoier to do it.

sw005320:
Please also fix the conflict. It does not look so difficult.

espnet/bin/asr_train.py
kamo-naoyuki: I reverted this part because asr_mix should not be included in this PR.

espnet/nets/pytorch_backend/e2e_asr_transformer.py
kamo-naoyuki: `lsm-type` seems to be used.

https://github.com/espnet/espnet/blob/8cbde3843ebec4d882d5ebf35aebda7ca72b7e80/espnet/nets/chainer_backend/e2e_asr.py#L65-L69

I don't know how to refactor, so please fix it yourself if necessary.

sw005320:
Thanks a lot @kamo-naoyuki!

Issue #784, opened 2019-05-30
noam for RNN


sw005320:
@ShigekiKarita, noam optimizer should be used for RNN (e.g., specaugment) but the current implementation could only use it for the transformer architecture. Can you change it?

PR #943, opened 2019-06-30
Implement OptimizerFactoryInterface

kamo-naoyuki:
#784

- [x] OptInterface for pytorch
- [x] Noam for pytorch
- [x] OptInterface for chainer
- [x] Noam for chainer
- [x] test codes
- [x] Change all yaml file to support new option


kamo-naoyuki:
This PR is breaking the backward compatibility:

- `--opt` -> `--opt-module`
  - `--opt` can be still used because it can identify `--opt-module`
- `--transformer-lr` -> `--lr`
- `--transformer-warmup-steps` -> `--noam-warmup`

ShigekiKarita:
I prefer `OptimizerInterface` to `OptInterface`. Maybe `Optim` is ok but `Opt` is too short.

espnet/opts/pytorch_backend/noam.py
```diff
@@ -0,0 +1,99 @@
...
+        optimizer = cls.optimizer_class.get(parameters, args)
+        # Note(kamo): The original lr of optimizer is ingored in Noam, so reuse args.lr
+        return NoamOptimizer(optimizer,
+                             model_size=args.adim,
+                             warmup=args.noam_warmup,
+                             factor=args.lr)
+
+
+class NoamAdam(NoamOptBase):
+    optimizer_class = Adam
```
ShigekiKarita: Can you change `NoamAdam`'s beta to (0.9, 0.98) and eps to 1e-9 as well as the current code
https://github.com/espnet/espnet/blob/master/espnet/nets/pytorch_backend/transformer/optimizer.py#L57

kamo-naoyuki:
> I prefer `OptimizerInterface` to `OptInterface`. Maybe `Optim` is ok but `Opt` is too short.

This interface itself, unlike ASRInterface, is not "Optimizer", but, really, this class generates Optimizer.
So, OptimizerInterface may be confusing, but OptInterface is also not good, I agree. I don't have good naming.

ShigekiKarita:
I think this pattern is known as "factory". How about `OptimizerFactory` (and `AdamFactory`, etc) or something?

kamo-naoyuki:
I see, this class could be called as Factory Pattern.

ShigekiKarita:
In addition, I suggest `espnet.opt` -> `espnet.optimizers`

kamo-naoyuki:
I feel `opt` is enough clear because the abbreviation is used often. e.g. tf.contrib.opt, tensor.opt, while I do not see much `optim` except for pytorch context.

Maybe we should make espnet.optimizer_factories rather, because this class is not optimizer, again, but `espnet.optimizer_factories.pytorch_backend.optimizer_factory_interface.py` looks too too long, though the most ugly part is `pytorch_backend`.

Current epsnet style prefer short words, like `utils`, `nets`, so `opts` looks natural to me.

ShigekiKarita:
I think `espnet.xxxxxxx.OptimizerFactoryInterface` is too long too. How about `espnet.opt.interface.OptimzierFactory`? because this interface seems to be common both for pytorch and chainer and `OptimzierFactory` is enough for a typical base name for factory pattern.

kamo-naoyuki:
The interface is diferent, because pytorch.optim requires parameters.

ShigekiKarita:
Does python (3) still need a unique name?  `xxx.chainer_backend`  cannot be `xxx.chainer`?

kamo-naoyuki:
Python3 doesn't require it, probably, but i never test.

sw005320:
I like this idea in general. My concern is that people may get lost when they try to introduce another optimizer (e.g., rmsprop, etc.). But maybe OK.

sw005320:
I want to try to use the warm-up optimizer for RNN and want to merge this PR. Is it ready for the merge? Is your change applied to all existing config files?

kamo-naoyuki:
I think I have finished. Please fix if you met errors.

espnet/opts/pytorch_backend/noam.py
```diff
@@ -0,0 +1,104 @@
...
+
+from torch.nn import Parameter
+from torch.optim import Optimizer
+
+from espnet.opts.pytorch_backend.adadelta import AdadeltaFactory
+from espnet.opts.pytorch_backend.adam import AdamFactory
+from espnet.opts.pytorch_backend.optimizer_factory_interface import OptimizerFactoryInterface
+
+
+# FIXME(kamo): Should inherit torch.optim.Optimizer to satisfy strict type check
```
ShigekiKarita: Do you wanna fix this in this PR?

espnet/opts/pytorch_backend/noam.py
```diff
@@ -0,0 +1,104 @@
...
+        self._step = 0
+        self.warmup = warmup
+        self.factor = factor
+        self.model_size = model_size
+
+    @property
+    def param_groups(self):
+        return self.optimizer.param_groups
+
+    def step(self):
```
ShigekiKarita: Could you cover these methods in unittest?

test/test_optimizer_factory.py
```diff
@@ -0,0 +1,32 @@
...
+@pytest.mark.parametrize('opt_factory', [Adadelta_th, Adagrad_th, Adam_th,
+                                         NoamAdam_th])
+def test_pytorch_opts(opt_factory):
+    parser = argparse.ArgumentParser()
+    opt_factory.add_arguments(parser)
+    args = parser.parse_args([])
+    args.adim = 100
+
+    net = torch.nn.Linear(1, 1)
+    opt_factory.create(net.parameters(), args)
```
ShigekiKarita: Here. 
```suggestion
    net.weight.grad = torch.ones(1, 1)
    opt = opt_factory.create(net.parameters(), args)
    opt.step()
    opt.zero_grad()
    d = opt.state_dict()
    opt = opt_factory.create(net.parameters(), args)
    opt.load_state_dict(d)
    assert d == opt.state_dict()
```

test/test_optimizer_factory.py
```diff
@@ -0,0 +1,32 @@
...
+from espnet.opts.pytorch_backend.adam import AdamFactory as Adam_th
+from espnet.opts.pytorch_backend.noam import NoamAdamFactory as NoamAdam_th
+
+
+@pytest.mark.parametrize('opt_factory', [AdaDelta_ch, AdaGrad_ch, Adam_ch])
+def test_chainer_opts(opt_factory):
+    parser = argparse.ArgumentParser()
+    opt_factory.add_arguments(parser)
+    args = parser.parse_args([])
+    opt_factory.create(args)
```
ShigekiKarita: ditto

ShigekiKarita:
LGTM except for the comments

espnet/opts/pytorch_backend/noam.py
kamo-naoyuki: No.

sw005320:
Can you fix the conflict and also apply it to new recipes (e.g., `https://github.com/espnet/espnet/tree/master/egs/commonvoice/asr1`)? Then, I'll merge this PR to the v.0.6.0 branch.

PR #947, opened 2019-07-01
Refactor  FileReaderWrapper and  FileWriterWrapper

kamo-naoyuki:
The current `FileReaderWrapper` and  `FileWriterWrapper` in  espnet/utils/cli_utils.py are very unreadable and it's difficult to test. These terrible codes are my fault, really sorry. I was lazy.

- [x] Refactor cli_utils.py to class base style
- [x] Rewrite `utils` to use this function.
- [x] Create unit test codes


espnet/utils/cli_readers.py
```diff
@@ -0,0 +1,225 @@
...
+    def __iter__(self):
+        with kaldiio.ReadHelper(
+                self.rspecifier, segments=self.segments) as reader:
+            for key, array in reader:
+                if self.return_shape:
+                    array = array.shape
+                yield key, array
+
+
+class HDF5Reader:
```
ShigekiKarita: Can you split these files into submodules?

ShigekiKarita:
I think now `espnet.utils` is too big. How about move these reader/writer into `espnet.io`?

sw005320:
> I think now `espnet.utils` is too big. How about move these reader/writer into `espnet.io` , `espnet.readers` or `espnet.writers`?

I think the current style would be fine. It's still manageable. My concern is `utils`.

PR #950, opened 2019-07-04
Multilingual End-to-End Speech Translation

hirofumi0810:


hirofumi0810:
@sw005320 I think dividing to each direction is quite inefficient in this case due to the storage issue (this data consumes a few TB).

sw005320:
> @sw005320 I think dividing to each direction is quite inefficient in this case due to the storage issue (this data consumes a few TB).

I see but is there any other option?
We're keeping the recipe directory consistent.

PR #954, opened 2019-07-09
run.sh should exit if sourcing path.sh return error

naxingyu:
Librispeech requires `sentencepiece` tools to be installed. Script should exist if path sourcing fails, otherwise the master script won't succeed.


naxingyu:
@sw005320 A recent change of CI script request a module called `stempeg`. If USE_CONDA=true, it will request installing ffmpeg using conda. This results in CI failure. I manage to resolve it in this PR too. Now the build is passed.

kan-bayashi:
Thanks a lot!

PR #955, opened 2019-07-10
FiX ER calculator

### LINKED ISSUE #952: ErrorCalculator does not work in corpus without <space>
hello，when I run aishell recipe，error is as follows：

  File "espnet/bin/asr_train.py", line 373, in <module>
    main(sys.argv[1:])
  File "espnet/bin/asr_train.py", line 361, in main
    train(args)
  File "espnet-0.4.0-py3.6.egg/espnet/asr/pytorch_backend/asr.py", line 257, in train
  File "espnet-0.4.0-py3.6.egg/espnet/nets/pytorch_backend/e2e_asr_transformer.py", line 116, in __init__
  File "espnet-0.4.0-py3.6.egg/espnet/nets/e2e_asr_common.py", line 108, in __init__
ValueError: "\<space>\" is not in list

bug?

Fhrozen:
fix #952, No space in char_list

### DIFF
--- a/espnet/nets/e2e_asr_common.py
+++ b/espnet/nets/e2e_asr_common.py
@@ -105,7 +105,10 @@ class ErrorCalculator(object):
         self.report_cer = report_cer
         self.report_wer = report_wer
         self.idx_blank = self.char_list.index(self.blank)
-        self.idx_space = self.char_list.index(self.space)
+        if self.space in self.char_list:
+            self.idx_space = self.char_list.index(self.space)
+        else:
+            self.idx_space = None
 
     def __call__(self, ys_hat, ys_pad, is_ctc=False):
         cer, wer = None, None
--- a/espnet/nets/pytorch_backend/e2e_asr_transformer.py
+++ b/espnet/nets/pytorch_backend/e2e_asr_transformer.py
@@ -109,7 +109,7 @@ class E2E(ASRInterface, torch.nn.Module):
         else:
             self.ctc = None
 
-        if args.report_cer or args.report_wer or args.mtlalpha > 0.0:
+        if args.report_cer or args.report_wer:
             from espnet.nets.e2e_asr_common import ErrorCalculator
             self.error_calculator = ErrorCalculator(args.char_list,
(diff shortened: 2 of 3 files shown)


kan-bayashi:
https://github.com/espnet/espnet/blob/5520987581c2026be367fd3491904fd589dec671/espnet/nets/pytorch_backend/e2e_asr_transformer.py#L112
I think `mtlalpha > 0.0` is not necessary.
Could you remove it?

Fhrozen:
Should I remove it? I put it in case the model is trained full attention and when both report variable were disabled.

kan-bayashi:
I see. 
Then it should be `(args.report_cer or args.report_wer) and args.mtlalpha > 0.0`, right?

Fhrozen:
sadly nop. `args.report_cer` => reports for CER values in Attention, `args.report_wer` reports for WER in Attention and  `args.mtlalpha > 0.0` report for CER in CTC. if I put an `and` then for cases where CER attention is required but not CTC is used then it will not work. That is why two `or` are employed

kan-bayashi:
What I want to do is disable this when `report_cer=False` and `report_wer=False`.
Currently, when `mtlalpha>0.0` it will be always True.

Fhrozen:
Ok I see, so report CER means report CER on Attention and CTC. I will modify and commit the fixes

r9y9:
It would be great if you add a test. In my option, we should try to add a test when we fix a bug.

Fhrozen:
@r9y9 there are already tests for ER 
https://github.com/espnet/espnet/blob/5b01766e5f7299b32854e266b02036e5418cfb75/test/test_utils.py#L146-L184

should be added more?

r9y9:
Do we have a test for https://github.com/espnet/espnet/issues/952? If I understand correctly, a test is missing for the particular issue, at line https://codecov.io/gh/espnet/espnet/pull/955/diff?src=pr&el=tree#D1-111, right?

kan-bayashi:
Thanks a lot!

PR #956, opened 2019-07-10
Fix a typo bug in computing guided attention loss

r9y9:
Fixed a bug causing index out of bounds error when elayers != dlayers.
Added a test case.


kan-bayashi:
Thanks!
Maybe merge current master is needed to pass travis check.

PR #957, opened 2019-07-10
transform: expose pad_mode for logmelspectrogram

r9y9:
I needed control to change `pad_mode` (I was using `pad_mode="constant"`) when computing log mel-spectrogram.


kan-bayashi:
Thanks!

Issue #961, opened 2019-07-15
Training performance issue


geekboood:
Hi! When I try to train on Aishell dataset, it trains very slow. The weird thing is that one core of my CPU is always running at 100% but my GPU is only used about 30% ~ 40%. I wonder is there any dataset loading overhead (I have met this before but I cannot figure it out in this scenerio) or other things that slow my tranining process down or I can parallellize these CPU heavy usage operations.

geekboood:
I found that if I remove the code which write info to tensorboard, this CPU bottleneck will be greatly migitated. It seems that the training program write info every iteration so could we set a parameter to report every certain number of iterations?

sw005320:
Thanks for your TIPS. I did not notice it. 
I agree that we don't need to set it every iteration.
In https://github.com/espnet/espnet/blob/master/espnet/asr/pytorch_backend/asr.py#L466,  can you change it something like below?
```
trainer.extend(TensorboardLogger(writer, att_reporter))
```
->
```
trainer.extend(TensorboardLogger(writer, att_reporter), trigger=(REPORT_INTERVAL, 'iteration'))
```
Please let me know if it solves your issue. We may do this as default.

geekboood:
@sw005320 I change the file and it does work. If the program writes info every iteration, it results in `2.9445 iters/sec`. After changing the code (I think this time it writes info every 100 iterations), it shows `4.4718 iters/sec`.

sw005320:
Very cool! Could you make a PR for this? or we'll do it by ourselves.

geekboood:
@sw005320 Please do it for me. By the way, since Transformer can train in FP16, and on RTX cards it can significantly increase the training speed and allowing a bigger batchsize, could we introduce such feature to accelerate the training process?

sw005320:
This is interesting. @ShigekiKarita, what do you think?

ShigekiKarita:
Wow, I will try that (but my RTX2080ti cards are not available in this week...). Previously I trained Transformer in FP16 with fairseq. It was really fast as seen in this paper https://arxiv.org/pdf/1806.00187.pdf

PR #966, opened 2019-07-16
make tensorboard logging done every 100 iters

sw005320:
This is related to #961


sw005320:
@hirofumi0810, can you do this for the MT part? This may improve the training speed.

PR #970, opened 2019-07-18
Update Librispeech recipe

kan-bayashi:


kan-bayashi:
@sw005320 Should we change the default config? I'm wandering because it requires 4 gpus.

@ShigekiKarita Could you check lm config is correct?

ShigekiKarita:
This is my/our best LM config https://github.com/ShigekiKarita/espnet/blob/asru2019/egs/librispeech/asr1/conf/lm_layer4_unit2048_dropout0.0_batchsize512.yaml

In fact, I added `--dropout` to `lm_train.py` https://github.com/ShigekiKarita/espnet/commit/ab34eb971ff137fa44591e116deaf95ce4d9df43#diff-8caa23dbf28dfc5bfc24bd8bf781e220

ShigekiKarita:
Technically, the `epoch` was 11 in our best system, and we need to add LM averaging to `run.sh`. It looks like
```bash
average_checkpoints.py --snapshots ${lmexpdir}/snapshot.ep* --num 6 --out $lmexpdir}/model.average
```

sw005320:
Can you make a unit test for `utils/average_checkpoints.py`?
We could also use the best valid score average as default later.

kan-bayashi:
Done.

PR #973, opened 2019-07-19
New RNN and Transformer result of AMI recipe(ihm)

Jzmo:


sw005320:
Thanks. By considering https://github.com/kaldi-asr/kaldi/blob/master/egs/ami/s5b/RESULTS_ihm, this result seems to be OK (we have absolute ~6% behind though). Did you encounter the same weird loss behavior as #819?

Jzmo:
Yes same big and small negative loss in these experiments except using conf/train.yaml (only small negative loss here).

PR #988, opened 2019-07-26
 Fix lm_n_average in lang_model

Magic-Bubble:
A little bug when running decode in librispeech run.sh, lang_model need to use lm_n_average, not n_average.


sw005320:
Thanks a lot!

PR #997, opened 2019-07-30
[WIP] New properties for data loader

kan-bayashi:


kan-bayashi:
It is better to revisit after some important changes.

PR #999, opened 2019-07-30
RFC: Add keep_all_data_on_mem option

r9y9:
An native implementation (this is what I had locally)

Not deeply investigated yet, but it had seemed to be not working as expected (memory usage wasn't increased when I treid this a while ago). Probably I am not fully understanding how data is processed in espnet. Any ideas?

### DIFF
--- a/espnet/tts/pytorch_backend/tts.py
+++ b/espnet/tts/pytorch_backend/tts.py
@@ -316,6 +316,7 @@ def train(args):
         use_second_target=args.use_second_target,
         preprocess_conf=args.preprocess_conf,
         preprocess_args={'train': True}  # Switch the mode of preprocessing
+        keep_all_data_on_mem=args.keep_all_data_on_mem,
     )
 
     load_cv = LoadInputsAndTargets(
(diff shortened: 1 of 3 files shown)


espnet/tts/pytorch_backend/tts.py
```diff
@@ -324,6 +325,7 @@ def train(args):
         use_second_target=args.use_second_target,
         preprocess_conf=args.preprocess_conf,
         preprocess_args={'train': False}  # Switch the mode of preprocessing
```
kan-bayashi:
```suggestion
        preprocess_args={'train': False},  # Switch the mode of preprocessing
```

espnet/tts/pytorch_backend/tts.py
```diff
@@ -316,6 +316,7 @@ def train(args):
         use_second_target=args.use_second_target,
         preprocess_conf=args.preprocess_conf,
         preprocess_args={'train': True}  # Switch the mode of preprocessing
```
kan-bayashi:
```suggestion
        preprocess_args={'train': True},  # Switch the mode of preprocessing
```

r9y9: Ah, this is due to cherry-picking conflict. Will push a fix shortly.

kan-bayashi:
LGTM
@kamo-naoyuki Do you have any comments?

kamo-naoyuki:
LGTM. 
(Perhaps I also modify this feature with #1012, but it's O.K in the current status.)

r9y9:
Yeah maybe we can close this PR. I would be grateful if you could consider the similar feature in #1012. cc: @kamo-naoyuki

PR #1002, opened 2019-07-30
fix unk scoring

### LINKED ISSUE #993: The scoring script is incorrect if any pruning is done on the dictionary or in the case of word models
Hi @sw005320 and other developers,

The command `json2trn.py ${dir}/data.json ${dic} --num-spkrs ${num_spkrs} --refs ${dir}/ref.trn --hyps ${dir}/hyp.trn` in the `score_sclite.sh` replaces all units in reference that are not in the dictionary with unk and then scores it.

This is wrong because you are modifying the reference transcription to match your dictionary. Let's say I train a model with dictionary with only 1 term "Mr" and the hypothesis is 
"Mr unk unk" and reference is "Mr John Doe". This would give a WER of 0% because the reference would be modified to "Mr unk unk".

sw005320:
This PR is to fix #993.
The reference must not be converted to unk.
@siddalmia, what do you think of this solution (this is not perfect though).

### DIFF
--- a/test_utils/test_json2trn.py.bats
+++ b/test_utils/test_json2trn.py.bats
@@ -1,0 +1,62 @@
+#!/usr/bin/env bats
+
+setup() {
+    [ ! -z $LC_ALL ] && export LC_ALL="en_US.UTF-8"
+
+    utils=$(cd $BATS_TEST_DIRNAME/..; pwd)/utils
+    tmpdir=$(mktemp -d testXXXXXX)
+
+    cat << EOF >> $tmpdir/valid.hyp
+A B C (spk1-uttid1)
+A <unk> <unk> (spk2-uttid2)
+EOF
+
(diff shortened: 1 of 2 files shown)


siddalmia:
Looks fine @sw005320!  

I don't remember how exactly you store your reference words. But if you append `<eos>` initially to the training, you would want to keep `ref_file.write(" ".join(seq).replace('<eos>', ''))` right? 

Thanks

sw005320:
> I don't remember how exactly you store your reference words. But if you append <eos> initially to the training, you would want to keep ref_file.write(" ".join(seq).replace('<eos>', '')) right?

Thanks. I just double-checked that the reference token is generated from https://github.com/espnet/espnet/blob/master/utils/text2token.py and it does not attach `<eos>`. So we should not need it. I added some comments about it in the script.

PR #1003, opened 2019-07-30
nbsphinx support

ShigekiKarita:
see this page for preview https://karita.xyz/espnet/notebook/asr_cli.html

### DIFF
--- a/ci/doc.sh
+++ b/ci/doc.sh
@@ -17,6 +17,8 @@ set -euo pipefail
     cd ./doc
     if [ ! -e notebook ]; then
         git clone https://github.com/espnet/notebook
+        cd notebook
+        git checkout de000a4fd000dcffa5c0ee14fc477f96b7af6fea
     fi
 )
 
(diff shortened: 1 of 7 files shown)


ci/doc.sh
```diff
@@ -13,6 +13,13 @@ set -euo pipefail
     ../doc/argparse2rst.py ./*.py > ../doc/_gen/utils_py.rst
 )
 
+(
+    cd ./doc
+    if [ ! -e notebook ]; then
+        git clone https://github.com/espnet/notebook
```
r9y9: To support off-line doc build in the future, we may consider adding this dependency as a git submodule, which is more explicit and sounds reasonable approach to me.

ShigekiKarita: Of course. I would like to use `git checkout xxxx` instead of git submodule (as well as `tools/Makefile`) because it is too difficult for us... (sorry)

r9y9: FYI I used to use git submodule to include notebooks to sphinx doc (see https://github.com/r9y9/nnmnkwii/tree/master/docs) and I get no problem so far. Off course I'm okay with git checkout though :)

ShigekiKarita: Thanks. We decided to use git submodule!

PR #1005, opened 2019-07-31
Add multi-speaker Transformer and FastSpeech config in LibriTTS

kan-bayashi:
This PR adds new config file for multi speaker transformer in libriTTS.
Should be merged after #1001 #1006


kan-bayashi:
To avoid OOM error, we need to tune the batch-bins.

Issue #969, opened 2019-07-17
Cupy Import Error.


Xiaofei-Wang:
Does someone have the cupy import error during making the ESPnet?
For example, if I 
`make KALDI=/path/to/kaldi CUDA_VERSION=9.0`
The cupy 6.0 still tries to find libraries of CUDA 10.0 on default, although I have correctly set the environment.

Instead of using 
`cupy.done: espnet.done
ifneq ($(strip $(CUPY_VERSION)),)
	. venv/bin/activate && pip install cupy==$(CUPY_VERSION)
	touch cupy.done
endif`

I have to manually change the code to
`cupy.done: espnet.done
ifneq ($(strip $(CUPY_VERSION)),)
	. venv/bin/activate && pip install cupy-cuda90==$(CUPY_VERSION)
	touch cupy.done
endif`

I would suggest making a small modification of the Makefile: )

sw005320:
Thanks for the report.
If so, it would be better to switch to use `cupy-cuda80`, `cupy-cuda90`, `cupy-cuda100`, etc.(?)
Hmm, this is not so cool option.

PR #1010, opened 2019-07-31
fix Cupy Import Error #969

Xiaofei-Wang:
Solve the issue described in Issues: Cupy Import Error. #969


ShigekiKarita:
Travis failure seems to be strange because it might have run #1003 (notebook-based doc). 

Anyway, this PR cannot be tested in Travis because it does not install cupy. I am running this branch in my local env. If it works well, I will merge this.

ShigekiKarita:
I will merge this.

PR #1016, opened 2019-07-31
[Docstrings] Removing empty init files to avoid docs

Fhrozen:
In order to start using flake8-docstring , it was found that empty `__init__.py` files can be removed from flake8 test.
This reduces from 1004 errors to 985. 
Referred flag:  D104 Missing docstring in public package


ShigekiKarita:
I recommend to rm these files because python3 does not require them.

sw005320:
@Fhrozen and @ShigekiKarita, do you think we should recover `__init__.py`? 
There is a report in #1831

Fhrozen:
I think that we can add, but we will need to add comments on all files. Otherwise, the coverage will be reduced to ~40% (If I can remember the number). Or add and exclusion flag for flake8.
ESPnet is usually installed with `-e` flag, so the error was not observed before (Until now?).

ShigekiKarita:
I agree to revert them.

sw005320:
Year, let's revert it.
@Fhrozen, could you do it?

Fhrozen:
OK,

ShigekiKarita:
I think coverage isn't problem because pytest-cov ignores modules without __init__.py never imported in test but measures ones with it never imported. So it is more strict.

PR #1026, opened 2019-08-01
Support of other recipe in recog_wav.sh

hiratake55:
This modification supports the other ASR recipe not only tedlium2.

### DIFF
--- a/utils/recog_wav.sh
+++ b/utils/recog_wav.sh
@@ -88,12 +88,6 @@ if [ -z $models ]; then
     fi
 fi
 
-# Check jq is installed
-if ! [ -x "$(command -v jq)" ]; then
-    echo 'Error: jq is not installed.' >&2
-    exit 1
-fi
-
 dir=${download_dir}/${models}
 mkdir -p ${dir}
 
@@ -216,7 +210,7 @@ if [ ${stage} -le 3 ] && [ ${stop_stage} -ge 3 ]; then
         ${recog_opts}
 
     echo ""
-    recog_text=$(cat ${decode_dir}/result.json | jq ".utts.\"${base}\".output[].rec_text" | sed -e 's/"\(.*\)<eos>.*/\1/')
+    recog_text=$(grep rec_text ${decode_dir}/result.json | sed -e 's/.*: "\(.*\)".*/\1/' | sed -e 's/<eos>//')	
     echo "Recognized text: ${recog_text}"
     echo ""
     echo "Finished"
--- a/utils/recog_wav.sh
+++ b/utils/recog_wav.sh
@@ -1,10 +1,11 @@
 #!/bin/bash
 
 # Copyright 2019 Nagoya University (Takenori Yoshimura)
+#           2019 RevComm Inc. (Takekatsu Hiramura)
 #  Apache 2.0  (http://www.apache.org/licenses/LICENSE-2.0)
 
 if [ ! -f path.sh ] || [ ! -f cmd.sh ]; then
-    echo "Please change directory to e.g., egs/tedlium2/asr1"
+    echo "Please change current directory to recipe directory e.g., egs/tedlium2/asr1"
     exit 1
 fi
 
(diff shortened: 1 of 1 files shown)


utils/recog_wav.sh
```diff
@@ -54,26 +71,45 @@ decode_cmd=
...
+if [ -z $models ]; then
+    if [[ -z $cmvn || -z $lang_model || -z $recog_model || -z $decode_config ]]; then
+    echo 'Error: models or set of cmvn, lang_model, recog_model and decode_config are not set.' >&2
+    exit 1
+    fi
+fi
+
+# Check jq is installed
+if ! [ -x "$(command -v jq)" ]; then
+    echo 'Error: jq is not installed.' >&2
```
sw005320: `jq` is not a standard tool, and it would be better to include the installation of it as an extra tool. (I used to include it in our Makefile, but now I removed it).

takekatsuhiramura: I see. I will include `jq` installation in Makefile.

sw005320:
> Removed usage of jq

Year, I think this is better.

takekatsuhiramura:
I considered adding `jq` installation to MakeFile, but it has possibility of making conflict to system installed `jq`. Therefore, it reverted to sed.

utils/recog_wav.sh
```diff
@@ -54,26 +71,39 @@ decode_cmd=
...
 fi
 
 set -e
 set -u
 set -o pipefail
 
+# Check model name or model file is set
+if [ -z $models ]; then
+    if [[ -z $cmvn || -z $lang_model || -z $recog_model || -z $decode_config ]]; then
+    echo 'Error: models or set of cmvn, lang_model, recog_model and decode_config are not set.' >&2
```
takenori-y:
```suggestion
        echo 'Error: models or set of cmvn, lang_model, recog_model and decode_config is required.' >&2
```

utils/recog_wav.sh
```diff
@@ -54,26 +71,39 @@ decode_cmd=
...
     exit 1;
 fi
 
 set -e
 set -u
 set -o pipefail
 
+# Check model name or model file is set
+if [ -z $models ]; then
+    if [[ -z $cmvn || -z $lang_model || -z $recog_model || -z $decode_config ]]; then
```
takenori-y: This line does not depend on `$use_lang_model`.

utils/recog_wav.sh
```diff
@@ -180,7 +210,7 @@ if [ ${stage} -le 3 ] && [ ${stop_stage} -ge 3 ]; then
         ${recog_opts}
 
     echo ""
-    recog_text=$(grep rec_text ${decode_dir}/result.json | sed -e 's/.*: "\(.*\)<eos>.*/\1/')
+    recog_text=$(grep rec_text ${decode_dir}/result.json | sed -e 's/.*: "\(.*\)".*/\1/' | sed -e 's/<eos>//')
```
takenori-y:
```suggestion
    recog_text=$(grep rec_text ${decode_dir}/result.json | sed -e 's/.*: "\(.*\)".*/\1/' | sed -e 's/<eos>//')
```

utils/recog_wav.sh
```diff
@@ -54,26 +71,39 @@ decode_cmd=
...
+    exit 1
+    fi
+fi
+
+dir=${download_dir}/${models}
+mkdir -p ${dir}
+
 function download_models () {
+    if [ -z $models ]; then
+        return 0
```
takenori-y: Return value may be omitted.
```suggestion
        return
```

utils/recog_wav.sh
```diff
@@ -54,26 +71,39 @@ decode_cmd=
...
 
 set -e
 set -u
 set -o pipefail
 
+# Check model name or model file is set
+if [ -z $models ]; then
+    if [[ -z $cmvn || -z $lang_model || -z $recog_model || -z $decode_config ]]; then
+    echo 'Error: models or set of cmvn, lang_model, recog_model and decode_config are not set.' >&2
+    exit 1
```
takenori-y:
```suggestion
        exit 1
```

takekatsuhiramura:
@takenori-y I appreciate for your review, I totally agree with your comment. I pushed revised edition of code.

sw005320:
Thanks, @hiratake55!

PR #1027, opened 2019-08-01
Add log of the inference speed of TTS models

kan-bayashi:
This PR adds log about the inference speed of TTS models.


kan-bayashi:
I see. I keep it in mind.

PR #1029, opened 2019-08-01
Use pypi sentencepiece

ShigekiKarita:
I also add compatibility check with previous C++ bin outputs. See test/test_sentencepiece.py


sw005320:
LGTM.

Does `spm_train` accept all arguments? (e.g.. https://github.com/espnet/espnet/blob/master/egs/swbd/asr1/run.sh#L150-L160)

ShigekiKarita:
I updated the test to use it. In fact `spm_train` seems to do the same thing in python
https://github.com/google/sentencepiece/blob/05fdffb51d4254baac32adb899978dc117b1e298/src/spm_train_main.cc#L177

PR #1028, opened 2019-08-01
Format Chainer E2E transformer forward

Fhrozen:
In order to make more similar the scripts between pytorch and chainer, this PR simplifies the script in forward function (E2E_transformer). 
Some functions have been moved into the encoder and decoder depending its required
This PR does not intent to check docs into the transformer so I would recommend to check and merge this one first and then #1017

PR #1017, opened 2019-07-31
[Docstrings] Doc for Chainer transformer

Fhrozen:
Added some docs for chainer transformer to reduce some error.

PR #1034, opened 2019-08-01
Format Chainer E2E transformer forward (fixed)

Fhrozen:
In order to make more similar the scripts between pytorch and chainer, this PR simplifies the script in forward function (E2E_transformer).
Some functions have been moved into the encoder and decoder depending its required
This PR does not intent to check docs into the transformer so I would recommend to check and merge this one first and then #1017

Replace of #1028


espnet/nets/chainer_backend/ctc.py
```diff
@@ -127,6 +129,35 @@ def __call__(self, hs, ys):
 
         return self.loss
 
+    def forward_from_transformer(self, hs, ys):
+        """Core function of the Warp-CTC layer.
+
+        Args:
+            hs (iterable of chainer.Variable | N-dimention array): Input variable from encoder.
+            ys (iterable of chainer.Variable | N-dimension array): Input variable of decoder.
```
ShigekiKarita: I do not think the `ys` can be chainer.Variable. see https://github.com/jheymann85/chainer_ctc/blob/master/chainer_ctc/warpctc.py#L34

espnet/nets/chainer_backend/ctc.py
```diff
@@ -127,6 +129,35 @@ def __call__(self, hs, ys):
 
         return self.loss
 
+    def forward_from_transformer(self, hs, ys):
```
ShigekiKarita: If this method is only for Transformer, I would like to find this in Transformer class

espnet/nets/chainer_backend/e2e_asr_transformer.py
```diff
@@ -169,80 +155,38 @@ def forward(self, xs, ilens, ys_pad, calculate_attentions=False):
...
-                cer_ctc = self.error_calculator(ys_hat, ys_pad_cpu, is_ctc=True)
+                cer_ctc = self.error_calculator(ys_hat, ys_pad, is_ctc=True)
 
-        # Compute cer/wer
-        with chainer.no_backprop_mode():
-            y_hats = ys.reshape((batch, length, -1))
-            y_hats = chainer.backends.cuda.to_cpu(F.argmax(y_hats, axis=2).data)
+        # 3. Decoder
+        ys = self.decoder(ys_pad, xs, x_mask)
+        if calculate_attentions:
```
ShigekiKarita: instead of return nothing, I suggest to create new method like `def calculate_attentions(self, xs, xs_mask, ys_pad)`

espnet/nets/chainer_backend/transformer/label_smoothing_loss.py
```diff
@@ -19,9 +19,17 @@ def __init__(self, smoothing, n_target_vocab, normalize_length=False):
...
-
-    def forward(self, concat_logit_block, t_block, batch, length):
+        self.ignore_id = ignore_id
+
+    def forward(self, ys_block, ys_pad, eos):
+        xp = self.xp
+        eos = np.array([eos], 'i')
+        ys_out = [np.concatenate([y, eos], axis=0) for y in ys_pad]
+        ys_out = F.pad_sequence(ys_out, padding=-1)
+        t_block = chainer.Variable(xp.array(ys_out.data))
```
ShigekiKarita: use `with chainer.no_backprop_mode():`

espnet/nets/chainer_backend/ctc.py
```diff
@@ -90,6 +90,8 @@ class WarpCTC(chainer.Chain):
 
     def __init__(self, odim, eprojs, dropout_rate):
```
ShigekiKarita: maybe you need to change `__call__` to `forward` in this class

espnet/nets/chainer_backend/transformer/decoder.py
```diff
@@ -48,27 +50,49 @@ def __init__(self, odim, args, initialW=None, initial_bias=None):
...
-            chainer.Chain: Decoder layer.
+    def forward(self, ys_pad, source, x_mask):
+        """Forward decoder
 
+        :param xp.array e: input token ids, int64 (batch, maxlen_out)
+        :param xp.array yy_mask: input token mask, uint8  (batch, maxlen_out)
+        :param xp.array source: encoded memory, float32  (batch, maxlen_in, feat)
+        :param xp.array xy_mask: encoded memory mask, uint8  (batch, maxlen_in)
+        :return e: decoded token score before softmax (batch, maxlen_out, token)
+        :rtype: chainer.Variable
```
ShigekiKarita: use Google style doc

Fhrozen: I am leaving this for #1017

espnet/nets/chainer_backend/transformer/label_smoothing_loss.py
```diff
@@ -44,4 +52,5 @@ def forward(self, concat_logit_block, t_block, batch, length):
             label_smoothing = F.sum(label_smoothing) / normalizer
             loss = self.confidence * loss + self.smoothing * label_smoothing
 
-        return loss
+        acc = F.accuracy(concat_logit_block, concat_t_block, ignore_label=self.ignore_id)
+        return loss, acc
```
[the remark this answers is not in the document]
Fhrozen: Ok, it is fine is it is stored inside of a self.acc in the label and then called from the e2e.forward?

espnet/nets/chainer_backend/transformer/decoder.py
ShigekiKarita: OK

espnet/nets/chainer_backend/transformer/label_smoothing_loss.py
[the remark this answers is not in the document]
ShigekiKarita: No. I do not want `LabelSmoothingLoss` to do too many things except for the loss calculation. I suggest:
- concat `eos` to `t_block` outside
- calculate acc outside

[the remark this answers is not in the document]
Fhrozen: done, pls check it

sw005320:
I changed the target branch of this PR from the master to v.0.6.0 since it includes several refactoring.

.gitignore
```diff
@@ -4,6 +4,7 @@
 \#*\#
 .\#*
 *DS_Store
+out.txt
```
sw005320: Do you need this?

.gitignore
```diff
@@ -21,6 +22,7 @@ shellcheck*
 check_shellcheck*
 test_spm.vocab
 test_spm.model
+custom*
```
sw005320: Do you need this?

.gitignore
```diff
@@ -36,6 +38,7 @@ egs/*/*/tensorboard
 egs/*/*/wav*
 
 # tools related
+tools/ideep*
```
sw005320: Do you need this?

espnet/nets/chainer_backend/transformer/subsampling.py
```diff
@@ -15,32 +15,29 @@ class Conv2dSubsampling(chainer.Chain):
...
-                                initialW=initialW(scale=stvd),
-                                initial_bias=initial_bias(scale=stvd))
-        self.add_link('conv.2', layer)
-        stvd = 1. / np.sqrt(dims)
-        layer = L.Linear(idim, dims, initialW=initialW(scale=stvd),
-                         initial_bias=initial_bias(scale=stvd))
-        self.add_link('out.0', layer)
         self.dropout = dropout
         with self.init_scope():
+            n = 1 * 3 * 3
```
sw005320: Can you add some comments about `n = 1 * 3 * 3`?

.gitignore
Fhrozen: Pytest generates this file, so maybe

sw005320: I see. Then, let's keep this.

.gitignore
Fhrozen: I was thinking about any custom training/decoding setup used for egs so they can be ignore, something like custom_training_my_rnn.yaml

.gitignore
Fhrozen: I was trying to speed up chainer for cpu so I installed it directly, i can remove it

espnet/nets/chainer_backend/transformer/subsampling.py
Fhrozen: ok

.gitignore
sw005320: ok, but it may not be started from `custom*`. So, we may not need this.

Fhrozen:
The main difference between the ctc for transformer and the rnn is because the target (`ys`) is already a list of arrays located in the cpu, while in rnn routine the target is a list of variables located in cpu. If the target of rnn becomes a list of cpu arrays then it should no be a problem.

sw005320:
> The main difference between the ctc for transformer and the rnn is because the target (`ys`) is already a list of arrays located in the cpu, while in rnn routine the target is a list of variables located in cpu. If the target of rnn becomes a list of cpu arrays then it should no be a problem.

OK.
Can you add this comment to the source? 
You may add it as a TODO to unify this module across the PyTorch and Chainer backends, and we may care this in later PRs if needed.

sw005320:
Thanks, @Fhrozen!

Issue #1038, opened 2019-08-03
espnet/nets/chainer_backend/e2e_asr.py has bug


boystray:
when I run egs/an4/asr1$ ./run.sh --backend chainer, there is a error

stage 4: Network Training
run.pl: job failed, log is in exp/train_nodev_chainer_train_mtlalpha1.0/train.log

the following is exp/train_nodev_chainer_train_mtlalpha1.0/train.log
 `^[[JException in main training loop: unsupported operand type(s) for *: 'float' and 'NoneType'
Traceback (most recent call last):
  File "/media/boystray/two/githome/espnet/tools/venv/lib/python3.7/site-packages/chainer/training/trainer.py", line 316, in run
    update()
  File "/media/boystray/two/githome/espnet/espnet/nets/chainer_backend/rnn/training.py", line 111, in update
    self.update_core()
  File "/media/boystray/two/githome/espnet/espnet/nets/chainer_backend/rnn/training.py", line 91, in update_core
    loss = optimizer.target(*x) / self.accum_grad
  File "/media/boystray/two/githome/espnet/tools/venv/lib/python3.7/site-packages/chainer/link.py", line 294, in __call__
    out = forward(*args, **kwargs)
  File "/media/boystray/two/githome/espnet/espnet/nets/chainer_backend/e2e_asr.py", line 119, in forward
    self.loss = alpha * loss_ctc + (1 - alpha) * loss_att
Will finalize trainer extensions and updater before reraising the exception.
Traceback (most recent call last):
  File "/media/boystray/two/githome/espnet/egs/an4/asr1/../../../espnet/bin/asr_train.py", line 380, in <module>
    main(sys.argv[1:])
  File "/media/boystray/two/githome/espnet/egs/an4/asr1/../../../espnet/bin/asr_train.py", line 365, in main
    train(args)
  File "/media/boystray/two/githome/espnet/espnet/asr/chainer_backend/asr.py", line 358, in train
    trainer.run()
  File "/media/boystray/two/githome/espnet/tools/venv/lib/python3.7/site-packages/chainer/training/trainer.py", line 349, in run
    six.reraise(*exc_info)
  File "/media/boystray/two/githome/espnet/tools/venv/lib/python3.7/site-packages/six.py", line 693, in reraise
    raise value
  File "/media/boystray/two/githome/espnet/tools/venv/lib/python3.7/site-packages/chainer/training/trainer.py", line 316, in run
    update()
[...]

PR #1039, opened 2019-08-03
Fix Loss Calculation

Fhrozen:
In response to #1038, the mtl loss calculation is being rolled back.


kan-bayashi:
Could you add test to check `mtlalpha=0.0` and `mtlaplha=1.0` in an another PR?
(I will merge this PR soon.)

PR #1040, opened 2019-08-03
Add test for MTLalpha

Fhrozen:
In response of #1038, #1039


kan-bayashi:
I modified test interface to check various options. See #1041.

PR #1042, opened 2019-08-03
Refactor E2E-TTS test

kan-bayashi:
- make hyperparameters smaller to fasten test
- add missing tests


ShigekiKarita:
Can you add `pytest --durations=0 ` (or setup.cfg) to check the running time? See also https://stackoverflow.com/a/55095253

ShigekiKarita:
Can you split circle ci's `config.yml` 
https://github.com/espnet/espnet/blob/3e24336b197843eb03a9f16583b7b9994efc23cc/.circleci/config.yml#L31-L41
into 
``` yaml
- run: sudo apt-get install -qq -y cmake g++-7 libsndfile1-dev bc 
- run: ./ci/install.sh 
- run: ./ci/test_shell.sh
...
```
because we cannot measure which part is slow [in the report](https://circleci.com/gh/espnet/espnet/4481) like Travis. Maybe it was my fault...

kan-bayashi:
Yes, I checked in local but it seems no problem.
It is strange since in travis the speed is ok.

ShigekiKarita:
Maybe you misunderstood my requests (??)

We need these two refinements `pytest --durations 0` and split config.yml not only for this PR but also future PRs to check CI slow down or bottlenecks. We need to know how the unit test matters for refactoring.

kan-bayashi:
I know. I will do it in another PR.

PR #1047, opened 2019-08-06
Add option to select the integration type of speaker embedding

kan-bayashi:
This PR adds option to select the integration type of speaker embedding in TTS-Transformer.


sw005320:
How about making this as an option?

kan-bayashi:
Yes, once I checked the results, I will do it.

kan-bayashi:
I found multi-speaker Transformer needs more strict guided attention loss to get diagonal encoder-decoder attention.
I attached attention in inference.

### speaker embed integration = concat / guided attention lambda = 1.0
This model can generate speech but not so good.

スクリーンショット 2019-08-09 午前2 05 37

### speaker embed integration = add / guided attention lambda = 1.0
This model cannot generate speech.

スクリーンショット 2019-08-09 午前2 05 56

### speaker embed integration = add / guided attention lambda = 100.0
This model can generate reasonable speech.

スクリーンショット 2019-08-09 午前2 06 19

We should check the results with concat + large lambda.

kan-bayashi:
I found that some interesting phenomena.
The attention weights are different between artificial (derived from guided attention loss) and natural ones.
The left bottom 2x2 figures represent artificial attention weights.
スクリーンショット 2019-08-11 午後11 21 20

kan-bayashi:
I finished experiment. 
I uploaded model and samples.
Ready to merge.

kan-bayashi:
From the experiment, I found some interesting points:
- In the case of `reduction-factor=2` + `spk-embed-integration-type="add"`, we do not need to use `guided_attention_loss`
- But we cannot learn a diagonal attention without `guided_attention_loss` in the case of `reduction-factor=2` + `spk-embed-integration-type="concat"`
- Large `warmup-steps` seems  to help to learn a diagonal attention
- `reduction_factor=1` causes a serious mispronunciation issue
- Attention weights are different between artificial (derived from guided attention loss) and natural ones.

sw005320:
> * n the case of `reduction-factor=2` + `spk-embed-integration-type="add"`, we do not need to use `guided_attention_loss`
> * But we cannot learn a diagonal attention without `guided_attention_loss` in the case of `reduction-factor=2` + `spk-embed-integration-type="concat"`

Very interesting. Thanks for the report.

r9y9:
Thanks for sharing the interesting insights! 

> Attention weights are different between artificial (derived from guided attention loss) and natural ones.

If I understand correctly, does that mean guided attention loss might affect if the weight coefficient for the loss term is high (or sigma is too small)? If so it matches my intuition that guided attention loss *helps* the attention model converge, but might have side-effects; penalize correct paths that are not strictly diagonal (e.g. long silence in the middle of an utterance). We may consider using dynamic weight scheduling for guided attention loss so that the weight is approaching zero when the model starts to learn diagonal attention.

kan-bayashi:
@r9y9 I see. It is a nice idea.
Not directly related to your opinion, but I will share some points.
I compared the model with / without guided attention loss in LJSpeech.
It can be trained without guided attention loss, but it seems the model without guided attention loss causes more repetition even if the attention weight are very clear and diagonal in training.
(I did not calculate the correct number, just listened to several samples.)

kan-bayashi:
Also, I found that it is nice to check the attention weight in decoding.
I attached some samples:


## Good examples
スクリーンショット 2019-08-14 午前10 30 30

スクリーンショット 2019-08-14 午前10 30 42

## Bad examples with deletion
スクリーンショット 2019-08-14 午前10 30 56

スクリーンショット 2019-08-14 午前10 31 35

As you said, the deletion of words are happened, and we can check it from attention weights.

Issue #1012, opened 2019-07-31
New style to describe input features


kamo-naoyuki:
I'd like to discuss about inputting data format for training program.

We encountered some issues with the current style of `data.json` now and I also feel some limitations about [LoadInputsAndTargets](https://github.com/espnet/espnet/blob/master/espnet/utils/io_utils.py) to describe training set.

#533

-  About JSON
    - JSON used much memory if the text file is too large because JSON doesn't support lazy loading.
    - The parse speed of JSON is slower than binary format, e.g. HDF5. This is critical problem if training corpus is very large.
- About `LoadInputsAndTarget`
    - `LoadInputsAndTarget` has less general versatility to give the additional feature to the DNN-model. Currently it has only the flag such as `use_speaker_embedding`  and we need to modify the class manually if more different features are need.

# New style
```yaml
data:
  # The key names can be used arbitrary
  input1:  
    path: feat1.scp
    type: hdf5
  input2: 
    path: feats2.scp
    type: ark
  text: 
    path: utt2tokenid
    type: csv
  lang: 
    path: utt2lang
  spk: 
    path: utt2spk
batch:
  sort: input1  # Sort utts in descending order of input1 length
  input1:
    shape: utt2feats1_shape  # (feat-Length, feat-dim)
    maxlen: 800
  text: 
    shape:  utt2text_shape  # (char-Length, N vocab)
    maxlen: 150
preprocess:
   input1: preprocess_feats.yaml
   text: preprocess_text.yaml
```

```bash
asr_train.py --train-data train.yaml 
```

```python
from espnet.utils.foo import Loader
d = yaml.load('train.yaml')
loader = Loader(...)
batches = make_batchset(d, ...)
for batch in batches:
    data = loader(batch)
    # data is dictionally
    input1_feat = data['input1']
    input2_feat = data['input2']
    text = data['text']
```

sw005320:
@r9y9, do you have comments? We want to have opinions from non-ASR people.

sw005320:
In addition to this change, I think it's better to use `sentencepiece` as default for all recipes and switch to use it or not by an option (e.g., `nbpe=0` to use a char-based one).

butsugiri:
I really like the idea of shifting to format that supports lazy loading.
As one of the NLPer, I sometimes have to handle several gigabytes of text and thus the memory footprint is a huge problem.

[...]

PR #1054, opened 2019-08-07
Fix tensorboard interval in validation

ShigekiKarita:
After this PR #966 (make tensorboard log every 100 iters), we found that iter 100 is too large to display validation stats. This PR aims to revert the validation TB log interval to 1.

This PR realize this by calling TB logger manually inside the new evaluator (espnet.utils.training.evaluator.BaseEvaluator).

- [x] BaseEvaluator
- [x] use args.report_interval_iters intead of the constant REPORT_INTERVAL. It is consistent to tts_train.py
- [x] add test


ShigekiKarita:
@kan-bayashi I feel we can merge ASR and TTS CustomEvaluator/Updater. It seems to be same. Do you agree with it?

ShigekiKarita:
Now Tensorboard on AN4 seems to display validation set logs
image

kan-bayashi:
Could you check acc.png is correctly working?

kan-bayashi:
> @kan-bayashi I feel we can merge ASR and TTS CustomEvaluator/Updater. It seems to be same. Do you agree with it?

I will check it lator.

ShigekiKarita:
this is cer.png because acc.png is not generated by mtlalpha=1.0 in the AN4 default
cer

ShigekiKarita:
> I will check it lator.

I think the unification is a necessary step for the input/output data format in the future #1012

kan-bayashi:
> this is cer.png because acc.png is not generated by mtlalpha=1.0 in the AN4 default

Thanks. It looks fine.
Recently, I met strange phenomena in TTS as follows:


スクリーンショット 2019-08-07 午後9 51 16

Evaluation was conducted every 4 epochs, not every epoch.
This might be a bug of TTS.

kan-bayashi:
> I think the unification is a necessary step for the input/output data format in the future #1012

I checked the code and as you mentioned, they are the almost same.
Let us unify them.

ShigekiKarita:
Hmm. I'm running AN4 tts but the all_loss.png seems to be OK. Did you test it LJSpeech?
all_loss

>  Let us unify them.

Yeah. We should split the PR for the unification

kan-bayashi:
>  Did you test it LJSpeech?

I test it on LibriTTS.
The reason may be the use of `ngpu=3` and `accum_grad=8`, but I’m not sure.

egs/an4/asr1/local/check_tblog.py
```diff
@@ -0,0 +1,21 @@
+#!/usr/bin/env python3
+# coding: utf-8
+
```
kan-bayashi: Is this file needed?

ShigekiKarita: no. I made this just for pre-testing.

ShigekiKarita: no. I made this just for pre-testing.

ShigekiKarita:
Yes. I added the test.

sw005320:
Many thanks!

PR #985, opened 2019-07-25
REVERB multi-channel E2E recipe

Xiaofei-Wang:
- Action items
  - [x] update RESULTS once I get it 
  - [x] Add more objective metrics

PR #1057, opened 2019-08-07
Updated REVERB multi-channel E2E recipe

Xiaofei-Wang:
In the previous pull request (#985 ), I just copied the speech enhancement scores from egs/reverb/asr1/RESULT, which seems not accurate (it only shows the dt scores). 
I just made some modifications of the recipes to support the calculation of the scores for the evaluation set (et).
  - [x] Once I get the new speech enhancement scores for et, I'll update the RESULTS.md this evening.


egs/reverb/asr1_multich/local/compute_se_scores_multich.sh
```diff
@@ -0,0 +1,68 @@
+#/bin/bash
+# Copyright 2018 Johns Hopkins University (Author: Aswin Shanmugam Subramanian)
+# Apache 2.0
+
+# This script computes the dereverberation scores given in REVERB challenge
+# Eg. local/compute_se_scores.sh --nch 8 /export/corpora5/REVERB_2014/REVERB ${PWD}/wav ${PWD}/local
+
+. ./cmd.sh
```
sw005320:
```suggestion
. ./cmd.sh || exit 1;
```

egs/reverb/asr1_multich/local/compute_se_scores_multich.sh
```diff
@@ -0,0 +1,68 @@
...
+. ./path.sh
```
sw005320:
```suggestion
. ./path.sh || exit 1;
```

sw005320:
In general, we should have results for both dev and eval sets. Can you do it?

sw005320:
Thanks a lot @Xiaofei-Wang!

PR #1060, opened 2019-08-08
Allow to extend the length of positional encoding at inference

takenori-y:
The current implementation does not allow to lengthen the positional encoding at inference due to using `register_buffer`.

### DIFF
--- a/espnet/nets/pytorch_backend/transformer/embedding.py
+++ b/espnet/nets/pytorch_backend/transformer/embedding.py
@@ -24,10 +24,10 @@ class PositionalEncoding(torch.nn.Module):
         pe = pe.unsqueeze(0)
         self.max_len = max_len
         self.xscale = math.sqrt(d_model)
-        self.register_buffer('pe', pe)
+        self.pe = pe
 
     def forward(self, x):
-        x = x * self.xscale + self.pe[:, :x.size(1)]
+        x = x * self.xscale + self.pe[:, :x.size(1)].to(x.device)
         return self.dropout(x)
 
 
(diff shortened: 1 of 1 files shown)


ShigekiKarita:
Can you add a tiny test to check that it is runnable?

ShigekiKarita:
Actually, I do not understand why `register_buffer` is not and why this code is OK. I am also afraid that this will make training slow because every iteration needs transfer CPU memory fragments (self.pe) to CUDA.

takenori-y:
OK, I will check the training speed. If `register_buffer` is used, `pe` is saved in `state_dict`. Making `max_len` different from that is used in training causes an error during model loading.

ShigekiKarita:
Because `self.pe` is constant (not learned), maybe you should override the `state_dict` and `load_state_dict` (just load self.scale) in PE class.

sw005320:
Is there any progress of this PR?

takenori-y:
In my environment, any significant difference in computational speed could not be observed. `self.xscale` is also constant. I think there is no need to change `state_dict`.

sw005320:
@ShigekiKarita, is it OK?

ShigekiKarita:
Did you tried CUDA training? If it was CPU, the speed should not be changed because it is originaly on CPU and requires no copy. I wanna confirm that first.

In my opinion, maybe we can save/load nothing for this class. There are two good points
1. reduce snapshot size (maybe the reduction is not large ... 5000 x 256 x 2 elements?)
2. we can reset arbitary `max_len` for decoding because there is no need to use the same value for loading trained parameters

```python
PositionalEncoding(torch.nn.Module):
    # just adding these two methods
    def state_dict(self):
        return {}

    def load_state_dict(self, d):
        return
```

takenori-y:
Yes, I tried CUDA training. `ScaledPositionalEncoding` class requires a modification because it has a trainable parameter. Do you know how to modify it?

PR #1062, opened 2019-08-08
Fix spm decode

ShigekiKarita:
Python version of `utils/spm_decode` (I picked this from fairseq) had different interface to C++ one. Therefore `utils/score_sclite.sh` raise errors when we use it with spm. 
- [x] fix spm_decode for score_sclite.sh
- [x] update mini AN4 to use spm to test this.

### DIFF
--- a/utils/spm_decode
+++ b/utils/spm_decode
@@ -43,7 +43,6 @@ def main():
         h = open(args.input, "r", encoding="utf-8")
     for line in h:
         print(decode(line.split()))
-        # print(decode(list(map(tok2int, line.rstrip().split()))))
 
 
 if __name__ == "__main__":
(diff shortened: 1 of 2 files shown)


utils/spm_decode
```diff
@@ -36,9 +37,13 @@ def main():
...
-    with open(args.input, "r", encoding="utf-8") as h:
-        for line in h:
-            print(decode(list(map(tok2int, line.rstrip().split()))))
+    if args.input is None:
+        h = sys.stdin
+    else:
+        h = open(args.input, "r", encoding="utf-8")
+    for line in h:
+        print(decode(line.split()))
+        # print(decode(list(map(tok2int, line.rstrip().split()))))
```
kan-bayashi:
```suggestion
```

PR #1065, opened 2019-08-08
RNN-Transducer

b-flo:
Hi,

Sorry for the delay, last months were rather complicated.

What needs to be done :

- _Experiments on large datasets and results_
 	Mingkun/@HawkAaron helps me for review and training, currently only he has the available resources to train big networks on datasets such as Librispeech, WSJ, etc. He is rather busy lately, if someone could help on that it would be nice!
Most of the experiments I could do right now were on small or medium datasets (e.g.: an4, timit, voxforge it-fr, etc).

- _Performance stabilization, optimizations and code refactoring/rework_
	There are various additions, alternative implementations and optimizations already made but not available in this PR. Some available options in MTL which can be used with RNN-T/RNN-T-Att were also removed for the moment.
Feedback would be good to know what is needed right now as default setting in terms of performance, speed, memory usage, etc.

- _Integration_
       For now I separated the implementation from the MTL one but some parts are similar. I guess a combined version should be discussed after all is done.

- _Tests_

What is in work / next :

- Batch decoding support
- Streaming recognizers based on RNN-T and RNN-T-Att


ShigekiKarita:
Is it possible to use `--model-module` in `asr_train.py` option to select the RNNT model instead of creating `asr_rnnt_train.py`? The `--model-module` is already used for selecting RNN encdec and Transformer.

b-flo:
Actually I used different finetuning methods for the following reasons : 
     - There was no option for pre-initialization when I begun to write them
     - You can extract the modules of your choices to do partial initialization of both encoder / decoder
     - You can initialize both encoder and decoder with different AM (or LM of prediction network pre-initialization)

It's only there right now because I wanted to try to reproduce some setup from Google and Baidu (pre-init partially specific modules from each part, freeze specific layers, etc). In the long term these methods should be removed.

For asr_train.py and asr_rnnt_train, as I said previously the separation is done to not mix mtl / rnnt right now. I can merge them all together but I don't know how you want me to handle mtlalpha mode with rnnt mode?

ShigekiKarita:
I see. Actually, `ASRInterface` supports model-specific arguments by overriding `add_arguments`. So you can put `--enc-init`, `--dec-init`, etc there. Also we are trying to remove more model-specific options in `asr_train.py` (see https://github.com/espnet/espnet/pull/941). However, this request can be accomplished in the later PR because this refactoring is wip. I just wanted to hear your idea.

b-flo:
Oh, that's nice! The main difference between MTL and RNNT/RNNT-Att is on the decoder so you're right it would be better to unify them and extend add_arguments in decoders_rnnt. There are other minor differences between asr/asr_rnnt and e2e_asr/e2e_asr_rnnt but it shouldn't be a problem.

However, we should first decide how to switch between MTL and RNNT. I'm no set but depending on your goals with RNNT it could be done differently. For example:
 Unified mode -> 0: CTC, ]0,1[: JCA, 1, Att, 2-RNNT, 3-RNNT-Att
 CTC extension -> add RNNT as extension argument for CTC (some tweaks needed however)

About the options -`-enc-init` and `--dec-init` in the long run it would be better to have a global method handling pre-initialization and freezing given multiples models and lists of modules. It would also be useful for MTL.

ShigekiKarita:
I agree but the smaller PR might be the best at this time. I think that this PR can add just new networks and its `arguments` for RNNT under the `nets` modul (and reuse `asr` parts). Then we can reconsider the MTL and loading pre-trained networks for global later.

b-flo:
That's fine with me! However, imo, encoder pre-initialization should be available in the smaller PR. I still lack some feedback but from what I experimented and what was reported, encoder initialization with pre-trained CTC model is the first thing you want to get acceptable performances, close to MTL.
I'll do the changes as soon as possible, tomorrow or the day after that if that's okay.

Also, I would like to discuss default settings and desired performances (TER/WER). Right now I'm not really sure what should be quickly available for RNN-T/RNN-T-Att (regularization, endpointing/end-block, auxiliary functions, prefix beam search, etc) and what already implemented MTL options (label smoothing, sampling prob, etc) should be accessible from RNN-T decoders. I can already add/restore most of the features and it would only affect the new decoder module. If @HawkAaron or someone else has some time in the following weeks for experiments/feedback, it would help!

sw005320:
Thanks. Once we merge your PRs, we can help your implementations. So, as @ShigekiKarita mentioned, I'd appreciate if you make your current PR smaller one.

ShigekiKarita:
Yes. We can add pre-initialization or any options into RNNT by `add_arguments` but I think adding them to the existing models is too much for this PR. And also the performance is not important now because we do not have tried RNNT before... We can improve it later.

b-flo:
@ShigekiKarita I made the changes as you requested, however there are small issues : 

- For now transducer is like a CTC extension, thus it only works with mtlalpha=1.0, criterion=loss and  early-stop-criterion="validation/main/loss" as I specified in egs/voxforge/asr1/train_transducer.yaml. The reason is that I don't deal with accuracy in transducer so there is no "validation/main/acc" reported (which throw an error if the previous options are not set as said) nor validation.acc.best created.

- I added two recog arguments to bin/asr_recog.py because there is no argument extension from there. If it's a problem I can remove the options (search_type=['greedy', 'beam'], score_norm=[True, False]).

ShigekiKarita:
Thanks. It looks much better now. The first one might be no problem. I will add a few comments.

espnet/bin/asr_recog.py
```diff
@@ -71,6 +73,13 @@ def get_parser():
                         help='Input length ratio to obtain min output length')
     parser.add_argument('--ctc-weight', type=float, default=0.0,
                         help='CTC weight in joint decoding')
+    # transducer related
+    parser.add_argument('--search-type', type=str, default='beam',
```
ShigekiKarita: Can you use `--beam-size 1` for greedy decoding instead of adding new option?

espnet/bin/asr_recog.py
```diff
@@ -71,6 +73,13 @@ def get_parser():
...
+                        choices=['greedy', 'beam'],
+                        help='Search algorithm to use for transducer.')
+    parser.add_argument('--score-norm', type=strtobool, nargs='?',
+                        default=True,
+                        help='Normalize transducer scores by length')
```
ShigekiKarita: Maybe `--score-norm-transducer` is better because this option can be seen from non-transducer models.

espnet/nets/pytorch_backend/e2e_asr_transducer.py
```diff
@@ -0,0 +1,414 @@
...
+    """
+
+    @staticmethod
+    def add_arguments(parser):
+        group = parser.add_argument_group("transducer model setting")
+        group.add_argument('--rnnt_type', default='warp-transducer', type=str,
+                           choices=['warp-transducer'],
+                           help='Type of transducer implementation to calculate loss.')
+        parser.add_argument('--rnnt-mode', default=0, type=int, choices=[0, 1],
+                            help='RNN-Transducing mode (0:rnnt, 1:rnnt-att)')
```
ShigekiKarita: `'--rnnt-mode', default=0, type=str, choices=['rnnt', 'rnnt-att'],` is more expressive

espnet/nets/pytorch_backend/e2e_asr_transducer.py
```diff
@@ -0,0 +1,414 @@
...
+            enhanced, hlens, mask = self.frontend(hs, ilens)
+            hs, hlens = self.feature_transform(enhanced, hlens)
+        else:
+            hs, hlens = hs, ilens
+
+        # 1. Encoder
+        h, _, _ = self.enc(hs, hlens)
+
+        # 2. Decoder (pred+joint or att-dec+joint)
+        if recog_args.search_type == 'greedy':
```
ShigekiKarita: I mean `recog_args.beam_size == 1` here

espnet/nets/pytorch_backend/e2e_asr_transducer.py
```diff
@@ -0,0 +1,414 @@
...
+        else:
+            hs_pad, hlens = xs_pad, ilens
+
+        # 1. Encoder
+        hs_pad, hlens, _ = self.enc(hs_pad, hlens)
+
+        # 2. Decoder
+        hlens = torch.tensor(list(map(int, hlens)))
+
+        if recog_args.search_type == 'greedy':
```
ShigekiKarita: ditto

espnet/nets/pytorch_backend/e2e_asr_transducer.py
```diff
@@ -0,0 +1,414 @@
...
+                hs_pad, hlens = xs_pad, ilens
+
+            # encoder
+            hpad, hlens, _ = self.enc(hs_pad, hlens)
+
+            # decoder
+            if self.rnnt_mode == 1:
+                att_ws = self.dec.calculate_all_attentions(hpad, hlens, ys_pad)
+            else:
+                att_ws = []
```
ShigekiKarita: I think you can `return []` at the beginning of this method. If so, we can skip the computation in frontend and encoder.

espnet/nets/pytorch_backend/e2e_asr_transducer.py
```diff
@@ -0,0 +1,414 @@
...
+            recog_args (namespace): argument Namespace containing options
+            char_list (list): list of characters
+            rnnlm (torch.nn.Module): language model module
+
+        Returns:
+           y (list): n-best decoding results
+        """
+
+        logging.error("Batch decoding for transducer is not supported yet.")
+        raise NotImplementedError
```
ShigekiKarita: You can simply remove this method because this method is dead code

b-flo: I should not raise an error if batch decoding is called?

ShigekiKarita: I see. Maybe we can move these docstring and `NotImplementedError` to `ASRInterface.recognize_batch` instead of here.

ShigekiKarita:
I want you to add tests similar to existing tests for encoder-decoder models that performs training and decoding with dummy data like https://github.com/espnet/espnet/blob/e37479397d0944f1d287560f22f5dbeebf311acd/test/test_e2e_asr.py#L207
because running voxforge for regression tests are hard.

b-flo:
I addressed your comments, I'll add the tests asap! One question though, I didn't know how you wanted me to extend encoder, decoder and attention arguments in` e2e_asr_transducer`? For example there are different "styles" in `e2e_asr` and `e2e_asr_transformer`.

Also, I only added a recipe to voxforge as an example. Do you want me to remove it?

espnet/nets/pytorch_backend/e2e_asr_transducer.py
b-flo: I removed the implementation as suggested for now. I can do another PR to do the changes in asr/asr.py when we are set on the approach.

ShigekiKarita:
Many thanks! 

> One question though, I didn't know how you wanted me to extend encoder, decoder and attention arguments in e2e_asr_transducer?

In fact, we are moving to the `e2e_asr_transformer` style. This PR https://github.com/espnet/espnet/pull/1065  makes `e2e_asr` arguments similar to `e2e_asr_transformer` and has been merged into master today. So I recommend you to follow the new style.

> Do you want me to remove it?

No, you can keep it. I just want you to add tests to prevent breaking changes in the future.

b-flo:
> In fact, we are moving to the e2e_asr_transformer style. This PR #1065 makes e2e_asr arguments similar to e2e_asr_transformer and has been merged into master today.

Sorry I think "style" was the wrong term. Right now `e2e_asr` defines all model arguments in `add_arguments` whereas `e2e_asr` has separated methods to get the arguments from the different model parts (e.g.: `encoder_add_arguments`, etc).
In the current version, I followed  `e2e_asr` _style_.

ShigekiKarita:
I got it. OK.

b-flo:
@ShigekiKarita Sorry to bother you with that but I'm confused by the codecov/project failure and report. What is the problem in this case?

kan-bayashi:
This is due to the decreasing of coverage.
If you add test, the coverage will increase and become status OK.

b-flo:
Thank you. What I don't understand is why it said the coverage is decreased by removed lines in `espnet/nets/pytorch_backend/fastspeech/length_regulator.py` after merging master. It should have been covered by #1083? Before that the coverage was ok.

Sorry for the questions, I'm not too familiar with codecov yet. I'll delete these messages after that to not mess with the PR.

kan-bayashi:
> Sorry for the questions, I'm not too familiar with codecov yet. 

Me too :) It is a little bit difficult to understand how codecov behave. 
The test script of `length_regulator` has randomness and sometimes the test script does not go through following part:
https://github.com/espnet/espnet/blob/f1bb2416de55f14610d291e3c95dde1ec5028d7c/espnet/nets/pytorch_backend/fastspeech/length_regulator.py#L74-L76
This is reason of decreasing of coverage, I think.
So you do not need to care about that.

b-flo:
Oh, that behaviour is... interesting. I'll add later tests as you said.
Many thanks for the explanation, I'm deleting previous message to no mess with current PR!

b-flo:
Hi,

Sorry I had other works to deal with, I just added test_e2e_asr_transducer.py with basic tests.
Interesting enough I found out my beam search could get stuck if I don't compute score for blank symbol in the beam loop. On rare occasion, If k_range is really small I could end up without emitting any blank symbol on the vertical transitions. I'll investigate and add optimizations or switch implementation.

sw005320:
> Edit : All tests in test_e2e_asr_transducer raise a ModuleNotFoundError for warprnnt_pytorch in CI however I can't reproduce the error locally. Does someone have any idea?

You may need to add the installation manually like https://github.com/espnet/espnet/blob/master/ci/install.sh#L53-L57

b-flo:
Thanks a lot! I don't really understand with I couldn't reproduce the error locally, even on another server/computer, but well it works now!

ShigekiKarita:
LGTM.

@b-flo  Please remove "WIP" from the PR title if you think it is ready to merge.

b-flo:
I modified the title. Note that I addressed your review for `recognize_batch` in `e2e_asr_transducer.py` however it's not handled in `asr/asr.py` as you suggested. Also, an error will be throw if mtlalpha=0.0 as said previously.

> I see. Maybe we can move these docstring and NotImplementedError to ASRInterface.recognize_batch instead of here.

About the initial objectives, if it's okay I'll open another PR to discuss experiments and optimizations/features in transducer after both this PR and transfer learning are merged.

sw005320:
LGTM. Many thanks, @b-flo ! 
@HawkAaron, we'll merge this PR, but we're still welcome your comments/reviews for this PR.

PR #1066, opened 2019-08-08
Add data augmentation to ami recipe

Jzmo:


egs/ami/asr1/run.sh
```diff
@@ -108,6 +108,13 @@ if [ ${stage} -le 0 ] && [ ${stop_stage} -ge 0 ]; then
 	PROCESSED_AMI_DIR=${AMI_DIR}
     fi
     local/ami_${base_mic}_data_prep.sh ${PROCESSED_AMI_DIR} ${mic}
+    # data augmentation
+    utils/perturb_data_dir_speed.sh 0.9 data/${mic}/train_orig data/ihm_tmp1
```
sw005320: `data/${mic}/train_orig` is first appeared.
I think you need to prepare it.

Jzmo: Hi, here `ami_${base_mic}_data_prep.sh` in line 110 automatically create the `data/ihm/train_orig` path when the ${mic} is ihm.

sw005320: I see. Thanks.

Jzmo: Also, do you think I need to upload the model files?

sw005320: This is very helpful. I sent an invitation to Google Drive. Please upload your model and modify the link accordingly.

sw005320:
Many thanks!
Can you also try specaugment?
If you check https://github.com/espnet/espnet/blob/master/egs/librispeech/asr1/run.sh#L23, you could see how you can use it.
It would be great if you check the following setups
speed purturb. vs. specaug. vs. speed purturb+specaug

PR #1072, opened 2019-08-09
Fix swbd recipe v2

sw005320:
This is a follow-up PR of #1024


sw005320:
We should also use an LM trained with fisher+swbd to make the experimental condition similar to the other reports.

sw005320:
I found that the number we got is not so bad compared with several references without LM found in https://arxiv.org/pdf/1904.08779.pdf 

|        | SWBD | CH   |
|--------|------|------|
| RWTH   | 11.9 | 23.7 |
| Google | 11.2 | 21.6 |
| Ours   | 12.2 | 22.1 |

I'll also play with the specaugment a little bit with another PR, but I think the current recipe is good enough for our future research/development.

PR #1074, opened 2019-08-11
Fix installation error in conda

kan-bayashi:
This PR fixes following error.
https://circleci.com/gh/espnet/espnet/5128


kan-bayashi:
I could not find out the reason :(

PR #1079, opened 2019-08-13
Replace character-level recipe with the BPE one in iwslt18

hirofumi0810:
- ASR: character -> BPE1k
- ST/NMT: character -> minimum BPE (equal to character)


egs/iwslt18/asr1/run.sh
```diff
@@ -43,6 +48,13 @@ case=lc.rm
 st_ted=/export/b08/inaguma/IWSLT
 # st_ted=/n/rd11/corpora_8/iwslt18
 
+# bpemode (unigram or bpe)
+nbpe=1000
+bpemode=bpe
+# NOTE: nbpe=69 means character-level ASR (case=lc.rm)
```
sw005320: I like this way, but another possibility is to get this number automatically when we set `nbpe=0`. But this may make `run.sh` complicated, and the current way is good for me.

egs/iwslt18/st1/conf/tuning/train_rnn_char.yaml
```diff
@@ -30,6 +30,7 @@ weight-decay: 1e-6
 # optimization related
 sortagrad: 0 # Feed samples from shortest to longest ; -1: enabled for all epochs, 0: disabled, other: enabled for 'other' epochs
 opt: adadelta
+eps-decay: 0.01
```
sw005320: This is interesting. Someone reported some improvement by making `eps-decay` larger. Did you also observe the improvement?

utils/spm_train
```diff
@@ -14,4 +14,4 @@ import sentencepiece as spm
 
 
 if __name__ == "__main__":
-    spm.SentencePieceTrainer.Train(" ".join(map(shlex.quote, sys.argv[1:])))
+    spm.SentencePieceTrainer.Train(" ".join(map(shlex.quote, sys.argv[1:])).replace("\'", ""))
```
sw005320: Can you add some comments on this change?

egs/iwslt18/asr1/run.sh
hirofumi0810: nbpe=0 causes an error in spm_train, which requires the second run. I don't prefer it.

egs/iwslt18/st1/conf/tuning/train_rnn_char.yaml
hirofumi0810: Yes. eps-decay: 0.01 is a little bit large for BPE models.

hirofumi0810:
All done.

hirofumi0810:
I prefer to merge into v.0.6.0 instead of v.0.5.2 for the compatibility with the upcoming PRs.

sw005320:
> I prefer to merge into v.0.6.0 instead of v.0.5.2 for the compatibility with the upcoming PRs.

We merge the v.0.5.2 change to v.0.6.0 frequently. So, it should be no problem to marge this one to the master. But if you still prefer to merge it to v.0.6.0, I'll do it.

hirofumi0810:
OK.

sw005320:
Thanks a lot!

PR #1083, opened 2019-08-14
Fix multi gpu LM issues and add hdf5 LM dataset dump

ShigekiKarita:
This PR aims to fix #1075 . I tested this by AN4 and TEDLIUM3. 

I found @kan-bayashi's comment was correct. The memory leak appeared when single GPU performs the frame reduction of loss values from multi GPUs. So I added the `ReduceFramewiseLoss` module to do it in multi GPUs.

In addition, I added HDF5 dump support for LM dataset because some large corpora, e.g., TEDLIUM3, results in 3GB of text dataset and it took 3 mins for preprocessing. After HDF5 dump performed, it only took 30 secs to load that. It really helps me testing this problem, and running rapid try-and-errors in the large LM experiment. (PS. I also tested pickle but it seems to perform very slow and save large bin.)

Issue #489, opened 2018-11-30
Discussion to refactor beam search


ShigekiKarita:
Recently, we have several issues like #429, #436, #470 etc and improvements like #318, #478 etc related to beam search. As far as I know, every PRs related to beam search (except for the wide refactoring #484) has been merged at this time. Maybe it is good time to refactor current complex and adhoc implentations to be simpler and generalized for v0.3.0.

These are my thoughts on the problems in current beam search
- Beam search should not be inside the Decoder class because it requires to reimplement beam search in very different model like Transformer.
- Beam search should not hard code initialization and state updates of scoring models (decoder, CTC, LM, etc)
- Hard to read many lines. Some lines should be function or class.
- Hard to add new scoring models (e.g., #429)

I propose two changes
- Define explicit interface class for scoring model (Decoder, CTC, LM, etc) that can be used in beam search
- Beam search class only call `.init_state`, `.score` and `.select_state` method of the scoring models.

This is just my draft what the new BeamSearch class looks like
```python
# interface for Decoder, CTC, RNNLM, etc
class ScoringBase(object):
    def score(self, token, enc_output, state):
        """return log_prob, new_state"""
        pass

    def init_state(self, src, src_lengths, enc_output, args):
        """return batch/beam state object"""
        pass

    def select_state(self, state, index):
        """return batch/beam state object selected on the index"""
        pass


class BeamSearch(object):
    def __init__(self, encoder, scorers, weight_dict, sos, eos):
        ...

    def init_state(self, n_batch, n_beam, device):
        ...

    def prune(self, global_best_score, local_best_score, eos_mask):
        ...

    def update_tokens(self, global_best_token_id, global_best_tokens, eos_mask):
        ...

    def recognize(self, src, src_length, args, device):
        # forward encoder
        enc_output = self.encoder(src, src_lengths)

        # prepare states
        states = dict():
        for s in self.scorers:
            states[s] = s.init_state(src, src_lengths, enc_output, args)

        n_batch = len(src_length)
        n_beam = args.beam_size
        global_best_token_seq = [[[] for _ in range(n_beam)] for _ in range(n_batch)]
        global_best_token, global_best_score, eos_mask = self.init_state(n_batch, n_beam, device)

        # prefix search
        for step in range(args.maxlen_decode):
            # forward scorers
            local_score = 0
            new_scores = dict()
            new_states = dict()
            for s in self.scorers:
                new_scores[s], new_states[s] = s.score(
                    global_best_token,
[...]

PR #980, opened 2019-07-21
Improve batch decoding

takaaki-hori:

PR #1092, opened 2019-08-15
New beam-search framework: ScorerInterface

ShigekiKarita:
This PR refactors beam search as discussed in #489 

- [x] wrap previous RNNLM as `espnet.nets.pytorch_backend.lm.legacy:LegacyRNNLM`
- [x] dynamic import using LMInterface
- [x] decoding with LMInterface abstraction instead of current hard-coded ones for RNNLM
  - [x] Transformer ASR Model
  - [x] Length Penalty
  - [x] RNN ASR Model
  - [x] RNNLM
  - [x] CTC Prefix Score
- [x] support more predefined/user-defined LMs, e.g., https://github.com/pytorch/examples/blob/master/word_language_model/model.py
- [x] GPU decoding
- [x] multi precision (float16/32/64) support
- [x] Transformer LM
- [x] fix wrong ppl reported in #1083 

I will implement these essential feature not to break current features
- custom language model training with `lm_train.py --model-module xxx`
- ASR decoding with custom language model `asr_recog.py --lm xxx` (since it is not `--rnnlm` anymore. I will leave the existing `--rnnlm` as-is.)

and I will not support the other existing features with custom LMs like ASR training with custom LMs, GPU batch decoding, streaming decoding, word LM, etc in this PR.


sw005320:
> and I will not support the other existing features with custom LMs like ASR training with custom LMs, GPU batch ASR decoding, etc.

Actually, the current GPU batch ASR decoding does not do special things for rnnlm. See https://github.com/espnet/espnet/blob/f1bb2416de55f14610d291e3c95dde1ec5028d7c/espnet/lm/pytorch_backend/lm.py#L134-L146
So, you could just do the same thing for now. 
I want to revisit it after #980.

ShigekiKarita:
@sw005320  As there is no breaking change, I expect master is better.

sw005320:
> @sw005320 As there is no breaking change, I expect master is better.

OK.

ShigekiKarita:
I have finished easy ones: Transformer and length penalty because they are **state-less** in the current implementation. Tomorrow, I will implement the **stateful** ones (RNN ASR, RNN LM, CTC Prefix Score). Currently, the old and new implementations results completely same scores. However, IMO, I do not want to reimplement this local pruning with `ctc_beam` (its value is hard-coded)
https://github.com/espnet/espnet/blob/6759e7fcb8386b845a8a48ef779db3977f599a05/espnet/nets/pytorch_backend/e2e_asr_transformer.py#L320-L333
I have no idea to share scores between `decoder` and `ctc` modules because every decoder runs independently in my impl.
https://github.com/espnet/espnet/blob/be342ae3ff19e31decb26ab9183126bfc6990a55/espnet/nets/pytorch_backend/beam_search.py#L130-L133

Does this `ctc_beam` really matter? If so, I will reconsider how to do that...

sw005320:
> Does this `ctc_beam` really matter? If so, I will reconsider how to do that...

`ctc_beam` would have a big impact when we have large numbers of output units (e.g., Japanese, Chinese, and BPE). If we don't have it, the beam search would be significantly slower. Actually, the current batch beam search does not have `ctc_beam` and causes significant speed degradation for Japanese, Chinese, and BPE. One of the benefits of #980 is to fix this.

ShigekiKarita:
OK. I got how to do it. For simplicity, I will add new option like `pre_beam_ratio` that is same to this
https://github.com/espnet/espnet/blob/6759e7fcb8386b845a8a48ef779db3977f599a05/espnet/nets/pytorch_backend/e2e_asr_transformer.py#L287

So first `beam_search` performs topk with pre-beam size on pre-decoders scores (s2s, nn-lm), then add post-decoders scores (ctc, fst-lm?), and finally topk with the full beam size.

ShigekiKarita:
Now I have implemented all the legacy functions above. I made unittests that are exactly matched to previous `model.recognize` methods in `mtlalpha=0.0,0.5,1.0` with or without LM and penalty in RNN and Transformer.

I am also testing new beam search api with AN4. Their final results are little bit different. Maybe is is ignorable?
```
$ ./run.sh --ngpu 1 --stage 5 --use-wordlm false --decode-config ./conf/decode_v2_ctcweight1.0.yaml
write a CER (or TER) result in exp/train_nodev_pytorch_train_mtlalpha1.0/decode_train_dev_decode_ctcweight1.0_lm/result.txt
|  SPKR     |  # Snt    # Wrd  |  Corr       Sub       Del       Ins      Err     S.Err   |
|  Sum/Avg  |   100      1915  |  86.5       5.8       7.7       3.7     17.1      77.0   |
write a CER (or TER) result in exp/train_nodev_pytorch_train_mtlalpha1.0/decode_test_decode_ctcweight1.0_lm/result.txt
|  SPKR    |  # Snt   # Wrd   |  Corr      Sub      Del      Ins       Err    S.Err  |
|  Sum/Avg |   130     2565   |  91.9      4.7      3.5      3.0      11.2     62.3  |
```

this result is made by v2
```
$ ./run.sh --ngpu 1 --stage 5 --use-wordlm false --decode-config ./conf/decode_v2_ctcweight1.0.yaml
write a CER (or TER) result in exp/train_nodev_pytorch_train_mtlalpha1.0/decode_train_dev_decode_v2_ctcweight1.0_lm/result.txt
|  SPKR     |  # Snt    # Wrd   |  Corr        Sub       Del       Ins       Err     S.Err   |
|  Sum/Avg  |   100      1915   |  84.9        5.9       9.2       2.5      17.5      77.0   |
write a CER (or TER) result in exp/train_nodev_pytorch_train_mtlalpha1.0/decode_test_decode_v2_ctcweight1.0_lm/result.txt
|  SPKR     |  # Snt   # Wrd   |  Corr      Sub       Del      Ins       Err    S.Err   |
|  Sum/Avg  |   130     2565   |  91.3      4.6       4.1      2.1      10.8     62.3   |
```

ShigekiKarita:
I have implemented CUDNN based RNNLM named `fast-rnn`. On AN4, it seems working but I have tuned nothing.

training

- fast-rnn: ppl 1.00, 13 sec (20 epoch)
- default rnn: ppl 4.96, 9 sec (6 epoch)

Our default rnn was really fast! (it does not use cudnn)

decoding results
```
./run.sh  --ngpu 1 --lm-config ./conf/lm_fast.yaml --use-wordlm false --decode-config ./conf/decode_v2_ctcweight1.0.yaml
write a CER (or TER) result in exp/train_nodev_pytorch_train_mtlalpha1.0/decode_train_dev_decode_v2_ctcweight1.0_lm_fast/result.txt
|   SPKR     |  # Snt     # Wrd   |   Corr       Sub        Del        Ins       Err      S.Err   |
|   Sum/Avg  |   100       1915   |   82.2       4.5       13.2        1.5      19.3       85.0   |
write a CER (or TER) result in exp/train_nodev_pytorch_train_mtlalpha1.0/decode_test_decode_v2_ctcweight1.0_lm_fast/result.txt
|  SPKR     |  # Snt    # Wrd   |  Corr        Sub       Del       Ins       Err     S.Err   |
|  Sum/Avg  |   130      2565   |  89.2        3.6       7.1       1.1      11.9      75.4   |

```

ShigekiKarita:
I also checked TEDLIUM3 stage 3 LM training (lm.yaml) by my gtx1080ti because it has more layers and maxlen than AN4

- default rnn: 1.80 iters/sec, ETF 3 days 6 hours
- cudnn rnn: 1.51 iters/sec, ETF 3 days 20 hours

hmm... our default looks very nice.

egs/mini_an4/asr1/run.sh
```diff
@@ -11,6 +11,7 @@ backend=pytorch
 stage=-1       # start from -1 if you need to start from data download
 stop_stage=100
 ngpu=0         # number of gpus ("0" uses cpu, otherwise use gpu)
+decode_ngpu=0
```
sw005320: Probably we should do it for all recipes. Let's do it with the other PR.

espnet/asr/pytorch_backend/recog.py
```diff
@@ -0,0 +1,94 @@
...
+from espnet.asr.pytorch_backend.asr import load_trained_model
+from espnet.nets.asr_interface import ASRInterface
+from espnet.nets.lm_interface import dynamic_import_lm
+from espnet.nets.pytorch_backend.beam_search import beam_search
+from espnet.nets.pytorch_backend.beam_search import LengthBonus
+from espnet.utils.deterministic_utils import set_deterministic_pytorch
+from espnet.utils.io_utils import LoadInputsAndTargets
+
+
+def recog_v2(args):
```
sw005320: We need some explanations between default and v2 here and the argument help.

espnet/nets/pytorch_backend/beam_search.py
```diff
@@ -0,0 +1,297 @@
...
+                and next state for ys
+        """
+        import torch
+        return torch.tensor([1.0]).expand(self.n), None
+
+
+def get_weighted_decoders(
+        decoders: Dict[str, DecoderInterface],
+        weights: Dict[str, float]
+) -> Dict[str, Tuple[DecoderInterface, float]]:
```
sw005320: Is this line correct?

espnet/nets/pytorch_backend/beam_search.py
```diff
@@ -0,0 +1,297 @@
...
+                token_list=None, maxlenratio=0.0, minlenratio=0.0,
+                pre_beam_ratio=1.5, pre_beam_score="decoder"):
+    """Beam search with scorers
+
+    Args:
+        x (torch.Tensor): Encoded speech feature (T, D)
+        sos (int): Start of sequence id
+        eos (int): End of sequence id
+        beam_size (int): The number of hypotheses kept during search
+        decoders (dict[str, DecoderInterface]): list of decoder modules
```
sw005320: Can you add an example of multiple decoders here (attention encoder-decoder, CTC, RNNLM?)?

sw005320:
@kan-bayashi and @takaaki-hori, could you check this review?
I want to make this PR better shape soon and also want to combine #979 in the later PR.
Then, I believe we could make the beam search implementation more fast and efficient.

egs/mini_an4/asr1/run.sh
ShigekiKarita: I agree

ShigekiKarita:
I have greatly changed the design of `beam_search.py` from a single function to torch.nn.Module.

- function https://github.com/espnet/espnet/blob/e9a20684aa804625dde75dfedcbffd04c225c7db/espnet/nets/beam_search.py
- nn.Module version https://github.com/espnet/espnet/blob/e0288b141173da88ee87e85060750af241bdb170/espnet/nets/beam_search.py

Pros of Module:
- easy to handle beam-search device (e.g., `BeamSearch(...).cuda()`) and precision (e.g., `BeamSearch(...).half()`)
- entensibility by inheritance
- chance to make this autograd
- change to use JIT in some methods (or porting to C++)

Cons of Module:
- it is longer than before
- difficulty to support chainer but it is still possible unless someone wanna support backprop through beam-search

As I listed above, there are both good and bad parts but I like the Module approach.

ShigekiKarita:
I supported float16/32/64 precision decoding in this PR. Maybe I will do float16/32/64 training in the later PR.

ShigekiKarita:
To test new LMs, I fixed PPL problem in #1083. Now I added Transformer LM example to AN4

WER: 
```
$ ./run.sh --ngpu 1 --lm-config $(change_yaml.py ./conf/lm.yaml -a model-module=transformer) --decode-config $(change_yaml.py ./conf/decode_ctcweight1.0.yaml -a api=v2)
write a CER (or TER) result in exp/train_nodev_pytorch_train_mtlalpha1.0/decode_train_dev_decode_v2_ctcweight1.0_lm_transformer/result.txt|   SPKR     |   # Snt      # Wrd   |   Corr         Sub        Del         Ins        Err       S.Err   |
|   Sum/Avg  |    100        1915   |   82.0         6.2       11.9         1.9       19.9        81.0   |
write a CER (or TER) result in exp/train_nodev_pytorch_train_mtlalpha1.0/decode_test_decode_v2_ctcweight1.0_lm_transformer/result.txt
|   SPKR     |   # Snt     # Wrd   |   Corr       Sub        Del        Ins        Err      S.Err   |
|   Sum/Avg  |    130       2565   |   88.5       5.0        6.5        1.7       13.3       79.2   |
```
In my PC, TransformerLM results decoding speed 3 times faster than RNNLM.


- existing char RNNLM result 

WER (api v1)
```
write a CER (or TER) result in exp/train_nodev_pytorch_train_mtlalpha1.0/decode_train_dev_decode_ctcweight1.0_lm/result.txt
|  SPKR     |  # Snt    # Wrd  |  Corr       Sub       Del       Ins      Err     S.Err   |
|  Sum/Avg  |   100      1915  |  85.8       5.8       8.3       3.7     17.9      81.0   |
write a CER (or TER) result in exp/train_nodev_pytorch_train_mtlalpha1.0/decode_test_decode_ctcweight1.0_lm/result.txt
|  SPKR    |  # Snt   # Wrd   |  Corr      Sub      Del      Ins       Err    S.Err  |
|  Sum/Avg |   130     2565   |  91.1      4.6      4.3      2.5      11.4     66.2  |
```
- existing char RNNLM result (api v2)
```
/decode_train_dev_decode_v2_ctcweight1.0_lm/result.txt
|  SPKR     |  # Snt    # Wrd   |  Corr        Sub       Del       Ins       Err     S.Err   |
|  Sum/Avg  |   100      1915   |  84.4        5.8       9.8       2.8      18.4      79.0   |
write a CER (or TER) result in exp/train_nodev_pytorch_train_mtlalpha1.0/decode_test_decode_v2_ctcweight1.0_lm/result.txt
|  SPKR     |  # Snt   # Wrd   |  Corr      Sub       Del      Ins       Err    S.Err   |
|  Sum/Avg  |   130     2565   |  90.5      4.1       5.4      1.8      11.3     68.5   |
```

sw005320:
> I have greatly changed the design of `beam_search.py` from a single function to torch.nn.Module.
> 
> * function https://github.com/espnet/espnet/blob/e9a20684aa804625dde75dfedcbffd04c225c7db/espnet/nets/beam_search.py#L57
> * nn.Module version https://github.com/espnet/espnet/blob/d9de4a468bdde94304c8bbb95db068c444ecb0c7/espnet/nets/beam_search.py#L58
> 
> Pros of Module:
> 
> * easy to handle beam-search device (e.g., `BeamSearch(...).cuda()`) and precision (e.g., `BeamSearch(...).half()`)
> * extensibility by inheritance
> * chance to make this autograd
> * chance to use JIT in some methods (or porting to C++)
> 
> Cons of Module:
> 
> * it is larger than before
> * difficulty to support chainer but it is still possible unless someone wanna support backprop through beam-search
> 
> As I listed above, there are both good and bad parts but I like the Module approach.

I like this change as well.
Now we don't have to have the beam search for each decoder, and the above flexibilities lead to solving the slow decoding problem.
Great job, @ShigekiKarita.

@Fhrozen, can you consider whether you can do the same thing for chainer?

espnet/nets/pytorch_backend/rnn/decoders.py
```diff
@@ -714,6 +715,34 @@ def _index_select_lm_state(rnnlm_state, dim, vidx):
...
         return new_state
 
+    # decoder interface methods
+    def init_state(self, x):
+        self.c_list = [self.zero_state(x.unsqueeze(0))]
+        self.z_list = [self.zero_state(x.unsqueeze(0))]
+        for _ in six.moves.range(1, self.dlayers):
+            self.c_list.append(self.zero_state(x.unsqueeze(0)))
+            self.z_list.append(self.zero_state(x.unsqueeze(0)))
+        # TODO(karita): what is this?
```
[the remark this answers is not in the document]
kan-bayashi: I think this is related to asr_mix.

kan-bayashi:
Sorry for the late reply.
I will check today.

espnet/nets/pytorch_backend/rnn/decoders.py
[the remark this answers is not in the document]
ShigekiKarita: OK

kan-bayashi:
I will check in detail from now, but I have some comments:

- If you can afford, it is better to follow the manner of `flake8-docstrings`.
- It is better to change some methods to private method.

kan-bayashi:
If you add the example to `ScorerInterface.score()` and `PartialScorerInterface.score_partial()`, they are very helpful to understand.

espnet/bin/asr_recog.py
```diff
@@ -153,7 +159,14 @@ def main(args):
...
         elif args.backend == "pytorch":
+            # Experimental API that supports custom LMs
+            if args.api == "v2":
+                from espnet.asr.pytorch_backend.recog import recog_v2
+                recog_v2(args)
+                return
             from espnet.asr.pytorch_backend.asr import recog
+            if args.dtype != "float32":
+                raise NotImplementedError(f"`--dtype {args.dtype}` is only available with `--api v2`")
             recog(args)
```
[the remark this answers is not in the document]
ShigekiKarita: OK

espnet/asr/pytorch_backend/recog.py
```diff
@@ -0,0 +1,97 @@
...
+        if args.preprocess_conf is None else args.preprocess_conf,
+        preprocess_args={'train': False})
+
+    lm = None
+    if args.rnnlm:
+        lm_args = get_model_conf(args.rnnlm, args.rnnlm_conf)
+        lm_class = dynamic_import_lm(lm_args.model_module, lm_args.backend)
+        lm = lm_class(len(train_args.char_list), lm_args)
+        torch_load(args.rnnlm, lm)
+        lm.eval()
```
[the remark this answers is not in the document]
ShigekiKarita: OK

espnet/lm/pytorch_backend/lm.py
```diff
@@ -307,14 +95,13 @@ def update_core(self):
...
         # (it is chainer.dataset.concat_examples by default)
         x, t = concat_examples(batch, device=self.device, padding=(0, -100))
-        loss, count = self.model(x, t)
-        loss = loss.sum()
-        count = count.sum()
-        reporter.report({'loss': float(loss.detach() / count)}, optimizer.target)
-        reporter.report({'count': int(count)}, optimizer.target)
+        loss, logp, count = self.model(x, t)
+        reporter.report({'train_loss': float(loss.mean())}, optimizer.target)
+        reporter.report({'loss': float(logp.sum())}, optimizer.target)
```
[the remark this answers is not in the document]
ShigekiKarita: Strongly agree. Let me explain why I made these three return values. It is complicated.

1. `loss`: this is only used for `loss.mean().backward()` to train NN. It can be different from logp because of normalization.
2. `logp`: this is used as numerator of perplexity. I rethink its should be named to `nll` (negative log likelihood) because it is negative.
3. `count`: this is used as the denominator of perplexity.

To rename the report key 'loss' into `nll` needs to change `compute_perplexity` function and  chainer backend as well. However, I will try that.

The reason why evaluator puts `logp / count` in `'loss'`
```python
reporter.report({'loss': logp / count}, self.model.reporter)
```
is that evaluator does not use `compute_perplexity` that computes ppl with averaged `results['main/loss'] / results['main/count']` online. So I think the code above should be something like:
```python
reporter.report({'test_perplexity': logp / count}, self.model.reporter)
```

espnet/lm/pytorch_backend/lm.py
```diff
@@ -337,26 +124,30 @@ def __init__(self, val_iter, eval_model, reporter, device):
...
         # report validation loss
         observation = {}
         with reporter.report_scope(observation):
-            reporter.report({'loss': float(loss.sum() / count.sum())}, self.model.reporter)
+            reporter.report({'loss': logp / count}, self.model.reporter)
         return observation
 
 
-def train(args):
+def train(args, model_class):
```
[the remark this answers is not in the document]
ShigekiKarita: OK

espnet/nets/beam_search.py
```diff
@@ -0,0 +1,294 @@
...
+        scorers (dict[str, ScorerInterface]): Dict of decoder modules e.g., Decoder, CTCPrefixScorer, LM
+            The scorer will be ignored if it is `None`
+        weights (dict[str, float]): Dict of weights for each scorers
+            The scorer will be ignored if its weight is 0
+        beam_size (int): The number of hypotheses kept during search
+        vocab_size (int): The number of vocabulary
+        sos (int): Start of sequence id
+        eos (int): End of sequence id
+        token_list (list[str]): List of tokens for debug log
+        pre_beam_score (str): key of scores to perform pre-beam search
```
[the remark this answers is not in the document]
ShigekiKarita: I think `pre_beam_score_key` might be better because it is key and it is not scorer itself.

espnet/nets/scorer_interface.py
```diff
@@ -0,0 +1,81 @@
...
+
+        Returns: initial state
+        """
+        return None
+
+    def score(self, y: torch.Tensor, state: Any, x: torch.Tensor) -> Tuple[torch.Tensor, Any]:
+        """Score new token
+
+        Args:
+            y (torch.Tensor): torch.int64 prefix token (U)
```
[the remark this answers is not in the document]
ShigekiKarita: It it the length of the prefix. I will add more description here.

ShigekiKarita:
> flake8-docstring

I haven't use that. I will follow all the docstring things.

> examples

I will add more notes on existing implementation of `score` (Transformer/RNN Decoder, RNNLM, ...) and `score_partial` (CTCPrefixScorer)

espnet/bin/lm_train.py
```diff
@@ -149,7 +148,7 @@ def main(args):
         train(args)
     elif args.backend == "pytorch":
         from espnet.lm.pytorch_backend.lm import train
-        train(args)
+        train(args, model_class)
```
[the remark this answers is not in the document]
ShigekiKarita: OK

test/test_beam_search.py
```diff
@@ -0,0 +1,174 @@
...
+        sos=model.sos,
+        eos=model.eos,
+    )
+    beam.to(device, dtype=dtype)
+    beam.eval()
+    with torch.no_grad():
+        enc = model.encode(torch.as_tensor(feat).to(device, dtype=dtype))
+        nbest_bs = beam(x=enc, maxlenratio=args.maxlenratio, minlenratio=args.minlenratio)
+        print(nbest_bs)
+    if dtype == torch.float16:
```
ShigekiKarita: memo: `if dtype != torch.float32` because `torch.float64` sometimes results differently from `model.recognize`

espnet/nets/pytorch_backend/transformer/attention.py
```diff
@@ -5,7 +5,7 @@
 from torch import nn
 
 
-MIN_VALUE = float(numpy.finfo(numpy.float32).min)
+MIN_VALUE = float(numpy.finfo(numpy.float16).min)
```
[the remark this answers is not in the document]
ShigekiKarita: I changed this line because the float16 tensor's `torch.masked_fill` does not accept the `float32.min` value because it is out-of numerical range. For strong compatibility, I will create a new function like
```python
def mask_by_min(x, mask):
    min_value = float(numpy.finfo(x.dtype).min)
    return x.maked_fill(min_value)
```

ShigekiKarita:
@kan-bayashi I have a suggestion for `flake8-docstrings`. I like this but it is not checked by our CI. So I want CI to do that as follows:
1. `pip install flake8-docstrings` in `ci/install.sh`
2. change `ci/test_python.sh` like
```bash
# --extend-ignore for wip files for flake8-docstrings
flake8 --extend-ignore=D espnet test utils

# white list of files that should support flake8-docstrings
flake8 \
  espnet/nets/beam_search.py \
  espnet/nets/lm_interface.py \
  espnet/nets/scorer_interface.py \
  ...
```
see how to disable flake8-docstrings temporally https://stackoverflow.com/questions/41860123/how-to-ignore-installed-flake8-plugin-easily-for-one-time

kan-bayashi:
That’s a nice idea.
Maybe the blacklist is better?
Anyway, let us introduce it.

ShigekiKarita:
I agree. I made the black list. No more invalid docstrings :beer: 

Now, I need to fix ppl things in `lm.py`. Just a moment.

espnet/lm/pytorch_backend/lm.py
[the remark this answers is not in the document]
ShigekiKarita: I have fixed this.

ShigekiKarita:
I believe it is done. Additionally, I split the flake8 part in `ci/test_python.sh` to `ci/test_flake8.sh` because it also helps rapid docstring annotation in local.  I also updated `doc/README.md` for this.

ShigekiKarita:
Fixed!

espnet/bin/asr_recog.py
```diff
@@ -153,8 +159,16 @@ def main(args):
             from espnet.asr.chainer_backend.asr import recog
             recog(args)
         elif args.backend == "pytorch":
-            from espnet.asr.pytorch_backend.asr import recog
-            recog(args)
+            # Experimental API that supports custom LMs
+            if args.api == "v2":
+                from espnet.asr.pytorch_backend.recog import recog_v2
+                recog_v2(args)
+                return
```
kan-bayashi:
```suggestion
```

espnet/lm/chainer_backend/lm.py
```diff
@@ -45,6 +46,26 @@
 from espnet.utils.training.train_utils import set_early_stop
 
 
+# TODO(karita): reimplement RNNLM with new interface
+class DefaultRNNLM(LMInterface, link.Chain):
+    """Default RNNLM wrapper to compute reduce framewise loss values
```
kan-bayashi:
```suggestion
    """Default RNNLM wrapper to compute reduce framewise loss values.
```

espnet/nets/asr_interface.py
```diff
@@ -49,3 +52,49 @@ def calculate_all_attentions(self, xs, ilens, ys):
     def attention_plot_class(self):
         from espnet.asr.asr_utils import PlotAttentionReport
         return PlotAttentionReport
+
+    def encode(self, feat):
+        '''encode feature in `beam_search` (optional)
```
kan-bayashi:
```suggestion
        '''Encode feature in `beam_search` (optional).
```

espnet/nets/asr_interface.py
```diff
@@ -49,3 +52,49 @@ def calculate_all_attentions(self, xs, ilens, ys):
...
+        Args:
+            x (numpy.ndarray): input feature (T, D)
+        Returns:
+            torch.Tensor for pytorch, chainer.Variable for chainer:
+                encoded feature (T, D)
+        '''
+        raise NotImplementedError("encode method is not implemented")
+
+    def scorers(self):
+        '''get scorers for `beam_search` (optional)
```
kan-bayashi:
```suggestion
        '''Get scorers for `beam_search` (optional).
```

espnet/nets/asr_interface.py
```diff
@@ -49,3 +52,49 @@ def calculate_all_attentions(self, xs, ilens, ys):
...
+    },
+    "chainer": {
+        "rnn": "espnet.nets.chainer_backend.e2e_asr:E2E",
+        "transformer": "espnet.nets.chainer_backend.e2e_asr_transformer:E2E",
+    }
+}
+
+
+def dynamic_import_asr(module, backend):
+    """Dynamic import ASR models
```
kan-bayashi:
```suggestion
    """Dynamic import ASR models.
```

espnet/nets/asr_interface.py
```diff
@@ -49,3 +52,49 @@ def calculate_all_attentions(self, xs, ilens, ys):
...
+
+    Args:
+        module (str): module_name:class_name or alias in `predefined_asr`
+        backend (str): NN backend. e.g., pytorch, chainer
+
+    Returns:
+        type: ASR class
```
kan-bayashi:
```suggestion
        type: ASR class

```

espnet/nets/asr_interface.py
```diff
@@ -49,3 +52,49 @@ def calculate_all_attentions(self, xs, ilens, ys):
...
         return PlotAttentionReport
+
+    def encode(self, feat):
+        '''encode feature in `beam_search` (optional)
+
+        Args:
+            x (numpy.ndarray): input feature (T, D)
+        Returns:
+            torch.Tensor for pytorch, chainer.Variable for chainer:
+                encoded feature (T, D)
```
kan-bayashi:
```suggestion
                encoded feature (T, D)

```

espnet/nets/asr_interface.py
```diff
@@ -49,3 +52,49 @@ def calculate_all_attentions(self, xs, ilens, ys):
...
+        '''
+        raise NotImplementedError("encode method is not implemented")
+
+    def scorers(self):
+        '''get scorers for `beam_search` (optional)
+
+        Returns:
+            dict[str, ScorerInterface]: dict of `ScorerInterface` objects
```
kan-bayashi:
```suggestion
            dict[str, ScorerInterface]: dict of `ScorerInterface` objects

```

sw005320:
Many thanks!

PR #1096, opened 2019-08-15
Chainer fix mixed input

### LINKED ISSUE #1091: chainer is not working with `--ngpu 0`
In my environment `master`, `egs/mini_an4/run.sh --ngpu 0 --backend chainer` is not working now. I attached train.log
[train.log](https://github.com/espnet/espnet/files/3504860/train.log)

Fhrozen:
To resolve #1091.

The problem is that for the first iteration, the custom updater return an `CpuDevice`, so when it is compared inside the `CustomConverter` the comparison return cuda.cupy. For some reason, in the first iter the CustomUpdater uses the super `CpuDevice` and then is set to an int `-1`, so it is not possible to use the xp from the  `CpuDevice`. Currently i applied the fix for both (rnn and transf) but transformer always uses np array in the CustomConverter so it does not show this problem.

### DIFF
--- a/ci/test_integration.sh
+++ b/ci/test_integration.sh
@@ -2,6 +2,13 @@
 
 set -euo pipefail
 
+if [ ! -z "${LD_LIBRARY_PATH}" ]; then
+    export LD_LIBRARY_PATH=${LD_LIBRARY_PATH}:$(pwd)/chainer_ctc/ext/warp-ctc/build
+else
+    export LD_LIBRARY_PATH=$(pwd)/chainer_ctc/ext/warp-ctc/build
+fi
+
+
 # test asr recipe
 (
     cd ./egs/mini_an4/asr1 || exit 1
--- a/ci/test_integration.sh
+++ b/ci/test_integration.sh
@@ -2,14 +2,6 @@
 
 set -euo pipefail
 
-if [ -z ${LD_LIBRARY_PATH+x} ]; then
-    LD_LIBRARY_PATH=$(pwd)/chainer_ctc/ext/warp-ctc/build
-else
-    LD_LIBRARY_PATH=${LD_LIBRARY_PATH}:$(pwd)/chainer_ctc/ext/warp-ctc/build
-fi
-
-export LD_LIBRARY_PATH
-
 # test asr recipe
 (
     cd ./egs/mini_an4/asr1 || exit 1
--- a/ci/test_integration.sh
+++ b/ci/test_integration.sh
@@ -1,5 +1,7 @@
 #!/usr/bin/env bash
 
+set -euo pipefail
+
 # test asr recipe
 (
     cd ./egs/mini_an4/asr1 || exit 1
--- a/ci/test_integration.sh
+++ b/ci/test_integration.sh
@@ -2,12 +2,13 @@
 
 set -euo pipefail
 
-if [ ! -z "${LD_LIBRARY_PATH}" ]; then
-    export LD_LIBRARY_PATH=${LD_LIBRARY_PATH}:$(pwd)/chainer_ctc/ext/warp-ctc/build
+if [ -z "${LD_LIBRARY_PATH}" ]; then
+    LD_LIBRARY_PATH=$(pwd)/chainer_ctc/ext/warp-ctc/build
 else
-    export LD_LIBRARY_PATH=$(pwd)/chainer_ctc/ext/warp-ctc/build
+    LD_LIBRARY_PATH=${LD_LIBRARY_PATH}:$(pwd)/chainer_ctc/ext/warp-ctc/build
 fi
 
+export LD_LIBRARY_PATH
 
 # test asr recipe
 (
--- a/ci/test_integration.sh
+++ b/ci/test_integration.sh
@@ -4,6 +4,7 @@
 (
     cd ./egs/mini_an4/asr1 || exit 1
     ./run.sh
+    ./run.sh --stage 3 --backend chainer
 )
 # test tts recipe
 (
--- a/ci/test_integration.sh
+++ b/ci/test_integration.sh
@@ -2,7 +2,7 @@
 
 set -euo pipefail
 
-if [ -z "${LD_LIBRARY_PATH}" ]; then
+if [ -z ${LD_LIBRARY_PATH+x} ]; then
     LD_LIBRARY_PATH=$(pwd)/chainer_ctc/ext/warp-ctc/build
 else
     LD_LIBRARY_PATH=${LD_LIBRARY_PATH}:$(pwd)/chainer_ctc/ext/warp-ctc/build
--- a/espnet/nets/chainer_backend/rnn/training.py
+++ b/espnet/nets/chainer_backend/rnn/training.py
@@ -17,9 +17,13 @@ from chainer.training.updaters.multiprocess_parallel_updater import gather_grads
 from chainer.training.updaters.multiprocess_parallel_updater import gather_params
 from chainer.training.updaters.multiprocess_parallel_updater import scatter_grads
 
-from cupy.cuda import nccl
-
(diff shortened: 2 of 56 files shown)


Fhrozen:
@sw005320 , btw I do not think that is possible to implement a test given that the error is generated for the CustomUpdater.

kan-bayashi:
Just add `run.sh --stage 3 --backend chainer` just after `run.sh` to check it.
https://github.com/espnet/espnet/blob/6759e7fcb8386b845a8a48ef779db3977f599a05/ci/test_integration.sh#L3-L7

sw005320:
Please check our new integration test https://github.com/espnet/espnet/blob/master/ci/test_integration.sh
This actually executes `run.sh`, and you can make sure our recipe is working without errors or not.

Fhrozen:
@sw005320. I just added the test.
I apologize for the inconveniences. Sometimes, I forgot about that new test. I  was just thinking about pytest.

sw005320:
No problem, but it seems that it is failed at https://travis-ci.org/espnet/espnet/jobs/572343376#L3162
I don't know why the CI was passed even it's failed. 
@kan-bayashi, @ShigekiKarita, do you have any ideas?

ci/test_integration.sh
```diff
@@ -4,6 +4,7 @@
 (
     cd ./egs/mini_an4/asr1 || exit 1
     ./run.sh
+    ./run.sh --stage 3 --backend chainer
```
ShigekiKarita: Add `set -euo pipefail` to the head of this filE. I forgot it.

ShigekiKarita:
https://5539-114054873-gh.circle-artifacts.com/0/home/circleci/repo/egs/mini_an4/asr1/exp/train_nodev_chainer_train/train.log

Circle CI saves train.log in artifact tab. It seems to still require cupy. Try uninstalling it when you fix that.

espnet/nets/chainer_backend/rnn/training.py
```diff
@@ -17,10 +17,14 @@
...
-from cupy.cuda import nccl
-
 import numpy as np
 
+try:
+    # Calls nccl when cupy is installed for CustomParallelUpdater (only for gpu support)
+    from cupy.cuda import nccl
+except Exception:
+    pass
+
```
kan-bayashi: Why don't you move above part to `CustomParallelUpdater.__init__()`?

espnet/nets/chainer_backend/transformer/training.py
```diff
@@ -19,10 +19,14 @@
...
 
 from chainer.training import extension
-from cupy.cuda import nccl
 
 import numpy as np
 
+try:
+    from cupy.cuda import nccl
+except Exception:
+    pass
```
kan-bayashi: ditto

espnet/nets/chainer_backend/rnn/training.py
kan-bayashi: Just call `from cupy.cuda import nccl` in `CustomParallelUpdater.__init__()`.

ci/test_integration.sh
```diff
@@ -1,9 +1,20 @@
...
 
+set -euo pipefail
+
+if [ -z ${LD_LIBRARY_PATH+x} ]; then
+    LD_LIBRARY_PATH=$(pwd)/chainer_ctc/ext/warp-ctc/build
+else
+    LD_LIBRARY_PATH=${LD_LIBRARY_PATH}:$(pwd)/chainer_ctc/ext/warp-ctc/build
+fi
+
+export LD_LIBRARY_PATH
```
kan-bayashi: It is better to add some comments here. (`path.sh`)
And you can use `shellcheck disable=SC...` to simplify

```bash
# NOTE: LD_LIBRARY_PATH is updated in `path.sh` but does not work. (Maybe?)
# shellcheck disable=SC<shell_check_code_number>
export LD_LIBRARY_PATH=${LD_LIBRARY_PATH}:$(pwd)/chainer_ctc/ext/warp-ctc/build

```

Fhrozen: The problem about D_LIBRARY_PATH is that it adds the chainer_ctc in tools (by default in pc env) but in ci it is install only in `pwd` (no tools) that is why raises error in Ubuntu tests.

I will disable SC for LD_LIBRARY_PATH

kan-bayashi: I see.
Another choice is as follows:
https://github.com/espnet/espnet/blob/6759e7fcb8386b845a8a48ef779db3977f599a05/ci/test_python.sh#L13

ShigekiKarita: I think fixing `install.sh` to put `chainer_ctc` in `tools/` is better because it is closer to our usage.

espnet/nets/chainer_backend/rnn/training.py
ShigekiKarita: I agree. Because nccl is only used for multi GPU training, it should be also optional for single GPU training.

espnet/nets/chainer_backend/rnn/training.py
```diff
@@ -73,6 +77,8 @@ def __init__(self, train_iter, optimizer, converter, device, accum_grad=1):
         self.forward_count = 0
         self.accum_grad = accum_grad
         self.start = True
+        # To solve #1091, it is required to set the variable inside this class.
+        self.device = device
```
ShigekiKarita: I do not think you need this because base class does
https://github.com/chainer/chainer/blob/a3c428a32905052e33fc6427f5bfd8ae4b782f79/chainer/training/updaters/standard_updater.py#L92

espnet/nets/chainer_backend/transformer/training.py
```diff
@@ -75,6 +79,7 @@ def __init__(self, train_iter, optimizer, converter, device, accum_grad=1):
         self.accum_grad = accum_grad
         self.forward_count = 0
         self.start = True
+        self.device = device
```
ShigekiKarita: ditto

espnet/nets/chainer_backend/rnn/training.py
kan-bayashi: Problem maybe be caused from this line.
https://github.com/espnet/espnet/blob/63b9cb451ab17e17506347f29cd0837d978647a8/espnet/nets/chainer_backend/rnn/training.py#L234
In base class, `self.device` is converted to `CpuDevice`.

https://github.com/chainer/chainer/blob/a3c428a32905052e33fc6427f5bfd8ae4b782f79/chainer/training/updaters/standard_updater.py#L71

Therefore, we should change if-condition like:
```python
xp = cuda.cupy if not isinstance(device, chainer.backends.CpuDevice) else np 
```

Fhrozen: The interesting part is that device is `chainer.backends.CpuDevice` in the first iter only, after that it becomes a int (-1 or the gpuid). I was also expecting to be a `CpuDevice` the whole training so it will be possible to extract the xp from the self device (So the if is no longer requires) but sadly it was not as expected. This is the only solution which I found without adding more lines. If you tried any other besides the mentioned let me know, but I could not find it.. T.T

kan-bayashi: Now I understood your point.
I doubt that following part:
https://github.com/espnet/espnet/blob/63b9cb451ab17e17506347f29cd0837d978647a8/espnet/asr/chainer_backend/asr.py#L286-L289
https://github.com/espnet/espnet/blob/63b9cb451ab17e17506347f29cd0837d978647a8/espnet/asr/asr_utils.py#L80
But easiest solution seems yours. 
Why don't your add `TODO(...):` comment in this time?
@ShigekiKarita how do you think?

ShigekiKarita:
Thank!

PR #1090, opened 2019-08-15
Update spm_encode

zh794390558:
UnicodeEncodeError: Ascii codec can't encode character u2581 in position 0: ordinal not in range(128)
when using SentencePiece

PR #1099, opened 2019-08-16
Add PYTHONIOENCODING=UTF-8 in path.sh

kan-bayashi:
This PR adds `PYTHONIOENCODING=UTF-8` in `path.sh` to avoid `UnicodeDecodeError` when `LC_ALL=C`.

Related to #1090


sw005320:
thanks!

PR #1101, opened 2019-08-16
Fix decoding for chainer transformer

Fhrozen:
This fix cover the decoding for chainer.transformer. 
Given than the encoder output has now 3 dimensions, need to set correctly the length of the output (to the axis 1)


sw005320:
LGTM. Thanks!

PR #1102, opened 2019-08-17
How2 speech translation recipe

hirofumi0810:
- `st1`
- `asr1`


egs/how2/asr1/conf/tuning/decode_pytorch_transformer.yaml
```diff
@@ -0,0 +1,7 @@
+batchsize: 0
```
sw005320: It seems that the decoding configs of RNN and transformer are very different. If you don't test either of them, you can remove that config.

hirofumi0810:
> Do you have some results? Can you add them as a `RESULTS.md`?

Let me do this in another PR. (due to some personal issues)

egs/how2/asr1/conf/tuning/decode_pytorch_transformer.yaml
hirofumi0810: I want to keep this as it is.

sw005320:
Thanks. Please let me know when it is ready to merge.

hirofumi0810:
This PR is ready to be merged.

egs/how2/st1/run.sh
```diff
@@ -0,0 +1,273 @@
...
+fi
+
+feat_tr_dir=${dumpdir}/${train_set}/delta${do_delta}; mkdir -p ${feat_tr_dir}
+feat_dt_dir=${dumpdir}/${train_dev}/delta${do_delta}; mkdir -p ${feat_dt_dir}
+if [ ${stage} -le 1 ] && [ ${stop_stage} -ge 1 ]; then
+    ### Task dependent. You have to design training and dev sets by yourself.
+    ### But you can utilize Kaldi recipes in most cases
+    echo "stage 1: Feature Generation"
+
+    # The fbank features is generated as pre-processing; 40-dimensional fbanks with 3-dimensional pitch on each frame
```
sw005320: Why don't you include the feature extraction process?
I want `run.sh` to be self-consistent.

hirofumi0810: How2 corpus provides pre-processed fbank features (40-dim fbank+3-dim pitch) only, which means we have no chance to access to the original audio files.

sw005320: Thanks!
Now I got it.
Could you add such comments more explicitly in `run.sh`?
Then, it's perfect.

sw005320:
Thanks a lot, @hirofumi0810!

PR #1103, opened 2019-08-18
[WIP] transfer learning

b-flo:
This PR adress @hirofumi0810 TODO for transfer learning in `asr/pytorch_backend/asr.py` : 

>better to simplify the E2E model interface by only allowing idim, odim, and args
the pretrained ASR and MT model arguments should be removed here and we should implement additional method to attach these methods

The proposed methods would be used to replace the following in `asr/pytorch_backend/asr.py`: 

```
asr_model, mt_model = None, None

if args.asr_model:
    asr_model, _ = load_trained_model(args.asr_model)
    assert isinstance(asr_model, ASRInterface)

if args.mt_model:
    mt_model, _ = load_trained_model(args.mt_model)
    assert isinstance(mt_model, ASRInterface)

 model_class = dynamic_import(args.model_module)

if asr_model is None and mt_model is None:
    model = model_class(idim, odim, args)
else:
    model = model_class(idim, odim, args, asr_model=asr_model, mt_model=mt_model)
assert isinstance(model, ASRInterface)

if args.asr_model:
    del asr_model

if args.asr_model:
    del asr_model
```

By:

```
if args.enc_init or args.dec_init:
    model = load_trained_modules(idim, odim, args)
else:
    model_class = dynamic_import(args.model_module)
    model = model_class(idim, odim, args)
```

And remove `load_trained_model` method in same file + pre-training code in `nets/pytorch_backend/e2e_asr.py`.

It also extend support by allowing user to specify the modules to transfer from the trained model, for both encoder and decoder. For the decoder part, a pre-trained ESPNET LM is now allowed to support decoder pre-initialization in vanilla transducer.

Before making any changes in  `asr/asr.py` and `nets/e2e_asr.py`, and remove current pre-init arguments (`asr_model` and `mt_model`), I have some questions : 

 - Should the specified trained model(s) architecture overwrite the architecture specified in training config?
 - For now I isolated the methods in `asr/pytorch_backend/asr_init.py`. I don't think I should add them in `asr/pytorch_backend/asr.py` for readibility nor in `asr/asr_utils.py` as it only applies to pytorch. Where should I put them?
 - Does current experiments need additional features such as freezing modules or multi-models support?

### DIFF
--- a/espnet/asr/pytorch_backend/asr_init.py
+++ b/espnet/asr/pytorch_backend/asr_init.py
@@ -1,0 +1,202 @@
+import logging
+import os
+import torch
+
+from collections import OrderedDict
+
+from espnet.asr.asr_utils import get_model_conf
+from espnet.asr.asr_utils import torch_load
+
+from espnet.nets.asr_interface import ASRInterface
+from espnet.nets.mt_interface import MTInterface
+
+from espnet.utils.dynamic_import import dynamic_import
+
+
+def transfer_verification(model_state_dict, partial_state_dict, modules):
+    """Verify tuples (key, shape) for input model modules match specified modules.
+
+    Args:
+        model_state_dict (odict): the initial model state_dict
+        partial_state_dict (odict): the trained model state_dict
+        modules (list): specified module list for transfer
+
+    Return:
+        (boolean): allow transfer
+    """
+
+    modules_model = []
+    partial_modules = []
+
+    for key_p, value_p in partial_state_dict.items():
+        if any(key_p.startswith(m) for m in modules):
+            partial_modules += [(key_p, value_p.shape)]
+
+    for key_m, value_m in model_state_dict.items():
+        if any(key_m.startswith(m) for m in modules):
+            modules_model += [(key_m, value_m.shape)]
+
+    len_match = (len(modules_model) == len(partial_modules))
+    module_match = (sorted([x for x in modules_model]) ==
+                    sorted([x for x in partial_modules]))
+
+    return len_match and module_match
+
+
+def get_trained_asr_mt_state_dict(model_state_dict, modules):
+    """Create state_dict with specified modules matching input model modules.
+
+    Args:
+        model_state_dict (odict): trained model state_dict
+        modules (list): specified module list for transfer
+
+    Return:
+        new_state_dict (odict): the updated state_dict
+    """
+
+    new_state_dict = OrderedDict()
+
+    for key, value in model_state_dict.items():
+        if any(key.startswith(m) for m in modules):
+            new_state_dict[key] = value
+
+    return new_state_dict
+
+
+def get_trained_lm_state_dict(model_state_dict, modules):
+    """Create compatible ASR state_dict from model_state_dict (LM).
+
+    Modify specified modules to match ASR decoder modules.
+
+    Args:
+        model_state_dict (odict): trained model state_dict
+        modules (list): specified module list for transfer
+
+    Return:
+        new_state_dict (odict): the updated state_dict
+        new_mods (list): the updated module list
+    """
+
+    new_state_dict = OrderedDict()
+    new_modules = []
+
+    for key, value in list(model_state_dict.items()):
+        if key == "predictor.embed.weight" \
+           and "predictor.embed." in modules:
+            new_key = "dec.embed.weight"
+            new_state_dict[new_key] = value
+            new_modules += [new_key]
+        elif "predictor.rnn." in key \
+             and "predictor.rnn." in modules:
+            new_key = "dec.decoder." + key.split("predictor.rnn.", 1)[1]
+            new_state_dict[new_key] = value
+            new_modules += [new_key]
+
+    return new_state_dict, new_modules
+
+
+def filter_modules(model_state_dict, modules):
+    """Filter non-matched modules in module_state_dict 
+
+    Args:
+        model_state_dict (odict): trained model state_dict
+        modules (list): specified module list for transfer
+
+    Return:
+        new_mods (list): the update module list
+    """
+
+    new_mods = []
+    incorrect_mods = []
+
+    mods_model = list(model_state_dict.keys())
+    for mod in modules:
+        if any(key.startswith(mod) for key in mods_model):
+            new_mods += [mod]
+        else:
+            incorrect_mods += [mod]
+
+    if incorrect_mods:
+        logging.info("Some specified module(s) in pre-trained model don\'t "
+                     "match (or partially match) available modules."
+                     " Ignoring the following modules: %s", incorrect_mods)
+
+    return new_mods
+
+
+def load_trained_model(model_path):
+    """Load the trained model.
+
+    Args:
+        model_path (str): Path to model.***.best
+
+    Return:
+        model.state_dict() (odict): the loaded model state_dict
+        (str): Type of model. Either ASR/MT or LM.
+    """
+
+    conf_path = os.path.join(os.path.dirname(model_path), 'model.json')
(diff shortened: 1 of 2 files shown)


sw005320:
Thanks, @b-flo!
Is it possible to make a simple unit test for this initialization with a pre-trained model?

b-flo:
Sure! I'll add unit test after review and receiving answer to the first and second question. Right now the code is not covered because the methods are not called in `asr/pytorch_backend/asr.py` and `nets/pytorch_backend/e2e_asr.py`

b-flo:
Ultimately I changed the transfer learning implementations in `asr/pytorch_backend/asr.py` and  `nets/pytorch_backend/e2e_asr.py`, and added tests (to extend) in `test/test_asr_init.py`. It may be better for review.

sw005320:
@hirofumi0810, could you review this PR?
This is a very nice extension for you as well.

kan-bayashi:
@hirofumi0810 ping

test/test_asr_init.py
```diff
@@ -0,0 +1,166 @@
...
+        model.recognize(in_data, args, args.char_list)
+
+    if not os.path.exists(".pytest_cache"):
+        os.makedirs(".pytest_cache")
+    utils = importlib.import_module('espnet.asr.asr_utils')
+
+    tmppath = tempfile.mktemp()
+    utils.torch_save(tmppath, model)
+
+    if enc_init:
```
hirofumi0810: if enc_init is not None:

test/test_asr_init.py
```diff
@@ -0,0 +1,166 @@
...
+        enc_init = tmppath
+    if dec_init:
```
hirofumi0810: if dec_init is not None:

espnet/asr/pytorch_backend/asr.py
```diff
@@ -309,34 +285,14 @@ def train(args):
...
-        assert isinstance(mt_model, MTInterface)
-
-    # specify model architecture
-    model_class = dynamic_import(args.model_module)
-    # TODO(hirofumi0810) better to simplify the E2E model interface by only allowing idim, odim, and args
-    # the pre-trained ASR and MT model arguments should be removed here and we should implement an additional method
-    # to attach these models
-    if asr_model is None and mt_model is None:
-        model = model_class(idim, odim, args)
+    if args.enc_init or args.dec_init:
```
hirofumi0810: if args.enc_init is not None or args.dec_init is not None:

espnet/asr/pytorch_backend/asr_init.py
```diff
@@ -0,0 +1,234 @@
...
+    for key_p, value_p in partial_state_dict.items():
+        if any(key_p.startswith(m) for m in modules):
+            partial_modules += [(key_p, value_p.shape)]
+
+    for key_m, value_m in model_state_dict.items():
+        if any(key_m.startswith(m) for m in modules):
+            modules_model += [(key_m, value_m.shape)]
+
+    len_match = (len(modules_model) == len(partial_modules))
+    module_match = (sorted([x for x in modules_model]) ==
```
hirofumi0810: Can you specify a key with lambda function?

espnet/asr/pytorch_backend/asr_init.py
```diff
@@ -0,0 +1,234 @@
...
+                    sorted([x for x in partial_modules]))
+
+    return len_match and module_match
```
hirofumi0810: You seem to skip shape matching. Is it OK? I think I added such parts...

b-flo: I don't skip shape matching, `modules_model` (main) and `partial_modules` (trained)  contains (name, shape) for matched in `modules`. Thus this test will return false if there is a shape mismatch!

espnet/asr/pytorch_backend/asr_init.py
b-flo: Ha yes, it would be more elegant!

b-flo:
@hirofumi0810 I addressed your reviews, thanks!

As it seems you'll be using transfer learning, do you think I should add in this PR simple features such as freezing layers/modules? It can be done by modifying `requires_grad` flag from specified parameters and filtering the parameters in optimizer (to avoid updating weights even if grad is 0). 
Also, I was wondering if we should allow the pre-trained model config to overwrite the config for specified modules?

hirofumi0810:
> @hirofumi0810 I addressed your reviews, thanks!
> 
> As it seems you'll be using transfer learning, do you think I should add in this PR simple features such as freezing layers/modules? It can be done by modifying `requires_grad` flag from specified parameters and filtering the parameters in optimizer (to avoid some optimizers to update the weights even if grad is 0).
> Also, I was wondering if we should allow the pre-trained model config to overwrite the config for specified modules?

I think such features are not necessary for this PR. (I'm not using parameter freezing)
With regard to overwriting config files, do you need it for RNN-T?

b-flo:
No I don't I was just wondering! It's just that it seems more user-friendly as the user doesn't need to set manually the config file to match pre-trained model config.

sw005320:
I prefer to have a smaller PR. We can always do it later.

sw005320:
I like the current design of splitting encoder/decoder as it would be very useful for RNN-T and ST. But at the same time, many people want to have more straightforward transfer learning for the whole encoder/decoder (e.g.,  we don't change any ASR model architecture and only re-train the parameter values without separating the encoder/decoder for the adaptation or teacher-student learning scenario). Can we easily do it with this approach? I personally feel that we could also have an option to do such things.

hirofumi0810:
@sw005320 We can do that easily because I moved ST from ASR modules in #1139.

sw005320:
OK. Is it ready for merge then? @b-flo and @hirofumi0810 ?

b-flo:
If it's okay for @hirofumi0810, the PR can be merged!
Btw if you want to transfer encoder-decoder from the same model, you can specify the same model for `--enc-init` or/and `--dec-init` and keep default `--enc-init-mods` and `dec-init-mods`. But at the same time, depending on the task, even when you want to transfer a complete pre-trained model you want to remove output layer and specific linear layers!

b-flo:
As it's related, I'm posting this here for the moment. Early results for transducer with pre-initialization : 

````
db: voxforge-dt_it               CER   |  WER
------------------------------------------------
CTC:                             17.1  |  64.4
Att:                             XX.X  |  XX.X
------------------------------------------------
Trans (no init):                 16.1  |  47.4
Trans (encoder pre-init):        11.7  |  42.2
------------------------------------------------
Trans-att (no init):             14.4  |  46.6
Trans-att (encoder pre-init):    11.7  |  43.3
Trans-att (enc+dec pre-init):    11.1  |  40.9
````

CTC model was trained with the same encoder architecture specified in `egs/voxforge/train_transducer.yaml` (minus dropout).
Outside better overall performances, the observations are in range with the ones I reported in #1129 for my initial experiments : CTC encoder pre-initialization seems more beneficial to vanilla transducer than transducer-attention. However adding decoder pre-initialization with attention-only model further improves transducer-att.
After investigation on why I had degraded performances by using LSTMCell in vanilla transducer, I'll change back and test decoder pre-init with LM to compare.

esalesky:
It looks like these updated flags (`enc_init`, `enc-init-mods`, `dec_init`, `dec-init-mods`) for asr/st pretraining didn't make it into mt_train.py, but they're used in the train.yaml files for at least one of the mt recipes (eg must-c). 

Has anyone updated mt_train.py locally (@hirofumi0810 @b-flo ?) 
If not, I can do so and push it.

PR #1105, opened 2019-08-19
 Allow to extend the length of positional encoding at training and inference

ShigekiKarita:
This is a rework of #1060 (by @takenori-y) not only for inference but also for inference. I found my PR including TransformerLM requires a much longer cache for PE. So I made this both for training and inference.

### DIFF
--- a/espnet/nets/pytorch_backend/transformer/embedding.py
+++ b/espnet/nets/pytorch_backend/transformer/embedding.py
@@ -63,7 +63,7 @@ class ScaledPositionalEncoding(PositionalEncoding):
     def __init__(self, d_model, dropout_rate, max_len=5000):
         """Initialize class.
 
-        :param int self.d_model: embedding dim
+        :param int d_model: embedding dim
         :param float dropout_rate: dropout rate
         :param int max_len: maximum input length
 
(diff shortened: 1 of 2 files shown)


espnet/nets/pytorch_backend/transformer/embedding.py
```diff
@@ -1,54 +1,90 @@
...
+    """Scaled positional encoding module.
 
     See also: Sec. 3.2  https://arxiv.org/pdf/1809.08895.pdf
+
     """
 
     def __init__(self, d_model, dropout_rate, max_len=5000):
+        """Initialize class.
+
+        :param int self.d_model: embedding dim
```
takenori-y:
```suggestion
        :param int d_model: embedding dim
```

espnet/nets/pytorch_backend/transformer/embedding.py
```diff
@@ -1,54 +1,90 @@
...
         self.dropout = torch.nn.Dropout(p=dropout_rate)
-        # Compute the positional encodings once in log space.
-        pe = torch.zeros(max_len, d_model)
-        position = torch.arange(0, max_len, dtype=torch.float32).unsqueeze(1)
-        div_term = torch.exp(torch.arange(0, d_model, 2, dtype=torch.float32) *
-                             -(math.log(10000.0) / d_model))
+        self.pe = None
+        self.reset_(torch.tensor(0.0).expand(1, max_len))
+
+    def reset_(self, x):
```
takenori-y: `extend_positional_encoding` or `extend_pe` may be better than `reset` as the function name since `reset(x)` may not be understandable.

sw005320:
Thanks a lot!

PR #1109, opened 2019-08-20
Merge master to v.0.6.0

ShigekiKarita:


ShigekiKarita:
@Fhrozen I have modified ./espnet/nets/chainer_backend/transformer/training.py because it was conflicted. Could you check that?

PR #1110, opened 2019-08-20
Small changes in chainer backend

Fhrozen:
- Remove Linear_tensor and replace with n_batch_axes
- Replace F.reshape for [var].reshape
- Replace F.transpose for [var].transpose
- Remove fixed files from blacklist (docstring)


ShigekiKarita:
LGTM. By the way, why do you replace function calls to method calls? Are they faster?

Fhrozen:
There are slightly faster, and to keep a similarity with pytorch

PR #1113, opened 2019-08-20
Fix: wrong variable in greedy decode

b-flo:
This PR fix a variable I forgot to change in both `recognize` methods from `nets/rnn/decoders_transducer.py`.
Previously, for vanilla transducer, embedding shape was **(odim, dunits)** and decoder shape for first layer was **(dunits, dunits)**. I changed for **(odim, dec_embed_dim)** and **(dec_embed_dim, dunits)** but didn't change initialization with **dunits** in greedy decode.

I also added a test covering the case were dunits != dec_embed_dim when beam_size=1 (greedy decode)


sw005320:
Thanks a lot!

Issue #1116, opened 2019-08-21
missing test_set_iwslt2019.en directory in how2


sw005320:
@hirofumi0810, `how2/asr1/run.sh` is failed due to the missing test_set_iwslt2019.en directory, e.g., 
```
utils/split_scp.pl: Error opening input scp file data/test_set_iwslt2019.en/feats.scp: No such file or directory
```
Could you fix it?

hirofumi0810:
Is test_set_iwslt2019 installed in `data/local/`?

sw005320:
I think so. I found data/test_set_iwslt2019 but not data/test_set_iwslt2019.en

sw005320:
Is it ok to use test_set_iwslt2019 instead of test_set_iwslt2019.en?

PR #1117, opened 2019-08-21
Fix bug in how2 asr1

hirofumi0810:
Bug fix for #1116


sw005320:
thanks for your quick fix!

PR #1119, opened 2019-08-21
Add --train-dtype option for float16/float32/float64 precision training in pytorch ASR and LM

ShigekiKarita:
This is similar to `--dtype floatXX` option in `asr_recog.py --api v2`. Because `--dtype` is already used for RNN ASR model, I named this option as `--train-dtype`.

### DIFF
--- a/espnet/bin/asr_train.py
+++ b/espnet/bin/asr_train.py
@@ -252,6 +252,10 @@ def get_parser(parser=None):
 def main(cmd_args):
     parser = get_parser()
     args, _ = parser.parse_known_args(cmd_args)
+    if args.backend == "chainer" and args.train_dtype != "float32":
+        raise NotImplementedError(
+            f"chainer backend does not support --train-dtype {args.train_dtype}."
+            "Use --dtype float32.")
 
     from espnet.utils.dynamic_import import dynamic_import
     if args.model_module is None:
@@ -267,10 +271,6 @@ def main(cmd_args):
         args.backend = 'chainer'
     if 'pytorch_backend' in args.model_module:
         args.backend = 'pytorch'
-    if args.backend == "chainer" and args.train_dtype != "float32":
-        raise NotImplementedError(
-            f"chainer backend does not support --train-dtype {args.train_dtype}."
-            "Use --dtype float32.")
 
     # logging info
     if args.verbose > 0:
--- a/espnet/bin/asr_train.py
+++ b/espnet/bin/asr_train.py
@@ -36,7 +36,7 @@ def get_parser(parser=None):
     parser.add_argument('--train-dtype', default="float32",
                         choices=["float16", "float32", "float64", "O0", "O1", "O2", "O3"],
                         help='Data type for training (only pytorch backend). '
-                        'O0,O1,.. flags require apex. apex automatically selects float precision.')
+                        'O0,O1,.. flags require apex. See https://nvidia.github.io/apex/amp.html#opt-levels')
     parser.add_argument('--backend', default='chainer', type=str,
                         choices=['chainer', 'pytorch'],
                         help='Backend library')
--- a/espnet/bin/asr_train.py
+++ b/espnet/bin/asr_train.py
@@ -18,11 +18,12 @@ from espnet.utils.training.batchfy import BATCH_COUNT_CHOICES
 
 
 # NOTE: you need this func to generate our sphinx doc
-def get_parser():
-    parser = configargparse.ArgumentParser(
-        description="Train an automatic speech recognition (ASR) model on one CPU, one or multiple GPUs",
-        config_file_parser_class=configargparse.YAMLConfigFileParser,
-        formatter_class=configargparse.ArgumentDefaultsHelpFormatter)
+def get_parser(parser=None):
+    if parser is None:
+        parser = configargparse.ArgumentParser(
+            description="Train an automatic speech recognition (ASR) model on one CPU, one or multiple GPUs",
+            config_file_parser_class=configargparse.YAMLConfigFileParser,
+            formatter_class=configargparse.ArgumentDefaultsHelpFormatter)
     # general configuration
     parser.add('--config', is_config_file=True, help='config file path')
     parser.add('--config2', is_config_file=True,
@@ -32,6 +33,8 @@ def get_parser():
 
     parser.add_argument('--ngpu', default=None, type=int,
                         help='Number of GPUs. If not given, use all visible devices')
+    parser.add_argument('--train-dtype', default="float32", choices=["float16", "float32", "float64"],
+                        help='Data type for training (only pytorch backend).')
     parser.add_argument('--backend', default='chainer', type=str,
                         choices=['chainer', 'pytorch'],
                         help='Backend library')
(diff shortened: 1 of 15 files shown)


sw005320:
Thanks, @ShigekiKarita!
I'm curious what happened when there are mixed precisions?

ShigekiKarita:
Do you mean "mixed precision" as different precisions in training and decoding? Essentially, it has no problem. As I tested, it seems to results almost same.
```
# asr 64, lm 64 -> decode 32
$ ./run.sh --stage 3 --lm-config $(change_yaml.py ./conf/lm.yaml -a train-dtype=float64) --train-config $(change_yaml.py ./conf/train.yaml -a train-dtype=float64) --decode-config $(change_yaml.py ./conf/decode.yaml -a api=v2 -a dtype=float32)
# asr 32, lm 64 -> decode 32
$ ./run.sh --stage 3 --lm-config $(change_yaml.py ./conf/lm.yaml -a train-dtype=float32) --train-config $(change_yaml.py ./conf/train.yaml -a train-dtype=float64) --decode-config $(change_yaml.py ./conf/decode.yaml -a api=v2 -a dtype=float32)
# asr 32, lm 32 -> decode 64
./run.sh --stage 3 --lm-config $(change_yaml.py ./conf/lm.yaml -a train-dtype=float32) --train-config $(change_yaml.py ./conf/train.yaml -a train-dtype=float32) --decode-config $(change_yaml.py ./conf/decode.yaml -a api=v2 -a dtype=float64)
```

sw005320:
Yes, and I'm curious how it works in training. I found that Warp CTC always use 32 while others are configurable. Then, there would be different precisions in training for example. I guess either of them is well casted and there would be no problem.

ShigekiKarita:
Hmm, I think it was correctly wraped warpctc. However, even without ctc, float16 training seems to result loss=nan easily (maybe due to its shorter numerical range?). Anyway float64 seems to work as same as float32.

ShigekiKarita:
Example memory usage in default AN4
- float16: 1142MiB
- float32: 1752MiB
- float64: 2899MiB

The loss values of float32 and float64 are almost same.
```
$ grep validation/main/loss ./exp/train_nodev_pytorch_train_mtlalpha1.0_train-dtypefloat64/results/log
        "validation/main/loss": 51.24402856826782,
        "validation/main/loss": 48.143794298172,
        "validation/main/loss": 44.937756299972534,
$ grep validation/main/loss ./exp/train_nodev_pytorch_train_mtlalpha1.0/results/log      
        "validation/main/loss": 51.244147539138794,
        "validation/main/loss": 48.14349985122681,
        "validation/main/loss": 44.91559827327728,
```

ShigekiKarita:
For LM training in AN4, float 16/32/64 seems to be same.

- float16: test perplexity: 14.264976096225666, time: 5 seconds (!!)
- float32: test-perplexity: 14.215787050249471, time: 20 seconds
- float64: test-perplexity: 14.215786138140388, time: 36 seconds

ShigekiKarita:
To fix loss=nan problem at float16 in ASR, I introduced [apex](https://github.com/NVIDIA/apex). 

AN4 benchmark (train_mtlalpha1.0)
- almost float16 (--train-dtype O2), memory: 1197MiB, 9.0 iter/sec
- float16/32 mixed (--train-dtype O1), memory: 2129MiB, 5.3 iter/sec
- float32, memory: 1719MiB, 8.3 iter/sec

espnet/bin/asr_train.py
```diff
@@ -32,6 +33,10 @@ def get_parser():
 
     parser.add_argument('--ngpu', default=None, type=int,
                         help='Number of GPUs. If not given, use all visible devices')
+    parser.add_argument('--train-dtype', default="float32",
+                        choices=["float16", "float32", "float64", "O0", "O1", "O2", "O3"],
+                        help='Data type for training (only pytorch backend). '
+                        'O0,O1,.. flags require apex. apex automatically selects float precision.')
```
kan-bayashi: I think it is better to add explanation of each option or add url of apex documentation.
Another option is to rename 01 to something like float16-float32-mixed.
How do you think?

espnet/nets/pytorch_backend/e2e_asr.py
```diff
@@ -135,6 +135,8 @@ def __init__(self, idim, odim, args, asr_model=None, mt_model=None):
         assert 0.0 <= self.mtlalpha <= 1.0, "mtlalpha should be [0.0, 1.0]"
         self.etype = args.etype
         self.verbose = args.verbose
+        # NOTE: for self.build method
+        args.char_list = getattr(args, "char_list", None)
         self.char_list = args.char_list
```
kan-bayashi: self.char_list = getattr(...) is enough?

espnet/nets/asr_interface.py
```diff
@@ -1,15 +1,38 @@
...
+            odim (int): The number of output vocab.
+
+        Returns:
+            ASRinterface: A new instance of ASRInterface.
+
+        """
+        args = argparse.Namespace(**kwargs)
+        args = fill_missing_args(
+            args, get_parser,
+            dummy_input="--outdir /tmp --dict /tmp/dummy.txt".split())
```
kan-bayashi: It seems that this line will overwrite **kwags.

espnet/utils/fill_missing_args.py
```diff
@@ -7,7 +7,7 @@
 import logging
 
 
-def fill_missing_args(args, add_arguments):
+def fill_missing_args(args, add_arguments, dummy_input=None):
```
kan-bayashi: dummy_input does not make sense.
cmd_line_args? not so good :(
Or if you use this method for only build method, revert this change and then add some operations in build method.

espnet/bin/asr_train.py
ShigekiKarita: I will add url to apex.amp doc. https://nvidia.github.io/apex/amp.html#opt-levels

espnet/utils/fill_missing_args.py
ShigekiKarita: pre_parse_str ?

espnet/nets/pytorch_backend/e2e_asr.py
ShigekiKarita: no, because some classes in this `__init__` also requires the `args` to have `char_list`

espnet/nets/asr_interface.py
ShigekiKarita: no. I added tests
https://github.com/ShigekiKarita/espnet/blob/train-dtype/test/test_asr_interface.py#L10-L13

espnet/utils/fill_missing_args.py
kan-bayashi: Now I understand why this option is needed.
Why don’t you remove required=True instead pf this option?

ShigekiKarita:
I agree. That is simplest!

2019年8月22日(木) 21:20 Tomoki Hayashi <notifications@github.com>:

ShigekiKarita:
I believe it is done

sw005320:
Thanks, @ShigekiKarita!!

PR #1129, opened 2019-08-23
Transducer v1.1

b-flo:
Hi,

I’m making this PR to fix some things for transducer :

- add missing argument `score_norm_transducer` for when report_cer=True or report_wer=True

- add argument relative to fix in `tests/test_e2e_asr_transducer.py`.

- raise error for temporary unavailable batch decoding for transducer (@ShigekiKarita  : I did as you mentioned in previous PR and put the assertion in `asr/asr.py`, is that okay like this?)

- modify voxforge example (initial config was used with pre-trained CTC encoder) to have performances closer to other E2E models. It's probably not the best config though.

Example of results for dt_it:

```
Trans (CER)	| 1082   79133 | 86.9    5.7    7.4    2.9   16.1   97.9 |
Trans (WER)	| 1082  13235 | 57.7   34.8    7.5    5.1   47.4   97.9 |
----------------------------------------------------------------------------------------
Trans-Att (CER)	| 1082   79133 | 89.5    5.2    5.3    2.8   13.3   97.5 |
Trans-Att (WER)	| 1082  13235 | 60.1   33.2    6.7    4.8   44.6   97.5 |
```

- In `rnn/decoders_transducer.py`: switch LSTMCell to LSTM for vanilla transducer as performances seems a bit better (I should investigate) + some rework/clean-up to have more consistent implementation between transducer and transducer-attention.
Note that first change will render unusable the decoder pre-initialization with LM option I added in #1103 but I think it’s not a problem right now.

Also, I’ll start working on performance improvment as I observed a high number of deletions on some setup compared to my initial experiments. However in most of these experiments I was using a CTC model to pre-init encoder, so I will maybe wait for #1103 to be merged to further investigate.
I should also adress memory and speed improvment.

For the remainining items (batch decoding and basic features support), I will add them asap when I have some time.

### DIFF
--- a/egs/voxforge/asr1/conf/tuning/train_transducer.yaml
+++ b/egs/voxforge/asr1/conf/tuning/train_transducer.yaml
@@ -13,26 +13,32 @@ patience: 3
 
 # network architecture
 ## encoder related
-etype: vggblstmp
+etype: vggblstm
 elayers: 4
-eunits: 320
-eprojs: 320
+eunits: 256
+eprojs: 256
+subsample: "1_1_1_1_1"
+dropout-rate: 0.5
 ## decoder related
 dtype: lstm
 dlayers: 1
-dunits: 300
-dec-embed-dim: 320
+dec-embed-dim: 256
+dunits: 256
+dropout-rate-decoder: 0.2
+dropout-rate-embed-decoder: 0.2
 ## attention related
 atype: location
-adim: 320
+adim: 256
+aheads: 1
+awin: 5
 aconv-chans: 10
 aconv-filts: 100
 ## joint network related
-joint-dim: 320
+joint-dim: 256
 
 # transducer related
 ## mtlalpha should be set to 1.0 (CTC) to use transducer
 mtlalpha: 1.0
-## transducer mode
-rnnt-mode: 'rnnt-att'
+## switch to 'rnnt-att' to use transducer with attention
+rnnt-mode: 'rnnt'
 model-module: "espnet.nets.pytorch_backend.e2e_asr_transducer:E2E"
--- a/espnet/asr/pytorch_backend/asr.py
+++ b/espnet/asr/pytorch_backend/asr.py
@@ -694,11 +694,6 @@ def recog(args):
                 new_js[name] = add_results_to_json(js[name], nbest_hyps, train_args.char_list)
 
     else:
-        if hasattr(train_args, "model_module"):
-            if 'transducer' in train_args.model_module:
-                logging.error("Batch decoding for transducer is not supported yet.")
-                raise NotImplementedError
-
         def grouper(n, iterable, fillvalue=None):
             kargs = [iter(iterable)] * n
             return zip_longest(*kargs, fillvalue=fillvalue)
--- a/espnet/asr/pytorch_backend/asr.py
+++ b/espnet/asr/pytorch_backend/asr.py
@@ -694,6 +694,11 @@ def recog(args):
                 new_js[name] = add_results_to_json(js[name], nbest_hyps, train_args.char_list)
 
     else:
+        if hasattr(train_args, "model_module"):
+            if 'transducer' in train_args.model_module:
+                logging.error("Batch decoding for transducer is not supported yet.")
+                raise NotImplementedError
+
         def grouper(n, iterable, fillvalue=None):
             kargs = [iter(iterable)] * n
             return zip_longest(*kargs, fillvalue=fillvalue)
--- a/espnet/nets/asr_interface.py
+++ b/espnet/nets/asr_interface.py
@@ -63,6 +63,19 @@ class ASRInterface:
         """
         raise NotImplementedError("recognize method is not implemented")
 
+    def recognize_batch(self, x, recog_args, char_list=None, rnnlm=None):
+        """Beam search implementation for batch
+
+        Args:
+            h (torch.Tensor): encoder hidden state sequences (B, Tmax, Henc)
+            recog_args (Namespace): argument Namespace containing options
+            rnnlm (torch.nn.Module): language module
+
+        Returns:
+            nbest_hyps (list of dicts): n-best decoding results
+        """
+        raise NotImplementedError(f"Batch decoding for transducer is not supported yet.")
+
     def calculate_all_attentions(self, xs, ilens, ys):
         """Caluculate attention.
 
--- a/espnet/nets/asr_interface.py
+++ b/espnet/nets/asr_interface.py
@@ -73,7 +73,7 @@ class ASRInterface:
         :return: N-best decoding results
         :rtype: list
         """
-        raise NotImplementedError(f"Batch decoding for transducer is not supported yet.")
+        raise NotImplementedError("Batch decoding is not supported yet.")
 
     def calculate_all_attentions(self, xs, ilens, ys):
         """Caluculate attention.
--- a/espnet/nets/asr_interface.py
+++ b/espnet/nets/asr_interface.py
@@ -64,15 +64,14 @@ class ASRInterface:
         raise NotImplementedError("recognize method is not implemented")
 
     def recognize_batch(self, x, recog_args, char_list=None, rnnlm=None):
-        """Beam search implementation for batch
+        """Beam search implementation for batch.
 
-        Args:
-            h (torch.Tensor): encoder hidden state sequences (B, Tmax, Henc)
-            recog_args (Namespace): argument Namespace containing options
-            rnnlm (torch.nn.Module): language module
-
-        Returns:
-            nbest_hyps (list of dicts): n-best decoding results
+        :param torch.Tensor x: encoder hidden state sequences (B, Tmax, Henc)
+        :param namespace recog_args: argument namespace containing options
+        :param list char_list: list of characters
+        :param torch.nn.Module rnnlm: language model module
+        :return: N-best decoding results
+        :rtype: list
         """
         raise NotImplementedError(f"Batch decoding for transducer is not supported yet.")
 
--- a/espnet/nets/pytorch_backend/rnn/decoders_transducer.py
+++ b/espnet/nets/pytorch_backend/rnn/decoders_transducer.py
@@ -356,7 +356,7 @@ class DecoderRNNTAtt(torch.nn.Module):
         """Initialize decoder states.
 
         Args:
-            ey (torch.Tensor): batch of input features (B, Lmax, X_dim)
+            ey (torch.Tensor): batch of input features (B, (Emb_dim + Eprojs))
 
         Return:
             z_list : list of L zero-init hidden state (B, Hdec)
--- a/espnet/nets/pytorch_backend/rnn/decoders_transducer.py
+++ b/espnet/nets/pytorch_backend/rnn/decoders_transducer.py
@@ -74,6 +74,15 @@ class DecoderRNNT(torch.nn.Module):
         self.blank = blank
 
     def zero_state(self, ey):
+        """Initialize decoder states
+
+        Args:
+            ey (torch.Tensor): batch of input features (B, Lmax, Emb_dim)
+
+        Returns:
+            (list): list of L zero-init hidden and cell state (1, B, Hdec)
+        """
+
         z_list = [ey.new_zeros(1, ey.size(0), self.dunits)]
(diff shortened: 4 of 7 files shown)


espnet/asr/pytorch_backend/asr.py
```diff
@@ -694,6 +694,11 @@ def recog(args):
                 new_js[name] = add_results_to_json(js[name], nbest_hyps, train_args.char_list)
 
     else:
+        if hasattr(train_args, "model_module"):
+            if 'transducer' in train_args.model_module:
```
ShigekiKarita: This `model_module` is str and it is not type. So this condition is always False. Instead, you can simply put this error in espnet.nets.asr_interface.ASRInterface:
```python
class ASRInterface:
    def recognize_batch(self, ...):
        """Recognize a speech batch."""
        raise NotImplementedError(f"Batch decoding for {type(self).__name__} is not supported yet.")
```

espnet/nets/pytorch_backend/rnn/decoders_transducer.py
```diff
@@ -69,23 +73,35 @@ def __init__(self, eprojs, odim, dtype, dlayers, dunits, blank,
...
+        c_list = [ey.new_zeros(1, ey.size(0), self.dunits)]
+
+        for _ in six.moves.range(1, self.dlayers):
+            z_list.append(ey.new_zeros(1, ey.size(0), self.dunits))
+            c_list.append(ey.new_zeros(1, ey.size(0), self.dunits))
 
-    def rnn_forward(self, ey, z_list, c_list, z_prev, c_prev):
+        return (z_list, c_list)
+
+    def rnn_forward(self, ey, dstate):
```
ShigekiKarita: I want docstrings for its args and return values (e.g., type, comment, and shape) because it is getting a little bit complex.

espnet/asr/pytorch_backend/asr.py
b-flo: Got it. Thanks

espnet/nets/asr_interface.py
```diff
@@ -63,6 +63,19 @@ def recognize(self, x, recog_args, char_list=None, rnnlm=None):
...
+
+        Args:
+            h (torch.Tensor): encoder hidden state sequences (B, Tmax, Henc)
+            recog_args (Namespace): argument Namespace containing options
+            rnnlm (torch.nn.Module): language module
+
+        Returns:
+            nbest_hyps (list of dicts): n-best decoding results
+        """
+        raise NotImplementedError(f"Batch decoding for transducer is not supported yet.")
```
ShigekiKarita: Because this interface is not only for transducer, the message should be more general
```suggestion
        raise NotImplementedError(f"Batch decoding is not supported yet.")
```

espnet/nets/pytorch_backend/rnn/decoders_transducer.py
```diff
@@ -343,10 +326,24 @@ def __init__(self, eprojs, odim, dtype, dlayers, dunits, blank, att,
...
+        z_list = [ey.new_zeros(ey.size(0), self.dunits)]
+        c_list = [ey.new_zeros(ey.size(0), self.dunits)]
+
+        for _ in six.moves.range(1, self.dlayers):
+            z_list.append(ey.new_zeros(ey.size(0), self.dunits))
+            c_list.append(ey.new_zeros(ey.size(0), self.dunits))
+
+        return z_list, c_list
+
+    def rnn_forward(self, ey, dstate):
```
ShigekiKarita: ditto

espnet/nets/pytorch_backend/rnn/decoders_transducer.py
```diff
@@ -35,15 +35,19 @@ def __init__(self, eprojs, odim, dtype, dlayers, dunits, blank,
...
 
         if dtype == "lstm":
-            dec_net = torch.nn.LSTMCell
+            dec_net = torch.nn.LSTM
         else:
-            dec_net = torch.nn.GRUCell
+            dec_net = torch.nn.GRU
 
-        self.decoder = torch.nn.ModuleList([dec_net(embed_dim, dunits)])
+        self.decoder = torch.nn.ModuleList([dec_net(embed_dim, dunits, 1,
```
ShigekiKarita: I believe `self.decoder = dec_net(embed_dim, dunits, dlayers, dropout=dropout, ...)` is faster than stacking each LSTM/Dropout layers because the layers are fused in cudnn implementation.

ShigekiKarita: However, if you think it is too much for this PR, you can skip this comment.

espnet/nets/pytorch_backend/rnn/decoders_transducer.py
b-flo: Hum, maybe the problem is the way I wrote the method? Because it should be almost the same as before.
Does the following docstring sound clear to you : 

```
    def rnn_forward(self, ey, dstate):
        """RNN Forward                                                                                                                                               
                                                                                                                                                                     
        Args:                                                                                                                                                        
            ey (torch.Tensor): batch of input features (B, Lmax, Emb_dim)                                                                                            
            dstate (tuple): tuple with hidden and cell state (NLayers, B, Hdec)                                                                                      
        Returns:                                                                                                                                                     
            output (torch.Tensor): batch of output features (B, Lmax, Hdec)                                                                                          
            dstate (tuple): tuple with hidden and cell state (NLayers, B, Hdec)                                                                                      
        """
```

espnet/nets/pytorch_backend/rnn/decoders_transducer.py
b-flo: If it's okay for you I'll compare both and change in another PR if it's faster!

espnet/nets/pytorch_backend/rnn/decoders_transducer.py
ShigekiKarita: LGTM. Because this method can be changed in the future (e.g., multi-layer cudnn states), it is nice to clarify what this method does at this time.

espnet/nets/pytorch_backend/rnn/decoders_transducer.py
b-flo: Done. Feel free to tell me if you think I should add more comments/doc!

ShigekiKarita: Cool. You almost added docstrings to everything public in this module. Maybe, you can remove this file from the blacklist 
https://github.com/espnet/espnet/blob/f9e8f676be21d530b473384fb8c45afb17396be4/ci/test_flake8.sh#L75

for [flake8-docstrings](https://pypi.org/project/flake8-docstrings/) by a few more changes.

b-flo: What remaining method needs a docstring? Does I need to add a blank docstring under __init__ and decoder_for?

b-flo: Done. One question though : what are the conditions for a first line to end up with `D401: First line should be in imperative mood`?
I'm confused by the fact that `"""RNN forward.` didn't trigger the error in rnn_forward() but did trigger for `Decoder forward.` in `forward()`...

ShigekiKarita: Wonderful!! The D401 seems to have high false positive than false negative. Maybe, its classifier is not enough smart.

espnet/nets/pytorch_backend/rnn/decoders_transducer.py
```diff
@@ -356,7 +356,7 @@ def zero_state(self, ey):
         """Initialize decoder states.
 
         Args:
-            ey (torch.Tensor): batch of input features (B, Lmax, Emb_dim)
+            ey (torch.Tensor): batch of input features (B, Lmax, X_dim)
```
b-flo: I modified this docstring because in some cases I'm using `hs_pad`, which can have a different dimension, for initialization. It is because I'm modyfying `ey` type and a `type error` is thrown in some calculations. As seen in the code, the third dimension is not needed for initialization.
Is that all right?

espnet/nets/pytorch_backend/rnn/decoders_transducer.py
```diff
@@ -356,7 +356,7 @@ def zero_state(self, ey):
         """Initialize decoder states.
 
         Args:
-            ey (torch.Tensor): batch of input features (B, Lmax, X_dim)
+            ey (torch.Tensor): batch of input features (B, (Emb_dim + Eprojs))
```
b-flo: Sorry I was not set on the right definition.
I modified the shape of `zero_state(...)` to match call through `rnn_forward(...)` however I also call `zero_state(...)` alone with `eys` (B, Lmax, Embed_dim) or `hs_pad` (B, Tmax, Henc) in some cases to avoid a type missmatch. Second and third dim are not important here but I though it would be important to be consistent with `rnn_forward(...)`. Is that okay?

ShigekiKarita: OK

ShigekiKarita:
>     * I'm very curious about how this can be easily incorporated with our new beam search (`recog_v2`). I think this is great if we unify the input and output synchronous beam search in a similar manner.

As seen in this part, it essentially consists of encoder-decoder like beam search in the `while True:` loop
https://github.com/espnet/espnet/blob/7569865bfdab094ef7d598a1d662b31489e95f2e/espnet/nets/pytorch_backend/rnn/decoders_transducer.py#L240-L253

I think it is possible but it requires additional loop here
https://github.com/espnet/espnet/blob/0063dc373029e94518d4e0ca58528e3fb32f4b74/espnet/nets/beam_search.py#L276-L285

``` python
def __init__(self, scorers, ...)
    ...
    # get them in advance
    self.win_scorers = [s for s in scorers) if isinstance(WindowedScorer, s)]

def forward(self, x, ...):
    ...
    running_hyps = [self.init_hyp(x)] 
    ended_hyps = []
    # input prefix search 
    for xlen, xi in enumerate(x):
        # output prefix search
        for i in range(maxlen):
            logging.debug('position ' + str(i))
            best = []
            for hyp in running_hyps:
                scores, states = self.score(hyp, x)
                win_scores, win_states =   # TODO
                part_ids = self.pre_beam(scores, device=x.device)
                part_scores, part_states = self.score_partial(hyp, part_ids, x)
        ...
        if len(self.win_scorers) == 0:
            # no need to iterate over x
            break
```
WindowedScorer is derived from [ScorerInterface](https://github.com/espnet/espnet/blob/master/espnet/nets/scorer_interface.py#L9)
```
class WindowedScorerInterface(ScorerInterface):
    pass
```
to identify that it requires the windowed input.

Then any WindowedScorerInterface can use any LMs (implementing ScorerInterface) in `BeamSearch` class.

egs/voxforge/asr1/conf/tuning/train_transducer.yaml
```diff
@@ -13,26 +13,32 @@ patience: 3
...
+eunits: 256
+eprojs: 256
+subsample: "1_1_1_1_1"
+dropout-rate: 0.5
 ## decoder related
 dtype: lstm
 dlayers: 1
-dunits: 300
-dec-embed-dim: 320
+dec-embed-dim: 256
```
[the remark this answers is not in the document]
b-flo: Yes, the number of deletions is reduced if I lower the number of units in all parts without enc pre-init. However I didn't really "tune" : I mainly tested number of layers, numbers of units and dropout_rate. As said, the config could be better I guess

sw005320:
Thanks, @b-flo!
Also, the result looks good by considering that it's just an initial implementation stage.

b-flo:
To be honest I'm quite surprised by the results! In my initial experiments on voxforge, with CTC pre-initialization, vanilla transducer had results a bit better than transducer-attention. For example on same `dt_it` CER was 12.7, between Transformer 100 epochs (12.2) and default RNN (12.9).
Also, from what I observed on small and medium dataset and different setup, degradation for WER is always "controlled". You could end-up with a not-so-competitive CER against others E2E models but be on-par with WER.

sw005320:
Cool.
We might get some reasonable performance with larger tasks without pre-training then.
I'll also start to touch it.

PR #1130, opened 2019-08-23
Updated symlink in Librispeech

kan-bayashi:
I forgot to update symlink of librispeech recipe in #970.


sw005320:
can you keep my single GPU config?

kan-bayashi:
> can you keep my single GPU config?

You mean keep as default?

sw005320:
sorry. Now I understand that you only changed the symlink.

PR #1139, opened 2019-08-25
Add E2E-ST specific modules

hirofumi0810:
- Add E2E-ST specific modules (bin, training, network)
- Add multi-task with ASR (JSON files are already adjusted for this purpose)
- Remove CTC dependencies
- Remove pre-training parts from ASR (related to #1103)
- Update egs/***/st1/run.sh
- Add test

Currently, only pytorch is supported


ci/test_flake8.sh
```diff
@@ -18,6 +18,8 @@ espnet/bin/asr_enhance.py
 espnet/bin/asr_train.py
 espnet/bin/mt_recog.py
 espnet/bin/mt_train.py
+espnet/bin/st_trans.py
+espnet/bin/st_train.py
```
ShigekiKarita: duplicated

ci/test_flake8.sh
```diff
@@ -56,6 +59,7 @@ espnet/nets/pytorch_backend/e2e_asr.py
 espnet/nets/pytorch_backend/e2e_asr_mix.py
 espnet/nets/pytorch_backend/e2e_asr_transducer.py
 espnet/nets/pytorch_backend/e2e_asr_transformer.py
+espnet/nets/pytorch_backend/e2e_st.py
```
ShigekiKarita: Can you remove those files from the blacklist (at least this file)? I think nobody can do this later.

egs/how2/st1/conf/tuning/train_pytorch_transformer.yaml
```diff
@@ -10,7 +10,7 @@ adim: 256
 aheads: 4
 
 # hybrid CTC/attention
-mtlalpha: 0.0  # this must be 0.0
+mtlalpha: 0.0
```
kan-bayashi: Is it  OK? (looks different from other yaml)

espnet/st/pytorch_backend/st.py
```diff
@@ -0,0 +1,632 @@
...
+import matplotlib
+matplotlib.use('Agg')
+
+if sys.version_info[0] == 2:
+    from itertools import izip_longest as zip_longest
+else:
+    from itertools import zip_longest as zip_longest
+
+
+class CustomUpdater(StandardUpdater):
```
ShigekiKarita: Is it possible to reuse this Updater and Converter from ASR?

egs/how2/st1/conf/tuning/train_pytorch_transformer.yaml
hirofumi0810: Transformer is not supported in ST modules (using ASR modules).

ci/test_flake8.sh
hirofumi0810: I think we cannot pass tests without this (even for ASR).

espnet/st/pytorch_backend/st.py
hirofumi0810: I'm using multitask learning with ASR. Therefore I made this PR.

egs/how2/st1/conf/tuning/train_pytorch_transformer.yaml
hirofumi0810: Of course, I can modify this PR for Transformer, but I think it's too much in a single PR.

ci/test_flake8.sh
kan-bayashi: We try to remove from these list.
Since we do not want to add new files, if you update your docstrings, it is very helpful.

espnet/nets/pytorch_backend/e2e_st.py
```diff
@@ -0,0 +1,497 @@
...
+        loss_data = float(self.loss)
+        if not math.isnan(loss_data):
+            self.reporter.report(loss_asr_data, loss_st_data, self.acc_asr, acc,
+                                 cer, wer, bleu, loss_data)
+        else:
+            logging.warning('loss (=%f) is not correct', loss_data)
+        return self.loss
+
+    def scorers(self):
+        return dict(decoder=self.dec, ctc=CTCPrefixScorer(self.ctc, self.eos))
```
ShigekiKarita:
```suggestion
        return dict(decoder=self.dec)
```

espnet/bin/st_trans.py
```diff
@@ -0,0 +1,171 @@
...
+        logging.error("It seems that both --rnnlm and --word-rnnlm are specified. Please use either option.")
+        sys.exit(1)
+
+    # trans
+    logging.info('backend = ' + args.backend)
+    if args.backend == "pytorch":
+        # Experimental API that supports custom LMs
+        if args.api == "v2":
+            from espnet.asr.pytorch_backend.recog import recog_v2
+            recog_v2(args)
```
ShigekiKarita: Can you add test for `espnet.nets.beam_search.BeamSearch` with ST models under test_beam_search?

egs/how2/st1/conf/tuning/train_pytorch_transformer.yaml
kan-bayashi: Then, is it needed to add here?

espnet/st/pytorch_backend/st.py
ShigekiKarita: OK. Now I can see the diference in CustomConverter. However, I expect that we can extend CustomConverter to support multiple output json. This strategy is hard to maintain for us because the code has many duplicated parts.

I also expect that CustomUpdater can be unified.

ci/test_flake8.sh
ShigekiKarita: You can pass if you add docstrings.

egs/how2/st1/conf/tuning/train_pytorch_transformer.yaml
kan-bayashi: OK. Just keep for next PR.

setup.py
```diff
@@ -41,6 +41,7 @@
         'PyYAML>=5.1.2',
         'sentencepiece>=0.1.82',
         'pysptk>=0.1.17',
+        'nltk',
```
kan-bayashi: Could you add version info? (it is OK to just add your using version with `>=`)

ShigekiKarita:
The merging branch should be v.0.6.0 because there are a lot of breaking changes in cmd options.

sw005320:
> The merging branch should be v.0.6.0 because there are a lot of breaking changes in cmd options.

I agree.

espnet/bin/st_trans.py
```diff
@@ -0,0 +1,171 @@
...
+                        help='Word RNNLM model config file to read')
+    parser.add_argument('--word-dict', type=str, default=None,
+                        help='Word list to read')
+    parser.add_argument('--lm-weight', type=float, default=0.1,
+                        help='RNNLM weight')
+    # streaming related
+    parser.add_argument('--streaming-mode', type=str, default=None,
+                        choices=['window', 'segment'],
+                        help="""Use streaming translator for inference.
+                        `--batchsize` must be set to 0 to enable this mode""")
```
kan-bayashi: ditto

espnet/st/pytorch_backend/st.py
```diff
@@ -0,0 +1,632 @@
...
+            # Don't create ComplexTensor and give it E2E here
+            # because torch.nn.DataParellel can't handle it.
+            xs_pad = {'real': xs_pad_real, 'imag': xs_pad_imag}
+        else:
+            xs_pad = pad_list([torch.from_numpy(x).float() for x in xs], 0).to(device, dtype=self.dtype)
+
+        ilens = torch.from_numpy(ilens).to(device)
+        ys_pad = pad_list([torch.from_numpy(np.array(y[0]) if isinstance(y, tuple) else y).long()
+                           for y in copy.deepcopy(ys)], self.ignore_id).to(device)
+        if self.asr_task:
```
ShigekiKarita: Instead of supporting multitask data loading in CustomConverter, maybe you can derive this class from `espnet.asr.pytorch_backend.asr.CustomConverter` and then
```python
def __call__(self, batch, device):
    xs_pad, ilens, ys_pad = super().__call__(batch, device)
    if self.asr_task:
    ...
``

File: espnet/nets/st_interface.py [hunk 10]
```diff
@@ -0,0 +1,31 @@
+"""ST Interface module."""
+
+from espnet.nets.asr_interface import ASRInterface
+
+
+class STInterface(ASRInterface):
```
ShigekiKarita: If you just wanna rename `recognize` to `translate`. STInterface should transfer them
```python
class STInterface(ASRInterface):
    def recognize(self, x, trans_args, char_list=None, rnnlm=None):
        return self.translate(self, x, trans_args, char_list, rnnlm)

    def recognize_batch(self, x, trans_args, char_list=None, rnnlm=None):
        return self.translate_batch(self, x, trans_args, char_list, rnnlm)

    ...
```

If not so, STInterface should not be derived from ASRInterface. This is because they are not `is-a and we cannot cast STInterface to STInterface.

espnet/st/pytorch_backend/st.py
```diff
@@ -0,0 +1,632 @@
...
+    """Decode with the given args.
+
+    Args:
+        args (namespace): The program arguments.
+    """
+    set_deterministic_pytorch(args)
+    model, train_args = load_trained_model(args.model)
+    assert isinstance(model, STInterface)
+    args.ctc_weight = 0.0
+    model.trans_args = args
```
ShigekiKarita: If you forward the `translate*` to `recognize*`, you can reuse `recog` here.

espnet/nets/st_interface.py
hirofumi0810: I don't use recognize* now, so this is not what I wanna do.
Also, I want to keep both recognize* and translate* for both ST and ASR branches when we use multi-task learning (ST+ASR).

ci/test_flake8.sh
hirofumi0810: OK. But this is not a problem for RNN-T...

ShigekiKarita: RNNT and flake8 docstrings were merged at the same time and I added that against my will... #1107

espnet/nets/st_interface.py
ShigekiKarita: OK. I understand. Can you describe this hierarchies in the Notes section of docstrings?

espnet/bin/st_trans.py
hirofumi0810: I will remove this for the clarity of this PR.

ShigekiKarita: OK!

test/test_e2e_st.py
```diff
@@ -0,0 +1,472 @@
...
+
+    with torch.no_grad(), chainer.no_backprop_mode():
+        in_data = np.random.randn(10, 40)
+        model.translate(in_data, args, args.char_list)  # decodable
+        if "pytorch" in module:
+            batch_in_data = [np.random.randn(10, 40), np.random.randn(5, 40)]
+            model.translate_batch(batch_in_data, args, args.char_list)  # batch decodable
+
+
+# def test_segment_streaming_e2e():
```
kan-bayashi: If you have special reason of comment out, it is better to add here.

espnet/bin/st_trans.py
```diff
@@ -0,0 +1,169 @@
...
+    parser.add_argument('--trans-json', type=str,
+                        help='Filename of translation data (json)')
+    parser.add_argument('--result-label', type=str, required=True,
+                        help='Filename of result label data (json)')
+    # model (parameter) related
+    parser.add_argument('--model', type=str, required=True,
+                        help='Model file parameters to read')
+    parser.add_argument('--model-conf', type=str, default=None,
+                        help='Model config file')
+    parser.add_argument('--num-spkrs', type=int, default=1,
```
kan-bayashi: Maybe not needed.

espnet/st/pytorch_backend/st.py
```diff
@@ -0,0 +1,525 @@
...
+        if args.preprocess_conf is None else args.preprocess_conf,
+        preprocess_args={'train': False})
+
+    if args.batchsize == 0:
+        with torch.no_grad():
+            for idx, name in enumerate(js.keys(), 1):
+                logging.info('(%d/%d) decoding ' + name, idx, len(js.keys()))
+                batch = [(name, js[name])]
+                feat = load_inputs_and_targets(batch)[0][0]
+                if args.streaming_mode == 'window':
```
kan-bayashi: In my understanding, streaming module is only working with `ctc-weight > 0.0`.
So it is better to remove the part related to `streaming` and also remove the related options.

espnet/st/pytorch_backend/st.py
```diff
@@ -0,0 +1,493 @@
...
+        Args:
+            subsampling_factor (int): The subsampling factor.
+            dtype (torch.dtype): Data type to convert.
+            asr_task (bool): multi-task with ASR task.
+
+        """
+        self.subsampling_factor = subsampling_factor
+        self.ignore_id = -1
+        self.dtype = dtype
+        self.asr_task = asr_task
```
ShigekiKarita:
```python
super().__init__(subsampling_factor=subsampling_factor, dtype=dtype)
self.asr_task = asr_task
```
is better. It can correctly initialize CustomConverter when it will be updated.

ShigekiKarita:
Because st.py is not tested at all, it can be easily broken. I hope that the integration test using the subset of iwslt18 is needed in `ci/test_integration.sh`. @kan-bayashi how do you think?

ShigekiKarita:
Can you try `git rebase --onto v.0.6.0 master move_e2est`?

kan-bayashi:
I agree to add integration test.
It is better to merge this PR into v.0.6.0 and then let us make a minimal recipe for integration test.

ShigekiKarita:
OK. Let's add the integration test in the future PR.

espnet/nets/pytorch_backend/transformer/utils.py
```diff
@@ -0,0 +1,24 @@
+import torch
+
+"""Unility funcitons for Transformer."""
+
+
+def add_sos_eos(ys_pad, sos, eos, ignore_id):
```
ShigekiKarita: As we prefer to tiny modules,  you should name this file as `add_sos_eos` instead of utils.

ShigekiKarita:
It is getting in good shape. Now I have nothing to leave except for one comment above.

espnet/st/pytorch_backend/st.py
```diff
@@ -0,0 +1,491 @@
...
+import matplotlib
+matplotlib.use('Agg')
+
+if sys.version_info[0] == 2:
+    from itertools import izip_longest as zip_longest
+else:
+    from itertools import zip_longest as zip_longest
+
+
+class CustomConverter_v2(CustomConverter):
```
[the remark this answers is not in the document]
kan-bayashi: Maybe he wants to inherit.
Why don't you modify as follows?

```python
from espnet.asr.pytorch_backend.asr import CustomConverter as ASRCustomConverter

class CustomConverter(ASRCustomConverter):
```

ShigekiKarita:
Many thanks for updating many docstrings, change maker!

PR #1153, opened 2019-08-29
fix recursive make

b-flo:
Hi,

`warprnnt_pytorch` will fails if `make -j X` is used. Switching from command name to variable MAKE resolve the issue.

### DIFF
--- a/tools/Makefile
+++ b/tools/Makefile
@@ -86,7 +86,7 @@ warp-ctc.done: espnet.done
 	if . venv/bin/activate && python -c 'import torch as t;assert t.__version__[0] == "1"' &> /dev/null; then \
         cd warp-ctc; git checkout -b pytorch-1.0.0 remotes/origin/pytorch-1.0.0; \
     fi
-	. venv/bin/activate; cd warp-ctc && mkdir build && cd build && cmake .. && make -j4 ; true
+	. venv/bin/activate; cd warp-ctc && mkdir build && cd build && cmake .. && $(MAKE); true
 	. venv/bin/activate; pip install cffi
 	. venv/bin/activate; cd warp-ctc/pytorch_binding && python setup.py install # maybe need to: apt-get install python-dev
 	touch warp-ctc.done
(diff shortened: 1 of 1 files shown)


kan-bayashi:
In my understanding, when we run 
cd tools; make -j4 
$(MAKE) becomes make -j4
Isn’t it opposite to your explanation?

b-flo:
> "In my understanding, when we run ... $(MAKE) becomes make -j4

That's right, parent's jobserver is coordinated with child's jobserver like this. If `make` in Makefile is used, parent and children could run different number of jobs in the same build. 

Also, from from GNU doc : 

> Recursive make commands should always use the variable MAKE, not the explicit command name ‘make’
https://www.gnu.org/software/make/manual/make.html#MAKE-Variable

Currently, if `make -j X` is used, the submake for `warprnnt_pytorch` end-up silently in background. My first guess was a issue with running jobs so I tried this fix. It works but I may be wrong on the reasons?

kan-bayashi:
I see. Actually I'm not familiar with make.
BTW, if so, we also should use $(MAKE) in the following line?
https://github.com/espnet/espnet/blob/b4e0b01d8c78d54d0ca6470bb1ebd8fdbecd4d2b/tools/Makefile#L89

b-flo:
I think we should, $(MAKE) is used everywhere except for `warp-ctc` and `warprnnt-pytorch` currently. Also, I'm not sure forcing `-j 4` for make works as intended. The following warning are displayed during build (sorry for the wall of text...):

>‘warning: -jN forced in submake: disabling jobserver mode.’
>
>This warning and the next are generated if make detects error conditions related to parallel processing on systems where sub-makes can communicate (see Communicating Options to a Sub-make). This warning is generated if a recursive invocation of a make process is forced to have ‘-jN’ in its argument list (where N is greater than one). This could happen, for example, if you set the MAKE environment variable to ‘make -j2’. In this case, the sub-make doesn’t communicate with other make processes and will simply pretend it has two jobs of its own.

>‘warning: jobserver unavailable: using -j1. Add `+' to parent make rule.’
>
> In order for make processes to communicate, the parent will pass information to the child. Since this could result in problems if the child process isn’t actually a make, the parent will only do this if it thinks the child is a make. The parent uses the normal algorithms to determine this (see How the MAKE Variable Works). If the makefile is constructed such that the parent doesn’t know the child is a make process, then the child will receive only part of the information necessary. In this case, the child will generate this warning message and proceed with its build in a sequential manner.

sw005320:
> Edit: I also modified make for warp-ctc. If there was a specific reason for using make -j4 here I'll revert this change!

I don't think we would have some specific reason.

This makes it very consistent.
Thanks, @b-flo!

PR #1159, opened 2019-08-30
using python3 as default

zh794390558:


Fhrozen:
Is this required? When you install espnet with the default python, Fonda links python3 as python, so this may be no required

kan-bayashi:
Maybe not needed.
But currently both python and python3 exists.
This may be confusing.
I think this PR helps to improve consistency.

ShigekiKarita:
I agree. 

In fact, we do not use conda sometimes like Google Colab. And some linux distribution e.g., [RHEL8](https://developers.redhat.com/blog/2018/11/14/python-in-rhel-8/) will remove unversioned `python` command. It is good to use explicit python3 now.

PR #1160, opened 2019-08-30
Update Russian open STT recipe for v0.5 of the dataset

akreal:
Current amount of training data is about 3800 hours.


egs/ru_open_stt/asr1/RESULTS.md
```diff
@@ -0,0 +1,24 @@
+# Transformer (elayers=12, dlayers=6, units=2048, 4 GPUs, specaug)
```
sw005320: How about keeping some of the old results to see the difference?

egs/ru_open_stt/asr1/conf/tuning/decode_pytorch_transformer_large.yaml
```diff
@@ -0,0 +1,6 @@
+beam-size: 2
```
sw005320: why did you use a very small beam?
Did you see some degradations or computational issues?

egs/ru_open_stt/asr1/RESULTS.md
akreal: The new results are not directly comparable to the old results because they use different subsets for testing. Current release of the dataset added special validation subsets with manually checked annotations and they can be used for comparisons in the future.

egs/ru_open_stt/asr1/conf/tuning/decode_pytorch_transformer_large.yaml
akreal: Out of several tested beam sizes (20, 5, 2, 1) this one resulted the best WER, while others were slightly worse.

akreal:
Thank you!

> Can you also review the information about this task in https://github.com/espnet/espnet/blob/master/egs/README.md?
> Please fix it if needed.

It looks fine.

PR #1173, opened 2019-09-03
Transducer v1.2

b-flo:
Hi,

Sorry I didn’t have too much time last weeks.

- Add remaining docstrings related to transducer and finetuning. Remove `asr/pytorch_backend/asr_init.py` and `nets/pytorch_backend/e2e_asr_transducer.py` from flake8 blacklist
- Fix and switch back to LSTMCell. Enable decoder initialization with CE-trained LM.
- Fix wrong tabulation in `asr/pytorch_backend/asr_init.py`. It only impacted dec init with LM, which was not used before.
- Add `egs/voxforge/asr1/config/tuning/transducer` directory containing all configs used to reproduce results (including for pre-trained models)
- Add complete results to `egs/voxforge/asr1/RESULTS.md`

The only missing item now is batch decoding. It’s quite straightforward to implement by adding a batch loop but I’m trying to figure out a trick to speed up batch beam search without too much degradation. Currently increasing batch also increase decoding time. Batch greedy search seems okay tough. If someone has any idea?


sw005320:
@b-flo, again great contributions!

> The only missing item now is batch decoding. It’s quite straightforward to implement by adding a batch loop but I’m trying to figure out a trick to speed up batch beam search without too much degradation. Currently increasing batch also increase decoding time. Batch greedy search seems okay tough. If someone has any idea?

Are you following our batch decoding? I'm curious.

FYI, ideally, I want to merge all our beam search implementations under our recog_v2 interface to avoid to maintain multiple implementations for beam search.
I and @ShigekiKarita are thinking of how we unify the input-synchronous (RNN-T) and output-synchronous beam search algorithms (or keeping both in a similar manner).
Then, we can also extend it to include batch decoding.

sw005320:
Can you add a recipe for RNN-T?
I think it would be complicated if we include pre-training and RNN-T/RNN-ATT with the current `run.sh`. So you may make `local/run_rnnt.sh` for now.

b-flo:
> Are you following our batch decoding? I'm curious.

I wanted to initially! But I couldn't make it works with the vertical transitions and the joint procedure..

> I want to merge all our beam search implementations under our recog_v2 interface to avoid to maintain multiple implementations for beam search. [...] Then, we can also extend it to include batch decoding.

If batch decoding is not critical I can shift my focus on adapting transducer beam search with the new interface if you want.

> Can you add a recipe for RNN-T?

Sure! I'll add that asap!

sw005320:
> > I want to merge all our beam search implementations under our recog_v2 interface to avoid to maintain multiple implementations for beam search. [...] Then, we can also extend it to include batch decoding.
> 
> If batch decoding is not critical I can shift my focus on adapting transducer beam search with the new interface if you want.
> 

Hmm. Difficult question.
In terms of the management perspective, the new beam search interface is better, but in terms of the function (especially RNN-T is aimed for on-line decoding) batch processing should be a higher priority.

@ShigekiKarita, do you have some comments?

b-flo:
> but in terms of the function (especially RNN-T is aimed for on-line decoding) batch processing should be a higher priority.

I'm not sure I understand what you mean with "on-line" decoding? Batch processing is for offline mode and online or streaming decoder would only process one input wav at the time. No?

sw005320:
Our current batch decoding parallelizes multiple hypotheses as well as multiple utterances (that is, we parallelize #hypos x #utts). So, even if we only use a single utterance, we can get significant speed improvement due to the hypotheses computation parallelization, which is necessary for on-line decoding (sorry that on-line and batch do not sound like they both exist, which confuses you).

b-flo:
Oh I got it now, thanks!

b-flo:
@sw005320 As I supposed you wanted a different run script for transducer to avoid switching configs for finetuning, I modified the usual run to incorporate pre-trainings and added finetuning options in it instead of config file. It should be temporary as it's not really user-friendly, we should maybe think of the run case when user want to train multiple models and transfer?
Also, I added `[WIP]` to this PR for now. I'm retraining everything to verify I didn't make a mistake. It shouldn't take long.

egs/voxforge/asr1/run_rnnt.sh
```diff
@@ -0,0 +1,289 @@
...
+lm_config=conf/tuning/transducer/lm_transducer.yaml
+
+if [ ${use_transfer} == true ]; then
+    exp_config=conf/tuning/transducer/train_transducer_${type_transfer}_init.yaml
+
+    sed '$a\' ${train_config} > ${exp_config}
+
+    case "${type_transfer}" in
+        enc|both)
+            echo "enc-init: 'exp/tr_it_pytorch_train_ctc/results/model.loss.best'" >> ${exp_config}
```
sw005320: Is it possible to change this `tr_it_pytorch_train_ctc` flexible (e.g., `tr_it_${backend}__$(basename ${ctc_config%.*}`)?
Then, we can change `ctc_config` to the other file.

Same comment for `train_rnnlm_pytorch_lm_transducer` and `tr_it_pytorch_train_att`.

egs/voxforge/asr1/run_rnnt.sh
```diff
@@ -0,0 +1,289 @@
...
+         data/dt_${lang} ${dict} > ${feat_dt_dir}/data.json
+    for rtask in ${recog_set}; do
+        feat_recog_dir=${dumpdir}/${rtask}/delta${do_delta}
+        data2json.sh --feat ${feat_recog_dir}/feats.scp \
+            data/${rtask} ${dict} > ${feat_recog_dir}/data.json
+    done
+fi
+
+if [ ${stage} -le 3 ] && [ ${stop_stage} -ge 3 ]; then
+    training_run() {
```
sw005320: It may be better to add some comments about `training_run()`.
This general function is used for ctc, att, and transducer (except for LM).

egs/voxforge/asr1/run_rnnt.sh
b-flo: I suppose the other files you're talking about are the default rnn and lm configs, right? If so, I made different configs to match parameters between model (and also avoid user problems...).

I think I'll isolate this part. I'll add an utility script taking as argument main config, finetuning options and optional configs (CTC, Att and LM). It will create or modify the different needed configs with matched parameters, and return all configs paths. It will be easier to maintain like this and only the network training part would be different from the usual running script. What do you think @sw005320, @kan-bayashi?

sw005320: This sounds good. I think the pre-training -> fine-tuning becomes more important and such utility function helps to clean the script.

kan-bayashi: I agree with your idea.
BTW, why don't you `utils/change_yaml.py` to change yaml file?
https://github.com/espnet/espnet#changing-the-configuration

kan-bayashi:
> If we put it local/run_rnnt.sh, it may not be well maintained.

I think so too :(
It may better to create the new directory `asr2` which contains pretraining part.
(The naming is not sure...)
And then create the README to explain the difference in `egs/voxforge`.

@b-flo Do you think that it is possible to apply the same pretraining flow to @hirofumi0810's ST?
(Just my curious. I'm not sure pretraining is still needed in ST.)

b-flo:
> @b-flo Do you think that it is possible to apply the same pretraining flow to @hirofumi0810's ST?
(Just my curious. I'm not sure pretraining is still needed in ST.)

Sure! Transfer learning can be done with any ASR/MT modules so the complete flow could be applied to ST.

Actually the main difficulty of this workflow is the config preparation. As you could mix any ASR models it's quite difficult to create default configs for each cases, give some freedom to the user for finetuning (e.g.: let him specify some changes in created default config without the need to provide a new config), etc, without hardcoding. In fact I think, if we want a general workflow, this step should be handled at python level.
For now, the utility script I'm making will strictly works for transducer but it can be a basis for discussion. I'll see if I can propose an acceptable solution to generalize this step.

egs/voxforge/asr1/run_rnnt.sh
b-flo: Because I was not aware of it! Thanks!
However it seems a bit overkill to use this script, no? Append/Delete entries to/from a file is rather straightforward with bash builtin and I don't think there is a gain in speed/memory.

egs/voxforge/asr1/run_rnnt.sh
b-flo: Done!

b-flo:
I modified the run script and added the utility script. It's a bit rough but it should work for every recipe as long as the transducer config architecture is not totally messed up. I also started working on the config generation/handler for finetuning directly in espnet framework, but I won't allocate too much time on it for now.

On another note, I'll re-run all recipes and modify `RESULTS.md` to remove WERs and put a 10-run-average CERs. The recipe can demonstrates slightly different results between runs compared to the ones I reported (+/- 0.2 CER for better or worse from what I observed). It would not be a problem for CER but WER difference could be misleading for users.

ShigekiKarita:
Sorry for late.

> Hmm. Difficult question.
> In terms of the management perspective, the new beam search interface is better, but in terms of the function (especially RNN-T is aimed for on-line decoding) batch processing should be a higher priority.
> 
> @ShigekiKarita, do you have some comments?

I think old beam search api (v1) is ok because we do not have input-sync decoding in `espnet/net/beam_search.py` yet. We (maybe I) can reimplement v2 api later.

b-flo:
Hi,

Sorry it took some time. In the end I tried to found a more stable setup as I observed bigger variations than previously (+/- 0.4%). Results are a bit different but it seems better like this.

sw005320:
Can you fix the conflict?
I'll merge it.

b-flo:
No problem but I don't really know what to fix... Is the conflict not due to the file references I removed in `ci/test_flake8.sh` (`asr/pytorch_backend/asr_init.py` and `nets/pytorch_backend/e2e_asr_transducer.py`)?

sw005320:
I think so.
Then, the master also removed `asr_mix` related, which seems to cause the conflict.

sw005320:
> Edit: Before merging, would it be possible to have someone test the recipe? It shouldn't take long for one full run with only one GPU (less than 3 hours with all trainings + decode for vanilla transducer).
> It may be good to have minimal feedback from different users.

@Masao-Someki, could you test the recipe again with this version?

Masao-Someki:
okay!

Masao-Someki:
I tested `run_rnnt.sh` with  `ngpu=1` .
The recipe test has been done successfully!
Results with default config in `tuning/transducer/*.yaml` :
```
write a CER (or TER) result in exp/tr_it_pytorch_train_transducer/decode_dt_it_decode_transducer/result.txt
  | SPKR        | # Snt  # Wrd | Corr     Sub    Del     Ins    Err   S.Err |
  | Sum/Avg     |   21    1452 |  3.8     0.1   96.1     0.0   96.2   100.0 |
write a WER result in exp/tr_it_pytorch_train_transducer/decode_dt_it_decode_transducer/result.wrd.txt
| SPKR         | # Snt  # Wrd  | Corr     Sub    Del     Ins     Err   S.Err  |
| Sum/Avg      |   21     250  |  0.0     8.4   91.6     0.0   100.0   100.0  |
write a CER (or TER) result in exp/tr_it_pytorch_train_transducer/decode_et_it_decode_transducer/result.txt
  | SPKR             | # Snt # Wrd | Corr    Sub    Del    Ins    Err  S.Err |
  | Sum/Avg          |   19   1573 |  3.0    0.0   97.0    0.0   97.0  100.0 |
write a WER result in exp/tr_it_pytorch_train_transducer/decode_et_it_decode_transducer/result.wrd.txt
| SPKR             | # Snt  # Wrd | Corr    Sub     Del    Ins    Err   S.Err |
| Sum/Avg          |   19     243 |  0.0    7.8    92.2    0.0  100.0   100.0 |
```

b-flo:
Thanks @Masao-Someki! It seems something went wrong though looking at the number of sentences and chars/words...
Previously, someone had an encoding problem with voxforge recipe (lot of `"The transcript for 'X'(user 'Y') is not properly normalized - [utterance]` in logs), is it similar for you?

Masao-Someki:
I checked the log file and find that there are 6240 UnicodeDecodeErrors...
Logs are at https://drive.google.com/open?id=1qpU-1ZiMn5sKSA5G2TEWHJ3AGybojN51

b-flo:
At least the error in obvious on your side, thanks!

`UnicodeDecodeError: 'ascii' codec can't decode byte 0xc3 in position 202: ordinal not in range(128)`

It seems `ascii` is called for decode, which is a problem with letters containing diatric...
I didn't encounter the problem, maybe it has to do with your `locale`? All mine are set with `fr_FR.UTF-8` so It could be the reason.

I have to travel today, I'll take a look when I return

Masao-Someki:
It fixed and ran successfully
Results:
```
write a CER (or TER) result in exp/tr_it_pytorch_train_transducer/decode_dt_it_decode_transducer/result.txt
| SPKR                  | # Snt  # Wrd | Corr    Sub    Del    Ins    Err  S.Err |
| Sum/Avg               | 1082   79133 | 89.8    5.0    5.3    2.6   12.8   97.0 |
write a WER result in exp/tr_it_pytorch_train_transducer/decode_dt_it_decode_transducer/result.wrd.txt
| SPKR                  | # Snt # Wrd | Corr    Sub    Del    Ins    Err  S.Err |
| Sum/Avg               | 1082  13235 | 62.1   31.7    6.2    4.7   42.6   97.0 |
2019-09-21 20:31:03,152 (concatjson:36) INFO: /work4/m_someki/espnet/tools/venv/bin/python3 /home/m_someki/workspace4/espnet/egs/voxforge/asr1/../../../utils/concatjson.py exp/tr_it_pytorch_train_transducer/decode_et_it_decode_transducer/data.1.json exp/tr_it_pytorch_train_transducer/decode_et_it_decode_transducer/data.10.json exp/tr_it_pytorch_train_transducer/decode_et_it_decode_transducer/data.11.json exp/tr_it_pytorch_train_transducer/decode_et_it_decode_transducer/data.12.json exp/tr_it_pytorch_train_transducer/decode_et_it_decode_transducer/data.13.json exp/tr_it_pytorch_train_transducer/decode_et_it_decode_transducer/data.14.json exp/tr_it_pytorch_train_transducer/decode_et_it_decode_transducer/data.15.json exp/tr_it_pytorch_train_transducer/decode_et_it_decode_transducer/data.16.json exp/tr_it_pytorch_train_transducer/decode_et_it_decode_transducer/data.2.json exp/tr_it_pytorch_train_transducer/decode_et_it_decode_transducer/data.3.json exp/tr_it_pytorch_train_transducer/decode_et_it_decode_transducer/data.4.json exp/tr_it_pytorch_train_transducer/decode_et_it_decode_transducer/data.5.json exp/tr_it_pytorch_train_transducer/decode_et_it_decode_transducer/data.6.json exp/tr_it_pytorch_train_transducer/decode_et_it_decode_transducer/data.7.json exp/tr_it_pytorch_train_transducer/decode_et_it_decode_transducer/data.8.json exp/tr_it_pytorch_train_transducer/decode_et_it_decode_transducer/data.9.json
2019-09-21 20:31:03,192 (concatjson:46) INFO: new json has 1055 utterances
2019-09-21 20:31:03,770 (json2trn:43) INFO: /work4/m_someki/espnet/tools/venv/bin/python3 /home/m_someki/workspace4/espnet/egs/voxforge/asr1/../../../utils/json2trn.py exp/tr_it_pytorch_train_transducer/decode_et_it_decode_transducer/data.json data/lang_1char/tr_it_units.txt --num-spkrs 1 --refs exp/tr_it_pytorch_train_transducer/decode_et_it_decode_transducer/ref.trn --hyps exp/tr_it_pytorch_train_transducer/decode_et_it_decode_transducer/hyp.trn
2019-09-21 20:31:03,770 (json2trn:45) INFO: reading exp/tr_it_pytorch_train_transducer/decode_et_it_decode_transducer/data.json
2019-09-21 20:31:03,787 (json2trn:49) INFO: reading data/lang_1char/tr_it_units.txt
write a CER (or TER) result in exp/tr_it_pytorch_train_transducer/decode_et_it_decode_transducer/result.txt
| SPKR                  | # Snt  # Wrd | Corr    Sub    Del    Ins    Err  S.Err |
| Sum/Avg               | 1055   77966 | 89.9    5.1    5.1    2.5   12.6   97.2 |
write a WER result in exp/tr_it_pytorch_train_transducer/decode_et_it_decode_transducer/result.wrd.txt
| SPKR                  | # Snt # Wrd | Corr    Sub    Del    Ins    Err  S.Err |
| Sum/Avg               | 1055  12990 | 61.8   31.9    6.3    5.2   43.4   97.2 |
```

b-flo:
Great, thanks a lot! What was the exact issue and the fix you used? We should maybe tackle the problem in espnet or notify users as you're not the first one to report an encoding problem with voxforge!

Masao-Someki:
The problem was very simple. The error was occured because the python3 uses 'ascii' when opening files. So I modified line31 in `local/make_trans.py` as follows.
```
<before> 
for l in open(sys.argv[1]):

<after>
for l in open(sys.argv[1], encoding='utf-8'):
```

b-flo:
Thanks, I added the fix in this PR! I wasn't aware that unlike python2, python3's `open` set a default encoding depending on your env:

`In text mode, if encoding is not specified the encoding used is platform dependent: locale.getpreferredencoding(False) is called to get the current locale encoding. `
https://docs.python.org/3.7/library/functions.html#open

sw005320:
Thanks a lot!
We may still need to think of the directory name, but it can be done later with our further investigations of this RNN-T.

b-flo:
Could you keep me updated on your investigations please? I have several additions I would like to test and add depending on feedbacks/observations. Some are already done but I didn't have time to test intensively!

PR #1180, opened 2019-09-06
Support multiple inputs

ruizhilijhu:
modified data2json.sh and io_utils.py to support multiple parallel input cases.

### DIFF
--- a/espnet/utils/io_utils.py
+++ b/espnet/utils/io_utils.py
@@ -165,8 +165,7 @@ class LoadInputsAndTargets(object):
 
         if self.preprocessing is not None:
             # Apply pre-processing all input features
-            x_names = [i for i in return_batch.keys() if i.startswith("input")]
-            for x_name in x_names:
+            for x_name in [i for i in return_batch.keys() if i.startswith("input")]:
                 return_batch[x_name] = \
                     self.preprocessing(return_batch[x_name], uttid_list,
                                        **self.preprocess_args)
--- a/espnet/utils/io_utils.py
+++ b/espnet/utils/io_utils.py
@@ -165,10 +165,10 @@ class LoadInputsAndTargets(object):
 
         if self.preprocessing is not None:
             # Apply pre-processing all input features
-            for x_name in [i for i in return_batch.keys() if i.startswith("input")]:
-                return_batch[x_name] = \
-                    self.preprocessing(return_batch[x_name], uttid_list,
-                                       **self.preprocess_args)
+            for x_name in return_batch.keys():
+                if x_name.startswith("input"):
+                    return_batch[x_name] = self.preprocessing(
+                        return_batch[x_name], uttid_list, **self.preprocess_args)
 
         # Doesn't return the names now.
         return tuple(return_batch.values())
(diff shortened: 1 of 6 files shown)


ruizhilijhu:
Does failing in codecov/patch prevent the PR from merging

sw005320:
`codecov` aims to increase unit test coverage. It would be good to increase the coverage but this is not mandatory.

ruizhilijhu:
Thanks! I will do a PR after this to support multi-encoder.

sw005320:
I think this is related to multichannel ASR. @ShigekiKarita and @kamo-naoyuki, do you have any comments for this extension?

ShigekiKarita:
Thanks for your fine-grained PR! I will check the detail later today. But can you add tests? Even stupid cases really help us understand what these changes are for.

For example, the `LoadInputsAndTargets` is tested at https://github.com/espnet/espnet/blob/master/test/test_utils.py#L58

Sadly, data2json does not have any existing test yet but you can refer this
https://github.com/espnet/espnet/blob/master/test_utils/test_scp2json_py.bats
how to write test for scripts.

ruizhilijhu:
@ShigekiKarita Thanks for the suggestions. I am trying to write test for data2json.sh, but there's a lot of KALDI dependencies inside data2json.sh. For example, a lot of files used in utils/ which is a softlink to kaldi. Any suggestions to this?

ShigekiKarita:
Thanks for adding the tests. It looks really helfull!

> For example, a lot of files used in utils/ which is a softlink to kaldi. Any suggestions to this?

Hmm. You can see CircleCI failed while Travis succeeded because Travis executes `ci/install_kaldi.sh` before `ci/test_shell.sh`. On the other hand, circle ci does not. So, can you put these lines in `ci/test_shell.sh`?
```bash
#!/usr/bin/env bash

if [ ! -e tools/kaldi ]; then
    git clone https://github.com/kaldi-asr/kaldi --depth 1 tools/kaldi
fi

...
```

If this also fails, I suggest that:
- move `ci/install_kaldi.sh` before `ci/test_shell.sh` in circleci that installs our prebuilt kaldi
- remove `ci/test_shell.sh` from centos7 image in circleci because we did not provide prebuilt kaldi for CentOS7 image (so it cannot execute `ci/install_kaldi.sh`).


I think I am the main CentOS7 user in the ESPnet but testing data2json.sh has higher priority than testing shell scripts in CentOS7. (@kan-bayashi maybe you are also CentOS7 user? Do you agree?)

ruizhilijhu:
@ShigekiKarita Thanks for all the help! It seems like it works out finally!

espnet/utils/io_utils.py
```diff
@@ -164,10 +164,10 @@ def __call__(self, batch):
             raise NotImplementedError
 
         if self.preprocessing is not None:
-            # Apply pre-processing only to input1 feature, now
-            if 'input1' in return_batch:
-                return_batch['input1'] = \
-                    self.preprocessing(return_batch['input1'], uttid_list,
+            # Apply pre-processing all input features
+            for x_name in [i for i in return_batch.keys() if i.startswith("input")]:
```
ShigekiKarita:
```python
for x_name in return_batch.keys():
    if x_name.startswith("input"):
        return_batch[x_name] = self.preprocessing(
            return_batch[x_name], uttid_list, **self.preprocess_args)
```
I think this more readable.

ShigekiKarita:
Looks great except for one tiny comment.

sw005320:
Thanks, @ruizhilijhu!

PR #1193, opened 2019-09-11
multi-encoder architecture with hierarchical attention and per-encoder CTC

ruizhilijhu:
I have implemented architecture of multiple parallel encoders with hierarchical attention fusion based on our ICASSP paper. https://arxiv.org/pdf/1811.04903.pdf. [pytorch backend only]

Details:
Implemented multiple encoder setup with test module.
Implemented hierarchical attention network (han).
Implemented per-encoder ctc.

@takaaki-hori , there are changes in terms of decoding, could you take a look at that part?  specially this file: espnet/nets/pytorch_backend/rnn/decoders.py Thanks a lot!


sw005320:
Thanks, @ruizhilijhu, and sorry for my late response.
Can you summarize the benefit of the multiple encoders?
Do you intend to use this for a multiple-input scenario like https://arxiv.org/pdf/1811.04903.pdf ?
I think this architecture could have more wide applications when we tackle multimodal input (e.g., audio and video), and want to make this PR more general extension.
Do you have any idea for this general extension?

ruizhilijhu:
Thanks @sw005320 for the reply. 
This PR of multiple-encoder (multiple-stream) is to:
1. have parallel encoders processing different stream of information, it does not have to be multiple microphone array. Each encoder could take in any sequence that fits into ESPnet input format. 
2. assign each encoder with designated attention mechanism to deal with alignment within each stream. On top of that, a hierarchical attention mechanism is to fuse contextual vectors from all streams. PR is written in the ways that we could use all the available attention modules as either frame-level attention or encoder-level attention.
3. have the function to either assign individual CTCs to each encoder or have a shared CTC across all encoders during training and decoding. 
That would be a great idea to push this PR into more general way. So far, there is no assumption about what kind of input each encoder should take as long as it is a sequential feature that fits ESPnet data format. For example, it should be used immediately for audio-visual following ESPnet data format.

sw005320:
@kan-bayashi and @ShigekiKarita, I think this is a very important and valuable change. With this extension, we can even handle multimodal input (The fusion is performed for the context vector and we can even handle different time units). Can you take a look at it and give @ruizhilijhu some suggestions on the design level? My current concern is that the CTC part is a bit ASR specific, but we can make it as a pure ATT mode by setting the CTC weight zero. Also, module design and name can be considered for more general use.

ci/test_flake8.sh
```diff
@@ -49,6 +49,7 @@ espnet/nets/e2e_asr_common.py
 espnet/nets/mt_interface.py
 espnet/nets/pytorch_backend/ctc.py
 espnet/nets/pytorch_backend/e2e_asr.py
+espnet/nets/pytorch_backend/e2e_asr_mulenc.py
```
ShigekiKarita: We do not want to add any new files here. And this file seems to be almost ready for passing the flake8-docstrings. Remove this line and add a few more docstrings to this file by following flake8-docstrings messages.

ShigekiKarita:
We test `espnet.asr.pytorch_backend.asr` by an integration test with small mock dataset `egs/asr1/mini_an4/asr1/run.sh`. Can you add new config `conf/train_multienc.yaml` to test `--num-encs 2`? CI servers can run the integration test like
```bash

# test asr recipe
(
    cd ./egs/mini_an4/asr1 || exit 1
    . path.sh  # source here to avoid undefined variable errors

    set -euo pipefail

    echo "==== ASR (backend=pytorch) ==="
    ./run.sh
    echo "==== ASR (backend=pytorch num-encs=2) ==="
    ./run.sh --stage 3 --train-config ./conf/train_multienc.yaml
    ...
```
at https://github.com/espnet/espnet/blob/master/ci/test_integration.sh

ci/test_flake8.sh
ruizhilijhu: Thanks, I have removed e2e_asr_mulenc.py from the list. and the file passed flake8-docstrings.

ruizhilijhu:
Thanks a lot @ShigekiKarita for the advice. I have added test script for multi-encoder setup in test_integration.sh.

sw005320:
@kan-bayashi, could you also review this PR?

kan-bayashi:
Oh, sorry I missed.
I will review today.

espnet/asr/asr_utils.py
```diff
@@ -576,3 +655,39 @@ def plot_spectrogram(plt, spec, mode='db', fs=None, frame_shift=None,
...
+# * ------------------ recognition related ------------------ *
+def format_mulenc_args(args):
+    """Format args for multi-encoder setup.
+
+    It deals with following situations:  (when args.num_encs=2):
+    1. args.elayers = None -> args.elayers = [4, 4];
+    2. args.elayers = 4 -> args.elayers = [4, 4];
+    3. args.elayers = [4, 4, 4] -> args.elayers = [4, 4].
+
+    """
```
kan-bayashi: Why don't you add `logging.warn` when this situation is happened?

espnet/asr/pytorch_backend/asr.py
```diff
@@ -693,6 +759,9 @@ def enhance(args):
     # read training config
     idim, odim, train_args = get_model_conf(args.model, args.model_conf)
 
+    # todo (ruizhili) implement enhance for multi-encoder model
```
kan-bayashi:
```suggestion
    # TODO(ruizhili): implement enhance for multi-encoder model
```

test/test_e2e_asr_mulenc.py
```diff
@@ -0,0 +1,407 @@
...
+            xs_list = [xs_list[idx].cuda() for idx in range(num_inputs)]
+            ilens_list = [ilens_list[idx].cuda() for idx in range(num_inputs)]
+            ys = ys.cuda()
+
+    return xs_list, ilens_list, ys
+
+
+@pytest.mark.parametrize(
+    "module, model_dict", [
+        ('espnet.nets.pytorch_backend.e2e_asr_mulenc', {}),
```
kan-bayashi: if this mulenc can work with `num_encs > 2`, it is better to add test case `('espnet.nets.pytorch_backend.e2e_asr_mulenc', {'num_encs': 3}),`

espnet/nets/pytorch_backend/e2e_asr_mulenc.py
```diff
@@ -0,0 +1,601 @@
+#!/usr/bin/env python3
+
+"""Define e2e module for multi-encoder network. https://arxiv.org/pdf/1811.04903.pdf."""
+# Copyright 2017 Johns Hopkins University (Shinji Watanabe)
+# Copyright 2017 Johns Hopkins University (Ruizhi Li)
+#  Apache 2.0  (http://www.apache.org/licenses/LICENSE-2.0)
+
+
+from __future__ import division
```
kan-bayashi: No need to support py2.
```suggestion
```

espnet/nets/pytorch_backend/e2e_asr_mulenc.py
```diff
@@ -0,0 +1,601 @@
...
+import argparse
+import logging
+import math
+import os
+
+import editdistance
+
+import chainer
+import numpy as np
+import six
```
kan-bayashi: ditto
```suggestion
```

espnet/nets/pytorch_backend/e2e_asr_mulenc.py
```diff
@@ -0,0 +1,601 @@
...
+            start, end = n // 4, n // 2
+            bias.data[start:end].fill_(1.)
+
+        lecun_normal_init_parameters(self)
+        # exceptions
+        # embed weight ~ Normal(0, 1)
+        self.dec.embed.weight.data.normal_(0, 1)
+        # forget-bias = 1.0
+        # https://discuss.pytorch.org/t/set-forget-gate-bias-of-lstm/1745
+        for l in six.moves.range(len(self.dec.decoder)):
```
kan-bayashi:
```suggestion
        for l in range(len(self.dec.decoder)):
```

espnet/nets/pytorch_backend/e2e_asr_mulenc.py
```diff
@@ -0,0 +1,601 @@
...
+            set_forget_bias_to_one(self.dec.decoder[l].bias_ih)
+
+    def forward(self, xs_pad_list, ilens_list, ys_pad):
+        """E2E forward.
+
+        :param List xs_pad_list: list of batch (torch.Tensor) of padded input sequences
+                                [(B, Tmax_1, idim), (B, Tmax_2, idim),..]
```
kan-bayashi: In this case, is it correctly working with `DataParallel`?
In my understanding, `DataParallel` split the data according to the first axis, so it seems to cause the error.
If it does not work with `DataParallel`, it is better to add `raise NotImplementedError` in `asr.py` when `num_enc > 1` and add `TODO` comment.

Also, adding the multi gpu case test is great.

espnet/asr/asr_utils.py
ruizhilijhu: Thanks, i have made the changes accordingly.

espnet/asr/pytorch_backend/asr.py
ruizhilijhu: Done!

espnet/nets/pytorch_backend/e2e_asr_mulenc.py
ruizhilijhu: Done

espnet/nets/pytorch_backend/e2e_asr_mulenc.py
ruizhilijhu: Done

espnet/nets/pytorch_backend/e2e_asr_mulenc.py
ruizhilijhu: Done

test/test_e2e_asr_mulenc.py
ruizhilijhu: Thanks, I have added test case for num_encs = 3.

espnet/nets/pytorch_backend/e2e_asr_mulenc.py
ruizhilijhu: Thanks a lot. It is not gonna work with DataParallel for now. I have made a TODO in the asr.py and made changes accordingly.

ShigekiKarita: I think data-parallel support can be delayed. do you agree @kan-bayashi?

kan-bayashi: I agree.

ruizhilijhu: Thanks @kan-bayashi @ShigekiKarita for the reviews and comments. Is there any other suggestions for this PR?

sw005320:
Many thanks, @ruizhilijhu.
I love this direction and will merge this PR, but it may have a lot of issues due to the changes across modules and may ask you to follow up it.

ruizhilijhu:
Thanks @sw005320 , I am glad to contribute on this direction. If there's any issue in the future, I would like to follow up closely.

PR #1204, opened 2019-09-19
support save snapshot by iteration

fanlu:
the epoch num may not  be very large on large dataset training because of time limit. And average_checkpoints is very important to achieve lower CER. So I suspect that it's necessary to support save snapshot when iterations(10000,20000 etc.).
Any idea will be appreciate, thanks

### DIFF
--- a/espnet/asr/asr_utils.py
+++ b/espnet/asr/asr_utils.py
@@ -283,6 +283,19 @@ def torch_snapshot(savefun=torch.save,
 
     return torch_snapshot
 
+def torch_snapshot_iter(savefun=torch.save,
+                   filename='snapshot.iter.{.updater.iteration}'):
+    """Extension to take snapshot of the trainer for pytorch.
+
+    Returns:
+        An extension function.
+
+    """
+    @extension.make_extension(trigger=(1000, 'iteration'), priority=-100)
+    def torch_snapshot(trainer):
+        _torch_snapshot_object(trainer, trainer, filename.format(trainer), savefun)
+
+    return torch_snapshot
 
 def _torch_snapshot_object(trainer, target, filename, savefun):
     # make snapshot_dict dictionary
--- a/espnet/asr/asr_utils.py
+++ b/espnet/asr/asr_utils.py
@@ -283,6 +283,7 @@ def torch_snapshot(savefun=torch.save,
 
     return torch_snapshot
 
+
 def torch_snapshot_iter(savefun=torch.save,
                    filename='snapshot.iter.{.updater.iteration}'):
     """Extension to take snapshot of the trainer for pytorch.
@@ -296,6 +297,7 @@ def torch_snapshot_iter(savefun=torch.save,
         _torch_snapshot_object(trainer, trainer, filename.format(trainer), savefun)
 
     return torch_snapshot
+
 
 def torch_snapshot_iter(savefun=torch.save,
                    filename='snapshot.iter.{.updater.iteration}'):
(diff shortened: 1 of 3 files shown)


espnet/asr/pytorch_backend/asr.py
```diff
@@ -428,11 +429,11 @@ def train(args):
     # we used an empty collate function instead which returns list
     train_iter = {'main': ChainerDataLoader(
         dataset=TransformDataset(train, lambda data: converter([load_tr(data)])),
-        batch_size=1, num_workers=args.n_iter_processes,
+        batch_size=1, num_workers=args.n_iter_processes, pin_memory=True,
```
kan-bayashi: Might be better not to include this PR.

espnet/asr/pytorch_backend/asr.py
```diff
@@ -428,11 +429,11 @@ def train(args):
...
         shuffle=not use_sortagrad, collate_fn=lambda x: x[0])}
     valid_iter = {'main': ChainerDataLoader(
         dataset=TransformDataset(valid, lambda data: converter([load_cv(data)])),
-        batch_size=1, shuffle=False, collate_fn=lambda x: x[0],
+        batch_size=1, pin_memory=True, shuffle=False, collate_fn=lambda x: x[0],
```
kan-bayashi: ditto

espnet/asr/asr_utils.py
```diff
@@ -284,6 +284,21 @@ def torch_snapshot(trainer):
     return torch_snapshot
 
 
+def torch_snapshot_iter(savefun=torch.save,
+                        filename='snapshot.iter.{.updater.iteration}'):
```
kan-bayashi: I think that it is not necessary to make a new function.
Just reuse `torch_snapshot` is fine.

espnet/asr/pytorch_backend/asr.py
```diff
@@ -490,6 +494,8 @@ def train(args):
 
     # save snapshot which contains model and optimizer states
     trainer.extend(torch_snapshot(), trigger=(1, 'epoch'))
+    if args.save_interval_iters > 0:
+        trainer.extend(torch_snapshot_iter(), trigger=(args.save_interval_iters, 'iteration'))
```
kan-bayashi: Why don't you reuse `torch_snapshot` instead of `torch_snapshot_iter`?
```suggestion
        trainer.extend(torch_snapshot(filename='snapshot.iter.{.updater.iteration}'),
                       trigger=(args.save_interval_iters, 'iteration'))
```

espnet/asr/pytorch_backend/asr.py
fanlu: OK. I will delete it

espnet/asr/pytorch_backend/asr.py
```diff
@@ -453,6 +453,9 @@ def train(args):
 
     # Evaluate the model with the test dataset for each epoch
     trainer.extend(CustomEvaluator(model, valid_iter, reporter, device, args.ngpu))
+    if args.save_interval_iters > 0:
```
sw005320: I think it is safe to use `if else` instead of set the CustomEvaluator twice.

espnet/asr/pytorch_backend/asr.py
```diff
@@ -490,6 +493,9 @@ def train(args):
 
     # save snapshot which contains model and optimizer states
     trainer.extend(torch_snapshot(), trigger=(1, 'epoch'))
+    if args.save_interval_iters > 0:
```
sw005320: Same comment. I think it is safe to use `if else` instead of set the torch_snapshot twice.

espnet/asr/pytorch_backend/asr.py
fanlu: OK

sw005320:
Thanks a lot!

PR #1210, opened 2019-09-23
Implementation of TCEN 

CherrieWang97:
https://arxiv.org/pdf/1909.07575.pdf


sw005320:
Thanks again for your great contribution.
I have several high-level thoughts.

1. in general it may take a long time to merge very large PR. It would be great if you split some of your PRs, and we can merge them one-by-one
2. it would be great if you also prepare the example script and showing the performance in README.md and possibly update the model. (but this makes the PR larger. For now, this seems to be fine.)
3. is it possible to still use `st_train.py` and `st_trans.py` and switch to TCEN by changing the configuration and specification of the prior models? (sorry that I do not deeply think of it, but I feel like we may do it).

CherrieWang97:
Thanks for your review. I'll try to split my PR and merge `tcen_train.py` and `tcen_trans.py` into existing scripts.
And I'll release the example script and trained model with the highest performance reported in our paper.

PR #1211, opened 2019-09-23
Fix a bug that <eos> prob is not applied in batch decoding with word LM

takaaki-hori:


sw005320:
Thanks!
This change seems to cause some shape errors.
Can you fix it?

takaaki-hori:
@sw005320 , the code is now ready to merge, which includes some tests for decoding with word LM.

sw005320:
Thanks a lot!

PR #1216, opened 2019-09-26
Improved JSUT TTS recipe

r9y9:
## Samples

- Transformer (phoneme input): https://drive.google.com/open?id=1_aTRb5JIw_JWzf0urV0veC5uwhF1abBa
- Tacotron 2 (phoneme input): https://drive.google.com/open?id=1TZ9MUlzJpxnxgjiAbOe9EHe9PtU2hw4H

My proposal:

1. Stop using neologd for grapheme-to-katakana conversion. Nelogd is not designed for TTS and they have a huge amount of useless entries (e.g., pair of お題 and オダイショウリャクカノウ; this is clearly wrong). I'd like to suggest using OpenJTalk frontend (they use mecab + ipadic internally) as an alternative, which comes with more sophisticated G2P. I made a tool once before https://github.com/r9y9/pyopenjtalk.
2. Downsample audio from 48k to 24k.  I think 24k is good enough for TTS.
3. Trim silences and noises more carefully. I wrote a blog post about trimming silences for JSUT (2 years ago!) https://r9y9.github.io/blog/2017/11/12/jsut_ver1/.

Important changes: 

- [x] Use OpenJTalk instead of mecab (+ nelogd) for G2P.  You can find differences between current/new version of G2P here https://gist.github.com/r9y9/900f65c44c21e518fd8438c8a280d257. Current one fails to estimate pronunciation especially for numbers, but OpenJTalk gives good estimates in most cases.
- [x] Trim silences and noises more carefully (see https://r9y9.github.io/blog/2017/11/12/jsut_ver1/).  I've just released alignment files for JSUT https://github.com/r9y9/jsut-lab. We can use this for trimming silences.
- [x] Downsample audio from 48k to 24k.

Minor changes:

- I'm trying similar feature extraction params as Tacotron. fmin=80, fmax=7600, shift=300, window_length=1200, fft size=2048.Let's see how it works.

TODO:

- [x] Merge tts2 to tts1
- [x] pyopenjtalk installation guide
- [x] Train Transformer
- [x] Train WaveNet vocoder. EDIT: samples will be updated once training is finished.
- [x] Try phoneme input?
- [x] Generate samples

Adding a new recipe as tts2 to experiment with new ideas, but we may consider adding changes to tts1 as well or simply merging them.


kan-bayashi:
This might help you when replace utt_id with segments file.
https://github.com/espnet/espnet/blob/master/egs/m_ailabs/tts1/local/trim_silence.sh

kan-bayashi:
And I think simply merging tts1 is fine.
You do not have to split into tts2.

kan-bayashi:
For resampling, why don’t you follow kaldi tool? I think this is easy.
https://github.com/kaldi-asr/kaldi/blob/master/egs/wsj/s5/utils/data/resample_data_dir.sh
In this case we do not need to modify feature extraction.

r9y9:
Thank you for comments. I'm going to address your comments soon. I will share GL samples for 1) kana input 2) phoneme input later today.

r9y9:
Pushed a fix https://github.com/espnet/espnet/pull/1216/commits/f73e5dbd3ef70d99f08c984f8d91795d6222ac50

If I read https://kaldi-asr.org/doc/data_prep.html correctly, utt_id is same as recording_id in this case, so we don't need to update utt_id, right?

kan-bayashi:
> If I read https://kaldi-asr.org/doc/data_prep.html correctly, utt_id is same as recording_id in this case, so we don't need to update utt_id, right?

Oh that is a great idea. I've never tried, but if it works, it looks more smart way!

r9y9:
Could anyone check the commit https://github.com/espnet/espnet/pull/1216/commits/b91adbc0fbc6f9e087385d78a32177f721cd3d77 if I'm doing right to support phoneme input? Is there anything I'm missing?

r9y9:
Yes, it works in the sense that utils/validate_data_dir.sh doesn't complain :)

kan-bayashi:
> Could anyone check the commit b91adbc if I'm doing right to support phoneme input? Is there anything I'm missing?
 
It seems no problem. But I've never used `phn` type so it is better to check`data.json`.

r9y9:
Finished training transformers. You can compare samples from the links below:

- Transformer, input type "char": https://drive.google.com/open?id=1Ec5TrdvJ_A0PBCjLCaiWvIC72izA9z5K
- Transformer, input type "phn": https://drive.google.com/open?id=1zmZFmX5Vptjzx6WtzCGIdSbM4it0fHjQ
- Transformer, previous one: https://drive.google.com/open?id=1Pav8bhRCZDtrjFEJDkfaPzFDvDRn_4z7

The accent is unnatural as expected (because we don't use accent as additional input), but I don't see big regressions compared with the previous version. Sound quality differs as I downsampled audio and used band-limited features, but this is also expected.

egs/jsut/tts1/run.sh
```diff
@@ -1,8 +1,14 @@
...
 
-# Copyright 2018 Nagoya University (Takenori Yoshimura)
+# Copyright 2018 Nagoya University (Takenori Yoshimura), Ryuichi Yamamoto
 #  Apache 2.0  (http://www.apache.org/licenses/LICENSE-2.0)
 
+# NOTE:
+# To run the script, you will need to manually installl pyopenjtalk.
+# Please see https://github.com/r9y9/pyopenjtalk for how to install the package.
+# Also, you need to install other dependences as follows:
+# pip install nnmnkwii jaconv
```
kan-bayashi: Why don't you move this NOTE into path.sh like:
https://github.com/espnet/espnet/blob/master/egs/chime4/asr1_multich/path.sh

r9y9:
I checked data.json it seems to be okay, and eval samples sound also reasonable.

Removed WIP from the title as I think this is ready to review. Just in case, cc @takenori-y.

r9y9:
Test failure:

```
E                   TypeError: prod() received an invalid combination of arguments - got (out=NoneType, axis=NoneType, ), but expected one of:
E                    * ()
E                    * (torch.dtype dtype)
E                    * (int dim, torch.dtype dtype)
E                         didn't match because some of the keywords were incorrect: out, axis
E                    * (int dim, bool keepdim, torch.dtype dtype)
E                    * (int dim, bool keepdim)
E                         didn't match because some of the keywords were incorrect: out, axis
```

This seems to be caused by the latest version of numpy; and not related to the PR. We will have to look into it in a separete PR.

kan-bayashi:
I will make PR to fix it.

kan-bayashi:
Fixed #1221

egs/jsut/tts1/run.sh
r9y9: Added installation check fo pyopenjtalk.

What should I do for other pip-installable external dependencies? I like the following approaches:

1. Add a comment (as is): you need to run `pip install nnmnkwii jaconv`
2. Add `pip install nnmnkwii jaconv` somewhere. 

To be honest I'm not a big fan of writing shell scripts to check pip-installable dependency... If we can take one of the approaches above, we just only needs a single line, which is very simple (and that's why I like it).

kan-bayashi: One of the advantage to add installation check in `path.sh` is that we can detect the installation necessity before starting the stages. (it failed in `path.sh || exit 1;`)
I think this is more user friendly than it fails at the middle of stages.

In the case of pip installable libraries, we can add them in `setup.py`.
Maybe here what we have to check is `openjtalk` installation, which is not easy to install.
(Pyopenjtalk cannot be installed before the installation of openjtalk, right?)
The best solution is adding like `openjtalk.done` which including all of the installations (openjtalk, or other pip installable libraries). But it is optional.

r9y9: I agree that run.sh shouldn't stop in the middle of stages, but I thought we could live with just notifying users to install some additional packages *before* running the script (e.g. adding a notice to README for example). Anyway, I added `pyopenjtalk.done` in Makefile by following the espnet-style.

r9y9:
Pushed a few more fixes to address comments. Tacotron2 GL samples will be ready later today.

egs/jsut/tts1/run.sh
kan-bayashi: Cool!
Yeah, I agree with you. If we put README in each recipe, that is perfect.

kan-bayashi:
Could you also update RESULT.md by following LJSpeech?
https://github.com/espnet/espnet/blob/master/egs/ljspeech/tts1/RESULTS.md

r9y9:
Sure, let me do this after tacotron training is finished.

egs/jsut/tts1/local/data_prep.sh
```diff
@@ -15,19 +16,11 @@ utt2spk=${data_dir}/utt2spk
...
 [ -e ${rawtext} ] && rm ${rawtext}
-if [ ! -e ${usr_dict_dir} -a ! -e ${esp_dict_dir} ]; then
-    echo "$0: Error: The mecab-ipadic-NEologd can not be found" >&2
-    echo "$0: Error: Please modify the dictionary path in this script or" >&2
-    echo "$0: Error: install it by running 'make mecab.done' at espnet/tools" >&2
-    exit 1
-fi
 
 # make scp, utt2spk, and spk2utt
 find ${db} -name "*.wav" | sort | while read -r filename; do
```
kan-bayashi: To find in symlink, better to use `-follow` option.
```suggestion
find ${db} -follow -name "*.wav" | sort | while read -r filename; do
```

egs/jsut/tts1/local/data_prep.sh
```diff
@@ -42,12 +35,10 @@ echo "finished making wav.scp, utt2spk, spk2utt."
 find ${db} -name "transcript_utf8.txt" | sort | while read -r filename; do
```
kan-bayashi:
```suggestion
find ${db} -follow -name "transcript_utf8.txt" | sort | while read -r filename; do
```

r9y9: Fixed. I didn't know the option. Thanks!

r9y9:
https://drive.google.com/drive/folders/1benxPbnUzRTgqGP6dhb3bZS85OfK4T7Z?usp=sharing Tacotron samples are ready.

I just started training WaveNet for JSUT. Will be finished in a week probably. 

@potato-inoue, Once WaveNet samples are ready, could you help me upload samples to espnet website?

r9y9:
## Samples

- Transformer (phoneme input): https://drive.google.com/open?id=1_aTRb5JIw_JWzf0urV0veC5uwhF1abBa
- Tacotron 2 (phoneme input): https://drive.google.com/open?id=1TZ9MUlzJpxnxgjiAbOe9EHe9PtU2hw4H

Uploaded samples and models. (omitted kana input version of Transformer, but I can upload if someone requested). There's no big differences as expected, but I slightly prefer Transformer ones.

I think this is ready for the second round review (hopefully this will be final). If there's anything I'm missing, please let me know. As for the wavenet vocoder samples, I will make another PR once training is finished.

PR #1220, opened 2019-09-27
Fixed a typo in swbd asr1

Shujian2015:


ShigekiKarita:
Thanks! Because CircleCI failures are fixed in #1221, it is OK.

PR #1221, opened 2019-09-28
Fix e2e_tts test error in CI

kan-bayashi:
Recent numpy changes cause errors when the arguments are tensor.


kan-bayashi:
@ShigekiKarita @r9y9 Travis errors are fixed.

PR #1111, opened 2019-08-20
[WIP] Support Chainer LM with new interface

Fhrozen:
Support Chainer LM with new interface
- [x] Api v1 for new interface (No batchsize, no gpu)
- [ ] Api v2
- [ ] Batchsize support
- [ ] GPU support
- [x] New models (SeqRNN, Transformer)
- [ ] MultiGPU training

PR #1223, opened 2019-09-28
Refact transformer for transformer LM

Fhrozen:
I will be dividing the #1111  into small PR so can make it more easy to review.
I am starting with the refact of transformer to accept transformer LM. 
As soon as it passes the tests and is accepted I will PR the API v.1 fixes.


ShigekiKarita:
LGTM but we need to fix tts CI tests in #1227

PR #1235, opened 2019-10-03
TTS: Update JSUT models

r9y9:
The previous verison didn't contain dict files, so I updated them.

followup of #1216


kan-bayashi:
Thanks a lot!

PR #1236, opened 2019-10-03
JESC recipe + Chainer MT

Fhrozen:
I was just playing with chainer and found a dataset for MT (EN <> JP). 
Adding as a draft in case is required.


sw005320:
This is very cool! @hirofumi0810, could you help it (by reviewing it) when you have time?

sw005320:
Once the recipe part of this PR is fixed, we can test it with both pytorch and chainer backends. I really love to have EN <-> JP (for my personal use).

Fhrozen:
Ohh, I just did it as suggestion, that is why I pushed as draft. (That is why it is too dirty)
There seems to have no problem for training in chainer (already training a model for a couple days).
So, I will try to fixed then for submit, but I will need to split this in two PR (one for the recipe and one for Chainer MT for better revision)

Fhrozen:
Added #1249 instead of this PR. I will be submiting chainer MT after finishing with chainer LM.

Issue #820, opened 2019-06-08
Common run.sh and new style egs/asr2


kamo-naoyuki:
In Kaldi/ESPNet style, each `egs` has different `run.sh` even though they are almost all parts are same. 
I think this system is not good for maintenance because all run.sh need to be modified if some new feature are appended. It also makes noises for reviewing when a new recipe is created. cmd.sh, path.sh and conf/slurm.conf etc. are also enough if one exists.

I'm planning that

1. Create it as `asr2` (I won't touch tts because I'm not familiar with it)
1. Kaldi-free (Maybe, just using librosa-FBANK instead of kaldi-FBANK with pitch)#736
1. Use HDF5 insatead of JSON #533
1. New egs structure

(though much not thinking)

```
egs/base (common?template?shared?)
  - asr2/
     | - steps/
          | - data.sh  # Create format data required by ESPNet e.g. data.json
          | - train_lm.sh
          | - train_asr.sh
          | - decode_asr.sh
     | - cmd.sh
     | - path.sh
     | - conf/
          | - queue.conf
egs/wsj
  - asr1/  # Keep as is
  - asr2/
     |- local/
          | - prepare_data.sh  # Create data/
     | - run.sh  #  Invoke train_lm.sh and train_asr.sh, decode_asr.sh
     | - path.sh  # Source ../base/path.sh and set some task depended requirements
     | - cmd.sh -> ../base/run.sh
     | - conf/
          | - queue.conf -> ../base/conf/queue.conf
```

Sometimes, it is necessary to run the only decoding, e.g. for evaluation for speech enhancement, so independent `decode_asr.sh` may be the clear for such people .

sw005320:
Thanks for your great suggestion. 
These are my comments/thoughts:
- I like your idea of working on such a new recipe procedure in `asr2`.
Once we fix it, we can follow the same style for TTS (@kamo-naoyuki and @takenori-y, please watch this issue carefully).
- For `cmd.sh` and `conf/*`, we can put them in the base directory, but I think it's better not to use symlinks because I'm changing them depending on the recipe (e.g., I added more CPU memory requests in `cmd.sh` or `queue.conf` if we use larger data). Do you have any ideas?
- I like the idea of having `data.sh`. It turns out that we don't have so many variations (BPE or not etc.), and we can modularize it.
[...]

PR #1241, opened 2019-10-07
Asr commonvoice recipe update

ftshijt:
Add recipe for more languages of commonvoice dataset
(default add language model training)

- [ ] update README.md


ShigekiKarita:
`asr2` is reserved for new style recipe e.g., #820. I recommend you to update `egs/commonvoice/asr1`

egs/commonvoice/asr1/local/data_prep.pl
```diff
@@ -46,7 +46,7 @@
   $text =~ tr/a-z/A-Z/;
   print TEXT "$uttId"," ","$text","\n";
   print GNDR "$uttId"," ","$gender","\n";
-  print WAV "$uttId"," sox $db_base/$filepath -t wav -r 16k -b 16 -e signed - |\n";
+  print WAV "$uttId"," ffmpeg -i $db_base/$filepath -f wav -ar 16000 -ab 16 - |\n";
```
ShigekiKarita: For me, using ffmpeg is fine but why did you update this? I just want to know the reason.

ftshijt: We found that sox doesn't support MP3 format because of some legal terms.

ShigekiKarita:
Can you also update the dataset description in https://github.com/espnet/espnet/blob/master/egs/README.md

ftshijt:
> Can you also update the dataset description (e.g., lang EN) in https://github.com/espnet/espnet/blob/master/egs/README.md

Sure~

egs/commonvoice/asr1/local/data_prep.pl
sw005320: The default sox does not support mp3 but it can be done with some plug-in install. `ffmpeg` is a robust solution, but it may not be installed as a default tool. So, could you add an installation check of `ffmpeg` like https://github.com/espnet/espnet/blob/83101a26177500542d1fcff9c22bf7752c2f49ad/egs/librispeech/asr1/local/data_prep.sh#L16-L20?

egs/commonvoice/asr1/local/filter_text.py
```diff
@@ -0,0 +1,35 @@
...
+from __future__ import print_function
+from __future__ import unicode_literals
+
+import argparse
+import codecs
+from io import open
+import sys
+
+
+PY2 = sys.version_info[0] == 2
```
sw005320: We stopped to support python 2. So you don't need this python 2 related lines.

egs/commonvoice/asr1/run.sh
```diff
@@ -55,23 +53,33 @@ set -e
...
-        local/data_prep.pl ${datadir}/ cv-${part} data/"$(echo ${part} | tr - _)"
+        local/data_prep.pl ${datadir} validated data/"$(echo ${part} | tr - _)"
     done
+
+    # Kaldi Version Split
+    # ./utils/subset_data_dir_tr_cv.sh data/validated data/valid_train data/valid_test_dev
+    # ./utils/subset_data_dir_tr_cv.sh --cv-spk-percent 50 data/valid_test_dev data/valid_test data/valid_dev
+
+    # ESPNet Version
+    # consider duplicated sentences
```
sw005320: You can also state that this does not consider the speaker split and this way of split is used for the voxforge recipe.

ftshijt: For train-dev-test split,
If we use the official test and dev set and use the rest of the validated set as training set as we discussed this morning, we may encounter the same problem as last week (duplicated word sequence in test and train). That will be a problem for some languages (e.g. welsh) as they will get a very small WER with 10s hours of training data. What do you think about it?

sw005320: how about removing the duplicated word sequence appeared in the dev/test set from the train set?

ftshijt: It is a possible solution, I will have a try~ Thanks

sw005320:
@YosukeHiguchi, can you give us your comments? This is an extension of your recipe.

sw005320:
Is there speaker information? Is it possible to also exclude test set speakers from the training data (I'm just asking the possibility. You don't have to do it, but this would be better if we still have enough amount of training data).

ftshijt:
> Is there speaker information? Is it possible to also exclude test set speakers from the training data (I'm just asking the possibility. You don't have to do it, but this would be better if we still have enough amount of training data).

There is speaker information. I will test it next.

ftshijt:
> Is there speaker information? Is it possible to also exclude test set speakers from the training data (I'm just asking the possibility. You don't have to do it, but this would be better if we still have enough amount of training data).

ftshijt:
> > Is there speaker information? Is it possible to also exclude test set speakers from the training data (I'm just asking the possibility. You don't have to do it, but this would be better if we still have enough amount of training data).
> 
> There is speaker information. I will test it next.

Unfortunately, it may not be possible to exclude the speakers in the validated set from the test set. I just tested some languages including both large dataset (e.g. de, fr) and small dataset (e.g. cy, tt). The test set has shown to include the majority of the whole speakers (3/5 or more).

sw005320:
OK.
Can you summarize it in a table to record our investigation? (the number of utterances for training/dev/test for each language w/ and w/o considering speaker overlaps)

ftshijt:
> OK.
> Can you summarize it in a table to record our investigation? (the number of utterances for training/dev/test for each language w/ and w/o considering speaker overlaps)


  | SPEAKER |   | UTTERANCE AFTER FILTER
-- | -- | -- | --
LANG | TOTAL_SPEAK | TEST/DEV_SPEAKER | VALIDATE_SPEAKER( /wo test&dev | TOTAL UTTERANCE | UTTERANCE (AFTER SPEAKER FILTER) | UTTERANCE (AFTER TEXT FILTER) | UTTERANCE (AFTER TEXT&SPEECH FILTER)
de | 4852 | 2911 | 1941 | 281207 | 257293 | 58814 | 58814
fr | 2903 | 2466 | 437 | 152362 | 124392 | 77356 | 77356
cy | 725 | 286 | 439 | 34188 | 32362 | 9019 | 9019
tt | 125 | 123 | 2 | 22083 | 15639 | 11859 | 11859
kab | 549 | 455 | 94 | 180582 | 156136 | 53452 | 53452
ca | 1778 | 1543 | 235 | 84818 | 63602 | 36231 | 36127
it | 577 | 549 | 28 | 27062 | 27062 | 9831 | 9831
fa | 1216 | 863 | 353 | 54050 | 47503 | 16341 | 16341
eu | 450 | 365 | 85 | 32962 | 26499 | 11160 | 10926
es | 585 | 546 | 39 | 23763 | 15424 | 6951 | 6930
ru | 61 | 58 | 3 | 19255 | 5718 | 7110 | 2805

ftshijt:
As discussed last Thursday, we use our own test&dev split just like voxforge dataset. If everything goes right, I will start to run for the recipe README

ftshijt:
As for speaker split, it can be found from the previous chart that the speech donation is very imbalance.  For example, the utterances of two speakers in the Tatar(tt) dataset take over around 70% of the total validation data while the other 123 speakers donate the other 30%.

sw005320:
@ftshijt, what is your current status? You don't need to finish all the experiments. I'll merge it if you confirm that it would work with your training/test split.

sw005320:
When you're ready, please edit to remove `[wip]` in the PR title.

ftshijt:
It's ready, I've checked two langs. They are working well

egs/commonvoice/asr1/local/filter_text.py
sw005320: Could you reflect on my comments?
As I mentioned, we'd not support python 2 and you can make your code very simple without these PY2 cases.

egs/commonvoice/asr1/run.sh
```diff
@@ -55,23 +53,34 @@ set -e
...
         # use underscore-separated names in data directories.
-        local/data_prep.pl ${datadir}/ cv-${part} data/"$(echo ${part} | tr - _)"
+        local/data_prep.pl ${datadir} ${part} data/"$(echo "${part}_${lang}" | tr - _)"
     done
+
+    # Kaldi Version Split
+    # ./utils/subset_data_dir_tr_cv.sh data/validated data/valid_train data/valid_test_dev
+    # ./utils/subset_data_dir_tr_cv.sh --cv-spk-percent 50 data/valid_test_dev data/valid_test data/valid_dev
+
+    # ESPNet Version (same as voxforge)
```
sw005320: This comment is very good.

sw005320:
Thanks!
Also, please add your results to `RESULTS.md` after you get them.

PR #1256, opened 2019-10-10
Indirect call  for training method [chainer]

Fhrozen:
I was wondering about the limitation of the training methods (Converter, updater...), because asr.py call it directly, so it is now rewritten in order to be call through the model. This is also helpful in case of a new e2e_foo.py because it only required to be declared in the model script and does not modify the asr.py script directly.

### DIFF
--- a/espnet/nets/chainer_backend/asr_interface.py
+++ b/espnet/nets/chainer_backend/asr_interface.py
@@ -1,9 +1,10 @@
 """ASR Interface module."""
+import chainer
 
 from espnet.nets.asr_interface import ASRInterface
 
 
-class ChainerASRInterface(ASRInterface):
+class ChainerASRInterface(ASRInterface, chainer.Chain):
     """ASR Interface for ESPnet model implementation."""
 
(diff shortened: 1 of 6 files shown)


sw005320:
I like this change (make sense for me), but it corresponds to refactoring and I want to ask @kan-bayashi and @ShigekiKarita's opinion.

ShigekiKarita:
Because this is related to chainer only, create new interface like ChainerASRInterface(ASRInterface)

espnet/nets/chainer_backend/asr_interface.py
```diff
@@ -0,0 +1,22 @@
+"""ASR Interface module."""
+
+from espnet.nets.asr_interface import ASRInterface
+
+
+class ChainerASRInterface(ASRInterface):
```
ShigekiKarita: Maybe this is a little bit better?
```suggestion
class ChainerASRInterface(ASRInterface, chainer.Chain):
```

PR #1270, opened 2019-10-13
Support batch decoding for streaming E2E

takenori-y:
I slightly modified SegmentStreamingE2E class to support batch decoding. This significantly speed up streaming-based decoding.

### DIFF
--- a/README.md
+++ b/README.md
@@ -402,7 +402,7 @@ You can recognize speech in a WAV file using pretrained models.
 Go to a recipe directory and run `utils/recog_wav.sh` as follows:
 ```sh
 cd egs/tedlium2/asr1
-../../../utils/recog_wav.sh --models tedlium2.rnn.v1 example.wav
+../../../utils/recog_wav.sh --models tedlium2.transformer.v1 example.wav
 ```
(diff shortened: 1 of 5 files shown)


sw005320:
Can you keep both batch and non-batch decoding functions with a test script to pass both functions?
We are now maintaining both functions.

takenori-y:
I modified a test script. Should I pass the codecov checks?

kan-bayashi:
No need.

takenori-y:
I uploaded a new model as an example of batch decoding and modified usage in `recog_wav.sh`.

sw005320:
LGTM!
Thanks a lot!

qmpzzpmq:
@takenori-y 
hi, but three is check for 
`SegmentStreamingE2E works only with batch size <= 1`
so, I am not sure is it support batch decoding literally?

sw005320:
We use the term `batch decoding` for two cases, one for across hypotheses and the other for across utterances. This PR is intended for the first case, and we got a significant speed improvement thanks to the across hypotheses batch decoding.

PR #1274, opened 2019-10-15
Chainer LM api v1.0

Fhrozen:
Implemented:

- LM decoding with api v1
- Train multiGPU
- Added models seq_rnn & transformer


Fhrozen:
@ShigekiKarita, I have a concern about the Scorer and BeamSearch. 
An idea was to merge pytorch and chainer in the decoding, but recently looking at the dtypes, I am not sure if this will be possible.
Chainer is moslty numpy in the CPU so it can accept FP16 in CPU without problem, but Pytorch seems to have no implemented this function. Also, in chainer, the mixed16 is available and in case of pytorch  the type is O#... 
It will be nice to merge both backends, but, should decoding be implemented in FP32 so it will be possible to mixed both backends or it is better to keep them separately so can take advantage of FP16 and Mixed16 in chainer.

sw005320:
@Fhrozen, can you revisit this and #1249?
I'm very sorry that I could not make time for the project.
I was very busy these days but now I'm getting to have some more time.

PR #1306, opened 2019-10-24
Add eval-interval-epochs for the tiny dataset

kan-bayashi:
In the case of a tiny dataset, the number of iterations in an epoch is small and then the evaluation interval is too short.
Too frequent evaluation is not necessary and it requires too much time.
This PR adds `eval-interval-epochs` to decide how frequent evaluation will be performed.
This enables us to avoid this issue by setting eval-interval-epochs to large number e.g. 10 or 100.


sw005320:
Very good idea!
Thanks!

Issue #1301, opened 2019-10-23
import librosa get OSError


kaituoxu:
Hi, after I pass `make check_install`, I got below import error when I tried to train models. After `apt-get install libsndfile1-dev`, this error gone. Maybe need to fix this problem in related install scripts. Thx.

```
>>> import librosa
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/home/Project/espnet/tools/venv/lib/python3.7/site-packages/librosa/__init__.py", line 12, in <module>
    from . import core
  File "/home/Project/espnet/tools/venv/lib/python3.7/site-packages/librosa/core/__init__.py", line 124, in <module>
    from .audio import *  # pylint: disable=wildcard-import
  File "/home/Project/espnet/tools/venv/lib/python3.7/site-packages/librosa/core/audio.py", line 10, in <module>
    import soundfile as sf
  File "/home/Project/espnet/tools/venv/lib/python3.7/site-packages/soundfile.py", line 142, in <module>
    raise OSError('sndfile library not found')
OSError: sndfile library not found
```

kan-bayashi:
Thank you for your report.
We will add the library as a pre-requirement in README.

PR #1309, opened 2019-10-25
Update README.md

kan-bayashi:
- Added a new paper citation
- Added dependency of libsndfile (#1301)
- Removed chainer and pytorch version because they are not pre-requirements.


sw005320:
LGTM!

PR #1299, opened 2019-10-22
Update CSMSC results

kan-bayashi:
- Added csmsc transformer + wavenet samples
- Added csmsc fastspeech + wavenet samples

CSMSC FastSpeech sounds really good.
https://drive.google.com/open?id=1LLrT-Dbk7Sov75NURZTAOlNoxyTqwLQb

PR #1314, opened 2019-10-26
Fixed pretrained model URL in CSMSC reicpe

kan-bayashi:
Fixed https://github.com/espnet/espnet/pull/1299#discussion_r339293252


r9y9:
Thanks! Could you also add train_fastspeech.v3.single.yaml?

PR #1315, opened 2019-10-28
Fixed error: "Expected object of device type cuda but got device type cpu" in decoder of transducer

rai4:
when decoding of transducer, occurred "Expected object of device type cuda but got device type cpu" error in decoders_transducer.py. 
Fixed error.

### DIFF
--- a/espnet/nets/pytorch_backend/rnn/decoders_transducer.py
+++ b/espnet/nets/pytorch_backend/rnn/decoders_transducer.py
@@ -200,7 +200,7 @@ class DecoderRNNT(torch.nn.Module):
 
         """
         z_list, c_list = self.zero_state(h.unsqueeze(0))
-        ey = torch.zeros((1, self.embed_dim))
+        ey = to_device(self, torch.zeros((1, self.embed_dim)))
 
         hyp = {'score': 0.0, 'yseq': [self.blank]}
 
@@ -214,7 +214,7 @@ class DecoderRNNT(torch.nn.Module):
                 hyp['yseq'].append(int(pred))
                 hyp['score'] += float(logp)
 
-                eys = torch.full((1, 1), hyp['yseq'][-1], dtype=torch.long)
+                eys = to_device(self, torch.full((1, 1), hyp['yseq'][-1], dtype=torch.long))
                 ey = self.dropout_embed(self.embed(eys))
 
                 y, (z_list, c_list) = self.rnn_forward(ey[0], (z_list, c_list))
@@ -239,7 +239,7 @@ class DecoderRNNT(torch.nn.Module):
         normscore = recog_args.score_norm_transducer
 
         z_list, c_list = self.zero_state(h.unsqueeze(0))
-        eys = torch.zeros((1, self.embed_dim))
+        eys = to_device(self, torch.zeros((1, self.embed_dim)))
 
         _, (z_list, c_list) = self.rnn_forward(eys, None)
 


b-flo:
Hi,

Could you give me an example to reproduce the error please?
I didn't encounter any error during training with either CPU and GPU and all test are OK for transducer. Decoding is only with CPU

rai4:
Hi,
Thank you. Sorry for not explaining it in detail. An error occurs when GPU decoding. 
I tried GPU decoding because CPU takes a long time. Also i tried Greedy Search(recognize() function not recognize_beam()). After modifying, GPU decoding works quickly well. 
If there is no specific problem, how about applying GPU decoding? Also if possible, i would like to participate in GPU decoding problem. 
I am training Korean corpus using rnn-transducer of espnet.

b-flo:
> I tried GPU decoding because CPU takes a long time. Also i tried Greedy Search(recognize() function not recognize_beam())

Yeah decoding takes too much time... I'm looking for some speed improvement right now for streaming.
Btw, could you also compare CPU/GPU beam search please? You also modified the function.

> If there is no specific problem, how about applying GPU decoding?

Yeah sure! As I said, I can't access my notes right now and I don't remember the problem why I _disabled_ GPU decoding, so if everything works like that it's great! Thanks!

sw005320:
Thanks!
The test seems to be failed due to errors in openslr to get some files. 
I just closed and reopened this PR to make travis ci run.

Can someone know how to rerun circleci?

@b-flo, can you review this PR?
The PR looks good for me.

rai4:
b-flo Hi,

> Btw, could you also compare CPU/GPU beam search please? You also modified the function.

Yeah sure! I compared CPU/GPU beam search. The choice of using CPU or GPU depends on the ngpu setting in the script. I also added an option for GPU beam search to work.
The result is faster CPU. (Perhaps the reason for blocking GPU search?) But on greedy search, GPU is much faster. So I think you should be able to choose GPU. 
As a result, CPU/GPU greedy and beam search work very well. If possible, I would like to make a pull request to implement the operation options of GPU beam search.(https://github.com/rai4/espnet/commit/04c0fb6138c332fbef8fa1a62252cfc248031cdc)

As you know, there is no implementation that increments the decoding batch size in espnet. (recognize_batch()). For this part I'm going to start implementing.

b-flo:
> The result is faster CPU. (Perhaps the reason for blocking GPU search?) But on greedy search, GPU is much faster. So I think you should be able to choose GPU.

Seems fair to me!

> As you know, there is no implementation that increments the decoding batch size in e2e_asr_transducer.py of espnet. (recognize_batch() in e2e_asr_transducer.py). For this part I'm going to start implementing.

I didn't add batch decoding for now because I couldn't : 1) speed-up beam search with batch decoding, 2) adopt the same approach as in JCA to adopt the new scoring interface (because of the joint procedure and vertical transitions). Feel free to work on it :)

espnet/nets/pytorch_backend/e2e_asr_transducer.py
```diff
@@ -134,6 +134,7 @@ def __init__(self, idim, odim, args):
         self.space = args.sym_space
         self.blank = args.sym_blank
         self.reporter = Reporter()
+        self.beam_size = args.beam_size
```
b-flo: You don't need to handle greedy/beam search selection here. It is already handled in asr and asrinterface modules : 

https://github.com/espnet/espnet/blob/ee0c82d6e74659bd174c90a261135e0901f8b39b/espnet/asr/pytorch_backend/asr.py#L688-L749

https://github.com/espnet/espnet/blob/ee0c82d6e74659bd174c90a261135e0901f8b39b/espnet/nets/asr_interface.py#L66-76

espnet/nets/pytorch_backend/e2e_asr_transducer.py
```diff
@@ -282,7 +283,11 @@ def forward(self, xs_pad, ilens, ys_pad):
             batch_nbest = []
 
             for b in six.moves.range(batchsize):
-                nbest_hyps = self.dec.recognize_beam(hs_pad[b], self.recog_args)
+                if self.beam_size == 1:
```
b-flo: ditto

espnet/nets/pytorch_backend/e2e_asr_transducer.py
rai4: Sorry for not explaining it in detail. This code is used to calculate cer and wer for each epoch when training. In other words, the train was slow because only beam search was supported. 
I want to use greedy search when calculating cer / wer on training.(not inference: asr_recog.py) 
So i think we need this. This error also happens in models(e2e_asr_transducer.py) that don't have the recognize_batch function when decoding (inference) after the train ends, setting the decode batch size to 1 or more.(ex. voxforge corpus, in decode_transducer.yaml, batch: 1)

espnet/nets/pytorch_backend/e2e_asr_transducer.py
rai4: ditto. This code applies only to training.

espnet/nets/pytorch_backend/e2e_asr_transducer.py
b-flo: Yes sorry, I mistook beamsize for batchsize! I made the review in a rush and I realized it after. Discard my review :)

For adding greedy decode option for cer/wer computation during training I don't see any problem. However you talk about an "error", it should not raise any error, it only not supported yet, right?

b-flo: Yes sorry, I mistook beamsize for batchsize! I made the review in a rush and I realized it after. Discard my comments!
For adding greedy decode option for cer/wer computation during training I don't see any problem. However you talk about an "error", it should not raise any error, it only not supported yet, right?

b-flo:
@rai4 I wanted to change the reviews to resolved but it seems my comments were deleted instead and it made duplicated as your answers. You can delete yours or put the resolve tag I guess?

espnet/nets/pytorch_backend/e2e_asr_transducer.py
rai4: Yes, that's right. it's only not supported yet, not error.

rai4:
@b-flo I already deleted my comments

sw005320:
Thanks @rai4 and @b-flo

rai4:
.

PR #1323, opened 2019-10-30
disabling nonlinear activation of the last encoder layer

simpleoier:


simpleoier:
I didn't update the results of an4, because my numbers are worse than the ones in `RESULT`.

Issue #1291, opened 2019-10-18
Attention weight of Transformer TTS is wrong in inference the cache


kan-bayashi:
We need to calculate the attention weight incrementally.

https://github.com/espnet/espnet/blob/03724e6577f7dc8e56c1db3e7b07551a307b3131/espnet/nets/pytorch_backend/e2e_tts_transformer.py#L692-L697

kan-bayashi:
Fixed in #1331

PR #1331, opened 2019-11-02
Fix TTS transformer attention weight calculation in inference

kan-bayashi:
This PR fixes a bug of transformer attention weight calculation in inference (#1291)


sw005320:
Thanks!

PR #1333, opened 2019-11-04
Support Parallel WaveGAN

kan-bayashi:
This PR allows using parallel wavegan in `synth_wav.sh`.


sw005320:
It's better to add some information for each vocoder (e.g., repository URL and brief descriptions).

kan-bayashi:
Yes, the mode training will be finished soon.
I will update README.

kan-bayashi:
Done.
I decided to keep old models in the demo scripts.

utils/synth_wav.sh
```diff
@@ -127,8 +136,12 @@ function download_vocoder_models () {
     case "${vocoder_models}" in
         "ljspeech.wavenet.softmax.ns.v1") share_url="https://drive.google.com/open?id=1eA1VcRS9jzFa-DovyTgJLQ_jmwOLIi8L";;
         "ljspeech.wavenet.mol.v1") share_url="https://drive.google.com/open?id=1sY7gEUg39QaO1szuN62-Llst9TrFno2t";;
```
r9y9: Did you forget to add`ljspeech.wavenet.mol.v2` perhaps?

kan-bayashi: Here I decided to list only the no limit ljspeech models not to be confused.
Instead of here, I made the table of all of the pretrained models in `README.md`. How do you think?

sw005320:
Can we add the language information in the table?

kan-bayashi:
I reformated the tables.
Now the information became much clearer than before.

sw005320:
great!
thanks!

Issue #1334, opened 2019-11-06
Parser in e2e_asr_transducer is missing a return statment


ahmedalbahnasawy:
In the code: https://github.com/espnet/espnet/blob/master/espnet/nets/pytorch_backend/e2e_asr_transducer.py

You're not returning parser in `add_arguments`. This throws an error when attempting to call `E2E.build()`

sw005320:
Thanks for the report.
@b-flo, could you confirm it?

b-flo:
Hi,

Thank you for the report!
I can confirm `parser` isn't returned as it should be. However before making any change, could you copy the error thrown please? I'm extensively using RNNT/RNNT-Att with arguments specific to the implementation and I did not encountered any errors. I'm a bit intrigued on why it wasn't reported/catched before.

rai4:
User classes are considered mutable. Python doesn't have (absolutely) private attributes, so you can always change a class by reaching into the internals. So the parser variable is considered a call by reference. I also did not encountered any errors.

```
from espnet.utils.dynamic_import import dynamic_import
    if args.model_module is None:
        model_module = "espnet.nets." + args.backend + "_backend.e2e_asr:E2E"
    else:
        model_module = args.model_module
    model_class = dynamic_import(model_module)
    model_class.add_arguments(parser)

    args = parser.parse_args(cmd_args)
    args.model_module = model_module
```

rai4:
User classes are considered mutable. Python doesn't have (absolutely) private attributes, so you can always change a class by reaching into the internals. So the parser variable is considered a call by reference. I also did not encountered any errors.

```
from espnet.utils.dynamic_import import dynamic_import
    if args.model_module is None:
        model_module = "espnet.nets." + args.backend + "_backend.e2e_asr:E2E"
    else:
        model_module = args.model_module
    model_class = dynamic_import(model_module)
    model_class.add_arguments(parser)

    args = parser.parse_args(cmd_args)
    args.model_module = model_module
```

b-flo:
> Python doesn't have (absolutely) private attributes, so you can always change a class by reaching into the internals. So the parser variable is considered a call by reference

That's why I suspected but I wanted to be sure given the error! Thanks for the info!

ahmedalbahnasawy:
Thanks for taking the time to review my issue.
[...]

PR #1337, opened 2019-11-06
Return parser in add_arguments method for transducer

b-flo:
This PR add a  `return parser` in `add_argument` method for transducer. There is no change in terms of use/behaviour within ESPNET but it will fix a problem explained in #1334. It is also more consistent with other modules.


sw005320:
Thanks @b-flo for your quick response!!!

PR #1341, opened 2019-11-08
Pre-trained CSJ model 

takenori-y:


sw005320:
Thanks!!!

sw005320:
> Thanks!
> Could you also update the result in README?

Yes, that is helpful. It would be great if you also update it.

PR #1344, opened 2019-11-09
Update libri_trans asr recipe

hirofumi0810:
- Change RNN to Transformer
- Change character unit to BPE1k unit

### DIFF
--- a/egs/libri_trans/asr1/conf/tuning/train_pytorch_transformer_bpe_long.yaml
+++ b/egs/libri_trans/asr1/conf/tuning/train_pytorch_transformer_bpe_long.yaml
@@ -16,7 +16,7 @@ mtlalpha: 0.3
 lsm-weight: 0.1
 
 # minibatch related
-batch-size: 64  # for BPE
+batch-size: 64
 maxlen-in: 512  # if input length  > maxlen-in, batchsize is automatically reduced
 maxlen-out: 150 # if output length > maxlen-out, batchsize is automatically reduced
 
@@ -26,7 +26,7 @@ opt: noam
 accum-grad: 2
 grad-clip: 5
 patience: 0
-epochs: 100
+epochs: 120
 dropout-rate: 0.1
 
 # transformer specific setting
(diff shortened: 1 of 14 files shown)


hirofumi0810:
@sw005320 I will update results with SpecAugment tomorrow.

hirofumi0810:
@sw005320 Ready to be merged.

egs/libri_trans/asr1/conf/tuning/train_pytorch_transformer_bpe_long.yaml
```diff
@@ -0,0 +1,44 @@
...
+maxlen-in: 512  # if input length  > maxlen-in, batchsize is automatically reduced
+maxlen-out: 150 # if output length > maxlen-out, batchsize is automatically reduced
+
+# optimization related
+sortagrad: 0 # Feed samples from shortest to longest ; -1: enabled for all epochs, 0: disabled, other: enabled for 'other' epochs
+opt: noam
+accum-grad: 2
+grad-clip: 5
+patience: 0
+epochs: 120
```
sw005320: The only difference is the number of epochs? Then, we could just rename `train_pytorch_transformer_bpe_long.yaml ` as `train_pytorch_transformer_bpe.yaml` and replace it.

hirofumi0810: I'd like to keep as they are in order to keep the training time for the baseline without SpacAugment.

sw005320: Oh, now I got it. This longer version is for the specaug, right? Then, maybe better to add some comments in `run.sh`? Another suggestion is to use specaug as default in `run.sh`.

hirofumi0810: I don't wanna use SpecAugment as the default setting. I added a comment in `run.sh`. Also, readers can easily follow it by looking at `RESULTS.md`.

sw005320: ok. thanks!

sw005320:
LGTM

PR #1366, opened 2019-11-16
Fix sphinx warning of TTS modules

kan-bayashi:
This PR fixed Sphnix warning of TTS modules.

### DIFF
--- a/doc/conf.py
+++ b/doc/conf.py
@@ -38,6 +38,7 @@ extensions = [
     'sphinx.ext.napoleon',
     'sphinx.ext.viewcode',
     "sphinx.ext.mathjax",
+    "sphinx.ext.todo",
     "sphinxarg.ext",
 ]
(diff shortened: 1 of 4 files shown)


doc/conf.py
```diff
@@ -38,6 +38,7 @@
     'sphinx.ext.napoleon',
     'sphinx.ext.viewcode',
     "sphinx.ext.mathjax",
+    "sphinx.ext.todo",
```
sw005320: What's this?
Will it automatically generate TODO?

kan-bayashi: This will support Todo section in docstring.
https://www.sphinx-doc.org/en/master/usage/extensions/todo.html
Without this extension sphinx raises warning for todo section.

sw005320: I see.
Thanks!

PR #1368, opened 2019-11-18
Fix ASR eval for TTS in the case of trans_type=phn

kan-bayashi:
This PR fixes a bug in the ASR eval for TTS in the case of `trans_type=phn`.

The ASR evaluation script assumed that the `text` is the character sequence but recently we support `trans_type=phn`. This causes the wrong evaluation results.
This PR fixes the issue by re-creating the `text`.


kan-bayashi:
Could you check it? @potato-inoue

potato-inoue:
I have checked the codes.
I think there is no problem.

Issue #1230, opened 2019-09-29
remove Chainer dependence maybe good for user


zh794390558:
Since most people only using pytroch and tensorflow, the  Chainer dependence is not good for user reading code. So remove Chainer or separate Chainer with pytroch is prefer.

kan-bayashi:
@ShigekiKarita is experimentally working on this issue #688
But we have to discuss the interface. (See also #1188)

zh794390558:
great work.

Issue #1188, opened 2019-09-09
OOP Task Trainer


ShigekiKarita:
Recently, ESPnet has a lot of duplicated lines in `def train(args): ...` of ASR, ASR-MIX, LM, MT, TTS, ST, etc. It is hard to maintain and update because lots of complicated things (custom fp, data loader, transfer learning) are copy-and-pasted in the huge GOD functions. I wanna unify those duplicated parts. How about unifying these with extensible OOP-based design?

```python
# task-dependent part
class TaskInterface:
    @staticmethod
    def add_arguments(parser):

    def init_dataset(self, ...):

    def init_model(self, ...):

    def init_criterion(self, ...):

    def init_sampler(self, ...):


# task-independent training loop
class TaskRunner:
    def train(self, task: TaskInterface):

    def eval(self, task: TaskInterface):

    def run(self):
```

see also: https://github.com/pytorch/fairseq/blob/master/fairseq/tasks/fairseq_task.py

ShigekiKarita:
I will rework chainer-free trainer #688  soon but we need to discuss this first.

kan-bayashi:
I agree with your idea but I'm wondaring that whether we can remove duplicated lines if we use this OOP design.

ShigekiKarita:
Yeah, I'm also afraid that copying fairseq is not enough for us. My goal is
not just splitting task-dependent and task-independent parts because the
task-dependent part still can be duplicated.

(I believe that TaskRunner is just a common part of existing Trainer and
Evaluator in all the tasks)

However, this idea can make relationship between tasks explicitly (instead
of randomly importing symbols). For example, ASR mix class can be derived
from ASR. And ST class can be derived from ASR and MT.

2019年9月9日(月) 21:28 Tomoki Hayashi <notifications@github.com>:

gs5264:
Are you still going on with the plan?

gs5264:
This will be very helpful for distributed training of pytorch.

PR #998, opened 2019-07-30
WIP: Add support for on-the-fly text cleanup for TTS

r9y9:
This will allow us to use more flexible input types and on-the-fly data augumentation. 

## Summary

- **Before**: text cleanup is done before making data.json
- **Proposed**: text cleanup is done while makinig minibatches (note: previous style is also supported)

## Motivation

- [x] On-the-fly char-to-phoneme substitution (see sec. 3.2 in DeepVoice3 https://arxiv.org/abs/1710.07654)
- On-the-fly char-phoneme representation mixing (https://arxiv.org/abs/1811.07240) (can be done in the next PR)

## Remarkable changes

- Add `espnet.tts.frontend` module, which defines TTS frontend interface and languege-dependent frontend (for now "en" and "text" are only available. we may add "ja" and more later). TTS frontend is responsible to convert input text to its numeric representation.
- Change `LoadInputsAndTargets` to use TTS frontend if specified. `tts_decode.py` and `tts_train.py` now accept `--frontend` argment.
- Change ljspeech/tts1 recipe to use on-the-fly cleanup. Note that previous recipes work as is.

###  Basic usage of TTS frontend

```py
In [1]: from espnet.tts import frontend

In [2]: en = frontend.get_frontend("en")

# tokenization
In [3]: en.text_to_sequence("hello world")
Out[3]: [35, 32, 39, 39, 42, 64, 50, 42, 45, 39, 31, 60, 1]

# de-tokenization (for convenience)
In [4]: en.sequence_to_text(en.text_to_sequence("hello world"))
Out[4]: 'hello world.~'
```

#### char

```py
In [1]: from espnet.tts.frontend import get_frontend

In [2]: fe = get_frontend("en", mix_prob=0.0)

In [3]: fe.text_to_sequence("hello world")
Out[3]: [35, 32, 39, 39, 42, 64, 50, 42, 45, 39, 31, 60, 1]

In [4]: fe.sequence_to_text(fe.text_to_sequence("hello world"))
Out[4]: 'hello world.~'
```

#### phoneme

```py
In [5]: fe = get_frontend("en", mix_prob=1.0)

In [6]: fe.sequence_to_text(fe.text_to_sequence("hello world"))
Out[6]: '{HH AH0 L OW1} .{W ER1 L D}~'
```

#### Joint representation of char and phoneme (same as DeepVoice3)

```py
In [7]: fe = get_frontend("en", mix_prob=0.5)

[...]

PR #1372, opened 2019-11-19
[Draft] [Nor for merged] [WIP] ESPNet2

### LINKED ISSUE #533: mergejson.py out of memory

kamo-naoyuki:
- I implemented a class named as `Task` to provide a common framework for our DNN training.
    - You can consider it as a replacement of the current 
`espnet/bin/*`, `espnet/asr/*/asr.py:train()`, and the trainer part of chainer.
    - The all tasks we have, ASR, LM, TTS, ST, and MT, could be unified by this framework.
    - The name of `Task` comes from `FairseqTask`, but any dependencies on fairseq doesn't exist.
- I put this new framework and the example in [espnet2](https://github.com/kamo-naoyuki/espnet/tree/espnet2/espnet2) and [egs2](https://github.com/kamo-naoyuki/espnet/tree/espnet2/egs2)
- I dropped Chainer options entirely from this new framework.
   - See also: https://preferred.jp/en/news/pr20191205/
- This PR is just changing of the trainer and `E2E` class, which is responsible for the definition of `loss`, so I believe that **the core part of network structure. e.g. transformer, tacotoron, etc. could be used as they are** with small modification to fit the new interface. 
- This is the first draft and all items are NOT fixed. Welcome any comments.

## Related issues and PR

- Trainer: #1188
- Kaldi-free: Not using any kaldi-binary-tools in data preparation part #736
- Chainer-free: Independent from Chainer's trainer #686 #1230
- Discarding Json format, at least for training corpus, to resolve #533
- More flexible IO system: #998, #1012
- And many refactoring  #820, #890, #942

### Simple example for new Task

See: 

- [espnet2/bin](https://github.com/kamo-naoyuki/espnet/tree/espnet2/espnet2/bin)
- [espnet2/tasks](https://github.com/kamo-naoyuki/espnet/tree/espnet2/espnet2/tasks)
- [espnet2/asr/model.py](https://github.com/kamo-naoyuki/espnet/tree/espnet2/espnet2/asr/model.py)

```python
# espnet2.tasks.foo
class TypeA:
    def __init__(self, a: int=3): pass

class TypeB:
    def __init__(self, b: int=4): pass

class YourModel(torch.nn.Module):
    # This is almost equivalent class to the current "E2E" class.
    def __init(self, bar, e2e_param_a = 3, e2e_param_b = 4.2): pass
    def forward(x, input, input_lengths, output, output_lengths):
        ...
        # Model is responsible to return loss, and the other stats.
        return loss, stats, weight

class FooTask(BaseTask):
    # Task class is not allowed to have any instance-method. 
    @classmethod
    def add_arguments(cls, parser):
        parser.add_arguments('--bar', choices=['TypeA', 'TypeB'])
        parser.add_arguments('--bar_conf', action=NetstedDictAction)
        parser.add_arguments('--e2e_conf', action=NetstedDictAction)
        return parser

    @classmethod
    def get_bar_class(cls, name: str):
        if name.lower() == 'typea':
            return TypeA
        elif name.lower() == 'typeb':
            return TypeB

    @classmethod
    def get_default_config(cls):
        # This method is used for --print_config
        config = BaseTask.get_default_config()
        ...
        bar_class = cls.get_bar_class(args.bar)
        bar_conf = get_default_kwargs(bar_class)
        e2e_conf = get_default_kwargs(YourE2E)
        config['bar_conf'] = bar_conf
        config['e2e_conf'] = e2e_conf
        return config
    
    @classmethod
    def build_model(cls, args):
        bar_class = cls.get_bar_class(args.bar)
        bar = bar_class(**args.bar_conf)
        return YourModel(bar=bar, **e2e_conf)
``` 

```python
# espnet2.bin.foo
from espnet2.tasks.foo import FooTask
if __name__ == '__main__':
    FooTask.main()
```

Only each tasks have to do is `Task:build_model()` and the other all parts,
data-loading, training, evaluation, reporting, resuming, and etc. are inherited from `BaseTask`. 

I carefully designed `Task` class as not to be too abstracted for our requirements.
ESPNet has many tasks, but the differences are a few as the aspect of Trainer.
I also keep `Task` as possible as simple and easy. I believe you can easily understand this new framework.

### Command line interface

```bash
# The common options for training are added by BaseTask
% python -m espnet2.bin.foo --optimizer adam --batch-size 256 ...
```

The most important feature is that you don't need explicitly 
to set options using `argparse.add_arguments()` 
for changeable class, `TypeA` and `TypeB`, because they are nested in `--bar_conf`.

```
# To give params to nested dicts
% python -m espnet2.bin.foo \
   --bar_conf a=10
   --optim_conf lr=0.1
```

You might worry how you can get to know what options exist because the nested options can't be shown by `--help`.
For this problem, I also implemented `--print_config` option.

```
# The default arguments of "TypeA" is automatically gathered ant printed there.
% python -m espnet2.bin.foo --print_config
bar: TypeA
bar_conf:
    a: 3
e2e_conf:
    e2e_param_a: 3
    e2e_param_b: 4.2
optim: adam
optim_conf:
    lr: 0.001
    betas:
    - 0.9
    - 0.999
    eps: 1.0e-08
    weight_decay: 0
    amsgrad: false
...

# You can also bind command-line-arguments to the yaml
# and "--bar_conf" is changed according to "--bar" 
% python -m espnet2.bin.foo --print_config --bar TypeB --optim adadelta
bar: TypeB
bar_conf:
    b: 4
e2e_conf:
    e2e_param_a: 3
    e2e_param_b: 4.2
optim: adadelta
optim_conf:
    lr: 1.0
    rho: 0.9
    eps: 1.0e-06
    weight_decay: 0
...

# To use configuration file
% python -m espnet2.bin.foo --print_config > config.yaml
% python -m espnet2.bin.foo --config config.yaml
```

### Data-IO: The flow of command-line args => mini-batch Iterator => DNN-Model
`BaseTask` also provides a common system of data-loading for each tasks.
I think this system could be used for almost all tasks,
thus `Task` class doesn't have the interface to define own iterator, such as `FooTask:build_iterator()`.

```bash
# --train_data_path_and_name_and_type can be repeated as you need and 
# it indicates inputting multiple features for your model.
% python -m  espnet2.bin.foo \
    --train_data_path_and_name_and_type=/some/path/wav.scp,input,sound \
    --train_data_path_and_name_and_type=/some/path/token_int,output,text_int
```
The above command indicates the following mini-batch.

```python
# The first element, /some/path/wav.scp, indicates the file path.
# The second element, input, specifies the dict-key of mini-batch. Any names can be used.
# The third element, sound, tells how to load the file.
dataset = ESPNetDataset([('/some/path/wav.scp','input','sound'), 
                         ('/some/path/token_int','output','text_int')])
data_loader = DataLoader(dataset=dataset, ...)
for batch in data_loader:
    # batch looks like {'input': tensor, 'input_lengths': tensor, 
    #                   'output': tensor, 'output_lengths': tensor}
    # Note that the sequence data are padded and concatenated to a Tensor,
    # so tensors named as `*_lengths` are automatically inserted to tell the original length.
    loss, stats, weight = model(**batch)
```

Note that you can freely add the loader class for your desired `type` to modify [espnet2/train/dataset.py](https://github.com/kamo-naoyuki/espnet/tree/espnet2/espnet2/train/dataset.py),
so this system is enough extensible.
Only the restriction is that the file-format must have key-value structure e.g. 

- kaldi-scp-like text
```
uttid_a /some/path/foo
uttid_b /some/path/bar
```
- HDF5
- NPZ

## TODO
- Trainer
    - [x] Resume
    - [x] Reporter
    - BatchSampler
        - [x] Constant-mini-batch and "seq" mode 
        - [ ] Batch-bin and Batch-frame sampler (not my todo)
    - [x] The system of scheduler: just using pytorch scheduler
        - [x] Noam scheduler
    - [ ] Distributed training: Very easy.
    - [x] Averaging model files
    -  Quality of life 
        - Managements of experiments, just like sacred
            - [ ] stdout log & its management
            - [ ] Better config file management
            - [ ] Dump system info 
        - [ ] Tensorboard support (not my todo)
        - [x] Matplotlib-based graph plotting
        - [x] Draw attention
- ASR task
    - [x] Transformer
    - [x] Frontend part: STFT, Fbank, etc.
    - [x] RNN
    - [ ] RNN-T
    - [ ] Multi-encoder
    - [x] Recognition tool based on v2 
    - recipe:
        - [x] Voxforge
        - [x] WSJ recipe
- LM task
    - [ ] Transformer (not my todo)
    - [x] RNN 
    - [x] A tool to calculate perplexity
- TTS task (not my todo)
    - [x] Tacotron2
    - [ ] Transformer
    - [ ] FastSpeech
    - [x] Decode tool
- [ ] MT task (not my todo)
- [ ] ST task (not my todo)
- Test codes
    - [ ] asr/
    - [ ] bin/
    - [x] layers/
    - [ ] lm/
    - [x] optimizers/
    - [x] schedulers/
    - [ ] tasks/
    - [ ] train/
    - [ ] tts/
    - [x] utils/


Fhrozen:
@kamo-naoyuki, Great work. Looks like a lot of work you done here.

I know is just a draft but I am wondering about a couple of things. I observed that there is no option backend so I supposed that the model is the one restricted to the backend. However looking at the task https://github.com/kamo-naoyuki/espnet/blob/espnet2/espnet2/tasks/asr.py. I observed that  it imports CTC from pytorch_backend and ASR from espnet2, both implemented for pytorch. So, it is that the new version will discard the use of chainer and will only focus on pytorch, or is just a current setup bc of the draft condition of this PR.

kamo-naoyuki:
(I inserted comment about chainer.)

I don't have the plan to add chainer option, but the idea of `Task` can be easily adopted for chainer, and in that case, it's enougth just to implement `ChainerBaseTask`, which should be independent from the `BaseTask` for pytorch.

Additional to say, I positively used  pytorch' `DataLoader` for this new framework and it is designed as loosely-coupled and I also keep new classes as loosely-coupled, e.g. Reporter, 
so that is not so hard work.

However in the current situation, I think we don't need such Task class for Chainer-backend because:

- As far as chainer's usage,  I think Chainer's Trainer is not bad. 
- The primary reason why we need new framework is we are using chainer's Trainer for pytorch backend.
- Many task, TTS, MT, ST, and asr-tranceducer, has already dropped the chainer options and the reason why we need **common** framework is that there are many tasks and also, there are more duplication codes than necessary, though they have only small difference between each other.

gs5264:
good work

kamo-naoyuki:
@Fhrozen 
> Also, I am wondering about the use of sacred for managing the experiments. I found it very useful. Is there any plan to use it?.

I have also used sacred and I agree that it is useful especially about the managing of experiments.
but, currently  I don't have the plan of sacred, too.

Sacred also provides, an unique command line interface and it conflicts the argparse based design. I decided to use argparse because it is simple and enough mature, because it is official python module. Really, several bugs of sacred bothered me and also I noticed that the design of sacred itself has inevitable issues. However, about this matter, I'm not sure I was right.

sw005320:
> @Fhrozen
> 
> > Also, I am wondering about the use of sacred for managing the experiments. I found it very useful. Is there any plan to use it?.
> 
> I have also used sacred and I agree that it is useful especially about the managing of experiments.
> but, currently I don't have the plan of sacred, too.
> 
> Sacred also provides, an unique command line interface and it conflicts the argparse based design. I decided to use argparse because it is simple and enough mature, because it is official python module. Really, several bugs of sacred bothered me and also I noticed that the design of sacred itself has inevitable issues. However, about this matter, I'm not sure I was right.

I also prefer to go with `argparse` due to the popularity and official python module.

bobchennan:
I also like sacred but it is hard to combine with argparse:
https://github.com/IDSIA/sacred/issues/164

There are some other alternates which might be friendly to combine with argparser:
https://old.reddit.com/r/MachineLearning/comments/bx0apm/d_how_do_you_manage_your_machine_learning/

egs2/mini_an4/asr1/pyscripts/feats/compute-cmvn-stats.py
```diff
@@ -0,0 +1 @@
+../../../../../utils/compute-cmvn-stats.py
```
sw005320: do you need this?
the current `path.sh` set a path in `utils`.

espnet2/train/abs_e2e.py
```diff
@@ -0,0 +1,36 @@
+from abc import ABC, abstractmethod
+from typing import Tuple, Dict
+
+import torch
+
+
+class AbsE2E(torch.nn.Module, ABC):
+    """The common abstract class among each tasks
+
+    "E22" is referred to as a class which inherits torch.nn.Module,
```
sw005320:
```suggestion
    "E2E" is referred to as a class which, inherits torch.nn.Module,
```
???

espnet2/tasks/lm.py
```diff
@@ -0,0 +1,195 @@
...
+import logging
+from typing import Any, Dict, Type, Tuple, Optional, Sequence, Callable
+
+import configargparse
+import numpy as np
+import torch
+from typeguard import check_return_type, check_argument_types
+
+from espnet2.lm.abs_model import AbsLM
+from espnet2.lm.e2e import LanguageE2E
```
sw005320: LanguageE2E sounds a bit strange for me.
It might be too later, but `e2e` does not sometimes sound correct.

kamo-naoyuki: It's not too later because class names and file names  can be easily renamed if using PyCharm, so I roughly name them always.

E2E is not good for LM training, I agree, but I don't have the other idea in my poor vocabulary.
Could you suggest any one?

The role of this class:

 - Delegate: Consists of  some internal blocks. e.g. Encoder, Decoder for ASR, and just a model for LM training.
 - Criterion: Define loss and return it to the Training part.
 - Mediator object between task depended structures and the common region. In other words, any functions in the common training part can't touch any task depended objects except for this class.

My candidates:

- LossModel
- ModelController
- ModelComposer
- ModelMaster
- ESPNetModel -> LanguageESPNetModel, ASRESPNetModel, ...
- Classifier
- E2E

E2E may be good if limiting for ESPNet user because this class has the similar role to the current E2E, but for language model, it sounds strange.

egs2/mini_an4/asr1/pyscripts/feats/compute-cmvn-stats.py
kamo-naoyuki: I didn't decide my mind about which scripts are placed under the `egs2/*asr1/` and which commands should invoked directly using `${PATH}`.

At least, shell scripts should be put under the `egs/` because they can't be used in the other space due to dependencies on `path.sh`, `utils`, etc.

sw005320:
Is there any Chainer related part? (I guess not).
By considering the Chainer situation, we may gradually remove Chainer related parts and only consider PyTorch as our backend.

kamo-naoyuki:
> Is there any Chainer related part? (I guess not).
> By considering the Chainer situation, we may gradually remove Chainer related parts and only consider PyTorch as our backend.

No, there isn't. Now espnet2 can performs without chainer entirely.

kamo-naoyuki:
Although there are many unfinished items, but I think now is good time to be used by some main contributors.

- I think, of course, the design of Task is not perfect yet, e.g. it can't support GAN obviously. 
   - If we will decide officially tackle un/semi supervised training such as ASR-TTS, then we need more high level abstracted system. I have some ideas, but it's not time to consider it.
- TTS part has bugs yet, maybe due to my simple mistaking . I haven't investigated the reason deeply.  Please look, @kan-bayashi 
- Some class names should be reconsidered, e.g. E2E, but it's O.K as the draft.
- I also implemented on-the-fly text processing at training time, including BPE encoding, but it's not used in egs2/ because If we use on-the-fly text processing at training time, then we couldn't get the length of tokens at the initial time of the training. I have some ideas to solve this, but it take more time to implement as better way.
   - Similar problem is also existing, if we use on-the-fly feature extracting, then we couldn't global statistics of mean and value of features.

sw005320:
I have merged this PR for further development.

PR #1388, opened 2019-11-26
Fix gradient accumlation

hirofumi0810:
In the current codebase, evaluation with the validation set happens only when the epoch is even.
In other words, the checkpoints in the odd epoch are skipped.
This PR forces the model to turn the gradient accumulation off when the last minibatch in the current epoch is given.


ShigekiKarita:
LGTM. Can you add comments (even just this PR number/URL is fine)?

Issue #1360, opened 2019-11-15
TTS Tacotron 2, "Weak" Alignment, any suggestions?


Imtinan1996:
So i am trying to train Tacotron 2 on some custom dataset, i have a single speaker dataset, that is roughly around 11 hours. 

I have trained other implementations of tacotron 1 before, and on this one implementation the alignment learnt was very good.

ESPNet though for some reason learns alignment, but its a bit "weak" meaning at every timestamp the predicted phonemes are a bit off sometimes.  

I am training this on 6 GPUs with batch size 32. I trained libritts as well from the default recipes, and according the the config, the model learnt very good alignment in only 30 epochs. But as can be seen from the GIF below, the alignment does get better, but it takes 800 epochs, and still its somewhat weak.

Can anyone give suggestions on what the problem could be, or what i could do to make things better, any help would be greatly appreciated.

EDIT:::

To not mess with the specifications alot, i have only changed the sampling rate in the config to match my datas SR, all other parameters like n_mels, nfft, etc remain the same. Could this be an issue? Should i resample my data to 24000 to match libritts specifications and try training the model again? My params are as follows

```
fs=16000      # sampling frequency
fmax=""       # maximum frequency
fmin=""       # minimum frequency
n_mels=80     # number of mel basis
n_fft=1024    # number of fft points
n_shift=256   # number of shift points
win_length="" # window length
 
```

Additionally this is what my melspectograms look like from feats.scp and feats.ark 

random feats.scp
![feat.scp](https://user-images.githubusercontent.com/30050508/68930536-0780c080-07a8-11ea-8709-32d8d10a690c.png)

random feats.ark
![feats.ark](https://user-images.githubusercontent.com/30050508/68930697-5dedff00-07a8-11ea-9d85-12f1a6d79dbf.png)

This is very different from what libritts features looked like, but my limited knowledge in signal processing does not help me identify what the problem could be

![Alignment spochs 280-870](https://user-images.githubusercontent.com/30050508/68928650-077ec180-07a4-11ea-886b-1babfc37106d.gif)

sw005320:
My suggestion is to use a single GPU. We may not need to use 6 GPUs for such small data. ESPnet multiple GPU training is based on a simple data parallelization, and your method corresponds to the use of 6x32=192 batch size, which is too large for this data. Then, according to the general online-training theory, you require a lot of iterations to get some reasonable convergence.

kan-bayashi:
[...]

PR #1397, opened 2019-11-28
Support length-weighted normalization in loss calculation

kan-bayashi:
This PR supports length-weighted normalization in loss calculation.
From #1360, it seems that the length weighted normalized loss is effective to get clear diagonal attention.

### Now
loss (B, T, D) -> average over all of the bins -> loss scalar

### length-weighted normalization
loss (B, T, D) -> average along with T -> loss (B, D) -> average over remaining bins -> loss scalar


kan-bayashi:
I added results. 
From this result, I also included ASR eval results.
This is nice to compare, I think.
It is worthwhile to try the dataset where there are various length samples.
In the case of the LJspeech, the results are almost the same between v3 and v4.

kan-bayashi:
@r9y9 Could you check it?
I think not so important option but it is worthwhile to try as we can see in #1360.

espnet/nets/pytorch_backend/e2e_tts_tacotron2.py
```diff
@@ -160,17 +160,26 @@ def _make_masks(ilens, olens):
...
     """Loss function module for Tacotron2."""
 
-    def __init__(self, use_masking=True, bce_pos_weight=20.0):
+    def __init__(self, use_masking=True, use_weighted_masking=False, bce_pos_weight=20.0):
         """Initialize Tactoron2 loss module.
 
         Args:
-            use_masking (bool): Whether to mask padded part in loss calculation.
+            use_masking (bool): Whether to apply masking for padded part in loss calculation.
+            use_weighted_masking (bool): Whether to applyg weighted for maskng padded part in loss calculation.
```
r9y9: Same as above

espnet/nets/pytorch_backend/e2e_tts_transformer.py
```diff
@@ -13,6 +13,7 @@
 
 from espnet.nets.pytorch_backend.e2e_asr_transformer import subsequent_mask
 from espnet.nets.pytorch_backend.e2e_tts_tacotron2 import GuidedAttentionLoss
+from espnet.nets.pytorch_backend.e2e_tts_tacotron2 import Tacotron2Loss as TransformerLoss
```
r9y9: Oh :)

We could consider naming it a more clear name that denotes what it actually does. Not an issue of the PR though.

espnet/nets/pytorch_backend/e2e_tts_transformer.py
```diff
@@ -325,7 +277,8 @@ def __init__(self, idim, odim, args=None):
                 - eprenet_dropout_rate (float): Dropout rate in encoder prenet.
                 - dprenet_dropout_rate (float): Dropout rate in decoder prenet.
                 - postnet_dropout_rate (float): Dropout rate in postnet.
-                - use_masking (bool): Whether to use masking in calculation of loss.
+                - use_masking (bool): Whether to apply masking for padded part in loss calculation.
+                - use_weighted_masking (bool): Whether to apply weighted masking for padded part in loss calculation.
```
r9y9: Same as the above; a bit confusing. Weighted mask (`w`) is applied for the non-padded part as well (0 is for the padded part) in my understanding.

kan-bayashi: Right. I removed `for padded part`.

kan-bayashi:
Added all of the ASR results.

PR #1409, opened 2019-12-04
Docker Updates

Fhrozen:
Finish to build, test and uploading the containers with fixes for new installation.

@sw005320 , BTW I am wondering if there is gonna be a soon support for pytorch 1.1, 
CUDA 10.1 cannot be used with pytorch 1.0.1 so I cannot add this container yet.

### DIFF
--- a/docker/build.sh
+++ b/docker/build.sh
@@ -51,20 +51,32 @@ cmd_usage() {
 build(){
     echo "Build docker containers"
     # build runtime and gpu based containers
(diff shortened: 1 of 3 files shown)


sw005320:
Thanks a lot, @Fhrozen!!!

> @sw005320 , BTW I am wondering if there is gonna be a soon support for pytorch 1.1,
CUDA 10.1 cannot be used with pytorch 1.0.1 so I cannot add this container yet.

Sure, we can exclude pytorch 1.0.1 and support pytorch 1.1.
Now, we also support pytorch 1.3 and this would not be an issue.

PR #1421, opened 2019-12-09
Fix a compatibility problem with PyTorch 1.3.0 in ESPnet (v0.6.0)

Emrys365:
I was running the wsj recipe with PyTorch 1.3.0 and the latest ESPnet.

And I found that some `loss=inf` warnings appeared due to some "short" samples (After subsampling, its length will be shorter than that of the corresponding label, which leads to `loss_ctc = inf`).

So in stage 2 (Data Preparation), I add a script to filer out such samples from the original `data.json` to avoid the warnings.
> It is a pseudo training script that uses the same configuration as the real one.
> It runs through one epoch with `batchsize=1` to detect the "short" samples and remove them from `data.json`. (in-place)
> This process takes about 0.5 ~ 1 hour(s).

BTW, I guess [asr.py#L199](https://github.com/espnet/espnet/blob/HEAD/espnet/asr/pytorch_backend/asr.py#L199) should be an in-place operation. I also fix it in this PR.


sw005320:
Is it possible to remove such samples during or after the data preparation stage? At least we could do it for the JSON file because it has enough information.

espnet/asr/pytorch_backend/asr.py
```diff
@@ -196,7 +196,7 @@ def update_core(self):
         if self.grad_noise:
             from espnet.asr.asr_utils import add_gradient_noise
             add_gradient_noise(self.model, self.iteration, duration=100, eta=1.0, scale_factor=0.55)
-        loss.detach()  # Truncate the graph
+        loss.detach_()  # Truncate the graph
```
ShigekiKarita: I do not think this line is needed because `loss` is not referenced below.

ShigekiKarita:
We can modify https://github.com/espnet/espnet/blob/master/utils/remove_longshortdata.sh (e.g., `--min-io-ratio 4`) to check filter samples of `ilen / olen < minratio` instead of that script.

Emrys365:
> Is it possible to remove such samples during or after the data preparation stage? At least we could do it for the JSON file because it has enough information.

Sorry I don't understand the question. The script I added is just used in the data preparation stage, and it modifies the `data.json` file by removing "short" samples.

Emrys365:
> We can modify https://github.com/espnet/espnet/blob/master/utils/remove_longshortdata.sh (e.g., `--min-io-ratio 4`) to filter samples of `ilen / olen < minratio` instead of that script.

Yes, this could be a simple solution when no preprocessing is used. Because in such cases, the input length is only changed by subsampling.

sw005320:
> > Is it possible to remove such samples during or after the data preparation stage? At least we could do it for the JSON file because it has enough information.
> 
> Sorry I don't understand the question. The script I added is just used in the data preparation stage, and it modifies the `data.json` file by removing "short" samples.

I meant to say that this is done with some bash based scripting during the data preparation stage, and @ShigekiKarita's comment (https://github.com/espnet/espnet/pull/1421#issuecomment-563220401) is more concrete.

espnet/asr/pytorch_backend/asr.py
Emrys365: Yes, you are right.

Emrys365:
> > > Is it possible to remove such samples during or after the data preparation stage? At least we could do it for the JSON file because it has enough information.
> > 
> > 
> > Sorry I don't understand the question. The script I added is just used in the data preparation stage, and it modifies the `data.json` file by removing "short" samples.
> 
> I meant to say that this is done with some bash based scripting during the data preparation stage, and @ShigekiKarita's comment ([#1421 (comment)](https://github.com/espnet/espnet/pull/1421#issuecomment-563220401)) is more concrete.

I see. So we simply parse the value of `subsampling` like in [train_rnn.yaml#L7](https://github.com/espnet/espnet/blob/master/egs/wsj/asr1/conf/tuning/train_rnn.yaml#L7), and then remove the samples that have an IO ratio lower than 4 (in case of `subsampling: 1_2_2_1_1`)?

sw005320:
(Yes, that's what I explained last time. Sorry that I did not give you more concrete procedures then.)

Emrys365:
Hi @sw005320  and @ShigekiKarita,  I just write a new script to filter out samples with IO ratio smaller than 4.

Interestingly, it can only detect two bad samples: `020c020c` and `020c020u`.
While my previous script will detect one additional sample that will cause `loss_ctc=inf`: `46hc040b`.

The following is the corresponding content in `data.json`:
```json
 ...
        "020c020c": {
            "input": [
                {
                    "feat": "xxx/feats.3.ark:18151651",
                    "name": "input1",
                    "shape": [
                        312,
                        83
                    ]
                }
            ],
            "output": [
                {
                    "name": "target1",
                    "shape": [
                        86,
                        52
                    ],
                    "text": "WE WERE NOT GOING TO BE ABLE TO BID AGAIN AND WE WERE HAPPY WITH THE PRICE WE ACHIEVED",
                    "token": "W E <space> W E R E <space> N O T <space> G O I N G <space> T O <space> B E <space> A B L E <space> T O <space> B I D <space> A G A I N <space> A N D <space> W E <space> W E R E <space> H A P P Y <space> W I T H <space> T H E <space> P R I C E <space> W E <space> A C H I E V E D",
                    "tokenid": "42 24 18 42 24 37 24 18 33 34 39 18 26 34 28 33 26 18 39 34 18 21 24 18 20 21 31 24 18 39 34 18 21 28 23 18 20 26 20 28 33 18 20 33 23 18 42 24 18 42 24 37 24 18 27 20 35 35 44 18 42 28 39 27 18 39 27 24 18 35 37 28 22 24 18 42 24 18 20 22 27 28 24 41 24 23"
                }
            ],
            "utt2spk": "020"
        },
 ...
        "020c020u": {
            "input": [
                {
                    "feat": "xxx/feats.3.ark:18879437",
                    "name": "input1",
                    "shape": [
                        676,
                        83
                    ]
                }
            ],
            "output": [
                {
                    "name": "target1",
                    "shape": [
                        174,
                        52
                    ],
                    "text": "FACED WITH MOUNTING COMPLAINTS THE STATE DEPARTMENT REMOVED FIDELITY FROM ITS LIST OF RECOMMENDED WAREHOUSES IN NINETEEN EIGHTY FIVE AND THE COMPANY DECIDED TO SHUT ITS DOORS",
                    "token": "F A C E D <space> W I T H <space> M O U N T I N G <space> C O M P L A I N T S <space> T H E <space> S T A T E <space> D E P A R T M E N T <space> R E M O V E D <space> F I D E L I T Y <space> F R O M <space> I T S <space> L I S T <space> O F <space> R E C O M M E N D E D <space> W A R E H O U S E S <space> I N <space> N I N E T E E N <space> E I G H T Y <space> F I V E <space> A N D <space> T H E <space> C O M P A N Y <space> D E C I D E D <space> T O <space> S H U T <space> I T S <space> D O O R S",
                    "tokenid": "25 20 22 24 23 18 42 28 39 27 18 32 34 40 33 39 28 33 26 18 22 34 32 35 31 20 28 33 39 38 18 39 27 24 18 38 39 20 39 24 18 23 24 35 20 37 39 32 24 33 39 18 37 24 32 34 41 24 23 18 25 28 23 24 31 28 39 44 18 25 37 34 32 18 28 39 38 18 31 28 38 39 18 34 25 18 37 24 22 34 32 32 24 33 23 24 23 18 42 20 37 24 27 34 40 38 24 38 18 28 33 18 33 28 33 24 39 24 24 33 18 24 28 26 27 39 44 18 25 28 41 24 18 20 33 23 18 39 27 24 18 22 34 32 35 20 33 44 18 23 24 22 28 23 24 23 18 39 34 18 38 27 40 39 18 28 39 38 18 23 34 34 37 38"
                }
            ],
            "utt2spk": "020"
        },
 ...
        "46hc040b": {
            "input": [
                {
                    "feat": "xxx/feats.9.ark:21351408",
                    "name": "input1",
                    "shape": [
                        483,
                        83
                    ]
                }
            ],
            "output": [
                {
                    "name": "target1",
                    "shape": [
                        120,
                        52
                    ],
                    "text": "I'VE NEVER SEEN AN ADJECTIVE USED SO MUCH IN CONNECTION WITH A PERSON'S NAME HE SAID IN A NINETEEN EIGHTY FOUR INTERVIEW",
                    "token": "I ' V E <space> N E V E R <space> S E E N <space> A N <space> A D J E C T I V E <space> U S E D <space> S O <space> M U C H <space> I N <space> C O N N E C T I O N <space> W I T H <space> A <space> P E R S O N ' S <space> N A M E <space> H E <space> S A I D <space> I N <space> A <space> N I N E T E E N <space> E I G H T Y <space> F O U R <space> I N T E R V I E W",
                    "tokenid": "28 5 41 24 18 33 24 41 24 37 18 38 24 24 33 18 20 33 18 20 23 29 24 22 39 28 41 24 18 40 38 24 23 18 38 34 18 32 40 22 27 18 28 33 18 22 34 33 33 24 22 39 28 34 33 18 42 28 39 27 18 20 18 35 24 37 38 34 33 5 38 18 33 20 32 24 18 27 24 18 38 20 28 23 18 28 33 18 20 18 33 28 33 24 39 24 24 33 18 24 28 26 27 39 44 18 25 34 40 37 18 28 33 39 24 37 41 28 24 42"
                }
            ],
            "utt2spk": "46h"
        },
 ...
```

sw005320:
Thanks for the investigation.

There would be three possible reasons.
1. rounding when subsampling
2. use of additional `<eos>` symbol (as far as I remember CTC does not use it but please double check).
3. padding

3 is a little bit difficult to make this procedure as an independent script. It's better to check such utterance when we make minibatches in https://github.com/espnet/espnet/blob/master/espnet/utils/training/batchfy.py

Emrys365:
> There would be three possible reasons.
> 
> 1. rounding when subsampling

This may not be the cause. Because we can see in the above example that, `hlens = ceil(ilens / 4)`. This can also be confirmed in [encoders.py#L181-L187](https://github.com/espnet/espnet/blob/master/espnet/nets/pytorch_backend/rnn/encoders.py#L181-L187).
Therefore, the IO ratio is not smaller than what we expected.

> 2. use of additional `<eos>` symbol (as far as I remember CTC does not use it but please double check).

I test several examples using the builtin CTCLoss in PyTorch 1.3.0, and find that I can easily get `loss=inf` when `(input_length - target_length)` is small (<6).

Here is the example code:
```python
count = 0 
for i in range(100): 
    T = random.randint(50, 200)    # Input sequence length
    C = 40                         # Number of classes (including blank)
    N = 16                         # Batch size
    S = T - 6                      # Target sequence length of longest target in batch
     
    input = torch.randn(T, N, C).log_softmax(2).detach().requires_grad_()
    target = torch.randint(low=1, high=C, size=(N, S), dtype=torch.long)
     
    input_lengths = torch.full(size=(N,), fill_value=T, dtype=torch.long)
    target_lengths = torch.LongTensor([S] * N)
    ctc_loss = nn.CTCLoss()
    loss = ctc_loss(input, target, input_lengths, target_lengths)
    if torch.isinf(loss.detach()):
        count+=1
        #print(f'T={T}, S={S}')
print(f'{count} `inf`s')
```

> 3. padding

Maybe this is not the cause either, because `ys_pad` is unpadded  in [ctc.py#L65](https://github.com/espnet/espnet/blob/master/espnet/nets/pytorch_backend/ctc.py#L65).

sw005320:
Thanks for your detail investigations. I'm not very sure why 6, but if it is deterministic, one solution is to prune the sentences met with this condition.

Emrys365:
> Thanks for your detail investigations. I'm not very sure why 6, but if it is deterministic, one solution is to prune the sentences met with this condition.

Oh 6 is not a hard boundary for infinite loss in the example.
I mean when `len(input) - len(target) = 7`, about 30% of the random samples result in infinite loss.
I mean when `len(input) - len(target) = 6`, about 50% of the random samples result in infinite loss.
When `len(input) - len(target) = 5`, about 60% of the random samples result in infinite loss.
When `len(input) - len(target) = 4`, about 80% of the random samples result in infinite loss.
...

sw005320:
Thanks for your further clarification. This does not seem to be deterministic...  Maybe we just use `len(input) - len(target)` as a hyperparameter and prune several utterances based on the hyperparameter? We can change this value depends on the recipe.

Emrys365:
OK. I just modify my script to detect such samples and remove them from `data.json`.
There will be two parameters: `min_io_ratio` and `min_io_delta`.
Samples satisfying `len(input) - len(output) * min_io_ratio < min_io_delta` will be removed from `data.json`.

> Since it is not easy to access the input and output shape from `feat.scp`, I do not use https://github.com/espnet/espnet/blob/master/utils/remove_longshortdata.sh to implement the filtering process.

egs/wsj/asr1/run.sh
```diff
@@ -134,6 +137,19 @@ if [ ${stage} -le 2 ] && [ ${stop_stage} -ge 2 ]; then
         data2json.sh --feat ${feat_recog_dir}/feats.scp \
             --nlsyms ${nlsyms} data/${rtask} ${dict} > ${feat_recog_dir}/data.json
     done
+
+    ### Filter out short samples which lead to `loss_ctc=inf` during training
+    ###  with the specified configuration.
+    # `subsample` is in ${train_config}
+    min_io_ratio=$(grep -P 'subsample' ${train_config} | grep -Po '[\d_]+' | \
```
sw005320: This is risky because subsample is only available when we use BLSTMP. It is disabled for BLSTM and VGGBLSTM*

Emrys365: Is there a simple way to get around this problem? I am not very sure how the subsampling option will be specified in different cases.
Or do we have to specify the parameter `min_io_ratio` manually?

sw005320: how about copying the code from   https://github.com/espnet/espnet/blob/212e5c54d2be9775a55199c278cd830f745f5317/espnet/nets/pytorch_backend/e2e_asr_mix.py#L178 and get the actual subsampling?
but this is a bit risky and the perfect solution is to make (an executable function) of the above part, import this for all subsampling parts (asr, st, etc.), and call this function inside this function.
thus we can get the actual subsampling from the training config file.

Emrys365:
> how about copying the code from
> 
> https://github.com/espnet/espnet/blob/212e5c54d2be9775a55199c278cd830f745f5317/espnet/nets/pytorch_backend/e2e_asr_mix.py#L178
> 
> and get the actual subsampling?
> but this is a bit risky and the perfect solution is to make (an executable function) of the above part, import this for all subsampling parts (asr, st, etc.), and call this function inside this function.
> thus we can get the actual subsampling from the training config file.

I add a unified function `get_subsample()` in `espnet/nets/pytorch_backend/nets_utils.py` for different architectures (rnn, rnn-t, rnn_mix, rnn_mulenc, transformer) and different tasks (asr, mt, st).
> I'm not sure whether it is okay to change the subsample value to [1] in [espnet/nets/chainer_backend/e2e_asr_transformer.py#L99](https://github.com/espnet/espnet/blob/master/espnet/nets/chainer_backend/e2e_asr_transformer.py#L99).
> But I just did this to make it consistent with other transformer implementations in pytorch backend.

And now `filtering_samples.py` is implemented by using `get_subsample()` to derive the `subsample` argument.

espnet/asr/pytorch_backend/asr.py
```diff
@@ -202,7 +202,6 @@ def update_core(self):
         if self.grad_noise:
             from espnet.asr.asr_utils import add_gradient_noise
             add_gradient_noise(self.model, self.iteration, duration=100, eta=1.0, scale_factor=0.55)
-        loss.detach()  # Truncate the graph
```
sw005320: is it OK?

Emrys365: @ShigekiKarita and I discussed about this before: https://github.com/espnet/espnet/pull/1421#discussion_r355423179

The removed line uses a non-inplace operation and does not make any difference.
It meant to detach the variable `loss`. But it is not used in the rest part of `update_core()`.

sw005320: ok

sw005320:
Thanks!
Please report whether this would not change the performance so much or not.

Emrys365:
Results using filtered data.json for training (word LM is used by default):

+ RNN (using config in [espnet/egs/wsj/asr1/conf/tuning/train_rnn.yaml](https://github.com/espnet/espnet/blob/master/egs/wsj/asr1/conf/tuning/train_rnn.yaml))

> + CER
>
> | dataset | Snt | Wrd | Corr | Sub | Del | Ins | Err | S.Err |
> | ---- | ---- | ---- | ---- | ---- | ---- | ---- | ---- | ---- |
> | decode_test_dev93_decode_rnn_lm_word65000 | 503 | 48634 | 96.7 | 1.6 | 1.7 | 0.7 | 4.0 | 66.8 |
> | decode_test_eval92_decode_rnn_lm_word65000 | 330 | 31739 | 99.1 | 0.4 | 0.4 | 0.2 | 1.1 | 30.0 |
>
> + WER
>
> | dataset | Snt | Wrd | Corr | Sub | Del | Ins | Err | S.Err |
> | ---- | ---- | ---- | ---- | ---- | ---- | ---- | ---- | ---- |
> | decode_test_dev93_decode_rnn_lm_word65000 | 503 | 8234 | 92.4 | 6.6 | 1.0 | 1.2 | 8.8 | 61.2 |
> | decode_test_eval92_decode_rnn_lm_word65000 | 330 | 5353 | 97.8 | 1.9 | 0.3 | 0.3 | 2.5 | 22.7 |

---
+ Transformer (using config in [espnet/egs/wsj/asr1/conf/tuning/train_pytorch_transformer.yaml](https://github.com/espnet/espnet/blob/master/egs/wsj/asr1/conf/tuning/train_pytorch_transformer.yaml))

> + CER
>
> | dataset | Snt | Wrd | Corr | Sub | Del | Ins | Err | S.Err |
> | ---- | ---- | ---- | ---- | ---- | ---- | ---- | ---- | ---- |
> | decode_test_dev93_decode_rnn_lm_word65000 | 503 | 48634 | 97.4 | 1.1 | 1.5 | 0.6 | 3.2 | 62.0 |
> | decode_test_eval92_decode_rnn_lm_word65000 | 330 | 31739 | 99.4 | 0.3 | 0.3 | 0.2 | 0.7 | 27.0 |
>
> + WER
>
> | dataset | Snt | Wrd | Corr | Sub | Del | Ins | Err | S.Err |
> | ---- | ---- | ---- | ---- | ---- | ---- | ---- | ---- | ---- |
> | decode_test_dev93_decode_rnn_lm_word65000 | 503 | 8234 | 93.7 | 5.3 | 0.9 | 1.1 | 7.4 | 55.5 |
> | decode_test_eval92_decode_rnn_lm_word65000 | 330 | 5353 | 98.5 | 1.3 | 0.1 | 0.3 | 1.8 | 19.1 |

sw005320:
Thanks!
The numbers look too good (eval92 WER is 4.X %).
Can you double-check it?

ShigekiKarita:
Because pytorch default initialization has been changed and Transformer uses it, the result should be changed. Can you try other init like `transformer-init: xavier_uniform`

kan-bayashi:
Amazing result! Is it a bug???

Emrys365:
Sorry for my late reply. I just committed the results and models of this experiment to my repository. You can check them here: https://github.com/Emrys365/espnet/tree/super_wsj/egs/wsj/asr1/exp_test

The Transformer models are not uploaded due to the their large file size.

Emrys365:
> Because pytorch default initialization has been changed and Transformer uses it, the result should be changed. Can you try other init like `transformer-init: xavier_uniform`

Thank you for the information. I am trying with `xavier_uniform` initialization in Transformer now.

Emrys365:
> Because pytorch default initialization has been changed and Transformer uses it, the result should be changed. Can you try other init like `transformer-init: xavier_uniform`

Results of Transformer with `xavier_uniform` initialization are very similar to the above ones (https://github.com/espnet/espnet/pull/1421#issuecomment-572672842).

+ CER

| dataset | Snt | Wrd | Corr | Sub | Del | Ins | Err | S.Err |
| ---- | ---- | ---- | ---- | ---- | ---- | ---- | ---- | ---- |
| decode_test_dev93_decode_rnn_lm_word65000 | 503 | 48634 | 96.8 | 1.3 | 1.9 | 0.7 | 3.9 | 65.0 |
| decode_test_eval92_decode_rnn_lm_word65000 | 330 | 31739 | 99.2 | 0.3 | 0.5 | 0.2 | 1.0 | 28.5 |

+ WER

| dataset | Snt | Wrd | Corr | Sub | Del | Ins | Err | S.Err |
| ---- | ---- | ---- | ---- | ---- | ---- | ---- | ---- | ---- |
| decode_test_dev93_decode_rnn_lm_word65000 | 503 | 8234 | 92.7 | 6.1 | 1.1 | 1.3 | 8.5 | 59.4 |
| decode_test_eval92_decode_rnn_lm_word65000 | 330 | 5353 | 98.1 | 1.6 | 0.3 | 0.3 | 2.3 | 21.2 |

Issue #1384, opened 2019-11-25
Toward fast decoding 


sw005320:
We're planning to work on the following items:

1. incorporate [batch decoding](https://github.com/espnet/espnet/blob/master/espnet/nets/pytorch_backend/rnn/decoders.py#L497) into our new [beam search interface](https://github.com/espnet/espnet/blob/master/espnet/nets/beam_search.py) (api_v2)
   - [x] batch beam search v2 for encoder-decoder, insertion penalty in transformer #1402
   - [ ] batch beam search v2 for CTC in transformer: wrap [CTCPrefixScoreTH](https://github.com/espnet/espnet/blob/6b4cb6aa6765b982cbf5b313c33f9334cca6048e/espnet/nets/ctc_prefix_score.py#L12) in [BatchScorerInterface](https://github.com/espnet/espnet/blob/6b4cb6aa6765b982cbf5b313c33f9334cca6048e/espnet/nets/scorer_interface.py#L83)
   - [ ] batch beam search v2 for RNN decoder: wrap [Decoder.recognize_beam_batch](https://github.com/espnet/espnet/blob/6b4cb6aa6765b982cbf5b313c33f9334cca6048e/espnet/nets/pytorch_backend/rnn/decoders.py#L632)
2. implement other scorers
    - [x] RNNLM, Transformer LM #1402
    - [x] N-gram LM #1946 #1992
    - [ ] Coverage
    - [ ] Multihead decoder
- The beauty of api_v2 is that we can combine all of them by adding a scorer class into the beam search algorithm with corresponding weights (see https://github.com/espnet/espnet/blob/e5bcce8d846efa65a963bbe49c7244ddc48f8143/espnet/nets/beam_search.py#L45-L48)
3. unify to use api_v2 for all other tasks including (probably we don't need additional implementation for some of them below).
    - [ ] asr_mix
    - [ ] mt
    - [ ] st
    - etc.? 
- Then, we should stop to have a beam search method for each decoder
4. RNNT is a different story as it is an input synchronous beam search. We may need to extend the beam search interface. Also, on-line processing would be another different story.

This change would perform the parallelization across both utterances and hypotheses and GPU computation becomes very fast (as we observed in batch decoding). Of course, we can expect significant speedups in CPU as well.
After item 1, we could distribute the implementation efforts to the others, and we may ask some of you.

mowayao:
Any plan for hot-word decoding?

sw005320:
[...]

PR #1422, opened 2019-12-09
Transducer v3 (w/ transformer support for encoder/decoder)

b-flo:
Hi,

This PR add primary modules to train and decode self-attention transducer (or transformer-transducer) models: https://arxiv.org/pdf/1910.12977v1.pdf / https://arxiv.org/pdf/1909.13037.pdf.
It's basically the same implementation as default transducer with self-attention blocks instead of RNNs. Some parts of the RNN transducer implementation were modified to be used in s-a transducer and all transducer parts were moved to a new folder `nets/pytorch_backend/transducer`.

~Also, I tried to implement it following transformer implementation however right now it's more a mix between old (RNN transducer) and new (Transformer) interface/implementations...~
_Edit: Unified `e2e_asr_transducer.py` is similar to `e2e_asr_transformer.py` now._

---

Needed:
 - Results for validation and recipes. I only have <= ~~12Mb~~ 8Mb (sigh) GPUs available right now so I can't do large scale testing. It would be great if someone run some recipes with dataset equal or larger than voxforge and configs similar to the default transformer ones.
I could only test on small datasets, validation loss seemed suspicious but results were comparable to transformer with the same configs (something may be wrong or missing though).
 - Cleaning, corrections and optimizations depending on previous item and reviews.
- ~Tests: to be added right after the completion of previous items.~
_Edit: RNN-Transducer and Transformer-Transducer tests are separated. I'll make an unified version soon to include mixed architecture (RNN - Transformer) testing._

Pending: 
 - Report CER/WER : Should I add another `ErrorCalculator` for transducer models or add report in `e2e_as_sa_transducer.py` similarly to `e2e_asr_transducer.py` for now?
_Edit: For testing and clarity purpose I chose the second option for now. I duplicated most of the methods pending reviews._
 - Batch decoding :  #1384
 - Unified transformer/RNN transducer implementation with encoder and decoder selection option (motivated by https://arxiv.org/pdf/1910.12977v1.pdf reports): as discussed on gitter, my initial attempt to unify implementations ended up with a broken trainer. It should be revisited according to kamo-san's work in #1372
_Edit: Added an unified version allowing mixing transformer and rnn through `etype`/`dtype`, except for rnnt-att. Default RNN-transducer is unchanged in terms of performance. Attention plot report was disabled temporary._

In work: 
 - ~Chunk-flow mechanism: https://arxiv.org/pdf/1909.13037.pdf~
_Edit: Pending other related works._
 - Path-aware regularization: https://arxiv.org/pdf/1909.13037.pdf
_Edit: Almost done._


espnet/nets/pytorch_backend/e2e_asr_transducer.py
```diff
@@ -44,397 +49,360 @@ class E2E(ASRInterface, torch.nn.Module):
...
+        General options encapsulate both modules options.
+
+        """
+        group = parser.add_argument_group("transformer model setting")
+
+        # Encoder - general
         group.add_argument('--etype', default='blstmp', type=str,
-                           choices=['lstm', 'blstm', 'lstmp', 'blstmp', 'vgglstmp', 'vggblstmp', 'vgglstm', 'vggblstm',
-                                    'gru', 'bgru', 'grup', 'bgrup', 'vgggrup', 'vggbgrup', 'vgggru', 'vggbgru'],
+                           choices=['transformer', 'lstm', 'blstm', 'lstmp', 'blstmp',
```
sw005320: "transformer" or "self attention (sa)"?

sw005320: sorry, keep to use transformer to be consistent with the other names. please ignore the above.

sw005320:
i think it's better to add tranceducer related integration tests in https://github.com/espnet/espnet/blob/master/ci/test_integration.sh (both rnn, transformer) to make them fully supported.

b-flo:
OK, I'll do that later this day! Should I also add integration test for mixed architecture (e.g.: transformer enc/RNN decoder)?

Also, about the mixed architecture, I'm wondering if I should support multiple optimizer, one for each part (transformer and RNN).

sw005320:
you can only add the normal ones. the test for various architectures should be supported by the unit test.

b-flo:
Is there any way to visualize travis artifact (e.g: `/home/circleci/repo/egs/mini_an4/asr1/exp/train_nodev_pytorch_train_transducer/train.log`)? I couldn't pinpoint the problem in travis build and I can't reproduce the error locally with `test_integration.sh` on ubuntu 16.04 and ubuntu 18.04. Maybe memory related?

espnet/nets/pytorch_backend/e2e_asr_transducer.py
```diff
@@ -44,397 +49,360 @@ class E2E(ASRInterface, torch.nn.Module):
...
+        return nbest_hyps
 
     def calculate_all_attentions(self, xs_pad, ilens, ys_pad):
         """E2E attention calculation.
 
         Args:
             xs_pad (torch.Tensor): batch of padded input sequences (B, Tmax, idim)
             ilens (torch.Tensor): batch of lengths of input sequences (B)
-            ys_pad (torch.Tensor): batch of padded character id sequence tensor (B, Lmax)
+            ys_pad (torc.Tensor): batch of padded character id sequence tensor (B, Lmax)
```
sw005320:
```suggestion
            ys_pad (torch.Tensor): batch of padded character id sequence tensor (B, Lmax)
```

b-flo:
Hi,

I'm wondering if I should end this PR here. Excluding more results what remains is path-aware regularization. I think the implementation I worked on (based on https://arxiv.org/pdf/1909.13037.pdf) is not really designed for espnet (even less for espnet2). It heavily rely on kaldi in the first part and it can easily create unnecessary work and issues.
I think it can be a great addition but some heavy modifications are needed for espnet. I had something in mind but I'm not really sure, when you have some time @kan-bayashi I would like to discuss this with you as I think you can help me !

sw005320:
I tend to agree with it.
The extensions in the source code seem to be OK, but the recipe may become complicated. Please let me know the conclusions once you have some discussions.

b-flo:
> The extensions in the source code seem to be OK, but the recipe may become complicated.

You're completely right. Extensions in source code are OK and can be easily controlled, however recipes are quite complicated for users not experienced with Kaldi models, pipeline and all fst stuff.

---

If someone can do some test and training with this PR (transformer-transducer and mixed rnn-transformer transducer), it can be merged after that. My resources are allocated for how2 and chime5 for now.

b-flo:
Hi,

I add a GPU available and free time so I did some experiments with transducer using vivos. I added recipes and results for RNN-transducer and Transformer/RNN-transducer.
I was a bit lazy though so there's probably a better config for both.

sw005320:
This is a really cool result!
Great job!
Do you think you can also get a normal transformer result?

egs/vivos/asr1/RESULT
```diff
@@ -22,4 +22,39 @@ exp/train_nodev_pytorch_train_mtlalpha1.0/decode_train_dev_decode_ctcweight1.0_l
 exp/train_nodev_pytorch_train_mtlalpha1.0/decode_test_decode_ctcweight1.0_lm_word7184/result.txt:|  SPKR         |  # Snt    # Wrd   |  Corr       Sub      Del       Ins       Err    S.Err   |
```
sw005320: Can you follow `RESULTS` with the other one (e.g., https://github.com/espnet/espnet/blob/master/egs/librispeech/asr1/RESULTS.md)?
- [ ] change it from text to markdown `RESULTS.md`
- [ ] change the order of the results starting from new one
- [ ] add more system information to run the recipe (see https://github.com/espnet/espnet/blob/master/CONTRIBUTING.md#how-to-contribute-to-espnet)
- [ ] upload your models and put a model link to `RESULTS.md`

sw005320: I'll send our Google drive link invitation later.

b-flo: Sure! My policy when working on an existing recipe is to follow the same style, that's why I didn't touch the results file. I'll update the file and add model to google drive!

sw005320: Oh, I see.
Thanks.

b-flo:
> Probably better to call self-attention transducer rather than transformer RNN-transducer? (SA-T?). Are you still using RNN somewhere (e.g., decoder part?)? But even there is an RNN component, I think self-attention transducer sounds good for me.

Transformer/RNN-transducer is a mix of Transformer (encoder) and RNN (decoder). I would prefer using "transducer" for both but the term is a bit generic I think...

> Do you think you can also get a normal transformer result?

Sure. You mean transformer CTC in this case?

sw005320:
> > Probably better to call self-attention transducer rather than transformer RNN-transducer? (SA-T?). Are you still using RNN somewhere (e.g., decoder part?)? But even there is an RNN component, I think self-attention transducer sounds good for me.
> 
> Transformer/RNN-transducer is a mix of Transformer (encoder) and RNN (decoder). I would prefer using "transducer" for both but the term is a bit generic I think...

I personally don't want to call the SA encoder as a transformer (I want to use the word `transformer` for entire encoder-decoder architecture with SA).
But your intention is to keep the terminology `RNN` for the decoder, right?
How about the SA-RNN-transducer?
If we also make the decoder SA, we can then call it transfromer-transducer.

> > Do you think you can also get a normal transformer result?
> 
> Sure. You mean transformer CTC in this case?

Year, this is also interesting, but my intentions are the transformer defined above (SA encoder-decoder)

b-flo:
Sorry I should have made another comment instead of editing.

> I personally don't want to call the SA encoder as a transformer (I want to use the word transformer for entire encoder-decoder architecture with SA).
But your intention is to keep the terminology RNN for the decoder, right?
How about the SA-RNN-transducer?
If we also make the decoder SA, we can then call it transfromer-transducer.

Current implementation is an unified version, transformer is specified through etype/dtype
This recipe use a LSTM decoder because I wanted to highlight composition and the new bottleneck (vgg2l instead of conv2d).

sw005320:
I see. Then it makes sense for me.

b-flo:
I renamed the PR to avoid confusion, thanks!

tu1258:
> Hi,
> 
> I had a GPU available and free time so I did some experiments with transducer using vivos. I added recipes and results for RNN-transducer and Transformer/RNN-transducer.
> I was a bit lazy though so there's probably a better config for both.
> 
> model	LM	CER (test)
> CTC (default)	None	22.2
> CTC (default)	char LM	19.1
> CTC (default)	word LM	20.2
> RNN-transducer	None	18.7
> RNN-transducer	char LM	18.3
> RNN-transducer	word LM	18.0
> Transformer/RNN-transducer	None	17.7
> Transformer/RNN-transducer	char LM	17.2
> Transformer/RNN-transducer	word LM	16.7

Hi, I am working on On-device Streaming/Online ASR. Recently, I am evaluating the latency/model size/accuracy of E2E streaming model because I would implement it on FPGA/ASIC. I may just adopt your streaming model since I am not experienced in machine learning(my main job is to implement software algorithm on hardware). So, what's the WER of RNN-T on English dataset(e.g. LibriSpeech, SWBD)? I just want to evaluate the performance of those models so if there is no test on English dataset maybe provide some similar performance comparison among streaming model. Also, CTC or RNN-T which would you recommend? My requirements are latency < 500ms(hopefully < 100ms), parameters < 100M (hopefully < 50M), WER < 20 (hopefully < 10), these condition may help evaluation. Any advice would be appreciate. Thanks very much!

b-flo:
Hi,

Online/streaming decoding is not available for transducers and this PR will not address items related to it unfortunately! The objective is to enable on-device streaming decoding with transducers but some things need to be discussed beforehand, related to the two requirements you mentioned (latency, number of parameters)! I can't really give a estimated time though, maybe a few weeks at best!

tu1258:
Hi,

Is there any online/streaming model on ESPnet? It would be great even if there is only online ctc model. I will stay tuned for your online RNN-T model though. Keep up the good work!

egs/vivos/asr1/RESULT
b-flo: Which model(s) do you want me to upload? Also, should I retrain default config to add missing infos and WER report?

b-flo: Oh and I'll try to to work a bit on the config files later when I have time for that. I didn't really tune each network so comparison is a bit pointless right now

ci/test_integration.sh
```diff
@@ -19,6 +19,15 @@ echo "==== ASR (backend=chainer) ==="
...
 ./run.sh --stage 2 --train-config ./conf/train_mulenc2.yaml --decode-config ./conf/decode_mulenc2.yaml --mulenc true
+
+# test transducer recipe
+echo "=== ASR (backend=pytorch, model=rnnt) ==="
+./run.sh --stage 2 --train-config conf/train_transducer.yaml \
+         --decode-config conf/decode_transducer.yaml
+echo "=== ASR (backend=pytorch, model=transformer-transducer) ==="
+./run.sh --stage 2 --train-config conf/train_transformer_transducer.yaml \
+         --decode-config conf/decode_transducer.yaml
+
```
ShigekiKarita: Good good good. This makes Transducer live forever. I think the can start from `--stage 3` because previous tests performed the data preparation.

egs/vivos/asr1/RESULTS.md
```diff
@@ -0,0 +1,123 @@
...
+# pytorch Transducer (enc: VGG2L + 8 x Transformer, dec: 2 x Transformer)
+
+- Environments
+  - date: `Sun Feb  9 20:50:30 CET 2020`
+  - python version: `3.7.3 (default, Mar 27 2019, 22:11:17)  [GCC 7.3.0]`
+  - espnet version: `espnet 0.6.1`
+  - chainer version: `chainer 6.0.0`
+  - pytorch version: `pytorch 1.0.1.post2`
+  - Git hash: `b19ae06fa0ee36d9a65ca701dd169d52cf1a24fd`
+  - Commit date: `Fri Feb 7 16:32:23 2020 +0100
```
ShigekiKarita: If you can upload the pretrained model later, that would be super great. However, it is not necessary.

b-flo: Sure! Which model(s) should I upload? RNN-Transducer, Transformer-Transducer or Transformer/RNN-Transducer? :)

ShigekiKarita:
I think the best one (Transformer?) is enough

2020年2月13日(木) 2:18 b-flo <notifications@github.com>:

b-flo:
OK, I'll add it! The "best" one is Transformer (enc) + LSTM (dec) but right now the comparison is a bit pointless, I didn't really tune any model! I'll do that when I have time.

ci/test_integration.sh
b-flo: I remember now why I started from stage 2, it is because transducer test is after `./run.sh --stage 2 --train-config ./conf/train_mulenc2.yaml --decode-config ./conf/decode_mulenc2.yaml --mulenc true` which would throw the following error if I started from stage 3 : 

```
...
File "/home/fboye/espnet/espnet/asr/pytorch_backend/asr.py", line 504, in <lambda>
    dataset=TransformDataset(train, lambda data: converter([load_tr(data)])),
  File "/home/fboye/espnet/espnet/asr/pytorch_backend/asr.py", line 257, in __call__
    xs, ys = batch[0]
ValueError: too many values to unpack (expected 2)
```

sw005320:
Is it ready?
Then, I'll merge it (after the CI check)

b-flo:
It is ready on my side, maybe you want to also review it @kan-bayashi ?

sw005320:
I think it is OK since I and Shigeki already checked it.
Thanks for your great work!

sw005320:
Can you also make another PR to advertize your implementation in the main `README.md`?

b-flo:
It is already in the main `README.md`!

qmpzzpmq:
@b-flo is Chunk-flow mechanism still in the pending?

b-flo:
Hi,

Yes, similar works for online decoding were already discussed (or done) by other espnet members so I put my focus on other works in the meantime. @sw005320 Do you have some comments on that maybe?

I'll resume tasks for online decoding when everything is set or after my current TODO list is finished, i.e.: other search algorithms and porting transducer to espnet2.

sw005320:
Sorry for my late response.
I think we need some discussions.
Can we have some sync after the interspeech deadline in the next week?
We have several other possible directions for online decoding and want to unify it (but probably it is difficult to completely unify them due to the time-sync and label-sync difference

b-flo:
Sure, I'm available next week! Also, it may be easier to implement an unified method after I finish implementing the other search algorithms.

MarkWuNLP:
Hi @b-flo ,

Thank you for your great work on ESPnet Transformer Transducer. Recently, I am trying to reproduce Google [Transformer Transducer](https://arxiv.org/pdf/2002.02562.pdf) result with ESPnet Transformer Transducer. However, the number is still far behind google's result 2.4/5.6 on Librispeech. Our best number is 3.3/7.7. Can you reproduce Google's number on Libirspeech?

b-flo:
Hi,

I couldn't try myself (all my ressources are allocated on other works since the beginning of the year) but one of my coworkers did try to reproduce the results from this paper. He ended-up with similar results (a bit better though IIRC) as the one you reported. After a bit of investigation, I found out there were several things he couldn't reproduce from the paper with current implementation (FWIR: some minor transformer architecture difference and training procedure).

I plan to work on the librispeech recipe when I have time but I can't give a timeline unfortunately. I may use a different architecture though, as I found out TDNN-Transformer encoder architecture and causalConv1D-Transformer decoder architecture suitable in some experiments.

b-flo:
@MarkWuNLP Would it be possible to give me access to your models (AM, LM)? I would like to try/apply one of my work for inference but I don't have any resources available to train a librispeech model and I didn't keep the models my coworker trained.

MarkWuNLP:
> @MarkWuNLP Would it be possible to give me access to your models (AM, LM)? I would like to try/apply one of my work for inference but I don't have any resources available to train a librispeech model and I didn't keep the models my coworker trained.

Ok.  Let's have a discussion via email (wumark@126.com)

keshawnhsieh:
> > @MarkWuNLP Would it be possible to give me access to your models (AM, LM)? I would like to try/apply one of my work for inference but I don't have any resources available to train a librispeech model and I didn't keep the models my coworker trained.
> 
> Ok. Let's have a discussion via email ([wumark@126.com](mailto:wumark@126.com))

@b-flo @MarkWuNLP I am doing the same task recently.I would be very grateful if you could share the latest update on T-T model performance on LibriSpeech with ESPNET. Any hints like training recipe, configurations,  would be helpful.

b-flo:
Hi,

Sorry, in the end I didn't ask @MarkWuNLP's model because I shifted my focus on other architecture. I don't have a transformer-based or conformer-based model compatible with current transducer implementation but it will be made available in ESPnet eventually.

Issue #1385, opened 2019-11-26
attention constrain of transformer TTS


xinqipony:
I noticed there're many issues especially when decoding long phone input for  transformer TTS(teacher), 
in fastspeech , there's attention constrain in decoding process to avoid bad cases, in this code base, I see a guided attention, so how's the effect of it? anyone test it?

kan-bayashi:
I did not explicitly evaluated objectively but if we add guided attention loss it helps to reduce repetition or deletion errors (when I compared some samples). But it is better to check the effect objectively using ASR results.

r9y9:
FWIW, constraining attention at inference time (e.g., monotonic) is worth trying and could alleviate alignment failures. A while ago I tried [attention constraints](https://github.com/r9y9/deepvoice3_pytorch/blob/6aef4c817745c58fe2fb56d36b533b86bd137fa3/deepvoice3_pytorch/deepvoice3.py#L150-L156) similar to [DeepVoice3](https://arxiv.org/abs/1710.07654), and it worked pretty well in practice. In my experience, guided attention loss doesn't help a lot at inference time. It does help convergence at training time though.

kan-bayashi:
@r9y9 This is interesting. Is it possible to integrate with Taco or Transformer?

r9y9:
I also see lots of alignment failures particularly in Transformer TTS and wanted to fix the practical issues. I wish I could have enough time to do it myself, but (I believe) we are always welcome to contributions.

r9y9:
To be honest, I'm not sure because I don't remember details (I guess I implemented it almost two years ago...) but I believe we can do similar in Taco and Transformer as well.

kan-bayashi:
I see. I will check the paper and if possible, I will try to implement it.

kan-bayashi:
I understood the technique in Deep Voice 3. It is easy to implement for Tacotron2 but for Transformer it is not straightforward. This is because in the case of Transformer there are several diagonal attention heads. So we need to predefine the diagonal heads to apply this technique.

xinqipony:
awesome, espnet team takes a real fast progress and action, great work! and could Transformer tts model be supported too?

kan-bayashi:
@xinqipony It is not straightforward in the case of Transformer because there are a lot of heads and only some of the heads are diagonal. I'm not sure what is happened we apply the attention constraint to all of the heads. But it is worthwhile to try. I will try it.

PR #1429, opened 2019-12-12
[EXPERIMENTAL] Transformer attention constraint

kan-bayashi:
This PR adds attention constraint for Transformer.
This is the experimental feature so I'm not sure about working well.

Related issue: #1385


kan-bayashi:
Here are the sample of encoder-decoder attentions.

## w/ constraint

スクリーンショット 2019-12-12 17 38 39
スクリーンショット 2019-12-12 17 39 18
スクリーンショット 2019-12-12 17 39 43

## w/o constraint



if we use current attention constraint, the generated samples become unstable, not-stopping well, and leading mispronunciation.
We need to consider 2-path decoding. (First, we decode the utterance without constraint and then decide the diagonal head. Second, we apply the constraint for only diagonal heads, which is different for each head.)

xinqipony:
nice work! maybe we should choose the diagonal attention for every decoder step.

PR #1430, opened 2019-12-13
[ESPnet2] Update TTS recipes

kan-bayashi:
This PR updates TTS recipes in espnet2.

- fixed all of shellcheck errors in `tts.sh`
- added training config example
- added decoding config example
- added options in `espnet2/tacotron2.py`
- added comments for each argument in `tts.sh`
- unified the log directory name to `log/`
- added `dumpdir`

Also I updated tag management.
- Added `--train_args` info in `tag` because it will overwrite the config.
- Added `--decode_model` info in `decode_tag` to identify the model used for the decoding.


kan-bayashi:
@kamo-naoyuki How do you think the integration of `tts_decode()` with `task`?

egs2/TEMPLATE/tts1/tts.sh
```diff
@@ -287,28 +293,29 @@ if [ ${stage} -le 5 ] && [ ${stop_stage} -ge 5 ]; then
...
 
         # 2. Submit decoding jobs
         log "Decoding started... log: '${_logdir}/tts_decode.*.log'"
+        # shellcheck disable=SC2086
         ${_cmd} --gpu "${_ngpu}" JOB=1:"${_nj}" "${_logdir}"/tts_decode.JOB.log \
             python3 -m espnet2.bin.tts_decode \
                 --ngpu "${_ngpu}" \
-                --data_path_and_name_and_type "${_data}/token_int,text,text_int" \
-                --key_file "${_logdir}"/keys.JOB.scp \
+                --data_path_and_name_and_type "${_logdir}/keys.JOB.scp,text,text_int" \
```
kan-bayashi: @kamo-naoyuki How do you think this change? For me, this is more simple.

kamo-naoyuki: Thank you for reading deeply.
I cared about the second inputs case for this.
If directly inputting from `data_path_and_name_and_type`, then the option of `key_file` is hidden and then users may think that the second scp file also needs to be split.
(and I think you'll add such option to `tts.sh`, if not so, your idea is better.)

egs2/TEMPLATE/tts1/tts.sh
```diff
@@ -188,37 +189,42 @@ if [ ${stage} -le 2 ] && [ ${stop_stage} -ge 2 ]; then
...
+    # "nlsyms_txt" should be generated by local/data.sh if need
+    if [ -n "${nlsyms_txt}" ]; then
+        nlsyms="$(<${nlsyms_txt})"
+    else
+        nlsyms=
+    fi
     echo "<unk>" > "${token_list}"
     if [ -n "${nlsyms}" ]; then
-        cat ${srctexts} | pyscripts/text/text2token.py -s 1 -n 1 -l "${nlsyms}"  \
+        pyscripts/text/text2token.py -s 1 -n 1 -l "${nlsyms}" < "${srctexts}" \
```
kamo-naoyuki: I expected that `srctexts` can be multiple texts, like srctexts="a.txt b.txt", but I don't know how to solve this with shell check.

egs2/TEMPLATE/tts1/tts.sh
```diff
@@ -9,125 +11,123 @@ set -o pipefail
...
+# check feature type
 if [ "${feats_type}" = fbank ]; then
-    data_feats=data_fbank
+    data_feats="${dumpdir}/fbank"
 elif [ "${feats_type}" = stft ]; then
-    data_feats=data_stft
+    data_feats="${dumpdir}/stft"
 elif [ "${feats_type}" = raw ]; then
-    data_feats=data_raw
+    data_feats="${dumpdir}/raw"
```
kamo-naoyuki: The idea of `dumpdir` looks good, I'll also use this in `asr.sh`

kamo-naoyuki:
> @kamo-naoyuki How do you think the integration of `tts_decode()` with `task`?

I also thought about this, and I'm not still sure.

Pros:
- Basically task is  designed for training, so we should keep it as the least implementation. Task may be going to be more big from now.
- We may need diffreent story decoding ways for ASR, e.g. RNN-T decode or frame-synchronous beam search, and split codes may be better for that case.

Cons:
- The decoding part is tightly coupled with task, so one file is more understandable.

egs2/TEMPLATE/tts1/tts.sh
```diff
@@ -9,125 +11,123 @@ set -o pipefail
...
+dev_set=        # name of development set
+eval_sets=      # names of evalaution sets
+srctexts=       # source text
+nlsyms_txt=     # If non-linguistic symbol list if existing
+trans_type=char # transcription type
 
 
 log "$0 $*"
-. utils/parse_options.sh
+. utils/parse_options.sh || exit 1;
```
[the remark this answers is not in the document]
kan-bayashi: I checked it. You are right. It is not needed.

egs2/TEMPLATE/tts1/tts.sh
kan-bayashi: I see. I wil revert it.

egs2/TEMPLATE/tts1/tts.sh
kan-bayashi: I see. `--key_file` is nice when we also want to use such as `--data_path_and_name_and_type x_vector.scp,spembs,...` because we do not need to split scp multiple times. I will leave comments about it.

kamo-naoyuki:
lgtm

kamo-naoyuki:
By the way, how do you think merging convert_fbank_to_wav.py into tts_decode?

kan-bayashi:
> By the way, how do you think merging convert_fbank_to_wav.py into tts_decode?

This is a nice idea. I will consider to merge in the next PR.

kan-bayashi:
@kamo-naoyuki I updated tag management to add `train_args`, `decode_args`, and `decode_model` info.
https://github.com/kan-bayashi/espnet/blob/4ce1803e272ebd0d83d8f075ad9c08f806cd88dd/egs2/TEMPLATE/tts1/tts.sh#L103-L126
I think this is nice when we want to manage several experiments using `--train_args` or `--decode_args` with for loop.

kamo-naoyuki:
> @kamo-naoyuki I updated tag management to add `train_args`, `decode_args`, and `decode_model` info.
> https://github.com/kan-bayashi/espnet/blob/4ce1803e272ebd0d83d8f075ad9c08f806cd88dd/egs2/TEMPLATE/tts1/tts.sh#L103-L126
> I think this is nice when we want to manage several experiments using `--train_args` or `--decode_args` with for loop.

Thank you. If possible, could you apply this for asr.sh? I'd like to keep same style between these scripts.

egs2/TEMPLATE/tts1/tts.sh
```diff
@@ -9,125 +11,132 @@ set -o pipefail
...
-
-# exp tag
-tag="" # tag for managing experiments.
-exp=exp
-# config files
-train_config=
-train_args=
-
+stage=1          # stage to start.
+stop_stage=100   # stage to stop.
```
kamo-naoyuki: Is it better to write the help comment in ${help_message}, isn't it? Though I didn't write both...

kan-bayashi:
> Thank you. If possible, could you apply this for asr.sh? I'd like to keep same style between these scripts.

I will do it in the next PR.

egs2/TEMPLATE/tts1/tts.sh
kan-bayashi: Right. I also updated help message.

PR #1439, opened 2019-12-15
[ESPnet2] Update ASR recipe

kan-bayashi:
This PR updated ASR recipe to unify with TTS recipe update (#1430).
- fixed shellcheck errors
- added comments for options
- updated help messages
- fixed typos
- added dumpdir
- updated tag management

### DIFF
--- a/ci/test_integration.sh
+++ b/ci/test_integration.sh
@@ -50,7 +50,8 @@ for t in ${token_types}; do
 done
 for t in ${feats_type}; do
     for t2 in ${token_types}; do
-        ./run.sh --ngpu 0 --stage 7 --feats-type "${t}" --token-type "${t2}" --asr-args "--max_epoch=1"
+        ./run.sh --ngpu 0 --stage 7 --feats-type "${t}" --token-type "${t2}" \
+            --asr-args "--max_epoch=1" --lm-args "--max_epoch=1"
     done
 done
 cd - || exit 1
--- a/egs2/TEMPLATE/asr1/asr.sh
+++ b/egs2/TEMPLATE/asr1/asr.sh
@@ -69,7 +69,7 @@ decode_asr_model=eval.acc.best.pt # ASR model path for decoding.
 train_set=     # Name of training set.
 dev_set=       # Name of development set.
 eval_sets=     # Names of evaluation sets. Multiple items can be specified.
-srctexts=      # Used for the training of BPE and the creation of a vocabulary list
+srctexts=      # Used for the training of BPE and the creation of a vocabulary list.
 lm_train_text= # Text file path of language model training set.
 lm_dev_text=   # Text file path of language model development set.
 lm_test_text=  # Text file path of language model evaluation set.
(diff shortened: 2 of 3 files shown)


sw005320:
Can you consider the following?

1. `cer` or `ter`? With BPE, it becomes a token error rate (ter), right?
Also, it is possible to display a real cer with some sclite options? 
Ideally, we want to have cer, ter, and wer for languages that have space-based word boundaries.
2. put all model files under `${expdir}/asr_${asr_tag}`?
Currently, we have some issues with packing models since the location of models (bpe, lm, asr, cmvn) are distributed. 
Can you put the actual file or put the symlink of them under `${expdir}/asr_${asr_tag}`?

@kamo-naoyuki,
I welcome your input about them.

sw005320:
More thought.

We may make model packing and RESULTS.md creation as an additional (optional) stage. With that, model sharing would be accelerated.

sw005320:
More thought2.

How about supporting the use of a trained model that only performs inference inside `asr.sh` and `tts.sh`?
Again, this boosts the model ecosystem.
If we don't specify a trained model (or we don't have any trained models), these scripts start to train the model.

kamo-naoyuki:
> Can you consider the following?
> 
>     1. `cer` or `ter`? With BPE, it becomes a token error rate (ter), right?
>        Also, it is possible to display a real cer with some sclite options?
>        Ideally, we want to have cer, ter, and wer for languages that have space-based word boundaries.

 I'm not familiar with sclite much. 
Could TER be calculated by just inputting BPE tokens as it is without any additional options? Looks good.

`sclite -c` can calculate cer level editdistance, but I'm not sure we could use it. 
    - it doesn't count space char, can it do with another option?
    - It can't handle non-ascii char well?
If my understanding is correct, I'll calculated CER with splitting texts to chars before inputting sclite by awk or sed.

kamo-naoyuki:
>        Currently, we have some issues with packing models since the location of models (bpe, lm, asr, cmvn) are distributed.
>        Can you put the actual file or put the symlink of them under `${expdir}/asr_${asr_tag}`?

> How about supporting the use of a trained model that only performs inference inside `asr.sh` and `tts.sh`?

Nice idea. I'll take into account. I'm thinking where is better to put model and the other files now.

> We may make model packing and RESULTS.md creation as an additional (optional) stage. With that, model sharing would be accelerated.

Yes, I agree. RESULTS.md is one of my 1st priority matters.

kan-bayashi:
I agree with most of your ideas.

> How about supporting the use of a trained model that only performs inference inside asr.sh and tts.sh?

I think it is better to separate the script because it make the script more complex.
Maybe the user who wants to perform only the inference likes the simple script, so why don't you make a single python script for the inference?

kan-bayashi:
And I want to avoid making the PR bigger, I want to merge this PR without additional extension.

kamo-naoyuki:
> Maybe the user who wants to perform only the inference likes the simple script, so why don't you make a single python script for the inference?

Such people may not intend to use parallel options of ${cmd}, 
then we can already do it just using `python -m espnet2.bin.asr_decode` since I have embed feature extractor in the model and implemented the text converter from bpe to text, and also you would merge `convert_fbank` to `tts_decode`. 

As far independent evaluation for ASR using pretrained model, I'm not sure which is better to include in `asr.sh` or to utilize independent script. I'll try to embed such options in `asr.sh` at the first and if I felt too complex, then I'll consider to create separate one.
I also add options to use pretrained model with training.
Anyway, to realize these items we need some packed expression for the model information.

> And I want to avoid making the PR bigger, I want to merge this PR without additional extension.

Maybe @sw005320 just used here for discussion place. Your PR looks already good for me.  Thank you.

kan-bayashi:
I will merge this PR.

PR #1441, opened 2019-12-16
[WIP] [ESPnet2] Collect stats mode for on-the-fly preprocessing

kamo-naoyuki:
- Implemented `AbsTask.stats_run()` which is used before running every training and derive shape information of input data and accumulated the value and square sum for cmvn.
    - I refer this new feature as `stats_run`.
- Implemented `espnet2.bin.aggregate_stats_dirs`:  `AbsTask.stats_run()`  can be executed in parallel using `${cmd}` and the outputs are gathered by this command into a directory. 
- Removed `prepare_token.sh` from `asr.sh` and I add these commands instead of it. 
   - I have not done for `tts.sh` yet, but this is my TODO.

How do you think?


kan-bayashi:
This is great. But the name `stats_run` is a little bit strange to me. How about `collect_stats` or `calculate_stats`?

kamo-naoyuki:
I'll reopen this PR with another shape because I noticed some parts needs more refactored.

PR #1442, opened 2019-12-16
[ESPnet2] Integrate Griffin-Lim with tts_decode()

kan-bayashi:
This PR integrates griffin-lim with `tts_decode()`.
Thanks to the integration, `stage 6` is no more needed.
Also, I changed to write both normalized and de-normalized features because some neural vocoder receives de-normalized features as inputs.

Some thought:
- I'm considering saving feature-related params (e.g. `--n_mels`) in training config even in the case of "feats_type=fbank". Then the script can be more simpler.
- I'm not sure where I should put `griffin_lim.py`. Currently, there is under `espnet2/utils`.


espnet2/bin/tts_decode.py
```diff
@@ -129,46 +164,44 @@ def tts_decode(
...
-
             logging.info(
                 f"({idx}/{len(batch_sampler)}) {key} "
                 f"(size:{insize}->{outs.size(0)})"
             )
             f[key] = outs.cpu().numpy()
+            g[key] = outs_denorm.cpu().numpy()
+
+            wav = spectrogram2wav(outs_denorm.cpu().numpy(), **spectrogram2wav_params)
+            sf.write(f"{output_dir}/wav/{key}.wav", wav, spectrogram2wav_params["fs"], "PCM_16")
```
kamo-naoyuki: I think wav scp file is useful for after processing, e.g. ASR-evaluation.
https://github.com/espnet/espnet/blob/f423c25e7081b4728064f494d2f2548611dea149/espnet2/utils/fileio.py#L218-L232
This class can generates wave files with scp text as the kaldi writer.

kamo-naoyuki:
Basic question, sorry, currently do we have options to change the vocoder to neural vocoder one at the recipe level? I has the following idea about this:

```python
y = tts(x)
if args.type == "griffin_lim":
    y = denormalize(y)
    wav = griffin_lim(y)
elif args.type == "vocoder2":
    wav = neural_vocoder(y)
elif args.type == "time":
    wav = y
elif args.type == "denormalized":
    y = denormalize(y)
    # Save y as array
elif args.type == "normalized":
    # Save y as array
elif args.type == "plot":
   # plot spec?
```

espnet2/bin/tts_decode.py
kan-bayashi: This is nice. I will introduce it.

kan-bayashi:
@kamo-naoyuki I think normed feature, de-normed feature, and plots of attention and stop token probability should be saved in the default due to the randomness of E2E-TTS model (The generated samples will be changed every inference in E2E-TTS, i.e., not deterministic). Therefore, we want to re-use generated features several times to generate a waveform with different vocoders.

kan-bayashi:
@kamo-naoyuki 
I think the word `frontend` for the TTS is very confusing.
How do you think following change?
- `espnet2/asr/frontend` -> `espnet2/speech_frontend`
And then maybe we can create `espnet2/text_frontend`.
Another idea is 
`espnet2/frontend/speech` / `espnet2/frontend/text`

kamo-naoyuki:
>     * `espnet2/asr/frontend` -> `espnet2/speech_frontend`
> 
> 
> And then maybe we can create
> 
>     * `espnet2/text_frontend`

Please give me some time. This is very sensitive.

Basically, I agree your idea, but I'd like to use another word than `frontend` because we often refer the speech enhancement part as the frontend in the speech recognition context.
Maybe it is better to separate enhancement block and feats extractor class, but give me time to think about this. For me, this structure is useful for ASR research.

I want to say that to sort our situation:

- I implemented two methods for `preprocessing`:  The feature extractor in the `E2E`, which can be embedded in computational graph, and the preprocessor used by `Dataset`. I'm tackling #1441 being aware of both usage. Currently the latter is named as `preprocessing`.
- Feats normalization is also one of the such processes.
- Some text processing is used for "post process" when decoding: e.g token->Text for ASR.

 Actually, I'm not sure we need abstract class for preprocessing and I need more inputs for this problem to generalize them in advance.
Could you give me your idea of preprocessing? What kinds or how many preprocessing will you implement? Or, could you give me some pseudocodes for them?

kan-bayashi:
I agree with your idea to separate `frontend` (enhancement) and `preprocessing`.
For TTS, we may want to implement text to text preprocessing like:
- `clean_text("no. 1")` -> `"number one"`
- `convert_grapheme_to_phoneme("こんにちは")` -> `"k o n n i ch i w a"`

Or text to token id processing like:
- `token2id("ABC")` -> `[1, 2, 3]`
(This may be already implemented in your PR)

@r9y9 Do you have comments about the frontend and preprocessing parts?

kamo-naoyuki:
Could you merge this? I'd like to rethink some parts taking into account your comment after this PR.

kan-bayashi:
OK. For now, I will keep the current frontend.
After CI, I will merge this.

espnet2/bin/tts_decode.py
kan-bayashi: `wav.scp` is useful for ASR evaluation but wav files are separately saved. TTS users may want to access the directory including all of wav files. For now, I did not come up with a good idea to achieve both wishes, so I keep current style.

PR #1443, opened 2019-12-17
Integrate Radam optimizer

minkyu119:
Integrate Radam optimizer into espnet 

Radam pytorch source from https://github.com/LiyuanLucasLiu/RAdam

i validate optimizer for switchboard task, but my GPU support only 1024 dim transformer

learning rate was working well for swbd when set to 0.001
baseline is 'noam' optimizer 
and the others are Rectified Adam Optimizer with different learning rate

image


sw005320:
Thanks!
Did you observe some improvements by using this?

minkyu119:
I am waiting for experiment result with Radam, Adam(noam) on SWBD basic recipe.(100 epoch, transformer)
Radam loss shows better than noam at 190k step now
I'll share full recognition result with loss graph

PR #1462, opened 2019-12-23
[ESPnet2] collect-stats mode, trainer class, etc.

kamo-naoyuki:
Sorry for large PR. I needed to think these items at the same time.

1. Implemented `collect-stats` mode in `*_train.py`.
related new commands:
   - `espnet2.bin.aggregate_stats_dirs`: `collect-stats mode` can run in parallel with `${cmd}` and this command is responsible to aggregate the output data from each jobs.
   -  `espnet2.bin.tokenize_text`: Just replacement of `text2token.py` and `text2vocaburary`. The most important point is that they are unified in core modules and we can guaranteed same normalization and tokenization among data preparation, training, and decoding.

Why do we need `collect-stats` mode?
Because  I prefer on-the-fly preprocessing for data loading at the training time for espnet2, including feature-extract and text-tokenization. However we couldn't know some required information with on-the-fly preprocessing in advance, e.g. the length of the features and the mean and variance of features.
To solve this problem,  this mode always runs before every training.

2. Trainer class

I cut off some method of `AbsTask` and move them to this new class because Task is too large
and for future developments e.g. GAN, or some tasks which requires more than one optimizers.

Trainer is defined as following:

- `Task` has a `Trainer` as a class attribute and the inheritance of `Task` can change this.

https://github.com/kamo-naoyuki/espnet/blob/8d996b229d854f40e5d53cd7227489df57942524/espnet2/tasks/abs_task.py#L98-L106

- Trainer is responsible to decide the procedures for one epoch.

```python
class AbsTask:
    @classmethod
    def run(cls):
        for epoch in range():
            cls.Trainer.train_one_epoch(model, optimizers, schedulers., ..)
            cls.Trainer.eval_one_epoch(model, ...)
```

3. Created `espnet2.tts.feats_extract` and modified some parts related to it. (we discussed at #1442)
    - This class is independent from asr's frontend class and I keep asr's frontend as it is.
4. Unify `--escheduler` and `--bscheduler` to `--scheduler`.
5. Change the API of NpyScpWriter and SoundScpWriter
6. Simplify Task API
7. nj: 50 -> 32
8. Change the model extension: `.pt` -> `.pth`
9. Append "black" to ci and change autopep8 to pycodestyle

Next TODO:

- Packed model and RESULTS.md
- Distributed training
- Create Text normalization class for TTS and put them as suitable shape.
- Test codes for all classes


espnet2/train/checkpoint.py
```diff
@@ -0,0 +1,160 @@
...
+        if obj is not None:
+            obj.load_state_dict(_st)
+    states = torch.load(resume_path / f"optim.pth")
+    for o, s, state in zip(optimizers, schedulers, states):
+        o.load_state_dict(state["optim"])
+        if s is not None:
+            s.load_state_dict(state["scheduler"])
+
+
+def save_checkpoint(
```
sw005320: Is there a way to control `save_checkpoint`?
If we save checkpoints every time, it may require a lot of storage... (espnet1 has such issue).

espnet2/train/checkpoint.py
```diff
@@ -0,0 +1,160 @@
...
+    )
+    torch.save(states, save_path / f"optim.pth")
+
+    # Write the datetime in "timestamp"
+    with (save_path / "timestamp").open("w") as f:
+        dt = datetime.now()
+        f.write(dt.strftime("%Y-%m-%d %H:%M:%S") + "\n")
+
+
+def load_pretrained_model(
```
sw005320: better to add comments to clarify the distinction between `load_pretrained_model` and `load_checkpoint`?

espnet2/train/trainer.py
```diff
@@ -0,0 +1,292 @@
...
+                    not_increment_count=True,
+                )
+
+            if iiter % log_interval == 0:
+                reporter.logging(nlatest=log_interval)
+        return all_steps_are_invalid
+
+    @classmethod
+    @torch.no_grad()
+    def eval_one_epoch(
```
sw005320: if we want to perform the evaluation for every `n` iterations, do we need to modify this part?
this trainer seems to assume epoch-based processing too much, and we'd have a lot of modifications in the future.
the chainer-based trainer has a lot of flexibility for such perspective, but this makes the trainer very complicated...

espnet2/train/checkpoint.py
kamo-naoyuki: I implemented `--keep_n_best_snapshots` for this (I'll change it to `--keep_n_best_checkpoints` to unify terms..)

- Except for the n-best snapshots, all files are removed at the every epochs.

https://github.com/espnet/espnet/blob/d21721fa1f9647d8c2a29c1c6ae8c5e2d2d765b4/espnet2/tasks/abs_task.py#L1055-L1074

- Regardless of this value, the snapshot of the latest epoch is always saved.

https://github.com/espnet/espnet/blob/d21721fa1f9647d8c2a29c1c6ae8c5e2d2d765b4/espnet2/tasks/abs_task.py#L1068

- The all checkpoints are kept among multiple criterions, e.g. `train.loss`, `eval.acc`.

https://github.com/espnet/espnet/blob/d21721fa1f9647d8c2a29c1c6ae8c5e2d2d765b4/espnet2/tasks/abs_task.py#L1058-L1062

- I'm not sure whether this is better, but it's also used as the number of  averaged model.

https://github.com/espnet/espnet/blob/d21721fa1f9647d8c2a29c1c6ae8c5e2d2d765b4/espnet2/tasks/abs_task.py#L835-L840

Another solution is something like `--save_checkpoints_interval`, but it is incompatible with averaging models, so I didn't append it.

espnet2/train/trainer.py
kamo-naoyuki: Your pointing out may be important. Could you tell me another idea of the non-epoch-based processes if you already have?

As related topics to it, we might need to redesign `AbsTask.run()`. I assumed that the epoch process are consists of `train()`, `eval()`, or `plot_attention()`, and another block can't be inserted, e.g. `train2()`, supervision generation, data cleaning by model, or changing model structure adoptively?? (I think this is not current matters.)

sw005320: If we augment the data (speed perturbation, process various RIRs, or adding noises), one epoch of the amount of training data would become several times of that of the original data. Then, epoch-based monitoring is too coarse. People often use iteration-based monitoring for this setup. Our method could perform such data augmentation dynamically and may still keep original epoch size though. This is just a note.

sw005320: As you know, the Chainer trainer has a trigger function to control it.

kamo-naoyuki: ### Idea1: Evaluate in the middle of training iterations.

```python
for epoch in range(max_epoch):
    for batch in enumerate(train_loader):
        train_one_batch(batch, reporter)
        if reporter.total_number_of_iterations() % eval_interval == 0:
            evaluate_one_epoch(eval_iter)
            plot_attention(eval_iter)
            scheduler_step()
            save_snapshot()
            check_early_stopping()
```

### Idea2: Use subset of training data in every epochs.

```python
for epoch in range(max_epoch):
   train_one_epoch(train_iter.generate_subset_iterator_randomly())
   evaluate_one_epoch(eval_iter)
   ...
```

The latter can't guaranteed using whole dataset, but seems natural as bootstrap method.

kamo-naoyuki: Regarding resuming of training: Unlike chainer's iterator, Pytorch dataloader doesn't provide the system for resuming from the middle of epoch, so if we'll go to idea1's direction, we need to save the number of iterations additionally and seek the batch-sampler at the staring time . 

And, as related topics to it, currently the random-seed is not considered for resuming. I know fairseq carefully handle it for data-loader. i.e. fairseq changes the random seed according to the epoch number. I'm not sure it is worth to do.

egs2/TEMPLATE/asr1/asr.sh
```diff
@@ -330,185 +332,322 @@ if [ ${stage} -le 3 ] && [ ${stop_stage} -ge 3 ]; then
...
+    cat ${srctexts} | \
+        python3 -m espnet2.bin.tokenize_text  \
+            --token_type "${token_type}" -f 2- --input - --output - ${_opts} \
+            --field 2- \
+            --write_vocabulary true \
+            --add_symbol "${blank}:0" \
+            --add_symbol "${oov}:1" \
+            --add_symbol "${sos_eos}:-1" \
+        > "${token_list}"
+
```
kamo-naoyuki: I removed `echo ${blank}...` and `cut -f` and and embed  `espnet2.bin.tokenize_text`.

espnet2/train/trainer.py
kamo-naoyuki: @sw005320 Could you decide the direction? We need a little big modification for the iterator structure with idea1. Idea2 is more simple implementation.

kamo-naoyuki: Anyway it should be done in next PR.

kan-bayashi: Recently I made the trainer which is similar to the idea 1. So +1 for idea 1.

https://github.com/kan-bayashi/ParallelWaveGAN/blob/2908bc5bce95f903f27a015f172ffd1b20017560/parallel_wavegan/bin/train.py#L208

sw005320: Sorry for my late reply.
Now, I'm back to normal.
I think idea 1 is intuitive.

setup.py
```diff
@@ -7,104 +7,106 @@
...
+        "pip>=19.0.0 is required, but your pip is {}. "
+        'Try again after "pip install -U pip"'.format(pip.__version__)
+    )
 
 requirements = {
-    'install': [
+    "install": [
         # 'torch==1.0.1',  # Installation from anaconda is recommended for PyTorch
-        'chainer==6.0.0',
+        "chainer==6.0.0",
```
sw005320: we may add some comments that we'll remove chainer installation in the near future.

sw005320:
LGTM.
@kamo-naoyuki, if this PR is ready, I'll merge it.

espnet2/train/trainer.py
kamo-naoyuki: O.K. I already have gone to the direction.

kamo-naoyuki:
Please merge. I'll send next PR.

sw005320:
sure! many thanks!

PR #1466, opened 2019-12-25
fixed a small problem in run.sh

Peidong-Wang:
changed the condition of running local/ami_text_prep.sh from ` [ ! -d data/local/downloads ] ` to ` [ ! -d data/local/downloads/annotations ] ` since data/local/downloads has already been created in stage -1


kan-bayashi:
Thanks a lot!

PR #1472, opened 2019-12-26
[WIP] DiPCo Recipe

Fhrozen:
Recipe for the Diner Party corpus (https://arxiv.org/ftp/arxiv/papers/1909/1909.13447.pdf)


Fhrozen:
Part of the error is like this:
```
2020-01-10 13:57:10,381 (asr_utils:626) INFO: groundtruth: hmmm hmmm
2020-01-10 13:57:10,381 (asr_utils:627) INFO: prediction : mmm<eos>
```
or
```
2020-01-10 13:53:21,712 (asr_utils:626) INFO: groundtruth: hmmm
2020-01-10 13:53:21,715 (asr_utils:627) INFO: prediction : mm<eos>
```
I am wondering if it may be possible to normalize the onomatopoeia(?)  with a label as  [humming] (?).
In the dipco corpus, the hmmm is more standard, while in chime 5 is can has h or not and more m's. 

But of course, this could be like 1 or 2 % of the WER. 
I am testing (only, bc of resources) the eval beamformit without LM and higher ctc (0.3). I will try to finetune the parameters of the decoding to check if it maybe possible to get better result.

sw005320:
Thanks for the report.
I think we should not have to normalize it since we cannot make a fair comparison with other reports.
Once you tune the search parameters, please update the result and fix this PR.

sw005320:
Is it ready for merge?
For now, we don't have to care about the number.

Fhrozen:
@sw005320, yes. it is ready to be merge.
Sadly, another tests only show worst performance than the current results.
I will try to take this later.

sw005320:
Thanks!
Year, we can make this and chime5 as our remaining challenges.
We need some research trials of how we make the e2e system work in this adverse environment.

PR #1479, opened 2019-12-27
Transformer NMT

hirofumi0810:
- Add Transformer-baed NMT model
- Visualize token strings (not index) for attention plotting

### DIFF
--- a/egs/must_c/mt1/conf/tuning/train_pytorch_transformer.yaml
+++ b/egs/must_c/mt1/conf/tuning/train_pytorch_transformer.yaml
@@ -1,0 +1,42 @@
+# network architecture
+# encoder related
+elayers: 6
+eunits: 2048
+# decoder related
+dlayers: 6
+dunits: 2048
+# attention related
+adim: 256
+aheads: 4
+tie-src-tgt-embedding: false
+tie-classifier: false
+
+# label smoothing
+lsm-weight: 0.1
+
+# minibatch related
+batch-size: 64  # for BPE
+maxlen-in: 100  # if input length  > maxlen-in, batchsize is automatically reduced
+maxlen-out: 100 # if output length > maxlen-out, batchsize is automatically reduced
+
+# optimization related
+sortagrad: 0 # Feed samples from shortest to longest ; -1: enabled for all epochs, 0: disabled, other: enabled for 'other' epochs
+opt: noam
+accum-grad: 1
+grad-clip: 5
+patience: 0
+epochs: 200
+dropout-rate: 0.1
+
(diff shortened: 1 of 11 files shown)


hirofumi0810:
New plotting function generate a figure like this:
ted_00767_0032400_0041270 decoder decoders 0 src_attn ep 200

sw005320:
> New plotting function generates a figure like this:
> ted_00767_0032400_0041270 decoder decoders 0 src_attn ep 200

it looks very good.
asr output or tts input may follow the same way.
btw, is it possible to make the input part better looking by avoiding overlaps?

hirofumi0810:
@sw005320 It is possible to rotate the source tokens, but it takes some time to check it.

egs/must_c/mt1/conf/tuning/train_pytorch_transformer.yaml
```diff
@@ -0,0 +1,42 @@
...
+lsm-weight: 0.1
+
+# minibatch related
+batch-size: 64  # for BPE
+maxlen-in: 100  # if input length  > maxlen-in, batchsize is automatically reduced
+maxlen-out: 100 # if output length > maxlen-out, batchsize is automatically reduced
+
+# optimization related
+sortagrad: 0 # Feed samples from shortest to longest ; -1: enabled for all epochs, 0: disabled, other: enabled for 'other' epochs
+opt: noam
```
butsugiri: Question: what are noam and transformer-lr for? Is vanilla Adam not good?
In Transformer-NMT, I think that we usually use Adam with warmup https://github.com/pytorch/fairseq/tree/master/examples/scaling_nmt

hirofumi0810: Noam in ESPnet means Adam with lr scheduling based on the original Transformer paper.

hirofumi0810: And also the hyperparameter such as beta is modified.

butsugiri: I see. Do you think adding a description for that is a good idea?

sw005320: good idea! maybe, we should add this in the main readme. I'll do it with another PR.

sw005320:
> @sw005320 It is possible to rotate the source tokens, but it takes some time to check it.

ok, you don't have to do it in this PR, but it would be great if you do it in the near future.

sw005320:
thanks!

PR #1481, opened 2019-12-27
Fix test_multi_spkr

simpleoier:
Modified the test_multi_spkr file for recognition. Previous fixed ground truth text comparison may cause assertions due to various reasons, which was highly dependent on initialization, network inner computation and so on. Now only check whether the recognition function works without interrupt and ensure the output is a sequence of characters.

### DIFF
--- a/test/test_multi_spkrs.py
+++ b/test/test_multi_spkrs.py
@@ -6,6 +6,7 @@
 import argparse
 import importlib
 import numpy
+import re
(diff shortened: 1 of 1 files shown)


simpleoier:
The error comes from the downloading process of Google drive. Can someone rebuild this PR?

sw005320:
Thanks for your modification!

Issue #1459, opened 2019-12-21
Combining multiple datasets to train a better pretrained model?


AndroYD84:
I'm really impressed with ESPnet results, it's the most accurate speech to text solution I've tried after Google Cloud, I got the best overall results using the Librispeech pretrained model, however, occasionally Common Voice (my second favourite) could detect some words that Librispeech couldn't, so, what do you think about combining more datasets together (ie. Librispeech + Common Voice) and train on them all? Couldn't that potentially deliver a pretrained model that surpasses all others? It would be great to have something comparable to Google Cloud's result, in fact, this is the very first time I've seen an open source solution that is almost capable of that. I was tempted to try it myself but I'd like to know exactly how you've trained your own as I've read often in the issues that others have tried to reproduce your results but they weren't just as accurate as yours. Thanks for the amazing repo.

sw005320:
Thanks for your encouraging words!
Indeed, ESPnet can easily combine the multiple training data (but the use of pre-training model is not officially supported) by simply using `utils/combine_data.sh`.
You can find some examples in https://github.com/espnet/espnet/blob/master/egs/li10/asr1/run.sh and https://github.com/espnet/espnet/blob/master/egs/fisher_swbd/asr1/run.sh

@ftshijt, can you also think of making a PR of your super multilingual recipe?
This would also be a good example.

s-tomic:
Did you manage to combine Common voice with Libri @AndroYD84 ?
I was trying to do so also, just common voice alone was failing on me in stage 1 due to Kaldi error.

sw005320:
We did not encounter the problem of simply combining librispeech and common voice data (see https://github.com/espnet/espnet/blob/master/egs/li42/asr1/run.sh#L208-L217)

nayankote:
@djo-koconi What exactly was the error and how did you fix it? I too am training on common voice and it is stuck at stage 1. It seems to run forever after executing steps/make_fbank_pitch.sh. 
Thanks in advance.

PR #1485, opened 2020-01-02
Update for Common Voice recipe & Multilingual training recipe

### LINKED ISSUE #1357: Sclite "Segmentation fault" encountered for Fr-Commonvoice
I recently found that the sclite for WER get "segmentation fault" when computing the decoded french trn files.

The error message is as follows:

`tools/espnet/utils/score_sclite.sh: line 121: 14394 Segmentation fault      sclite -r ${dir}/ref.trn trn -h ${dir}/hyp.trn trn -i rm -o all stdout > ${dir}/result.txt`

The two relating trn file is in the following links (for you to check the error):
https://drive.google.com/file/d/1vh7TbR_zHZI5yR3F3zTEufEIjID2VZS1/view?usp=sharing

ftshijt:
1. update RESULTS.md with the new setting (it seems the WERs drop much comparing to the original one, the reason should be the text-split. The previous train-dev-test split does not consider the text split, all the text in dev&test were in the train set.)
2. remove text with "{}" (only in French data
Fixes #1357 )
3. add multi-lingual recipe for #1459


sw005320:
> 3. add multi-lingual recipe for #1459

This should not be under the commonvoice directory.
It's better to make a new recipe directory (e.g., `li35`).

ftshijt:
> > 1. add multi-lingual recipe for #1459
> 
> This should not be under the commonvoice directory.
> It's better to make a new recipe directory (e.g., `li35`).

Got it, I would open up another PR regarding this

egs/li42/asr1/conf/lm.yaml
```diff
@@ -0,0 +1,8 @@
+layer: 2
```
sw005320: Is it used?
(I don't think so).

ftshijt:
> 1. can you upload all models?
> 2. add the li42 data descriptions to https://github.com/espnet/espnet/blob/master/egs/README.md

For the models, do I need to upload to your spack or simply provide it in my google drive?

sw005320:
> For the models, do I need to upload to your spack or simply provide it in my google drive?

Either is fine.

egs/commonvoice/asr1/run.sh
```diff
@@ -100,11 +100,9 @@ if [ ${stage} -le 1 ] && [ ${stop_stage} -ge 1 ]; then
     for x in ${train_set} ${recog_set}; do
       # Remove features with too long frames in training data
       max_len=3000
```
sw005320: Are `max_len=3000` and `max_chars=10000` appropriate?
Especially `max_chars=10000` looks very strange.

ftshijt: I mean to set a large threshold on character-level to follow the previous results (with only a 3000 frame threshold).

sw005320: why not using default one?
Some corpus includes very long sentences (by mistakes), and 10,000 will break training.

ftshijt: The default one should be enough, I will change it by next commit

sw005320:
Thanks a lot!

PR #1487, opened 2020-01-03
[ESPnet2] Tensorboard, --num_iters_per_epoch, etc.

kamo-naoyuki:
- Support Tensorboard
  - Always enable.
  - Saving in ${traindir}/tensorboard
- Add options of `--cudnn_enabled`, `--cudnn_ benchmark`, `cudnn_deterministic`
- ~~Fix ci: flake8 didn't work due to typo for espnet2.~~
   - I'll fix this in another PR.
- Added supporting distributed training (almost done, but not all)
  - I won't finish this feature in this PR because I don't have multiple machines now.
- Change `checkpoint` and unity `--resume_path` and `--resume_epoch` to `--resume <true|false>`

  - Before

  ```
  1epoch/
    - model.pth
    - reporter.pth
    - optim.pth
  ```

    - After

  ```
  checkpoint.pth  # Includes the latest model, reporter, and all states
  1epoch.pth  # Includes model states only
  2epoch.pth
  ```

  Note that the model state is saved twice for each epochs.

- Add `--num_iters_per_epoch` option

  ```python
  for iepoch in range(1, max_epoch):
      # build_iter() returns a batch-iterator with `--num_iters_per_epoch` samples.
      train_one_epoch(train_iter_factory.build_iter(iepoch))
      # eval_iters are same in all epochs now.
      eval_one_epoch(eval_iter_factory.build_iter(iepoch))
      optim_scheduler_step()
      save_checkpoint()
      save_model()
      check_early_stopping()
  ```

  This option changes the definition of `epoch` to shorten the interval between training and evaluation.  `build_iter` simply changes the samples used for each epochs and give them to dataloader .  The selected samples for each epochs are scheduled to be equivalent to the following pseudo codes.

  ```python
   for iepoch in range(1, max_epoch):
        for batch in train_iter:
              train_one_batch(batch)
              if num_iterations % eval_interval == 0:
                   eval(eval_iter)
  ```

     Note that the random seed for shuffling data are decided as `epoch + args.seed`, so the order can be reproduced when resuming of process.


espnet2/train/batch_sampler.py
```diff
@@ -37,22 +34,20 @@ def build_batch_sampler(
         shape_files: Text files describing the length and dimension
             of each features. e.g. uttA 1330,80
         max_lengths: Used for "seq" mode
-        shuffle: If False, the batch are sorted by ascending order.
```
kamo-naoyuki: The options of shuffle are removed from sampler and instead of it,  the iterator factory class shuffles the samples

espnet2/torch_utils/load_pretrained_model.py
```diff
@@ -0,0 +1,52 @@
+from pathlib import Path
+from typing import Any
+from typing import Union
+
+import torch
+import torch.nn
+import torch.optim
+
+
+def load_pretrained_model(
```
kamo-naoyuki: I deleted `espnet2.train_utils.checkpoint.py`  and created  `espnet2.train_utils.load_pretrained_model.py`. load_checkpoint and save_checkpoint are removed because the procedures are simplified and such functions seem redundant.

sw005320: is this used for inference as well?
it would be better to unify all model load functions via this?

kamo-naoyuki: Sorry for not clear explanations. 
For inference, torch.load() is used simply. `load_pretrained_model` is used for training for fine tuning or distillation.

kamo-naoyuki: `load_pretrained_model`  can apply partially parameters of some parts of the model.
However, to be honest, there are practical example of fine-tuning, so you don't need to care this much. We may need to modify this in the future.

sw005320: ok

sw005320:
I read through it and it seems good.
@kan-bayashi, can you also check it?

espnet2/tasks/abs_task.py
```diff
@@ -840,14 +885,75 @@ def main(cls, args: argparse.Namespace = None, cmd: Sequence[str] = None) -> Non
...
+        )
+
+        batches = list(batch_sampler)
+        if num_batches is not None:
+            batches = batches[:num_batches]
+
+        if distributed:
+            world_size = torch.distributed.get_world_size()
+            rank = torch.distributed.get_rank()
+            batches = [batch[rank::world_size] for batch in batches]
```
kamo-naoyuki: About distributed training: Note that we can't use pytorch DistributedSampler because we need custom batch sampler to implement variable batch-size.
Indeed, DistributedSampler simply split the samples by `world_size`,

https://github.com/pytorch/pytorch/blob/master/torch/utils/data/distributed.py

so instead of it, split samples here.

espnet2/train/trainer.py
```diff
@@ -127,15 +162,11 @@ def train_one_epoch(
...
-                    if v is not None
-                    else None
-                    for k, v in stats.items()
-                }
-                weight = weight.sum()
+            if ngpu > 1 or options.distributed:
+                # Weighted averaging
+                loss = (loss * weight).sum() / weight.sum()
+                stats, weight = recursive_average(stats, weight, options.distributed)
+
```
kamo-naoyuki: About distributed training: In pytorch distributed training, the training process can be done transparently with `DistributedDataParallel`, just like `DataParallel`, because `all-reduce` operations are hidden in the wrapper class when `backward()`.

The `stats` need manual all reduce operation since it's not relevant to `backward()` operation, so I defined `recursive_average()` here.

espnet2/train/trainer.py
```diff
@@ -127,15 +162,11 @@ def train_one_epoch(
                 loss, stats, weight = data_parallel(
                     model, (), range(ngpu), module_kwargs=batch
```
kamo-naoyuki: Some developers in pytorch and fairseq says that DataParallel, 1-Node-n-GPU with multithread, is not efficient and use DistributedDataParallel, m-Node-mn-GPU with multiprocess.
Multithread has disadvantage due to Global-interpreter-lock of Python interpreter in general.
We might remove the supporting DataParallel and unify it to DistributedDataParallel.

espnet2/train/trainer.py
```diff
@@ -83,7 +118,7 @@ def add_arguments(cls, parser: argparse.ArgumentParser):
     @classmethod
     def train_one_epoch(
         cls,
-        model: AbsE2E,
+        model: torch.nn.Module,
```
kamo-naoyuki: I loosen the type check for AbsE2E because it can be also DistributedDataParallel.

espnet2/tasks/abs_task.py
```diff
@@ -330,44 +394,46 @@ def get_parser(cls) -> configargparse.ArgumentParser:
...
             type=str2bool,
             default=False,
             help="Just only iterating data loading without "
             "model forwarding and training",
         )
+        group.add_argument(
+            "--resume",
+            type=str2bool,
+            default=False,
+            help="Enable resuming if checkpoint is existing",
```
kan-bayashi: Now we cannot specify the checkpoint, right?

espnet2/tasks/abs_task.py
```diff
@@ -620,88 +686,66 @@ def main(cls, args: argparse.Namespace = None, cmd: Sequence[str] = None) -> Non
...
+
+        # 1. Set random-seed
+        set_all_random_seed(args.seed)
+        torch.backends.cudnn.enabled = args.cudnn_enabled
+        torch.backends.cudnn.benchmark = args.cudnn_benchmark
+        torch.backends.cudnn.deterministic = args.cudnn_deterministic
+
+        # 2. Build iterator factories
+        train_iter_factory = cls.build_iter_factory(
+            args=args,
```
kan-bayashi: args is a little bit confusing since this function has other arguments.

espnet2/tasks/abs_task.py
```diff
@@ -840,14 +888,75 @@ def main(cls, args: argparse.Namespace = None, cmd: Sequence[str] = None) -> Non
                 nbest=args.keep_n_best_checkpoints,
             )
 
+    @classmethod
+    def build_iter_factory(
+        cls,
+        args: argparse.Namespace,
```
kan-bayashi: As I mentioned above, args is a little bit confusing.
Is it ok to convert args to separated keyword arguments?

espnet2/train/epoch_iter_factory.py
```diff
@@ -0,0 +1,96 @@
...
+    def build_iter(self, epoch: int, shuffle: bool = None) -> DataLoader:
+        raise NotImplementedError
+
+
+class EpochIterFactory(AbsIterFactory):
+    def __init__(
+        self,
+        dataset,
+        batches: Sequence[Sequence[Any]],
+        num_iters_per_epoch: int = None,
```
kan-bayashi: If the len(batches) = 4, 1 epoch w/ num_iters_per_epoch = None and 4 epochs w/ num_iters_per_epoch = 1 will be exactly the same?

espnet2/train/trainer.py
kan-bayashi: I think it is better to add NOTE about your comments.

espnet2/tasks/abs_task.py
```diff
@@ -228,20 +257,55 @@ def get_parser(cls) -> configargparse.ArgumentParser:
...
-            "training phase. If None is given, it is decided according the number "
-            "of training samples automatically .",
+            "--dist_backend", default="nccl", type=str, help="distributed backend",
+        )
+        group.add_argument(
+            "--dist_init_method",
+            type=str,
+            default="env://",
+            help='if init_method="env://", env values of "MASTER_PORT", "MASTER_ADDR", '
+            '"WORLD_SIZE", and "RANK" are referred.',
```
kan-bayashi: By default, rank is passed as —local_rank.
(This is just a note)
https://github.com/pytorch/pytorch/blob/34561dadcddf6ce3c76daf16f09a09adc9c7b73b/torch/distributed/launch.py#L179

espnet2/tasks/abs_task.py
kamo-naoyuki: Yes. I concluded that many people wouldn't do and simplicity is more important.

espnet2/tasks/abs_task.py
kamo-naoyuki: I agree this is bad way, but I have no idea about this.

My intention of giving args is  to make `build_iter_factory` abstracted and enable to be overwritten.

```python
class SomeTask(AbsTask):
    def build_iter_factory(cls, args, ...):
       ...
```

Basically, the task class assumed same data-iterator among many tasks, but
some tasks, e.g. LM training with Truncated BPTT,  require splitting the sequence into pieces and this iterator doesn't support such flexibility. I'd like to add such features.

If args is converted explicit names, we can't give additional arguments for such unseen iterator.

espnet2/train/epoch_iter_factory.py
kamo-naoyuki: Yes if the implementation is correct.

espnet2/tasks/abs_task.py
kamo-naoyuki: Well, I need more time to think about this, but in this PR, I'd like to keep this shape.

kan-bayashi: I see. OK.

kamo-naoyuki:
if looks good, then please merge. Ill send next

sw005320:
> if looks good, then please merge. Ill send next

OK, I just merged it.
Thanks!

PR #1491, opened 2020-01-08
[ESPnet2] Fix ci of flake8 part

kamo-naoyuki:
Flake8 was not applied when ci by mistake.

### DIFF
--- a/espnet2/asr/encoder/transformer_encoder.py
+++ b/espnet2/asr/encoder/transformer_encoder.py
@@ -2,7 +2,8 @@
 #  Apache 2.0  (http://www.apache.org/licenses/LICENSE-2.0)
 
 """Encoder definition."""
-from typing import Tuple, Optional
+from typing import Optional
+from typing import Tuple
 
 import torch
 from typeguard import check_argument_types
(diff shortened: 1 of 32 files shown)


espnet2/asr/encoder/transformer_encoder.py
```diff
@@ -15,7 +16,7 @@
 from espnet.nets.pytorch_backend.transformer.multi_layer_conv import Conv1dLinear
 from espnet.nets.pytorch_backend.transformer.multi_layer_conv import MultiLayeredConv1d
 from espnet.nets.pytorch_backend.transformer.positionwise_feed_forward import (
-    PositionwiseFeedForward,
+    PositionwiseFeedForward,  # noqa: H301
```
kamo-naoyuki: This comma, `PositionwiseFeedForward,` is added by Black.
Black style allows multiple import and they say that even one-import statement should end with comma if parenthesis is used.

https://github.com/psf/black/issues/1220

but it conflicts with hacking-extension. `hacking` sends error if it ends with comma.

I agree almost that Black developer said, and  I also think this hacking rule is reasonable for me.
Thus, finally, I decided to exclude them one by one.

kan-bayashi:
Ready to merge?

kamo-naoyuki:
yes

PR #1501, opened 2020-01-11
[ESPnet2]  remove model_size arguments from noam lr scheduler

kamo-naoyuki:
Originally "model_size" is derived from the model, when instantiation of this class, but this parameter is just argument and it's not related to actual model size in the current implementation in ESPnet2.

https://github.com/espnet/espnet/blob/e88a477cb72be7e5a03595ead5c233f8d211f6b6/espnet/asr/pytorch_backend/asr.py#L431-L433

Indeed, this model_size is just multiplied to learning rate and this parameter seems redundant. I removed this parameter, so the learning rate must be multiplied `model_size ** 0.5` from the previous configuration.

- Before PR, base lr is:
 `lr / model_size ** 0.5`

- After PR, Base lr is:
   `lr`

Where lr is the optimizer's lr.

### DIFF
--- a/espnet2/schedulers/noam_lr.py
+++ b/espnet2/schedulers/noam_lr.py
@@ -14,14 +14,11 @@ class NoamLR(_LRScheduler, AbsBatchStepScheduler):
     FIXME(kamo): PyTorch doesn't provide _LRScheduler as public class,
      thus the behaviour isn't guaranteed at forward PyTorch version.
 
-    TODO(kamo): "Model_size" is shared from the model, but it can't be now.
-
     """
 
     def __init__(
         self,
         optimizer: torch.optim.Optimizer,
-        model_size: Union[int, float] = 320,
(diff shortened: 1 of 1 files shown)


espnet2/schedulers/noam_lr.py
```diff
@@ -39,8 +35,6 @@ def __init__(
     def get_lr(self):
         step_num = self.last_epoch + 1
         return [
-            lr
-            * self.model_size ** -0.5
```
sw005320: Does this mean that `lr` would be changed from the previous one by a factor of `self.model_size ** -0.5`?
Does this modification change `lr` tuning range?

kamo-naoyuki: Sorry, this implementation had a bug, that is not my original intention of PR. The lr will be multiplied by `self.model_size ** -0.5` at every steps.

kamo-naoyuki:
I'll reopen.

PR #1506, opened 2020-01-12
[ESPnet2] distributed training

kamo-naoyuki:
Anyone test this.

This PR is too complex to explain, so I'd like to show the examples only:

See: https://github.com/espnet/espnet/wiki/About-distributed-training

I added `drop_last` arguments for Batch Sampler. 
  - ~~For training, drop_last is true. In Distributed training, mini-batch is divided by worldsize and each worker must have 1 or more batch-size. To avoid 0-batchsize, drop_last=true for training~~.
  - For training, drop_last is false.
 - For validation, drop_last is false. ~~Validation mode perform only at RANK==0 worker.~~
 - For inference, drop_last is false.
 - By default,  drop_last is false.


sw005320:
@kan-bayashi, can you test this PR?

@kamo-naoyuki, do you have any suggestions on how to use it in the grid engine?

kamo-naoyuki:
@kan-bayashi Please wait to use if you'll, I'll test this a bit.

kan-bayashi:
OK. And I have a question about multi-node distributed training.
My servers are not connected with a strong cable (e.g. InfiniBand), just connected w/ 1Gb or 10Gb Ethernet cable.
Are there any requirements for the connection b/w servers?

kamo-naoyuki:
@kan-bayashi 

I tested with 10gb  Ethernet now and Ib is not required.
pytorch document mentioned it here: https://pytorch.org/docs/stable/distributed.html#which-backend-to-use 
I didn't measure the performance.

kamo-naoyuki:
> @kamo-naoyuki, do you have any suggestions on how to use it in the grid engine?

Firstly, I'd like to note:

- **PyTorch distributed can be done with launching python normally at multiple hosts simply, e.g. via ssh**. There are no requirements except for Ethernet or Infiniband. You don't need even Slurm or MPI originally.
- There are two mode for distributed training: `--multiprocessing_distributed` or not.  You can know about `--multiprocessing_distributed` here: https://github.com/pytorch/examples/tree/master/imagenet
    - `--multiprocessing_distributed` is default for now.
    -  On `--multiprocessing_distributed` mode, user needs to launch with `1process-per-node`. Each processes in each nodes spawn same number process to the number of GPUs.
    ```
    # --multiprocessing_distributed
    HOST1: Process1 (launched by user) -> Nprocesses (launched by `train.py` itself, N=number of GPUs)
    HOST2: Process2 (launched by user) -> Nprocesses (launched by `train.py` itself, N=number of GPUs)
   ```
    - Not multiprocessing mode assumes that all processes are launched by uses. You can also use [pytorch.distributed.launch](https://github.com/pytorch/pytorch/blob/master/torch/distributed/launch.py) with this mode.
- Single-node multiple GPUs with `--multiprocessing_distributed`  is more important in this PR for practical situation.
    - **It's significantly faster than DataParallel in my environment**.
- I implemented two casual way to launch for each nodes: --launcher slurm or mpi. MPI assumes  `--multiprocessing_distributed` and slurm supports both mode. 
   - I wrote the example for the above box.
- Note that ` --launcher=mpi` doesn't mean `backend=mpi` and MPI used for assignment of RANK and WORLD_SIZE only. The NCCL backend can be used with ` --launcher=mpi`.

` --launcher=mpi`  is my original idea and it's prepared for not Slurm environment actually. 
SGE or  torque support MPI, but I haven't used MPI with SGE, so I don't know how useful it is. 

Please ask me if there are not clear point. Maybe It's helpful for the other people.

kamo-naoyuki:
@kan-bayashi O.K. I have tested on slurm and mpirun mode. Please test this if you have time( I think this PR is not high priority matter)

ShigekiKarita:
FYI when I previously made prototype distributed training. I also made simple wrapper script for slurm and others https://github.com/espnet/espnet/blob/310dee39174513ce6a21ce0b6183f93e9eb03a68/utils/distributed.sh
```bash
    distributed.sh --gpu ${ngpu} --node ${nnode} \
        ${expdir}/train.log \
        asr_train.py \
        --ngpu ${ngpu} \
        --backend ${backend} \
        --outdir ${expdir}/results ...
```
If slurm is available, it take advantage of that. Otherwise it will allocate greedy single gpu jobs for `$ngpu * $nnode` times.

kamo-naoyuki:
@ShigekiKarita 

Your script corresponds to `not  multiprocessing_distributed` pattern and I don't think it can work without Slurm because

1. By NCCL restriction, not only each device id used by each processes, all device id within a node is necessary to be visible by each processes in a node.
1.  Thus, we need to give the all device ids to be used to all process in a node.
    - Slurm can give this information using ${CUDA_VISIBLE_DEVICES}, no this is no problem.
    - But, without slurm, your code can't tell this information because the jobs are launched one by one.


 `multiprocessing_distributed` is more easy.  The requirements is only `1task-per-node`.

kamo-naoyuki:
Sorry, I misunderstood. Forget my comment.

kamo-naoyuki:
I implemented similar utility as https://github.com/espnet/espnet/blob/310dee39174513ce6a21ce0b6183f93e9eb03a68/utils/distributed.sh

The usage is here:

https://github.com/espnet/espnet/blob/395fac59d68962e840fb903f677612c5c1102d36/egs2/TEMPLATE/asr1/asr.sh#L650-L676

The multi-node mode can work only with Slurm now.
Theoretically, it is not impossible with SGE, but very hard.  I wrote the requirements here.

https://github.com/espnet/espnet/blob/395fac59d68962e840fb903f677612c5c1102d36/espnet2/bin/launch.py#L182-L188

SGE doesn't support exclusive allocation for GPU devices, so we may need some tricks like `espnet/utils/free-gpu.sh`. MPI with SGE may be more smart.

I think SGE is too obsolete and recommend all to use Slurm or Torque if you really want to tackle distributed training.

gs5264:
Can you tell me what is your version of pytorch? Thanks a lot

kamo-naoyuki:
@gs5264  I'm using 1.3. 1.0.1 may work.

kamo-naoyuki:
I also implemented an option to submit jobs via SSH for distributed training. 

```bash
# 2Nodes and 4gpu-per-node
python -m espnet2.bin.launch --envfile path.sh --log train.log \
    --host host1,host2 --ngpu 4 -- python -m espnet2.bin.asr_train
```

kan-bayashi:
Sorry for late, I will test it in our slurm cluster.

kan-bayashi:
I am testing with `egs2/an4/asr1`.
I found that `collect_stats` stage allocates gpu memory and many `nj` causes OOM errors without gpu allocation through the slurm.

kan-bayashi:
Single-node multi-gpu training with slurm seems to be working well.
(via `run.sh --stage 5 --num_nodes 1 --ngpu 2`)
But multi-node single gpu per node training is not working well.
(via `run.sh --stage 5 --num_nodes 2 --ngpu 1`)
The log itself said "successfully training is done" but when I checked gpu allocation, only one node was used.

kamo-naoyuki:
> I am testing with `egs2/an4/asr1`.
> I found that `collect_stats` stage (e.g. stage 4) allocates gpu memory without gpu allocation through the slurm and many `nj` causes OOM errors .

The GPU issue was fixed by  #1526
Is OOM errors  due to your environment?

kan-bayashi:
> The GPU issue was fixed by #1526

Thanks!

> Is OOM errors due to your environment?

If we use slurm backend, all of the job in job array uses gpu 0 memory and it causes OOM errors. #1526 will fix this issue.

kan-bayashi:
> srun: Warning: can't run 1 processes on 2 nodes, setting nnodes to 1

Oh I missed this warning.
Maybe our environment problem?

kamo-naoyuki:
The default parameter differs from my slurm and such information is really helpful.

Can you try this slurm.conf?

```
# Default configuration
command sbatch --export=PATH
option time=* --time $0
option mem=* --mem-per-cpu $0
option mem=0          # Do not add anything to qsub_opts
option num_threads=* --cpus-per-task $0
option num_threads=1 --cpus-per-task 1
option num_nodes=* --nodes $0 --ntasks-per-node 1
default gpu=0
option gpu=0 -p cpu
option gpu=* -p gpu --gres=gpu:$0
# note: the --max-jobs-run option is supported as a special case
# by slurm.pl and you don't have to handle it in the config file.
```

kamo-naoyuki:
Hmm,  I couldn't understand how to reproduce the log. Maybe there are something I don't know about Slurm.

If you are stucked, Could you try `srun` directly?  I believe this can work.

```bash
% srun -c2 -N2 --ntasks-per-node 1 --gres gpu:2 python -m espnet2.bin.asr_train \
    --ngpu 2 \
    --dist_master_port 20000 \
    --dist_launcher slurm
```

kan-bayashi:
I tested the above command but got the error `NCCL WARN Connect to 192.168.1.51<51890> failed : No route to host`. It seems to be necessary to open all of the ports between nodes. I will try to modify firewall setting tomorrow.

kan-bayashi:
Finally, I could run multi-node distributed training.
The things I need to set are as follows:

1. Update PyTorch (1.0.1 -> 1.3.1, not sure I will test with other version)
2. Disable related settings in iptables (or firewall)
3. Set `export NCCL_SOCKET_IFNAME=<appropriate_interface>`
4. Add `command sbatch --export=PATH,NCCL_SOCKET_IFNAME` in `slurm.conf`.

kan-bayashi:
I found that multi-node multi-gpu distributed training causes another error.
I will investigate the reason tomorrow.

kan-bayashi:
I successfully ran the multi-node multi-gpu per node distributed training in the case of each node with 3 gpus.
In our server, the network interface name is different among the nodes, so I made `/etc/nccl.conf` in all of the nodes and set `NCCL_SOCKET_IFNAME=<each_node's_interface>` appropriately.

But I failed in the case of each node with 10 gpus due to the `NCCL WARN peer mapping resources exhausted`.
It will be happened when we use more than 8 gpus (but in my case only 2x2 gpus, I'm not sure why this is happened.)
https://devtalk.nvidia.com/default/topic/970010/cuda-programming-and-performance/cuda-peer-resources-error-when-running-on-more-than-8-k80s-aws-p2-16xlarge-/post/4994583/#4994583
Anyway, maybe this is the problem of our node hardware limitation, so the code itself is OK, I think.

egs2/TEMPLATE/asr1/asr.sh
```diff
@@ -26,6 +26,7 @@ SECONDS=0
 stage=1          # Processes starts from the specified stage.
 stop_stage=10    # Processes is stopped at the specified stage.
 ngpu=1           # The number of gpus ("0" uses cpu, otherwise use gpu).
+num_nodes=1      # The number of nodes
```
kan-bayashi: We might want to unify the naming. Why don't you use `nnode`?

kan-bayashi:
I summarized the results.

pytorch version | single-node mult-gpu | multi-node single gpu | multi-node multi-gpu | note
-- | -- | -- | -- | --
1.0.1 | ✔ | ✔ | ✔ |  
1.1 | ✔ | ✔ | ✔ | Require `pip install tensorboard future`
1.2 | ✔ | ✔ | ✔ | Require `pip install tensorboard future`
1.3.1 | ✔ | ✔ | ✔ | Require `pip install tensorboard`

Consequently, all of the versions work well.
I could not reproduce the above error https://github.com/espnet/espnet/pull/1506#issuecomment-576627357 even if we use before `slurm.conf`...
Maybe port setting is related?

kamo-naoyuki:
> I successfully ran the multi-node multi-gpu per node distributed training in the case of each node with 3 gpus. In our server, the network interface name is different among the nodes, so I made `/etc/nccl.conf` in all of the nodes and set `NCCL_SOCKET_IFNAME=<each_node's_interface>` appropriately.

This is information is really helpful! I also encountered such problem.

- So, finaly, does you need `--ntasks-per-node 1`?
- What does requires `future`? I didn't realize it.
- Maybe we should remove `--export PATH` part from slurm.conf. Slurm exports all environment variables by default if it's not specified.
- For reference, can you insert `torch.cuda.nccl.version()` in your table? 
   - I'm using pytorch=1.3.1 and `nccl=2408`. This is installed by anaconda.

egs2/TEMPLATE/asr1/asr.sh
kamo-naoyuki: Maybe num_* is more clear than n*, in my recent impression. `nnodes` is really difficult.
I would rather change ngpu to num_gpus.

sw005320:
how about the scalability with respect to the number of gpus?

kan-bayashi:
> does you need --ntasks-per-node 1?

Not needed.

> What does requires future? I didn't realize it.

This is dependency of `torch.utils.tensorboard`.
https://github.com/pytorch/pytorch/issues/22389

> Maybe we should remove --export PATH part from slurm.conf. Slurm exports all environment variables by default if it's not specified.

If we use different variable for each node (e.g. `NCCL_SOCKET_IFNAME`), `--export=ALL` might be problem.

> For reference, can you insert torch.cuda.nccl.version() in your table?
I'm using pytorch=1.3.1 and nccl=2408. This is installed by anaconda.

NCCL version 2.3.7+cuda10.0

kan-bayashi:
> how about the scalability with respect to the number of gpus?

I have not enough gpu resources to calculate the speed now...

kamo-naoyuki:
I summarized our discussion in wiki. https://github.com/espnet/espnet/wiki/About-distributed-training

@kan-bayashi You don't need to change NCCL_SOCKET_IFNAME for each nodes because it can accept multiple names.

kan-bayashi:
> You don't need to change NCCL_SOCKET_IFNAME for each nodes because it can accept multiple names.

I tested `export NCCL_SOCKET_IFNAME=en,eth,em` and worked well when our servers have various types of sockets. Thanks!

kamo-naoyuki:
Let's merge. Some notes:

-  I may change `ngpu` to `num_gpus` in the future.
-  I also added some configuration for multi nodes in conf/queue.conf and conf/pbs.conf.
   - I don't have both environment. I need someone to complete this feature.
-  I also implemented SSH based launching. I think this feature is most general pattern,  but it is still experimental.
- Indeed, the distributed training has too many branch, e.g. single node distributed, Slurm mode, MPI mode, etc. Several bugs may remain there. It's very hard to maintain without CI... 
- torch.FFT doesn't work   with DataParallel mulit GPUs if pytorch>1.1, https://github.com/pytorch/pytorch/issues/31412, 
and current 1.4 doesn't include the bug fix, so we should provide single node distributed quickly.

kamo-naoyuki:
I found a bug

kamo-naoyuki:
The gradients was not calculated correctly and it's fixed now.

kamo-naoyuki:
I wonder whether to add NCCL_SOCKET_IFNAME to path.sh. 
It is too strong assumption to user's environment, but almost all people might encounter this issue, so I decided to add it.

kamo-naoyuki:
O.K. Please merge.
`all_reduse()` is used multiple times for gathering statistics at each forwarding and it might takes times. I'll check it when I have times.

kan-bayashi:
Once the conflict is resolved, I will merge this PR.

kamo-naoyuki:
o.k.

kan-bayashi:
Thanks a lot!

PR #1509, opened 2020-01-13
[ESPnet2] Implement --max_cache_size option

kamo-naoyuki:
Refer: 

https://github.com/ptrblck/pytorch_misc/blob/master/shared_dict.py

I'm not sure that shared memory by Manager is really efficient, but I couldn't find another idea.

### DIFF
--- a/espnet2/tasks/abs_task.py
+++ b/espnet2/tasks/abs_task.py
@@ -16,6 +16,7 @@ from typing import Tuple
 from typing import Union
 
 import configargparse
+import humanfriendly
 import numpy as np
 import torch
 import torch.nn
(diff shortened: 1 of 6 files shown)


sw005320:
Very interesting.
Please let me know how we use this.

kamo-naoyuki:
> Please let me know how we use this.

Just append --max_cache_size:

```bash
python -m espnet2.bin.asr_train --max_cache_size 10G
```

Some notes:

- I used `Manager` for sharing memory between workers: https://docs.python.org/3/library/multiprocessing.html#proxy-objects
  - I don't understand how it work correctly with pytorch tensor.
- Train dataset and validation dataset has `Manager` respectively.
  - By default, 5% of max_cache_size is assigned for valiation dataset.

PR #1513, opened 2020-01-14
[ESPnet2] Support "segments" file in asr.sh and tts.sh

kamo-naoyuki:
(#1504)

CSJ, CHIME5, and many tasks distributes "unsegmented" wav files in raw corpus, so we need to split them some where.
Espnet2 generates segmented wav files using "segments" if feats_type=raw,  while espnet1 and kaldi creates it as fbank.

### DIFF
--- a/egs2/TEMPLATE/asr1/asr.sh
+++ b/egs2/TEMPLATE/asr1/asr.sh
@@ -299,7 +299,7 @@ if [ ${stage} -le 2 ] && [ ${stop_stage} -ge 2 ]; then
             fi
             # shellcheck disable=SC2086
             scripts/audio/format_wav_scp.sh --nj "${nj}" --cmd "${train_cmd}" \
-                --audio-format "${audio_format}" --fs "${fs}" ${opts} \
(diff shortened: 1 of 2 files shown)


egs2/TEMPLATE/asr1/asr.sh
```diff
@@ -293,8 +293,13 @@ if [ ${stage} -le 2 ] && [ ${stop_stage} -ge 2 ]; then
...
         for dset in "${train_set}" "${dev_set}" ${eval_sets}; do
             utils/copy_data_dir.sh data/"${dset}" "${data_feats}/${dset}"
+            _opts=
+            if [ -e data/"${dset}"/segments ]; then
+                _opts+="--segments data/${dset}/segments "
+            fi
+            # shellcheck disable=SC2086
             scripts/audio/format_wav_scp.sh --nj "${nj}" --cmd "${train_cmd}" \
-                --audio-format "${audio_format}" --fs "${fs}" \
+                --audio-format "${audio_format}" --fs "${fs}" ${opts} \
```
ftshijt: Hi, I think that the ${opts} should be ${_opt} here

PR #1514, opened 2020-01-15
Use Markdown table instead for WER results

lijunzh:
By looking at the cotent in the snippets, I think the original intention is to use the Markdown table.


sw005320:
oh, yes.
thanks!

PR #1515, opened 2020-01-17
Fix bug in MT training script

hirofumi0810:
- Fix bug for slow convergence of Transformer MT
- Update training scripts of all tasks to keep consistency
- Update Must-C results
- Upload 16 Transformer models


espnet/bin/mt_train.py
```diff
@@ -222,6 +240,9 @@ def main(cmd_args):
             else:
                 ngpu = len(p.stderr.decode().split('\n')) - 1
     else:
+        if is_torch_1_2_plus:
+            assert args.ngpu == 1, "There are some bugs with multi-GPU processing in PyTorch 1.2+" \
```
ShigekiKarita: because this bug only related to cudnn RNN (flatten params?), this is annoying for other nnets.  I think it can be logging.info or something just useful for users

hirofumi0810:
After this change, the results are updated as follows:

|model                   |De   |Pt   |Fr   |Es   |Ro   |Ru   |Nl   |It   |
|------------------------|-----|-----|-----|-----|-----|-----|-----|-----|
|[[Di Gangi et al.]](https://www.isca-speech.org/archive/Interspeech_2019/pdfs/3045.pdf)|25.30|31.10|35.50|29.90|22.60|14.00|30.30|25.80|
|Before|26.07|32.13|38.44|32.09|24.83|16.98|31.31|26.80|
|After|27.63|33.34|38.58|32.61|25.92|18.40|32.08|27.68|

All model are Transformer (joint BPE8k, lc.rm->tc).

sw005320:
This is really good!
I think we should these results in the main `README.md`!
Now NMT is in good shape.
Thanks, @hirofumi0810!

PR #1525, opened 2020-01-21
[ESPnet2 ] Fix --accum_grad

kamo-naoyuki:


kan-bayashi:
Thanks for your quick fixing!

PR #1526, opened 2020-01-21
[ESPnet2] Fix bug of using GPU even if CPU mode

kamo-naoyuki:
DataLoaders sent tesnor to GPU memory  because pin_memory=True.


kan-bayashi:
Thank you for quick fixing!
(I wrongly merged #1525 since I confused with this PR...)

PR #1531, opened 2020-01-26
[ESPnet2] Remove short data

kamo-naoyuki:
We discussed at #1504 
I appended a stage to remove short data in asr.sh and tts.sh both.

- Remove short data only and and not removing long data.
- My intention is not limitation for sequence length here, but removing invalid sequence having about 0 length.
   - We can't correctly get  the length of tokens before `collect_stats` mode. 
   - Instead of this I'll also add the limitation of length in BatchSampler in next PR. 
-  `--lm_train_text` is removed.
   - This is a bug not related to this PR and my original intention is that `--srctexts` is also used for LM training.
   - I'm not sure why I did so.


egs2/TEMPLATE/tts1/tts.sh
```diff
@@ -255,43 +255,94 @@ if [ ${stage} -le 2 ] && [ ${stop_stage} -ge 2 ]; then
...
 
-token_list="data/token_list/${trans_type}/tokens.txt"
 if [ ${stage} -le 3 ] && [ ${stop_stage} -ge 3 ]; then
-    log "Stage 3: Generate character level token_list from ${srctexts}"
-    mkdir -p "$(dirname ${token_list})"
+    log "Stage 3: Remove short data: ${data_feats}/* -> ${data_feats}/*_fix"
+
+    for dset in "${train_set}" "${dev_set}" ${eval_sets}; do
+        # Copy data dir
+        utils/copy_data_dir.sh "${data_feats}/${dset}" "${data_feats}/${dset}_fix"
```
sw005320: using the `_fix` suffix for all data looks ugly.
How about moving the original data under the data directory (e.g.,  `${data_feats}/${dset}/org/${dset}`) and rename the fixed directory with `${data_feats}/${dset}`

kamo-naoyuki:
btw is it permitted to change

kamo-naoyuki:
As my philosophy, I'd rather not to move and remove intermediate file for shell based experiments.
It's really un-debuggable and it sometimes causes unexpected behaviour when resuming middle stages. 

I just changed the directory name to`${data_feats}/org` at stage2.

sw005320:
OK, I just accept your philosophy, but then, I'm expecting that we'll have a lot of `data/*_xxx_yyy_zzz` (but this is one style and I still accept it). Another possibility is to use `org` or `org2` etc. and make the final directory name simple, as you mention. I'm fine with any of them.

kamo-naoyuki:
Is it good to you now?
Anyway, It's not time to worry about creation of a lot of duplication of such directories. We should change the style in each case.
(I think we'll soon encounter this issue. )

sw005320:
Yes, of course, I'm OK with it.

kamo-naoyuki:
Please merge

kamo-naoyuki:
Please merge. I want to finish current PRs soon. I'll be busy.

PR #1533, opened 2020-01-27
[ESPnet2] fix rnn_type when bidirectional is used

kamo-naoyuki:
(Cherry pick from #1522)
GRU is always used if bidirectional mode.


sw005320:
Thanks!

PR #1539, opened 2020-01-28
[ESPnet2] Fix transformer

kamo-naoyuki:
I just didn't create subsequent mask.
@ShigekiKarita Is this correct?
acc
lr_0


kamo-naoyuki:
@ShigekiKarita Could you take a look? I'll be busy at this month, so  I'd like to finish quickly.

ShigekiKarita:
@kamo-naoyuki If you add tests like this, that would be great https://github.com/espnet/espnet/blob/5e191ecf6068136235f649543661a8faf7300f5b/test/test_e2e_asr_transformer.py#L110-L122

kamo-naoyuki:
Did you read my implementation? The test  code checks `model.decoder.decoders[0].self_attn` and I just borrowed your module, so already it is tested in the previous test code.

ShigekiKarita:
OK LGTM!

2020年2月1日(土) 23:52 Naoyuki Kamo <notifications@github.com>:

PR #1540, opened 2020-01-28
[ESPNet2] Minor fix for CSJ recipe

kamo-naoyuki:
@YosukeHiguchi   "$@" is different from "$*".


YosukeHiguchi:
@kamo-naoyuki 
Sorry, thank you for the fix.

PR #1541, opened 2020-01-28
[ESPnet2] Use load_scp_sequential in formart_wav_scp.py

kamo-naoyuki:
load_scp_sequential() is faster than load_scp for segmentation.

@YosukeHiguchi @ftshijt  I think  the stage of formart_wav_scp.py took much times. Sorry for wasting your resource.


sw005320:
Thanks!

PR #1543, opened 2020-01-31
Fix IWSLT16 Script Permission

butsugiri:
This PR adds `execute` permission to scripts under `local`.
I am sorry that I forgot to add the permission in the original PR.


sw005320:
Thanks for your quick fix!

PR #1545, opened 2020-02-01
Add Mboshi-French corpus

hirofumi0810:


utils/update_json.sh
```diff
@@ -0,0 +1,87 @@
+#!/bin/bash
```
sw005320: It's better to add a test for `utils/update_json.sh`.
Are you also planning to use `utils/update_json.sh` instead of `local/update_json.sh` in the other recipe? (which is great)

hirofumi0810: @sw005320 Yes. After this PR, I will remove the existing bash files under `local` in each recipe.

sw005320:
LGTM.
Is it ready for merge, @hirofumi0810?
Also, will you upload the model later?

hirofumi0810:
> LGTM.
> Is it ready for merge, @hirofumi0810?
> Also, will you upload the model later?

@sw005320 It's ready to be merged. I will upload How2 models next.

PR #1546, opened 2020-02-02
[ESPnet2] Change CTC default to builtin

kamo-naoyuki:


sw005320:
Thanks. (this means that we'd not support pytorch 1.0.0> in espnet2, right?)

PR #1550, opened 2020-02-03
[ESPnet2] Fix nj of scripts/audio/format_wav_scp.sh

kamo-naoyuki:
The maximum number of split equals to the number of lines of segments file.


sw005320:
Oh, I see.
Thanks!

PR #1551, opened 2020-02-03
Use solve() instead of  gesv() in time_warp and add new mode of "torch_interpolate"

kamo-naoyuki:
gesv() becomes deprecated at 1.1.0 and removed at torch>=1.2.0.
https://pytorch.org/docs/1.1.0/torch.html?highlight=gesv#torch.gesv

I also add test codes, but the result of time_warp between mode=PIL and sparse_image_warp looks quite different.
Is this expected behaviour? If so, I'll remove the assertion check.

### DIFF
--- a/espnet/transform/spec_augment.py
+++ b/espnet/transform/spec_augment.py
@@ -5,6 +5,7 @@ import random
 import numpy
 from PIL import Image
 from PIL.Image import BICUBIC
+import torch
 
 from espnet.transform.functional import FuncTrans
 
@@ -36,14 +37,19 @@ def time_warp(x, max_time_warp=80, inplace=False, mode="PIL"):
             return x
         return numpy.concatenate((left, right), 0)
     elif mode == "sparse_image_warp":
-        import torch
-
-        from espnet.utils import spec_augment
+        from espnet.utils.spec_augment import time_warp as _time_warp
 
         # TODO(karita): make this differentiable again
-        return spec_augment.time_warp(torch.from_numpy(x), window).numpy()
+        return _time_warp(torch.from_numpy(x), window).numpy()
+    elif mode == "torch_interpolate":
+        from espnet.utils.spec_augment import time_warp2
+
+        return time_warp2(torch.from_numpy(x), window, inplace).numpy()
+
     else:
-        raise NotImplementedError("unknown resize mode: " + mode + ", choose one from (PIL, sparse_image_warp).")
+        raise NotImplementedError(
+            "unknown resize mode: " + mode + ", choose one from (PIL, sparse_image_warp, torch_interpolate)."
+        )
 
 
 class TimeWarp(FuncTrans):
--- a/espnet/utils/spec_augment.py
+++ b/espnet/utils/spec_augment.py
@@ -25,7 +25,7 @@ LIABILITY, WHETjjHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
 OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
 SOFTWARE.
 """
-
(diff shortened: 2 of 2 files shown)


kamo-naoyuki:
I'll implement new time_warp using torch.interpolate. just a note.

sw005320:
I'm not sure because I usually don't use time-warping. @bobchennan, could you check this PR?

bobchennan:
Looks fine to me.

BTW also as I provided to yiming, we can use torch.nn.interpolate (mode=bicubic) instead of PIL. Not sure if it is faster but we don't need to include PIL in this case. 

@kamo-naoyuki You can try to compare those three. I think PIL and torch should be similar.

kamo-naoyuki:
These three are similar each other, but I don't have any ideas to test them... I removed the test codes.

I also new time_warp using torch.nn.functional.interpolate

```python
In [1]: import numpy as np                                                                        
In [2]: x = np.random.randn(1000, 80)                                                             
In [3]: from espnet.transform.spec_augment import time_warp                                       
In [4]: %timeit time_warp(x, mode="sparse_image_warp")                                            
14.9 ms ± 96.7 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)
In [5]: %timeit time_warp(x, mode="PIL")                                                          
657 µs ± 16.2 µs per loop (mean ± std. dev. of 7 runs, 1000 loops each)
In [6]: %timeit time_warp(x, mode="torch_interpolate")                                            
16.3 µs ± 624 ns per loop (mean ± std. dev. of 7 runs, 100000 loops each)
```

interpolate is significantly faster.

bobchennan:
Nice! I think we can remove `sparse_image_warp` since it is slow and we only need to warp in time. `sparse_image_warp` is more general but we don't really need it so far for specaug.

sw005320:
If @bobchennan and @kamo-naoyuki are OK with this, I'll merge this PR.

kamo-naoyuki:
Sorry! Sorry! There are bugs and the benchmark was wrong, maybe slower than PIL. I'll check it again tomorrow.

sw005320:
No problem. I'm looking forward to the update.

ShigekiKarita:
I wanna know how GPU accelerates specaug. Can you also benchmark that?

kamo-naoyuki:
```python
# k80
In [7]: %timeit time_warp2(x_t)
251 µs ± 1.08 µs per loop (mean ± std. dev. of 7 runs, 1000 loops each)
```

Speed order:
torch (GPU) > PIL > torch(CPU) >  sparse_image_warp

bobchennan:
@kamo-naoyuki could you show us other numbers?

kamo-naoyuki:
The others I reported  were correct:

```python
In [1]: import numpy as np                                                                        
In [2]: x = np.random.randn(1000, 80)                                                             
In [3]: from espnet.transform.spec_augment import time_warp                                       
In [4]: %timeit time_warp(x, mode="sparse_image_warp")                                            
14.9 ms ± 96.7 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)
In [5]: %timeit time_warp(x, mode="PIL")                                                          
657 µs ± 16.2 µs per loop (mean ± std. dev. of 7 runs, 1000 loops each)
In [6]: %timeit time_warp(x, mode="torch_interpolate")                                            
2.19 ms ± 81.1 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)
# k80
In [7]: from espnet.utils.spec_augment import time_warp
In [8]: x_t = torch.from_numpy(x).cuda()
In [9]: %timeit time_warp2(x_t)
251 µs ± 1.08 µs per loop (mean ± std. dev. of 7 runs, 1000 loops each)
```

PIL and torch gpu might be swapped by machine performance.

kamo-naoyuki:
I'll reopen PR.

PR #1565, opened 2020-02-08
[ESPnet2] Lower python requirements to 3.6

kamo-naoyuki:
ESPnet2 required python>=3.7 because 3.7 has utf-8 mode and it can release us from all encoding issues entirely. However Google colab doesn't support 3.7 yet. I think this is fatal problem, so I decided to lower the requirements.

If python3.6, the following causes UnicodeEncodeError:

```bash
# error
LC_ALL=C python -c "f = open('a.txt', 'w'); f.write('あ')"
# o.k.
LC_ALL=C python -c "f = open('a.txt', 'w', encoding='utf-8'); f.write('あ')"
```

We should always use all open() with encoding='utf-8', but I didn't change about it.
In `path.sh` we write `export LC_ALL=C`, so espnet2 always fails at everywhere for non-ascii task.


kamo-naoyuki:
I changed my mind. I add encoding="utf-8" for all open().
Note that "utf-8" is always used for encoding now even if using python<=3.6:

- open(): I set "utf-8"
- sys.stdin, sys.stdout: PYTHONIOENCODING=UTF-8
- str.decode(), str.encode(): The default is always utf-8 in python3 regardless of OS and locale.

sw005320:
Thanks! I think this is a safe and practical solution for now.

PR #1544, opened 2020-01-31
[WIP] add MT into key feature

butsugiri:
This PR adds description of MT into the `Key Features`.

I wonder if I should put MT results in the README, too (given the results that we have right now).

PR #1576, opened 2020-02-13
Add MT result

butsugiri:
This PR add MT result on IWSLT'16 corpus. (cont. from [this PR](https://github.com/espnet/espnet/pull/1544))


sw005320:
Thanks!!!

Issue #1508, opened 2020-01-13
Pure CTC decoding bug


luyizhou4:
Hi, there!

I'm currently decoding my CTC model (mtlalpha 1.0) with pure CTC mode (ctc_weight 1.0), and find that in current implementation, ctc-prefix-search will also compute \<blank\> token's next-label-score, which to my understanding shall be excluded. Details below: 

With ctc_weight=1.0, **ctc_beam** becomes the size of whole vocabulary (including \<blank\> token)

https://github.com/espnet/espnet/blob/28ce90a17148afcb36f4e593966911b9c3a6230b/espnet/nets/pytorch_backend/e2e_asr_transformer.py#L291-L296

Then \<blank\> token-id will be included in **local_best_ids**, and prefix-score computed. To my understanding, \<blank\> should be excluded here.

https://github.com/espnet/espnet/blob/28ce90a17148afcb36f4e593966911b9c3a6230b/espnet/nets/pytorch_backend/e2e_asr_transformer.py#L328-L335

sw005320:
I see.
Did you find any degradation?

@takaaki-hori, I think we should not consider the blank.
Can you check it?

takaaki-hori:
Yeah, the array `local_best_ids` should not include 0 (\<blank\>).
When ctc_weight != 1, there is no problem since the attention model gives a very low probability to \<blank\>, and `local_best_ids` does not include 0.
But, if ctc_weight = 1,  `local_best_ids` has all IDs.
So, we need to exclude 0 from `local_best_ids`, e.g., by making it with np.arange(1, odim).
It is also possible that we always set zero to the prefix probability for \<blank\> in CTCPrefixScore.
Anyway, I will handle this matter.

sw005320:
Thanks!

luyizhou4:
Thanks for reply! Excluding 0 from local_best_ids improves my CTC result from 12.54% to 12.08%, and for reference greedy decoding WER is 12.10%.

sw005320:
thanks, this is significant.

sw005320:
@takaaki-hori, could you think of fixing this?

takaaki-hori:
Yes, I am working on it now.

PR #1580, opened 2020-02-14
Fix pure CTC decoding

takaaki-hori:
Update for #1508. Changed CTC score computation so that it is always P(y_i=blank|y_1,...,y_i-1)=0


sw005320:
Thanks!
Can you check the CI error?
We may update the reference result.

takaaki-hori:
Yeah, I'm working on it, and have a little struggle with the problem that the normal and batch decoding results don't exactly match when using the test script. There is no problem when using the WSJ recipe. Now I am looking into each CTC score to figure it out.

sw005320:
Thanks!!
I think this will improve the performance!

PR #1589, opened 2020-02-17
ru_open_stt recipe bug fix

qmpzzpmq:
if "tag" is setted, since there is no variable "expname" before line 147.

so I copy the setting from librispeech


sw005320:
Thanks!

PR #1601, opened 2020-02-20
[ESPnet2]Update voxforge RESULTS

kamo-naoyuki:
- There are no gain by pitch for this task.
   - Raw is slightly better than pitch?
- As far as RNN, espnet2 won with espnet1.
   - xavier init seems good for RNN.


sw005320:
Sounds very good.
We can distribute your setup to others and redo their experiments.

PR #1602, opened 2020-02-20
Update run_rnnt.sh

qmpzzpmq:
should same with run.sh, variable _lang_ could be changed by utils/parse_options.sh

### DIFF
--- a/egs/voxforge/asr1/run_rnnt.sh
+++ b/egs/voxforge/asr1/run_rnnt.sh
@@ -19,11 +19,6 @@ lm_resume=     # Resume the LM training from snapshot
 voxforge=downloads # original data directory to be stored
 lang=it # de, en, es, fr, it, nl, pt, ru
 
-# datasets
-train_set=tr_${lang}
-train_dev=dt_${lang}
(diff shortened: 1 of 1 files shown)


egs/voxforge/asr1/run_rnnt.sh
```diff
@@ -43,6 +38,11 @@ tag="" # tag for managing experiments.
 
 . utils/parse_options.sh || exit 1;
 
+# datasets
```
sw005320: Thanks!
The [other recipe](https://github.com/espnet/espnet/blob/master/egs/librispeech/asr1/run.sh#L69) put the dataset variables after the bash environment setting
```
# Set bash to 'debug' mode, it will exit on :
# -e 'error', -u 'undefined variable', -o ... 'error in pipeline', -x 'print commands',
set -e
set -u
set -o pipefail

train_set=train_960
train_dev=dev
recog_set="test_clean test_other dev_clean dev_other"
```

So, this way is more consistent.
Could you do that?

qmpzzpmq: yes, sure

PR #1608, opened 2020-02-22
[ESPnet2] Implement chunk iterator

kamo-naoyuki:
This is preliminary PR for not seq2seq task for ESPnet2.

Now our data-iterator can  generate only a batch of sequences, so it can support only some task which inputs variable length sequence. I'd like to extend espnet for more general task, e.g. speech enhancement network often uses "fixed length to same length", and speaker recognition task typically uses a DNN of "fixed length to a vector" at training.

I implemented `ChunkIterFactory` in this PR. This iterator creates fixed length segments, which I refer it as chunk, cut from a sequence.

- Chunks are equal length for a sequence and they have half of length overlap by default like STFT. The length of chunk and shift length are parameters.
- Multiple chunk lengths can be also specified and randomly pickup a value from them for each sequences.

```python
# Select from 10,15, or 16
ChunkIterFactory(dataset, batches, batch_size, chunk_length="10,15,16")
# Select a value between 10 and 16
ChunkIterFactory(dataset, batches, batch_size, chunk_length="10-16")
```

- This iterator has some limitations unlike sequence iterator. I mean, it dynamically cuts segments from per utterance data loader, so we can't know the number of iteration counts.
- When distributed training, originally we can split data loading in each processes if using sequence type iterator and it also reduces data IO, but for this chunk iterator, we need load same data for each processes because it splits chunks after loading same sequences.
   - If really thinking about scalability of chunk type training, it's better to dump chunks as files in advance just like Kaldi.

### DIFF
--- a/espnet2/iterators/abs_iter_factory.py
+++ b/espnet2/iterators/abs_iter_factory.py
@@ -1,0 +1,8 @@
+from abc import ABC
+from abc import abstractmethod
+
+
+class AbsIterFactory(ABC):
+    @abstractmethod
+    def build_iter(self, epoch: int, shuffle: bool = None):
+        raise NotImplementedError
--- a/espnet2/iterators/chunk_iter_factory.py
+++ b/espnet2/iterators/chunk_iter_factory.py
@@ -1,0 +1,192 @@
+from typing import Any
+from typing import Dict
+from typing import Iterator
+from typing import List
+from typing import Sequence
+from typing import Tuple
+from typing import Union
+
+import numpy as np
+import torch
+from typeguard import check_argument_types
+
+from espnet2.iterators.abs_iter_factory import AbsIterFactory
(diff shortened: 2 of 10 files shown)


kamo-naoyuki:
I found a problem. I'll reopen this PR.

kamo-naoyuki:
Chunk iterator is incompatible with feats_type=raw because it changes the length of sequence, but I have no idea to deal it for now. Even though this iterator is useful for some tasks, so I decided to merge this.

sw005320:
Probably, we need some more considerations for the on-the-fly feature extraction/text processing during training since we'd not easily get an access to the sequence length. I think we need some two-pass procedure??? These problems also happened in the sentence pruning preprocessing.

kamo-naoyuki:
Anyway I'll merge this PR and I'll send some tweak PR.

PR #1610, opened 2020-02-24
[ESPnet2] Renaming

kamo-naoyuki:
Just renaming

- E2E -> ESPnetModel
- asr_decode, tts_decode -> asr_inference, tts_inference
- --e2e_conf -> --model_conf


sw005320:
LGTM!
I actually like such changes are done in espnet2 (it is very hard to do it in espent1).
@kan-bayashi, do you agree with this change?

PR #1611, opened 2020-02-24
Support markdown-table for sphinx

kamo-naoyuki:


sw005320:
I just set some labels and milestones for our releasement.

PR #1623, opened 2020-02-27
Embed defaultlm

qmpzzpmq:
add to parameter to control embed unit in default rnnlm, method same with transformer LM.


qmpzzpmq:
I don't know what's going on with
https://github.com/espnet/espnet/blob/8db0c2996486a763452beccfb835b8ead84d8860/test/test_e2e_compatibility.py#L40
https://github.com/espnet/espnet/blob/8db0c2996486a763452beccfb835b8ead84d8860/test/test_e2e_compatibility.py#L74
https://github.com/espnet/espnet/blob/8db0c2996486a763452beccfb835b8ead84d8860/test/test_e2e_compatibility.py#L50
it seems a bug with downloading pretrain model?

qmpzzpmq:
I don't know what's going on with 
https://github.com/espnet/espnet/blob/8db0c2996486a763452beccfb835b8ead84d8860/test/test_e2e_compatibility.py#L40
https://github.com/espnet/espnet/blob/8db0c2996486a763452beccfb835b8ead84d8860/test/test_e2e_compatibility.py#L74
https://github.com/espnet/espnet/blob/8db0c2996486a763452beccfb835b8ead84d8860/test/test_e2e_compatibility.py#L50
it seems a downloading pre-train model's problem.
Can you guys give me a clue?

qmpzzpmq:
I don't know what's going on with
https://github.com/espnet/espnet/blob/8db0c2996486a763452beccfb835b8ead84d8860/test/test_e2e_compatibility.py#L40
https://github.com/espnet/espnet/blob/8db0c2996486a763452beccfb835b8ead84d8860/test/test_e2e_compatibility.py#L50
https://github.com/espnet/espnet/blob/8db0c2996486a763452beccfb835b8ead84d8860/test/test_e2e_compatibility.py#L74
it seems a downloading protein model problem.
can you guys give me a clue?

sw005320:
This happens on [the other PR](https://travis-ci.org/espnet/espnet/builds/655741975?utm_source=github_status&utm_medium=notification) as well.
I could manually download the zip file, so it might be a tentative issue in the Google drive. 
Please ignore it for now.

sw005320:
Now it's working (this kind of thing happens several times...)

espnet/nets/pytorch_backend/lm/default.py
```diff
@@ -27,6 +27,8 @@ def add_arguments(parser):
                             help='Number of hidden layers')
         parser.add_argument('--unit', '-u', type=int, default=650,
                             help='Number of hidden units')
+        parser.add_argument('--embed-unit', type=int, default=128,
```
sw005320: Currently, the default uses `args.unit`, so could you also set it like that instead of `default=128`? (maybe we can just set with `default=None`?

qmpzzpmq: Actually, I think about it last night. I agree with change default with 650 keeping same with unit.

sw005320: I think it's better to make `default=None`.
Then, it just uses the same dimension as the number of hidden units according to https://github.com/espnet/espnet/blob/860cbd249f5d2ed48bd3f97cd82b984d7a7939a3/espnet/nets/pytorch_backend/lm/default.py#L242
Thus, this becomes the exact same behavior unless your new option of `--embed-unit` is specified.
If you agree, please change it, and also add such comments in the argument help.

qmpzzpmq: ok, I can do it, but which place you suggest my add some comment? in parser.add_argument 's help parameter?

sw005320: yes!

espnet/nets/pytorch_backend/lm/default.py
```diff
@@ -28,7 +28,8 @@ def add_arguments(parser):
         parser.add_argument('--unit', '-u', type=int, default=650,
                             help='Number of hidden units')
         parser.add_argument('--embed-unit', default=None,
-                            help='Number of hidden units in embedding layer')
+                            help='Number of hidden units in embedding layer, \
+                            if it is not setted, it keep same with hidden units.')
```
sw005320:
```suggestion
                            if it is not specified, it keeps the same number with hidden units.')
```

sw005320:
Thanks a lot!

PR #1625, opened 2020-02-27
[ESPnet2]  scripts/utils/perturb_data_dir_speed.sh

kamo-naoyuki:
Copied from Kaldi and deleted parts related to wav duration because we shouldn't use kaldi's command and we don't need the file actually.


sw005320:
> Copied from Kaldi and deleted parts related to wav duration because we shouldn't use kaldi's command and we don't need the file actually.

How about adding the above description to the script?

kamo-naoyuki:
@sw005320 You are right

kamo-naoyuki:
(This branch was merged)

PR #1641, opened 2020-03-04
VCC2020 baseline recipe

unilight:
Recipe for VCC2020 ASR+TTS baseline.

### Essential
- [x] Task 1
- [x] pack model in each TTS pretraining recipe
- [x] integrate with Parallel WaveGAN (PWG)
- [x] put pretrained TTS model
- [x] put finetuned TTS models for all target speakers
- [x] put pretrained PWG model
- [x] use source training set as the development set

### Refactor
- [x] Integrate the repeated scripts using symbolic links
- [x] move `train-embedding` to `tts_train.py`
- [x] add some checks whether the directory in other task exists or not, and add some appropriate error handling messages when it does not exist.
- [x] remove pack model part in `tts1_en_[de/fi/zh]`
- [x] move `conf/lists` to `local/lists`
- [x] add download scripts for downloadable datasets

### Documentation
- [x] add documentation about the VC function in the main `README.md`
- [x] add a short description in egs/README.md
- [x] add egs/vcc2020/README.md to explain the flow.
- [x] add description on how to manually download the CSS10 and VCC2020 dataset
- [x] add some documents of how TTS and ASR text normalizations are different.
- [x] add a comment that you decrease the beam size for fast decoding.
- [x] add an explanation of which model is used in `vc1_task1/local/recognize.sh`.
- [x] add the careful description of both downloaded and user's own models

### If have time
- [ ] unit testing `text2token.py` ?
- [ ] modularize to `local/decode.sh`
- [x] put converted samples in main README.md

### Renaming
- [x] `vcc2020` -> `vcc20`
- [x] `vc1/run_task1.sh`, `vc1/run_task2.sh` -> `vc1_task1/run.sh`, `vc1_task2/run.sh`


sw005320:
Please also put your action items (checkbox) in the first comment field (as I put them) so that we can monitor your progress easily.

kan-bayashi:
It is better to add `egs/vcc2020/README.md` which describes the basic explanation of the vc1 recipe and the whole flow.

unilight:
Just in case you did not notice, 
I said that I wouldn't have to modify the core of Espnet but I found that I had to.
In the finetuning stage I wanted to freeze the embedding table, 
so I added an argument in `e2e_tts_transformer` to achieve this function.
Please check.

egs/vcc2020/csmsc_mailabs/local/text/LICENSE
```diff
@@ -0,0 +1,19 @@
+Copyright (c) 2017 Keith Ito
```
kan-bayashi: It seems that while `vcc2020/tts1/local/text` is a symbolic link, the others are copied one.
To avoid too many duplicated scripts, could you change all of `local/text` to a symbolic link?

espnet/nets/pytorch_backend/e2e_tts_transformer.py
```diff
@@ -202,6 +202,8 @@ def add_arguments(parser):
                            help="Dropout rate in postnet")
         group.add_argument("--pretrained-model", default=None, type=str,
                            help="Pretrained model path")
+        group.add_argument("--train-embedding", default=True, type=strtobool,
```
kan-bayashi: How about moving this option to `tts_train.py`?
My suggestion is to make it similar to following options:
https://github.com/espnet/espnet/blob/043baca91d461022b1f3ae04f3152890aab9b830/espnet/bin/tts_train.py#L126-L130
For example, `--freeze-mods` which recieves comma separeted module names to be freezed (e.g. `freeze-mods: encoder.embed`).
And then you can set `require_grad = False` before optimizer setup, around here.
https://github.com/espnet/espnet/blob/043baca91d461022b1f3ae04f3152890aab9b830/espnet/tts/pytorch_backend/tts.py#L325-L328
Then, we can use freeze function for arbitrary models.

egs/vcc2020/tts1/run.sh
```diff
@@ -0,0 +1,395 @@
...
+
+if [ -z "$pretrained_model" ]; then
+    echo "Please specify pretrained model."
+    exit 1
+fi
+
+if [ ${stage} -le -1 ] && [ ${stop_stage} -ge -1 ]; then
+    echo "stage -1: Data Download"
+    # local/data_download.sh ${download_dir} ${spk}
+    # local/pretrained_model_download.sh ${download_dir} ${pretrained_model}
```
kan-bayashi: TODO: Fix this part

unilight:
Some discussions:

1. Some TTS datasets used here are also used in other existing recipes (e.g. CSMSC, M-AILABS). Is it okay to tell users to download them in the original recipes?

2. Can you provide space for me to upload the pretrained models? @sw005320 

3. In the finetuning stage, users can either use pretrained models we provide, or the models trained by the users. What would be a better way to show these options in, say, stage -1 in `vc1/run.sh` ? @kan-bayashi 

4. The VCC2020 training data would be on the Nagoya Univ. cloud space, w/ password.
So, instead of writing some script to download it automatically,
I plan to just write in the readme and tell the participant to put the data in the right place.
We can fix this part as the dataset is publicly released after the challenge.

5. Since the release of the training data does not come with the evaluation set,
and we don't really use the training data of the source English speakers,
I'm thinking about using them as development set.

sw005320:
> Some discussions:
> 
> 1. Some TTS datasets used here are also used in other existing recipes (e.g. CSMSC, M-AILABS). Is it okay to tell users to download them in the original recipes?

Yes.

> 2. Can you provide space for me to upload the pretrained models? @sw005320

I just sent an invitation.

> 3. In the finetuning stage, users can either use pretrained models we provide, or the models trained by the users. What would be a better way to show these options in, say, stage -1 in `vc1/run.sh` ? @kan-bayashi

You can refer to the speech translation recipe (e.g., https://github.com/espnet/espnet/blob/master/egs/fisher_callhome_spanish/st1/run.sh#L37-L38). @kan-bayashi would also suggest some ideas.

> 4. The VCC2020 training data would be on the Nagoya Univ. cloud space, w/ password.
>    So, instead of writing some script to download it automatically,
>    I plan to just write in the readme and tell the participant to put the data in the right place.
>    We can fix this part as the dataset is publicly released after the challenge.

This is no problem.
Many of the corpora are not downloadable, and we put the path of the directory and ask the user to change it with their own environments, e.g.,  the famous WSJ recipe is like that https://github.com/espnet/espnet/blob/master/egs/wsj/asr1/run.sh#L44-L45

> 5. Since the release of the training data does not come with the evaluation set,
>    and we don't really use the training data of the source English speakers,
>    I'm thinking about using them as development set.

OK.

utils/text2token.py
```diff
@@ -87,22 +87,22 @@ def main():
...
-                    i = end_pos
-                else:
-                    chars.append(a[i])
-                    i += 1
-            a = chars
-
         if(args.trans_type == "phn"):
             a = a.split(" ")
         else:
+            if len(match_pos) > 0:
```
sw005320: Can you add a unit test to validate this change in https://github.com/espnet/espnet/blob/master/test_utils/test_text2token_py.bats?

egs/vcc2020/vc1/local/recognize.sh
```diff
@@ -0,0 +1,120 @@
+#!/bin/bash
+
+# Copyright 2020 Nagoye University (Wen-Chin Huang)
+#  Apache 2.0  (http://www.apache.org/licenses/LICENSE-2.0)
+
+echo "$0 $*"  # Print the command line for logging
+. ./path.sh
+
```
sw005320: How about introducing `stage` like `run.sh`?
Then, we can also add `recog_stage` in `vc1/run.sh` and add that option when we call `local/recognize.sh`, e.g.,
```
local/recognize.sh --stage ${recog_stage}
```

utils/text2token.py
unilight: Can you tell me how this works? For example this change is to solve incorrect results when the text contains non-linguistic symbols and the trans_type is phn. So should I add such an example text in the unit test?

egs/vcc2020/vc1/local/recognize.sh
unilight: I can add this, but I wonder if this feature would be used, because the whole recognition stage is just one step in the complete VC flow, so users may hardly ever want to separately execute the recognition stages...?

utils/text2token.py
sw005320: Yes, I'd like to add the case for the phoneme.

egs/vcc2020/vc1/local/recognize.sh
sw005320: I see. I think we don't have to do it then.

unilight:
Some discussion about task 1:

1. Since task 1 and 2 share many codes in common, currently I'm splitting `vc1/run.sh` to `vc1/run_task1.sh` and `vc1/run_task2.sh`, instead of having two directories. This design is for reducing repeated codes, but if the latter way is clearer to you, we can change.

2. Since the LibriTTS recipe already exists, I think there's no need to create a recipe here, say, `vcc2020/tts1_en`. I will just tell the user to use the existing LibriTTS recipe if they want to tune the pretrained model or do experiments. Users can also just use the pretrained model provided in the LibriTTS recipe.
One question would be that we provide a model packing stage for finetuning in all `tts1_en_xx/run.sh`. Shall we also add it in `LibriTTS/tts1/run.sh`? Or we just tell the users to pack by themselves?

3. In the LibriTTS dataset, normalized text is provided, which is also used in the recipe. 
Sample: `100_121669_000001_000000 Tom, the Piper's Son`
However, the output of ASR is always uppercased.
Sample: `IN REALITY THE EUROPEAN PARLIAMENT IS PRACTISING DIALECTICS (SEF1_E30001)`
I plan to just simply lowercase the ASR results. This will cause a small mismatch but I think it should be acceptable.

unilight:
@kan-bayashi About the integration with Parallel WaveGAN, last time you mentioned that if we use pip to install, we can use something like `parallel-wavegan-train`. I am okay with this, but I'm not sure about the recipe design. Can you suggest some ideas?

sw005320:
> Some discussion about task 1:
> 
> 1. Since task 1 and 2 share many codes in common, currently I'm splitting `vc1/run.sh` to `vc1/run_task1.sh` and `vc1/run_task2.sh`, instead of having two directories. This design is for reducing repeated codes, but if the latter way is clearer to you, we can change.

I want to stick to have one `run.sh` in one directory.
Almost all recipes follow this convention.
So, my suggestion is `vc1_task1/run.sh` and `vc1_task2/run.sh`
You can use a symlink to specify to avoid redundant files in `local` (but not for `config` since the user may change this file depending on the task).
I think this would be good for people who participated in either task 1 or task2. 

> 2. Since the LibriTTS recipe already exists, I think there's no need to create a recipe here, say, `vcc2020/tts1_en`. I will just tell the user to use the existing LibriTTS recipe if they want to tune the pretrained model or do experiments. Users can also just use the pretrained model provided in the LibriTTS recipe.

OK

>    One question would be that we provide a model packing stage for finetuning in all `tts1_en_xx/run.sh`. Shall we also add it in `LibriTTS/tts1/run.sh`? Or we just tell the users to pack by themselves?

My suggestion is to specify the working directory containing models rather than using the packed models.
Packed models would be used for people who would skip training.
How about switching an option of specifying the working directory or using a packed model downloaded from our Google drive?

> 3. In the LibriTTS dataset, normalized text is provided, which is also used in the recipe.
>    Sample: `100_121669_000001_000000 Tom, the Piper's Son`
>    However, the output of ASR is always uppercased.
>    Sample: `IN REALITY THE EUROPEAN PARLIAMENT IS PRACTISING DIALECTICS (SEF1_E30001)`
>    I plan to just simply lowercase the ASR results. This will cause a small mismatch but I think it should be acceptable.

I think this is fine.
In other words, recovering the true case is not easy, and training ASR with the true case is not so successful.

kan-bayashi:
I will make the following options for parallel-wavegan-train:
- —train-wavscp
- —train-featscp
- —dev-wavscp
- —dev-featscp
Then you can just add a stage to train vocoder by specifying already dumped scpfiles.
I can finish within today (hopefully)

kan-bayashi:
I implemented the above features in https://github.com/kan-bayashi/ParallelWaveGAN/pull/96.
We can this feature from version 0.3.3. 
I will upload this version to pypi soon.

You can add a new stage like:
```bash
echo "stage X: neural vocoder training"
${cuda_cmd} --gpu "${ngpu}" "<voc_expdir>/train.log" \
	parallel-wavegan-train \
		--train-wav-scp data/train_nodev/wav.scp \
		--train-feats-scp dump/train_nodev/feats.scp \
		--dev-wav-scp data/train_dev/wav.scp \
		--dev-feats-scp dump/train_dev/feats.scp \
		--outdir <voc_expdir>/results \
		--config <voc_conf>
```

unilight:
@kan-bayashi Thank you very much!! 

I wonder how should we tell users to install extra packages. Should I specify it somewhere? (So far I've been doing this by executing `. ./path.sh` first and install by pip.)

kan-bayashi:
> Does it mean that I still cannot use this feature if I install parallel-wavegan through pip in Espnet?

I uploaded to pypi. Now you can install via `pip install parallel_wavegan==0.3.3`

> Is decoding also supported?

Of course. You can use like:
```
parallel-wavegan-decode \
	--checkpoint <voc_expdir>/checkpoint-XXsteps.pkl \
	--feats-scp dump/test_yesno/feats.scp \
	--outdir <path>/<to>/<directory>
```

> I wonder how should we tell users to install extra packages. Should I specify it somewhere? (So far I've been doing this by executing . ./path.sh first and install by pip.)

You can add installation check in `path.sh` e.g.:
```
if ! command -v parallel-wavegan-train > /dev/null; then
    echo "Error: parallel_wavegan is not installed." >&2
    echo "Error: Please install via \`. ./path.sh && pip install -U parallel_wavegan\`" >&2
    return 1
fi
```
And you can write about the extra requirements in `egs/vcc20/README.md`.

utils/text2token.py
unilight: I tried to follow https://github.com/espnet/espnet/blob/master/CONTRIBUTING.md#unit-testing but I can not figure out the right way to do it... Can you give a little more details on how to do this? Thank you!

unilight:
@sw005320 @kan-bayashi 
I am thinking about the recipe for integrating parallel WaveGAN (PWG) training. Although Hayashi-san provided a convenient way to train a PWG in `vc1_taskX/run.sh`, I think it would be clearer if we have a separate directory, say `pwg1`, and provide training recipe in it. We may also provide PWG pretrained models. Then, in `vc1_taskX/run.sh`, at the decoding stage where we convert fbanks back to waveform, we let users choose whether they want to use Griffin-Lim, our provided pretrained PWG or the PWG they trained in `pwg1`.
Does it sound okay to you?

egs/vcc2020/voc1/local/preprocess.py
```diff
@@ -0,0 +1,184 @@
...
+# -*- coding: utf-8 -*-
+
+# Copyright 2020 Nagoya University (Wen-Chin Huang)
+#  Apache 2.0  (http://www.apache.org/licenses/LICENSE-2.0)
+
+"""Perform preprocessing and raw feature extraction.
+   This is slightly different from the original version,
+   because we want to train 16kHz-24kHz PWG, 
+   so we skip sampling rate check.
+"""
```
kan-bayashi: If the difference only the check, I can make `--allow-different-sampling-rate` option in PWG.

egs/vcc2020/vc1/run_task1.sh
```diff
@@ -372,32 +382,75 @@ if [ ${stage} -le 12 ] && [ ${stop_stage} -ge 12 ]; then
...
+        echo "Normalizing..."
+        ${train_cmd} --num-threads "${nj}" "${hdf5_norm_dir}/normalize.log" \
+            local/normalize_for_pwg.py \
+                --config "${voc_conf}" \
+                --stats "${voc_stats}" \
+                --rootdir ${hdf5_dir} \
+                --dumpdir ${hdf5_norm_dir} \
+                --n_jobs "${nj}" \
+                --verbose "${verbose}"
+        echo "successfully finished normalization."
```
kan-bayashi: I made new option for this part.
You can use `--skip-wav-copy` option as follows
```bash
parallel-wavengan-normalize \
	--skip-wav-copy \
	--config ${voc_conf} \
	--feats-scp ${outdir}_denorm/${pairname}/feats.scp \
	--dumpdir ${hdf5_norm_dir}
```

egs/vcc2020/vc1/run_task2.sh
```diff
@@ -376,32 +385,75 @@ if [ ${stage} -le 12 ] && [ ${stop_stage} -ge 12 ]; then
...
+        echo "Normalizing..."
+        ${train_cmd} --num-threads "${nj}" "${hdf5_norm_dir}/normalize.log" \
+            local/normalize_for_pwg.py \
+                --config "${voc_conf}" \
+                --stats "${voc_stats}" \
+                --rootdir ${hdf5_dir} \
+                --dumpdir ${hdf5_norm_dir} \
+                --n_jobs "${nj}" \
+                --verbose "${verbose}"
+        echo "successfully finished normalization."
```
kan-bayashi: ditto

egs/vcc2020/vc1/run_task1.sh
kan-bayashi: Note that if the stats are the same in both tts and voc, we can directly use generated feats.scp.

egs/vcc2020/voc1/conf/parallel_wavegan.task2.yaml
```diff
@@ -0,0 +1,120 @@
+# This is the hyperparameter configuration file for Parallel WaveGAN.
+# This configuration requires 12 GB GPU memory and takes ~3 days on TITAN V.
+
+###########################################################
+#                FEATURE EXTRACTION SETTING               #
+###########################################################
+sampling_rate: 24000     # Sampling rate.
+fft_size: 1024           # FFT size.
+hop_size: 384            # Hop size.
```
kan-bayashi: Is this parameter intended?

egs/vcc2020/voc1/run.sh
```diff
@@ -0,0 +1,199 @@
...
+    [ ! -e ${hdf5_norm_dir} ] && mkdir -p ${hdf5_norm_dir}
+    ${train_cmd} --num-threads "${n_jobs}" "${hdf5_norm_dir}/normalize.log" \
+        local/custom_normalize.py \
+            --config "${conf}" \
+            --stats "${dumpdir}/${train_set}/stats.h5" \
+            --rootdir ${hdf5_dir} \
+            --dumpdir ${hdf5_norm_dir} \
+            --n_jobs "${n_jobs}" \
+            --verbose "${verbose}"
+    echo "successfully finished normalization."
```
kan-bayashi: This part can also be replaced with the above `parallel-wavegan-normalize` with `skip-wav-copy` option.

kan-bayashi:
@unilight Hi WenChin, thank you for your great work!
Since my pwg code lacks flexibility, you made slightly customed code.
To avoid this, I made several new options.
https://github.com/kan-bayashi/ParallelWaveGAN/pull/98
I can implement more options, so if you have a request, please let me know.

egs/vcc2020/voc1/conf/parallel_wavegan.task2.yaml
unilight: In all TTS model feature extraction, the `hop_size` is 256 and the waveform is 16kHz,
so I set `hop_size` to 384 here so that we can upsample the 16kHz fbanks to 24kHz waveform.

But I think there should be some problem here.
In PWG feature extraction, we first downsample 24kHz to 16kHz, so here hop size should be still 256.

I wrote some custom codes in my own PWG experiment directory to overcome these problems,
but, anyway, can you implement some option so that we can make <24kHz waveform, 16kHz mel fbank>?
Maybe `parallel-wavegan-preprocess` can take two kinds of wav scps, one is for ground truth (24kHz), and another one is the downsampled one for feature extraction (16kHz). In the config file, we can also split the `hop_size` into two, one for feature extraction and one for checking the upsample scale of the generator.
Sorry for asking you to implement such complicated feature...

kan-bayashi: OK. Let me confirm some points
- The target waveform is 24kHz
- In feature extraction, downsample from 24kHz -> 16kHz
- Instead of the hopsize in 24 kHz, hopsize * 16 kHz / 24 kHz will be used
Is it correct? If so, I will support sampling_rate_in_mel_ext option in config. (the hopsize will be calculated automatically and wav is downsampled in the scripts.)

egs/vcc2020/voc1/local/preprocess.py
kan-bayashi: Again, could you explain the difference from the original script?

egs/vcc2020/voc1/conf/parallel_wavegan.task2.yaml
kan-bayashi: @unilight I implemented such option.
https://github.com/kan-bayashi/ParallelWaveGAN/pull/99
Could you check it? If OK, I will update version and upload to pypi.

egs/vcc2020/vc1/local/feats2hdf5.py
```diff
@@ -0,0 +1,68 @@
+#!/usr/bin/env python3
```
kan-bayashi: Maybe no more needed?

egs/vcc2020/voc1/conf/parallel_wavegan.task2.yaml
unilight: @kan-bayashi Thank you very much! It works fine. Please upload to pypi.

kan-bayashi: Uploaded. 0.3.4 is available.

egs/vcc2020/vc1_task2/run.sh
```diff
@@ -0,0 +1,452 @@
...
+griffin_lim_iters=64            # the number of iterations of Griffin-Lim
+
+# pretrained model related
+pretrained_model_dir=downloads  # If use provided pretrained models, set to desired dir, ex. `downloads`
+                                # If use manually trained models, set to `..`
+pretrained_model_name=          # Recommended choices: tts1_en_[de,fi,zh]
+
+# dataset configuration
+db_root=downloads/official_v1.0_training
+eval_db_root=../vc/downloads    # Same as `db_root` in training
```
sw005320: ditto.
Also, the `vc` directory may not exist. Please double-check it.

egs/vcc2020/README.md
```diff
@@ -0,0 +1,62 @@
...
+
+The following packages do not come with the installation of ESPnet. Please install them via `pip`.
+
+```
+cd <recipe>
+. ./path.sh
+pip install -U <package>
+```
+
+- [pypinyin](https://pypi.org/project/pypinyin/): used in `tts1_en_zh` and `vc1_task2`.
```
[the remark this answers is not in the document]
sw005320: I found pwg one in https://github.com/unilight/espnet/blob/vcc2020-baseline/egs/vcc2020/voc1/path.sh#L19-L23
Please add a similar check for pypinyin where it is used.

egs/vcc2020/tts1_en_de/run.sh
```diff
@@ -0,0 +1,325 @@
...
+eval_set="eval"
+
+if [ ${stage} -le -1 ] && [ ${stop_stage} -ge -1 ]; then
+    echo "stage -1: Data Download"
+    echo "Please download the M-AILABS datasets in the recipe."
+fi
+
+if [ ${stage} -le 0 ] && [ ${stop_stage} -ge 0 ]; then
+    echo "stage 0: Data preparation"
+
```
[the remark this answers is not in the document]
sw005320: The above assumes that the data download is in stage -1, and also it only uses data download. If the script requires to do more, please change the above part appropriately.

unilight:
@sw005320 Thank you for your review and valuable comments !!!!

> * I cannot fully understand why you pack the model, but it might require more time and we can skip this point for now (correction, your script seems to accept both?)

I'm sorry that I did not notice that part. I decided not to make packing the model as an optional stage in `tts1_en_xx/run.sh`.
If users want to pretrain their own models, I just let them specify the directory. One small problem is that if you look at this part (https://github.com/unilight/espnet/blob/vcc2020-baseline/egs/vcc2020/vc1_task1/run.sh#L209-L220,) only the first experiment will be used. If users do several pretraining exps, they need to hack this part by themselves.

> Please add some check to confirm the existence of `${db_root}`
> 
> ```
> if [ -e ${db_root} ]; then
> 	...
> else
> 	echo "no ${db_root} found"
> 	echo "cd ${db_root}; ./run.sh --stop_stage -1; cd -"
> 	exit 1;
> fi
> ```

I am not really sure where to put this. Currently I put it in stage 0, right before data preparation.

Also, can I keep the name to `vcc2020`?

unilight:
I have finished all the documentation. Please check if you have time.

sw005320:
> @sw005320 Thank you for your review and valuable comments !!!!
> 
> > * I cannot fully understand why you pack the model, but it might require more time and we can skip this point for now (correction, your script seems to accept both?)
> 
> I'm sorry that I did not notice that part. I decided not to make packing the model as an optional stage in `tts1_en_xx/run.sh`.
> If users want to pretrain their own models, I just let them specify the directory. One small problem is that if you look at this part (https://github.com/unilight/espnet/blob/vcc2020-baseline/egs/vcc2020/vc1_task1/run.sh#L209-L220,) only the first experiment will be used. If users do several pretraining exps, they need to hack this part by themselves.

I see.
This is very good.
I think you can also add the careful description of both downloaded and user's own models in your README.md (sorry if you already added. I did not check the README part yet).
This part is a bit complicated but very beneficial for the users.

> > Please add some check to confirm the existence of `${db_root}`
> > ```
> > if [ -e ${db_root} ]; then
> > 	...
> > else
> > 	echo "no ${db_root} found"
> > 	echo "cd ${db_root}; ./run.sh --stop_stage -1; cd -"
> > 	exit 1;
> > fi
> > ```
> 
> I am not really sure where to put this. Currently I put it in stage 0, right before data preparation.

Very good!

> Also, can I keep the name to `vcc2020`?

I personally want to change it to `vcc20` since the all other recipes (iwslt and blizzard) follow this naming convention.

egs/README.md
```diff
@@ -51,6 +51,7 @@ See: https://espnet.github.io/espnet/tutorial.html
 | timit                   | TIMIT Acoustic-Phonetic Continuous Speech Corpus             | ASR                                        | EN             | https://catalog.ldc.upenn.edu/LDC93S1                        |                               |
 | tweb                    | The World English Bible                                      | TTS                                        | EN             | https://www.kaggle.com/bryanpark/the-world-english-bible-speech-dataset                      |                               |
 | vais1000                | VAIS-1000                                                    | TTS                                        | VI             | https://ieee-dataport.org/documents/vais-1000-vietnamese-speech-synthesis-corpus  |             |
+| vcc2020                 | Voice Conversion Challenge 2020                              | VC                                         | EN->{DE, FI, ZH} | http://www.vc-challenge.org/                               |                               |
```
sw005320: This description is only for task2?

README.md
```diff
@@ -53,6 +54,10 @@ and also follows [Kaldi](http://kaldi-asr.org/) style data processing, feature e
 - Transformer based end-to-end ST (new!)
 - Transformer based end-to-end MT (new!)
 
+### VC: Voice conversion
+- End-to-end VC based on cascaded ASR+TTS (new!)
+- Baseline system for Voice Conversion Challenge 2020!
+
```
sw005320: Like the TTS function, you can add VC results (audio example link) and pre-trained models in this main README.md (advertise your work!).
Probably it would be better to have an additional cool audio example webpage, but you can do it later.

egs/vcc2020/README.md
```diff
@@ -0,0 +1,63 @@
...
+
+This recipe describes baseline for the Voice Conversion (VC) Challenge 2020 (VCC2020 for short). VCC2020 contains two tasks. In either tasks, the source speech is always an English speech from a native speaker. Task 1 requires to convert to a English target speaker with a small amount of English parallel training set, and task 2 requires to convert to a non-English (German/Finnish/Mandarin) target speaker with a non-English training set. Task 2 is also referred as cross-lingual VC.
+
+## Method
+
+A naive approach for VC is a cascade of an automatic speech recognition (ASR) model and a text-to-speech (TTS) model. In this recipe we revisit this method under an end-to-end (E2E) framework.
+
+- Specifically, we train a (speaker independent) ASR model, and a separate speaker dependent TTS model for each target speaker. First, the ASR model takes the source English speech as input and output the transcribed text. Then, the TTS model of the target speaker synthesizes the converted speech using the recognition result.
+- Since the size of training set of each target speaker is to limited for E2E TTS learning, we use a pretraining-finetuning scheme to improve the performance.
```
sw005320:
```suggestion
- Since the size of the training set of each target speaker is too limited for E2E TTS learning, we use a pretraining-finetuning scheme to improve the performance.
```

egs/vcc2020/README.md
```diff
@@ -0,0 +1,63 @@
+# Recipe for Voice Conversion Challenge 2020 baseline : Cascade ASR + TTS
+
+Official homepage: [http://www.vc-challenge.org/](http://www.vc-challenge.org/)
+
+## Introduction
+
+This recipe describes baseline for the Voice Conversion (VC) Challenge 2020 (VCC2020 for short). VCC2020 contains two tasks. In either tasks, the source speech is always an English speech from a native speaker. Task 1 requires to convert to a English target speaker with a small amount of English parallel training set, and task 2 requires to convert to a non-English (German/Finnish/Mandarin) target speaker with a non-English training set. Task 2 is also referred as cross-lingual VC.
```
sw005320:
```suggestion
This recipe describes a baseline for the Voice Conversion (VC) Challenge 2020 (VCC2020 for short). VCC2020 contains two tasks. In either task, the source speech is always an English speech from a native speaker. Task 1 requires to convert to an English target speaker with a small amount of English parallel training set, and task 2 requires to convert to a non-English (German/Finnish/Mandarin) target speaker with a non-English training set. Task 2 is also referred to as a cross-lingual VC.
```

egs/vcc2020/tts1_en_de/run.sh
```diff
@@ -0,0 +1,310 @@
...
+set -u
+set -o pipefail
+
+train_set="train"
+dev_set="dev"
+eval_set="eval"
+
+if [ ${stage} -le -1 ] && [ ${stop_stage} -ge -1 ]; then
+    echo "stage -1: Data Download"
+    echo "Please download the M-AILABS datasets in the recipe."
```
sw005320: Sorry for the confusion.
I think it's better to add actual download scripts here by assuming that most people have never touched `m_ailabs`'s recipe.
```
local/download.sh ${db_root} de_DE
local/download.sh ${db_root} en_US
```
Then, please add a symlink of `local/download.sh` from `m_ailabs/tts1/local/download.sh`

Please also do the same things for the other `run.sh`

egs/vcc2020/tts1_en_de/run.sh
```diff
@@ -0,0 +1,310 @@
...
+fi
+
+if [ ${stage} -le 0 ] && [ ${stop_stage} -ge 0 ]; then
+    echo "stage 0: Data preparation"
+
+    if [ ! -e ${db_root} ]; then
+        echo "${db_root} not found."
+        echo "cd ${db_root}; ./run.sh --stop_stage -1; cd -"
```
sw005320: If we perform the above modification, you can just do like that
```suggestion
        echo "./run.sh --stage -1"
```
Or you can delete this check.

egs/vcc2020/tts1_en_fi/run.sh
```diff
@@ -0,0 +1,314 @@
...
+        echo "${fin_db} not found."
+        echo "cd ${fin_db}; ./run.sh --stop_stage -1; cd -"
+        exit 1;
+    fi
+    
+    if [ ! -e ${mailabs_db} ]; then
+        echo "${mailabs_db} not found."
+        echo "cd ${mailabs_db}; ./run.sh --stop_stage -1; cd -"
+        exit 1;
+    fi
```
sw005320: Delete these lines if you put the appropriate download scripts in the above.

egs/vcc2020/tts1_en_fi/run.sh
```diff
@@ -0,0 +1,314 @@
...
+set -u
+set -o pipefail
+
+train_set="train"
+dev_set="dev"
+eval_set="eval"
+
+if [ ${stage} -le -1 ] && [ ${stop_stage} -ge -1 ]; then
+    echo "stage -1: Data Download"
+    echo "Please download the CSS10 and M-AILABS datasets in respective recipes"
```
sw005320: Same comment to the above.
We can put the CSS10 and M-AILABS download scripts here.

egs/vcc2020/tts1_en_de/run.sh
```diff
@@ -0,0 +1,310 @@
...
+# config files
+train_config=conf/train_pytorch_transformer+spkemb.yaml
+decode_config=conf/decode.yaml
+
+# decoding related
+model=model.loss.best
+griffin_lim_iters=64  # the number of iterations of Griffin-Lim
+
+# specify the downloaded database directories
+db_root=../m_ailabs/tts1/downloads
```
sw005320: Maybe?
```suggestion
db_root=../../m_ailabs/tts1/downloads
```

egs/vcc2020/tts1_en_zh/run.sh
```diff
@@ -0,0 +1,313 @@
...
+set -u
+set -o pipefail
+
+train_set="train"
+dev_set="dev"
+eval_set="eval"
+
+if [ ${stage} -le -1 ] && [ ${stop_stage} -ge -1 ]; then
+    echo "stage -1: Data Download"
+    echo "Please download the CSMSC and M-AILABS datasets in respective recipes."
```
sw005320: ditto

egs/vcc2020/tts1_en_zh/run.sh
```diff
@@ -0,0 +1,313 @@
...
+        echo "${csmsc_db} not found."
+        echo "cd ${csmsc_db}; ./run.sh --stop_stage -1; cd -"
+        exit 1;
+    fi
+    
+    if [ ! -e ${mailabs_db} ]; then
+        echo "${mailabs_db} not found."
+        echo "cd ${mailabs_db}; ./run.sh --stop_stage -1; cd -"
+        exit 1;
+    fi
```
sw005320: ditto

unilight:
>I think you can also add the careful description of both downloaded and user's own models in your README.md (sorry if you already added. I did not check the README part yet).
>This part is a bit complicated but very beneficial for the users.

I'll add this part tomorrow.

>I personally want to change it to vcc20 since the all other recipes (iwslt and blizzard) follow this naming convention.

I see. I'll do this right before the release.

I have done the spelling check of the README files w/ Grammarly. Please check them!

egs/vcc2020/vc1_task1/README.md
```diff
@@ -0,0 +1,78 @@
...
+- Stage 3: X-vector extraction. This is based on the pre-trained, Kaldi-based x-vector extraction.
+- Stage 4: Model training.
+- Stage 5: Decoding. We decode the development set.
+
+## Conversion phase
+
+Execute the main script to convert `SEF1` to `TEF1`:
+
+```
+$ ./run_task2.sh --stage 11 \
```
sw005320: I guess `run_task2.sh` should be changed.

egs/vcc2020/vc1_task1/README.md
```diff
@@ -0,0 +1,78 @@
...
+
+## Notes
+
+### Text normalization mismatch between ASR and TTS
+
+LibriTTS itself provides normalized text, and the pretrained model is directly trained on such text. Ex:
+
+`100_121669_000001_000000 Tom, the Piper's Son`
+
+However, the output of the pretrained ASR is always uppercased:
```
sw005320: You may also note the punctuation mismatch issue.

kan-bayashi:
Hi @unilight, this is really great work!
I will check the scripts again from now.

@tomokitoda Could you also add your comments?

sw005320:
> I will check the scripts again from now.

Many thanks.
I'd not have enough bandwidth to actually execute the recipe. It would be great if you try it.

egs/vcc2020/README.md
```diff
@@ -0,0 +1,63 @@
...
+- **VCC2020** contains the main training data of the challenge. Please follow the instruction from the organizers and put in the desired directory. (default is `vc1_task[1/2]/downloads/`)
+
+## Usage
+
+Please see the readme in each recipe.
+
+## Author
+
+Main development: Wen-Chin Huang @ Nagoya University (2020/03)  
+Advisors: Tomoki Hayashi (Nagoya U), Shinji Watanabe (JHU), Tomoki Toda (Nagoya U)
```
sw005320: Could we also add an acknowledgment to Matthew Wiesner for his help with phoneme transcription preparations?
E.g.,
```
Special thanks to Matthew Wiesner (JHU)
```
You could also add other people as well.

egs/vcc2020/vc1_task1/run.sh
```diff
@@ -0,0 +1,459 @@
...
+
+# dataset configuration
+db_root=downloads/official_v1.0_training
+eval_db_root=downloads/official_v1.0_training    # Same as `db_root` in training
+list_dir=local/lists
+spk=TEF1 
+
+# vc configuration
+srcspk=
+trgspk=
```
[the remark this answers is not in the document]
kan-bayashi: And how about adding the check whether to correctly set these variables which needed to set manually in the beginning of the conversion stage?

egs/vcc2020/vc1_task2/local/clean_text_finnish.py
```diff
@@ -0,0 +1,105 @@
...
+    else:
+        ssl._create_default_https_context = _create_unverified_https_context
+    nltk.download("punkt")
+
+
+def g2p(text):
+    """Convert grapheme to phoneme."""
+    tokens = filter(lambda s: s != " ", f_g2p(text))
+    return ' '.join(tokens)
+
```
kan-bayashi:
```suggestion


```

egs/vcc2020/vc1_task2/local/clean_text_finnish.py
```diff
@@ -0,0 +1,105 @@
...
+def custom_finnish_cleaners(text):
+    text = lowercase(text)
+    text = expand_numbers(text)
+    text = expand_abbreviations(text)
+    text = expand_symbols(text)
+    text = remove_unnecessary_symbols(text)
+    text = uppercase(text)
+    text = collapse_whitespace(text)
+    return text
+
```
kan-bayashi:
```suggestion


```

egs/vcc2020/vc1_task2/local/clean_text_german.py
```diff
@@ -0,0 +1,92 @@
...
+    else:
+        ssl._create_default_https_context = _create_unverified_https_context
+    nltk.download("punkt")
+
+
+def g2p(text):
+    """Convert grapheme to phoneme."""
+    tokens = filter(lambda s: s != " ", f_g2p(text))
+    return ' '.join(tokens)
+
```
kan-bayashi:
```suggestion


```

egs/vcc2020/vc1_task2/local/clean_text_mandarin.py
```diff
@@ -0,0 +1,114 @@
...
+    else:
+        ssl._create_default_https_context = _create_unverified_https_context
+    nltk.download("punkt")
+
+
+def g2p(text):
+    """Convert grapheme to phoneme."""
+    tokens = filter(lambda s: s != " ", f_g2p(text))
+    return ' '.join(tokens)
+
```
kan-bayashi:
```suggestion


```

egs/vcc2020/voc1/run.sh
```diff
@@ -0,0 +1,157 @@
...
+. utils/parse_options.sh || exit 1;
+
+train_set="train_${task}"
+dev_set="dev_${task}"
+
+set -euo pipefail
+
+if [ "${stage}" -le -1 ] && [ "${stop_stage}" -ge -1 ]; then
+    echo "Stage -1: Data download"
+    echo "Please download the dataset following the README."
```
kan-bayashi: Add existence check?

kan-bayashi:
I think you should add comments about providing pretrained models. Actually we can skip `tts_*` and `voc1` recipes to run `vc_task*` since the pretrained model will be downloaded in the `vc_task*` recipe. Is it correct? This is nice for the users who want to run only fine-tuning part as example.

egs/vcc2020/tts1_en_fi/README.md
```diff
@@ -0,0 +1,31 @@
...
+- Stage 0: Data preparation. This includes transcription parsing.
+- Stage 1: Feature extraction. The features are normalized using the stats calculated using the pooled dataset.
+- Stage 2: JSON format data preparation.
+- Stage 3: X-vector extraction. This is based on the pre-trained, Kaldi-based x-vector extraction.
+- Stage 4: Model training.
+- Stage 5: Decoding and synthesis. We decode the development set and synthesize the waveform using the Griffin-Lim algorithm for a quick performance check.
+
+## Notes
+
+- All texts are capitalized.
```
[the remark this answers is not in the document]
unilight: @kan-bayashi  I think the `custom_english_cleaners` function preserves the punctuation, right?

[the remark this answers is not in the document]
kan-bayashi: Right. E.g. comma, period, and apostrophe.
```
LJ049-0076 THE COMMISSION'S REVIEW OF THE PROVISIONS FOR PRESIDENTIAL PROTECTION AT THE TIME OF PRESIDENT KENNEDY'S TRIP TO DALLAS DEMONSTRATES THE NEED FOR SUBSTANTIAL IMPROVEMENTS.
```

egs/vcc2020/tts1_en_zh/local/data_prep_csmsc.sh
```diff
@@ -0,0 +1,52 @@
...
+
+# make text
+find ${db}/PhoneLabeling -name "*.interval" -follow | sort | while read -r filename;do
+    id="$(basename ${filename} .interval)"
+    content=$(tail -n +13 ${filename} | grep "\"" | grep -v "sil" | sed -e "s/\"//g" | tr "\n" " " | sed -e "s/ $//g")
+    start_sec=$(tail -n +14 ${filename} | head -n 1)
+    end_sec=$(head -n -2 ${filename} | tail -n 1)
+    echo "${spk}_${id} <${lang}> ${content}" >> ${text}
+done
+echo "Successfully finished making text. One of the segment info is wrong so let's not to use it."
```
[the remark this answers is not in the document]
unilight: For one of the utterances, the segment label does not match the waveform. (I don't quite remember but the segment label says the waveform is 9 seconds long and has a 4 seconds long silence in the beginning, but the utterance is only like 3 seconds long? Something like that.) So the segmentation in the preprocessing stage fails.

egs/vcc2020/vc1_task1/run.sh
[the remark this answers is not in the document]
unilight: > And how about adding the check whether to correctly set these variables which needed to set manually in the beginning of the conversion stage?

I can add the comments. As for argument checking, I found that in some ESPnet recipes the check does not always exist, and since we set bash to debug mode, any undefined argument (variable?) will cause the program to instantly exit. So...

unilight:
> I think we should add comments about providing pretrained models. Actually we can skip `tts_*` and `voc1` recipes to run `vc_task*` since the pretrained model will be downloaded in the `vc_task*` recipe. Is it correct? This is nice for the users who want to run only fine-tuning part as an example, so why don't you add comments about it?

Yes, it is correct. I will add comments to the main README. Regarding pretrained models, I have some additional concerns:

1. I am considering adding pretrained finetuned models for each target speaker. Do you think it is necessary?

2. I actually did not mention about the training of the ASR model. My thought is that the current pretrained ASR model is already good enough (we already have CER/WER ~ 2-3% for the test set). Even if we further, say, adapt the model to each source speaker, I don't think there would be a big improvement. But I am afraid some users would want to do it. What do you think?

egs/vcc2020/tts1_en_fi/README.md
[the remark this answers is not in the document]
unilight: I see. I will add comments about this.

egs/vcc2020/vc1_task1/run.sh
```diff
@@ -0,0 +1,459 @@
...
+voc=                            # GL or PWG
+voc_expdir=downloads/pwg_task1  # If use provided pretrained models, set to desired dir, ex. `downloads/pwg_task1`
+                                # If use manually trained models, set to `../voc1/exp/<expdir>`
+voc_checkpoint=                 # If not specified, automatically set to the latest checkpoint 
+griffin_lim_iters=64            # the number of iterations of Griffin-Lim
+
+# pretrained model related
+pretrained_model_dir=downloads  # If use provided pretrained models, set to desired dir, ex. `downloads`
+                                # If use manually trained models, set to `../libritts`
+pretrained_model_name=          # Recommended choices: tts1
```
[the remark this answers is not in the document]
unilight: Hmm... If the user chooses to use the manually pretrained model, then it is the recipe directory name. But if the user chooses to use the pretrained model, then it is just the name of the directory containing the downloaded model. So I just call it `pretrain_model_name`. Don't know if it sounds reasonable to you...

egs/vcc2020/voc1/run.sh
unilight: Existence checks are in stage 0.

sw005320:
After PWG pre-training is finished, we will merge your PR to the master (and continue some existing TODO and get feedbacks from @tomokitoda). 

> > I think we should add comments about providing pretrained models. Actually we can skip `tts_*` and `voc1` recipes to run `vc_task*` since the pretrained model will be downloaded in the `vc_task*` recipe. Is it correct? This is nice for the users who want to run only fine-tuning part as an example, so why don't you add comments about it?
> 
> Yes, it is correct. I will add comments to the main README. Regarding pretrained models, I have some additional concerns:
> 
> 1. I am considering adding pretrained finetuned models for each target speaker. Do you think it is necessary?

How many speakers?
I highly recommend you to do it.

> 2. I actually did not mention about the training of the ASR model. My thought is that the current pretrained ASR model is already good enough (we already have CER/WER ~ 2-3% for the test set). Even if we further, say, adapt the model to each source speaker, I don't think there would be a big improvement. But I am afraid some users would want to do it. What do you think?

You can just mention the above things in `egs/vcc2020/README.md` as follows. We may not have to have a recipe level support of it.

ASR training. A [Transformer-based ASR model](https://github.com/espnet/espnet/tree/master/egs/librispeech/asr1) is used. ESPnet provides a state-of-the-art [LibriSpeech pretrained model](https://github.com/espnet/espnet/blob/master/egs/librispeech/asr1/RESULTS.md#pytorch-large-transformer-with-specaug-4-gpus--large-lstm-lm), which is already good enough (CER/WER ~ 2-3% for the test set) and we did not include ASR training in our recipe. If you want to further refine the ASR pretrained model, please revisit our Librispeech recipe or other ASR recipes in the ESPnet.

unilight:
I have put the pretrained PWGs on Google Drive.

> How many speakers?
> I highly recommend you to do it.

There are 4 and 6 target speakers in tasks 1 and 2, respectively.
I think it is a good chance to check the scripts so I am re-finetuning their respective TTS models right now. It takes ~6 hrs to fine-tune one and depending on the GPU usage I expect to finish all the finetuning by the end of tomorrow. We can merge this PR by then.

@kan-bayashi I want to pack the average x-vector of each target speaker along with the TTS models. Is it possible to modify `pack_model.sh` to pack extra files? Or we should let the user calculate at local?

sw005320:
> I have put the pretrained PWGs on Google Drive.

Cool!

> > How many speakers?
> > I highly recommend you to do it.
> 
> There are 4 and 6 target speakers in tasks 1 and 2, respectively.
> I think it is a good chance to check the scripts so I am re-finetuning their respective TTS models right now. It takes ~6 hrs to fine-tune one and depending on the GPU usage I expect to finish all the finetuning by the end of tomorrow. We can merge this PR by then.

OK.
Or we can merge it now (if @kan-bayashi is OK).
We can make a PR later.

> @kan-bayashi I want to pack the average x-vector of each target speaker along with the TTS models. Is it possible to modify `pack_model.sh` to pack extra files? Or we should let the user calculate at local?

Yes.
Also, you may use `--etc`

egs/vcc2020/tts1_en_zh/local/data_prep_csmsc.sh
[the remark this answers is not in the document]
kan-bayashi: OK. Thanks for your information.

egs/vcc2020/vc1_task1/run.sh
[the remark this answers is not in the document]
kan-bayashi: OK. I understand.

kan-bayashi:
LGTM for me except for minor comments.
We can add the fine-tuned model later in another PR.
If you reflect @tomokitoda's comments,

unilight:
Just a brief report on the progress.
I have modified the documents based on comments from Toda-sensei.
I have also finished pretraining (pre-finetuning) all the models.
But the required modification to make the script accept downloaded pre-finetuned models is more than I expected.
Since such modification might greatly change the main scripts (`vc1_task[1/2]/run.sh`), please give me another day to finish it.
Then we can merge this PR.

sw005320:
OK for me.

unilight:
@sw005320 @kan-bayashi 

Sorry for the wait.
I have put the finetuned models up and made final modifications to the documents.
Toda-sensei has also checked the converted samples.
I think this baseline is ready to go!

sw005320:
Many thanks, @unilight!

PR #1648, opened 2020-03-05
[WIP] DNN HMM Hybrid ASR

kamo-naoyuki:


b-flo:
Hi,

Really interesting! When you're back could you comment a bit on this PR please? As I mentioned on slack, one of the problem for path-aware regularization in espnet was that it heavily relies on kaldi and fst stuff. If HMM-based models are planned in espnet2 I think I can figure out an adequate solution!

Also feel free if you need any help in this PR! I would be glad to contribute!

kamo-naoyuki:
- On a whim. Not authorized by @sw005320 
   - I often need pytorch based DNN-HMM system simply to compare espnet or combine some techniques for DNN-HMM to end-to-end. It's useful for me if integrated in espnet itself.
- Just like pytorch-kaldi, rely on Kaldi for GMM-HMM part, forced alignment, and WFST decoding.
   - @b-flo may need some python native? This PR doesn't aim such stuff.
- I implemented simple classifying task of HMM state for each frames.
- For decoding, I also implemented a tool to dump posterior from DNN model as ark file, then we can pipe it to kaldi's decoder in shell level.
- I don't intend to implement more.
- (I will be stopped for 2-3 weeks)

Issue #1643, opened 2020-03-05
Cannot reproduce WSJ Transformer RESULTS


wolverineq:
Hi I have noticed the latest WSJ RESULTS reporting WER 2.x% in eval92 using ctc-transformer models.
However I cannot reproduce the results with the same configurations.

My results:
PyTorch:1.1.0 espnet: 0.6.1 cuda: 9.0
data:using sample filtering (however with no utts excluded for train set )
model: ctc(PyTorch builtin)-transformer
WER: 5.2% CER:2.5%

Even I cannot decode similar results using the uploaded model files
I get WER 5.6% with the provided model files.

Maybe I have missing some important settings. Could you help me?

sw005320:
@Emrys365, could you follow up on this discussion?

Emrys365:
I just looked at the [RESULTS](https://github.com/espnet/espnet/blob/master/egs/wsj/asr1/RESULTS.md). It seems that I mistakenly used the wrong eval set (330 utterances) instead of the true eval92 set (333 utterances). I will check the setup again.

Emrys365:
The best result in [RESULTS](https://github.com/espnet/espnet/blob/master/egs/wsj/asr1/RESULTS.md) turns out to be evaluated on eval92_5k, instead of eval92...
Can't believe I hadn't seen the bug in my setup for so long! Sorry for this.

I will test the model again on the true eval92 set and update the RESULTS.

wolverineq:
> The best result in [RESULTS](https://github.com/espnet/espnet/blob/master/egs/wsj/asr1/RESULTS.md) turns out to be evaluated on eval92_5k, instead of eval92...
> Can't believe I hadn't seen the bug in my setup for so long! Sorry for this.
> 
> I will test the model again on the true eval92 set and update the RESULTS.

OK...that's a pity

wolverineq:
BTW, anyone tries specaug on WSJ? any good results?

PR #1537, opened 2020-01-28
Update WSJ results in the main README.md

Emrys365:
update the new WSJ results reported in https://github.com/espnet/espnet/pull/1527 in the main README.md

PR #1527, opened 2020-01-22
Update WSJ results using PyTorch 1.3.1 and builtin CTC

Emrys365:

PR #1652, opened 2020-03-05
correct results with builtin CTC and PyTorch 1.3 in WSJ recipe

Emrys365:
Previous results with builtin CTC and PyTorch 1.3 were mistakenly obtained from the `eval92_5k` set. Now I update the results on the standard dev and eval sets generated by the recipe.

This is related to https://github.com/espnet/espnet/pull/1421,  https://github.com/espnet/espnet/pull/1527, https://github.com/espnet/espnet/pull/1537 and https://github.com/espnet/espnet/issues/1643.

### DIFF
--- a/README.md
+++ b/README.md
@@ -98,8 +98,8 @@ We list the character error rate (CER) and word error rate (WER) of major ASR ta
 | TEDLIUM2 test          | N/A     | 8.1     | same as above                                                                                                                                                         |
 | TEDLIUM3 dev           | N/A     | 9.7     | [link](https://github.com/espnet/espnet/blob/master/egs/tedlium3/asr1/RESULTS.md#transformer-elayers12-dlayers6-units2048-8-gpus-specaug--large-lm)                   |
 | TEDLIUM3 test          | N/A     | 8.0     | same as above                                                                                                                                                         |
-| WSJ dev93              |   3.2   |   7.4   | [link](https://github.com/espnet/espnet/blob/master/egs/wsj/asr1/RESULTS.md#transformer-pytorch-13--builtin-ctc)                                                      |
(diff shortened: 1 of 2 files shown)


sw005320:
Thanks for your quick fix!

kamo-naoyuki:
@Emrys365 The results in README.md says that the best CER are 3.2 and 2.1 respectively , while egs/wsj/asr1/RESULTS.md seems to say that 4.1 and 2.7?

Emrys365:
@kamo-naoyuki 
I just reverted the results in README.md to its previous numbers. And the best results in egs/wsj/asr1/RESULTS.md is here (CERs: 2.3 (dev) and 2.1 (eval)): https://github.com/Emrys365/espnet/blob/update_wsj/egs/wsj/asr1/RESULTS.md#pytorch-4-layer-blstmp-with-no-subsampling--wide-rnn-wordlm--add-attention

kamo-naoyuki:
I see. thx.

It might be better to resort the result, I mean, we should put the RNN results of 1.3 at the top in RESULST.md since we always put the best result at the top. 

In this case, we also found the degradation due to pytorch version, or ctc implementation, so it's better to note some warnings for the difference from the best score, too.

I'm not sure which is better to put the previous best or the 1.3's best score at the top, but I think 1.3 is better because we need to move the pytorch version in the near future.

Emrys365:
If we put the [1.3's best score](https://github.com/Emrys365/espnet/blob/update_wsj/egs/wsj/asr1/RESULTS.md#rnn-pytorch-13--builtin-ctc) (CERs are 4.0 on dev and 2.4 on eval) at the top, should the numbers in the main README.md also be changed to larger ones?

sw005320:
Year, I think it's better to keep the best number in the front.
We can put a lot of detailed comments in the recipe directory.

Issue #1646, opened 2020-03-05
New argument `embed_unit` in language model breaks backward compability


unilight:
I'm using the pretrained Transformer ASR model on Librispeech as in [here](https://github.com/espnet/espnet/blob/master/egs/ljspeech/tts1/local/ob_eval/evaluate_cer.sh#L36-L40). Similar to issue #1638 , I think the new argument `embed_unit` in language model is breaking backward compatibility. In the original issue the solution seemed to be retraining it from scratch. I'm wondering if it is possible to fix the commit so I can still use the pretrained model normally.

b-flo:
Hi,

>  I'm wondering if it is possible to fix the commit so I can still use the pretrained model normally.

I guess you could use `git revert commit-sha`?

kan-bayashi:
This issue seems to affect all of the pretrained models. We should carefully keep backward compatibility.

kan-bayashi:
I will fix this issue this evening.

kan-bayashi:
Maybe fixed in #1649 
@unilight Could you check?

unilight:
Thank you for the fix!

Sorry because in [the ASR evaluation script](https://github.com/espnet/espnet/blob/master/egs/ljspeech/tts1/local/ob_eval/evaluate_cer.sh), v2 was used. After using the fix, I still have trouble. But I can now successfully run by setting api to v1.
Could you also fix, I guess `espnet/asr/pytorch_backend/recog.py` for api v2?
Also I wonder what is the difference between v1 and v2?

kan-bayashi:
Thanks, I will fix it in #1653.

kan-bayashi:
Fixed.

Issue #1638, opened 2020-03-03
New argument `embed_unit` in language model breaks an4 recipe


qmeeus:
Following [this commit](https://github.com/espnet/espnet/commit/943c9933fc047133a1ba8adc07be79957416316e):
```
Traceback (most recent call last):
  File "/users/spraak/qmeeus/spchdisk/repos/espnet/espnet/bin/asr_recog.py", line 208, in <module>
    main(sys.argv[1:])
  File "/users/spraak/qmeeus/spchdisk/repos/espnet/espnet/bin/asr_recog.py", line 190, in main
    recog(args)
  File "/esat/spchdisk/scratch/qmeeus/repos/espnet/espnet/asr/pytorch_backend/asr.py", line 664, in recog
    len(word_dict), rnnlm_args.layer, rnnlm_args.unit, rnnlm_args.embed_unit))
AttributeError: 'Namespace' object has no attribute 'embed_unit'
```
This is happening for recipe `an4/asr1` but I suspect that it is causing more issues in other recipes

sw005320:
Did it happen even you train it from scratch?

qmeeus:
No you are right, when trained from scratch the error disappears, thanks for the pointer !

PR #1653, opened 2020-03-06
Fix lm compatibility for v2

kan-bayashi:
This PR fix LM backward compatibility issue for api v2.

Related: #1646 #1638 #1623


qmpzzpmq:
Thanks a lot

Issue #1631, opened 2020-03-02
Update support version on check_install.py


Fhrozen:
I observed that the current ESPnet with Pytorch backend can support torch > 1.0 (1.1 ...) but the supported versions on check_install has not been updated. 
Is there any problem to update the supported version on check_install. 
I cannot install docker with cuda 10.1 bc torch 1.0.1 does not support it. 
Should I update it?

Fhrozen:
Yes, I will do.
Just to clarify:
In case of pytorch 1.2+, warpctc seems to not be working with (https://github.com/espnet/warp-ctc#for-pytorch-12).
So should I add an `if` in the Makefile https://github.com/espnet/espnet/blob/9bcab950d076808045bf48ece4be6385f1bd929a/tools/Makefile#L81 to avoid installing warpctc if `pytorch >=1.2`
and also in check_install.py 
to avoid checking warpctc version in the same scenario, given that calling warpctc generates a ImportError?

sw005320:
Yes, I think this is a very good idea.
Probably, it would be better to have such handlings in https://github.com/espnet/espnet/blob/master/espnet/nets/pytorch_backend/ctc.py#L20 ?

Also, we may not seem to include the test for buildin ctc.
I think it is worth to add it.

PR #1659, opened 2020-03-07
Update Install for Pytorch version

Fhrozen:
REF: #1631 

Installation checking for new pytorch versions.

BTW, I also quiet pytest. I am not sure if it is fine. IMO, the only important details are the failed tests and this option does not affect them. Showing which test has passes generates a really long list and now even larger because of the multiple tests. Let me know if this is fine or if I should reverse it.


Fhrozen:
@sw005320 ready. Maybe need @kamo-naoyuki @kan-bayashi @ShigekiKarita overview?

espnet/nets/pytorch_backend/ctc.py
```diff
@@ -22,7 +22,10 @@ def __init__(self, odim, eprojs, dropout_rate, ctc_type='warpctc', reduce=True):
         self.dropout_rate = dropout_rate
         self.loss = None
         self.ctc_lo = torch.nn.Linear(eprojs, odim)
-        self.ctc_type = ctc_type
+
+        # In case of Pytorch >= 1.2.0, CTC will be always builtin
+        torch_ver = int(torch.__version__.replace('.', '').replace('post2', ''))
```
kan-bayashi: It is better to use LooseVersion (see setup.py for example).
Also it is better to add logging.warning for notification if ctc_type is different.

tools/Makefile
```diff
@@ -6,8 +6,13 @@ PYTHON :=
 PYTHON_VERSION := 3.7.3
 CUPY_VERSION := 6.0.0
 CUDA_VERSION := 10.0
-# PyTorch version: 0.4.1 or 1.0.0 or 1.0.1
+# PyTorch version: 0.4.1, 1.0.0, 1.0.1, 1.1.0, 1.2.0, 1.3.0, 1.3.1
```
kan-bayashi: 1.4 is not working?

tools/check_install.py
```diff
@@ -13,22 +13,34 @@ def main(args):
...
         ('chainer', ("6.0.0")),
         ('chainer_ctc', None),
-        ('warpctc_pytorch', ("0.1.1", "0.1.3")),
         ('warprnnt_pytorch', ("0.1"))
     ]
 
     if not args.no_cupy:
         MANUALLY_INSTALLED_LIBRARIES.append(('cupy', ("6.0.0")))
+    
+    torch_ver = int(args.torch_version.replace('.', '').replace('post2', ''))
```
kan-bayashi: ditto

tools/Makefile
Fhrozen: I did not test the version but should be working if there is no big difference from 1.3.1

sw005320: @Emrys365 already tested and it's working (see https://github.com/espnet/espnet/pull/1527)

sw005320:
Thanks!

qmpzzpmq:
hi, According your previous benchmark of different ctc, it seems that the result of model training from warpctc is better than from builtinctc.
Have you guys test these benchmark in these newer versions pytorch. It seems that I found the built in ctc is much faster than waprtctc especially in with accerlerating.

sw005320:
We tested it with several pytorch versions, and our conclusion is that warpctc is slightly better but we don't observe so much difference in various builtin pytorch CTC versions. We did not check the speed though. Could you share your speed results with us?

qmpzzpmq:
@sw005320 I am testing recently, but I fund there is some conflict between apex with builtin ctc.

sw005320:
Could you report that conflict? I'm curious.

sw005320:
Sorry I saw your report in #1751 
Many thanks!
I'll take a look at it.

PR #1624, opened 2020-02-27
Docker updates

Fhrozen:
This is an update for docker issues (been a long time since the last update).
Related issues: #1619 #1562 #1512

- Overall updates:
  - [x] CUDA 10.1
  - [x] Docker 2 (detect whether docker 1 or 2 is installed)
  - [x] check for local docker

PR #1674, opened 2020-03-15
Update Dockerfile

goznauk:
typo & sed line change


sw005320:
Could you check #1624, which fixed some of them?

PR #1675, opened 2020-03-16
missing torch version in check_install.py

beckgom:
torch version is set to 1.0.1 but it is not included in check_install.py


kan-bayashi:
Thanks a lot!

Issue #1619, opened 2020-02-26
Error when building docker 


wentaoxandry:
When I building docker, I run espnet/docker/build.sh: sudo ./build.sh local 10.0, but I got this error:

. venv/bin/activate && pip install cupy-cuda==6.0.0

ERROR: Could not find a version that satisfies the requirement cupy-cuda==6.0.0 (from versions: none)
ERROR: No matching distribution found for cupy-cuda==6.0.0
Makefile:77: recipe for target 'cupy.done' failed
make: *** [cupy.done] Error 1
The command '/bin/sh -c if [ -z "$( nvcc -V )" ]; then         make KALDI=/kaldi CUPY_VERSION='';     else         sed -i '161s|install.py|install.py --no-cupy|' Makefile &&         sed -i '21s|nvidia-smi|nvcc|' Makefile &&         make KALDI=/kaldi CUDA_VERSION=${CUDA_VER};     fi' returned a non-zero code: 2

I have installed espnet with no error but I don't know why I got this Error by building docker.
Please help me and tell me how can I fix it.
Thank you

Fhrozen:
I am testing this. I hope to solve soon. No ETC.

Arafat4341:
I want to run docker in t2.micro ec2 instance of aws. I don't have GPU.
`./run.sh --docker_gpu 0 --docker_egs chime4/asr1 --docker_folders /export/corpora4/CHiME4/CHiME3 --dlayers 1 --ngpu 1`
is this command okay for me to work?

Fhrozen:
@Arafat4341, For not GPU should be `./run.sh --docker_gpu -1 --docker_egs chime4/asr1 --docker_folders /export/corpora4/CHiME4/CHiME3 --dlayers 1 --ngpu 0 `

Arafat4341:
@Fhrozen Thank you so much! Lets try!

Arafat4341:
`useradd: UID 0 is not unique`
I get this after completing pulling.

Fhrozen:
Are you executing `run.sh` as Root?, which is the ID=0. IF that is the case then use `./run.sh docker_user false ...`

Arafat4341:
I used `sudo` . 
After adding `--docker_user false` I get `./run.sh: invalid option --dlayers`

`Using image espnet/espnet:cpu-u18.
Executing application in Docker
[...]

PR #1677, opened 2020-03-17
Docker local updates

Fhrozen:
This is a fix for #1619.
The CUDA_VER variable was not passing to docker build so It generates the cupy installation error.


sw005320:
thanks

PR #1678, opened 2020-03-17
Update asr.sh

zh794390558:


kamo-naoyuki:
`-f 2` is appended to discard the first column. I mean that assuming the text format is same as the transcription for asr training here and we need to give the ids for each texts always at the data preparation part. LM training also requires ids for texts in espnet2.

```
uttidA (...text)
uttidB (...text)
```

zh794390558:
I think  `-f 2- ` is same to `--field 2-`.

kamo-naoyuki:
oh, you are right. sorry

PR #1689, opened 2020-03-17
Reformat the RESULTS.md in vivos

sw005320:


b-flo:
Hi shinji,

FYI it seems I made a mistake when reporting results (or overwrote correct results rather): CER for `decode_train_dev_decode_transducer_lm` in ` pytorch Transducer (enc: 4 x BLSTMP, dec: 1 x LSTM)`is wrong!

I'll fix that in my next PR, as other results will change given current work.

sw005320:
OK!

PR #1691, opened 2020-03-18
Update trainer.py

zh794390558:


kamo-naoyuki:
This will be fixed in #1685 .

(Too less information. Please write some description).

PR #1692, opened 2020-03-18
Update asr.sh

zh794390558:


kamo-naoyuki:
This --max_length is needed. It's repeatable option.

zh794390558:
Yes, I know now.

PR #1710, opened 2020-03-20
Create documentation for espnet2

kamo-naoyuki:


doc/espnet2_task.md
```diff
@@ -0,0 +1,224 @@
...
+    def optional_data_names(cls, inference: bool = False):
+        retval = ("auxially_feature",)
+        return retval
+```
+
+
+`required_data_names()` determines the mandatory data names and `optional_data_names()` gives optional data.  It means that the other names is allowed to given by command line arguments.
+
+```bash
+# The following is expected argument
```
sw005320:
```suggestion
# The following is the expected argument
```

doc/espnet2_task.md
```diff
@@ -0,0 +1,224 @@
...
+
+```python
+for ids, batch in data_loader:
+    model(**batch)
+```
+
+We provide common collate_fn and this function can support many cases, so you might not need to customize it. This collate_fn is aware of variable sequence features for seq2seq task:
+
+- The first axis of the sequence tensor from dataset must be length axis: e.g. (Length, Dim), (Length, Dim, Dim2), or (Length, ...)
+- It's not necessary to be unified the lengths of each samples and they are stacked with zero-padding.
```
sw005320:
```suggestion
- It's not necessary to make the lengths of each sample unified and they are stacked with zero-padding.
```
(???)

doc/espnet2_task.md
```diff
@@ -0,0 +1,224 @@
...
+    @classmethod
+    def build_collate_fn(cls, args):
+        # float_pad_value is used for float-tensor and int_pad_value is used for int-tensor
+        return CommonCollateFn(float_pad_value=0.0, int_pad_value=-1)
+    ```
+- The a tensor to keep the length of each samples is also appended
+    ```python
+    batch = {"speech": ..., "speech_lengths": ..., "text": ..., "text_lengths": ...}
+    ```
+- If the feature is not sequence data, this behavior can be disable.
```
sw005320:
```suggestion
- If the feature is not sequential data, this behavior can be disabled.
```

doc/espnet2_training_option.md
```diff
@@ -0,0 +1,209 @@
...
+```
+
+## Configuration file
+You can find the configuration files for DNN training in `conf/train_*.yaml`.
+
+```bash
+ls conf/
+```
+
+We adopts [ConfigArgParse](https://github.com/bw2/ConfigArgParse) for this configuration system. The configuration in yaml format has equivalent effect to the command line argument. e.g. The following two are equivalent:
```
sw005320:
```suggestion
We adopt [ConfigArgParse](https://github.com/bw2/ConfigArgParse) for this configuration system. The configuration in yaml format has an equivalent effect to the command line argument. e.g. The following two are equivalent:
```

doc/espnet2_training_option.md
```diff
@@ -0,0 +1,209 @@
...
+
+```bash
+python -m espnet2.bin.asr_train --config conf/config.yaml
+python -m espnet2.bin.asr_train --foo 3 --bar 4
+```
+
+
+## Change the configuration for dict type value
+
+Come parameters are named as `*_conf`, e.g. `optim_conf`, `decoder_conf` and they has the `dict` type value. We also provide the way to configure the nested value in such dict object.
```
sw005320: ???
```suggestion
Some parameters are named as `*_conf`, e.g. `optim_conf`, `decoder_conf` and they has the `dict` type value. We also provide a way to configure the nested value in such a dict object.
```

doc/espnet2_training_option.md
```diff
@@ -0,0 +1,209 @@
...
+- Saving model and checkpoint
+- Show result in the epoch
+
+Sometimes the examination after training with whole corpus is too coarse if using large corpus: `--num_iters_per_epoch` restrict the number of iteration of each epochs.
+
+```bash
+python -m espnet2.bin.asr_train --num_iters_per_epoch 1000
+```
+
+Note that he training process can't be resumed at the middle of an epoch because data iterators is stateless, but instead of it, the iterators for each epochs can be built with the specific epoch number deterministically, just like:
```
sw005320:
```suggestion
Note that the training process can't be resumed at the middle of an epoch because data iterators are stateless, but instead of it, the iterators for each epoch can be built with the specific epoch number deterministically, just like:
```

doc/espnet2_training_option.md
```diff
@@ -0,0 +1,209 @@
...
+
+## The relation between mini-batch size and number of GPUs
+
+In ESPnet1, we support three type of mini-batch type:
+
+- --batch-count seq
+- --batch-count bin
+- --batch-count frame
+
+For now, ESPnet2 supports only `seq` mode. Do as following to change the batch-size.
```
sw005320: ???
```suggestion
For now, ESPnet2 supports only `seq` mode. The batch-size can be changed as follows:
```

doc/espnet2_training_option.md
```diff
@@ -0,0 +1,209 @@
...
+- ESPnet2: The batch-size is not changed regardless of the number of GPUs.
+    - Therefore, you should more number of batch-size than the number of GPUs.
+    ```bash
+    python -m espnet.bin.asr_train --batch_size 10 --ngpu 2  # Actual batch_size is 10 and each GPU devices are assigned to 5
+    ```
+
+Note that even espnet1, if using `bin` or `frame` batch-count, this changing of batch_size is not done.
+
+## Gradient accumulating
+There are several measures to deal large model architecture than your GPU device memory when training.
```
sw005320:
```suggestion
There are several ways to deal with larger model architectures than the capacity of your GPU device memory during training.
```

doc/espnet2_training_option.md
```diff
@@ -0,0 +1,209 @@
...
+
+- Using more number of GPUs
```
sw005320:
```suggestion
- Using a larger number of GPUs
```

doc/espnet2_training_option.md
```diff
@@ -0,0 +1,209 @@
...
+- Using half decision tensor
```
sw005320:
```suggestion
- Using a half decision tensor
```

doc/espnet2_training_option.md
```diff
@@ -0,0 +1,209 @@
...
+python -m espnet.bin.asr_train --train_dtype float16  # Just training with half precision
+python -m espnet.bin.asr_train --train_dtype float32  # default
+python -m espnet.bin.asr_train --train_dtype float64
+python -m espnet.bin.asr_train --train_dtype O0  # opt_level of apex
+python -m espnet.bin.asr_train --train_dtype O1  # opt_level of apex
+python -m espnet.bin.asr_train --train_dtype O2  # opt_level of apex
+python -m espnet.bin.asr_train --train_dtype O3  # opt_level of apex
+```
+
+## Reprodusibility and determinism
```
sw005320:
```suggestion
## Reproducibility and determinization
```
???

doc/espnet2_training_option.md
```diff
@@ -0,0 +1,209 @@
...
+There are some possibilities to make training not-reproducible.
```
sw005320:
```suggestion
There are some possibilities to make training non-reproducible.
```

doc/espnet2_training_option.md
```diff
@@ -0,0 +1,209 @@
...
+
+- Initialization of parameters which comes from pytorch/espnet version difference.
+- Reducing order for float values when multi GPUs
+    - I don't know whether NCCL is deterministic or not.
+- Random seed difference
+    - We fixed the random seed for each epochs, so the randomness should be reproduced even if the process is resumed.
+- CuDNN or some non-deterministic operations for CUDA: See https://pytorch.org/docs/stable/notes/randomness.html
+
+By default, CuDNN performs non-deterministic mode and it can be changed by:
```
sw005320: I thought that we perform `--cudnn_deterministic true` as default.
Could you double-check it?
It's better to use deterministic as default.

doc/espnet2_tutorial.md
```diff
@@ -0,0 +1,172 @@
...
+- **Chainer free**
+  - Discarding Chainer completely.
+  - The development of Chainer is stopped at v7: https://chainer.org/announcement/2019/12/05/released-v7.html
+- **Kaldi free**
+  - It's not mandatory to compile Kaldi.
+      - **If you find some recipes requiring Kaldi mandatory, please report it. It should be dealt as a bug in ESPnet2.**
+  - We still support the features made by Kaldi optionally.
+  - We still follow Kaldi style. i.e. depending on `utils/` of Kaldi.
+- **On the fly** feature extraction & text preprocessing for training
+  - You don't need to create the features file before training, but just inputting wave data directly.
```
sw005320:
```suggestion
  - You don't need to create the feature file before training, but just input wave data directly.
```

doc/espnet2_tutorial.md
```diff
@@ -0,0 +1,172 @@
...
+      - We support both raw wave inputting and extracted feature.
```
sw005320:
```suggestion
      - We support both raw wave input and extracted feature.
```

doc/espnet2_tutorial.md
```diff
@@ -0,0 +1,172 @@
...
+  - The preprocessing for Text, tokenization to characters, or sentencepieces, can be also applied in training time.
```
sw005320:
```suggestion
  - The preprocessing for text, tokenization to characters, or sentencepieces, can be also applied during training.
```

doc/espnet2_tutorial.md
```diff
@@ -0,0 +1,172 @@
...
+- Discarding JSON format for describing training corpus in espnet1.
```
sw005320:
```suggestion
- Discarding the JSON format for describing the training corpus in espnet1.
```

doc/espnet2_tutorial.md
```diff
@@ -0,0 +1,172 @@
...
+    - Why do we discard JSON format? Because a dict object generated from large JSON file requires much memory and it also takes much times to parse such large JSON file.
```
sw005320:
```suggestion
    - Why do we discard the JSON format? Because a dict object generated from a large JSON file requires much memory and it also takes much time to parse such a large JSON file.
```

doc/espnet2_tutorial.md
```diff
@@ -0,0 +1,172 @@
...
+    cd egs2/an4/
+    ./asr1/run.sh
+    ```
+
+    ```bash
+    # Doesn't work
+    cd egs2/an4/asr1/local/
+    ./data.sh
+    ```
+
```
sw005320: It's better to add the case that it does work.

doc/espnet2_tutorial.md
```diff
@@ -0,0 +1,172 @@
...
+      - steps/     # Kaldi utilities
+      - utils/     # Kaldi utilities
+      - db.sh      # The directory path of each corpora
+      - path.sh    # Setup script for environment variables
+      - cmd.sh     # Configuration for your backend of job scheduler
+      - run.sh     # Entry point
+      - asr.sh     # Invoked by run.sh
+    ```
+
+    - You need to modify `db.sh` for specifying your corpus before executing `run.sh`. For example, you'll the recipe of `egs2/wsj`, you need to change the value of `WSJ0` and `WSJ1`.
```
sw005320:
```suggestion
    - You need to modify `db.sh` for specifying your corpus before executing `run.sh`. For example, when you' touch the recipe of `egs2/wsj`, you need to change the paths of `WSJ0` and `WSJ1` in `db.sh`.
```

doc/espnet2_tutorial.md
```diff
@@ -0,0 +1,172 @@
...
+        - Some corpus can be freely obtained from WEB and they are written as "downloads/" at the initial state. You can also change them to your corpus path if it's already downloaded.
```
sw005320:
```suggestion
        - Some corpora can be freely obtained from the WEB and they are written as "downloads/" at the initial state. You can also change them to your corpus path if it's already downloaded.
```

doc/espnet2_tutorial.md
```diff
@@ -0,0 +1,172 @@
...
+    - `path.sh` is used to setup the environment for `run.sh`. Note that the Python interpreter used for ESPnet is not the current Python of your terminal, but it's the Python which was installed at `tools/`. Thus you need to source `path.sh` to use the Python.
```
sw005320:
```suggestion
    - `path.sh` is used to set up the environment for `run.sh`. Note that the Python interpreter used for ESPnet is not the current Python of your terminal, but it's the Python which was installed at `tools/`. Thus you need to source `path.sh` to use this Python.
```

doc/espnet2_tutorial.md
```diff
@@ -0,0 +1,172 @@
...
+```
+
+### Use tensorboard
+
+```bash
+tensorboard --logdir exp/asr_train_<some-name>/tensorboard/
+```
+
+# Instruction for run.sh
+We uses all python commands via `run.sh` and you may need to invoke Python script directly in some cases, but `run.sh` itself is configurable and you can live comfortably with it in many cases without any modifications.
```
sw005320:
```suggestion
We use all python commands via `run.sh` and you may need to invoke a Python script directly in some cases, but `run.sh` itself is configurable and you can live comfortably with it in many cases without any modifications.
```

doc/espnet2_tutorial.md
```diff
@@ -0,0 +1,172 @@
...
+```
+
+Then you can change the value as following:
+
+```bash
+$ ./run.sh --ngpu 2
+echo 2
+```
+
+You can also show the help message if it supports.
```
sw005320:
```suggestion
You can also show the help message whether it is supported.
```
???

doc/espnet2_tutorial.md
```diff
@@ -0,0 +1,172 @@
...
+The procedures in `run.sh` can be divided into some stages, e.g. data preparation, training, and evaluation. You can specify the starting stage and the stopping stage.
+
+```bash
+./run.sh --stage 2 --stop-stage 6
+```
+
+## Change the configuration for training
+See  about the usage of training tool.
+
+(The following is the case of ASR training and you need to replace accordingly)
```
sw005320:
```suggestion
(The following is the case of ASR training and you need to replace it accordingly)
```

doc/espnet2_tutorial.md
```diff
@@ -0,0 +1,172 @@
...
+
+```bash
+# Give configuration file
```
sw005320:
```suggestion
# Give a configuration file
```

kamo-naoyuki:
@sw005320 Many thanks and sorry for typos... I didn't expect that you would fix.

doc/espnet2_training_option.md
kamo-naoyuki: Deterministic mode shouldn't be true by default because it might be lower performance than non-deterministic.

https://pytorch.org/docs/stable/notes/randomness.html#cudnn

doc/espnet2_distributed.md
```diff
@@ -0,0 +1,165 @@
...
+You can disable distributed mode and switch to threading based data parallel as following:
+
+```bash
+% python -m espnet2.bin.asr_train --ngpu 4 --multiprocessing_distributed false
+```
+
+### 2Nodes and 2GPUs for each node with multiprocessing distributed mode
+Note that multiprocessing distributed mode assumes same number of GPUs for each nodes.
+
+```bash
```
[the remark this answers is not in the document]
kamo-naoyuki: `--help` option shows the explanations.

doc/espnet2_training_option.md
sw005320: Yes, but I want to stick to make it deterministic as much as possible for the following reasons.

- Many researchers want to have deterministic implementations for fair comparisons. I know that it is not deterministic due to the GPU, GPU driver, CUDA, CUDNN etc., but it still reproduces the same result with a specific condition
- It is very hard to debug it without the deterministic mode (the deterministic mode with the fixed seed assures that the forward computation is the same at least).
- If people wrongly set the non-deterministic mode in the inference, it would be more serious (ASR/TTS results would be changed every trial).

Could you reconsider to make the deterministic mode default?
(I also once tested the non-deterministic and deterministic mode and I could not observe so much difference).

doc/espnet2_distributed.md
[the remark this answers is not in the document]
sw005320: Is it possible to add very brief explanations? I think it would help the user a lot if the document is self-consistent.

doc/espnet2_training_option.md
kamo-naoyuki: O.K. I'll follow you though I still hesitate because the performance difference is dependent on the model architecture and GPU device. ESpnet happens not to use deep cnn arch, so this was not big issue.

In my small benchmark,  it's 1% slower for resnet152 with Geforce RTX 2080ti, v100, tesla-k80, Geforce 1080ti. This is acceptable for me.

```python
import torch
from torchvision.models import resnet

def main():
    print(torch.backends.cudnn.version())
    torch.manual_seed(0)
    conv = resnet.resnet152().cuda()
    x = torch.randn(30, 3, 224, 224).cuda()
    ntime = 20

    for benchmark in [True, False]:
        for deterministic in [False, True]:
            with torch.backends.cudnn.flags(
                deterministic=deterministic, benchmark=benchmark, verbose=True,
            ):
                start_event = torch.cuda.Event(enable_timing=True)
                end_event = torch.cuda.Event(enable_timing=True)

                start_event.record()
                for i in range(ntime):
                    y = conv(x)
                    loss = y.mean()
                    loss.backward()
                    conv.zero_grad()
                    if i == 0:
                        end_event.record()
                        torch.cuda.synchronize()
                        first = start_event.elapsed_time(end_event)
                        start_event.record()

                    end_event.record()
                    torch.cuda.synchronize()
                    times = start_event.elapsed_time(end_event)

                print(
                    f"benchmark={benchmark}, deterministic={deterministic}: "
                    f"{times / ntime:.4e}[s/iter] "
                    f"(at the first iteration: {first:.4e}[s/iter])"
                )
main()
```
```
7603
benchmark=True, deterministic=False: 5.9307e+02[s/iter] (at the first iteration: 9.6397e+02[s/iter])
benchmark=True, deterministic=True: 6.0100e+02[s/iter] (at the first iteration: 6.2919e+02[s/iter])
benchmark=False, deterministic=False: 6.0787e+02[s/iter] (at the first iteration: 6.3496e+02[s/iter])
benchmark=False, deterministic=True: 6.1303e+02[s/iter] (at the first iteration: 6.4359e+02[s/iter])
```

doc/espnet2_distributed.md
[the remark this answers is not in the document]
kamo-naoyuki: I added

doc/espnet2_training_option.md
kamo-naoyuki: I changed determinic mode to default

doc/espnet2_tutorial.md
```diff
@@ -0,0 +1,176 @@
...
+    ```
+
+    - You need to modify `db.sh` for specifying your corpus before executing `run.sh`. For example, when you touch the recipe of `egs2/wsj`, you need to change the paths of `WSJ0` and `WSJ1` in `db.sh`.
+      - Some corpora can be freely obtained from the WEB and they are written as "downloads/" at the initial state. You can also change them to your corpus path if it's already downloaded.
+    - `path.sh` is used to set up the environment for `run.sh`. Note that the Python interpreter used for ESPnet is not the current Python of your terminal, but it's the Python which was installed at `tools/`. Thus you need to source `path.sh` to use this Python.
+        ```bash
+        source path.sh
+        python
+        ```
+    - `cmd.sh` is used for specifying the backend of the job scheduler. If you don't have such a system in your machine environment, you don't need to change anything about this file. See [Using Job scheduling system](./parallelization.md)
```
sw005320, line 91:
```suggestion
    - `cmd.sh` is used for specifying the backend of the job scheduler. If you don't have such a system in your local machine environment, you don't need to change anything about this file. See [Using Job scheduling system](./parallelization.md)
```

doc/espnet2_tutorial.md
```diff
@@ -0,0 +1,176 @@
...
+
+```bash
+# Accuracy plot
+# (eog is Eye of GNOME Image Viewer)
+eog exp/asr_train_<some-name>/images/acc.img
+# Attention plot
+eog exp/asr_train_<some-name>/att_ws/<sample-id>/<param-name>.img
+```
+
+### Use tensorboard
```
sw005320, line 124: Maybe better to add some descriptions about how to install tensorboard and a simple usage like https://espnet.github.io/espnet/tutorial.html#logging.
We don't include the tensorboard installation as default (we could do it though).

sw005320: Maybe it is OK to leave it as it is but in the near future, we may need to prepare a complete README for espnet2 that merges both espnet1 and espnet2 documentations.

doc/espnet2_training_option.md
```diff
@@ -0,0 +1,422 @@
...
+```
+
+The behavior for batch-size during multi-GPU training is **different from that of ESPNet1**.
+
+- ESPNet1: The batch-size will be multiplied by the number of GPUs.
+  ```bash
+  python -m espnet.bin.asr_train --batch_size 10 --ngpu 2 # Actual batch_size is 20 and each GPU devices are assigned to 10
+  ```
+- ESPnet2: The batch-size is not changed regardless of the number of GPUs.
+  - Therefore, you should more number of batch-size than the number of GPUs.
```
sw005320, line 180:
```suggestion
  - Therefore, you should set a more number of batch-size than that of GPUs.
```

doc/espnet2_task.md
```diff
@@ -0,0 +1,267 @@
...
+On the contrary, unlike `fairseq` or training system such as `pytorch-lightining`, 
+our `Task` class doesn't have an interface for building the dataset explicitly.
+This is because we aim at the task related to speech/text only, 
+so we don't need such general system so far.
+
+The following is an example of the command lint arguments:
+
+```bash
+python -m espnet2.bin.asr_train \
+  --train_data_path_and_name_and_type=/some/path/tr/wav.scp,speech,sound \
```
sw005320, line 67: `audio` might be better (more specific and appropriate) than `sound`?
This is just a comment.

kamo-naoyuki: I also thought this, but I couldn't decide. `sound` comes from `sndfile`.
I don't know whether it's necessary,  `audio` is left for `torchaudio` ( or `sox`)

doc/espnet2_tutorial.md
kamo-naoyuki: Sorry, I already added tensorboard.

https://github.com/espnet/espnet/blob/eca4fa073a826957c91069501844d4e2b9392cc9/setup.py#L33-L34

This is  because pytorch1.1? officially start to support tensorboard logging and it requires to install tensorboard. 
I decided to use the pytorch's SummaryWriter instead of tensorboardx if pytorch>=1.1, so we need to install tensorboard.

sw005320: Make sense for me.
This is very good.

doc/espnet2_task.md
sw005320: OK.

kamo-naoyuki:
@sw005320 Thank you for proofreading.
 I'll merge this PR, but I haven't written some important topics yet, just a note:

- Calc stats mode
- On the fly preprocessing
- Chunk iterator
- launch.py

kamo-naoyuki:
@kan-bayashi v.0.7.0 is ready now. Please merge it into master if the other completed PRs are merged.

PR #1711, opened 2020-03-21
[WIP] add recipe for French Polyphone: ELRA-S0030_02

AdolfVonKleist:
This is a draft recipe proposal for the French Polyphone [ELRA-S0030_02] corpus.  The language is French, the locale is Switzerland, the domain is general telephony.  It includes results, configs, and a data preparation script derived from the default directory structure as received from any purchasing entity via ELRA.  It is a work-in-progress.


b-flo:
Hi,

I'm not sure what is the policy on that @sw005320, but I think "swiss french" should be specified here?

AdolfVonKleist:
> I'm not sure what is the policy on that @sw005320, but I think "swiss french" should be specified here?

good point. `french_polyphone` -> `swiss_french_polyphone` perhaps?

sw005320:
> > I'm not sure what is the policy on that @sw005320, but I think "swiss french" should be specified here?
> 
> good point. `french_polyphone` -> `swiss_french_polyphone` perhaps?

Sure, that is good.

sw005320:
Could you fix the CI error (https://travis-ci.org/github/espnet/espnet/jobs/665092097#L2367-L2374)? We'll start the review then.

AdolfVonKleist:
> Could you fix the CI error (https://travis-ci.org/github/espnet/espnet/jobs/665092097#L2367-L2374)? We'll start the review then.

I have made these, and the changes to the naming conventions, hopefully in the intended way.
Some things that are maybe a bit different: the data prep script is written in python, and uses a yaml configuration file stored in `conf/`.   The data prep script allows for re-generation of new train/dev/test splits but provides the fixed lists used for the two reported experiments [with/without specaug].  The `local/` dir contains a `.pylintrc` used to lint the script; I did not find one in the espnet repo so I used a local one; can switch or amend.  The `beam-size` is quite low; but we have not found major advantages to much larger beams so far; this setting seems to achieve good accuracy and runs well under 1xRT, but it is not optimized here.

egs/polyphone_swiss_french/asr1/run.sh
```diff
@@ -0,0 +1,225 @@
...
+# decoding parameter
+recog_model=model.acc.best # set a model to be used for decoding: 'model.acc.best' or 'model.loss.best'
+n_average=10
+
+# bpemode (unigram or bpe)
+nbpe=800
+bpemode=unigram
+
+# exp tag
+tag="specaug" # tag for managing experiments.
```
sw005320: We can make a tag variable blank and add specaug related tag through the `preprocess_conf` variable, e.g., 
https://github.com/espnet/espnet/blob/master/egs/librispeech/asr1/run.sh#L203-L213

sw005320:
> Some things that are maybe a bit different: the data prep script is written in python, and uses a yaml configuration file stored in `conf/`. The data prep script allows for re-generation of new train/dev/test splits but provides the fixed lists used for the two reported experiments [with/without specaug]. 

This is a very interesting way (also it is pythonic and very nice).
I was first confused because it requires `conf/fbank.conf`, but I found that it is to get the correct sampling rate.
I think this is a very cool way and may also be used for the other recipes. (@kamo-naoyuki, it might be a good way for your espnet2).

> The `local/` dir contains a `.pylintrc` used to lint the script; I did not find one in the espnet repo so I used a local one; can switch or amend. 

We did not prepare it.
@kan-bayashi, do you think we should provide it?
It may reduce CI errors.

> The `beam-size` is quite low; but we have not found major advantages to much larger beams so far; this setting seems to achieve good accuracy and runs well under 1xRT, but it is not optimized here.

Year, I think our default beam size is too large.
Our intention is to get a better number, but this confuses a lot of people (they just thought that espnet is slow), so we should definitely set better numbers for the beam size.
Thanks for your suggestions!

sw005320:
LGTM, but I want to get a comment about `.pylintrc ` from @kan-bayashi. We may put it in the root directory.

AdolfVonKleist:
> This is a very interesting way (also it is pythonic and very nice).
> I was first confused because it requires `conf/fbank.conf`, but I found that it is to get the correct sampling rate.

Yes I just wanted to avoid duplicating the setting in multiple places, and we also work mainly in 8kHz.  Probably there is still a better way.

> > The `local/` dir contains a `.pylintrc` used to lint the script; I did not find one in the espnet repo so I used a local one; can switch or amend.
> We did not prepare it.
> @kan-bayashi, do you think we should provide it?
> It may reduce CI errors.

It is derived from one of the google variants, but there are numerous custom elements disabled, you can check them here, just in case, and should make sure you agree with them:

https://github.com/espnet/espnet/blob/c821466b20d723811d6be754013a3d289bfb2f0e/egs/french_polyphone/asr1/local/.pylintrc#L68

> Year, I think our default beam size is too large.
> Our intention is to get a better number, but this confuses a lot of people (they just thought that espnet is slow), so we should definitely set better numbers for the beam size.

Yes! I wanted to mention it because I have seen many comments to this effect, but it is actually quite performant and accurate at smaller `beam-size`s even when decoding in CPU only mode.

sw005320:
> LGTM, but I want to get a comment about `.pylintrc ` from @kan-bayashi. We may put it in the root directory.

@kan-bayashi, do you have some comments about it?

kamo-naoyuki:
@sw005320 We should merge quickly. Keeping long time will be a burden for the commiter.

Though we should add a linter for local to our CI, but it's not related to this PR. We should discuss it in another thread. 
Are there some reasons why we don't put it for local scripts? @sw005320 

Anyway, we doesn't use pylint but use pycodestyle, so could your remove for now? @AdolfVonKleist

sw005320:
ok

kamo-naoyuki:
@AdolfVonKleist Thank you. I'll merge.

AdolfVonKleist:
One question, and sorry if this should now be separate: I'm not familiar with `pycodestyle`, I would like to use your rules for any future similar prs.  Can you indicate how I should do that? Thanks!

kamo-naoyuki:
@AdolfVonKleist We are just applying pycodestyle and flake8 here: https://github.com/espnet/espnet/blob/master/ci/test_python.sh
(But we don't apply it to local/*.py for now.)

If you asked about the setting, then please see https://github.com/espnet/espnet/blob/master/setup.cfg 

The CI automatically warns you when PR if the code style differs, so basically you don't need to care much.


BTW, is this a python file? https://github.com/espnet/espnet/blob/master/egs/polyphone_swiss_french/asr1/local/data_prep
Sorry after merging, but why didn't you name it as `data_prep.py`?

AdolfVonKleist:
> BTW, is this a python file? https://github.com/espnet/espnet/blob/master/egs/polyphone_swiss_french/asr1/local/data_prep
> Sorry after merging, but why didn't you name it as `data_prep.py`?

Yes it is a python file.  It is just a habit from writing `postinst`, `prerm`, etc. configuration scripts for deb packages, where the 'script' should have the same name but might sometimes be bash, sometime perl, python, etc.  I guess in _this_ case it is probably why the internal `pycodestyle` check was not triggered?  I would like to add the asr2 version of this recipe for comparison, when you have a transition guide ready; maybe then I can change this to `.py` in a next pr?

kamo-naoyuki:
Yeah, your guess is exactly right, but not only. I think such habit is only used for such init scripts in traditional unix system. The file extension can assist our editor, or github, to detect the file type, so.

Then would you rename it data_prep.py?

AdolfVonKleist:
> Yeah, your guess is exactly right, but not only. I think such habit is only used for such init scripts in traditional unix system. The file extension can assist our editor, or github, to detect the file type, so.
> 
> Then would you rename it data_prep.py?

Yes, is it ok to amend it in a next pr with the v2 example? Then I will also make sure to run `pycodestyle` check locally before submitting.

kamo-naoyuki:
If not troublesome, please send it in another.

PR #1712, opened 2020-03-21
--max_length -> --fold_length

kamo-naoyuki:


sw005320:
@kamo-naoyuki, thanks!
This change looks good but I'm concerning that you remove recipe specific values since this will change the result. Do you want us to tune it again?

kamo-naoyuki:
@sw005320 Hopefully, this changing doesn't affect result much, but I don't know.

Anyway we need to retune some recipes because the condition is unified, e.g. pytorch-version, warp-ctc or cudnn-ctc, and w/ or w/o pitch, but I think it's not time. I'll implement batch-count=bin or frame for espnet2 and we should use it.

sw005320:
Sure, this is just a note that. I think we may need to redo our experiments anyway.

kamo-naoyuki:
I'll merge this PR.

PR #1714, opened 2020-03-21
Add --name option for training job

kamo-naoyuki:
I changed the job-name of a training command to the log file name. This is my fashion for experiments and we can distinguish jobs with qstat/squeue command by their job names.


sw005320:
I like this idea.
Thanks!

PR #1724, opened 2020-03-23
Add wsj results for espnet2

kamo-naoyuki:


sw005320:
Hmm, we still seem to have a gap compared with https://github.com/espnet/espnet/blob/master/egs/wsj_mix/asr1/run.sh (also it requires only 100 epochs). Do you think you can manage to get a comparable result?

kamo-naoyuki:
@sw005320  wsj_mix?
The result is already better than espnet1 about transformer with pytorch=1.4 
 
espnet1: 4.1, 2.7
espent2: 3.6,2.6

We can't compare the result directly because the LM is different.

sw005320:
Sorry I mean to put https://github.com/espnet/espnet/blob/master/egs/wsj/asr1/RESULTS.md#transformer-pytorch-13--builtin-ctc

OK.
Do we require to have 200 epochs?

kamo-naoyuki:
Sorry, I didn't evaluate 100epoch model. The loss looks converged at more early epoch. 

Tuning note:

I tried to change batch-size/accum_gard larger as possible in this experiment for training speed and I found

- accum_gard=8 may be upper limit for WSJ. 16 performs worse obviously.
- It seems better to set warmup_step so that the lr is kept to increate at the end of training or a few epochs ago.

kamo-naoyuki:
I got similar results with 100epoch model.

sw005320:
LGTM.
If it's ready, please merge it.

PR #1731, opened 2020-03-24
Fix the gpu device when resuming

kamo-naoyuki:
(I think this is a bug of pytorch)

```python
# Always cuda:0
state = torch.load("a.pth", map_location="cuda")
# Always cuda:0
state = torch.load("a.pth", map_location=torch.device("cuda"))
# Current device
state = torch.load("a.pth", map_location=f"cuda:{torch.cuda.current_device())}")

```

Therefore I just changed it.


kamo-naoyuki:
Current code is not fixed, don't merge.

kamo-naoyuki:
O.K. Ready to be merged.

This bug will be fixed at pytorch1.5: https://github.com/pytorch/pytorch/issues/35289

sw005320:
Thanks!

PR #1732, opened 2020-03-24
Change cudnn deterministic mode to default

kamo-naoyuki:
We discussed in #1710 and decided to turn  cudnn deterministic mode on by default. (This is the default behavior of espnet1).

Note that the performance might be slower in some cases, but in my benchmark that is enough small.


kamo-naoyuki:
I'll merge this PR.

PR #1733, opened 2020-03-24
Change the type of --chunk_length to str_or_int

kamo-naoyuki:
--chunk_length can be int or str like "100,200,400" (=[100, 200, 400]) or "100-300" (=range(100, 300)).

(type=str is no problem)


kamo-naoyuki:
I'll merge this PR.

PR #1736, opened 2020-03-25
Update e2e_asr_common.py

zh794390558:
fix label smoothing bug


sw005320:
Could you explain what would be a problem in the previous one?

zh794390558:
I think `transcript` is a filepath, so can not get `len`. So need cnt how mayn transcripts in the file.

sw005320:
Oh, I see.
Another possibility is to get the `<eos>` count from `len(trans_json)`, i.e., we can count the `<eos>` even when we don't have any token in an utterance.

@takaaki-hori, I think it is a bug.
Can you also confirm it?

zh794390558:
I have no time to fix this, please take over this PR.

PR #1737, opened 2020-03-25
Log  rotation for launch.py

kamo-naoyuki:
Kaldi's run.pl, and etc. overwrite the log file regardless of the options. This is hard coding, so we can't change it to append mode. This issue might be addressed in Kaldi, but I won't discuss about it.

In espnet2 we always submit training jobs via launch.py for distributed training, so I implemented log-rotation in launch.py.

### DIFF
--- a/espnet2/bin/launch.py
+++ b/espnet2/bin/launch.py
@@ -28,6 +28,11 @@ def get_parser():
         "--log", help="The path of log file used by cmd", default="run.log",
     )
     parser.add_argument(
+        "--max_num_log_files",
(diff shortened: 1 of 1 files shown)


sw005320:
Thanks! 
This is very useful.
Do you think 30 is enough? 
(how about 100? I don't have a good sense, but this is my personal opinion)

kamo-naoyuki:
O.K. I'll change it to 1000. There are no strong reasons for 30, just for avoiding keep infinite number.

kamo-naoyuki:
I'll merge this PR.

PR #1742, opened 2020-03-26
Implement BatchBinSampler

kamo-naoyuki:


kamo-naoyuki:
When distributed, the current code splits each mini-batch by samples, e.g. `bs=12` -> `bs=12//world_size`, but the data size for each process differs in this method although batch-bin sample unified the size for each batch.

One of the solution is splitting mini-batches into sub groups instead of splitting within batch, e.g. `num_batches=40` -> `num_batches=40//world_size`, but in this case we must discard the remainder mini-batches of world_size because the number of iterations must equal to each process. I adopts this method for ChunkIterator because it's only way if the iterator is a generator in the term of python. 

I think that this restriction in distributed data parallel is very unuseful, so I'll do something to solve this issue by another way.

sw005320:
@ShigekiKarita, could you take a look at this PR?

espnet2/samplers/abs_sampler.py
```diff
@@ -0,0 +1,16 @@
...
+from abc import abstractmethod
+from typing import Iterator
+from typing import Tuple
+
+from torch.utils.data import Sampler
+
+
+class AbsSampler(Sampler, ABC):
+    @abstractmethod
+    def __len__(self):
```
ShigekiKarita:
```suggestion
    def __len__(self) -> int:
```

espnet2/samplers/batch_bin_sampler.py
```diff
@@ -0,0 +1,153 @@
...
+
+from espnet2.samplers.abs_sampler import AbsSampler
+from espnet2.utils.fileio import load_num_sequence_text
+
+
+class BatchBinSampler(AbsSampler):
+    def __init__(
+        self,
+        batch_bins: int,
+        shape_files: Union[Tuple[str, ...], List[str]],
```
ShigekiKarita: Maybe `typing.Sequence[str]` is better?

ShigekiKarita: Maybe `typing.Sequence[str]` is better?

espnet2/samplers/batch_bin_sampler.py
```diff
@@ -0,0 +1,153 @@
...
+            f"sort_in_batch={self.sort_in_batch}, "
+            f"sort_batch={self.sort_batch})"
+        )
+
+    def __len__(self):
+        return len(self.batch_list)
+
+    def __iter__(self) -> Iterator[Tuple[str, ...]]:
+        for batch in self.batch_list:
+            yield batch
```
ShigekiKarita:
```suggestion
        return iter(self.batch_list)
```

espnet2/samplers/batch_bin_sampler.py
kamo-naoyuki: (You don't need to check typing though there are some miss, that is not related to this PR)

I basically don't use Sequence[str].
str is also `Sequence[str]`, so the desired type is actually `not str and Sequence[str]`, but this looks a little bit ugly, so.

espnet2/samplers/batch_bin_sampler.py
```diff
@@ -0,0 +1,153 @@
...
+                )
+            self.batch_list.append(tuple(minibatch_keys))
+
+        if sort_batch == "ascending":
+            pass
+        elif sort_batch == "descending":
+            self.batch_list.reverse()
+        else:
+            raise ValueError(
+                f"sort_batch must be ascending or descending: {sort_batch}"
```
ShigekiKarita: It would be better If this validation were at the beginning of this func

espnet2/samplers/batch_bin_sampler.py
ShigekiKarita: It looks crazy. I hate it too indeed.

espnet2/samplers/batch_bin_sampler.py
kamo-naoyuki: applied.

espnet2/samplers/build_batch_sampler.py
```diff
@@ -0,0 +1,94 @@
...
+    sort_in_batch: str = "descending",
+    sort_batch: str = "ascending",
+    drop_last: bool = False,
+    min_batch_size: int = 1,
+    padding: bool = True,
+) -> AbsSampler:
+    """Helper function to instantiate BatchSampler
+
+    Args:
+        type: mini-batch type. "constant", "seq", "bin", or, "frame"
```
ShigekiKarita: Here are readability issues and the discrepancy between this `type` names and the implementation class names.  And we can forget espnet1 compatibility and reconsider better names. For example

- `const_no_sort` `ContantBatchSampler` => `unsorted` `UnsortedBatchSampler`
- `const` `ConstantSortedBatchSampler` => `sorted` `SortedBatchSampler`
- `seq` `SequenceBatchSampler` => `target-length` `TargetLengthBatchSampler`
- `frame` NA => `source-length` `SourceLengthBatchSampler`
- `bins` `BatchBinSampler` => `source-target-numel` `SourceTargetNumelBatchSampler`

ref: [numel](https://pytorch.org/docs/stable/torch.html#torch.numel)

espnet2/samplers/sequence_batch_sampler.py
```diff
@@ -0,0 +1,135 @@
...
+            f"sort_in_batch={self.sort_in_batch}, "
+            f"sort_batch={self.sort_batch})"
+        )
+
+    def __len__(self):
+        return len(self.batch_list)
+
+    def __iter__(self) -> Iterator[Tuple[str, ...]]:
+        for batch in self.batch_list:
+            yield batch
```
ShigekiKarita:
```suggestion
        return iter(self.batch_list)
```

espnet2/samplers/constant_sorted_batch_sampler.py
```diff
@@ -0,0 +1,96 @@
...
+            f"sort_in_batch={self.sort_in_batch}, "
+            f"sort_batch={self.sort_batch})"
+        )
+
+    def __len__(self):
+        return len(self.batch_list)
+
+    def __iter__(self) -> Iterator[Tuple[str, ...]]:
+        for batch in self.batch_list:
+            yield batch
```
ShigekiKarita:
```suggestion
        return iter(self.batch_list)
```

espnet2/samplers/constant_batch_sampler.py
```diff
@@ -0,0 +1,68 @@
...
+            f"batch_size={self.batch_size}, "
+            f"key_file={self.key_file}, "
+        )
+
+    def __len__(self):
+        return len(self.batch_list)
+
+    def __iter__(self) -> Iterator[Tuple[str, ...]]:
+        for batch in self.batch_list:
+            yield batch
```
ShigekiKarita:
```suggestion
        return iter(self.batch_list)
```

espnet2/samplers/build_batch_sampler.py
kamo-naoyuki: I didn't care about the naming. `unsorted` and `sorted` sound good.

What does `target-length` means? SequenceBatchSampler takes into account all input sequences i.e.  the input and the target?

ShigekiKarita: > SequenceBatchSampler takes into account all input sequences i.e. the input and the target?

Oh, yes. You are right. The `seq` option is too complicated!

kamo-naoyuki: hmm, I don't have good name.
seq and frame have similar behaviour and we need to name explicitly...

Or, do we rather to discard seq and frame mode?
"bins" and "constant" may be enough? Actually. I'm sicking of deciding `max_length` values. They looks  ugly in new recipes.


https://github.com/espnet/espnet/blob/4799542be4a0579752f953627d1cf8cd3b13668c/egs2/TEMPLATE/asr1/asr.sh#L763-L781

kamo-naoyuki: @ShigekiKarita @sw005320 Sounds odd?

- `const_no_sort` `ContantBatchSampler` => `unsorted` `UnsortedBatchSampler`
- `const` `ConstantSortedBatchSampler` => `sorted` `SortedBatchSampler`
- `seq` `SequenceBatchSampler` => `folded` `FoldedLengthBatchSampler` 
- `frame` `NA` => `len` `LengthBatchSampler`
-  `bins` `BatchBinSampler` => `numel` `NumElementsBatchSampler`

sw005320: All the method folds the original batchsize anyway.
How about distinguishing it with the **unit** to use folding?

kamo-naoyuki: @sw005320 Sorry, I'm not sure what you mean. Could you say more?

sw005320: all methods first define the unit (e.g., input/output length, number of elements of input and output), and use this unit as a threshold to define the batch size.
Probably I'm wrong.

kamo-naoyuki: seq mode is not matched this pattern?

```
If any feats lengths in the mini-batch are more than N * fold_length:
    => bs // N
```

sw005320: OK.
Then you can keep this name.

PR #1743, opened 2020-03-27
Fix --best_model_criterions

kamo-naoyuki:
- Change `action="append"` -> `nargs="+"`
   -  `action="append"`  appends the given arg to the default list
   -  `nargs="+"` remove the default. This is expected behavior
- Fix problem when best_model_criterions is given by config file:
   - The interval behavior of configargpase is that:
       1. Parse the value from yaml to Python object: e.g. [a, b] -> ['a', 'b']
       2. Change it to text: "['a', 'b']"
       3. Give it to type function which is given for add_arguments():  
           - e.g. ---add_arguments("--foo",type=int): --foo 3 -> return int("3")
       4. The value derived from the above step is the exact attribute in the parsed objects.

### DIFF
--- a/espnet2/tasks/abs_task.py
+++ b/espnet2/tasks/abs_task.py
@@ -398,7 +398,7 @@ class AbsTask(ABC):
         group.add_argument(
             "--best_model_criterion",
             type=str2triple_str,
-            action="append",
+            nargs="+",
             default=[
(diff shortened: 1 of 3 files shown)


kamo-naoyuki:
I'll merge this PR.

PR #1744, opened 2020-03-27
Not to create the averaged model if --keep_nbest_models=1

kamo-naoyuki:
The averaged model shouldn't be created  if --keep_nbest_models=1. Instead of it, create a symbolic link to the the best epoch model.


sw005320:
LGTM

sw005320:
Maybe better to add a unit test?

kamo-naoyuki:
Yeah,  but, to be honest I'm not sure whether the current way is better; Averaging the models in the training tool or doing it in the shell script, so I hesitate to create a unit test.

The current tool doesn't support any number of averaging, but the recipe looks simple.

Issue #1636, opened 2020-03-03
Too often failed ci at downloading from google drive


kamo-naoyuki:
We all may know, ci is often failed with the following part:

https://github.com/espnet/espnet/blob/b55cbcc50090e56808b1b2f3abc7f9327b1af27b/test/test_e2e_compatibility.py#L33-L57

e.g.

https://github.com/espnet/espnet/pull/1570/checks?check_run_id=482370468

Can't we solve this?

sw005320:
How about using https://github.com/wkentaro/gdown that @kan-bayashi introduced?

kamo-naoyuki:
@kan-bayashi What do you think if we remove this test? 

Possibly there seems to be some limitations to access google drive per day(?), so it's not appropriate to test in the main repo's ci, anyway.
How about creating another repository to manage our pre-trained model and test there?

kan-bayashi:
> @kan-bayashi What do you think if we remove this test? 

I agree. It is no problem.

PR #1772, opened 2020-04-06
Skip compatibility test

kan-bayashi:
This PR makes the compatibility test to be skipped since the download from google drive is often failed.

Related to #1636.

### DIFF
--- a/espnet/nets/pytorch_backend/e2e_asr_transducer.py
+++ b/espnet/nets/pytorch_backend/e2e_asr_transducer.py
@@ -305,6 +305,7 @@ class E2E(ASRInterface, torch.nn.Module):
 
         Returns:
             x (torch.Tensor): encoded features (T, attention_dim)
+
         """
         self.eval()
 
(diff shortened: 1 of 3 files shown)


kan-bayashi:
The flake8 warning may be caused by new version of `hacking`?

kamo-naoyuki:
Hmm... W503 was selected by me before. Why does it happen now?

https://github.com/espnet/espnet/blob/cb83e1b9d42ad6b13c6b195223884f647b076cd8/setup.cfg#L8-L11

```
#W503

# bad
income = (gross_wages
          + taxable_interest)
# good
income = (gross_wages +
          taxable_interest)
```

```
#W504

# bad
income = (gross_wages +
          taxable_interest)
# good
income = (gross_wages
          + taxable_interest)
```

(Note that there are no strong recommendations between W503 and W504, but black forces W504)

kan-bayashi:
Hmm. Let me check again.

PR #1782, opened 2020-04-07
missing argument for local/data_prep.sh added

beckgom:
in local/data_prep.sh

if [ $# != 3 ]; then
    echo "Usage: $0 <db> <data_dir> <trans_type>"
    exit 1
fi

char or phn should be added in the command line (seems like bug)


kan-bayashi:
Thank you. Could you make it an option as the same as `tts1/run.sh`?

kan-bayashi:
I will do it by myself.
Thanks!

beckgom:
Sorry. I missed your request.

Thanks~!

Issue #1780, opened 2020-04-07
Probably no dropout applied to the encoder in the RNN encoder/decoder ASR


qmeeus:
Hi,

When trying to apply dropout on the rnn encoder, I get this warning:
```
/lib/python3.6/site-packages/torch/nn/modules/rnn.py:50: UserWarning: dropout option adds dropout after all but last recurrent layer, so non-zero dropout expects num_layers greater than 1, but got dropout=0.5 and num_layers=1
  "num_layers={}".format(dropout, num_layers))
```
Indeed, [after a quick check in the code](https://github.com/espnet/espnet/blob/master/espnet/nets/pytorch_backend/rnn/encoders.py#L35):
```python
rnn = torch.nn.LSTM(inputdim, cdim, dropout=dropout, num_layers=1, bidirectional=bidir,
                                batch_first=True) if "lstm" in typ \
                else torch.nn.GRU(inputdim, cdim, dropout=dropout, num_layers=1, bidirectional=bidir, batch_first=True)
```

b-flo:
Hi,

Thanks for the report. We are aware of the problem: dropout won't be applied if `use_projection=True`, as RNNP is a stack of 1-layer torch.nn.LSTM and dropout is disabled if `num_layers=1`.

@sw005320 It was already reported in multiple discussions but I'm not sure why we didn't fix that? There may have been a reason.

qmeeus:
I wrote a fix, I will do a merge request after testing

sw005320:
> @sw005320 It was already reported in multiple discussions but I'm not sure why we didn't fix that? There may have been a reason.

It should be fixed...

> I wrote a fix, I will do a merge request after testing

That is really helpful.
It would be great if you make a PR about it.

qmeeus:
Here you go
https://github.com/espnet/espnet/pull/1784

qmeeus:
Before closing this, is it intentional to use the same dropout rate for the attention layer and the encoder? If not, I suggest adding a dropout rate specific to the attention layer and renaming the argument "dropout_rate" to "encoder_dropout" or something like that

sw005320:
I think this is a good idea in general. Personally, I did not have good success in introducing dropout anyway and making dropout related development items in a lower priority due to that. But I think this function is quite reasonable.

qmeeus:
My model overfits on the encoder side (CTC) but if I'm using dropout in the attention, the decoder is not able to attend to the right parts of the input anymore. I will look into it

PR #1784, opened 2020-04-07
fix https://github.com/espnet/espnet/issues/1780

qmeeus:
Summary:
- Remove dropout argument in RNN init
- Add dropout attribute to RNNP
- Call `F.dropout` in all but last layer
-  Some minor other changes that make the code more readable in my opinion

### DIFF
--- a/espnet/nets/pytorch_backend/rnn/encoders.py
+++ b/espnet/nets/pytorch_backend/rnn/encoders.py
@@ -78,7 +78,7 @@ class RNNP(torch.nn.Module):
             projection_layer = getattr(self, 'bt%d' % layer)
             projected = projection_layer(ys_pad.contiguous().view(-1, ys_pad.size(2)))
             xs_pad = projected.view(ys_pad.size(0), ys_pad.size(1), -1)
-            if layer < self.elayers - 1:              
+            if layer < self.elayers - 1:
                 xs_pad = torch.tanh(F.dropout(xs_pad, p=self.dropout))
 
         return xs_pad, ilens, elayer_states  # x: utt list of frame x dim
--- a/espnet/nets/pytorch_backend/rnn/encoders.py
+++ b/espnet/nets/pytorch_backend/rnn/encoders.py
@@ -32,21 +32,22 @@ class RNNP(torch.nn.Module):
                 inputdim = idim
             else:
                 inputdim = hdim
-            rnn = torch.nn.LSTM(inputdim, cdim, dropout=dropout, num_layers=1, bidirectional=bidir,
-                                batch_first=True) if "lstm" in typ \
-                else torch.nn.GRU(inputdim, cdim, dropout=dropout, num_layers=1, bidirectional=bidir, batch_first=True)
+
+            RNN = torch.nn.LSTM if "lstm" in typ else torch.nn.GRU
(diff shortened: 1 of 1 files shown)


espnet/nets/pytorch_backend/rnn/encoders.py
```diff
@@ -32,21 +32,22 @@ def __init__(self, idim, elayers, cdim, hdim, subsample, dropout, typ="blstm"):
                 inputdim = idim
             else:
                 inputdim = hdim
-            rnn = torch.nn.LSTM(inputdim, cdim, dropout=dropout, num_layers=1, bidirectional=bidir,
-                                batch_first=True) if "lstm" in typ \
-                else torch.nn.GRU(inputdim, cdim, dropout=dropout, num_layers=1, bidirectional=bidir, batch_first=True)
+
+            RNN = torch.nn.LSTM if "lstm" in typ else torch.nn.GRU
+            rnn = RNN(inputdim, cdim, num_layers=1, bidirectional=bidir, batch_first=True)
+            num_directions = 1 + bidir
```
sw005320, line 38: This is right but the https://github.com/espnet/espnet/pull/1784/files#diff-12afbdd087fdf0d238a6a6abe4b582e0L40-L43 is more readable.
Maybe we can keep this part as it is?

qmeeus: I restored the lines 40-43 to their original form

sw005320:
Thanks!

boeddeker:
In the new code you use `F.dropout(xs_pad, p=self.dropout)`, but in pytorch the function is
```
def dropout(input: Tensor, p: float = 0.5, training: bool = True, inplace: bool = False) -> Tensor:
    ...
```
So the signature contains a `training` flag, to indicate, it dropout should be used or not.
Could you either make it deterministic for evaluation or add a small comment to the source code, why it is choosen to be non-deterministic?

sw005320:
Thanks. It can be a bug.
There is a discussion in https://github.com/espnet/espnet/pull/3539

@pengchengguo, I still found it in 
- https://github.com/espnet/espnet/blob/master/espnet/nets/pytorch_backend/rnn/encoders.py#L90
- https://github.com/espnet/espnet/blob/master/espnet2/asr/ctc.py#L142

Can you deal with it?

@kan-bayashi,
You told me that it is intentional, then, I suggest you add a comment about it.
- https://github.com/espnet/espnet/blob/master/espnet/nets/pytorch_backend/tacotron2/decoder.py#L146

kan-bayashi:
Please read note
https://github.com/espnet/espnet/blob/40202fca64529982fa7c1c428137243f306a0d5d/espnet/nets/pytorch_backend/tacotron2/decoder.py#L106

sw005320:
Yes, but I thought that it is better to add the comment in the actual code.

sw005320:
@kan-bayashi, see #4322 
I think this is safe.

boeddeker:
Ah, ok. I wrote in this PR, because before this PR the dropout was disabled at eval time.

I didn't know, that some NNs keep the dropout at eval time. In my mind, this should at least be an option for core modules (like RNNP), so the user can decide if he/she wants to follow the classical dropout idea or the always active dropout idea.

For me, I copied now the RNNP code with a hint to ESPnet and disabled the dropout at eval time.

PR #1785, opened 2020-04-07
Add note in synth_wav.sh

kan-bayashi:
I added note for `synth_wav.sh` to make it clear the usage.


unilight:
Is it possible to generalize the current script to support synthesizing multiple utterances at one?
For example, generate an utterance for each line in the input file.

kan-bayashi:
It can be. I think not so difficult.

kan-bayashi:
@unilight I supported such options. See #1788.

PR #1800, opened 2020-04-10
Data preparation for TTS on Babel 

lwang114:
Generate the wav.scp and text file with segmented utterances for training the Tacotron TTS system. To run the scripts, you may need to change the path to the dataset to your own. The VAD is disabled by default.


sw005320:
Thanks!
Could you first add a brief explanation about the corpus/task in https://github.com/espnet/espnet/blob/master/egs/README.md?
It seems that this PR combines `egs/babel/tts1` and `egs/discophone/asr1`, and I'd like to know the relationship between them (also, it might be better to make a two PRs instead of making one big PR).

PR #1815, opened 2020-04-12
add badge

kamo-naoyuki:


kamo-naoyuki:
@sw005320 or @kan-bayashi 
Could you change the setting of protection branch? Maybe we need add mergify to “Restrict who can push to matching branches” list.

kan-bayashi:
Updated.

PR #1819, opened 2020-04-12
Added some TIMIT results

sknadig:
These are some results from a grid search experiment I did. I can add the final PER, CER of all the experiments if that is useful.

@sw005320 @kamo-naoyuki


kamo-naoyuki:
@sknadig Thank you! This results are very useful for me.
Could you add the configuration files? It's enough to put only the best config.

```
conf/train.yaml -> conf/tuning/train2.yaml   # symlink
conf/tuning/train.yaml    # The existing file
conf/tuning/train2.yaml  # The best
``

sknadig:
@kamo-naoyuki I will add the best config for PER and CER from my grid search experiment. Although I did not explore pure CTC or pure Attention - all of these experiments are with `mtlalpha: 0.5`.

sw005320:
> Although I did not explore pure CTC or pure Attention - all of these experiments are with `mtlalpha: 0.5`.

This is no problem.

sknadig:
@kamo-naoyuki The best config for both PER and CER turns out to be same.

kamo-naoyuki:
Thanks! I'll merge

PR #1840, opened 2020-04-15
Add test for torch>1.1

### LINKED ISSUE #1832: transformer bug in Pytorch 1.2.0
espnet/espnet/nets/pytorch_backend/transformer/mask.py 
`is_torch_1_2_plus = LooseVersion(torch.__version__) >= LooseVersion('1.2.0')
datatype = torch.bool if is_torch_1_2_plus else torch.uint8

def subsequent_mask(size, device="cpu", dtype=datatype):
    """Create mask for subsequent steps (1, size, size).
    :param int size: size of mask
    :param str device: "cpu" or "cuda" or torch.Tensor.device
    :param torch.dtype dtype: result dtype
    :rtype: torch.Tensor
    >>> subsequent_mask(3)
    [[1, 0, 0],
     [1, 1, 0],
     [1, 1, 1]]
    """
    ret = torch.ones(size, size, device=device, dtype=dtype)
    return torch.tril(ret, out=ret)
`
RuntimeError: "tril" not implemented for 'Bool'

pytorch 1.2.0 with GPU or CPU

kamo-naoyuki:
- Add test for torch>1.1
- Now the tested items are duplicated, so I changed the range
    - https://github.com/espnet/espnet/blob/34b21ca720d698de72d20b1d601ab4671d7badac/README.md
    - Travis-ci: only builds the documentation and codecov without testing with python=3.8.
    - Circleci: Testing of different OS with `conda`, Python=3.6, Pytorch=1.4
    - Github-actions: Testing of different PyTorch version with `pip`, ubuntu18, Python=3.7 or 3.8
- I removed the test for nightly. Maybe we don't need it?
- Fix #1832

@ShigekiKarita Would you check the CI error of `test/test_recog.py`?  I couldn't fix this.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -19,7 +19,7 @@ jobs:
       - image: circleci/buildpack-deps:18.04
         environment:
           ESPNET_PYTHON_VERSION: 3.6
-          TH_VERSION: 1.1.0
+          TH_VERSION: 1.4.0
           CHAINER_VERSION: 5.0.0
           USE_CONDA: true
           CC: gcc-7
@@ -69,7 +69,7 @@ jobs:
       - image: circleci/buildpack-deps:16.04
         environment:
           ESPNET_PYTHON_VERSION: 3.6
-          TH_VERSION: 1.1.0
+          TH_VERSION: 1.4.0
           CHAINER_VERSION: 5.0.0
           USE_CONDA: true
           CC: gcc-7
@@ -122,7 +122,7 @@ jobs:
       - image: centos:7
         environment:
           ESPNET_PYTHON_VERSION: 3.6
-          TH_VERSION: 1.1.0
+          TH_VERSION: 1.4.0
           CHAINER_VERSION: 5.0.0
           USE_CONDA: true
           CC: /opt/rh/devtoolset-7/root/usr/bin/gcc
@@ -180,7 +180,7 @@ jobs:
       - image: debian:9
         environment:
           ESPNET_PYTHON_VERSION: 3.6
-          TH_VERSION: 1.1.0
+          TH_VERSION: 1.4.0
           CHAINER_VERSION: 5.0.0
           USE_CONDA: true
           CC: gcc-6
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -18,13 +18,14 @@ jobs:
     docker:
       - image: circleci/buildpack-deps:18.04
         environment:
-          # NOTE(kan-bayashi): Use 3.7.3 to avoid sentencepiece installation error
-          ESPNET_PYTHON_VERSION: 3.7.3
+          ESPNET_PYTHON_VERSION: 3.6
           TH_VERSION: 1.1.0
           CHAINER_VERSION: 5.0.0
           USE_CONDA: true
           CC: gcc-7
           CXX: g++-7
+          # To avoid UnicodeEncodeError for python<=3.6
+          LC_ALL: en_US.UTF-8
     working_directory: ~/repo
 
     steps:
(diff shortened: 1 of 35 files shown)


kan-bayashi:
> test/test_recog.py

I asked Shigeki before. This test is to check the randomness of the initialization.
Maybe the initialization of some parameters is changed from 1.1.
I'm not sure we should keep this test as is.

ShigekiKarita:
You can skip this. It is impossible to figure out the reason.

.travis.yml
```diff
@@ -21,24 +21,24 @@ addons:
 
 env:
   # NOTE(kan-bayashi): Use 3.7.3 to avoid sentencepiece installation error
```
kan-bayashi, line 23: Not related to this PR but this already fixed so it is better to remove.
```suggestion
```

.circleci/config.yml
```diff
@@ -18,13 +18,14 @@ jobs:
     docker:
       - image: circleci/buildpack-deps:18.04
         environment:
-          # NOTE(kan-bayashi): Use 3.7.3 to avoid sentencepiece installation error
-          ESPNET_PYTHON_VERSION: 3.7.3
+          ESPNET_PYTHON_VERSION: 3.6
           TH_VERSION: 1.1.0
           CHAINER_VERSION: 5.0.0
```
kan-bayashi, line 23: In `setup.py` we use chainer==6.0.0.
Maybe it is better to follow it.
```suggestion
          CHAINER_VERSION: 6.0.0
```

.circleci/config.yml
```diff
@@ -66,13 +68,14 @@ jobs:
     docker:
       - image: circleci/buildpack-deps:16.04
         environment:
-          # NOTE(kan-bayashi): Use 3.7.3 to avoid sentencepiece installation error
-          ESPNET_PYTHON_VERSION: 3.7.3
+          ESPNET_PYTHON_VERSION: 3.6
           TH_VERSION: 1.1.0
           CHAINER_VERSION: 5.0.0
```
kan-bayashi, line 73: ditto
```suggestion
          CHAINER_VERSION: 6.0.0
```

.circleci/config.yml
```diff
@@ -117,13 +121,14 @@ jobs:
     docker:
       - image: centos:7
         environment:
-          # NOTE(kan-bayashi): Use 3.7.3 to avoid sentencepiece installation error
-          ESPNET_PYTHON_VERSION: 3.7.3
+          ESPNET_PYTHON_VERSION: 3.6
           TH_VERSION: 1.1.0
           CHAINER_VERSION: 5.0.0
```
kan-bayashi, line 126: ditto
```suggestion
          CHAINER_VERSION: 6.0.0
```

.circleci/config.yml
```diff
@@ -173,13 +179,14 @@ jobs:
     docker:
       - image: debian:9
         environment:
-          # NOTE(kan-bayashi): Use 3.7.3 to avoid sentencepiece installation error
-          ESPNET_PYTHON_VERSION: 3.7.3
+          ESPNET_PYTHON_VERSION: 3.6
           TH_VERSION: 1.1.0
           CHAINER_VERSION: 5.0.0
```
kan-bayashi, line 184: ditto
```suggestion
          CHAINER_VERSION: 6.0.0
```

sw005320:
> Maybe it is time to update the default pytorch version in Makefile.

I agree.

kamo-naoyuki:
NOTE: About LC_ALL

The locale affects the default encoding in Python<=3.6, and the only way to change the encoding is giving encoding argument to open: `open(encoding="utf-8")`.
`sentencepiece` and `jsondiff` doesn't use open with encoding explicitly and there are no solutions except for changing the locale.
The locale is not utf-8 in circle-ci, so I changed the locale.

Note that our units tests are failed if LC_ALL=C and python=3.6. 
In integration test, we set LC_ALL=C in path.sh, but mini_an4 doesn't include non-ascii chars, thus this integration test succeeds.

kamo-naoyuki:
@ShigekiKarita I'd like you to check too.

kamo-naoyuki:
@ShigekiKarita Thanks

Issue #1839, opened 2020-04-15
model link for transformer in espnet/egs/tedlium3/asr1 not working


ORlGlN:
The model link with the following details returned 404 on Google Drive, anyone able to fix the link?
file path: espnet/egs/tedlium3/asr1/RESULTS.md
model name: Transformer (elayers=12, dlayers=6, units=2048, 8 GPUs, specaug) + large LM
model link: https://drive.google.com/open?id=1wYYTwgvbB7uy6agHywhQfnuVWWW_obmO

Thanks in advance.

sw005320:
This is the same problem as https://github.com/espnet/espnet/issues/1590.
@ahmedalbahnasawy, do you also have a tedlium3 model?

ahmedalbahnasawy:
@sw005320 done 
I uploaded it to your drive.

sw005320:
Many thanks!

PR #1841, opened 2020-04-15
fix the missing link of tedlium3

sw005320:
This is related to #1839 

@ahmedalbahnasawy, could you check that it's working?


ahmedalbahnasawy:
The link is not working. I checked the file locally and nothing is missing.

sw005320:
> The link is not working. I checked the file locally and nothing is missing.

Probably my setting is wrong?
I checked it with the other account and I could download it.

image

PR #1857, opened 2020-04-20
remove files to reduce disk usage in integrated test

kamo-naoyuki:


sw005320:
Probably it's better to add some comments of removing these files in order to reduce disk usage in an integrated test

PR #1868, opened 2020-04-24
HDF5 corpus: Out-of-core data loading

kamo-naoyuki:
### Motivation
In ESPNet1, we dump scp files into a JSON file and  I removed this json format for espnet2 because the JSON structure has redundant information and it causes a huge amount of memory footprint.

Instead of JSON, espnet2 can directly accept scp files, but it's still not scalable because it is a just text file and if they are large files themselves, then causes memory error. #1801

### About this PR
I decided to use HDF5 format to describe a corpus. I referred it as `HDF5 corpus` in the comment.

- HDF5 is lazyio format, i.e. the data loading is done when accessing data, so we can easily implement out-of-core loading.
- This format can be generated by combining multiple scp files:
    ```bash
   python3 -m espnet2.bin.gen_hdf5_corpus \
              --data_path_and_name_and_type "wav.scp,speech,sound" \
              --data_path_and_name_and_type "train/text,text,text" \
              --shape_file "train/speech_shape" \
              --shape_file "train/text_shape" \
              --out "train/corpus.h5"
    ```
- This format is just a replacement of the JSON file of epsnet1. It has text information only and  the value indicates the reference path to wav files, ark file, npy, just like scp files and **not having the real array data** in each entry.
- The structure of HDF5 corpus:
    ```
    speech/type="sound"
    speech/data/
       sample_id1="some/path/a.wav"
       sample_id2="some/path/b.wav"
       ...
    text/type="direct"
    text/data/
       sample_id1="hello world"
       sample_id2="foo bar"
       ...
    shape_files/0
       sample_id1=(18000,)
       sample_id2=(13900,)
       ...
    shape_files/1
       sample_id1=(2,)
       sample_id2=(2,)
       ...
    ```
- Scp files inputting is also still supported for training tool. i.e. There are two methods to describe corpus.
    ```
     # The following two are equivalent
     python -m espnet2.bin.asr_train  \
         --train_data_path_and_name_and_type ... \
         --train_shape_file ... \
         --train_data_path_and_name_and_type ... \
         --train_shape_file ...
    python -m espnet2.bin.asr_train --train_hdf5_corpus corpus.h5
    ```

### Remained limitation

This PR removed the memory error caused by `Dataset`, but I couldn't avoid keeping sample ids for mini-batching in memory for `BatchSampler`.

```
batches = [["sample_id1", "sample_id2", ...], ["sample_id3", "sample_id4", ...], ...]]
```

If too many samples to keep in memory, a memory error happens. To avoid this problem, we need to dump shuffled ids to an external file. If some guys complain about it, I'll implement it.

### DIFF
--- a/ci/test_integration.sh
+++ b/ci/test_integration.sh
@@ -61,9 +61,12 @@ for t in ${token_types}; do
 done
 for t in ${feats_types}; do
     for t2 in ${token_types}; do
-        echo "==== feats_type=${t}, token_types=${t2} ==="
-        ./run.sh --ngpu 0 --stage 6 --stop-stage 100 --feats-type "${t}" --token-type "${t2}" \
-            --asr-args "--max_epoch=1" --lm-args "--max_epoch=1"
+        for use_hdf5_corpus in true false; do
+            echo "==== feats_type=${t}, token_types=${t2}, use_hdf5_corpus=${use_hdf5_corpus} ==="
+            rm -rf exp
+            ./run.sh --ngpu 0 --stage 6 --stop-stage 100 --feats-type "${t}" --token-type "${t2}" \
+                --asr-args "--max_epoch=1" --lm-args "--max_epoch=1" --use_hdf5_corpus "${use_hdf5_corpus}"
+        done
     done
 done
 # Remove generated files in order to reduce the disk usage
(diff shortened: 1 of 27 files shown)


sw005320:
Very cool.
Did you test the scalability of HDF5 in this implementation (it is not easy but we could just duplicate a lot of data and test it anyway)?

@kan-bayashi, you have several experiences of handling hdf5, right?
What is your opinion about it?

kamo-naoyuki:
> Did you test the scalability of HDF5 in this implementation (it is not easy but we could just duplicate a lot of data and test it anyway)?

I met a problem now, I'll rethink.

kamo-naoyuki:
1. About the limitation HDF5

Writing large amount of number of datasets in HDF5 is really slow.
For small number, the speed is 5000iter/second, but after ~600,000 keys are registered, the speed is decreased to ~10iter/second. 

https://github.com/h5py/h5py/issues/1055

https://stackoverflow.com/questions/40071425/writing-large-amounts-of-numbers-to-a-hdf5-file-in-python

The number of keys in a `group` causes this problem. I mean, if we split keys into sub groups, the speed degradation doesn't happen.

```python
f = h5py.File("out.h5")

# Slow
for i in range(1000_000_000):
     f[str(i)] = i

# Fast
for i in range(1000_000_000):
     if i % 600_000 == 0:
          g = f.create_group(f"group{i // 600_000}")
     g[str(i)] = i
```

2. My thought

We always sort all samples by the length. I think sorting and shuffling for all samples is not necessary for large dataset. Sorting and shuffling also take times and memory.

We may need some system to train with multiple datasets splited by a large dataset. Each datasets are independently sorted and shuffled and expanded in memory one by one, then we can handle a large dataset. 

An easy way to implement such system is distributed training. Current distributed training implementation requires to load same scp files for each process and split keys after loading, but this implementation is not efficient. Originally it can split at the file level before training and we can give the splited files to each training process.

I'll think for a while.

PR #1881, opened 2020-04-28
Add test for pytorch1.5

kamo-naoyuki:
- Add test for pytorch1.5
- Modified Makefile and create tools/install_*.sh
    - Makefile is unreadable, so I separated each installation steps to each shell script.
- Use Makefile in ci to validate Makefile, too.
    - Currently Makefile is not tested at all.


sw005320:
Thanks.
This looks good for me.

PR #1906, opened 2020-05-06
fix reverse option for cmvn

magictron:
The reverse operation order needs to be divide -> subtract (rather than subtract -> divide) w.r.t normalization operation order add + multiply

### DIFF
--- a/espnet/transform/cmvn.py
+++ b/espnet/transform/cmvn.py
@@ -106,11 +106,11 @@ class CMVN(object):
             if self.norm_vars:
                 x = np.multiply(x, self.scale[spk])
 
-        else:
-            if self.norm_means:
-                x = np.subtract(x, self.bias[spk])
+        else:            
             if self.norm_vars:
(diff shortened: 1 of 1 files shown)


kamo-naoyuki:
Oh, thanks.

kamo-naoyuki:
Would you fix the CI error?

magictron:
> Would you fix the CI error?

sorry that I am not that familiar with the process, I just edited the file from the web. Can you take it over?

Thanks

espnet/transform/cmvn.py
```diff
@@ -106,11 +106,11 @@ def __call__(self, x, uttid=None):
             if self.norm_vars:
                 x = np.multiply(x, self.scale[spk])
 
-        else:
-            if self.norm_means:
-                x = np.subtract(x, self.bias[spk])
+        else:
```
kamo-naoyuki, line 109: Please delete the extra spaces.
```suggestion
        else:
```

PR #1918, opened 2020-05-12
FIX E523,E541,E741

kamo-naoyuki:
I fixed new errors of flake8

### DIFF
--- a/ci/install_kaldi.sh
+++ b/ci/install_kaldi.sh
@@ -13,6 +13,6 @@ set -euo pipefail
 
 # download pre-built kaldi binary
 # TODO(karita) support non ubuntu env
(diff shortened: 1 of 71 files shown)


kamo-naoyuki:
@ShigekiKarita https://18-198329952-gh.circle-artifacts.com/0/home/circleci/repo/ubuntu16-featbin.tar.gz is not found?

ShigekiKarita:
Oh ... it looks expired. Maybe we can use this one instead https://github.com/espnet/kaldi-bin/releases

ShigekiKarita:
Hmm, the CI error looks related to
> Meaning, if you downloaded this from a web server, sometimes Gzip compression get’s applied to web content on the server level to speed up content download. But if not properly set on the server to ignore already compressed content such as this, it can inadvertently double-Gzip files.
https://superuser.com/questions/841865/extracting-a-tar-gz-file-returns-this-does-not-look-like-a-tar-archive

I could decompress the file by  `gzip -dc ./ubuntu16-featbin.tar.gz  | tar -xvzf -`

PR #1922, opened 2020-05-13
Master to develop

kamo-naoyuki:


kamo-naoyuki:
@sw005320 @kan-bayashi @ShigekiKarita  Does anyone know this error?

ShigekiKarita:
It looks like sentencepiece breaking change:
> not ok 30 spm_xxx
> (in test file test_utils/test_spm.bats, line 25)
>   `diff ${bpemodel}.vocab test/tedlium2.vocab' failed

ShigekiKarita:
3 hours ago, sp released 1.90 https://github.com/google/sentencepiece/releases
Maybe we should specify the older one in `setup.py`.

kamo-naoyuki:
I'm not sure whether we should specify the version, or rather to fix the test. I don't know the detail.
Please check the update.

ShigekiKarita:
I cannot understand why it changed. However, according to the diff, at
least the vocab looks different ([tag] is included?). For reproducibility,
we should not update this.

And spm seems to have no deps. It won't be a blocker in the future.

2020年5月13日(水) 17:56 Naoyuki Kamo <notifications@github.com>:

PR #1897, opened 2020-05-01
Finetuning doc + freezing parameters option

b-flo:
Hi,

This PR add freezing parameters option + documentation for finetuning options (transfer learning + freezing). Feel free to tell me or modify docs if it lacks clarity!

PR #1934, opened 2020-05-15
fix freezing modules

atozto9:
`state_dict()` is not suitable for freezing weights.

### DIFF
--- a/espnet/tts/pytorch_backend/tts.py
+++ b/espnet/tts/pytorch_backend/tts.py
@@ -329,15 +329,19 @@ def train(args):
 
     # freeze modules, if specified
     if args.freeze_mods:
-        for mod, param in model.state_dict().items():
+        for mod, param in model.named_parameters():
             if any(mod.startswith(key) for key in args.freeze_mods):
                 logging.info(f"{mod} is frozen not to be updated.")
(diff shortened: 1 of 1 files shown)


b-flo:
Hi,

Thanks for the PR!

@kan-bayashi Sorry, I forgot to notify you about the problem in the TTS part when I made #1897 . Also we should add something like:

```
model_params = filter(lambda x: x.requires_grad, model.parameters())
```

and pass it to optimizer to avoid update when weights value are 0 in some pytorch version.

kan-bayashi:
Sorry for late reply.
Could you also apply @b-flo suggestions?

atozto9:
Hi, @kan-bayashi 
I applied @b-flo 's suggestion.

Could you check it if it's applied properly?

kan-bayashi:
@b-flo We should fix the following part as well.
https://github.com/espnet/espnet/blob/39731696531dedfe2cc6ce3178d4572f56c6d503/espnet/nets/pytorch_backend/transformer/optimizer.py#L74

b-flo:
Oh yes, I also forgot to change it in #1897, I can add make modifications in that PR as it's not merged yet. Let me finish some things and I'll do that!

PR #1952, opened 2020-05-22
Fix bug of espnet2 asr_inference.py

kamo-naoyuki:
I found a fatal bug.

The model is not changed to eval mode when inference. BeamSearch.eval() is invoked, so the decoder part is set on eval mode, but the encoder part is still train mode, i.e. dropout is used for the encoder part.

### DIFF
--- a/espnet2/bin/asr_inference.py
+++ b/espnet2/bin/asr_inference.py
@@ -81,6 +81,7 @@ def inference(
     asr_model, asr_train_args = ASRTask.build_model_from_file(
         asr_train_config, asr_model_file, device
     )
+    asr_model.eval()
 
     decoder = asr_model.decoder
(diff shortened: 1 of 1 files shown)


sw005320:
Oh, this is super critical...
But this means espnet2 numbers would be improved more!
Lucky to find such a serious bug in this stage!

kamo-naoyuki:
Yeah, really sorry...

sw005320:
Maybe, we should revisit the espnet2 ASR recipe with this fix (and other improvements)? We can help it.

kamo-naoyuki:
Yes, it's better to do about the transformer result.
(Unfortunately I don't have any resources for training due to personal reason for now, so I can't help)

I also added some new features e.g. specaug and numel mode (batch-bin mode) from that time, so we can also use them for the second round.

PR #1972, opened 2020-05-25
Solve memory issue with super large corpus training

kamo-naoyuki:
It's hard to handle large coprus, such as over 10,000,000 samples now because the number of samples in scp files directly impacts the memory footprint.

I prepared two ways to handle this problem.

1. IterableDataset

This class is not used training, but used for calc_stats and inference mode.
Current `ESPnetDataset` is dict-type:

```python
data = dataset["uttid"]
```

I newly implemented iterable type dataset: `IterableESPnetDataset`

```python
for data in dataset:
    ...
```

Text files are read line-by-line, so we no longer consume memory.
Note that iterable dataset for pytorch DataLoader is supported from pytorch=1.2, so if pytorch<=1.1, it falls back to the conventional ESPnetDataset.

2. Multiple iterator

IterableDataset can't be used for training because it can't shuffle the samples order.
I think it's one of the reasonable choices for some task, but for seq2seq task, it's not acceptable.

It's hard to handle large corpus as it is with finite memory size. We need some out-of-core techniques both for Dataset class and Sampler class for it. 
Instead of using such hacking, I rather select splitting the corpus and build iterators from each subset.

- At first, split scp files:
    ```sh
    % python -m espnet2.bin.split_scps \
      --scps wav.scp text speech_shape text_shape \
      --num_splits 100 \
      --output_dir out
    # out/wav.scp/split.0 out/wav.scp/split.1
    # out/text/split.0 out/text/split.1
    # out/speech_shape/split.0 out/speech_shape/split.1
    # out/text_shape/split.0 out/text_shape/split.1
    ```
- it can be input as same way
    ```sh
    % python -m espnet2.bin.asr_train \
      --train_data_path_and_name_and_type out/wav.scp,speech,sound \
      --train_data_path_and_name_and_type out/text,text,text \
      --train_shape_file out/speech_shape \
      --train_shape_file out/speech_shape
    ```
- Iterators are built using each subsets. They are destroyed after used as follows:
    ```python
    # Pseudo code
    for epoch in range(max_epoch):
        for n in numpy.random.permutations(num_splits):
            iter_factory = build_iter_factory(
                f"out/wav.scp/split.{n}",
                f"out/text/split.{n}",
                f"out/speech_shape/split.{n}",
                f"out/speech_shape/split.{n}", ...
            )
           for batch in iter_factory.build_iter(epoch):
                 train(batch)
    ```

Theoretically we can handle any size of corpus now. Actually I could start to train with 84,170,000  samples.

### DIFF
--- a/egs2/TEMPLATE/asr1/asr.sh
+++ b/egs2/TEMPLATE/asr1/asr.sh
@@ -62,6 +62,7 @@ lm_config=        # Config for language model training.
 lm_args=          # Arguments for language model training, e.g., "--max_epoch 10".
                   # Note that it will overwrite args in lm config.
 use_word_lm=false # Whether to use word language model.
+num_splits_lm=1   # Number of splitting for lm corpus
 # shellcheck disable=SC2034
 word_vocab_size=10000 # Size of word vocabulary.
 
@@ -71,6 +72,7 @@ asr_config= # Config for asr model training.
 asr_args=   # Arguments for asr model training, e.g., "--max_epoch 10".
             # Note that it will overwrite args in asr config.
 feats_normalize=global_mvn  # Normalizaton layer type
+num_splits_asr=1   # Number of splitting for lm corpus
 
 # Decoding related
 decode_tag=    # Suffix to the result dir for decoding.
@@ -140,6 +142,7 @@ Options:
                       # Note that it will overwrite args in lm config.
     --use_word_lm     # Whether to use word language model (default="${use_word_lm}").
     --word_vocab_size # Size of word vocabulary (default="${word_vocab_size}").
+    --num_splits_lm=1   # Number of splitting for lm corpus (default="${num_splits_lm}").
 
     # ASR model related
     --asr_tag    # Suffix to the result dir for asr model training (default="${asr_tag}").
@@ -147,6 +150,7 @@ Options:
     --asr_args   # Arguments for asr model training, e.g., "--max_epoch 10" (default="${asr_args}").
                  # Note that it will overwrite args in asr config.
     --feats_normalize # Normalizaton layer type (default="${feats_normalize}").
+    --num_splits_asr=1   # Number of splitting for lm corpus  (default="${num_splits_asr}").
 
     # Decoding related
     --decode_tag       # Suffix to the result dir for decoding (default="${decode_tag}").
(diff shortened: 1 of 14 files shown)


kamo-naoyuki:
I'll merge this PR.

sw005320:
OK.

Did you confirm that it's working with huge lines of files?

kamo-naoyuki:
Yes, it works with 84,170,000 samples.

Cescfangs:
I also tried about 20,000,000 utt dataset, it  worked well

RobotGF:
When training with Multiple iterator, it also couldn't shuffle across different split scp？

kamo-naoyuki:
The samples are shuffle within the each scp and the order of multiple iterators is randomly selected for each epoch.

RobotGF:
> The samples are shuffle within the each scp and the order of multiple iterators is randomly selected for each epoch.

Thanks, Got it

kamo-naoyuki:
Actually, I'm not sure whether this strategy is good or not. Please notify me if any problems happen.

RobotGF:
> Actually, I'm not sure whether this strategy is good or not. Please notify me if any problems happen.

Tensorflow is also use this strategy. I'll show my experiments results later.

kamo-naoyuki:
Oh, what tool?

Cescfangs:
I have quetions using split scps when set `num_iters_per_epoch`,  I split scp into 4 files, each have about 35000 batches:
```
ESPnetDataset(
  speech: {"path": "exp/asr_stats/splits4/feats.scp/split.1", "type": "kaldi_ark"}
  text: {"path": "exp/asr_stats/splits4/text/split.1", "type": "text"}
  preprocess: <espnet2.train.preprocessor.CommonPreprocessor object at 0x7f06f4827128>)
[transformer-specaug-2w2-stream-rjy94de2k:0/8] 2020-06-23 11:21:29,845 (abs_task:1354) INFO: [train] Batch sampler: FoldedBatchSampler(N-batch=35427, batch_size=140, shape_files=['exp/asr_stats/splits4/speech_shape/split.1', 'exp/asr_stats/splits4/text_shape/split.1'], sort_in_batch=descending, sort_batch=descending)
[transformer-specaug-2w2-stream-rjy94de2k:0/8] 2020-06-23 11:21:29,851 (abs_task:1356) INFO: [train] mini-batch sizes summary: N-batch=35427, mean=134.4, min=31, max=140
```

I set `num_iters_per_epoch=50000` and every subset ended after 12500 batches, which sums to 50000, I guess it was OK if samples are shuffled within the each scp in each epoch, is that supposed to work like that?

kamo-naoyuki:
Yes, I intended it. What is problem?

Cescfangs:
no problem, just want to be sure it was expected to work that way

PR #1983, opened 2020-05-29
Fix freezing modules (when using  multi-gpu)

atozto9:
Hi, @kan-bayashi, @b-flo 

After we've wrapped model in `nn.DataParallel`, the name of the parameter is prepended with '`module.`'

Thanks,

### DIFF
--- a/espnet/tts/pytorch_backend/tts.py
+++ b/espnet/tts/pytorch_backend/tts.py
@@ -330,10 +330,12 @@ def train(args):
     # freeze modules, if specified
     if args.freeze_mods:
         if hasattr(model, "module"):
-            prefixed_freeze_mods = ["module." + x for x in args.freeze_mods]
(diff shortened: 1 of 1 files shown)


espnet/tts/pytorch_backend/tts.py
```diff
@@ -329,8 +329,11 @@ def train(args):
 
     # freeze modules, if specified
     if args.freeze_mods:
+        if hasattr(model, "module"):
+            prefixed_freeze_mods = ["module." + x for x in args.freeze_mods]
```
kan-bayashi, line 333: We must handle the single gpu case as well.

atozto9: Hi, I reflected your comment
thx.

kan-bayashi:
Thanks a lot! @atozto9

PR #1984, opened 2020-05-29
Implicitly ignore check for `warp_ctc`

Syzygianinfern0:
Ignore redundant check for `warp_ctc` implicitly. Doesn't terminate the check negatively if `--torch-version` is not passed but is greater than `1.2.0`

### DIFF
--- a/tools/check_install.py
+++ b/tools/check_install.py
@@ -61,7 +61,7 @@ def main(args):
             logging.info("--> %s is installed." % name)
             is_correct_installed_list.append(True)
 
-            if name=="torch":
+            if name == "torch":
                 if LooseVersion(importlib.import_module(name).__version__) > LooseVersion('1.2.0'):
                     try:
                         library_list.remove(('warpctc_pytorch', ("0.1.1", "0.1.3")))
--- a/tools/check_install.py
+++ b/tools/check_install.py
@@ -49,8 +49,7 @@ def main(args):
 
     logging.info("python version = " + sys.version)
 
-    library_list = []
-    library_list.extend(MANUALLY_INSTALLED_LIBRARIES)
+    library_list = MANUALLY_INSTALLED_LIBRARIES.copy()
 
     # check library availableness
     logging.info("library availableness check start.")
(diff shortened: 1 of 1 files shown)


tools/check_install.py
```diff
@@ -61,6 +60,16 @@ def main(args):
             importlib.import_module(name)
             logging.info("--> %s is installed." % name)
             is_correct_installed_list.append(True)
+
+            if name=="torch":
```
kan-bayashi, line 64:
```suggestion
            if name == "torch":
```

kan-bayashi:
I think this is already included. 
https://github.com/Syzygianinfern0/espnet/blob/d18907bc1cb468cf3315d262c6e6ce9fe665fbb8/tools/check_install.py#L43-L44

Syzygianinfern0:
> I think this is already included.

This is when `--torch-version` is not passed as an arg. 
> Doesn't terminate the check negatively if --torch-version is not passed but is greater than 1.2.0

During my installation, I had a tough time trying to understand why warp_ctc was to be installed. It is not mentioned anywhere in the installation docs that it doesn't need to be if torch.version > 1.2.0. Nor is it made clear that `--torch-version` has to be explicitly passed. 

Also trying to install warp_ctc with a higher torch version leads to problems in the python environment.

kan-bayashi:
In Makefile, we check the pytorch version and decide whether to install warp_ctc.
https://github.com/espnet/espnet/blob/005fcc00ec9a7c71eda8c3a9a2c1802236f1e8e7/tools/Makefile#L10-L15
And we pass the `--torch-version` in Makefile.
https://github.com/espnet/espnet/blob/005fcc00ec9a7c71eda8c3a9a2c1802236f1e8e7/tools/Makefile#L162

In your experiment, why did you meet the above problem?

kan-bayashi:
Why don't you remove the default value of `--torch-version` and set to `required=True`? I think this can solve your problem much simpler.

Syzygianinfern0:
> In your experiment, why did you meet the above problem?

I had not installed from the MakeFile for 1 particular reason. I already had most of those dependencies installed on my local machine.

Syzygianinfern0:
> Why don't you remove the default value of `--torch-version` and set to `required=True`? I think this can solve your problem much simpler.

If the torch version can be identified by the script itself, it would be more comfortable to automatically do so. Just my opinion though.

Syzygianinfern0:
> > Why don't you remove the default value of `--torch-version` and set to `required=True`? I think this can solve your problem much simpler.
> 
> If the torch version can be identified by the script itself, it would be more comfortable to automatically do so. Just my opinion though.

@kan-bayashi What do you feel?

kan-bayashi:
I reflected your comments in #1991. 
Could you check it?

Syzygianinfern0:
> Could you check it?

Sure

kan-bayashi:
Fixed #1991

kan-bayashi:
Thank you for your recommendations, @Syzygianinfern0.

PR #1991, opened 2020-05-31
Update `check_install.py`

kan-bayashi:
Discussed in #1984 

- remove `--torch-version` option
- check torch installation at first
- check warp_ctc installation according to the torch version
- apply black
- change `--no-cupy` -> `--no-cuda` and if enable, skip cuda availableness check.


Syzygianinfern0:
Seems to be much better now!! Although, I feel instead of having a `--no-cupy` we can change it to `--no-cuda` that even disables the last part of the checks.

kan-bayashi:
@Syzygianinfern0 Thank you for your checking! That is a nice idea. I will reflect.

kan-bayashi:
@Fhrozen Could you check the change of Dockerfile?

PR #1998, opened 2020-06-03
SCC ultrasound image speech recognition

amadeusuzx:
Silent speech recognition using ultrasound image features. Modified basing on WSJ recipe.

The dateset is from SSC dataset(Silent Speech Challenge, https://ftp.espci.fr/pub/sigma/)

Instructions:

Replaced stage 3: LM Preparation with downloading WSJ pretrained LM. （See https://github.com/espnet/espnet/blob/master/egs/wsj/asr1/RESULTS.md). This is becauese, in SSC data, train set transcripts are from TIMIT corpus, but test set transcripts are from WSJ corpus.
Default features are 30 dimentions DCT feature.According to the paper Updating the silent speech challenge benchmark with deep learning, DCT features performance is better than auto endcoder. Also you can try it by yourself. Just download features from the SSC dataset (https://ftp.espci.fr/pub/sigma/Features/) to the data directory, then replace the path intest.scp with your work path.
Actually this recipe uses parameters based on the Voxfage recipe, tuned batchsize 64 >> 16 and transformer-lr 10 >> 20.
Here's my decode results with batch size 12 and transformer-lr 20. https://storage.googleapis.com/szxs/batch12lr20.zip


egs/timit/ssr1/featprepare.py
```diff
@@ -0,0 +1,26 @@
+#!/usr/bin/env python3
```
sw005320, line 1: Can you move this script to `local/` and modify `run.sh` accordingly?

amadeusuzx: Got it! Thanks for your response!

egs/timit/ssr1/local/wsj_format_data.sh
```diff
@@ -0,0 +1,35 @@
+#!/bin/bash
```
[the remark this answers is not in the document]
amadeusuzx: Thanks...

egs/timit/ssr1/RESULTS.md
```diff
@@ -0,0 +1,35 @@
+- Environments
```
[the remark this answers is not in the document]
sw005320: If you upload your model, that is great!
I can provide our Google drive space. 
So, please send me (shinjiw@ieee.org) an email. I'll send an invitation for the Google drive space.

egs/timit/ssr1/RESULTS.md
```diff
@@ -0,0 +1,35 @@
...
+  - pytorch version: `pytorch 1.0.1`
+
+
+- Results(`batch-size 16` `transformer-lr 20`using SpecAug)
+  ### CER
+
+  |dataset|Snt|Wrd|Corr|Sub|Del|Ins|Err|S.Err|
+  |---|---|---|---|---|---|---|---|---|
+  |decode_test_decode_lm_word65000|100|5767|89.6|5.0|5.4|2.3|12.7|72.0
+  ### WER(removed sentences including abbreviations because there is no abbreviations in timit corpus)
```
sw005320, line 16: Maybe you can provide the number as it is?
Also, I extracted the WER numbers below from the PDF file in the ftp site. It seems that your result is quite good.
```
• Visual WSJ0 5k + Word-loop bigram (PCA features)
o Word level: 43.50%
o Phone level: 77.87%
• Visual WSJ0 5k + WSJ0 5k NVP bigram (PCA features)
o Word level: 79.08%
o Phone level: 90.21%
• Visual WSJ0 5k + WSJ0 5k NVP bigram (DCT features)
o Word level: 84.16%
o Phone level: 93.44%
```

sw005320:
The numbers in https://arxiv.org/pdf/1709.06818.pdf look very good. I'm not sure whether we can fairly compare it or not.

egs/timit/ssr1/RESULTS.md
sw005320: I think it is a limitation of the end-to-end system for this range of the amount of training data.
The HMM/DNN system was well designed to handle the issue.
We could do the same thing for the end-to-end system but we lose the simplicity, which we really want to avoid because this is the most important benefit.
So, I think the current number would be reasonable anyway as a simple recipe that everyone uses. We can do such additional efforts as a research item!

amadeusuzx: Sorry for taking so long to fix this PR....
I believe now the recipe can work properly. But I don't know how to solve the `No module named 'numba.decorators'` problem in CI checking.. Could please tell me what to do next for merging this PR?

sw005320:
Could you update the latest espnet and solve the conflict?
Then, it would work.

amadeusuzx:
@sw005320 
Thanks! I've updated the repo and solved the conflict files. Could you please check and merge this PR?

egs/timit_ssc/ssr1/local/featprepare.py
```diff
@@ -0,0 +1,30 @@
+#!/usr/bin/env python3
+
+if __name__ == "__main__":
+    for dataset in ["data/train/", "data/test/"]:
```
sw005320, line 4: This is a little bit hard-corded with a strong assumption for these directory structures, but it's OK. This is just a note.

amadeusuzx: Many thanks for your help!
I'd like to fix it anytime if  necessary!

PR #2009, opened 2020-06-08
Fix BLEU/PPL calculation during training

hirofumi0810:
- Fixed PPL calculation
- Fix BLEU calculation
- Refine ST/MT scripts to keep consistency

### DIFF
--- a/espnet/nets/pytorch_backend/e2e_mt_transformer.py
+++ b/espnet/nets/pytorch_backend/e2e_mt_transformer.py
@@ -14,8 +14,8 @@ import math
 import numpy as np
 import torch
 
-from espnet.nets.e2e_mt_common import ErrorCalculator
 from espnet.nets.e2e_asr_common import end_detect
+from espnet.nets.e2e_mt_common import ErrorCalculator
 from espnet.nets.mt_interface import MTInterface
 from espnet.nets.pytorch_backend.e2e_mt import Reporter
 from espnet.nets.pytorch_backend.nets_utils import get_subsample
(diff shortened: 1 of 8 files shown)


espnet/nets/pytorch_backend/e2e_mt_transformer.py
```diff
@@ -461,8 +446,6 @@ def translate(self, x, trans_args, char_list=None, rnnlm=None, use_jit=False):
                     remained_hyps.append(hyp)
 
             # end detection
-            from espnet.nets.e2e_asr_common import end_detect
-
             if end_detect(ended_hyps, i) and trans_args.maxlenratio == 0.0:
```
sw005320, line 449: This is not related to this PR.

I'm just curious whether this `end_detect` is working well for MT.
The [hyper-parameters](https://github.com/espnet/espnet/blob/6f6452a3891d11b0b0306084dd46a464ddf8e25c/espnet/nets/e2e_asr_common.py#L19) in the `end_detect` function were determined by ASR.
I think they are based on a similar problem and we could use the same hyper-parameters though.
@takaaki-hori, what do you think?

hirofumi0810:
@sw005320 I pushed a tiny modification. Now it's good to me, too.
Regarding the `end_detect` function, I just moved it to the top. I can remove it if needed.

hirofumi0810:
Maybe this is the last change.

hirofumi0810:
bleu
ppl
BLEU and PPL are falling within reasonable range but BLUE is relatively high. 
This is probably because corpus-level BLEU scores are calculated sentence by sentence and chainer reporter summarizes the results.
To calculate the exact BLEU scores, we need to get the entire sentences and then pass them though nltk.
In order to do that, we need to modify the extension function in chainer.

sw005320:
I think this is acceptable.
More or less, CER/accuracy has the same issue.

sw005320:
I checked the rest of the modifications, and it looks good for me.
Many thanks!

sw005320:
One note. This includes some refactoring, but this is still a minor modification for me. So, I merged it into the mater instead of the dev branch.

PR #2013, opened 2020-06-09
Turn off BLEU calculation during training

hirofumi0810:
In the current implementation, BLEU scores are calculated at each forward pass.
This PR modifies it so that BLEU scores are calculated in the validation only for faster training.
Also, WER calculation is turned off in the training mode when auxiliary ASR task is incorpolated.


sw005320:
LGTM!

PR #2047, opened 2020-06-16
Create espnet2 jsut/tts recipe

kamo-naoyuki:
- Create jsut/tts recipe
- Fix several bugs in tts.sh
- Implement some classes for tts frontend.
    ```sh
    ./run.sh --cleaner jaconv --g2p pyopenjtalk
    ./run.sh --cleaner jaconv --g2p pyopenjtalk_kana
    ./run.sh --cleaner tacotron --g2p g2p_en
    ```

I got comparable result with feats_type=fbank, but I couldn't get normal result with feats_type=raw. There are still some bugs about on-the-fly feature extraction, maybe which is around recipe revel. I'll check this at the next time.


egs2/jsut/tts1/run.sh
```diff
@@ -0,0 +1,40 @@
...
+    opts="--audio_format flac "
+fi
+
+train_set=tr_no_dev
+dev_set=dev
+eval_set=eval1
+
+tts_config=conf/train_tacotron2.v3.yaml
+decode_config=conf/decode.yaml
+g2p=pyopenjtalk
```
kan-bayashi, line 23: Why don't you add example?
```suggestion
# pyopenjtalk case: m i z u o m a r e e sh i a k a r a k a w a n a k U t e w a n a r a n a i n o d e s U
# pyopenjtalk_kana case: ミズヲマレーシアカラカワナクテワナラナイノデス。
g2p=pyopenjtalk
```

espnet2/text/cleaner.py
```diff
@@ -0,0 +1,40 @@
...
+
+        if normalize_types is None:
+            self.normalize_types = []
+        elif isinstance(normalize_types, str):
+            self.normalize_types = [normalize_types]
+        else:
+            self.normalize_types = list(normalize_types)
+
+    def __call__(self, text: str) -> str:
+        for t in self.normalize_types:
```
kan-bayashi, line 30: Concatenation of several cleaners is nice :)

espnet2/tasks/tts.py
```diff
@@ -121,6 +121,21 @@ def add_task_arguments(cls, parser: argparse.ArgumentParser):
             type=str_or_none,
             help="non_linguistic_symbols file path",
         )
+        parser.add_argument(
+            "--cleaner",
+            type=str_or_none,
+            choices=[None, "tacotron", "jaconv", "vietnamese"],
+            default=None,
+            help="Apply text cleaning",
+        )
```
kan-bayashi, line 130: Currently, support only a single cleaner, but the use of multiple cleaners can be, right?

kamo-naoyuki: Exactly, yes. You are great...
Multiple cleaners are also expressed as a class, so I'm not sure which is better.
It's better to consider when such additional cleaner is required really.

kamo-naoyuki:
I'll add test codes later. After it, I'll merge this PR.

kamo-naoyuki:
I fixed a bug of phoneme tokenizer.
(I thought pyopenjtalk returns list of str, but it returns a str separated by spaces)

I also added cleaner and g2p option for asr task.

kamo-naoyuki:
I'll merge this PR, but there might be still some bugs.

PR #2049, opened 2020-06-16
Update install_chainer_ctc.sh

jmiller0711:
Updated the pip install command to work on MARCC, I don't think this will have any effect on other systems.

### DIFF
--- a/tools/install_chainer_ctc.sh
+++ b/tools/install_chainer_ctc.sh
@@ -11,5 +11,5 @@ git clone https://github.com/jheymann85/chainer_ctc.git
 pip install cython
 ( 
     set -euo pipefail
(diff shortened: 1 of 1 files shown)


sw005320:
Thanks.

Unfortunately, this change affects to some environment (see https://github.com/espnet/espnet/pull/2049/checks?check_run_id=777498362#step:6:464).

Any idea to solve this?

kamo-naoyuki:
Basically, we assume using a python which is owned by each user and not using system python. This is our style. 
If users want to use system python, then they can rewrite the script easily. I don't think we need to change this.

jmiller0711:
> Basically, we assume using a python which is owned by each user and not using system python. This is our style.
> If users want to use system python, then they can rewrite the script easily. I don't think we need to change this.

PR #2059, opened 2020-06-18
Change default normalize from 16 to 1.

aadyotb:
When experimenting with a decent range of Python filter-bank feature extraction libraries (`librosa`, `torchaudio`, and `pykaldi`), I found that supplying `normalize=16` to `utils/make_fbank.sh` (instead of `normalize=1`) resulted in vastly different features from those extracted by Kaldi. Changing this value to `normalize=1` allowed me exactly recover the features extracted by Kaldi when using `torchaudio` and `pykaldi` (`librosa`'s algorithm is different enough that the comparison was not one-to-one).

More importantly, when I tried to use ESPNet v1 to train a transformer model for ASR on WSJ, using the features extracted by Kaldi (`steps/make_fbank.sh`) gave me good performance, whereas the features extracted by `utils/make_fbank.sh` were unusable. Simply changing `normalize=16` to `normalize=1` resulted in identical performance to Kaldi. 

Because this hyperparameter is incredibly easy to overlook, I suggest setting the default to be 1 instead of 16 (which I imagine is useful for PCM/multi-channel stuff, but not so much for regular WAV). For my specific use case, this ended up being the bugfix that the Kaldi-free ESPNet v2 data processing pipeline needed to regain parity with ESPNet v1's Kaldi-dependent data processing pipeline.

### DIFF
--- a/utils/make_fbank.sh
+++ b/utils/make_fbank.sh
@@ -16,7 +16,7 @@ window=hann
 write_utt2num_frames=true
 cmd=run.pl
 compress=true
-normalize=16  # The bit-depth of the input wav files
+normalize=1  # The bit-depth of the input wav files
 filetype=mat # mat or hdf5
 # End configuration section.
 
--- a/utils/make_stft.sh
+++ b/utils/make_stft.sh
@@ -13,7 +13,7 @@ window=hann
 write_utt2num_frames=true
 cmd=run.pl
 compress=true
-normalize=16  # The bit-depth of the input wav files
+normalize=1  # The bit-depth of the input wav files
 filetype=mat # mat or hdf5
 # End configuration section.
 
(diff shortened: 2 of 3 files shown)


kamo-naoyuki:
make_fbank.sh is not created for asr recipe. The normalization is the default behaviour of sndfile and matlab audioread,
This is just a option. You can give the option by yourself. We shouldn't change.

aadyotb:
My argument is that if `utils/make_fbank.sh` is to be used as a drop-in replacement for Kaldi's `steps/make_fbank.sh` (which seems to be the goal of ESPNet v2), it should have similar behavior. If the usage includes different things as well, I understand keeping the default to 16 instead of changing it to 1.

kamo-naoyuki:
No. utils/make_fbank.sh  is not aimed at kaldi compatibility at all.
Note that espnet2 is using same feature internally if feats_type=raw.

kamo-naoyuki:
And, utils/make_fbank.sh is a script for tts recipe. We can't change this.

PR #2062, opened 2020-06-19
Ngram batch decoding

qmpzzpmq:
run.sh in aishell glitch fix

I implement it through for loop since in Kenelm these aren't a better method to parallel computing.
However, I couldn't test them in any run.sh script. But I tried to add something in the batch decoding test, I need help about it.


kamo-naoyuki:
Hi, could you configure your git? The commit name is different from your git account now, but it's inconvenient for some reasons, e.g. when PR, automatic linking to reviewer can't work now.

```bash
git config --global user.name qmpzzpmq
git config --global user.email <your-email>
```

PR #2066, opened 2020-06-20
ngram batch implement

qmpzzpmq:
https://github.com/espnet/espnet/pull/2065 fixed
@kamo-naoyuki @sw005320 fix the username and email here.


espnet/nets/scorers/ngram.py
```diff
@@ -78,7 +74,29 @@ def score(self, y, state, x):
                 and next state list for ys.
 
         """
-        return self.score_partial_(y, torch.tensor(range(len(self.chardict))), state, x)
+        return self.score_partial_(y, torch.tensor(range(self.charlen)), state, x)
+
+    def batch_score(self, ys, states, xs):
```
ShigekiKarita, line 79: This for-loop based implementation looks general. Is it possible to pull this function into `BatchScorerInterface` as default impl?

qmpzzpmq: Could you explain it?
I am not sure I totally understand your words

ShigekiKarita: I think it is nice to have this for-loop based implementation for all scorers by default when the scorer inherits `BatchScorerInterface`. This is because your implementation requires nothing but the `score` method.

ShigekiKarita: I suggest you to replace this default `NotImplementedError`
https://github.com/espnet/espnet/blob/7e6be051bd575bd82310d7d460ab68b7394c7926/espnet/nets/scorer_interface.py#L86-L103
with your new for-loop implementation.

qmpzzpmq: yes, good suggestion.
Since that, and I think we can have a test from run.sh through batchscorerinterface right now.

qmpzzpmq: but it still needs a variable of 'charlen' which means how much chars in the dict. And suggestion?

espnet/nets/scorer_interface.py
```diff
@@ -100,7 +100,14 @@ def batch_score(
                 and next state list for ys.
 
         """
-        raise NotImplementedError
+        scores = list()
+        outstates = list()
+        for i, (y, state, x) in enumerate(zip(ys, states, xs)):
```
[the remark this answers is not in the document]
qmpzzpmq: Since this function will be called many times, I suggest adding some comment as warning. is it ok? for you guys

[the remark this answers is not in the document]
kamo-naoyuki: Same message from same place is shown only once.

```python
for i in range(10):
    warnings.warn("aaa")
```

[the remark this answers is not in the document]
qmpzzpmq: is it looks ok? @kamo-naoyuki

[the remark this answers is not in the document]
kamo-naoyuki: What about? I think warning is good idea.

espnet/nets/scorer_interface.py
```diff
@@ -100,7 +101,19 @@ def batch_score(
                 and next state list for ys.
 
         """
-        raise NotImplementedError
+        warnings.warn(
+            "this {} batch score is implemented throught for loop not parallel".format(
```
qmpzzpmq, line 105: I mean these changes @kamo-naoyuki

kamo-naoyuki: LGTM

sw005320:
Many thanks!
This is a simple but very important extension to deal with other non-parallelized processing with the same interface.

Can you also consider whether we can unify `BatchScorerInterface` and `ScorerInterface` later?
This would further simplify our beam search interface.

espnet/nets/scorer_interface.py
sw005320, line 105: Typo and some English correction

"this {} batch score is implemented throught for loop not parallel"
->
"this {} batch score is implemented through for loop and not parallelized"

qmpzzpmq: lol, might I leave English environment for a too long time, I will fix it quickly.

Issue #2071, opened 2020-06-21
Downloading large files using `download_from_google_drive.sh` is not working


unilight:
I found that downloading large files using `download_from_google_drive.sh` is not working well now. For example, the following is the log when downloading pre-trained model in `egs/vcc20/vc1_task2/` recipe:
```
$ ./run.sh --stop_stage -1 --pretrained_model_name tts1_en_zh
stage -1: Data and Pretrained model download
Please download the dataset following the README.
Downloading pretrained TTS model...
--2020-06-22 00:29:17--  https://drive.google.com/uc?export=download&id=1E6vzNaXT6r7Zybefat_p9ncnMOQCXtem
Resolving drive.google.com (drive.google.com)... 216.58.196.238, 2404:6800:400a:808::200e
Connecting to drive.google.com (drive.google.com)|216.58.196.238|:443... connected.
HTTP request sent, awaiting response... 200 OK
Length: unspecified [text/html]
Saving to: 'downloads/tts1_en_zh/agUQal..tar.gz'

downloads/tts1_en_zh/agUQal..tar.gz                    [ <=>                                                                                                            ]   3.19K  --.-KB/s    in 0s

2020-06-22 00:29:18 (30.0 MB/s) - 'downloads/tts1_en_zh/agUQal..tar.gz' saved [3263]


gzip: stdin: not in gzip format
tar: Child returned status 1
tar: Error is not recoverable: exiting now
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100  3263    0  3263    0     0   9217      0 --:--:-- --:--:-- --:--:--  9191
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100  3280    0  3280    0     0   9479      0 --:--:-- --:--:-- --:--:--  9479

gzip: stdin: not in gzip format
tar: Child returned status 1
tar: Error is not recoverable: exiting now
```
I believe similar problems occur when downloading large files in other recipes. Is anyone having the same issue?

qmpzzpmq:
I recommend using https://github.com/circulosmeos/gdown.pl

unilight:
@qmpzzpmq Thanks! I tried and it worked like a charm :)

Do you consider this as an alternative way to download from google drive, maintainers?

sw005320:
[...]

PR #2074, opened 2020-06-22
Fix error when downloading large files using `download_from_google_drive.sh`

unilight:
#2071 
Use https://github.com/circulosmeos/gdown.pl. It is distributed under GPL 3, so I guess it is okay to just copy to `utils/`?

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -17,6 +17,7 @@ requirements = {
         "pillow>=6.1.0",
         "editdistance==0.5.2",
         "numba==0.49",
+        "gdown",
         # DNN related
         # 'torch==1.0.1',  # Installation from anaconda is recommended for PyTorch
         "chainer==6.0.0",
--- a/utils/download_from_google_drive.sh
+++ b/utils/download_from_google_drive.sh
@@ -41,21 +41,10 @@ decompress () {
     fi
 }
 
-# Try-catch like processing
-(
-    wget "https://drive.google.com/uc?export=download&id=${file_id}" -O "${tmp}"
-    decompress "${tmp}" "${download_dir}"
-) || {
-    # Do not allow error from here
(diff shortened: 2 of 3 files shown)


sw005320:
> Use https://github.com/circulosmeos/gdown.pl. It is distributed under GPL 3, so I guess it is okay to just copy to `utils/`?

Year, as long as the header information is correctly provided, it should be no problem.

kamo-naoyuki:
How about using this one, https://github.com/wkentaro/gdown?
Copying is bad idea because we can't reflect the updating.

unilight:
> How about using this one, https://github.com/wkentaro/gdown?
> Copying is bad idea because we can't reflect the updating.

Thank you for the info, @kamo-naoyuki!
I tried this package and it worked pretty well. Both are okay with me.
It's just that this package needs to be installed via pip, but I don't really know where I should modify to include this package into the installation.
Can anyone suggest me how to do this?

sw005320:
https://github.com/espnet/espnet/blob/master/setup.py ???

sw005320:
Also, you may add some check like https://github.com/espnet/espnet/blob/94a066e7b669f5e78be6aba5863255d0a1a2d44e/egs/vcc20/vc1_task1/path.sh#L18-L23

kamo-naoyuki:
> Also, you may add some check like
> 
> >https://github.com/espnet/espnet/blob/94a066e7b669f5e78be6aba5863255d0a1a2d44e/egs/vcc20/vc1_task1/path.sh#L18-L23

If adding setup.py, you don't need to check. It's should be installed by default.

kan-bayashi:
+1 for default

sw005320:
Many thanks!

PR #2075, opened 2020-06-23
Remove long wave in template recipe

kamo-naoyuki:
I added an option for removing long wave data in espnet2 template recipe.
Note that

- Removing short data is already existing in the recipe.
- We can't use `remove_longshortdata.sh` in espnet2 because it assumes exracted feature, but espnet2 recommend on-the-fly feature extraction. 
- Removing long/short text is not applied in the template recipe yet. This is difficult, unlike wave data, because of on-the-fly text processing.
   - Instead of it, just removing empty text here.

### DIFF
--- a/egs2/TEMPLATE/tts1/tts.sh
+++ b/egs2/TEMPLATE/tts1/tts.sh
@@ -40,8 +40,10 @@ expdir=exp       # Directory to save experiments.
 local_data_opts= # Options to be passed to local/data.sh.
 
 # Feature extraction related
-feats_type=raw    # Feature type (fbank or stft or raw).
-audio_format=flac # Audio format (only in feats_type=raw).
+feats_type=raw      # Feature type (fbank or stft or raw).
+audio_format=flac   # Audio format (only in feats_type=raw).
+min_wav_duration=0.1   # Minimum duration in second
+max_wav_duration=20    # Maximum duration in second
 # Only used for feats_type != raw
 fs=16000          # Sampling rate.
 fmin=80           # Minimum frequency of Mel basis.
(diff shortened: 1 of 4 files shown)


egs2/TEMPLATE/tts1/tts.sh
```diff
@@ -295,22 +299,27 @@ if [ ${stage} -le 3 ] && [ ${stop_stage} -ge 3 ]; then
         # Remove short utterances
         _feats_type="$(<${data_feats}/${dset}/feats_type)"
         if [ "${_feats_type}" = raw ]; then
-            min_length=2560
+            # 256 is typical number of shift length
```
sw005320, line 302: Could you explain why 256?
16000 / 1000 = 16 samples / ms
10 ms * 16 = 160 samples?
Anyway, the number of shift sample length depends on the sampling rate and shift size (ms).

kamo-naoyuki: 256 is typical number for 16khz audio. 160, which is used for ASR recipe, is also okay.

More accurately, it's calculated from sampling rate and shift size, but it's impossible because shift size is not defined here since it's raw data.
We need two option, --max_wav_frames and --max_wav_seconds for each feature type,  but it looks verbose.
We don't need to calculate the accurate number here. It's enough to cut long/short data roughly.

sw005320: Sorry, I meant to say that is for ASR but I wrongly put the comment in the TTS script, but anyway both use the same 256, which is risky for me...

I don't recommend to make a rough estimate.
Note that we often use 8/16kHz data, and if you keep it 256, this will change the shift sample length two times different, and this threshold changes affect this pruning process very significantly...
Also, if such a number is embedded in `asr.sh` or `tts.sh`, people do not notice this issue, and they wrongly cut very large numbers of utterances unintentionally.

How about using `--max_wav_seconds` and `--min_wav_seconds`?
With the `fs` information, we can cut it with the exact number of samples.
In my experience, I actually set such thresholds based on the audio length (second), e.g, 20 seconds (which corresponds to 2,000 frames in the ASR case).
Why 20 seconds? This is a usual maximum length of the utterance (e.g., AMI and CHiME set this number during the manual transcription process).
Librispeedh does not follow this convention since it is a read speech, and I set 30 seconds by checking the histogram of the wave lengths of training files. (3,000 frames)

kamo-naoyuki: I'll rethink

kamo-naoyuki: I changed it to --min_wav_duration.
- If feats_type=raw: Calculate using fs
- If feats_type=fbank_pitch: Calculate using frame_shift. frame_shift is derived from fbank.conf if it can else using 10ms.

kamo-naoyuki:
I also removed the dependencies on remove_longshortdata.sh from local/data.sh. This is my mistake

sw005320:
LGTM!

Quick comment.
It's probably better to display how many utterances are pruned, but probably `utils/fix_data_dir.sh` can give us such information.

kamo-naoyuki:
Yes, utils/fix_data_dir.sh shows it.

PR #2082, opened 2020-06-23
[WIP] transducer v4

b-flo:
This PR add/modify a bunch of things:

To do / in progress:

- Default beam search:
  - [x] Optimization techniques
    - [x] prediction net caching
- N-Step Constrained beam search (modified version of: https://arxiv.org/pdf/2002.03577.pdf): 
  - [x] RNN-T
  - [x] RNN-T w/ att.
  - [x] T-T
  - [x] VIVOS decode config
  - [x] Voxforge decode config
  - [x] VIVOS results
  - [x] Voxforge results
  - [x] Optimization techniques
    - [x] prediction net caching
  - [ ] Code optimization
- Time Synchronous Decoding (https://ieeexplore.ieee.org/document/9053040): 
  - [x] RNN-T
  - [x] RNN-T w/ att.
  - [x] T-T
  - [x] VIVOS decode config
  - [x] Voxforge decode config
  - [x] VIVOS results
  - [x] Voxforge results
  - [x] Optimization techniques
    - [x] prediction net caching
    - [x] prediction net bashing
    - [x] joint net caching
  - [ ] Code optimization
- Alignment-Length Synchronous Decoding (https://ieeexplore.ieee.org/document/9053040):
  - [x] RNN-T
  - [x] RNN-T w/ att.
  - [x] T-T
  - [x] VIVOS decode config
  - [x] Voxforge decode config
  - [x] VIVOS results
  - [x] Voxforge results
  - [x] ! Fix bug final hypothesis empty with transformer (Edit: lazy patch for now, I'll investigate later)
  - [x] Optimization techniques
    - [x] prediction net caching
    - [x] prediction net bashing
    - [x] joint net caching
  - [ ] Code optimization
- Transformer-Transducer:
  - [x] Customizable architecture
  - [x] TDNN-BN-ReLU blocks for encoder part
  - [x] Conformer blocks for encoder part
  - [x] CausalConv1d for decoder part
- General:
  - [x] Beam search interface (first version)
  - [x] Beam search tests (I'll add more tests in another PR I think)
  - [x] Benchmark: decoding speed (edit: only rough benchmark for now, complete will be done after optimizations)
  - [x] Fix mtl_mode (#2162)
  - [x] Fix attention weights saving condition + PlotAttentionReport selection for unified transducer architecture
  - [x] Fix CER/WER reporting CPU-GPU bug (#2232)
  - [ ] Fix CER/WER reporting multi GPUs bug (#2084)
  - [x] Documentation
      - [x] T-T w/ customizable architecure
      - [x] New features in main README
- Recipe
  - [x] Voxforge: remove transducer run script + transfer learning scripts for voxforge (too complicated to maintain and documentation is now available for finetuning)

Note/Important: Modifications related to customizable architecture should be discussed. This version is only intended to start the discussion: it works correctly however it's not user-friendly. This feature would be better suited with espnet2 but I think it should be accessible in espnet1 for now as it is necessary to reproduce some (if not most) transformer-transducer related papers.

To be added later: prefix tree when dealing with BPE.

~P.S: I'm now focused on porting transducer to espnet2 so I'll probably keep remaining work on standby.~


sw005320:
Many thanks, @b-flo!
This is a really great effort.
I like your structured configuration and we can discuss the format and so on through this PR.

kamo-naoyuki:
(As I said same thing) If using external files for configuration, you have to care about packing model. 

https://github.com/espnet/espnet/blob/master/utils/pack_model.sh 

Basically, I don't recommend to create additional files any more because separated configuration files causes some bugs.
Please consider merge the network configuration file to the current configuration file.

b-flo:
Fine with me! I used JSON and external configs because it was more convenient. I'll add the structured architectures to the main YAML config. On that note, do you have a prefered format or this one is fine?

espnet/nets/pytorch_backend/transducer/custom_layers.py
```diff
@@ -0,0 +1,222 @@
+"""Convolution-related modules definition for transformer-transducer."""
```
ShigekiKarita, line 1: Use more informative file name, for example, `conv_layers.py`?

espnet/nets/pytorch_backend/transducer/blocks.py
```diff
@@ -0,0 +1,280 @@
...
+    Returns:
+        (function): function to create Transformer layer
+
+    """
+    if not {"d_hidden", "d_ff", "heads"} <= block_arch.keys():
+        logging.warning(
+            "Transformer layer format is: {'type: transformer', "
+            "'d_hidden': int, 'd_ff': int, 'heads': int}"
+        )
+        exit(1)
```
ShigekiKarita, line 106: Instead of `exit(1)`, how about `raise ValueError("Transformer layer format is ...")`. This can provide where the error happened.

espnet/nets/pytorch_backend/transducer/blocks.py
```diff
@@ -0,0 +1,280 @@
...
+        pos_conv_kernel_size (int) : kernel size for positionwise conv1d layer
+        pos_dropout_rate (float): dropout rate for positional encoding
+        att_dropout_rate (float): dropout rate for attention
+        dropout_rate (float): dropout rate
+
+    Returns:
+        (function): function to create Transformer layer
+
+    """
+    if not {"d_hidden", "d_ff", "heads"} <= block_arch.keys():
```
ShigekiKarita, line 101:
```suggestion
    if {"d_hidden", "d_ff", "heads"} > block_arch.keys():
```

espnet/nets/pytorch_backend/transducer/blocks.py
```diff
@@ -0,0 +1,280 @@
...
+    Returns:
+        (function): function to create CausalConv1d layer
+
+    """
+    if not {"idim", "odim", "kernel_size"} <= block_arch.keys():
+        logging.warning(
+            "CausalConv1d layer format is: {'type: causal-conv1d', "
+            "'idim': int, 'odim': int, 'kernel_size': int}"
+        )
+        exit(1)
```
ShigekiKarita, line 147: ditto

espnet/nets/pytorch_backend/transducer/blocks.py
```diff
@@ -0,0 +1,280 @@
...
+        (function): function to create TDNN layer
+
+    """
+    if not {"idim", "odim", "ctx_size", "dilation", "stride"} <= block_arch.keys():
+        logging.warning(
+            "TDNN block format is: {'type: tdnn', "
+            "'idim': int, 'odim': int, 'ctx_size': int, "
+            "'dilation': int, 'stride': int}"
+        )
+        exit(1)
```
ShigekiKarita, line 172: ditto

kamo-naoyuki:
> Fine with me! I used JSON and external configs because it was more convenient. I'll add the structured architectures to the main YAML config. On that note, do you have a prefered format or this one is fine?

Thank you to accept.  About format, for now, it's okay to embed the configuration into the yaml file, like,

```yaml
enc-block-arch: 
  - type: transformer
    d_ff: 30
    d_hidden: 32
    heads: 4
  - ...
```

Note that type must be `eval`. This is a restriction of configargparse.

```python
parser.add_arguments("--enc-block-arch", type=eval)
```

(This is not fixed format. We need to discuss more.)

espnet/nets/pytorch_backend/transducer/custom_layers.py
ShigekiKarita: Or how about splitting this file into three, e.g., `vgg2l.py`, `causal_conv1d.py`, and `tdnn.py`. I think it is the best.

kamo-naoyuki: agree

b-flo:
@ShigekiKarita I adressed your comments, thanks!
@kamo-naoyuki I included architecture configuration in the yaml file as you suggested!

siddalmia:
Hi I noticed that the transducer loss being used might be suboptimal. Have you tried using https://github.com/1ytic/warp-rnnt ? They claim to be 10x faster in the loss compute than the warprnnt_pytorch.

sw005320:
> Hi I noticed that the transducer loss being used might be suboptimal. Have you tried using https://github.com/1ytic/warp-rnnt ? They claim to be 10x faster in the loss compute than the warprnnt_pytorch.

This is really cool and it's worth trying. I'm also curious about their installation. If it is robust, that is also great. @b-flo, do you have some bandwidth to test it? If not, I may do it for you.

b-flo:
@sw005320  I'm aware of this repo but I didn't have time to run some experiments. If it's consistent with what is reported, we can consider replacing warp-transducer when training with GPU! I currently don't have any bandwidth for that task so any help is welcomed! Thanks!

b-flo:
@kamo-naoyuki when you have time, could you review the NSC part (either for RNN-T, RNN-T Att or T-T) please? Performances are great but I think I need some help/guidance on the optimization part. I kind of butchered the decoding speed I think (for N=1+ at least)...

Also, be aware that for NSC to work as if with T-T some tricky things need to be done such as:
1) left-padding for token sequence and cache
2) trimming cache in some cases to ensure cache shape is (B, [maxlen token sequence] - 1, dim) because the padded cache is conserved while the padded token sequence is not conserved (because of prefix search for exampled)

While performances are better than expected, I maybe missing some general understanding because some things seems weird to me in theory.

kamo-naoyuki:
> @kamo-naoyuki when you have time, could you review the NSC part (either for RNN-T, RNN-T Att or T-T) please? Performances are great but I think I need some help/guidance on the optimization part. I kind of butchered the decoding speed I think (for N=1+ at least)...

@b-flo The result of NSC looks nice! Sorry, but I'll get some tasks done, so probably I can't read your code soon.

b-flo:
I modified the following conditions: https://github.com/espnet/espnet/blob/master/espnet/asr/pytorch_backend/asr.py#L662-L664, attention weights saving was disabled for all cases because `mtl_mode` was not set to "transducer" (mentioned in #2162). 

Also, when I unified RNN-T and T-T E2E classes I discarded a decorator used for transformer: https://github.com/espnet/espnet/blob/master/espnet/nets/pytorch_backend/e2e_asr_transformer.py#L204-L207. An error for T-T occurs because the wrong attention_plot_class is called. I proposed a dummy fix for now, I'm not sure what is the proper approach here..?

@hirofumi0810 As you introduced these changes recently, feel free to correct me! I may have misunderstood or missed something, sorry if that's the case!

espnet/nets/pytorch_backend/transducer/blocks.py
```diff
@@ -0,0 +1,277 @@
+"""Set of methods to create transformer-based block."""
```
sw005320, line 1: This is just a note.
Using this block scheme for transducer only is a bit inconsistent with other modules, but I just want to leave them as they are.
espnet2 does not have model-specific encoder/decoder architectures (https://github.com/espnet/espnet/tree/master/espnet2/asr), and we could consistently refactor this issue in the near future there.

espnet/nets/pytorch_backend/transducer/transformer_decoder.py
```diff
@@ -295,3 +290,242 @@ def recognize_beam(self, h, recog_args, rnnlm=None):
             ]
 
         return nbest_hyps
+
+    def recognize_beam_nsc(self, h, recog_args, rnnlm=None):
```
sw005320, line 294: Is there some condition that both beam search methods produce identical results (maybe beam=1?). This would be a good unit test.

espnet/nets/pytorch_backend/transducer/utils.py
```diff
@@ -48,3 +48,98 @@ def prepare_loss_inputs(ys_pad, hlens, blank_id=0, ignore_id=-1):
     target_len = target_len.to(device)
 
     return ys_in_pad, target, pred_len, target_len
+
+
```
sw005320, line 52: Can you make a unit test for each function?

egs/vivos/asr1/conf/tuning/transducer/train_transformer-rnn_transducer.yaml
```diff
@@ -21,24 +21,24 @@ transformer-attn-dropout-rate-encoder: 0.4
 # network architecture
 ## encoder related
 etype: transformer
-transformer-input-layer: vgg2l
-elayers: 6
-eunits: 320
-dropout-rate: 0.4
+transformer-enc-input-layer: vgg2l
```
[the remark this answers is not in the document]
b-flo: Currently it works as follow: `etype/dtype` is used to determine the encoder/decoder type. If `etype/dtype: transformer` is selected, `enc-block-arch/dec-block-arch` should be filled. Each block type (Transformer and TDNN) has its ones independant parameters. Input layer are defined for both parts outside the block and dropout-rate are specified globally. Also, `enc-block-repeat` can be used to repeat the specified block (similarly to how `elayers` works in default transformer)

I can stick to a espnet2 style if I rename `etype/dtype` to `encoder/decoder` and do some refactoring to includes RNN blocks, and rename `enc-block-arch/dec-block-arch` to `enc-conf/dec-conf`. I can also add all parameters you specified to each transformer blocks.

espnet/nets/pytorch_backend/transducer/tdnn.py
```diff
@@ -0,0 +1,92 @@
+"""TDNN modules definition for transformer encoder."""
+
+import torch
+import torch.nn.functional as F
+
+
+class TDNN(torch.nn.Module):
```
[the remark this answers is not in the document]
b-flo: > Can we also call this from transformer or RNN encoder-decoder?

Sure but just didn't have time to experiment outside T-T. Maybe make a more general one without masks

espnet/nets/pytorch_backend/transducer/blocks.py
b-flo: To be honest that feature was intended in espnet2 at first. However, 1) a lot of users (including myself) are still extensively using espnet1 and 2) I think it's a necessary to reproduce some paper results using transducer, thus it should also be available in espnet1.

Also, this PR also serve me as a draft. I wanted to finish my first stack of pending tasks before doing the porting and decide what should be done next for transducer in espnet2. I have several things I want to work on but each one will take some time: streaming, general and transducer optimizations for mobile deployment, own transducer implementation (including monornnt, modified encoder and prediction output combination, etc), new recipes, etc.

espnet/nets/pytorch_backend/transducer/transformer_decoder.py
b-flo: Good question, I'll have to investigate! However it should be noted that: 1) NSC has a fixed number of expansions for all beam hypothesis and has local pruning after each expansion, 2) prefix search was removed in default beam search

b-flo:
> Can you add some simple unit tests for the NSC beam search? The beam search is the most buggy part, and the unit test helps to figure out what happened easily.

I'm finishing last two items and I'll add unit tests for all of them, no problem!

espnet/nets/pytorch_backend/transducer/transformer_decoder.py
```diff
@@ -213,8 +205,8 @@ def recognize(self, h, recog_args):
 
         return [hyp]
 
-    def recognize_beam(self, h, recog_args, rnnlm=None):
-        """Beam search implementation for transformer-transducer.
+    def recognize_beam_default(self, h, recog_args, rnnlm=None):
+        """Beam search implementation.
```
kamo-naoyuki, line 209: Sorry if it's a dump question. 
Is the beam search method of transformer different from that of rnn? Will you make/unify this beam search like ScorerInterface and BeamSearch class? Basically, we should define it as an external function/class rather than a method of each class to reduce the duplication part. 

(Even if you will, you don't need to do it in this PR. This PR is already too large.)

b-flo: > Is the beam search method of transformer different from that of rnn?

There are some minor differences but the methods is the same.

> Will you make/unify this beam search like ScorerInterface and BeamSearch class?

I guess so, I made an interface at the beginning but I wasn't too fond of it. I'll work on a new one after last beam search is implemented and optimization is done!

kamo-naoyuki: okay, thanks.

kamo-naoyuki:
If you could, would you summarize the current speed of each beam search method, like?

|method|RNN speed|Transducer speed|
|---|---|---|
|Default|||
|NSC nstep=1|||
|NSC nstep=?|||

Also, I think it's better to optimize the speed in another PR if necessary because this PR is too large.

b-flo:
I plan to do that but at the end! There'll be 4 new beam search algorithms so I really need to re-think a general interface for all of them before any optimization. I will benchmark on the first version.

b-flo:
Hi,

  I made a v0 interface for search algorithms and modified transducer decoders according to the changes. Feel free to tell me if you have any advice or idea for the interface, right now it's mainly a draft for now.
 It should be easier to review though.

sw005320:
> I made a v0 interface for search algorithms and modified transducer decoders according to the changes. Feel free to tell me if you have any advice or idea for the interface, it's mainly a draft for now.
> It should be easier to review though.

Very cool!
@ShigekiKarita, could you take a look at it?
I want to have a clear design of the whole beam search algorithms, and want your help.

espnet/nets/pytorch_backend/transducer/search.py
```diff
@@ -0,0 +1,563 @@
...
+        nbest_hyps (list of dicts): n-best decoding results
+
+    """
+    if hasattr(decoder, "att"):
+        decoder.att[0].reset()
+
+    if recog_args.beam_size == 1:
+        nbest_hyps = greedy_search(decoder, h, recog_args)
+    else:
+        if recog_args.search_type == "default":
```
ShigekiKarita, line 552: `elif` is more readable than `else: if`

ShigekiKarita:
LGTM for v0 impl. It is nice to have an explicit interface (i.e., abstract) class for decoders to describe what methods are required/called in beam search (e.g., zero_state, forward_one_step) like `espnet.nets.scorer_interface.ScorerInterface`.

b-flo:
Fine with me! @ShigekiKarita Just to be sure, you want me to add a TransducerDecoderInterface and inherits in Decoder{TT,RNNT,RNNTAtt} alongside torch.nn.Module?

ShigekiKarita:
Yes! I want those methods to have consistent args/returns types and names.

2020年7月30日(木) 19:00 b-flo <notifications@github.com>:

b-flo:
@ShigekiKarita I added an "interface" for transducer decoders but I'm not sure if it's done correctly. To be honest I'm not too experienced with that...

espnet/nets/pytorch_backend/transducer/decoder_interface.py
```diff
@@ -0,0 +1,109 @@
...
+
+import torch
+
+
+class TransducerDecoderInterface:
+    """Decoder interface for transducer models."""
+
+    def zero_state(
+        self, init_tensor: torch.Tensor = None
+    ) -> Union[Tuple[List[torch.Tensor], List[torch.Tensor]], List[torch.Tensor]]:
```
ShigekiKarita, line 16: LGTM overall. I think the type of states should be `typing.Any` because it deprends on model architecture. And there is no need to restrict it if beam search implementations don't care about that.

espnet/nets/pytorch_backend/transducer/decoder_interface.py
```diff
@@ -0,0 +1,109 @@
...
+            hyps: batch of hypothesis
+            state: batch of decoder states
+            att_w : batch of attention weights
+            att_params : attention parameters
+
+        Returns:
+            tgt: decoder outputs
+            new_state: new batch of decoder states
+            att_w: new batch of attention weights
+            lm_tokens: input token ids for LM
```
ShigekiKarita, line 69: I wonder if these `state, att_w, att_params` can be merged into one `state` or `new_state`. It looks cleaner for decoder models without `att_w` or `att_params`, e.g., DecoderTT.

espnet/nets/pytorch_backend/transducer/decoder_interface.py
```diff
@@ -0,0 +1,109 @@
...
+
+        """
+        raise NotImplementedError("forward_batch_one_step method is not implemented")
+
+    def get_idx_dec_state(
```
ShigekiKarita, line 74: Can you rename these methods consistent to the [ScorerInterface](https://github.com/espnet/espnet/blob/master/espnet/nets/scorer_interface.py)? It should be useful to unify these two in the future. For example:

- `zero_state` -> `init_state`
- `forward_one_step` -> `score`
- `forward_batch_one_step` -> `batch_score`
- `get_idx_dec_state` -> `select_state`
- `get_batch_dec_states` (ScorerInterface does not have this though)

espnet/nets/pytorch_backend/transducer/tdnn.py
```diff
@@ -0,0 +1,92 @@
...
+        self.odim = odim
+
+        self.ctx_size = ctx_size
+        self.stride = stride
+        self.dilation = dilation
+
+        self.batch_norm = batch_norm
+        self.relu = relu
+
+        self.kernel = torch.nn.Linear((idim * ctx_size), odim)
```
GNroy, line 47: Many thanks for adding TDNN!
Though I wonder why use Linear here? For example, in Espresso they use Conv1d (https://github.com/freewym/espresso/blob/master/espresso/models/speech_tdnn.py#L137).

b-flo: Hi,

After some testing, I preferred following cqluu's approach, as mentionned in the method's doc:

```
    This implementation exploits Conv1d's dilation argument.
    Thus, temporal context is defined as a value encapsulating following constraints:
        - Context must be symmetric
        - Spacing between each element of the context must be equal.
    Unfold usage is based on cvqluu's implementation (https://github.com/cvqluu/TDNN)
```

However, given I didn't extensively tested TDNN (notably speed) and what espresso does, @yuekaizhang or I may revisit that part before merging :)

espnet/nets/pytorch_backend/transducer/decoder_interface.py
b-flo: Thanks for the tip, done!

espnet/nets/pytorch_backend/transducer/decoder_interface.py
b-flo: I may remove `get_batch_dec_states` after the code optimization part!

espnet/nets/pytorch_backend/transducer/decoder_interface.py
b-flo: I merged them into a tuple, is that okay like that?

b-flo:
OK, all features are implemented. What remains is:
- code optimization (!) and benchmarking speed decoding of all search algorithms.
- discuss the "final" form of the customizable architecture and add corresponding documentation. 

Also @yuekaizhang, could you report on your experiments with TDNN please? We should investigate a bit more, notably the parts discussed in @GNroy's review.

espnet/nets/pytorch_backend/transducer/tdnn.py
GNroy: Hi,

Sorry for the late response.

As I understood from your docs, there are strong constraints for your TDNN. I think, with equal splicing and symmetric context, Conv1d implementation can be at least more straightforward (without the unfold part in particular).
Also, I found Conv1d TDNN variant in another cqluu's repo (https://github.com/cvqluu/Factorized-TDNN/blob/13506c3cd3c3b6b41de560d2d2ca2ef09cf96438/models.py#L212)

On the other hand, Kaldi uses Linear, but they do not have the limitations mentioned above (so their network structure can be almost arbitrary).

b-flo: > I think, with equal splicing and symmetric context, Conv1d implementation can be at least more straightforward (without the unfold part in particular).

Looking again at espresso code and the one in FTDNN, I guess you're right.

> On the other hand, Kaldi uses Linear, but they do not have the limitations mentioned above (so their network structure can be almost arbitrary).

Yes, performance degradation with these constraints should be evaluated more carefully. I can add TDNN with "free definition" but a lot of checking need to be done compared to current version, where the constraints are encapsulated with single values (instead of list of values).

espnet/nets/pytorch_backend/transducer/decoder_interface.py
```diff
@@ -0,0 +1,100 @@
...
+from typing import Any
+from typing import Dict
+from typing import List
+from typing import Tuple
+from typing import Union
+
+import torch
+
+
+class TransducerDecoderInterface:
```
sw005320, line 12: Is it possible to add this class to https://github.com/espnet/espnet/blob/master/espnet/nets/scorer_interface.py or moving this file as https://github.com/espnet/espnet/blob/master/espnet/nets/scorer_interface_transducer.py ?

@ShigekiKarita, what do you think?
I want to place it in the same manner as our output-sync scorer interface.

espnet/nets/pytorch_backend/transducer/search.py
```diff
@@ -0,0 +1,654 @@
+"""Search algorithms for transducer models."""
```
sw005320, line 1: ditto.
maybe you can move this as https://github.com/espnet/espnet/blob/master/espnet/nets/beam_search_transducer.py ???

Again, @ShigekiKarita, what do you think?

espnet/nets/pytorch_backend/transducer/decoder_interface.py
ShigekiKarita: I agree with scorer_interface_transducer.py or transducer_decoder_interface.py (the module name should be equal to the class name)

espnet/nets/pytorch_backend/transducer/search.py
ShigekiKarita: agree

yuekaizhang:
> Also @yuekaizhang, could you report on your experiments with TDNN please? We should investigate a bit more, notably the parts discussed in @GNroy's review.

Actually, I didn't use the tdnn+self-attention arch in this PR. I have some results using contextnet (implementation from @pengchengguo )

Aishell 1(170hrs) + speed perturbation + specaug w/o LM using rnn-t
| Exp | WER (test, %) | 
| :-----: | :-----------------------: |
| contextnet(M) + global context| 10.1 | 
| context_window 32 frames| 11.3 |
| context_window 16 frames | 12.4 | 
| vgg-blstm w/o specaug, w/o sp | 13.0 |
|ref:transformer+ctc|~6.0|

However, since the SE layer in contextnet would bring a lot of latency for streaming task. E.g. orignial latency N frames, if we use P frames as context of SE layer, the total latency would be NxP frames.

So, I am now trying to some other variants network and also try to improve the SE layer.

espnet/nets/pytorch_backend/transducer/decoder_interface.py
b-flo: If it's okay for you, I prefer changing to `transducer_decoder_interface.py` as it's not really the same interface as ScorerInterface and other ones.

espnet/bin/asr_recog.py
```diff
@@ -144,6 +144,31 @@ def get_parser():
                         If margin=0 (default), this function is disabled""",
     )
     # transducer related
+    parser.add_argument(
+        "--search-type",
+        type=str,
+        default="default",
+        choices=["default", "nsc", "tsd", "alsd"],
```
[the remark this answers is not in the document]
b-flo: Done. Is that okay or you wanted me to add a brief explanation for each algorithms?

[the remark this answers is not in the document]
sw005320: This looks fine for me.

espnet/nets/pytorch_backend/transducer/decoder_interface.py
sw005320: OK for now, but we'd like to unify them in the near future.
So, could you please leave some documents of how they are different and the possibility of unifying them?

sw005320:
@b-flo, I think the PR is getting close to a good shape.
Do you want to keep on working on this PR for the rest of your action items?
We can also make some of them in the later PR.
Please let me know your opinion.

b-flo:
Outside a bug with multiple GPU and CER/WER report I can fix later, I think remaining items should be adressed.

- Code optimization and benchmarking

All added beam search algorithms should be optimized a bit as speed degradation with batching is too big. I plan to benchmark the methods after that

- Documentation: T-T w/ customizable architecure

Are we okay or does the current T-T architecture definition needs some modification? It it's okay I'll add documentation for explain the difference with other models.

sw005320:
> Are we okay with the current T-T architecture definition or some modifications are needed? It it's okay, I'll add documentation for explain the difference with other models.

Sounds good for me.
And again you can also do it with the later PR.
If you're OK, let's merge this PR!

b-flo:
@kamo-naoyuki @ShigekiKarita @sw005320 Could you take a look again at the PR when you have time please? Except for optimization related to bashing, I think everything is done on my side!
Also, @pengchengguo could you review the conformer definition in customizable architecture please? I can't add you as reviewer but I wanted your opinion on what should be a general option and block option

sw005320:
Many thanks, @b-flo!
Sure, I'll review this PR in the next week.

b-flo:
I forgot to say: please report if you review TSD, ALSD or NSC and have any idea to optimize the bashing part, even if it's a major refactor. I'll resume porting transducer to espnet2 starting tomorrow, it may be a good time to rethink that part from scratch!

@pengchengguo I also forgot to ask you if you're working on adding conformer to espnet2? If so please tell me as I'll add missing items related to transducer models in espnet1

pengchengguo:
Hi @b-flo, I'm not working on conformer in espnet2.

kamo-naoyuki:
> @pengchengguo I also forgot to ask you if you're working on adding conformer to espnet2? 

I'll do it.

kamo-naoyuki:
@b-flo Could you try to prepare  more unittest as possible? At least, we need to keep the current coverage.

kamo-naoyuki:
and if possible, please use type annotation instead of docstring.

espnet/nets/beam_search_transducer.py
```diff
@@ -0,0 +1,634 @@
...
+
+    Returns:
+        hyp (list of dicts): 1-best decoding results
+
+    """
+    init_tensor = h.unsqueeze(0)
+    dec_state = decoder.init_state(init_tensor)
+
+    hyp = {
+        "score": 0.0,
```
kamo-naoyuki, line 32: Could you use dataclass or namedtuple instead of dict to represent Hypothesis?

kamo-naoyuki: https://docs.python.org/3/library/dataclasses.html

kamo-naoyuki:
@b-flo Can you let me to port the transducer part to espnet2? I need to consider how to combine the feature in this PR in espnet2  without breaking the current style.

egs/vivos/asr1/RESULTS.md
```diff
@@ -1,120 +1,487 @@
...
 |---|---|---|---|---|---|---|---|---|
-|decode_test_decode_transducer|760|32771|85.4|10.8|3.8|3.0|17.6|94.2|
-|decode_test_decode_transducer_lm|760|32771|85.6|10.6|3.8|3.0|17.4|93.3|
-|decode_test_decode_transducer_lm_word7184|760|32771|86.1|10.2|3.7|2.8|16.7|92.1|
-|decode_train_dev_decode_transducer|100|4007|84.7|12.8|2.6|3.2|18.6|97.0|
-|decode_train_dev_decode_transducer_lm|100|4007|84.7|12.6|2.7|3.3|18.6|97.0|
-|decode_train_dev_decode_transducer_lm_word7184|100|4007|84.9|12.4|2.7|3.3|18.4|97.0|
+|decode_test_decode_alsd|760|32771|85.3|11.5|3.2|3.6|18.3|93.8|
+|decode_test_decode_alsd_lm|760|32771|85.7|11.1|3.2|3.6|17.9|93.0|
+|decode_test_decode_alsd_lm_word7184|760|32771|86.2|10.7|3.1|3.5|17.3|92.0|
```
kamo-naoyuki, line 18: What is difference between lm and lm_word7184?
I think it's enough to keep the best result about lm.

egs/vivos/asr1/RESULTS.md
```diff
@@ -1,120 +1,487 @@
...
+|decode_test_decode_default|760|32771|84.6|11.6|3.8|3.2|18.6|93.3|
+|decode_test_decode_default_lm|760|32771|84.1|10.9|5.0|2.9|18.8|93.0|
+|decode_test_decode_default_lm_word7184|760|32771|84.4|10.7|4.9|2.8|18.4|92.0|
+|decode_test_decode_nsc|760|32771|85.2|11.4|3.4|3.5|18.3|93.0|
+|decode_test_decode_nsc_lm|760|32771|85.5|11.1|3.4|3.4|17.9|92.5|
+|decode_test_decode_nsc_lm_word7184|760|32771|85.9|10.8|3.3|3.3|17.4|92.0|
+|decode_test_decode_tsd|760|32771|85.0|11.4|3.5|3.2|18.2|93.6|
```
kamo-naoyuki, line 25: Would you note the decoding parameter somewhere? e.g. The n of `nsc`.

b-flo:
> @b-flo Could you try to prepare more unittest as possible? At least, we need to keep the current coverage.

I think something went wrong somewhere in recent commits. As mentionned in a previous edit the coverage should be increased by about 0.26/0.27%, or at least it was before adding conformer. For which I also added some test...

> @b-flo Can you let me to port the transducer part to espnet2? I need to consider how to combine the feature in this PR into espnet2 without breaking the current style.

If it's okay for you I would like to port transducer to espnet2 under your guidance! I'm also rethinking the whole implementation to follow current style and resolve current limitations (bashing being one of them).

egs/vivos/asr1/RESULTS.md
b-flo: The meaning of "N" is explained in the tutorial part alongside other algorithm parameters and the `n-step` parameter is given in `decode_nsc.yaml` config, is that okay?

egs/vivos/asr1/RESULTS.md
b-flo: `lm` is a char LM and the other a word LM, I just followed the style from previous CTC results! I can remove one of them but the best result can be different in accordance to CER or WER.

espnet/nets/beam_search_transducer.py
b-flo: Sure, I'll take a look!

egs/vivos/asr1/RESULTS.md
kamo-naoyuki: Did you test n==3?
I didn't read the papers detail, but just only according to your report, I understood that only `NSC [nstep=1, prefix-alpha=1]` or `ALSD [u-max=20]` are faster than your new default, so I'd like to the result, too.

b-flo: Yes, I tested N=[1, 3] for NSC and max_sym_exp=[2, 4] for TSD (note: N=(max_sym_exp+1)) . The results are for respectively N=3 and max_sym_exp=4.

I'll re-run experiments for N=1 and max_sym_exp=2 but from what I remember CER is about 20% for NSC and 60% for TSD with vanilla RNNT. The difference between the two can be explained by a condition in NSC I added, related to blank.

b-flo: It should be noted that with most corpora (such as Librispeech, WSJ, ESTER, etc) you would use N=1 and max_sym_exp=2. Vivos is a one of the special "case" which motivated the modification of OSC into NSC

egs/vivos/asr1/RESULTS.md
kamo-naoyuki: Word LM looks also better for CER ?

b-flo: There are less cases now than I thought, I count only three cases where CER with word lm is greater than with char lm. I'll delete entries with char lm.

kamo-naoyuki:
> If it's okay for you I would like to port transducer to espnet2 under your guidance! I'm also rethinking the whole implementation to follow current style and resolve current limitations (bashing being one of them).

Okay.

egs/vivos/asr1/RESULTS.md
b-flo: With vanilla RNN-T (performance difference for NSC and TSD with same parameters would be the same across models):

| Algorithm [beam=10] | CER (test) | WER (test) |
| - | - | - |
| Default (new) | 18.6 | 39.0 |
| NSC [nstep=1, prefix-alpha=1] | 24.3 | 50.0 |
| NSC [nstep=3, prefix-alpha=2] | 18.3 | 38.7 |
| ALSD [u-max=10] | 33.2 | 50.8 | 
| ALSD [u-max=100] | 18.3 | 38.4 |
| TSD [max_sym_exp=2] | 64.4 | 80.1 |
| TSD [max_sym_exp=4] | 18.2 | 38.5 |

I also forgot to mention other things:
- The condition I added in NSC is for N=1 is not complete in the espnet version: it shouldn't be applied depending on other parameters (e.g.: if monoRNNT is used, if a prior validation step give a theoretical N equal to 1, etc). However, I confirmed on several dataset (Librispeech, WSJ, french corpus, etc) that even in this cases the condition doesn't decrease performance in any ways (and could even increase).
- If I add the condition to TSD algorithm, I can reach similar performance for max_sym_exp=2 vs N=1.

espnet/nets/beam_search_transducer.py
b-flo: Hum, one question: did you have something in mind when using dataclasses instead of dict? I'm wondering if some refactoring could be done for other algorithms dealing with list of dict originally.

kamo-naoyuki: Sorry, I'm not sure what you mean. Do you ask if dict can be replaced to dataclasses?

kamo-naoyuki: This is very basic strategy in Python programming, or generally, prefer style in programming language.
If you use dict object, no one can inspect what attribute exists without accessing it. You can refer some websites about namedtuple and you may know why we use it.

b-flo: Sorry, I guess I badly explained: I understand replacing dict by a dataclass is straightforward in case of greedy decoding. However, for other algorithms we are dealing with list of dict. Instead of simply replacing dict by a dataclass in that cases, I was wondering if something could be done to speed up a bit (e.g.: using list of objects in dataclass and relying on inner methods for sorting, etc, or some other scheme)

kamo-naoyuki: The beam search class by @ShigekiKarita is good example of using NamedTuple.

kamo-naoyuki: What code specifically do you have problem with?

kamo-naoyuki: Basically, dataclasses is just a rich dict, so I don't think speedup or slowdown can be done.

b-flo: I don't have problem to replace dict to dataclass or namedtuple! As I'm not too familiar with these modules I was wondering if I could rethink in a clever way how I deal with hypothesis.

That was just a question I had on top of my head :)

kamo-naoyuki:
@b-flo By the way, do you have some resource to test https://github.com/1ytic/warp-rnnt ? I think this is also important.

b-flo:
Yes but I won't adress that part before finishing this PR and porting transducer to espnet2 because the next item on my todo list is replacing warp-transducer by our own kernel (w/ some variant rnnt loss and optimisation techniques) :)

I'm a bit busy right now but I'll resume and finish current items in a few days

espnet/nets/beam_search_transducer.py
b-flo: There is a slight degradation in terms of decoding speed but I guess it's expected given changes!

b-flo:
@kamo-naoyuki I'm honestly clueless on why codecov reports such decrease. Code related to transducer (new additions included) should be covered at ~92%... It seems some cases are missed, for example `vgg2l.py` coverage is reported at 23.8% while it should be at 100%.

kamo-naoyuki:
Okay, please ignore for now. I'll check later.

kamo-naoyuki:
Oh, circle ci doesn't exist in this PR. Coverage is uploaded in circle ci test, so, but I can't understand why it doesn't work...

b-flo:
I couldn't find the cause but the decrease happened between commit `cdd566e` and `51eab66`, the latter changes are related to a conflict I resolved directly via github.

kamo-naoyuki:
Could you recreate this PR? Maybe it fix.

PR #2087, opened 2020-06-24
Fix bug for feats_type=extracted

kamo-naoyuki:
Fix minor bug in asr template recipe


kamo-naoyuki:
I'll merge though ci error exists.

PR #2091, opened 2020-06-24
Add cupy.done in make python

kan-bayashi:
@kamo-naoyuki I found `cupy.done` is not included in `make python`.
This causes an error in `make check_install`.
Is it OK to add or are there any reason?


kamo-naoyuki:
It should be a bug, I think.

PR #2095, opened 2020-06-24
CTC forced alignment in E2E ASR Transformer model

simpleoier:


sw005320:
@jnishi, could you review this PR?

espnet/bin/asr_ctc_align.py
```diff
@@ -0,0 +1,293 @@
+#!/usr/bin/env python3
```
sw005320, line 1: Can you remove some unused options?

espnet/nets/pytorch_backend/e2e_asr_transformer.py
```diff
@@ -634,3 +634,74 @@ def calculate_all_attentions(self, xs_pad, ilens, ys_pad):
                 ret[name + "_time"] = m.attn_t.cpu().numpy()
                 ret[name + "_freq"] = m.attn_f.cpu().numpy()
         return ret
+
+    def ctc_align(self, x, y):
```
sw005320, line 638: How about moving this function to https://github.com/espnet/espnet/blob/master/espnet/nets/pytorch_backend/ctc.py or nets_utils.py?

kan-bayashi:
This is a great feature.
I think we can use this force alignment results for non-AR TTS models.

espnet/nets/pytorch_backend/e2e_asr_transformer.py
```diff
@@ -634,3 +634,74 @@ def calculate_all_attentions(self, xs_pad, ilens, ys_pad):
...
+            numpy.zeros((lpz.size(0), len(y_int))) - 100000000000.0
+        )  # log of zero
+        state_path = (
+            numpy.zeros((lpz.size(0), len(y_int)), dtype=numpy.int16) - 1
+        )  # state path
+
+        logdelta[0, 0] = lpz[0][y_int[0]]
+        logdelta[0, 1] = lpz[0][y_int[1]]
+
+        import six
```
jnishi, line 672: How about moving this "import" to the beginning of the file?

espnet/bin/asr_ctc_align.py
```diff
@@ -0,0 +1,293 @@
...
+    # recog
+    logging.info("backend = " + args.backend)
+    if args.num_spkrs == 1:
+        if args.backend == "pytorch":
+            if args.num_encs == 1:
+                # Experimental API that supports custom LMs
+                if args.api == "v2":
+                    raise NotImplementedError(f"--api {args.api} is not supported")
+                else:
+                    from espnet.asr.pytorch_backend.asr import ctc_align
```
jnishi, line 277: How about moving this "import" to the beginning of the file?

espnet/nets/pytorch_backend/e2e_asr_transformer.py
simpleoier: Thanks for the suggestion. I will try to move it to ctc.py.

espnet/nets/pytorch_backend/e2e_asr_transformer.py
simpleoier: Thanks for the suggestion.

espnet/bin/asr_ctc_align.py
simpleoier: Thanks for the suggestion.

espnet/bin/asr_ctc_align.py
simpleoier: done.

sw005320:
One more request. Could you add an example in https://github.com/espnet/espnet/blob/master/README.md#asr-demo ?

Also, question. When there is an OOV token, how would it behave? You don’t have to deal with it, but I’m curious how it would expect to work. If you’d add some warning for such a case in `ctc_align_wav.sh`, that is perfect.

simpleoier:
> One more request. Could you add an example in https://github.com/espnet/espnet/blob/master/README.md#asr-demo ?

Done.

> Also, question. When there is an OOV token, how would it behave? You don’t have to deal with it, but I’m curious how it would expect to work. If you’d add some warning for such a case in `ctc_align_wav.sh`, that is perfect.

From my experience, there will be an "<unk>" symbol being aligned at the position of OOV token.

sw005320:
Many thanks!

Do you think other pre-trained models can be used?
Many of our pre-trained models have CTC branches and I think it is possible.

simpleoier:
> Do you think other pre-trained models can be used?
> Many of our pre-trained models have CTC branches and I think it is possible.
In my opinion, any pre-trained model have CTC branches can be used. The wsj example I used was from the egs/wsj/asr1 (not from myself). It has no special requirements.

sw005320:
That is again very cool!

PR #2098, opened 2020-06-24
Update egs2/TEMPLATE/README.md

kamo-naoyuki:


sw005320:
Very good idea to make such a document.
How about making a guideline to port the Kaldi and espnet1 recipes to the espnet2 recipe?
(cf https://github.com/espnet/espnet/wiki/How-to-port-the-Kaldi-recipe-to-the-ESPnet-recipe%3F)

kamo-naoyuki:
I'll think a little.

Issue #2099, opened 2020-06-25
commonvoice pretrained model bpe missing


Fpooyan:
Hi
I am trying to use commonvoice pretrained models provided in the repo. while running run.sh in strage 2, bpe model should be provided but it is missing from the downloaded model dir.
any help is highly appreciated!

sw005320:
@ftshijt, could you re-pack the model by including the BPE model?

ftshijt:
Sure, I will repack it soon.

Fpooyan:
Hi.
Thanks for your quick response, but I can not find the added model file in the Persian model directory.
Is it also included?

ftshijt:
> Hi.
> Thanks for your quick response, but I can not find the added model file in the Persian model directory.
> Is it also included?

Thanks for bringing it out. Please check again.

PR #2101, opened 2020-06-26
updatemodel link (add pre-trained bpe model and lm model)

ftshijt:
Fix issue of https://github.com/espnet/espnet/issues/2099


sw005320:
Many thanks!

PR #2102, opened 2020-06-26
Fix feature extractor of TTS for compatibility

kamo-naoyuki:
I fixed feature extractor of TTS for espnet2 for compatibility.

- Use hann window in stft layer: The default of torch.stft is all 1 window. I forgot it, sorry...  **This changing also affects ASR results**.
- Change to log_e -> log_10

Note that we use different definition for log-spec between TTS and ASR now.

- ESPnet2
    - TTS: log_10(abs(torch.stft))
    - ASR: log_e(power(torch.stft))
- ESPnet1
    - TTS: log_10(abs(librosa.stft))
    - ASR: From Kaldi: log_e(power(kaldi-stft))

I found that the default behaviour of torch.stft is almost compatible with librosa.stft, while scipy.signal.stft is different.

I also added a compatibility testing between espnet2 and espnet1 feature for TTS. Now I assure you that feats_type=raw and feats_type=fbank in TTS recipe uses **almost** same feature (numerical error exists.).

The definition for ASR follows Kaldi's spectrogram basically, but scaling and padding of spectrogram and mel definition are based on librosa, thus it is different from Kaldi.

### DIFF
--- a/espnet2/layers/stft.py
+++ b/espnet2/layers/stft.py
@@ -1,6 +1,5 @@
 from typing import Optional
 from typing import Tuple
-from typing import Union
 
 import torch
 from typeguard import check_argument_types
@@ -13,8 +12,9 @@ class Stft(torch.nn.Module, InversibleInterface):
     def __init__(
         self,
         n_fft: int = 512,
-        win_length: Union[int, None] = 512,
+        win_length: int = None,
         hop_length: int = 128,
+        window: Optional[str] = "hann",
         center: bool = True,
         pad_mode: str = "reflect",
         normalized: bool = False,
@@ -32,6 +32,9 @@ class Stft(torch.nn.Module, InversibleInterface):
         self.pad_mode = pad_mode
         self.normalized = normalized
         self.onesided = onesided
+        if window is not None and not hasattr(torch, f"{window}_window"):
+            raise ValueError(f"{window} window is not implemented")
+        self.window = window
 
     def extra_repr(self):
         return (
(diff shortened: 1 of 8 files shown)


espnet2/layers/stft.py
```diff
@@ -64,14 +66,23 @@ def forward(
...
+        # NOTE(kamo):
+        #   The default behaviour of torch.stft is compatible with librosa.stft
+        #   about padding and scaling.
+        #   Note that it's different from scipy.signal.stft
+
         # output: (Batch, Freq, Frames, 2=real_imag)
         # or (Batch, Channel, Freq, Frames, 2=real_imag)
+        window = torch.hann_window(
+            self.win_length, dtype=input.dtype, device=input.device
+        )
```
kan-bayashi, line 78: Is it OK not to optionize?

kamo-naoyuki: We may use only hann, but I added just in case.

kamo-naoyuki:
@kan-bayashi 
The quality both of raw and fbank is still worse than espnet1.
I found the decoding for every utterances get the following warning now, so I think there are some bugs. 

```
WARNING: output length reaches maximum length (BASIC5000_0251).
```

Do you have any idea?

kan-bayashi:
Hmm. It seems that stop token prediction is failed.
I will check in detail the hyperparameters.

kan-bayashi:
https://github.com/espnet/espnet/blob/e1873ef5cd514fd5aec0ec23d71be98e70ae6448/egs2/jsut/tts1/conf/train_tacotron2.v3.yaml#L55
It seems that num_epoch is very small. Did you use this config?

kamo-naoyuki:
Actually I trained with 15epoch now, but I also stopped espnet1 training at 15ep and espnet1's result doesn't have such problem, so different problem exists.

kan-bayashi:
Hmm. In the case of TTS, the validation best is not the best. Anyway, I will run the recipe and check the performance today.

Issue #2106, opened 2020-06-26
espnet2/bin/split_scps.py is slow


sw005320:
I found that `split_scps.py` is very slow.
I'm now trying to split the Librispeech text data (40,699,501), which is normal size as a text data, and it takes 20 minutes to process 1,000,000.
So, it would take approximately 800 minutes to split the data.

It may be specific to my own environments to some extent (because CPU load average is ~30%, which means that some data access bottleneck happens).

kamo-naoyuki:
Maybe the following part is the reason: https://github.com/espnet/espnet/blob/e1873ef5cd514fd5aec0ec23d71be98e70ae6448/espnet2/bin/split_scps.py#L58-L62

PR #2109, opened 2020-06-27
espnet2 librispeech recipe

sw005320:
Will check the following items in later PRs
- [x] add `README.md`
- [ ] check the STFT analysis condition in detail
- [x] temporally avoid `integration_test_ctc_align_wav.bats` (see https://github.com/espnet/espnet/issues/2146)

- known issues
  - `spm_train` requires huge memory (currently skipping)
  - LM training is very slow (currently skipping) -> fixed in #2114


kamo-naoyuki:
>   - LM training is very slow (currently skipping)

I heard @ShigekiKarita seq_rnn (= CUDNN-RNN), which is used in espnet2, is slower than RNN-Cell. I don't know why.
Hmm, but very slowness is strange. I don't feel it

sw005320:
> > * LM training is very slow (currently skipping)
> 
> I heard @ShigekiKarita seq_rnn (= CUDNN-RNN), which is used in espnet2, is slower than RNN-Cell. I don't know why.
> Hmm, but very slowness is strange. I don't feel it

This seems to come from some preprocessing before actual training. Probably, some bottleneck due to the huge (but normal range as LM training data) lines of scp files. This is why I’m also testing the scp file splitting but it is also very slow, as I reported in #2106, but I need some time to figure out what happened after I get some extra GPUs.

sw005320:
> I don’t feel it.

Good information.
I guess you already tested it in the librispeech LM training data or similar scale of data.
Then, could you share your config file?

kamo-naoyuki:
The preprocessing includes bpe, so it might be the bottleneck, hmm, but it seems enough fast.
The log file when training, iter_time and train_time, can tell you more information.

kamo-naoyuki:
Note that huge size doesn't affect the training speed. Your problem should be reproduced using small  dataset.
In addition, if the reason is bpe proprocess, it should also make ASR speech degraded. If you see  it in only LM training, it's more strange.

sw005320:
I got the first tentative result.
I think this number is OK by considering that both ASR/LM training is on-going and the beam size is not large.

|decode_dev_clean_decode_asr_beam_size20_lm_train_lm_bpe_valid.loss.best_asr_model_valid.acc.best|2703|54402|97.2|2.4|0.3|0.3|3.1|35.2|
|decode_dev_other_decode_asr_beam_size20_lm_train_lm_bpe_valid.loss.best_asr_model_valid.acc.best|2864|50948|93.2|6.0|0.8|0.8|7.6|54.7|
|decode_test_clean_decode_asr_beam_size20_lm_train_lm_bpe_valid.loss.best_asr_model_valid.acc.best|2620|52576|97.0|2.6|0.4|0.4|3.4|37.4|
|decode_test_other_decode_asr_beam_size20_lm_train_lm_bpe_valid.loss.best_asr_model_valid.acc.best|2939|52343|92.9|6.1|1.0|0.9|8.0|58.3|

egs2/librispeech/asr1/local/data_prep.sh
```diff
@@ -0,0 +1,85 @@
...
+  for chapter_dir in $(find -L $reader_dir/ -mindepth 1 -maxdepth 1 -type d | sort); do
+    chapter=$(basename $chapter_dir)
+    if ! [ "$chapter" -eq "$chapter" ]; then
+      echo "$0: unexpected chapter-subdirectory name $chapter"
+      exit 1
+    fi
+
+    find -L $chapter_dir/ -iname "*.flac" | sort | xargs -I% basename % .flac | \
+      awk -v "dir=$chapter_dir" '{printf "%s %s/%s.flac\n", $0, dir, $0}' >>$wav_scp|| exit 1
+
```
kamo-naoyuki, line 57: (Not problem of this PR)
This format violates the format of wav.scp in Kaldi (=flac format is used), so this wav.scp can't perform with feats_type=pitch_fbank, but this is useful for feats_type=raw.

I'll rethink how they can coexist.

sw005320:
@simpleoier, FYI, I just renamed `integration_test_ctc_align_wav.bats` to `fixme_integration_test_ctc_align_wav.bats` to skip the error in this PR, as we discussed in  https://github.com/espnet/espnet/issues/2146.

simpleoier:
> @simpleoier, FYI, I just renamed `integration_test_ctc_align_wav.bats` to `fixme_integration_test_ctc_align_wav.bats` to skip the error in this PR, as we discussed in #2146.

I see. I am fixing it.

PR #2114, opened 2020-06-27
Make read_text faster

kamo-naoyuki:
We found `read_2column_text()` and `load_num_sequence_text` is very slow now. They re used for reading `text` and `*_shape.txt`. @Cescfangs may also know this problem?

I deleted some bottlenecks, but it might be still slow.

### DIFF
--- a/espnet2/fileio/read_text.py
+++ b/espnet2/fileio/read_text.py
@@ -33,7 +33,7 @@ def read_2column_text(path: Union[Path, str]) -> Dict[str, str]:
             k, v = sps
             if k in data:
                 raise RuntimeError(f"{k} is duplicated ({path}:{linenum})")
-            data[k] = v.rstrip()
+            data[k] = v
     return data
 
 
--- a/espnet2/fileio/read_text.py
+++ b/espnet2/fileio/read_text.py
@@ -1,12 +1,10 @@
-from io import StringIO
 import logging
 from pathlib import Path
 from typing import Dict
+from typing import List
 from typing import Union
 
-import numpy as np
 from typeguard import check_argument_types
-from typeguard import check_return_type
 
 
 def read_2column_text(path: Union[Path, str]) -> Dict[str, str]:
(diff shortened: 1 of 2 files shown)


Cescfangs:
how about using tools like pandas? this could be 3x faster
read2column

kamo-naoyuki:
Great idea! Thanks!

kamo-naoyuki:
@Cescfangs Maybe did you test it with the current implementation? For me, pandas is comparable performance with, or slower than, this PR's implementation.

```python
In [1]: import numpy as np 
   ...: import pandas as pd 
   ...: from espnet2.fileio.read_text import read_2column_text 
   ...:  
   ...:  
   ...: def read_2column_text_pandas(filename): 
   ...:     df = pd.read_csv(filename, names=[0, 1], delim_whitespace=True) 
   ...:     d = {} 
   ...:     for key, value in zip(df[0], df[1]): 
   ...:         if key in d: 
   ...:             raise RuntimeError 
   ...:         d[str(key)] = value 
   ...:     return d 
   ...:   
   ...: with open("a.scp", "w") as f: 
   ...:     for i in range(100000): 
   ...:         n = int(np.random.randint(0, 10, 1)) 
   ...:         f.write(f"{i} ") 
   ...:         for j in range(n - 1): 
   ...:             f.write(f"{i + j},") 
   ...:         f.write(f"{i + n}") 
   ...:         f.write(f"\n") 

In [2]: %timeit read_2column_text("a.scp")                                                                
104 ms ± 1.79 ms per loop (mean ± std. dev. of 7 runs, 10 loops each)

In [3]: %timeit read_2column_text_pandas("a.scp")                                                         
186 ms ± 15.1 ms per loop (mean ± std. dev. of 7 runs, 10 loops each)
```

Cescfangs:
Wow! I tried with your new implementation, it is about 4x faster than before, and 1x faster than my pandas version(with small, medium and large feats.scp), that's is really fast!

sw005320:
I also confirmed the speed improvement based on this fix (but mainly the bottleneck comes from the `np.loadtxt` part. Now, I can reduce the preprocessing (file reading and batching) time with 5 min even with our slow environment for 40 million scp files, which is acceptable.
Many thanks for your quick fix @kamo-naoyuki!

kamo-naoyuki:
Thanks to confirm, then I'll merge this PR.

PR #2128, opened 2020-06-30
[WIP] using Nemo's pretrained model to score

yuekaizhang:
This PR is for discussion. Trying to use  [nemo](https://github.com/NVIDIA/NeMo/tree/master/nemo) to replace the espnet model to score. I am now cut off the 83 dims feature into 64 dims to fit nemo's model. The decoing result looks acceptable comparing with changing espnet's fbank config to 64 dims.
I didn't add the code for convert nemo's model to torchscript model for now. It is not very related to espnet and I am not sure where to add it.

Actions
- [ ] using espnet2 to implement nemo's feature extraction (which could allow on-the-fly feature extraction)
- [ ] fix the issue for eos and sos token id (since nemo didn't have it)

### DIFF
--- a/espnet/asr/pytorch_backend/recog.py
+++ b/espnet/asr/pytorch_backend/recog.py
@@ -182,7 +182,8 @@ def recog_nemo(args):
         params = yaml.load(f)
     vocab = params["labels"]
     vocab.append('<blank>')
-    sos_id = 27 # check for the vocab to see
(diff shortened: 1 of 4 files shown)


sw005320:
Many thanks!
This effort must be a good example of how we port models from other frameworks.
Please keep on working, and feel free to discuss some design level issues.

PR #2134, opened 2020-07-02
Support TTS-Transformer in ESPnet2

kan-bayashi:
This PR supports TTS-Transformer in ESPnet2.

- Add `espnet2/tts/transformer.py`
- Remove deprecated options from `espnet2/tts/tacotron2.py`
- Add options available in transformer for tacotron2
- Add test code for epsnet2 tts models.


kan-bayashi:
I confirmed it works. I will merge this PR.

PR #2139, opened 2020-07-02
Support GST in ESPnet2 TTS

kan-bayashi:
This PR adds GST introduced in [Style Tokens: Unsupervised Style Modeling, Control and Transfer in End-to-End Speech Synthesis](https://arxiv.org/pdf/1803.09017.pdf).
We can build both GST-Tacotron2 and GST-Transformer.
This may enable us to build multi-speaker model without pretrained speaker embedding and kaldi.
I already checked GST-Tacotron2 works well in the case of single speaker model.
I will make VCTK recipe and check the performance.

- Support GST
- Support teacher forcing inference


egs2/TEMPLATE/tts1/tts.sh
```diff
@@ -620,18 +620,32 @@ if [ ${stage} -le 7 ] && [ ${stop_stage} -ge 7 ]; then
...
 
+    # If we use teacher forcing or GST embedding in inference,
+    # we need to pass speech as well as text
+    # TODO(kan-bayashi): Make this part more smarter
+    _use_teacher_forcing="$(pyscripts/utils/get_yaml.py "${decode_config}" use_teacher_forcing)"
+    _use_gst="$(pyscripts/utils/get_yaml.py "${train_config}" tts_conf.use_gst)"
+    _use_speech=false
+    if [ "${_use_teacher_forcing,,}" = "true" ] || [ "${_use_gst,,}" = "true" ]; then
+        _use_speech=true
+    fi
```
kan-bayashi, line 647: I think this part is not so good.
Do you have any idea? @kamo-naoyuki 
I want to add additional opts when the specific option is True in config.

egs2/TEMPLATE/tts1/tts.sh
```diff
@@ -620,18 +620,32 @@ if [ ${stage} -le 7 ] && [ ${stop_stage} -ge 7 ]; then
...
+    if [ "${_feats_type}" = fbank ] || [ "${_feats_type}" = stft ]; then
         _opts+="--vocoder_conf n_fft=${n_fft} "
         _opts+="--vocoder_conf n_shift=${n_shift} "
         _opts+="--vocoder_conf win_length=${win_length} "
         _opts+="--vocoder_conf fs=${fs} "
+        _scp=feats.scp
+        _type=kaldi_ark
     fi
-    if [ "${_feats_type}" == fbank ]; then
+    if [ "${_feats_type}" = fbank ]; then
```
kan-bayashi, line 633: Fixed typo.

egs2/TEMPLATE/tts1/tts.sh
```diff
@@ -620,18 +620,32 @@ if [ ${stage} -le 7 ] && [ ${stop_stage} -ge 7 ]; then
     _feats_type="$(<${data_feats}/${train_set}/feats_type)"
 
     # NOTE(kamo): If feats_type=raw, vocoder_conf is unnecessary
-    if [ "${_feats_type}" == fbank ] || [ "${_feats_type}" == stft ]; then
+    _scp=wav.scp
+    _type=sound
+    if [ "${_feats_type}" = fbank ] || [ "${_feats_type}" = stft ]; then
```
kan-bayashi, line 625: Fixed typo

espnet2/tts/gst/style_encoder.py
```diff
@@ -0,0 +1,238 @@
...
+        gru_layers (int, optional): The number of GRU layers in the reference encoder.
+        gru_units (int, optional): The number of GRU units in the reference encoder.
+
+    """
+
+    def __init__(
+        self,
+        idim=80,
+        conv_layers: int = 6,
+        conv_chans_list: List[int] = [32, 32, 64, 64, 128, 128],
```
ShigekiKarita, line 115: The mutable default arg is stateful and gonna be bug. How about tuple or None instead? https://docs.python-guide.org/writing/gotchas/

kan-bayashi: Thanks. Chaneged to Tuple.

kan-bayashi: @ShigekiKarita `Union[Tuple, List] = (...)` is OK?

ShigekiKarita:
How about `Sequence[Int] = (...)`? For example,
https://github.com/espnet/espnet/blob/0fe5068a356f5e48beac25117fa1d6c2ad799452/espnet2/iterators/chunk_iter_factory.py#L41


2020年7月3日(金) 16:53 Tomoki Hayashi <notifications@github.com>:

kan-bayashi:
Maybe finished.
After this PR, I will make VCTK recipe and FastSpeech PRs.

kan-bayashi:
@kamo-naoyuki Could you check `tts.sh` and `tts_inference.py` part?

espnet2/tts/tacotron2.py
```diff
@@ -1,10 +1,13 @@
-# Copyright 2018 Nagoya University (Tomoki Hayashi)
+# -*- coding: utf-8 -*-
```
kamo-naoyuki, line 1: `coding: utf-8` has no effects if python3.
pep8 recommends not writing `coding: utf-8` for now.

egs2/TEMPLATE/tts1/tts.sh
```diff
@@ -620,18 +620,40 @@ if [ ${stage} -le 7 ] && [ ${stop_stage} -ge 7 ]; then
...
+        _use_gst=$(pyscripts/utils/get_yaml.py \
+            "${train_config}" tts_conf.use_gst)
+    fi
+    if [ -n "${decode_config}" ]; then
+        _use_teacher_forcing=$(pyscripts/utils/get_yaml.py \
+            "${decode_config}" use_teacher_forcing)
+    fi
+    if [ "${_use_teacher_forcing,,}" = "true" ] || [ "${_use_gst,,}" = "true" ]; then
+        _use_speech=true
+    fi
```
kamo-naoyuki, line 655: My suggestion:

1. Always give `--data_path_and_name_and_type ${_data}/${_scp},speech,${_type}`
2. Add something like `tts.use_speech` as an attribute or public method to the interface to check `use_speech` option
3. Then

    ```python
    if not tts.use_speech:
        data_path_and_name_type = list(filter(lambda x: x[1] != "speech", data_path_and_name_type))
```

File: espnet2/tts/gst/style_encoder.py [hunk 7]
```diff
@@ -0,0 +1,243 @@
+# -*- coding: utf-8 -*-
```
kamo-naoyuki, line 1: `coding: utf-8` has no effect if python3.
Pep8 recommends not writing the statement now.

egs2/TEMPLATE/tts1/tts.sh
```diff
@@ -651,10 +673,17 @@ if [ ${stage} -le 7 ] && [ ${stop_stage} -ge 7 ]; then
         # shellcheck disable=SC2086
         utils/split_scp.pl "${key_file}" ${split_scps}
 
-        # 2. Submit decoding jobs
+        # 2. Check extra options
+        # NOTE(kan-bayashi): Using opts causes duplicated option errors
+        _ex_opts=
+        if "${_use_speech}"; then
+            _ex_opts+="--allow_variable_data_keys true "
```
kamo-naoyuki, line 680: https://github.com/espnet/espnet/blob/4168d53a9bef5aa3b529c1555a9e4fb423afc438/espnet2/tasks/tts.py#L198-L204

We can add `speech` to this list.

(BTW, possibly I'll  change the system of `--data_path_and_name_and_type`... This style seems too flexible than necessary and it's hard to understand)

espnet2/tts/tacotron2.py
kan-bayashi: OK! I will remove.

egs2/TEMPLATE/tts1/tts.sh
kan-bayashi: OK. I will follow your suggestions!

kan-bayashi:
@r9y9 If you have a time, could you take a look at `espnet2/tts/gst/style_encoder.py`?

r9y9:
Sure, I will look into it when I have some time (probably tomorrow). GST is very nice to have :)

espnet2/tts/gst/style_encoder.py
```diff
@@ -0,0 +1,241 @@
...
+        for i in range(conv_layers):
+            conv_in_chans = 1 if i == 0 else conv_chans_list[i - 1]
+            conv_out_chans = conv_chans_list[i]
+            convs += [
+                torch.nn.Conv2d(
+                    conv_in_chans,
+                    conv_out_chans,
+                    kernel_size=conv_kernel_size,
+                    stride=conv_stride,
+                    padding=padding,
```
r9y9, line 143: I think `bias=False` is better for the conv layer since the next BatchNorm layer has a bias term. It wouldn't cause performance degradation, but it is just redundant.

espnet2/tts/gst/style_encoder.py
```diff
@@ -0,0 +1,241 @@
...
+            Tensor: Reference embedding (B, gru_units)
+
+        """
+        batch_size = speech.size(0)
+        xs = speech.unsqueeze(1)  # (B, 1, Lmax, idim)
+        hs = self.convs(xs).transpose(1, 2)  # (B, Lmax', conv_out_chans, idim')
+        # NOTE(kan-bayashi): We need to care the length?
+        time_length = hs.size(1)
+        hs = (
+            hs.transpose(1, 2).contiguous().view(batch_size, time_length, -1)
```
r9y9, line 180: `hs.transpose(1,2)` is not needed here, right? You've already done it at line 176.

espnet2/tts/gst/style_encoder.py
```diff
@@ -0,0 +1,241 @@
...
+        dropout_rate: float = 0.0,
+    ):
+        """Initilize style token layer module."""
+        assert check_argument_types()
+        super(StyleTokenLayer, self).__init__()
+
+        gst_embs = torch.randn(gst_tokens, gst_token_dim)
+        self.register_parameter("gst_embs", torch.nn.Parameter(gst_embs))
+        self.projection = torch.nn.Linear(ref_embed_dim, gst_token_dim)
+        self.mha = MultiHeadedAttention(gst_heads, gst_token_dim, dropout_rate)
```
r9y9, line 222: I am not fully sure about the dimension of the style token embedding (256 vs. 256//num_heads). According to the paper, 
> When using h attention heads, we set the token embedding size to be 256/h and concatenate the attention outputs, such that the final style embedding size remains the same.

I think the current code would work okay, but if we strictly follow the original paper's description, would it be better to use 256//num_heads dim for the style token embedding? I am wondering like modifying https://github.com/espnet/espnet/blob/6f6452a3891d11b0b0306084dd46a464ddf8e25c/espnet/nets/pytorch_backend/transformer/attention.py#L33-L34 to 
```py
        self.linear_k = nn.Linear(key_dim, n_feat) # 256//num_heads  -> 256
        self.linear_v = nn.Linear(value_dim, n_feat) # 256//num_heads -> 256
```
for example (BUT I am not sure to be honest...) Would be nice to check the TTS quality.

espnet2/tts/gst/style_encoder.py
kan-bayashi: Agree. I will remove it.

espnet2/tts/gst/style_encoder.py
kan-bayashi: OMG 
This is a critical bug 😑

espnet2/tts/gst/style_encoder.py
kan-bayashi: Yeah, maybe you are right.
Since our multihead attention assumes the same dimension, I use embedding size = 256 (256 -> 64*4 -> 256).
Let us check the performance after fixing transpose bug ...

kan-bayashi: I added note about the difference.

kan-bayashi:
> Would be great to see some TTS outputs if you have them already!

The quality is varied significantly among the samples (speaker characteristics are changed significantly even if we provide the same speaker speech). But it may be transpose bug.
Now I re-run the experiments. Let us check the results tomorrow.

kan-bayashi:
I changed to follow the original paper:
```python
      (stl): StyleTokenLayer(
        (mha): MultiHeadedAttention(
          (linear_q): Linear(in_features=128, out_features=512, bias=True)
          (linear_k): Linear(in_features=64, out_features=512, bias=True)
          (linear_v): Linear(in_features=64, out_features=512, bias=True)
          (linear_out): Linear(in_features=512, out_features=512, bias=True)
          (dropout): Dropout(p=0.0, inplace=False)
        )
      )
```

kan-bayashi:
@r9y9 Here is the sample @ 100 epoch + PWG.
https://drive.google.com/drive/folders/1MbjtiO9MjAv-ClFrgBqE2ddAhGoTioq2?usp=sharing

It seems to be working but women voice is dominant.
Sometimes, a man voice appeared. I'm not sure this is correct behavior or not...
I should check the behavior when we manually select the weights.

r9y9:
Well, not sure why one gender is dominant to the other. Do you mean that even we give a male utterance as a reference, an e2e model produces a female voice? It could be possibly due to the imbalance of data. According to the speakers-info.txt in the VCTK corpus, it contains 47 male speakers data and 61 female speakers data. Anyway, it would be better to inspect style embedding weights to check what (interpretable) attributes it leaned.

kan-bayashi:
I found a bug in inference, gst is not added 😂
After fixing the bug, it works reasonably, the speaker characteristics are changed according to a given speech.
(Something like an average speaker? voice was generated before? It is interesting.)

https://drive.google.com/drive/folders/1MbjtiO9MjAv-ClFrgBqE2ddAhGoTioq2?usp=sharing

kan-bayashi:
I will merge after CI.

Issue #2126, opened 2020-06-30
`dev_set` and `eval_sets` in espnet2


sw005320:
I've found some issues in the current design for `dev_set` and `eval_sets` in https://github.com/espnet/espnet/blob/master/egs2/TEMPLATE/asr1/asr.sh

The role of the development set is different depending on as a part of the test set to tune some search parameters/hyperparameters **after recognition** or monitoring the loss **during the training process** (and also tune the optimizer based on this dev score).
So, this role must be separated depending on the case, and actually we often use different sets for both cases.
For example, in the librispeech recipe, we use the combined and pruned development set during training while we use the original dev set (dev_clean and dev_other) during inference. Note that they are slightly different but mostly the same.
Then, the current script tries to recognize both `dev_set` and `eval_sets` in stage 11.
Since the dev sets used in training and inference are different, we need to add the original dev sets to `eval_sets`, which means that we almost recognize the same dev sets twice. Also, adding the dev sets is inconsistent with the name of the variable `eval_sets`.

My suggestion is to first rename them as follows (basically try to avoid to use the name dev and eval in the variable name level).
- `dev_set` -> `val_set` (or `train_dev` like in espnet1)
- `eval_sets` -> `test_sets` (or `recog_set` like in espnet1)

Also, instead of always recognizing both `dev_set` and `eval_sets`, we only recognize `test_sets`, in stage 11 (https://github.com/espnet/espnet/blob/master/egs2/TEMPLATE/asr1/asr.sh#L944) and stage 12 (https://github.com/espnet/espnet/blob/master/egs2/TEMPLATE/asr1/asr.sh#L999). Since it can accept multiple data sets, we can include the original development sets in `test_sets`.
Actually, espnet1 is designed like that to deal with such an issue with some consideration of the above multiple dev set roles.

PR #2142, opened 2020-07-02
fix dev_set/eval_sets issues

sw005320:
This issue is raised in https://github.com/espnet/espnet/issues/2126

- [x] If this is acceptable, I'll modify the other all ASR recipes. I'm currently only applying it to `an4`.
- [x] I keep the LM part as it is since the LM part is currently consistent. 
- [x] I found the same style in the TTS script, and the same issue may happen (I could fix it in the same manner, if needed).
- [x] fix the description in https://github.com/espnet/espnet/blob/master/egs2/TEMPLATE/README.md accordingly


kamo-naoyuki:
Would you apply it for tts.sh? You also need to change the argument in all run.sh.

sw005320:
> Would you apply it for tts.sh? You also need to change the argument in all run.sh.

OK.
(Please check my initial comments. These are already considered ;) but I need your confirmation before moving to these items)

kamo-naoyuki:
> (Please check my initial comments. These are already considered ;) but I need your confirmation before moving to these items)

I missed. LGTM.

sw005320:
I'm testing some of them, and it seems to be no problem.
There are known issues when I checked the scripts

- Babel needs some test
- Commonvoice dev/eval split will be updated
- DIRHA sim (dev) vs real (eval) split is not ideal

sw005320:
I'd like to ask @kamo-naoyuki and @kan-bayashi to check the change of the asr and tts recipe templated parts, respectively.

PR #2144, opened 2020-07-03
Support Conformer model.

pengchengguo:
Implementation of Google's paper "Conformer: Convolution-augmented Transformer for Speech Recognition" (https://arxiv.org/abs/2005.08100).


sw005320:
Many thanks, @pengchengguo!!!

@ShigekiKarita and @kan-bayashi, what do you think?
I like @pengchengguo's implementation of making a conformer module as a separate module from the transformer module. Another option is to put them under the transformer module like lightweight conv. Which direction we should do?

@pengchengguo, could you at least try to remove duplications in the function levels?
For example, you can import https://github.com/espnet/espnet/blob/master/espnet/nets/pytorch_backend/transformer/positionwise_feed_forward.py instead of putting the copy under the conformer module. You could do it for many of your functions.

Anyway, I believe your implementation can be minimized with some preparation code for this new conformer module, and conformer own functions (conformer forward including Makaron layers and conv1d, and Swish Activation, relative position embedding). You can import all the others from the transformer module.

pengchengguo:
Ok, I will remove those duplicate functions and also post some results when I finish training.

espnet/nets/pytorch_backend/conformer/attention.py
```diff
@@ -0,0 +1,190 @@
...
+        self.linear_q = nn.Linear(n_feat, n_feat)
+        self.linear_k = nn.Linear(n_feat, n_feat)
+        self.linear_v = nn.Linear(n_feat, n_feat)
+        # linear transformation for positional ecoding
+        self.linear_pos = nn.Linear(n_feat, n_feat, bias=False)
+        # these two hyperparameters are used in matrix c and matrix d
+        # as described in https://arxiv.org/abs/1901.02860 Section 3.3
+        self.pos_bias_u = nn.Parameter(torch.Tensor(self.h, self.d_k))
+        self.pos_bias_v = nn.Parameter(torch.Tensor(self.h, self.d_k))
+        torch.nn.init.kaiming_uniform_(self.pos_bias_u, a=math.sqrt(5))
```
hirofumi0810, line 107: What is the reference for this part?

espnet/nets/pytorch_backend/conformer/embedding.py
```diff
@@ -0,0 +1,136 @@
...
+        if self.rel_pos:
+            x = x * self.xscale
+            pos_emb = self.pe[:, : x.size(1)]
+            return self.dropout(x), self.dropout(pos_emb)
+        else:
+            x = x * self.xscale + self.pe[:, : x.size(1)]
+            return self.dropout(x)
+
+
+class ScaledPositionalEncoding(PositionalEncoding):
```
hirofumi0810, line 102: Is this class necessary?

espnet/nets/pytorch_backend/conformer/encoder_layer.py
```diff
@@ -0,0 +1,155 @@
...
+            x = residual + self.dropout(self.conv_block(x))
+            if not self.normalize_before:
+                x = self.norm3(x)
+
+        # feed forward module
+        residual = x
+        if self.normalize_before:
+            x = self.norm4(x)
+        if self.macaron_style:
+            x = residual + 0.5 * self.dropout(self.feed_forward(x))
```
hirofumi0810, line 140: IMPORTANT: this should be self.feed_forward2.

espnet/nets/pytorch_backend/conformer/encoder_layer.py
```diff
@@ -0,0 +1,155 @@
...
+        else:
+            x = residual + self.dropout(self.feed_forward(x))
```
hirofumi0810, line 142: ditto

espnet/nets/pytorch_backend/conformer/encoder_layer.py
pengchengguo: Yes, it should be self.feed_forward2. Before initing the PR, I just shared the parameters of FNN modules and forgot to modify this part. I will change this part now.

espnet/nets/pytorch_backend/conformer/embedding.py
pengchengguo: Ok. We don't use this class in Conformer. I modified the Conformer's embedding.py file based on Transformer's  embedding.py and neglected the unused class. I will delete this part.

espnet/nets/pytorch_backend/conformer/attention.py
pengchengguo: The reference is the Transformer-XL paper. They introduce two learnable bias named u and v for computing rel. pos ecoding. I named these two tensors as described in the paper. Since `bias_u` and `bias_v` are not easy to understand, I just cite the paper here. Should we change another name? Thanks so much for your correction and advice.

hirofumi0810: My question is about initialization. Original Transformer-XL initialized parameters by sampling values from normal distribution with std 0.02. Where did you find the reference to initialize these bias parameters by sampling from kaiming_uniform?

espnet/nets/pytorch_backend/conformer/convolution.py
```diff
@@ -0,0 +1,86 @@
...
+        kernel_size,
+        stride=1,
+        padding=0,
+        dilation=1,
+        bias=False,
+        activation=nn.ReLU(),
+    ):
+        """Construct an ConvolutionBlock object."""
+        super(ConvolutionBlock, self).__init__()
+        self.pad_left = nn.ConstantPad1d((kernel_size - 1, 0), 0)
```
hirofumi0810, line 33: Why is this padding asymmetric?

espnet/nets/pytorch_backend/conformer/encoder_layer.py
hirofumi0810: OK. Or It might be better to have such an option to share FFN parameters.

espnet/nets/pytorch_backend/conformer/attention.py
pengchengguo: I saw the initialization of the learnable bias is `self.r_w_bias = nn.Parameter(torch.Tensor(self.n_head, self.d_head))` in the official implementation of transformer-xl (https://github.com/hirofumi0810/neural_sp/blob/master/neural_sp/models/seq2seq/encoders/transformer.py#L165). I found the bias values of this method can be very large, such as `6.0630e+33, 4.5836e-41`. So I used the same initialization function as `torch.nn.Linear.reset_parameters()`. I am not sure about the influence between different initialization functions, maybe I should change it to `torch.nn.init.xavier_normal_`.

espnet/nets/pytorch_backend/conformer/encoder_layer.py
pengchengguo: Yes, if `marcon_style = False`, we only use `self.feed_forward2`.

espnet/nets/pytorch_backend/conformer/convolution.py
pengchengguo: In the Gated CNN LM paper (https://arxiv.org/pdf/1612.08083.pdf), they shift the convolutional inputs to prevent the kernels from seeing the future context. I think we shouldn't follow this setup since we can see future information in the speech scenario. I will modify this, thanks so much.

hirofumi0810: For streaming experiments in future research, how about making it a option?
We can use symmetric padding as default in the offline experiments.
But it might be redundant because Transformer encoder generally adopts blockwise encoding instead of framewise encoding.

espnet/nets/pytorch_backend/conformer/positionwise_feed_forward.py
```diff
@@ -0,0 +1,31 @@
+#!/usr/bin/env python3
```
hirofumi0810, line 1: This file is still redundant.

espnet/nets/pytorch_backend/conformer/subsampling.py
```diff
@@ -0,0 +1,55 @@
+#!/usr/bin/env python3
```
hirofumi0810, line 1: This file is still redundant.

espnet/nets/pytorch_backend/e2e_asr_conformer.py
```diff
@@ -0,0 +1,601 @@
+# Copyright 2020 Johns Hopkins University (Shinji Watanabe)
```
hirofumi0810, line 1: Maybe we can override the Transformer ASR class because the only difference should be the encoder. What do you think about? @ShigekiKarita @sw005320

sw005320: I also feel that override would be a better solution.
I agree that the code is still redundant and it's difficult to maintain it.

@pengchengguo, could you try that?
Sorry to ask many things, but this makes your code more useful ant many people would have the benefits.

sw005320:
We should also have some unit/integration tests.

espnet/nets/pytorch_backend/conformer/positionwise_feed_forward.py
pengchengguo: Thanks so much for your help. I keep this file because I make the activation function as an optional item, such as ReLU() and Swish() etc. Should I merge this difference to the Transformer's positionwise_feed_forward.py or just keep this file now?

espnet/nets/pytorch_backend/conformer/subsampling.py
pengchengguo: Same as above, include some differences in activation functions and whether to use rel_pos.

espnet/nets/pytorch_backend/e2e_asr_conformer.py
pengchengguo: I agree. I will do this.

espnet/nets/pytorch_backend/conformer/positionwise_feed_forward.py
hirofumi0810: Since most lines are the same, you can slightly modify the original file.

espnet/nets/pytorch_backend/conformer/subsampling.py
hirofumi0810: In the original paper, the Swish activation is not placed in frontend CNN layers. Also, relative positional encoding shouldn't appear here.

espnet/nets/pytorch_backend/conformer/attention.py
```diff
@@ -0,0 +1,190 @@
...
+
+    :param int n_head: the number of head s
+    :param int n_feat: the number of features
+    :param float dropout_rate: dropout rate
+
+    """
+
+    def __init__(self, n_head, n_feat, dropout_rate):
+        """Construct an MultiHeadedAttention object."""
+        super(MultiHeadedAttention, self).__init__()
```
ShigekiKarita, line 28: this init looks completely same to https://github.com/espnet/espnet/blob/cc75da63b4fb3975e01b6ea3e3ba7580a0c99384/espnet/nets/pytorch_backend/transformer/attention.py#L16

Can this class inherit it and reuse the init?

espnet/nets/pytorch_backend/conformer/attention.py
```diff
@@ -0,0 +1,190 @@
...
+"""Multi-Head Attention layer definition."""
+
+import math
+
+import numpy
+import torch
+from torch import nn
+
+
+class MultiHeadedAttention(nn.Module):
```
ShigekiKarita, line 17: This class looks same to https://github.com/espnet/espnet/blob/cc75da63b4fb3975e01b6ea3e3ba7580a0c99384/espnet/nets/pytorch_backend/transformer/attention.py#L16

Can you delete this class and reuse that class?

espnet/nets/pytorch_backend/conformer/attention.py
```diff
@@ -0,0 +1,190 @@
...
+        # these two hyperparameters are used in matrix c and matrix d
+        # as described in https://arxiv.org/abs/1901.02860 Section 3.3
+        self.pos_bias_u = nn.Parameter(torch.Tensor(self.h, self.d_k))
+        self.pos_bias_v = nn.Parameter(torch.Tensor(self.h, self.d_k))
+        torch.nn.init.xavier_uniform_(self.pos_bias_u)
+        torch.nn.init.xavier_uniform_(self.pos_bias_v)
+
+        self.linear_out = nn.Linear(n_feat, n_feat)
+        self.attn = None
+        self.dropout = nn.Dropout(p=dropout_rate)
```
ShigekiKarita, line 112: Can you reuse the MultiHeadedAttention to init the common fields like
```python
class RelPositionMultiHeadedAttention(MultiHeadedAttention):
    def __init__(self, n_head, n_feat, dropout_rate):
        super().__init__(n_head, n_feat, dropout_rate)
        # these two hyperparameters are used in matrix c and matrix d
        # as described in https://arxiv.org/abs/1901.02860 Section 3.3
        self.linear_pos = nn.Linear(n_feat, n_feat, bias=False)
        self.pos_bias_u = nn.Parameter(torch.Tensor(self.h, self.d_k))
        self.pos_bias_v = nn.Parameter(torch.Tensor(self.h, self.d_k))
        torch.nn.init.xavier_uniform_(self.pos_bias_u)
        torch.nn.init.xavier_uniform_(self.pos_bias_v)
```

espnet/nets/pytorch_backend/conformer/attention.py
```diff
@@ -0,0 +1,190 @@
...
+        p_attn = self.dropout(self.attn)
+        x = torch.matmul(p_attn, v)  # (batch, head, time1, d_k)
+        x = (
+            x.transpose(1, 2).contiguous().view(n_batch, -1, self.h * self.d_k)
+        )  # (batch, time1, d_model)
+        return self.linear_out(x)  # (batch, time1, d_model)
+
+
+class RelPositionMultiHeadedAttention(nn.Module):
+    """Multi-Head Attention layer with relative position encoding.
```
ShigekiKarita, line 83: Can you add paper link to the rel potion mha?

espnet/nets/pytorch_backend/conformer/attention.py
```diff
@@ -0,0 +1,190 @@
...
+            )  # (batch, head, time1, time2)
+        else:
+            self.attn = torch.softmax(scores, dim=-1)  # (batch, head, time1, time2)
+
+        p_attn = self.dropout(self.attn)
+        x = torch.matmul(p_attn, v)  # (batch, head, time1, d_k)
+        x = (
+            x.transpose(1, 2).contiguous().view(n_batch, -1, self.h * self.d_k)
+        )  # (batch, time1, d_model)
+        return self.linear_out(x)  # (batch, time1, d_model)
```
ShigekiKarita, line 190: These lines (+ q,k,v forward) are same to the last parts of MultiHeadedAttention.forward. Can you split those into common methods like
```python
# in transformer/attention.py
class MultiHeadedAttention(nn.Module):
    def forward_qkv(self, query, key, value):
        n_batch = query.size(0)
        q = self.linear_q(query).view(n_batch, -1, self.h, self.d_k)
        k = self.linear_k(key).view(n_batch, -1, self.h, self.d_k)
        v = self.linear_v(value).view(n_batch, -1, self.h, self.d_k)
        q = q.transpose(1, 2)  # (batch, head, time1, d_k)
        k = k.transpose(1, 2)  # (batch, head, time2, d_k)
        v = v.transpose(1, 2)  # (batch, head, time2, d_k)
        return q, k, v

    def forward_attention(self, v, scores, mask):
        if mask is not None:
            mask = mask.unsqueeze(1).eq(0)  # (batch, 1, time1, time2)
            min_value = float(
                numpy.finfo(torch.tensor(0, dtype=scores.dtype).numpy().dtype).min
            )
            scores = scores.masked_fill(mask, min_value)
            self.attn = torch.softmax(scores, dim=-1).masked_fill(
                mask, 0.0
            )  # (batch, head, time1, time2)
        else:
            self.attn = torch.softmax(scores, dim=-1)  # (batch, head, time1, time2)

        p_attn = self.dropout(self.attn)
        x = torch.matmul(p_attn, v)  # (batch, head, time1, d_k)
        x = (
            x.transpose(1, 2).contiguous().view(n_batch, -1, self.h * self.d_k)
        )  # (batch, time1, d_model)
        return self.linear_out(x)  # (batch, time1, d_model)

    def forward(self, query, key, value, mask):
        q, k, v = self.forward_qkv(query, key, value)
        scores = torch.matmul(q, k.transpose(-2, -1)) / math.sqrt(self.d_k)
        return self.forward_attention(v, scores, mask)

# in conformer/attention.py
class RelPositionMultiHeadedAttention(MultiHeadedAttention):

     def forward(self, query, key, value, pos_emb, mask):
        q, k, v = self.forward_qkv(query, key, value)

        n_batch_pos = pos_emb.size(0)
        p = self.linear_pos(pos_emb).view(n_batch_pos, -1, self.h, self.d_k)
        p = p.transpose(1, 2)  # (batch, head, time1, d_k)
        # (batch, head, time1, d_k)
        q_with_bias_u = (q + self.pos_bias_u).transpose(1, 2)
        # (batch, head, time1, d_k)
        q_with_bias_v = (q + self.pos_bias_v).transpose(1, 2)

        # compute attention score
        # first compute matrix a and matrix c
        # as described in https://arxiv.org/abs/1901.02860 Section 3.3
        # (batch, head, time1, time2)
        matrix_ac = torch.matmul(q_with_bias_u, k.transpose(-2, -1))
        # compute matrix b and matrix d
        # (batch, head, time1, time2)
        matrix_bd = torch.matmul(q_with_bias_v, p.transpose(-2, -1))
        matrix_bd = self.rel_shift(matrix_bd)
        # (batch, head, time1, time2)
        scores = (matrix_ac + matrix_bd) / math.sqrt(self.d_k)
        return self.forward_attention(v, scores, mask)
```

espnet/nets/pytorch_backend/conformer/embedding.py
```diff
@@ -0,0 +1,99 @@
...
+def _pre_hook(
+    state_dict,
+    prefix,
+    local_metadata,
+    strict,
+    missing_keys,
+    unexpected_keys,
+    error_msgs,
+):
+    """Perform pre-hook in load_state_dict for backward compatibility.
```
ShigekiKarita, line 24: This function should be deleted because we don't need backward compatibility for a new module.

espnet/nets/pytorch_backend/conformer/encoder.py
```diff
@@ -0,0 +1,241 @@
...
+from espnet.nets.pytorch_backend.conformer.subsampling import Conv2dSubsampling
+from espnet.nets.pytorch_backend.nets_utils import rename_state_dict
+from espnet.nets.pytorch_backend.transducer.vgg import VGG2L
+from espnet.nets.pytorch_backend.transformer.layer_norm import LayerNorm
+from espnet.nets.pytorch_backend.transformer.multi_layer_conv import Conv1dLinear
+from espnet.nets.pytorch_backend.transformer.multi_layer_conv import MultiLayeredConv1d
+from espnet.nets.pytorch_backend.transformer.repeat import repeat
+
+
+def _pre_hook(
```
ShigekiKarita, line 32: ditto (no backward compatibility is required)

espnet/nets/pytorch_backend/conformer/embedding.py
```diff
@@ -0,0 +1,99 @@
...
+                return
+        pe = torch.zeros(x.size(1), self.d_model)
+        # Does it matter to input position in reversed order
+        # when using relative position encoding?
+        if self.rel_pos:
+            position = torch.arange(
+                x.size(1) - 1, -1, -1.0, dtype=torch.float32
+            ).unsqueeze(1)
+        else:
+            position = torch.arange(0, x.size(1), dtype=torch.float32).unsqueeze(1)
```
ShigekiKarita, line 72: Can you simply reuse `PositionalEncoding` in transformer/embedding.py and add this option? I think the name should be `reverse` rather than `rel_pos`.

espnet/nets/pytorch_backend/conformer/embedding.py
```diff
@@ -0,0 +1,99 @@
...
+
+        Args:
+            x (torch.Tensor): Input. Its shape is (batch, time, ...)
+
+        Returns:
+            torch.Tensor: Encoded tensor. Its shape is (batch, time, ...)
+
+        """
+        self.extend_pe(x)
+        if self.rel_pos:
```
ShigekiKarita, line 93: I prefer
```python
from espnet.nets.pytorch_backend.transformer.embedding import PositionalEncoding

class RelativePositionalEncoding(PositionalEncoding):
     def __init__(self, d_model, dropout_rate, max_len=5000):
        super().__init__(d_model, dropout_rate, max_len, reverse=True)

    def forward(self, x):
        self.extend_pe(x)
        x = x * self.xscale
        pos_emb = self.pe[:, : x.size(1)]
        return self.dropout(x), self.dropout(pos_emb)
```

espnet/nets/pytorch_backend/conformer/encoder.py
```diff
@@ -0,0 +1,241 @@
...
+    unexpected_keys,
+    error_msgs,
+):
+    # https://github.com/espnet/espnet/commit/21d70286c354c66c0350e65dc098d2ee236faccc#diff-bffb1396f038b317b2b64dd96e6d3563
+    rename_state_dict(prefix + "input_layer.", prefix + "embed.", state_dict)
+    # https://github.com/espnet/espnet/commit/3d422f6de8d4f03673b89e1caef698745ec749ea#diff-bffb1396f038b317b2b64dd96e6d3563
+    rename_state_dict(prefix + "norm.", prefix + "after_norm.", state_dict)
+
+
+class Encoder(torch.nn.Module):
```
ShigekiKarita, line 47: Can you refactor this class as well as attention.py and embedding.py?

espnet/nets/pytorch_backend/conformer/encoder_layer.py
```diff
@@ -0,0 +1,156 @@
...
+            x = self.norm4(x)
+
+        if self.conv_block is not None:
+            x = self.norm5(x)
+
+        if cache is not None:
+            x = torch.cat([cache, x], dim=1)
+
+        if self.rel_pos:
+            return tuple([x, pos_emb]), mask
```
ShigekiKarita, line 154:
```suggestion
            return (x, pos_emb), mask
```

espnet/nets/pytorch_backend/conformer/encoder_layer.py
```diff
@@ -0,0 +1,156 @@
...
+        super(EncoderLayer, self).__init__()
+        self.self_attn = self_attn
+        self.feed_forward = feed_forward
+        self.feed_forward2 = feed_forward2
+        self.norm1 = LayerNorm(size)  # for the first FNN module
+        self.norm2 = LayerNorm(size)  # for MHA module
+        self.norm3 = LayerNorm(size)  # for CNN module
+        self.norm4 = LayerNorm(size)  # for the second FNN module
+        self.norm5 = LayerNorm(size)
+        self.conv_block = conv_block
```
ShigekiKarita, line 66: Instead of norm1 ... 5 with comments, I prefer
```python
self.norm_ff1 = LayerNorm(size)
self.norm_ff2 = LayerNorm(size)
self.norm_mha = LayerNorm(size)
if self.conv_block is not None:
     self.norm_cnn1 = LayerNorm(size)
     self.norm_cnn2 = LayerNorm(size)
```

ShigekiKarita:
"+1,531 -3 lines" is hard to review at one time! Can you add RESULTS.md or share some WIP values here?

espnet/nets/pytorch_backend/conformer/attention.py
pengchengguo: Sure, I am moving convolutional block as an option to transformer's encoder now. I will merge conformer related codes into transformer. Thanks so much for your suggestions.

espnet/nets/pytorch_backend/conformer/embedding.py
pengchengguo: In the forward(), we also use `self.rel_pos` to identify different outputs. For using relative positional encoding, we want return both `x` and `pos_emb`, while for normal positional encoding, we only return `x+pos_emb`. I think using `reverse` may be a little ambiguous in the forward() part. Should I change the option to `use_rel_pos`?

espnet/nets/pytorch_backend/conformer/embedding.py
pengchengguo: Thanks so much. If I merge conformer as an option encoder to transformer, it's better to modify `class PositionalEncoding` to support relative positional encoding or implement another `class RelativePositionalEncoding`.

espnet/nets/pytorch_backend/conformer/encoder_layer.py
pengchengguo: Thanks so much. I think it's better to use `self.norm_ff1` etc. I will modify these name.

espnet/nets/pytorch_backend/conformer/attention.py
pengchengguo: Yes, I will do it.

pengchengguo:
Hi @sw005320, @hirofumi0810, @ShigekiKarita,
Thanks for your help. I moved Conformer as an optional encoder type of Transformer and modified some codes following your suggestions. I don't finish the integration test file now. Should I just create a new `test_e2e_asr_conformer.py` or just modify the `test_e2e_asr_transformer.py`? Besides, I found the PR has conflicts with the master branch.

espnet/nets/pytorch_backend/nets_utils.py
```diff
@@ -392,7 +392,7 @@ def get_subsample(train_args, mode, arch):
     Returns:
         np.ndarray / List[np.ndarray]: subsampling factors.
     """
-    if arch == "transformer":
+    if arch == "transformer" or arch == "contextnet" or arch == "conformer":
```
sw005320, line 395: Please remove `contextnet`

espnet/nets/pytorch_backend/transformer/encoder_layer.py
```diff
@@ -36,34 +42,62 @@ def __init__(
...
         self.feed_forward = feed_forward
-        self.norm1 = LayerNorm(size)
-        self.norm2 = LayerNorm(size)
+        self.ff_scale = 1.0
+        self.conv_module = conv_module
+        self.macaron_style = macaron_style
+        self.norm_ff = LayerNorm(size)  # for the FNN module
+        self.norm_mha = LayerNorm(size)  # for the MHA module
+        if self.macaron_style:
+            self.feed_forward_macaron = copy.deepcopy(feed_forward)
```
sw005320, line 61: why did you use deep copy instead of giving a pointer of `feed_forward` (this means sharing the parameters and it would be wrong though) or newly introduce the liner layer?

espnet/nets/pytorch_backend/transformer/encoder_layer.py
```diff
@@ -6,6 +6,7 @@
```
sw005320, line 6: I feel that this part is very risky as we have a lot of "if", and might introduce some bugs. OK for now, but we could consider to refactor this and use the inheritance from transformer instead (as @hirofumi0810 and @ShigekiKarita suggested) in the later PR.

espnet/nets/pytorch_backend/transformer/subsampling.py
```diff
@@ -8,19 +8,18 @@
...
 
     :param int idim: input dim
     :param int odim: output dim
     :param flaot dropout_rate: dropout rate
+    :param nn.Module pos_enc_class: positional encoding layer
 
     """
 
-    def __init__(self, idim, odim, dropout_rate):
+    def __init__(self, idim, odim, dropout_rate, pos_enc_class):
```
sw005320, line 22: Does this extension keep the compatibility? (Can we use the old model?)

espnet/nets/pytorch_backend/transformer/subsampling.py
```diff
@@ -41,10 +39,13 @@ def forward(self, x, x_mask):
         :param torch.Tensor x_mask: input mask
         :return: subsampled x and mask
         :rtype Tuple[torch.Tensor, torch.Tensor]
+               or Tuple[Tuple[torch.Tensor, torch.Tensor], torch.Tensor]
         """
         x = x.unsqueeze(1)  # (b, c, t, f)
         x = self.conv(x)
         b, c, t, f = x.size()
+        # if RelPositionalEncoding, x: Tuple[torch.Tensor, torch.Tensor]
```
sw005320, line 47: Can you remove unrelated comments or if you leave these lines please explain more why you put these comments.

sw005320:
Please also add your conformer result to `RESULTS.md`.

espnet/nets/pytorch_backend/transformer/encoder_layer.py
pengchengguo: The previous implementation of Conformer construct encoder likes this:
```
self.encoders = repeat(
    num_blocks,
    lambda: EncoderLayer(
        attention_dim,
        attention_layer(attention_heads, attention_dim, attention_dropout_rate),
        positionwise_layer(*positionwise_layer_args),
        positionwise_layer(*positionwise_layer_args),
        convolution_layer(*convolution_layer_args) if use_cnn_module else None,
        dropout_rate,
        normalize_before,
        concat_after,
        rel_pos,
        macaron_style,
    ),
)
```
However, Transformer doesn't need another feed-forward layer. So I have to define another layer inside the EncoderLayer(). I think deepcopy will create a new object.

espnet/nets/pytorch_backend/transformer/encoder_layer.py
pengchengguo: Now we are exploring the improvement obtained by each part, such as `macaron style`, `cnn_module`. I think for a totally Conformer implement, we can remove the `if` of `macaron style`.

espnet/nets/pytorch_backend/transformer/subsampling.py
pengchengguo: Actually, I want to discuss this with you in our group meeting. I added this argument because for relative position encoding, we may need to input a position vector with reverse order. @ShigekiKarita suggest I implement another class `RelPositionalEncoding` inheritance the class `PositionalEncoding`. I didn't found a better way to change the `pos_enc_class` inside the class `Conv2dSubsampling` and passed a pos enc entity to `Conv2dSubsampling` provides options for `PositionalEncoding`, `RelPositionalEncoding` or a CNN-based pos enc which shows effcient in @hiratake55 's toolkit.

espnet/nets/pytorch_backend/transformer/subsampling.py
pengchengguo: Ok, I will remove these comments.

espnet/nets/pytorch_backend/nets_utils.py
pengchengguo: Yes, I will remove it.

espnet/nets/pytorch_backend/transformer/encoder_layer.py
sw005320: I see, but why don’t you use `torch.nn.Linear` to create an object?
You can obtain the dimension information from `feed_forward`.
I think deepcopy would be OK, but this irregular init method is not consistent with the other ways, and we might introduce some bugs.

pengchengguo: We didn't pass these arguments `attention_dim, linear_units, dropout_rate` to `EncoderLayer()`. Maybe I can use the same way as the convolution layer, like this
```
self.encoders = repeat(
    num_blocks,
    lambda: EncoderLayer(
        attention_dim,
        encoder_attn_layer(*encoder_attn_layer_args),
        positionwise_layer(*positionwise_layer_args),
        positionwise_layer(*positionwise_layer_args) if use_cnn_module else None,  # for macaron style
        convolution_layer(*convolution_layer_args) if use_cnn_module else None,
        dropout_rate,
        normalize_before,
        concat_after,
        macaron_style,
    ),
)
```

espnet/nets/pytorch_backend/transformer/encoder_layer.py
sw005320: Buggy here does not mean your conformer implementation.
This complicated `if` logic may break existing transformer implementations if someone else (not you) touch this part, which we should avoid.

Again, I think we should have `conformer` module inherited from the transformer module instead of modifying the transformer module. This can avoid complicated `if` logics and duplicated code, but we can do such refactoring in the later PR.

espnet/nets/pytorch_backend/transformer/encoder_layer.py
pengchengguo: Besides, the `positionwise_layer` contains 2 linear layers and a dropout layer. Construct a new `positionwise_layer ` will need some redundant codes.

sw005320:
> Hi @sw005320, @hirofumi0810, @ShigekiKarita,
> Thanks for your help. I moved Conformer as an optional encoder type of Transformer and modified some codes following your suggestions. I don't finish the integration test file now. Should I just create a new `test_e2e_asr_conformer.py` or just modify the `test_e2e_asr_transformer.py`? Besides, I found the PR has conflicts with the master branch.

Sorry that I did not notice this question. The current way of having `test_e2e_asr_transformer.py` is fine, but I want to keep the compatibility of `test_e2e_asr_conformer.py`, so please make sure that the original transformer is working correctly without any changes.

espnet/nets/pytorch_backend/transformer/encoder_layer.py
pengchengguo: Sorry, I found I misunderstand your and @hirofumi0810's suggestion. You mean to override the Transformer class, but I merged Conformer into Transformer.

pengchengguo:
Hi @sw005320, @hirofumi0810, @ShigekiKarita,
I merged Conformer as an encoder type of Transformer. Should I change the Conformer as a separate module with most of the classes import from Transformer? After merging Conformer and Transformer, we can't load previously trained models because of the mis-match of parameter keys. Maybe implement  Conformer as a separate module is better. BTW, here are some results of Conformer. I didn't finish the training.

| Model |  Test CER (%) |
| - | - |
| Transformer | 6.5 (avg ep41~ep50)|
| Conformer | 5.9 (avg ep18~ep27) |
| - CNN | 6.3 (avg ep17~ep26) |

sw005320:
> Hi @sw005320, @hirofumi0810, @ShigekiKarita,
> I merged Conformer as an encoder type of Transformer. Should I change the Conformer as a separate module with most of the classes import from Transformer? After merging Conformer and Transformer, we can't load previously trained models because of the mis-match of parameter keys. Maybe implement Conformer as a separate module is better. BTW, here are some results of Conformer on Aishell1 corpus. I didn't finish the training.

Year, the transformer pre-trained model is very important, so if it is possible, I want you to make a conformer module instead of putting it inside the transformer module (this is what I mentioned in the first comment of this PR. I should have given you more detailed instructions.).
Can you do that?

> Model	Test CER (%)
> Transformer	6.5 (avg ep41~ep50)
> Conformer	5.9 (avg ep18~ep27)
> - CNN	6.3 (avg ep17~ep26)

Great results!!!

pengchengguo:
Ok, I will try to finish it tomorrow. Thanks so much for your suggestions.

espnet/nets/pytorch_backend/transformer/convolution.py
```diff
@@ -0,0 +1,73 @@
...
+
+    """
+
+    def __init__(self, channels, kernel_size, bias=True):
+        """Construct an ConvolutionModule object."""
+        super(ConvolutionModule, self).__init__()
+        # kernerl_size should be a odd number for 'SAME' padding
+        assert (kernel_size - 1) % 2 == 0
+
+        self.pointwise_cov1 = nn.Conv1d(
```
hirofumi0810, line 28: self.pointwise_conv1?

espnet/nets/pytorch_backend/transformer/convolution.py
```diff
@@ -0,0 +1,73 @@
...
+            channels,
+            channels,
+            kernel_size,
+            stride=1,
+            padding=(kernel_size - 1) // 2,
+            groups=channels,
+            bias=bias,
+        )
+        self.norm = nn.BatchNorm1d(channels)
+        self.pointwise_cov2 = nn.Conv1d(
```
hirofumi0810, line 41: self.pointwise_conv2?

espnet/nets/pytorch_backend/transformer/convolution.py
```diff
@@ -0,0 +1,73 @@
...
+
+        # GLU mechanism
+        x = self.pointwise_cov1(x)  # (batch, 2*channel, dim)
+        x = nn.functional.glu(x, dim=1)  # (batch, channel, dim)
+
+        # 1D Depthwise Conv
+        x = self.depthwise_conv(x)
+        x = self.activation(self.norm(x))
+
+        x = self.pointwise_cov2(x)
```
hirofumi0810, line 63: self.pointwise_conv2?

espnet/nets/pytorch_backend/transformer/convolution.py
```diff
@@ -0,0 +1,73 @@
...
+        """Compute covolution module.
+
+        :param torch.Tensor x: (batch, time, size)
+        :return torch.Tensor: convoluted `value` (batch, time, d_model)
+        """
+        # exchange the temporal dimension and the feature dimension
+        x = x.transpose(1, 2)
+
+        # GLU mechanism
+        x = self.pointwise_cov1(x)  # (batch, 2*channel, dim)
```
hirofumi0810, line 56: self.pointwise_conv1?

espnet/nets/pytorch_backend/transformer/encoder_layer.py
```diff
@@ -73,22 +107,43 @@ def forward(self, x, mask, cache=None):
...
-            x = self.norm2(x)
-        x = residual + self.dropout(self.feed_forward(x))
+            x = self.norm_ff(x)
+        x = residual + self.ff_scale * self.dropout(self.feed_forward(x))
         if not self.normalize_before:
-            x = self.norm2(x)
+            x = self.norm_ff(x)
+
+        if self.conv_module is not None:
+            x = self.norm_final(x)
```
hirofumi0810, line 141: Is this necessary? (maybe redundant)

espnet/nets/pytorch_backend/transformer/subsampling.py
hirofumi0810: @sw005320 I mostly agree with @pengchengguo. Regarding compatibility, this PR controls sinusoidal and relative positional encoding with return values. I personally suggest that we add sinusoidal positional encoding to inputs outside `Conv2dSubsampling` as done in `RelPositionalEncoding` in order to make them simpler.

espnet/nets/pytorch_backend/transformer/attention.py
```diff
@@ -71,8 +79,107 @@ def forward(self, query, key, value, mask):
...
+        x_padded = x_padded.view(*x.size()[:2], x.size(3) + 1, x.size(2))
+        x = x_padded[:, :, 1:].view_as(x)
+
+        if zero_triu:
+            ones = torch.ones((x.size(2), x.size(3)))
+            x = x * torch.tril(ones, x.size(3) - x.size(2))[None, None, :, :]
+
+        return x
+
+    def forward(self, query, key, value, pos_emb, mask):
```
ShigekiKarita, line 146: Is `pos_emb` constant for any layers, right? If so, I want to make this module as a drop-in replacement for MultiHeadedAttention. It can minimize the diff from transformer. How about this?
```python
class RelPositionMultiHeadedAttention(MultiHeadedAttention):
    def __init__(self, n_head, n_feat, dropout_rate, pos_enc):
        ...
        self.pos_enc = pos_enc

    def forward(self, query, key, value, mask):
        pos_emb = self.pos_enc(query)
        ...
```
its overhead is only dropout in `self.pos_enc.forward` because the embedding is cached in advance.

espnet/nets/pytorch_backend/transformer/encoder_layer.py
pengchengguo: As described In the Conformer paper Sec 2.4, they use LayrNrom not only in each module, but also at the end of the block. So I think there are five LayerNorms totally.

espnet/nets/pytorch_backend/transformer/convolution.py
pengchengguo: Sorry, I didn't get the question here. Could you please give me more detailed suggestions?

espnet/nets/pytorch_backend/transformer/attention.py
pengchengguo: Yes, `pos_emb` is constant for each layer. Move `pos_enc` as an attribute of class `RelPositionMultiHeadedAttention` can really reduce the change of transformer, but the `Conv2dSubsampling` also has `PositionalEncoding`.

espnet/nets/pytorch_backend/transformer/encoder_layer.py
hirofumi0810: I checked the paper again. Yes, it seems true. But if we use pre-norm, the first layer norm (self.norm_ff_macaron()) in the next layer corresponds to the self.norm_final() in the current layer.
Check the original implementation https://github.com/zhuohan123/macaron-net/blob/master/translation/fairseq/models/transformer.py#L595.
We need some discussions.

pengchengguo: Yes, I also feel a little confused about the `self.norm_final`. I think in the conformer paper, the final `LayerNorm` is different from the `self.norm_ff_macaron` in the next block, also described in eq.(1). However, in fairseq's implementation, I think they didn't use an additional `LayerNorm` for the final output. They only use `LayerNorm` before or after each module. I am not sure whether `self.norm_final` is necessary for Conformer. I can do some experiments latter.

espnet/nets/pytorch_backend/transformer/convolution.py
iiSeymour: You are missing the `n` in `conv`.

pengchengguo:
Hi @sw005320, @hirofumi0810, @ShigekiKarita,

I re-organized the implementation of conformer. Here are some changes:

- As @ShigekiKarita suggested, I split `forward()` function in `transformer/attention.py` into three functions, `forward_qkv()`, `forword_attention()` and `foward()`.
- Add a new class `RelPositionMultiHeadedAttention(MultiHeadedAttention)` in transformer's directory.
- Add a new class `RelPositionalEncoding(PositionalEncoding)` in transformer's directory.
- Add `e2e_asr_conformer.py` which is inherited `e2e_asr_transformer.py`.
- Add `encoder.py`, `encoder_layer.py`, and `convolution.py` in conformer's directory to support conformer architecture.
- Modify the original `subsampling.py` to take a pos_enc objective as input in conformer's directory.
- Add files related to the integration test.

However, there has a ci test error is in `ASR Mix (backend=pytorch, model=transformer)`. I didn't have this error when I tested in my local machine.

sw005320:
> There has a ci test error in `ASR Mix (backend=pytorch, model=transformer)`. I think this error comes from the `matplotlib` version. The latest `matplotlib==3.3.0` is released on Jul 16, 2020. When using `matplotlib==3.3.0`, the ci test fails, you can see the detail log at https://github.com/pengchengguo/espnet/runs/884156277?check_suite_focus=true. When I force the `matplotlib==3.1.1`, the ci test will succeed, the detail log can be seen at https://github.com/pengchengguo/espnet/actions/runs/173449729.

Thanks!
I made a PR in https://github.com/espnet/espnet/pull/2181

espnet/nets/pytorch_backend/conformer/subsampling.py
sw005320: @pengchengguo, could you add the above descriptions of how it is different from the original transformer subsampling in the code?

pengchengguo:
Yes, I will do it soon.

sw005320:
@pengchengguo, could you merge the update in the master? It would pass the CI error.

sw005320:
I just re-formated the `RESULTS.md` file.
I'll merge this PR after it passes the CI.

Again, many thanks for your great contributions!

PR #2150, opened 2020-07-04
update espnet2 aishell results 

Cescfangs:
lower lr to 0.002
**acc with lr=0.00395**
acc
lr_0

**acc with lr=0.002**

acc

lr_0
githubusercontent.com/11382612/86505586-4deb6400-bdf9-11ea-832c-d4e094009372.png)


sw005320:
Many thanks!

PR #2155, opened 2020-07-08
fix bug making dict 

Cescfangs:
If we limit `vocab_size` or `cutoff` which will cause early break, there would be no words with id < 0(sos/eos)


kamo-naoyuki:
Thank you, but the part of `add_symbol` looks complicated, (this is my fault), so I'll fix it and also refactor it as another PR.

kamo-naoyuki:
Thanks.  I fixed in #2158

PR #2157, opened 2020-07-08
Implement configargparse in espnet2

kamo-naoyuki:
### Issue
- Now the configuration from config file is overwritten by argument.
    ```bash
    % python3 -m espnet2.bin.tts_train --print_config --config conf/tuning/train_gst_tacotron2.yaml --tts_conf bce_pos_weight=10.0
    # The value of tts_conf in conf/tuning/train_gst_tacotron2.yaml is ignored
    ```
- The type checking of configargparse is likely to cause some bugs. I'm sick of it.

### This PR
I decided discarding configargparse from espnet2 and I implemented more simple configargparse in espnet2.utils. New argparse is lack of  some features which original configargparse has.

- Config file is used to change **default** values for `ArgumentParser`.
    ```python
    parser.set_defaults(**from_config_file)
    ```
- Not supporting multiple config files and automatically adding "--config" as an option.
    ```bash
    python asr_train.py --config config.yaml
    ```
- Not supporting any formats other than yaml
- Not checking argument type, e.g. Even if `type=int`, you can give str value from config file.
    ```python
    parser.add_argument("--foo", type=int)
    ```
    ```yaml
    # config.yaml
    foo:  "bar"
    ```
- If `name` is different from `dest`, you must write `dest`  in  the config file for new argparse, while the config file of `configargparse` accepts `name`.
    ```python
    parser.add_argument("--foo", dest="bar", type=int)
    ```
    - Good
        ```yaml
        bar: 4
        ```
    - Bad
        ```yaml
        foo: 4
        ```

### DIFF
--- a/espnet2/bin/asr_inference.py
+++ b/espnet2/bin/asr_inference.py
@@ -1,4 +1,5 @@
 #!/usr/bin/env python3
+import argparse
 import logging
 import sys
 from typing import Optional
@@ -6,7 +7,6 @@ from typing import Sequence
 from typing import Tuple
 from typing import Union
 
-import configargparse
 import torch
 from typeguard import check_argument_types
 
@@ -22,6 +22,7 @@ from espnet2.text.build_tokenizer import build_tokenizer
 from espnet2.text.token_id_converter import TokenIDConverter
 from espnet2.torch_utils.device_funcs import to_device
 from espnet2.torch_utils.set_all_random_seed import set_all_random_seed
+from espnet2.utils import config_argparse
 from espnet2.utils.types import str2bool
 from espnet2.utils.types import str2triple_str
 from espnet2.utils.types import str_or_none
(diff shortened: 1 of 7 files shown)


kan-bayashi:
Sorry for late review.
I will test it this evening.

kan-bayashi:
I checked the behavior with `--print_config`.
However, unfortunately, it still have a problem.
```bash
# Works, not overwrite the `tts_conf` in config
$ python3 -m espnet2.bin.tts_train --print_config --config conf/tuning/train_gst_tacotron2.yaml --tts_conf bce_pos_weight=1.0

# Does not work well, overwrite `tts_conf` in config
$ python3 -m espnet2.bin.tts_train --print_config --config conf/tuning/train_gst_tacotron2.yaml --tts_conf "{bce_pos_weight: 1.0, dropout_rate 0.5}"

```

kamo-naoyuki:
> ```
> # Does not work well, overwrite `tts_conf` in config
> $ python3 -m espnet2.bin.tts_train --print_config --config conf/tuning/train_gst_tacotron2.yaml --tts_conf "{bce_pos_weight: 1.0, dropout_rate 0.5}"
> ```

I also forgot this behaviour, but it is expected.

- `--tts_conf value` => tts_conf=value
- `--tts_conf.key value` => tts_conf[key] = value

It implies replacing the specified key and not updating always because if it indicates updating there are no ways to remove keys.

```python
tts_conf = {"a": 4}
# If updating, we can't remove "a"
tts_conf.update({"b": 5})
tts_conf = {"b": 5}
```

However, we may not need to delete keys and updating might be more intuitive. What do you think?

kan-bayashi:
I have no idea when we want to delete some arguments.
But if there is such a situation, we can use like:
```
$ python3 -m espnet2.bin.tts_train --print_config --config conf/tuning/train_gst_tacotron2.yaml --tts_conf bce_pos_weight=5.0 --tts_conf use_masking=false
```

It is also acceptable for me.
(But it is better to update the document https://espnet.github.io/espnet/espnet2_training_option.html#change-the-configuration-for-dict-type-value)

kamo-naoyuki:
I changed it to update the values.

PR #2158, opened 2020-07-08
Fix bug of when espnet2/bin/tokenize_text.py --cutoff or --vocabulary_size is used

kamo-naoyuki:
I fixed the bug reported in #2155 and refactor for readability.
@Cescfangs Would you check this PR?


kamo-naoyuki:
Yes, I agree, but I'd like to keep it just in case for now.

PR #2164, opened 2020-07-10
fix bug in espnet2 load_pretrained_model

Cescfangs:
I found It won't work when key!=null because of key prefix missing


kamo-naoyuki:
Thanks a lot. 
I'll explain what is expected here because there are no documentation about this, sorry, and I didn't use this option in practical case actually, so we should discuss about the following behaviour. 

We need two arguments originnaly: source key and destination key to determine how to load the parameters, but source key is missing now.

```python
# Psuedo code

# Source key
state = torch.load("pretrain.pth")
source_state = state[source_key]

# Destination_key
model.destination_key.load_state_dict(source_state)
```

You may misunderstand that this option specifies source_key, but it means destination_key. i.e. I expected that the pretrained model comes from any other framework and set it as an attribute of espnet model.

Again, I didn't think much about this. Please give me your idea.

Cescfangs:
I think loading from any framework is nice, makes it more useful👍
BTW I understand it is dest key, it will not work either way I guess 🤔

Naoyuki Kamo <notifications@github.com> 于 2020年7月10日周五 19:31写道：

kamo-naoyuki:
Hmm, I don't know what you expected.
I mean that the level of keys is same between pretrained model and destination model, so we don't need to give prefix keys.

Cescfangs:
Have you ever used this ？if so could you show me the right way to use this
pretrained key other than null, I may misunderstood something badly, thanks
😊


Naoyuki Kamo <notifications@github.com> 于 2020年7月10日周五 22:03写道：

kamo-naoyuki:
How did you use it?


```bash
python -m espnet2.bin.asr_train --pretrain_path decoder.pth --pretrain_key decoder
```

Cescfangs:
Now I know what is going wrong here, I use model.pth instead of
decoder.pth, well I guess I have to get the decoder part out of the model
first 🙃

Naoyuki Kamo <notifications@github.com> 于 2020年7月10日周五 23:07写道：

PR #2165, opened 2020-07-10
Add VCTK recipe for ESPnet2 TTS

kan-bayashi:
This PR adds VCTK recipe and remove unused `fbank.conf` from TTS recipe.


kan-bayashi:
I will merge after CI.

PR #2092, opened 2020-06-24
Python API for inference

kamo-naoyuki:
I implemented API class of espnet2 for inference mode by refactoring the current code.
(Actually I'm on researching side, so I was lack of caring about such demand)

```python
>>> from espnet2.bin.asr_inference import Speech2Text
>>> import soundfile
>>> speech2text = Speech2Text("asr_config.yml", "asr.pth", "lm_config.yaml", "lm.pth", nbest=10)
>>> audio, rate = soundfile.read("speech.wav")
>>> speech2text(audio)  # Return n-best list of recognized results
[(text, token, token_int, hypothesis object), ...]
```

```python
>>> from espnet2.bin.tts_inference import Text2Speech
>>> import soundfile
>>> text2speech = Text2Speech("config.yaml", "tts.pth")
>>> wav = text2speech("Hello World")[0]
>>> soundfile.write("out.wav", wav.numpy(), text2speech.fs, "PCM_16")
```

@ShigekiKarita I fixed the type annotation of Hypothesis class because it causes type check error now.

PR #2168, opened 2020-07-12
Implement espnet2.bin.zenodo_upload

kamo-naoyuki:
I implemented a script to upload pretrained model to zenodo. This is an example: https://sandbox.zenodo.org/record/646767#.Xwqq6577TOQ

I also added --asr_exp and --lm_exp option in asr.sh and --tts_exp in tts.sh. It specifies the path of expdir directly.

```bash
unzip ${packed_model}
exp=$(basename ${packed_model} .zip)
./run.sh --skip_data_prep false --skip_train true --asr_exp ${exp}/asr --decode_asr_model pretrain.pth --lm_exp ${exp}/lm --decode_lm pretrain.pth
```

### DIFF
--- a/ci/test_integration.sh
+++ b/ci/test_integration.sh
@@ -94,7 +94,7 @@ done
 for t in ${feats_types}; do
     for t2 in ${token_types}; do
         echo "==== feats_type=${t}, token_types=${t2} ==="
-        ./run.sh --ngpu 0 --stage 6 --stop-stage 100 --feats-type "${t}" --token-type "${t2}" \
+        ./run.sh --ngpu 0 --stage 6 --stop-stage 13 --feats-type "${t}" --token-type "${t2}" \
             --asr-args "--max_epoch=1" --lm-args "--max_epoch=1"
     done
 done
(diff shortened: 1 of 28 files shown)


kamo-naoyuki:
> The sampling rate is the most FAQ for the pre-trained model, and if this sampling rate information is also added to a keyword, we may avoid such FAQs, but this would be a bit complicated.
> * If we add some language IDs, that would be good, but we cannot automatically extract such information, and it can be optional.

Sampling rate and language id is surely important for general users. I'll think.

> * We can generate a usage in the description, e.g.,
> 
> ```
> $ cd <espnet dir>/egs2/an4/tts1
> $ wget <this model> 
> $ packed_model=`echo <this model> | awk -F "/" '{print $NF}' | awk -F "?" '{print $1}'`
> $ unzip ${packed_model}
> $ ./run.sh --stage 8 --tts_exp $(basename ${packed_model} .zip) --decode_model pretrain.pth
> ```

> * how about including zenodo model downloading in the inference stage (10 in ASR and 8 in TTS) by adding a model link variable in `asr.sh` or `tts.sh`? If the model exists and model download is enabled, people can skip the training stage and directly perform the inference. Then, the above usage description is simplified by `./run.sh --stage 8 --model_link <this model>`

Okay, but let me rethink in #2092 about downloading. My goal is similar system like https://github.com/mpariente/asteroid.

kamo-naoyuki:
@sw005320 

I updated the description: https://sandbox.zenodo.org/record/646767#.Xwqq6577TOQ

> The sampling rate is the most FAQ for the pre-trained model, and if this sampling rate information is also added to a keyword, we may avoid such FAQs, but this would be a bit complicated.
> * If we add some language IDs, that would be good, but we cannot automatically extract such information, and it can be optional.

Done. Note that we have to give --lang option in run.sh in order to derive language name. This option might be also useful for future development.

> * We can generate a usage in the description, e.g.,

Done.

> * how about including zenodo model downloading in the inference stage (10 in ASR and 8 in TTS) by adding a model link variable in `asr.sh` or `tts.sh`? 

I'll add the option at the next PR.

PR #2170, opened 2020-07-13
fix integration_test_ctc_align_wav.bats with a small model

simpleoier:
The size of the small model is 15MB. I still use the google drive to store the model. I'm not sure whether there is still a problem when too many download trials.


sw005320:
> The size of the small model is 15MB. I still use the google drive to store the model. I'm not sure whether there is still a problem when too many download trials.

Many thanks!
I believe it would not be a problem.

PR #2171, opened 2020-07-13
CHiME-6 recipe

GNroy:
This is a CHiME-6 recipe implementation according to https://arxiv.org/abs/2004.10799

We obtained 55.3 WER on the default beam search algorithm against 55.5 in the paper.

Please note that we had to put a stub to line 754 of asr.py for the model to be able to train. Thus the request is WIP and needs clarification on how to resolve this without breaking the code.

Thanks,
Alex Laptev


egs/chime6/asr1/local/prepare_dict.sh
```diff
@@ -0,0 +1,126 @@
+#!/bin/bash
```
sw005320, line 1: Probably we don't need this?

egs/chime6/asr1/local/run_beamformit.sh
```diff
@@ -0,0 +1,87 @@
+#!/usr/bin/env bash
```
sw005320, line 1: We may not need this?

egs/chime6/asr1/local/train_lms_srilm.sh
```diff
@@ -0,0 +1,261 @@
+#!/usr/bin/env bash
```
sw005320, line 1: We may not need this?

egs/chime6/asr1/conf/preprocess.yaml
```diff
@@ -0,0 +1,21 @@
+process:
```
sw005320, line 1: this looks similar to `specaug.yaml`.
could you use either of them?

egs/chime6/asr1/conf/preprocess.yaml
```diff
@@ -0,0 +1,21 @@
...
+    n_mask: 2
+    inplace: true
+    replace_with_zero: false
+  - type: "time_mask"
+    T: 40
+    n_mask: 2
+    inplace: true
+    replace_with_zero: false
+  - type: "token_augment"
+    gtm_path: "/mnt/k05/laptev/exp/espnet_chime6_asr_recipe_exp/jsondata/train_worn_simu_u400k_cleaned_trim_sp/gtm.csv"
```
sw005320, line 18: This should be removed.
Probably this "token_augment" is an experimental function?

sw005320:
@sas91, could you also review this PR?
This PR has some unique CHiME processing including GSS processing etc., but we can just leave this enhancement part as it is.
I hope multichannel e2e ASR or other e2e work will replace this module in the future.

sw005320:
@b-flo, you may also check this PR. I think this is a very successful example of your transducer ;)

egs/chime6/asr1/local/prepare_dict.sh
GNroy: Yes, this file is redundant. We removed it and some other unused scripts.

egs/chime6/asr1/local/run_beamformit.sh
GNroy: Yes, this file is redundant. We removed it and some other unused scripts.

egs/chime6/asr1/local/train_lms_srilm.sh
GNroy: This file is used in the data preparation too. They build LM for further use in the data cleanup.

egs/chime6/asr1/conf/preprocess.yaml
GNroy: Thanks for pointing out! It leaked from another experiment. We removed it.

egs/chime6/asr1/conf/preprocess.yaml
GNroy: We made it specaug.yaml

GNroy:
The details of the problem are as follows:
When using the standard adadelta optimizer, there is a comparison between this epoch' model loss and the model.best.loss. If the current loss is worse than the best previous one, then the snapshot model weights will be taken for the following epoch, instead of the current weights.
In our case, this often led the training process to end up in a local optimum, thus resulting in about 60%WER.
Thereby, we put a stub in this comparison and obtained our 55.3%WER.

GNroy:
@sw005320 We need a piece of advice on how to make this comparison optional, thus to sustain our best result and not to affect existing ones in other recipes.

GNroy:
I think I fixed the CI errors related to the recipe.

sw005320:
Could you merge the update in the master branch?
We fixed a CI issue.

@sas91, 
This is a reminder.
Could you review this PR?

sw005320:
> The details of the problem are as follows:
> When using the standard adadelta optimizer, there is a comparison between this epoch' model loss and the model.best.loss. If the current loss is worse than the best previous one, then the snapshot model weights will be taken for the following epoch, instead of the current weights.
> In our case, this often led the training process to end up in a local optimum, thus resulting in about 60%WER.
> Thereby, we put a stub in this comparison and obtained our 55.3%WER.

I see.
How about adding another argument to `args.criterion`?
e.g.,

```
        elif args.criterion == "loss_eps_decay_only":
            trainer.extend(
                adadelta_eps_decay(args.eps_decay),
                trigger=CompareValueTrigger(
                    "validation/main/loss",
                    lambda best_value, current_value: best_value < current_value,
                ),
            )
```
Then, could you also add some comments around this code and config that this is particularly important for this task to avoid local optimum?

GNroy:
@sw005320 Thank you for your suggestions!
I've added "loss_eps_decay_only" argument and related comments.
Unfortunately, codecov CI had failed since now there is more code in asr.py

egs/chime6/asr1/local/install_pb_chime5.sh
```diff
@@ -0,0 +1,24 @@
+#!/bin/bash
+
+# Installs pb_chime5
+# miniconda should be installed in $HOME/miniconda3/ 
+
```
sas91, line 5: Can we reuse the python venv from espnet installation here to install GSS instead of reinstalling miniconda. @sw005320 What do you think?

egs/chime6/asr1/conf/train.yaml
```diff
@@ -0,0 +1,46 @@
+# minibatch related
```
sas91, line 1: Can you also please add the transformer transducer configuration from your paper in a different file? (not the default option, just in case if someone wants to try it)

egs/chime6/asr1/local/run_gss.sh
```diff
@@ -0,0 +1,66 @@
...
+
+. ./cmd.sh
+if [ -f ./path.sh ]; then . ./path.sh; fi
+
+# Config:
+cmd=run.pl
+nj=4
+multiarray=outer_array_mics   #-- 16 channels for gss; usually better than 24 channels gss
+#multiarray=True              #-- 24 channels for gss
+bss_iterations=5
```
sas91, line 13: 24 channel GSS seems to be better from your paper when also used with train GSS option. This parameter "multiarray" doesn't seem to be passed from "run.sh", can you please add options in "run.sh" and pass it from there.

egs/chime6/asr1/local/install_pb_chime5.sh
sw005320: I think this is a good idea.

egs/chime6/asr1/local/prepare_baseline_chime6_data.sh
```diff
@@ -0,0 +1,326 @@
...
+  if [ ! -f pb_chime5/cache/chime6.json ]; then
+    (
+    cd pb_chime5
+    miniconda_dir=$HOME/miniconda3/
+    export PATH=$miniconda_dir/bin:$PATH
+    export CHIME6_DIR=$chime6_corpus
+    make cache/chime6.json
+    )
+  fi
+
```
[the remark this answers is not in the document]
GNroy: Thank you for your suggestion!
Enhancing training data does lower the WER, but getting such data is extremely resource-intensive. In our research, we didn't do it ourselves. This train_gss data were provided by an STC team researching GSS improvements. Some information about the modified GSS can be found on page 6 of https://chimechallenge.github.io/chime2020-workshop/presentations/CHiME_2020_slides_medennikov.pdf
Even if we include this in the recipe, we will not be able to test it due to a lack of computing resources. Thus, I would prefer not to include it in this pool request.

egs/chime6/asr1/run.sh
```diff
@@ -0,0 +1,188 @@
...
+if [ ${stage} -le 0 ] && [ ${stop_stage} -ge 0 ]; then
+    ### Task dependent. You have to make data the following preparation part by yourself.
+    ### But you can utilize Kaldi recipes in most cases
+    echo "stage 0: Data preparation"
+    local/prepare_baseline_chime6_data.sh --chime5_corpus ${chime5_corpus}
+fi
+
+train_set=train_worn_simu_u400k_cleaned
+train_dev=dev_gss
+
```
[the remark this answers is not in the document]
GNroy: Thank you for this suggestion!
Measuring WER on eval was beyond the scope of our research. We have neither results for it, nor the data itself.
Thus, if it is not necessary for the recipe, I would prefer not to include eval in the request.

egs/chime6/asr1/conf/train.yaml
GNroy: Thanks for your suggestion!
I'll do it.

egs/chime6/asr1/local/install_pb_chime5.sh
GNroy: Thanks for this suggestion!
I'll try to do it.

egs/chime6/asr1/local/run_gss.sh
GNroy: Thanks for your suggestion!
Indeed, according to our results, WER on dev_gss24 is better than on dev_gss12. However, this gss24 is built according to the STC pipeline (as I mentioned above) that is different from the baseline gss24. In our local experiments, the baseline gss24 performed about 0.5-0.6% worse than the gss12. Yet, I can add an option with proper comments for it in run.sh

sw005320:
How about adding the transformer transducer result in `RESULTS.md` even it is worse then RNN-T?
This is good information

b-flo:
Hi,

> @b-flo, you may also check this PR. I think this is a very successful example of your transducer ;)

Sorry, I don't think I'll have time to thoroughly check the PR before it's merged. I'll do that after my PR as I'll need to update the recipe :) 
On that note @GNroy, you may want to test some new features for transducer at that time!

sw005320:
I think this is a very good shape now.
@GNroy, please let me know if it is ready.
I'll merge it.

GNroy:
@sw005320 It is ready.
Thank you for your guidance on this PR!

GNroy:
@b-flo I'm looking forward to testing this new transducer features, especially TDNN!

sw005320:
Many thanks!

PR #2179, opened 2020-07-17
Fix ST/MT models for compatibility with ASR

hirofumi0810:
I added lightweight and dynamic convolutions to Transformer-based ST/MT models.


sw005320:
Thanks!
LGTM!

Please let us know if our lightweight and dynamic convolutions are effective for MT/ST

PR #2181, opened 2020-07-18
Freeze the matplotlib version with 3.1.0

sw005320:


setup.py
```diff
@@ -13,7 +13,7 @@
         "dataclasses",  # For Python<3.7
         "humanfriendly",
         "scipy>=1.4.1",
-        "matplotlib>=3.1.0",
+        "matplotlib>=3.1.1",
```
pengchengguo, line 16: Hi Shinji, I think this may still install `matplotlib==3.3.0` (the latest version). We should force it to install `matplotlib` less than 3.3.0. I also found people have the same error as me (https://stackoverflow.com/questions/62953704/valueerror-the-number-of-fixedlocator-locations-5-usually-from-a-call-to-set). I am not sure whether it's a common error.

sw005320: Thanks. I'll fix it with 3.2.2 tentatively.
If matplotlib could solve this issue, I'll change it to use the latest one.

jackz314: Is this freeze still necessary? matplotlib 3.1.0 doesn't have a pre-built wheel file on Windows, which means it has to be built from source on Windows, this is making installing espnet on Windows pretty difficult. Other than this, everything else works pretty much out of the box on Windows.

sw005320: Year, we need to revisit it. 
Do you have some recommended versions with respect to the pre-built wheel file on Windows?

jackz314: Currently my matplotlib version is 3.3.3, which does have pre-built wheel files. However my use case of espnet doesn't really use matplotlib (I'm only doing TTS inference for now), so I don't know if the bug mentioned above is fixed.

PR #2064, opened 2020-06-20
Voice Transformer Network

unilight:
Hi, I plan to open-source the Voice Transformer Network a.k.a. Transformer-VC.
Tl.dr: Parallel VC based on mel-spectrogram using TTS-pretrained Transformer (Tacotron 2 also supported). #1670 
https://arxiv.org/pdf/1912.06813.pdf

This is still WIP since I am still cleaning my local repo.
I am just opening this PR to ask you to help me check the design since I am touching some core functions (those in `espnet/`).

My original paper proposed pertaining with TTS (please check my paper), but I think it is troublesome to provide the script for pertaining. (Training from scratch does not reach satisfactory performance)
So I am thinking about providing the pretrained model (for training) only.

### Essential
- [x] Clean up `arctic/vc/run.sh`
- [x] Rename the main config files
- [x] Upload the pretrained model for training and write downloading scripts
- [x] Rename `vc` to `vc1`
- [x] write tests

### Documentation
- [x] Modify `egs/README.md`
- [x] Modify `README.md`
- [x] Modify `egs/arctic/README.md` to explain how to use this recipe?

### Misc
- [x] move `local/make_pair_json.py` to `utils`?
- [x] <del>Make symbolic links for some files in `conf/`?</del>

### Future work
- integration tests
- inherit the class from TTS ones and override some methods. (c.f., `e2e_asr_conformer`)

PR #2199, opened 2020-07-22
Update shellcheck URL in test_shell.sh

unilight:
The URL to download shellcheck is out-dated. Updating to the new one.

Later in this PR, I might also update this doc page: https://github.com/espnet/espnet/blob/master/CONTRIBUTING.md#unit-testing since it is also a bit out-dated. For example, ./ci/test_bash.sh does not exist anymore and has been integrated into ./ci/test_shell.sh.


sw005320:
I don't know why but circleci suddenly started to check the shell of several files that we have not checked in this PR....

sw005320:
Maybe we can leave it as it is?

unilight:
I will do this update in PR #2064 . If the same problem happens there too, let's just leave as it is...

PR #2209, opened 2020-07-23
Append a missing space in cmd-line args in utils/dump_pcm.sh

yistLin:
Without a space there, an error could happen when the variable **opts** is appended with **--format wav** because there's no space in between **--write-num-frames=xxx** and **--format wav**.


kamo-naoyuki:
Thanks!

PR #2210, opened 2020-07-23
Add integration test for ST/MT tasks

hirofumi0810:


sw005320:
This is great!
But...
It seems that this would increase the testing time significantly...
Can you prune some of the integration tests?

sw005320:
Many many thanks!!!

PR #2214, opened 2020-07-23
Select torchaudio version according to torch version

kamo-naoyuki:
We can't add torchaudio as a requirement in setup.py because it requires specific pytorch version, but we need to control it.
(Now, pytorch=1.5.1 is always installed)


espnet2/layers/stft.py
```diff
@@ -126,12 +126,28 @@ def inverse(
...
+            except ImportError:
+                raise ImportError(
+                    "Please install torchaudio>=0.3.0 or use torch>=1.6.0"
+                )
+
+            if not hasattr(torchaudio.functional, "istft"):
+                raise ImportError(
+                    "Please install torchaudio>=0.3.0 or use torch>=1.6.0"
+                )
+            istft = torchaudio.functional.istft
```
mthrok, line 143: Looks good.

kamo-naoyuki:
@LiChenda @Emrys365 @shincling 
Please check. I updated some parts to pass unit tests. Note that you can call double() method explicitly, but you can't call float() in the froward function because we need to keep an option to calculate all in double precision.

Emrys365:
> Please check. I updated some parts to pass unit tests. Note that you can call double() method explicitly, but you can't call float() in the froward function because we need to keep an option to calculate all in double precision.

Thank you! I checked the updated code and it looks good to me.

test/espnet2/enh/nets/test_beamformer_net.py
```diff
@@ -190,6 +191,7 @@ def test_beamformer_net_consistency(
 def test_beamformer_net_wpe_output(ch, num_spk, use_dnn_mask_for_wpe):
     torch.random.manual_seed(0)
     inputs = torch.randn(2, 16, ch) if ch > 1 else torch.randn(2, 16)
+    inputs = inputs.float()
```
Emrys365, line 194: How about adding a type consistency check on the model output?
```python
    assert spec.dtype == torch.float
```

kamo-naoyuki:
Note that the setup.py changes the requirements depending on torch version dynamically, but wheel file always has static requirements.
When uploading espnet package, we don't install torch, so our wheel doesn't include torchaudio in the requirements.

```sh
pip install espnet
pip install torchaudio=0.5.1
```

Issue #2211, opened 2020-07-23
WSJ dataset processing error - AttributeError: 'Namespace' object has no attribute 'etype'


Jxu-Thu:
When run the run.sh at espnet/egs/wsj/asr1/run.sh,  run this command 
`
local/filtering_samples.py \
        --config ${train_config} \
        --preprocess-conf ${preprocess_config} \
        --data-json ${feat_tr_dir}/data.json \
        --mode-subsample "asr" \
        --arch-subsample "rnn" \
        ${min_io_delta:+--min-io-delta $min_io_delta} \
        --output-json-path ${feat_tr_dir}/data.json
`

get this error 

Traceback (most recent call last):
  File "local/filtering_samples.py", line 65, in <module>
    args, mode=args.mode_subsample, arch=args.arch_subsample
  File "/mypath/espnet/espnet/nets/pytorch_backend/nets_utils.py", line 411, in get_subsample
    if train_args.etype.endswith("p") and not train_args.etype.startswith("vgg"):
AttributeError: 'Namespace' object has no attribute 'etype'

I print the args parameters and do not find the 'etype'.

And why arch-subsample is rnn? this script run the asr model with transformer?

https://github.com/espnet/espnet/blob/74de653ef7d37fc204562e4142c86d3c65ce96fe/egs/wsj/asr1/run.sh#L149

Emrys365:
You are right, [espnet/egs/wsj/asr1/run.sh#L149](https://github.com/espnet/espnet/blob/master/egs/wsj/asr1/run.sh#L149) should not specify `rnn` explicitly. It should be determined by the architecture in the model conf.
For a quick fix, could you try changing `--arch-subsample "rnn"` to `--arch-subsample "transformer"` to make it work?

Emrys365:
I just made a PR to fix this issue. Now the argument `--arch-subsample` is removed, and it will be parsed from `model-module` in the config file.
@Jxu-Thu, Could you try it?

Jxu-Thu:
Ok, Thanks!

sw005320:
@Jxu-Thu, please let me know if it solves the problem. Then, I’ll merge #2211

jaesong:
I confirmed the #2211 (c0cd327fd69446cca6ec77a0b51da14288c32d28) solves the issue on WSJ.

PR #2219, opened 2020-07-24
Fix issue https://github.com/espnet/espnet/issues/2211

Emrys365:
This PR fixes the bug in espnet/egs/wsj/asr1/local/filtering_samples.py reported by https://github.com/espnet/espnet/issues/2211.


sw005320:
Thanks for your quick fix, @Emrys365!
(and sorry for my late action about this PR).

PR #2221, opened 2020-07-25
CTC posterior visualization during training

hirofumi0810:
This PR adds CTC posterior visualization during training as in attention weight plotting.
CTC visualization in the auxiliary ASR task during ST model training is also supported.

An example is as follows:
20051009_182032_217_fsp-A-000576-000919 ep 34


sw005320:
LGTM!
If it is ready, please merge it (or I can do it for you).

Issue #2188, opened 2020-07-20
Update chainer to v6.7.0?


kamo-naoyuki:
I'm ignoring the version conflict of protobuf between chainer=6.0.0 and tensorboardX in a long time because we doesn't use chainer except for its trainer and also note that we doesn't need to use tensorboardX because pytorch already includes its function. It's really waste of time to spend our energy for this topic. 
However, now I faced a problem about version conflicting, so I must solve this problem.

I know that chainer=7 is incompatible with our espnet, so we need to use chainer=6 at least.
I found that chainer=6.5,6.6,6.7 doesn't include the version restriction for protobuf, https://github.com/chainer/chainer/blob/v6.7.0/setup.py, so maybe we should use v6.7.0.

I don't have any resources to test this, but I'd like to solve this problem soon. Can anyone test espnet with chainer=6.7.0?

kamo-naoyuki:
Can anyone help, please.

kan-bayashi:
I will check it this evening!

kan-bayashi:
I tested with:

- chainer 6.7.0
- cupy 6.7.0

I confirmed that `egs/yesnot/asr1` worked well with both --backend chainer and pytorch.
And `ci/test_ingregration.sh` finished without errors.

kamo-naoyuki:
Okay, thanks! Multi gpus also works well?

kan-bayashi:
Hmm... Multi-GPU did not work well with chainer backend.

|    | cupy 6.0.0 | cupy 6.5.0 | cupy 6.7.0 | 
| :-: | :-: | :-: | :-: |
| chainer 6.0.0 | ○ | - | - | 
| chainer 6.5.0 | x | × | - |
| chainer 6.7.0 | x | - | × |

kan-bayashi:
And I found that ESPnet1 with torch=1.5.1 + multi-gpu did not work well.

```bash
Exception in main training loop: Caught StopIteration in replica 0 on device 0.
Original Traceback (most recent call last):
  File "/home/t_hayashi/workspace3/work/espnet/tools/venv/lib/python3.7/site-packages/torch/nn/parallel/parallel_
apply.py", line 60, in _worker
    output = module(*input, **kwargs)
  File "/home/t_hayashi/workspace3/work/espnet/tools/venv/lib/python3.7/site-packages/torch/nn/modules/module.py"
, line 550, in __call__
    result = self.forward(*input, **kwargs)
[...]

Issue #2169, opened 2020-07-12
Multi-GPU training failure with PyTorch 1.5.1


unilight:
E.g. in `egs/libritts/tts1`, `$ CUDA_VISIBLE_DEVICES=0,1 ./run.sh --ngpu 2 --trans_type phn` will result in the following error in the log file,
```
Exception in main training loop: Caught StopIteration in replica 0 on device 0.
Original Traceback (most recent call last):
  File "/home/huang18/VC/Experiments/espnet/tools/venv/lib/python3.6/site-packages/torch/nn/parallel/parallel_apply.py", line 60, in _worker
    output = module(*input, **kwargs)
  File "/home/huang18/VC/Experiments/espnet/tools/venv/lib/python3.6/site-packages/torch/nn/modules/module.py", line 550, in __call__
    result = self.forward(*input, **kwargs)
  File "/nas01/internal/wenchin-h/VC/Experiments/espnet/espnet/nets/pytorch_backend/e2e_tts_tacotron2.py", line 746, in forward
    after_outs, before_outs, logits, att_ws = self.dec(hs, hlens, ys)
  File "/home/huang18/VC/Experiments/espnet/tools/venv/lib/python3.6/site-packages/torch/nn/modules/module.py", line 550, in __call__
    result = self.forward(*input, **kwargs)
  File "/nas01/internal/wenchin-h/VC/Experiments/espnet/espnet/nets/pytorch_backend/tacotron2/decoder.py", line 436, in forward
    att_c, att_w = self.att(hs, hlens, z_list[0], prev_att_w)                                                                                                                                           File "/home/huang18/VC/Experiments/espnet/tools/venv/lib/python3.6/site-packages/torch/nn/modules/module.py", line 550, in __call__
    result = self.forward(*input, **kwargs)                                                                                                                                                             File "/nas01/internal/wenchin-h/VC/Experiments/espnet/espnet/nets/pytorch_backend/rnn/attentions.py", line 365, in forward
    self.mask = to_device(self, make_pad_mask(enc_hs_len))
  File "/nas01/internal/wenchin-h/VC/Experiments/espnet/espnet/nets/pytorch_backend/nets_utils.py", line 24, in to_device
    device = next(m.parameters()).device
StopIteration
Traceback (most recent call last):                                                                                                                                                                      File "/home/huang18/VC/Experiments/espnet/tools/venv/lib/python3.6/site-packages/chainer/training/trainer.py", line 316, in run
    update()
[...]

PR #2225, opened 2020-07-27
Update Makefile

kamo-naoyuki:
- Change to remove chainer from requirements of setup.py and install it from Makefile
    - chainer=6.0.0 has incompatible requirement of protobuf to tensorboardX #2188 
      As workaround, I removed it from setup.py.
   - We still have conflicted packages due to underthesea or languageflow.
- I changed the default PyTorch to 1.4.0.
   - We should use pytorch=1.5.1, but espnet1 doesn't work, so I set 1.4.0 for now #2169
- I removed kaldi installation part from Makefile.
   - KALDI is mandatory argument: `$ make KALDI=<kaldi-path>`
- New variables:
    - CPU_ONLY: Perform on CPU mode
    - CONDA: You can specify existing anaconda path
    - CONDA_ENV_NAME
    - CHAINER_VERSION
    - USE_PIP: Use pip to install torch even if it's anaconda
- Romoved:
    - CUDA_VERSION: Detecting cuda version from `nvcc`
    - CUPY_VERSION: Use CHAINER_VERSION instead
- Makefile creates `tools/activate_python.sh`. `path.sh` invokes it.
    - To fix a current issue: If already in another anaconda env when make, it's failed to install packages because `conda deactivate` is not called.
    - Now we can easily replace python environment rewriting `tools/activate_python.sh`.
- Implement `tools/setup_cuda_env.sh`
    - A util to setup CUDA environment variables, $PATH, $LD_LIBRARY_PATH, etc.
    - This file is not called in Makefile.
- Rename `tools/install_*.sh` -> `tools/installers/install_*.sh`


doc/installation.md
```diff
@@ -185,27 +182,47 @@ $ cd <espnet-root>/tools
...
+You can skip the installation of new environment by creating `activate_python.sh`.
+
+```sh
+$ cd <espnet-root>/tools
+$ rm -f activate_python.sh; touch activate_python.sh
+$ make KALDI=<kaldi-root> USE_PIP=0
+```
 
-To install in a terminal that does not have a GPU installed, just clear the version of `CUPY` as follows:
+If your Python is anaconda, you don't need to give `USE_PIP`.
```
sw005320, line 202:
```suggestion
If your Python is anaconda, you don't need to provide `USE_PIP`.
```

setup.py
```diff
@@ -75,22 +75,25 @@
     ],
 }
 try:
+    # NOTE(kamo): There packages are not listed if installing from pypi server
```
sw005320, line 78:
```suggestion
    # NOTE(kamo): These packages are not listed if installing from the PyPI server
```
??? Could you check it?

kamo-naoyuki:
@sw005320 Sorry, I'll consider editor setting. After ci, I'll merge

sw005320:
> @sw005320 Sorry, I'll consider editor setting. After ci, I'll merge

No problem at all.
Better documentation is important teamwork. I'll try to help as much as possible.

Issue #2200, opened 2020-07-22
Development plan for ESPnet2 speech enhancement 


LiChenda:
We have been developing (enhancing) speech enhancement functions in ESPnet2.
We particularly design it to tightly integrate it with downstream speech processing tasks (I.e., ASR and ST). 
Phase 1 is to implement a stand alone basic speech enhancement and separation techniques as a reference technique. In Phase 2, they are flexibly called as a front-end processing subnetwork from ASR/ST/SLU/DIAR with maintaining computational graphs for joint training.

**We also welcome your requests for this particular directions and contributions!**

## Progress & Plan  #2124 

- [x] A new task for speech enhancement
- [x] Speech enhancement recipe template
- [x] Evaluate with pretrained ASR
- [x] Joint training with ASR/ST/SLU/DIAR tasks
- [x] Single-channel multi-talker ASR
- [ ] Support integration of models from other tools 
    - [ ] [PaderTorch](https://github.com/fgnt/padertorch)
    - [x] [Asteroid](https://github.com/mpariente/asteroid)

#### frequency domain
- [x] Conformer
- [x] [DCCRN](https://www.isca-speech.org/archive/interspeech_2020/hu20g_interspeech.html)
- [x] [Deep Attractor Network](https://arxiv.org/abs/1611.08930)
- [x] [Deep Clustering](https://ieeexplore.ieee.org/abstract/document/7471631/) series
- [ ] [DPT-FSNet](https://arxiv.org/abs/2104.13002)
- [ ] [Speaker-Conditional Chain Model](https://www.isca-speech.org/archive/interspeech_2020/shi20e_interspeech.html)
- [ ] [TFPSNet](https://ieeexplore.ieee.org/document/9747554/)
- [x] [TF-GridNet](https://arxiv.org/abs/2209.03952)
- [x] Transformer
- [x] [uPIT](https://ieeexplore.ieee.org/abstract/document/7979557/)

#### time domain
- [ ] [CDiffuSE](https://arxiv.org/abs/2202.05256)
- [x] [Conv-TasNet ](https://arxiv.org/abs/1809.07454)
- [x] [DPRNN](http://arxiv.org/abs/1910.06379v2)
- [x] [DPTNet](https://arxiv.org/abs/2007.13975)
- [ ] [MixIT](https://arxiv.org/abs/2006.12701) (PRing, see #4263)
- [x] [SkiM](https://arxiv.org/abs/2201.10800)
[...]

PR #2238, opened 2020-07-29
Add some new features and a new recipe for the enhancement task

Emrys365:
This PR is related to https://github.com/espnet/espnet/issues/2200.

1. Now different types of losses are supported for mask based networks, e.g. loss on enhanced spectrum, loss on predicted masks, etc.
2. A new recipe `egs2/mini_an4/enh1/` is created and also added to `ci/test_integration.sh` for testing.


Emrys365:
An additional commit from @LiChenda is needed for fixing some issues in enh.sh.

ci/test_integration.sh
```diff
@@ -157,6 +157,18 @@ done
 rm -rf exp dump data
 cd "${cwd}" || exit 1
 
+# [ESPnet2] test enh recipe
```
kamo-naoyuki, line 160: Thank you! Could you also add validation part for the config files?

https://github.com/espnet/espnet/blob/933c94870a2ed6b1748fb34dbd13406b61f65a75/ci/test_integration.sh#L164-L177

Emrys365: Done.

LiChenda:
> An additional commit from @LiChenda is needed for fixing some issues in enh.sh.

It's solved.

Emrys365:
> LGTM.
> I’m just curious why you use MSE for the mask loss? I thought that it would be better to use CE (in the binary case), KL or whatever (information theoretic loss).

Yes, I think we should make it flexible to choose any loss function that is appropriate for the selected loss type and mask type. I'll work on this later.
Currently, we are mainly testing the phase-sensitive mask (PSM), and I think it's okay to use mse loss for this mask type (according to [this paper](https://ieeexplore.ieee.org/document/7178061)).

Emrys365:
It seems torch 1.1.0 is installed on the CI machine, so torchaudio won't be installed in this case.
Could you check it? @kamo-naoyuki 
https://github.com/espnet/espnet/pull/2238/checks?check_run_id=928189635#step:6:1158

kamo-naoyuki:
So, you need to skip it in ci if <1.1.0.

espnet2/enh/layers/dnn_wpe.py
```diff
@@ -84,11 +84,31 @@ def forward(
...
+            masks (torch.Tensor): (B, T, C, F)
+            ilens (torch.Tensor): (B,)
+        """
+        if self.use_dnn_mask:
+            (mask,), ilens = self.mask_est(data.permute(0, 3, 2, 1).float(), ilens)
+            # (B, F, C, T) -> (B, T, C, F)
+            mask = mask.transpose(-1, -3)
+        else:
+            mask = None
+        return mask, ilens
```
sas91, line 114: In this file, you seem to be using mask_estimator from espnet1. I think using sigmoid activation might not be best for dereverberation mask. I think torch.clamp would be better. Can we pass arguments to control this activation instead of sigmoid being the only option?

sas91: Also, for dereverberation, doing a direct signal approximation (instead of masking approximation) is better when trained as a pipeline. Is it possible to add this option?

Emrys365:
> So, you need to skip it in ci if <1.1.0.

OK, now I add the conditional block.

espnet2/enh/layers/dnn_wpe.py
Emrys365: Thank you for your suggestions!

> Can we pass arguments to control this activation instead of sigmoid being the only option?

Yes. I have added two additional arguments to allow choosing the nonlinear type for both wpe and beamformer.

> Also, for dereverberation, doing a direct signal approximation (instead of masking approximation) is better when trained as a pipeline. Is it possible to add this option?

This is a bit complicated. Because the training loss is calculated in `espnet2/enh/espnet_model.py`. For now, it only gives an option to choose the loss type for all masks. If we want to choose the loss type for a specific module of a specific model, the code will be a bit messy.

A current workaround may be training the WPE and beamformer part (using different loss types) separately and combine them later.

kamo-naoyuki:
ready to merge?

Emrys365:
> ready to merge?

Yes, you can merge now. We will make a new PR later when the frontend-backend joint model is completed.

kamo-naoyuki:
Thanks.

kamo-naoyuki:
Hi. The design of `ESPnetEnhASRModel` is bad.
It is a combination of ESPnetASRModel and ESPnetEnhancementModel. In this case, delegate or inherit should be used according to common programming strategy.

- Delegate

```python
class ESPnetEnhASRModel(AbsESPnetModel):
    def __init__(
         self, 
         enh_model: ESPnetEnhancementModel,
         asr_model: ESPnetASRModel,
         ...
     ):
```

- Inherit

```python
class ESPnetEnhASRModel(ESPnetEnhancementModel, ESPnetASRModel):
```

Delegate is better.

Emrys365:
Thank you for your suggestion! @shincling Could you try redesigning it?

shincling:
Ok. I will do it.

shincling:
> Thank you for your suggestion! @shincling Could you try redesigning it?

Ok. I will do it.

PR #2245, opened 2020-07-30
Remove  installation part of Python from Makefile

kamo-naoyuki:
I removed  installation part Python from Makefile and create setup_anaconda.sh and setup_python.sh. I also removed KALDI option from Makefile. I think this is more clear and make Makefile more simple. 

Installation flow becomes as follows.

1. Install Kaldi and create a symlink at tools/
2. Setup Python (= create `tools/activate_python.sh`)
3. make


kamo-naoyuki:
I added conda_packages target, which installs:

- libflac
- sox
- ffmpeg
- libsndfile

kamo-naoyuki:
> Frankly, I'm bit hesitating to go in this direction. I know that the current Makefile is messy and many confusions come from it. However, this provides one command building of all required tools (Yes, I know that this causes a problem).

You know, the python environment depends on user style and our Makefile seems busybody. People ask and meet some errors many times and many times. Actually I was tired to answer them and I decided to change it.

(However, I believe that people should create a new environment when they start a new experiment in order to make each experiment independent. Our default style was reasonable. 
Additionaly, I'm sure that people will probably ask some questions due to their own incomplete Python environment...)

PR #2254, opened 2020-08-02
Take integration tests into coverage

ShigekiKarita:
I replaced `python xxx.py` with `coverage run --append -m espnet.bin.xxx` for including integration tests into coverage stats (see [coverage.py usage](https://coverage.readthedocs.io/en/coverage-5.2.1/cmd.html#cmd-run)). The reason why I used `-m xxx` instead of `xxx.py` is that coverage.py cannot find non-local files.

I also disabled some codecov lines in CI without integration tests because their reports are different from (i.e. fewer than) ones with integration tests.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -170,11 +170,11 @@ jobs:
       #     name: test integration
       #     command: |
       #       ./ci/test_integration.sh
-      - run:
-          name: codecov
-          command: |
-            source ./tools/venv/bin/activate
-            bash <(curl -s https://codecov.io/bash)
+      # - run:
+      #     name: codecov
+      #     command: |
+      #       source ./tools/venv/bin/activate
+      #       bash <(curl -s https://codecov.io/bash)
       - store_artifacts:
           path: egs/mini_an4/asr1/exp
 
--- a/.travis.yml
+++ b/.travis.yml
@@ -34,7 +34,7 @@ install:
   # - travis_retry ./ci/install_kaldi.sh
 
 script:
-  # NOTE(kamo): Codecov and build documentation only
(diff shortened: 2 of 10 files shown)


ShigekiKarita:
Maybe the above codecov's comment is outdated. I found the total coverage is **78.37%** at this commit (from 65.35% at master)
https://codecov.io/gh/espnet/espnet/tree/986cbb9890bf8279a6f59e1d516fa5d1905a66e5

kamo-naoyuki:
> Merging #2254 into master will increase coverage by 13.02%.

Great.

- Add an installation check for coverage in egs/mini_an4/*/path.sh because it's not installed by default.
- About ESPnet2?

```sh
# egs2/TEMPLAE/asr1/asr.sh
python=python3
...
$(python) -m espnet2.bin.asr_train ...
```

```sh
# ci/test_integration.sh
...
./run.sh --python "coverage run --append" ...
```

ShigekiKarita:
>   Add an installation check for coverage in egs/mini_an4/*/path.sh because it's not installed by default.

Your suggestion `--python "coverage run --append"` (dependency injection?) for ESPnet2 looks better in other cases. So, instead of path.sh checks, I added the cmd opt.

> About ESPnet2?

However, I felt the `asr.sh`  (and tts.sh) looks too complicated for me to simply replace all the "python3" into `${python}`. And I'm not familiar with sanity results. Can you update it after my PR?

kamo-naoyuki:
> However, I felt the `asr.sh` (and tts.sh) looks too complicated for me to simply replace all the "python3" into `${python}`. And I'm not familiar with sanity results. Can you update it after my PR?

okay

PR #2255, opened 2020-08-02
Support larger num_iters_per_epoch than the number of batches in small corpus

kamo-naoyuki:
- Before: If num_iters_per_epoch than the number of batches in the corpus, num_iters_per_epoch is ignored.
    ```python
    if num_iters_per_epoch> len(batches):
        num_iters_per_epoch = len(batches)
    ```
- After: Support if


kan-bayashi:
Cool!

PR #2257, opened 2020-08-02
Support torch native automatic mixed precision for espnet2

kamo-naoyuki:
- Support torch.cuda.amp, which is implemented from torch=1.6.0
   - We can use it with `--use_amp True`
- Remove apex.amp (=We can't use amp if torch<1.6)
   - `--train_dtype O0, O1, O2, O3` are removed.
- Disable autocast of the feature extraction part in asr model and tts model.
   - Probably, we should enable it if the feature extractor part has learnable part, but they don't have now.
   - I keep enh model as it is because I don't know how we should.

Note that RNN model doesn't work with amp now ( pytorch/pytorch#36428 )

@ShigekiKarita or @kan-bayashi 
Could you cross check this implementation? https://pytorch.org/docs/stable/notes/amp_examples.html#amp-examples

### DIFF
--- a/doc/espnet2_training_option.md
+++ b/doc/espnet2_training_option.md
@@ -106,7 +106,7 @@ Checkpoint includes the following states.
 - Optimizer states
 - Scheduler states
 - Reporter state
-- apex.amp state
+- torch.cuda.amp state (from torch=1.6)
 
 ## Change logging interval
 The result in the middle state of the training will be shown by the specified number:
@@ -393,18 +393,12 @@ python -m espnet.bin.asr_train --accum_grad 2
 - Some statistical layers based on mini-batch e.g. BatchNormalization
 - The case that the batch_size is not unified for each iteration.
 
-## Using apex: mixed-precision training
+## Automatic Mixed Precision training
+
 ```bash
-python -m espnet.bin.asr_train --train_dtype float16 # Just training with half precision
-python -m espnet.bin.asr_train --train_dtype float32 # default
-python -m espnet.bin.asr_train --train_dtype float64
-python -m espnet.bin.asr_train --train_dtype O0 # opt_level of apex
-python -m espnet.bin.asr_train --train_dtype O1 # opt_level of apex
-python -m espnet.bin.asr_train --train_dtype O2 # opt_level of apex
-python -m espnet.bin.asr_train --train_dtype O3 # opt_level of apex
+python -m espnet.bin.asr_train --use_amp true
 ```
 
-Note that you need to install apex manually to use mixed-precision: https://github.com/NVIDIA/apex#linux
 
 ## Reproducibility and determinization
 There are some possibilities to make training non-reproducible.
--- a/espnet2/asr/espnet_model.py
+++ b/espnet2/asr/espnet_model.py
@@ -1,3 +1,5 @@
+from contextlib import contextmanager
+from distutils.version import LooseVersion
 from typing import Dict
 from typing import List
 from typing import Optional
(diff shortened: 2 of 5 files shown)


ShigekiKarita:
How about wrapping RNN `with autocast(enabled=False):`?

kamo-naoyuki:
Maybe the problem of RNN will be fixed soon. It's their high priority matter.

kamo-naoyuki:
I'll merge.

Gqwert123:
Hello, when using AMP，we must use python1.6. But when use pytorch1.6，builin-ctc will be used not warp-ctc. This will
cause the error [](https://github.com/espnet/espnet/pull/1421). Some loss will be nan or inf by using my own data. The only way is "egs/wsj/asr1/local/filtering_samples.py "?  And another question is how to set the parameter "args.min_io_delta" ?

kamo-naoyuki:
Please open an issue if you have a question. If you talk in some PRs, it's hard to find it for the other people even if they are interested in it.

Sorry, I didn't have enough time to investigate amp. If you'll try it and report the result, it would help me.

It might be better to disable auto casting for ctc loss calculation:  
https://github.com/espnet/espnet/blob/180af1103997d4eb42ece85cdb92b049db5570c7/espnet2/asr/ctc.py#L49

@ShigekiKarita Do you have some comments?

> The only way is "egs/wsj/asr1/local/filtering_samples.py "? And another question is how to set the parameter "args.min_io_delta" ?

No, we don't have something like `min_io_delta` option in common recipe. Please implement it if you need it. It's easy.

@sw005320 Do we need the option in common recipe? I don't know much about the motivation.

sw005320:
> Please open an issue if you have a question. If you talk in some PRs, it's hard to find it for the other people even if they are interested in it.
> 
> Sorry, I didn't have enough time to investigate amp. If you'll try it and report the result, it would help me.
> 
> It might be better to disable auto casting for ctc loss calculation:
> https://github.com/espnet/espnet/blob/180af1103997d4eb42ece85cdb92b049db5570c7/espnet2/asr/ctc.py#L49
> 
> @ShigekiKarita Do you have some comments?
> 
> > The only way is "egs/wsj/asr1/local/filtering_samples.py "? And another question is how to set the parameter "args.min_io_delta" ?
> 
> No, we don't have something like `min_io_delta` option in common recipe. Please implement it if you need it. It's easy.
> 
> @sw005320 Do we need the option in common recipe? I don't know much about the motivation.

The script is to exclude the utterance of which (subsampled) input length would be shorter than the output length with some mergins, which yields to avoid the problem pointed by the above during the CTC loss computation. We added it to the WSJ recipe since it tends to happen because the WSJ recipe uses the character-based modeling and the output length is longer. We don't observe such issues so much in the other recipes because we often use spm, which makes the output length shorter, and we don't encounter the issue so much. We can make it a common stage in a recipe, but mostly we don't need it.

ShigekiKarita:
If you had no error with fp32, can you try nvidia apex.amp instead? The apex.amp excludes ctc from fp16 but maybe pytorch 1.6 does not.
https://github.com/NVIDIA/apex/blob/93cabd5df060b3783228dd75941ad359be5d391d/apex/amp/lists/functional_overrides.py#L67

> Remove apex.amp (=We can't use amp if torch<1.6) 

I don't  know why @kamo-naoyuki removed this. apex.amp could run with pytorch1.6.

For warpctc, we manually cast the loss to fp32 because it didn't support fp16, we should do this in builtin-ctc too.
https://github.com/espnet/espnet/blob/01eaefef7698453d3c6497c17ba679c98e9f5f85/espnet/nets/pytorch_backend/ctc.py#L106

ShigekiKarita:
In addition, I think more layers should be fp32 when you use builtin pytorch amp(?). For example, loss function and normalization are manually set to fp32 in apex and fairseq.
- https://github.com/NVIDIA/apex/blob/93cabd5df060b3783228dd75941ad359be5d391d/apex/amp/lists/functional_overrides.py#L29
- https://github.com/pytorch/fairseq/blob/master/fairseq/modules/fp32_group_norm.py

kamo-naoyuki:
Apex.amp has many problems and we can't use it in espnet2, e.g. it doesn't support disable auto casting for some parts. We need this feature because espnet2 extracts features on the fly. 

If both apex.amp and builtin amp are supported, the code becomes too complex. I don't have no resources to maintain it. Now builtin amp is upper replacement of apex.amp. I think it's enogth to support apex.amp

sean37:
I got the same issue working on fp16 based training using espnet2

> espnet2.bin.asr_train --train_dtype float16 --use_amp true

> RuntimeError: "ctc_loss_cuda" not implemented for 'Half'

Is there any update or working solution to do fp16 based training?

kamo-naoyuki:
If you want to use mix precision training, just `--use_amp true`. 
There are no ways to half precision training with ctc loss.

PR #2259, opened 2020-08-03
Add metric option to checkpoint averaging for Transformer

hirofumi0810:
This PR adds some metric options (e.g., BLEU) to pick up top checkpoints for checkpoint averaging in stage-5.
It does not affect the existing recipes unless the metric is specified.


hirofumi0810:
@sw005320 The coverage was recovered now. This was due to the conflict in the first commit.

sw005320:
I see. Thanks!

PR #2260, opened 2020-08-03
Update comments in MultiHeadAttention

placebokkk:
The description about shape is wrong, which may confuse user. Also fix typos in conformer subsampling class.

### DIFF
--- a/espnet/nets/pytorch_backend/transformer/encoder_layer.py
+++ b/espnet/nets/pytorch_backend/transformer/encoder_layer.py
@@ -57,7 +57,7 @@ class EncoderLayer(nn.Module):
         """Compute encoded features.
 
         :param torch.Tensor x: encoded source features (batch, max_time_in, size)
-        :param torch.Tensor mask: mask for x (batch, max_time_in)
+        :param torch.Tensor mask: mask for x (batch, 1, max_time_in)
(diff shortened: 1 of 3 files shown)


espnet/nets/pytorch_backend/transformer/encoder_layer.py
```diff
@@ -71,7 +71,6 @@ def forward(self, x, mask, cache=None):
             assert cache.shape == (x.shape[0], x.shape[1] - 1, self.size)
             x_q = x[:, -1:, :]
             residual = residual[:, -1:, :]
-            mask = None if mask is None else mask[:, -1:, :]
```
sw005320, line 74: Could you explain this change as well?

placebokkk: When mask shape is [batch, 1, max_length], mask[:, -1:, :] is same with mask. I did not find the case that mask.shape[1] != 1 in the code. Please correct me if I am wrong. Thanks!

placebokkk: Sorry I just find the mask is also used in LM transformer part. I will read the code thoroughly and then refine the comments. close this PR now.

sw005320: OK, but you can make a separate PR to just fix the documentation part.

PR #2269, opened 2020-08-04
aidatatang conformer ESPnet1 recipe

nzhoward:
added conformer recipe and results

### DIFF
--- a/egs/aidatatang_200zh/asr1/RESULTS.md
+++ b/egs/aidatatang_200zh/asr1/RESULTS.md
@@ -1,6 +1,35 @@
+# Conformer result
+
+## MODEL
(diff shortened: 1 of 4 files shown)


sw005320:
Very good!!!
Maybe we can make the conformer set up as default? (change the pointer of the symlink of https://github.com/espnet/espnet/blob/master/egs/aidatatang_200zh/asr1/conf/train.yaml)

nzhoward:
> Very good!!!
> Maybe we can make the conformer set up as default? (change the pointer of the symlink of https://github.com/espnet/espnet/blob/master/egs/aidatatang_200zh/asr1/conf/train.yaml)

done

sw005320:
Many thanks!!

PR #2299, opened 2020-08-13
ngram single decoding bug fix

qmpzzpmq:
it is a basic bug fix, I will make a unit test later.
or you guys want to I merge them together?


sw005320:
Many thanks!
I’ll just merge it, and yes it would be great if you make a unit test in a later PR to avoid further bugs.

Issue #2192, opened 2020-07-21
CTC-segmentation of large corpora


lumaku:
In our work, we used the segmentation with CTC as originally proposed by Graves et al. in his original publication on CTC (for protein structures etc.).

With that, our approach to CTC-segmentation uses an ASR model pre-trained with CTC to boot-strap more training data from unsegmented or unlabeled training data.
Utterances are then extracted from label probabilities obtained from the network trained with CTC to determine segment alignments.
We used this to generate over 1700h of labeled speech training data, see https://arxiv.org/abs/2007.09127 .

The source code for this method can be found here: https://github.com/cornerfarmer/ctc_segmentation
Please advise on how to merge this with ESPnet, e.g., by adding it to `espnet/utils`.

sw005320:
This is very impressive work (and quite timely for me, @jnishi, and @simpleoier).
Many thanks for introducing it!
And, yes, we definitely want to include it in espnet.
(As you may notice), we recently made a prototype [alignment tool](https://github.com/espnet/espnet/blob/master/utils/ctc_align_wav.sh), but your tool is more complete.

What I'm thinking is to include it in `espnet/utils` as you mentioned or regarding it as a recipe. 
The current tool is rather complicated as a stand-alone tool, and my feeling is that we can put some of your scripts in `espnet/utils/align_xxx.{py.sh}` and then it is probably better to create some recipes, e.g., `egs/tedlium2/align1/`.
I and @jnishi have a similar project in Japanese, and we can follow or help your development.

@jnishi, could you also take a look at this tool and give us some comments?

jnishi:
That sounds great. I agree with @sw005320.  It is good to add their core method to `espnet/utils` and align with pre-trained model listed in https://github.com/espnet/espnet#asr-demo. After that, it is good to add data preparation code as recipes to align with the model trained by users. 

@sw005320 I think the recipe of an alignment tool requires the corresponding asr recipe. Which is better that the alignment recipe have the (symbolic) link to the asr recipe or the alignment recipe includes asr training flow?

lumaku:
@sw005320 Yes, I noticed your current work on `ctc_align`; that was also the motivation to post our work.

[...]

PR #2301, opened 2020-08-13
CTC segmentation

lumaku:
Integrate [CTC segmentation](https://arxiv.org/abs/2007.09127) into the ESPnet toolkit, as previously discussed in #2192, extending the previous `ctc_align` tool.

The front-end to this algorithm is in `espnet/bin/asr_align.py`. Input is json formatted, as in regular ESPnet ASR recipes, and its output is a plain `segments` file with utterance positions in the audio files. The core algorithm can be found in `espnet/utils/ctc_segmentation.py`.

**CTC segmentation parameters:** There are several notable parameters to adjust the working of the algorithm:

* `min_confidence_score` (in the recipe `run.sh`): This is minium confidence score in log space as described in the paper. Utterances with a low confidence score are discarded. This parameter may need adjustment depending on dataset, ASR model and language. For the german ASR model, a value of -1.5 worked very well, but for TEDlium, a lower value of about -5.0 seemed more practical.
* `--min-window-size`: Minimum window size considered for a single utterance. The current default value should be OK in most cases.

Two parameters are needed to correctly map the frame indices to a time stamp in seconds:
* `--subsampling-factor`: If the encoder sub-samples its input, the number of frames at the CTC layer is reduced by this factor. A BLSTMP encoder with subsampling 1_2_2_1_1 has a subsampling factor of 4. 
* `--frame-duration`: This is the non-overlapping duration of a single frame in milliseconds (the inverse of frames per millisecond).
* Currently, if those parameters deviate from default values, they have to be set manually. If the ESPnet api supports it, those parameters could also be automatically determined.

**Localization:** The character set is taken from the model dict, i.e., usually are generated with SentencePiece. An ASR model trained in the corresponding language and character set is needed. For asian languages, no changes to the CTC segmentation parameters should be necessary. One exception: If the character set contains any punctuation characters, "#", or the Greek char "ε", adapt the setting in `CtcSegmentationParameters` in `segmentation.py`.

**ASR recipe:** We provided an ASR-style demo recipe for the TEDlium2 test and dev set. It should be easily adapted to any other audio files.


sw005320:
Many thanks, @lumaku!!!
As you know, we have a lot of requests about this function, and I'll do my best to help you to make this PR in a good shape.

@simpleoier and @jnishi, could you start to review it?

kamo-naoyuki:
Your module looks nice. How about creating another repository to manage ctc_segmentation? Then, it can be used in non espnet tool easily. If you are not familiar to distributing python package, I can help you.

```sh
pip install ctc_segmentation
```

```python
from ctc_segmentation import ctc_segmentation
ctc_segmentation(config, lpz, ground_truth_mat)
```

egs/tedlium2/align1/run.sh
```diff
@@ -0,0 +1,194 @@
...
+            --data-json ${dumpdir}/${rtask}/delta${do_delta}/data.json \
+            --model ${align_model} \
+            --subsampling-factor ${subsampling_factor} \
+            --api ${api} \
+            --utt-text data/${rtask}/utt_text \
+            --output data/${rtask}/aligned_segments || exit 1;
+    done
+fi
+
+if [ ${stage} -le 3 ] && [ ${stop_stage} -ge 3 ]; then
```
simpleoier, line 184: Should be 4 here?

simpleoier:
LGTM!

egs/tedlium2/align1/run.sh
```diff
@@ -0,0 +1,194 @@
...
+if [ -z "${align_config}" ]; then
+    align_config=$(find ${download_dir}/${models} -name "decode*.yaml" | head -n 1)
+fi
+if [ -z "${dict}" ]; then
+    dict=$(find ${download_dir}/${models}/data/lang_*char -name "*.txt" | head -n 1) || dict=""
+
+    if [ -z "${dict}" ]; then
+        mkdir ${download_dir}/${models}/data/lang_autochar/ -p
+        model_config=$(find ${download_dir}/${models}/exp/*/results/model.json | head -n 1)
+        cat $model_config | python -c 'import json,sys;obj=json.load(sys.stdin);[print(char + " " + str(i + 1)) for i, char in enumerate(obj[2]["char_list"])]' > ${download_dir}/${models}/data/lang_autochar/dict.txt
```
jnishi, line 101: @sw005320 I wonder if `python` command uses python in which the ESPnet is installed. Is it correct to use the `python` command to run python scripts in the ESPnet environment?

sw005320:
There are several CI errors (e.g., https://github.com/espnet/espnet/pull/2301/checks?check_run_id=984626673#step:7:109)
You can also run the same test in your local by following https://github.com/espnet/espnet/blob/master/CONTRIBUTING.md#unit-testing

egs/tedlium2/align1/run.sh
sw005320: I think this would be OK.

egs/tedlium2/align1/run.sh
lumaku: Yes, that is correct.

egs/tedlium2/align1/local/download_data.sh
```diff
@@ -0,0 +1,38 @@
...
+    # because of the --continue switch.
+    wget --continue http://www.openslr.org/resources/19/TEDLIUM_release2.tar.gz || exit 1
+    tar xf "TEDLIUM_release2.tar.gz"
+  else
+    echo "$0: not downloading or un-tarring TEDLIUM_release2 because it already exists."
+  fi
+fi
+
+
+num_sph=$(find TEDLIUM_release2/ -name '*.sph' | wc -l)
```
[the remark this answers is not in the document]
lumaku: Added!

egs/tedlium2/align1/run.sh
```diff
@@ -0,0 +1,194 @@
...
+download_dir=models
+align_model=
+align_config=
+align_dir=align
+api=v1
+dict=
+
+if [ ${stage} -le -2 ] && [ ${stop_stage} -ge -2 ]; then
+    echo "stage -1: Data Download"
+    local/download_data.sh
```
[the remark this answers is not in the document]
lumaku: Done.

egs/tedlium2/align1/run.sh
lumaku: Changed `python` to `${python}` in this case.

sw005320:
You can pip install black and `$ black espnet espnet2 test utils setup.py egs*/*/*/local egs2/TEMPLATE/asr1/pyscripts` to fix most of format issues automatically ;)

lumaku:
Thank you for the feedback!

Regarding a distribution package. In order to make it easier to package the CTC segmentation algorithm separately later, ESPnet-specific functions (`ctc_align`) are now separate from the CTC core functions (`CtcSegmentationParameters`, `ctc_segmentation`, `prepare_text`, `determine_utterance_segments`). And the Cython-module is currently imported via pyximport as a temporary solution; distribution packages have this technical advantage, as they can cythonize the module in its `setup()`. Porting this module to other toolkits is probably possible without major changes in this state, given that the CTC likelihoods are universal to CTC-based systems and the text is read in as list of strings.

Temporarily, I removed the unit test for `ctc_align_wav`, as its interface changed. A unit test for the new `espnet/bin/asr_align.py` would be the same regarding to setting up the data directory. The algorithm additionally needs a list of utterances (not necessary to fully cover the transcription) and its ouput can be checked at the `segments` file. An example `utt_text` for the small unit test:
```
PART_1 SALE OF THE HOTELS
PART_2 SELL OFF ASSETS AND CONCENTRATE
```

lumaku:
The recent commits added the `--use-dict-blank` parameter in `asr_align.py`. This copies over the Blank character from the dictionary to the configuration. Useful if in the model dictionary e.g. "<blank>" instead of the default "_" is used - if the Blank in the configuration and in the dictionary mismatch, the algorithm raises an IndexError at backtracking.

lumaku:
The pull request would be OK to merge from our side.
@sw005320 The codecov report seems to be not yet fully happy. What else is needed here?

kamo-naoyuki:
We should not include .pyx file in the espnet repo. It may causes several problems. Please move it to another repository and import it, at least.

kamo-naoyuki:
and, pyximport is not allowed in any cases. Please compile.

lumaku:
@kamo-naoyuki Done.
The code is now at https://github.com/lumaku/ctc-segmentation
If you require any changes to the algorithm, e.g. for Japanese, please let me know.

kamo-naoyuki:
Great! Really thanks!

Will you register it to pypi? If so, please add it in [setup.py](https://github.com/espnet/espnet/blob/3b68f3a7ef6b334003d1a69279a11765f3ce3f3d/setup.py#L8-L57)
If not, please add the installation flow in https://github.com/espnet/espnet/blob/master/tools/Makefile

(Now you module can't be installed by pip. please also check it.
`pip install git+https://github.com/lumaku/ctc-segmentation` )

lumaku:
Thanks for pointing that out, it seems to be a compatibility issue. Rebuilding the .pyx file seems to have fixed it. I have no idea why that happened.
The package is in PyPI as `ctc-segmentation`. I added this to the ESPnet `setup.py`.

lumaku:
OK!

Regarding test coverage, tests are in `test_utils/integration_test_asr_align_wav.bats` and in ` test/test_ctc_segmentation.py`.

kamo-naoyuki (approved):
LGTM. Many thanks, this is great job!

sw005320:
Many thanks!
I'll merge this PR.
Could you make a second PR to add some documents about this function to the following places?
- https://github.com/espnet/espnet#key-features
- https://github.com/espnet/espnet#results-and-demo

PR #2303, opened 2020-08-14
Remove channel: setup_anaconda.sh

kamo-naoyuki:


Gqwert123:
@kamo-naoyuki   can you tell  the reason you did it ?

kamo-naoyuki:
There are no reason to use anaconda channel, so we should use the default channel.

kamo-naoyuki:
Note that anaconda channel is just a mirror of main channel.
https://docs.anaconda.com/anaconda/user-guide/tasks/using-repositories/

PR #2324, opened 2020-08-18
sliding window multihead attention

qmpzzpmq:
add a sliding multihead attention
paper: https://arxiv.org/abs/1909.13037 figure 2

### DIFF
--- a/espnet/nets/pytorch_backend/transducer/blocks.py
+++ b/espnet/nets/pytorch_backend/transducer/blocks.py
@@ -21,6 +21,7 @@ from espnet.nets.pytorch_backend.transducer.vgg2l import VGG2L
 from espnet.nets.pytorch_backend.transformer.attention import (
     MultiHeadedAttention,  # noqa: H301
     RelPositionMultiHeadedAttention,  # noqa: H301
+    ChunkedMultiHeadedAttention,
 )
 from espnet.nets.pytorch_backend.transformer.encoder_layer import EncoderLayer
 from espnet.nets.pytorch_backend.transformer.embedding import (
@@ -305,7 +306,7 @@ def build_input_layer(
         raise NotImplementedError("Support: linear, conv2d, vgg2l and embed")
 
 
-def build_transformer_block(net_part, block_arch, pw_layer_type, pw_activation_type):
+def build_transformer_block(net_part, block_arch, pw_layer_type, pw_activation_type, enc_win_left=0, enc_win_right=0):
     """Build function for transformer block.
 
     Args:
@@ -339,12 +340,17 @@ def build_transformer_block(net_part, block_arch, pw_layer_type, pw_activation_t
 
     if net_part == "encoder":
         transformer_layer_class = EncoderLayer
+        if enc_win_left != 0 or enc_win_right != 0:
+            MHA = ChunkedMultiHeadedAttention(heads, d_hidden, att_dropout_rate, enc_win_left, enc_win_right)
+        else:
+            MHA = MultiHeadedAttention(heads, d_hidden, att_dropout_rate)
     elif net_part == "decoder":
         transformer_layer_class = DecoderLayer
+        MHA = MultiHeadedAttention(heads, d_hidden, att_dropout_rate)
 
     return lambda: transformer_layer_class(
         d_hidden,
-        MultiHeadedAttention(heads, d_hidden, att_dropout_rate),
+        MHA,
         pw_layer(*pw_layer_args),
         dropout_rate,
     )
@@ -526,11 +532,15 @@ def build_blocks(
         if block_type == "tdnn":
             module = build_tdnn_block(blocks_arch[i])
         elif block_type == "transformer":
+            enc_win_left = blocks_arch[i].get('win_left', 0)
+            enc_win_right = blocks_arch[i].get('win_right', 0)
             module = build_transformer_block(
                 net_part,
                 blocks_arch[i],
                 positionwise_layer_type,
                 positionwise_activation_type,
+                enc_win_left,
+                enc_win_right,
             )
         elif block_type == "conformer":
             module = build_conformer_block(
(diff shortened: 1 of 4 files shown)


qmpzzpmq:
@b-flo that is what I mentioned. do you think how merge into your part?

b-flo:
> @b-flo that is what I mentioned. do you think how merge into your part?

Sorry, I totally forgot this PR until I received emails about your last commits... I'm wrapping up my draft for customizable architecture in espnet2 and I'll review your PR! Maybe later this day or tomorrow (CET time).
If you need help for some remaining work feel free to ask me!

qmpzzpmq:
@b-flo just relax, I haven't finished it totally.
I will let you know when I finish it.

qmpzzpmq:
@b-flo done

qmpzzpmq:
baseline:
accum-grad=2 1 with 4 GPU
dev:24.1/53.0
test:25.6/56.5

sliding window multihead  attention:
accum-grad=2 1 with 4 GPU
dev:26.3/56.5
test:26.4/57.8

espnet/nets/pytorch_backend/transducer/blocks.py
```diff
@@ -526,11 +541,15 @@ def build_blocks(
         if block_type == "tdnn":
             module = build_tdnn_block(blocks_arch[i])
         elif block_type == "transformer":
+            enc_win_left = blocks_arch[i].get("win_left", 0)
```
b-flo, line 544: If `enc_win_left` and `dec_win_left` are block parameters, you should get values within build_transformer_block! However I think it should be a global parameter passed to `build_blocks` as we won't change window between decoder blocks? Also, as we can mix blocks, some checking should be done I guess

espnet/nets/pytorch_backend/transducer/blocks.py
```diff
@@ -305,7 +306,14 @@ def build_input_layer(
         raise NotImplementedError("Support: linear, conv2d, vgg2l and embed")
 
 
-def build_transformer_block(net_part, block_arch, pw_layer_type, pw_activation_type):
+def build_transformer_block(
+    net_part,
+    block_arch,
+    pw_layer_type,
+    pw_activation_type,
+    enc_win_left=0,
```
b-flo, line 314: ditto

b-flo (commented):
LGTM to me except for comments!
Thank you again btw!

espnet/nets/pytorch_backend/transformer/attention.py
```diff
@@ -204,3 +204,97 @@ def forward(self, query, key, value, pos_emb, mask):
...
+        left_context (int): Sliding windows left length
+        right_context (int): Sliding windows right length
+
+    """
+
+    def __init__(self, n_feat, left_context=5, right_context=2):
+        """Construct a SlidingWindow object."""
+        super(SlidingWindow, self).__init__()
+        self.n_feat = n_feat
+        self.left_context = left_context
```
b-flo, line 226: Btw, you should maybe assert left/right context have correct values? You only check value is != 0 during block creation

qmpzzpmq: what means of "correct value", from my opinion, all >0 int value is "correct"

espnet/nets/pytorch_backend/transducer/blocks.py
qmpzzpmq: 1. Any suggestion for " get values within build_transformer_block"?
2. these window is for restrict the attention in encoder, I don't decoder attention need it. 
3. I will add some checks anyway, but I am not sure whether it satisfied your suggestion

espnet/nets/pytorch_backend/transformer/attention.py
b-flo: Negative values and kernel size bigger than input size are allowed/handled?

espnet/nets/pytorch_backend/transducer/blocks.py
b-flo: 1. It's an optional parameter so you can follow same pattern as dropout parameters:
e.g.: `enc_win_left = block_arch["enc_win_left"] if "enc_win_left" in block_arch else 0.0`
2. Yes, not sure what you implied? You already used `net_part` to select attention class accordly
3. No problem! My suggestion is to define a `sliding_window` with left-right values in `asr_recog.py `to pass to `build_blocks`. If you see any usage where you need to define different sliding window by blocks, we can go with 1.

espnet/nets/pytorch_backend/transformer/attention.py
qmpzzpmq: I don't understand which kernel you specified?

b-flo: Sorry, wanted to say `chunk_size` (left-window, right-window)!

espnet/nets/pytorch_backend/transducer/blocks.py
qmpzzpmq: 1.  done
2.  forget about it
3. good suggestion, the window size actually could be different in train and test phase. But it need a smart way to implement it. any  idea? since all model parameters in decoding phase is read in model.json

espnet/nets/pytorch_backend/transformer/attention.py
qmpzzpmq: since chunk_size is the sum of left and right window, I just add a check with these two.

espnet/nets/pytorch_backend/transducer/blocks.py
b-flo: It was misleading to point out `asr_recog.py`, I'm okay to keep the same values for training and inference for now. I'll think about it.
What I wanted to highlight is that if you define `enc_win_left` and `enc_win_right` as transformer block parameters, you can change values between blocks. If it's intended, let's keep it, otherwise it should be defined as a "global" parameter (in `e2e_asr_transducer` for training and `asr_recog` for inference)

qmpzzpmq: I thinks these parameters could be different in each block. 
Yes, intended.

xzwworkplace:
为什么我把这种注意机制用到transoformer里，训练时长和解码时长反而变长了呀？

PR #2332, opened 2020-08-19
Add installers for sctk and sph2pipe and create tools/extra_path.sh

kamo-naoyuki:
- New installers: install_sctk.sh, install_sph2pipe.sh
  - sclite and sph2pipe become independent from Kaldi installation.
- `tools/extra_path.sh`:  Set all environment variable for extra tools.  From now, `path.sh` must refer this file instead of directly setting $PATH


kamo-naoyuki:
I also created `egs2/*/*/local/path.sh`, which set the environment variables and check required tools for local task.
Now all `path.sh` in egs2/ is symlink to `TEMPLATE/*/path.sh`.

PR #2356, opened 2020-08-24
remove ff_scale from conformer constructor arguments

koji-okabe-hub:
I found a bug of conformer.

The constructor of EncoderLayer has 3 optional arugments (ff_scale, normalize_before, concat_after).
https://github.com/espnet/espnet/blob/3ce695ea7098b8df8c5bdec513b4aafd4885a5e8/espnet/nets/pytorch_backend/conformer/encoder_layer.py#L42-L53

However, it is called with only latter ones (normalize_before and concat_after) without named arguments.
https://github.com/espnet/espnet/blob/3ce695ea7098b8df8c5bdec513b4aafd4885a5e8/espnet/nets/pytorch_backend/conformer/encoder.py#L173-L182

As a result, ff_scale is initialized by normalize_before, normalize_before is initialized by concat_after, and concat_after is initialized by the default value (=False).

ff_scale is 0.5 if macaron_type FFs are used, otherwise 1.0.
I think there is no other reason to change the ff_scale.
So I'd like to suggest to remove ff_scale from the constructor arguments.


sw005320:
Thanks, @koji-okabe-hub !
@pengchengguo, could you review this PR?

pengchengguo:
Yes, I think it is great to remove `ff_scale` from the constructor arguments. I add it to the construct function because I want to find the influence of different `ff_scale` values. But I think most of the papers just use `ff_scale=0.5` for macaron style.

sw005320:
There are some circle CI errors.
Could you marge the latest master and fix the CI errors?
Then, I'll merge it.

sw005320:
Many thanks, @koji-okabe-hub!

PR #2377, opened 2020-08-29
Implement batch_score() method for ASR decoder and LM

kamo-naoyuki:
- Implement Transformer LM for espnet2.
- Put  batch_score() method in ASR transformer decoder and RNNLM and Transformer LMfor espnet2.
- Copied LightweightConvolution and DynamicConvolution for decoder part for espnet2.. (Encoder part is not yet.)

@ShigekiKarita Could you check batch_score part? I read almost all your BeamSearch part today, so maybe it's okay, but I'm not sure.


ShigekiKarita:
LGTM. If you will change the integration test with batch scorers, that would be the best.

espnet2/lm/transformer_lm.py
```diff
@@ -0,0 +1,131 @@
...
+                torch.stack([states[b][i] for b in range(n_batch)])
+                for i in range(n_layers)
+            ]
+
+        # batch decoding
+        h, _, states = self.encoder.forward_one_step(
+            self.embed(ys), self._target_mask(ys), cache=batch_state
+        )
+        h = self.decoder(h[:, -1])
+        logp = h.log_softmax(dim=-1)
```
ShigekiKarita, line 127: maybe you can share this part with `score`

test/espnet2/asr/decoder/test_transformer_decoder.py
```diff
@@ -26,14 +50,132 @@ def test_TransformerDecoder_backward(input_layer, normalize_before, use_output_l
...
+    [
+        TransformerDecoder,
+        LightweightConvolutionTransformerDecoder,
+        LightweightConvolution2DTransformerDecoder,
+        DynamicConvolutionTransformerDecoder,
+        DynamicConvolution2DTransformerDecoder,
+    ],
+)
+def test_TransformerDecoder_init_state(decoder_class):
+    decoder = decoder_class(10, 12)
```
ShigekiKarita, line 64: I'm afraid that this model is too large for unit tests. For example, default `linear_units` is 2048. That should be smaller.

kamo-naoyuki: I see

kamo-naoyuki:
> LGTM. If you will change the integration test with batch scorers, that would be the best.

Yes, I agree. I'll change it at the next PR.

ShigekiKarita:
For simplicity, we don't need `score` or `ScorerInterface` anymore in espnet2 because we have `BatchScorerInterface` and `BatchBeamSearch` now.

espnet2/lm/transformer_lm.py
kamo-naoyuki: Yes, but... I think, it' okay? Not the duplication is not so large.

ShigekiKarita: OK

kamo-naoyuki:
We can use BatchScorerInterface for non batch scorer mode? Do you mean `batch_score` can fallback to `score()`?

ShigekiKarita:
Yes. The `BatchScorerInterface` has for-loop based default implementation of `batch_score` using `score` https://github.com/espnet/espnet/blob/eca726a4a1866116779a77108662e1e638800943/espnet/nets/scorer_interface.py#L99-L128

kamo-naoyuki:
Okay, then I removed score() method.

Issue #2365, opened 2020-08-26
ctc-segmentation installation is failed with pip in colab


kan-bayashi:
Recently `ctc-segmentation` is addded in requirements (#2301).
However, this causes the error in the installation with pip in colab.
(You can reproduce in [this notebook](https://colab.research.google.com/github/espnet/notebook/blob/master/espnet2_tts_realtime_demo.ipynb).)
This affects the demo notebook so I want to fix it as soon as possible.

@lumaku Could you check it?

```bash
!pip install -vvv ctc-segmentation
```

```bash
Created temporary directory: /tmp/pip-ephem-wheel-cache-334qq_lm
Created temporary directory: /tmp/pip-req-tracker-dwauuihb
Created requirements tracker '/tmp/pip-req-tracker-dwauuihb'
Created temporary directory: /tmp/pip-install-98c9o0q4
1 location(s) to search for versions of ctc-segmentation:
* https://pypi.org/simple/ctc-segmentation/
Getting page https://pypi.org/simple/ctc-segmentation/
Found index url https://pypi.org/simple
Looking up "https://pypi.org/simple/ctc-segmentation/" in the cache
Request header has "max_age" as 0, cache bypassed
Starting new HTTPS connection (1): pypi.org:443
https://pypi.org:443 "GET /simple/ctc-segmentation/ HTTP/1.1" 304 0
Analyzing links from page https://pypi.org/simple/ctc-segmentation/
  Found link https://files.pythonhosted.org/packages/5c/49/70f81b78a2c67bcfe2e72a8c3be91430f0f822857f5b8f5587cc8a1d0de0/ctc_segmentation-1.0.1.tar.gz#sha256=f146adf31fc225a3594cd52dd57b6da754808b00ca1865aeab0ad978b38909fb (from https://pypi.org/simple/ctc-segmentation/) (requires-python:>=3), version: 1.0.1
  Found link https://files.pythonhosted.org/packages/67/be/5fdbb4e536cfa07affe57f947074a5f5b5325a22aebf4c661e21ca10187c/ctc_segmentation-1.0.2.tar.gz#sha256=84584a4251988414262941293f31d295fd9c406ce1295ea50306fca8a28bdf30 (from https://pypi.org/simple/ctc-segmentation/) (requires-python:>=3), version: 1.0.2
[...]

PR #2383, opened 2020-08-31
Change the source of CTC-segmentation from pypi to git

kan-bayashi:
Temporary fix.

Related #2365


kan-bayashi:
@kamo-naoyuki This requires the recent pip. Is it OK?

kamo-naoyuki:
No, we can't use git+ syntax because it can't be uploaded to pypi server. 
I think just removing is better.

kan-bayashi:
OK. I will chage.

PR #2396, opened 2020-09-01
Add bug report template

sw005320:


kamo-naoyuki:
Why don't you use python -m  torch.utils.collect_env? 
I think, cuda version is necessary at least.

sw005320:
> Why don't you use python -m torch.utils.collect_env?
> I think, cuda version is necessary at least.

Done!

kamo-naoyuki (approved):
LGTM

kan-bayashi (approved):

PR #2400, opened 2020-09-02
Fix score_bleu.sh

### LINKED ISSUE #2363: Bug: score_bleu.sh removes good tokens
Hello,

While reproducing an experiment on MuST-C, I realized that the reference file used for computing BLEU is different from the original reference file: tokens such as `(_Gelächter_)` and `(_Applaus_)` are removed from both the reference and the hypothesis.

After checking `score_bleu.sh`, I think that the issue lies at [these lines](https://github.com/espnet/espnet/blob/d0846517c1a69cd32ac6c09dabd883ba8417ca48/utils/score_bleu.sh#L33-L36) for removing utterance IDs (they seem to remove everything between parentheses).

Could you please check this? The issue seems critical as it may impact results reported in published papers.

Thank you in advance!

hirofumi0810:
Fix #2363.


sw005320:
Many thanks, @hirofumi0810 !

PR #2401, opened 2020-09-02
Added griffin-lim generated audio to tensorboard

Piteryo:
After each epoch model is inferring and griffin-lim generated audio is being added to Tensorboard.


kamo-naoyuki:
Thank you, but your modification assumes tts task's argument, i.e. it's not allowed to refer feats_extract_conf at trainer and abs_task, so we can't merge your PR, sorry.

Trainer class and AbsTask class is carefully designed for generalization, so it might be hard to modify for other than me, so I don't recommend you touching these part. 

However, the flexibility for such monitoring stage is important, so I keep griffin-lim in mind.

@kan-bayashi How do you think about this feature?

Piteryo:
@kamo-naoyuki Ok. But maybe there are other options to get access to parameters for griffin-lim in trainer class?

kamo-naoyuki:
No. Do you have something in mind?
Basically, task depended options are only available from each inheritance of AbsClass, so we need to define such a function in them, but we don't provide such framework for monitoring stage.

kan-bayashi:
Hmm. This feature looks nice but we do not have such a flexible reporter design now.
We need to consider the new interface.

PR #2413, opened 2020-09-04
Support Conformer-based FastSpeech / FastSpeech2

kan-bayashi:
This PR supports Conformer for FastSpeech & FastSpeech2.
I've not yet checked the performance.


kan-bayashi:
I checked conformer-based fastspeech2 with JSUT, CSMSC, and LJspeech recipe.
It seems working well but the training curve itself is almost the same as the Transformer-based.
https://drive.google.com/drive/folders/1R2pGsLEla69oTL5-8wRpRPT15tl-Tjb6?usp=sharing

Issue #2433, opened 2020-09-08
Error: in data/train, recording-ids extracted from segments and wav.scp


happyjin:
I got an error during stage 1 of run.sh using hkust dataset. The error message is as below:
![bug2](https://user-images.githubusercontent.com/18534928/92446945-2e9df800-f1e9-11ea-8b75-3cc47c627fb8.PNG)
Is there anybody who knows how to figure out this error? Thanks in advance!

b-flo:
Hi,

You already opened the same issue a few days ago (#2415), I'm closing the other one then.

@sw005320 It seems you created the recipe a few years back, maybe you have some comments on that? Or maybe you know someone who used it recently?

happyjin:
@b-flo  Thanks for your swift comment. I installed the latest espnet and kaldi, then follow the original recipe of hkust asr dataset. I get stuck during stage 1 as the screenshot displayed. I am wondering why it happens? And more importantly, how to fix this error.

b-flo:
Sorry, I'll let others familiar with the dataset comment on the issue. A general fix would be to use `utils/fix_data_dir.sh` but I don't think such mismatch is expected, something went wrong during data preparation imo.

kamo-naoyuki:
Did you see #2321 ?

speechLearner:
Hi @happyjin - I have faced this error when I accidentally rerun the stage 1; even though the features were extracted during the previous run . I would suggest you clear the train and dev directory and rerun from stage 0 data preparation

happyjin:
@kamo-naoyuki Thanks for your advice regarding #2321 . I follow the problem by following the detailed steps.
@speechLearner Thanks for your suggestion and I also remove the previous generated data etc folders and re-run it to get new folders. It works.

kamo-naoyuki:
@speechLearner Thanks, we must fix it since hkust recipe has too much problem...

kamo-naoyuki:
@speechLearner Thanks, we must fix it since hkust recipe has too much problem...

kamo-naoyuki:
@speechLearner Thanks, we must fix it since hkust recipe has too much problem...

kamo-naoyuki:
@speechLearner Thanks, we must fix it since hkust recipe has too much problem...

kamo-naoyuki:
@speechLearner Thanks, we must fix it since hkust recipe has too much problem...

kamo-naoyuki:
[...]

PR #2440, opened 2020-09-09
Fix HKUST recipe

### LINKED ISSUE #2321: [Dependency] mmseg on egs/hkust/asr1
## Summary

In run.sh stage 0: Data preparation

script requires mmseg on [hkust_segment.py](https://github.com/42qu/mmseg)
however, mmseg using pyhon 2.7
and in [hkust_data_prep.sh](https://github.com/espnet/espnet/blob/master/egs/hkust/asr1/local/hkust_data_prep.sh)
```
#transcripts normalization and segmentation
#(this needs external tools),
python3 -c "import mmseg" 2>/dev/null || \
  (echo "mmseg is not found. Checkout tools/extra/install_mmseg.sh" && exit 1;)

cat $train_dir/transcripts.txt |\
  sed -e 's/<foreign language=\"[a-zA-Z]\+\">/ /g' |\
  sed -e 's/<\/foreign>/ /g' |\
  sed -e 's/<noise>\(.\+\)<\/noise>/\1/g' |\
  sed -e 's/((\([^)]\{0,\}\)))/\1/g' |\
  local/hkust_normalize.pl |\
  python3 local/hkust_segment.py |\
  awk '{if (NF > 1) print $0;}' > $train_dir/text || exit 1;

cat $dev_dir/transcripts.txt |\
  sed -e 's/<foreign language=\"[a-zA-Z]\+\">/ /g' |\
  sed -e 's/<\/foreign>/ /g' |\
  sed -e 's/<noise>\(.\+\)<\/noise>/\1/g' |\
  sed -e 's/((\([^)]\{0,\}\)))/\1/g' |\
  local/hkust_normalize.pl |\
  python3 local/hkust_segment.py |\
  awk '{if (NF > 1) print $0;}' > $dev_dir/text || exit 1;
```
seems use python3 to execute python2 script

I also tried [py3mmseg](https://github.com/bxshi/py3mmseg)
unfortunately, it didn't work.

kamo-naoyuki:
Fix #2321#2433

- I forked py3mmseg for python3.7: https://github.com/kamo-naoyuki/py3mmseg
- Add -L option to find. This recipe is failed if the specified corpus directory is a symlink. We should use find command more carefully.
- Now speed perturbation is applied after feature extraction. It causes the problem of #2433. We should use utils/fix_data_dir.sh more carefully. 
  - Such a ridiculous problem never happens in espnet2. We should move all to espnet2 quickly.

### DIFF
--- a/ci/install.sh
+++ b/ci/install.sh
@@ -19,7 +19,7 @@ ${CXX:-g++} -v
     fi
     make TH_VERSION="${TH_VERSION}"
 
-    make nkf.done moses.done mwerSegmenter.done pesq pyopenjtalk.done
+    make nkf.done moses.done mwerSegmenter.done pesq pyopenjtalk.done py3mmseg.done
     rm -rf kaldi
 )
 . tools/activate_python.sh
--- a/egs/hkust/asr1/local/hkust_data_prep.sh
+++ b/egs/hkust/asr1/local/hkust_data_prep.sh
@@ -7,7 +7,9 @@ if [ $# != 2 ]; then
   echo " $0 /export/corpora/LDC03S04 /export/corpora/LDC03T19"
   exit 1;
 fi
+set -euo pipefail
 
+sph2pipe=sph2pipe
 hkust_audio_dir=$1
 hkust_text_dir=$2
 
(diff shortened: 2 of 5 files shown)


ci/install.sh
```diff
@@ -19,7 +19,7 @@ ${CXX:-g++} -v
     fi
     make TH_VERSION="${TH_VERSION}"
 
-    make nkf.done moses.done mwerSegmenter.done pesq pyopenjtalk.done
+    make nkf.done moses.done mwerSegmenter.done pesq pyopenjtalk.done py3mmseg.done
```
ShigekiKarita, line 22: Why you need this in CI?

kamo-naoyuki: These targets are installed to test the installation script. I'd like to test Makefile because sometimes it was broken due to my modification.
(Sometimes, pesq installation is failed due to the source server, so I'm thinking to delete pesq)

ShigekiKarita: ideally it should be in a different script like .github/workflows/make.yml
```yml
on:
  push:
    paths:
    - tools/Makefile

jobs:
  extra:
    steps:
    - run: make extra
```
to reduce the testing time

kamo-naoyuki: Hmm, ideally, yes. More ideally, we should split integration test and unittest, and linter test  independently.
Okay, I keep it in mind, but not in this PR.

ShigekiKarita: OK. I'm not sure if codecov supports independent coverage xml files from different jobs.

ShigekiKarita: I think linter before testing is good early stopping. Anyway, the independent linter script is nice for local env.

kamo-naoyuki: I see, it sounds hard, so we should keep integration tests and unit tests in one CI for now.

kan-bayashi:
Let us merge this PR.

PR #2451, opened 2020-09-11
Support audio_format=flac.ark, wav.ark

kamo-naoyuki:
This is a new feature to use some **extended** ark format: ark file of flac or wav.

In espnet2, we normally extract features on the fly, thus the source files of training corpus is now separated flac files or wav files for each utterances.

In general, separated files cause more heavy IO stress than glued archive file, so I decided to support flac.ark format in common recipe.

Please use this feature if you have troubles with IO stress.


kamo-naoyuki:
To reduce the number of open and close system call, I added`--max_cached_fd` option. This option is used for only "kaldi_ark" mode.
This option changes the number of file descriptors to be kept as opened for ark files in a scp file. The default value is 32.

Note that the maximum number of file descriptors is limited in unix system. Be careful for the value.

PR #2452, opened 2020-09-11
ADD CSJ RNN pretrained model 

jnishi:
I uploaded the CSJ RNN pre-trained model [here](https://drive.google.com/file/d/1ALvD4nHan9VDJlYJwNurVr7H7OV0j2X9/view?usp=sharing).

It is useful for CTC Segmentation, though it is worse than the transformer model in terms of accuracy.


sw005320:
Many thanks!

PR #2453, opened 2020-09-11
Install warpctc-pytorch wheel when torch version is 1.1, 1.2 or 1.3

ysk24ok:
https://pypi.org/project/warpctc-pytorch/
Now `warpctc-pytorch` is released, which works with PyTorch 1.1, 1.2 and 1.3.

warp-ctc installation is like this,

- `torch >= 1.4`, warpctc will not be installed.
- `1.1 <= torch <= 1.3` , `warpctc-pytorch` wheel will be installed.
- `torch == 1.0` , `warpctc-pytorch10-cudaXX` wheel will be installed.


sw005320:
Many thanks!

PR #2455, opened 2020-09-11
Update CommonVoice for Latest Version

ftshijt:
Similar to the previous setting, we only include langs with more than 20 hrs of validated audio.

Since we got 23 langs (more than 5000 hrs) for this setting, it will need a significantly amount of time to run the model. So I would suggest first merge this shell-level change. And I will start a new PR regarding the trained model & result.


sw005320:
Many thanks!

LGTM.
Yes, I agree with your strategy.
I’ll merge this PR after you fix some CI errors.

PR #2457, opened 2020-09-12
Support Debian9 and CentOS7 in Github Actions

ShigekiKarita:
As circle CI supports those OSs and stopped due to short of 2500 free credits
[image](
https://app.circleci.com/settings/plan/github/espnet/overview?return-to=https%3A%2F%2Fapp.circleci.com%2Fpipelines%2Fgithub%2Fespnet)
I propose Github Actions based redundant CI by its similar functionality https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#jobsjob_idcontainer


egs/mini_an4/asr1/cmd.sh
```diff
@@ -28,7 +28,7 @@
 
 
 # Select the backend used by run.sh from "local", "stdout", "sge", "slurm", or "ssh"
-cmd_backend='local'
+cmd_backend='stdout'
```
kan-bayashi, line 31: I remembered `stdout` does not stop when the error happened.
Is it OK?

ShigekiKarita: I found it was not true. Now stdout.pl correctly stopped when the exist status != 0. See  https://github.com/espnet/espnet/runs/1105039198#step:9:46

I suspect that maybe we forgot `set -euo pipefail`

ShigekiKarita: Well, however, I'm not sure. I will revert this.

kan-bayashi (approved):
LGTM

PR #2466, opened 2020-09-12
[WIP] Recipe for Puebla Nahuatl

ftshijt:
Here some action items:

- [x] set the corpus partition split
- [x] update the downloading from openslr
- [ ] update the pre-trained model link and results


ftshijt:
Unlike Mixtec, the Nahuatl corpus only contains few speakers (less than 8). So it seems not a good idea to split the data by speakers. For now, I offer a random split based on our previous method of corpus partition.

Do you @sw005320  think we should, for example, fix a random partition as the official split of the corpus?

kamo-naoyuki:
@sw005320 @ftshijt I think we should use espnet2 and stop to create new recipe and new feature.

sw005320:
> Unlike Mixtec, the Nahuatl corpus only contains few speakers (less than 8). So it seems not a good idea to split the data by speakers. For now, I offer a random split based on our previous method of corpus partition.

OK

> Do you @sw005320 think we should, for example, fix a random partition as the official split of the corpus?

Year, I think this is good.
We can put the utterance list of the official split in the local directory.

sw005320:
> @sw005320 @ftshijt I think we should use espnet2 and stop to create new recipe and new feature in espnet1.

I agree.
This project is related to the yoloxochitl_mixtec recipe only used in espnet1 and we want to have it in the espnet1 for now, but we can think of making the espnet2 recipe later.
The other espnet1 recipes I'm involved have been started before espnet2 became stable, and we would still have some espnet1 recipes in future PRs, but basically we can ask people to make espnet2 recipes instead of (or in addition to depending on the situations) espnet1 recipes.

sw005320 (commented):
- Please upload the model once you solve the Google drive issue
- The symlink should direct conformer one.

sw005320:
Many thanks!
Please deal with the rest of the items in the later PR.

Issue #2459, opened 2020-09-12
Top 20 slowest tests


ShigekiKarita:
I found running test is really slower now! 

- master: [30:45](https://app.circleci.com/jobs/github/espnet/espnet/20951)
- v 0.9.0: [10:38](https://app.circleci.com/pipelines/github/espnet/espnet/5617/workflows/d0068615-59ab-41f7-9291-59049c3e5186/jobs/18906)
- v 0.8.0: NA
- v 0.7.0: [13:45](https://app.circleci.com/pipelines/github/espnet/espnet/4748/workflows/01d7470c-ea49-4241-bb57-fc4198e76e24/jobs/15785)
- v 0.6.0: [5:09](https://app.circleci.com/pipelines/github/espnet/espnet/2768/workflows/9f28786d-0ee9-44be-854e-f618500e28aa/jobs/9100)

in CircleCI's ubuntu-18 image (running out of free credits!)

I think unit tests should be fast because test-driven dev requires that property (while the integration tests can be slow). Here are the top 20 slowest test files:

- [x] 0 ('test/test_beam_search.py', 536.879999999997) -> 1.91 (#2462)
- [x] 1 ('test/test_batch_beam_search.py', 289.2799999999999) -> 1.45 (#2462)
- [x] 2 ('test/test_e2e_asr_mulenc.py', 216.11999999999998) -> 8.86 (#2480)
- [x] 3 ('test/espnet2/enh/nets/test_beamformer_net.py', 212.21000000002525) -> 12.76s (#2489)
- [x] 4 ('test/espnet2/tts/test_transformer.py', 69.75999999999985) -> 5.30 (#2461)
- [x] 5 ('test/test_e2e_st_transformer.py', 49.809999999999995) -> 8.0 (#2464)
- [x] 6 ('test/test_asr_init.py', 42.18) -> 5.4 (#2444)
- [x] 7 ('test/test_e2e_asr_sa_transducer.py', 38.73000000000001) -> (#98dof55: 43.5) -> 5.89 (#2444)
- [x] 8 ('test/test_e2e_asr_transformer.py', 34.8) -> 5.0 (#2464)
- [x] 9 ('test/test_e2e_asr_conformer.py', 32.07) -> 5.0 (#2464)
- [x] 10 ('test/test_recog.py', 22.89) -> 5.18 (#2486)
- [x] 11 ('test/test_e2e_asr.py', 21.729999999999997) -> 9.79s (#2488)
[...]

PR #2486, opened 2020-09-16
Make test_recog.py faster

ShigekiKarita:
Related to #2459 

- Remove chainer lines because nothing was tested
- Remove etype/dtype variants because there is no difference in batch/non-batch search impl
- Decrease model and data sizes

43.85s -> 5.18s in local


sw005320 (approved):
LGTM!

PR #2491, opened 2020-09-17
shorten test time for tasnet

LiChenda:
this is the PR for testing faster #2459 

Now, testing for TasNet takes 7.01s in my local.


ShigekiKarita (approved):
LGTM. Thanks!

PR #2444, opened 2020-09-09
Transducer v4

b-flo:
**Note: PR re-created in a tentative to fix codecov, refer to #2082 for discussions**

This PR add/modify a bunch of things:

To do / in progress:

- Default beam search:
  - [x] Optimization techniques
    - [x] prediction net caching
- N-Step Constrained beam search (modified version of: https://arxiv.org/pdf/2002.03577.pdf): 
  - [x] RNN-T
  - [x] RNN-T w/ att.
  - [x] T-T
  - [x] VIVOS decode config
  - [x] Voxforge decode config
  - [x] VIVOS results
  - [x] Voxforge results
  - [x] Optimization techniques
    - [x] prediction net caching
  - [ ] Code optimization
- Time Synchronous Decoding (https://ieeexplore.ieee.org/document/9053040): 
  - [x] RNN-T
  - [x] RNN-T w/ att.
  - [x] T-T
  - [x] VIVOS decode config
  - [x] Voxforge decode config
  - [x] VIVOS results
  - [x] Voxforge results
  - [x] Optimization techniques
    - [x] prediction net caching
    - [x] prediction net bashing
    - [x] joint net caching
  - [ ] Code optimization
- Alignment-Length Synchronous Decoding (https://ieeexplore.ieee.org/document/9053040):
  - [x] RNN-T
  - [x] RNN-T w/ att.
  - [x] T-T
  - [x] VIVOS decode config
  - [x] Voxforge decode config
  - [x] VIVOS results
  - [x] Voxforge results
  - [x] ! Fix bug final hypothesis empty with transformer (Edit: lazy patch for now, I'll investigate later)
  - [x] Optimization techniques
    - [x] prediction net caching
    - [x] prediction net bashing
    - [x] joint net caching
  - [ ] Code optimization
- Transformer-Transducer:
  - [x] Customizable architecture
  - [x] TDNN-BN-ReLU blocks for encoder part
  - [x] Conformer blocks for encoder part
  - [x] CausalConv1d for decoder part
- General:
  - [x] CUDA-Warp RNN-T support (https://github.com/1ytic/warp-rnnt) 
  - [x] Beam search interface (first version)
  - [x] Beam search tests (I'll add more tests in another PR I think)
  - [x] Benchmark: decoding speed (edit: only rough benchmark for now, complete will be done after optimizations)
  - [x] Fix mtl_mode (#2162)
  - [x] Fix attention weights saving condition + PlotAttentionReport selection for unified transducer architecture
[...]

PR #2496, opened 2020-09-18
Transducer float16 loss bug fix

GNroy:
I've just encountered a problem with transducer training in the float16 mode. It seems like warprnnt_pytorch doesn't support float16.

This fix casts a tensor of predicted sequences to float32 if necessary.

Thanks,
Aleksandr Laptev.


b-flo (commented):
Thanks for the fix! I'm wondering some things though:
  - I'm not sure what I prefer: casting or raising an error. We should at least warns users I think. @sw005320 do you have an opinion on that?
  - What about others types? To be honest I never tested these options.
  - #2444 introduces `warp-rnnt`, could you also test it with the same options?

If you don't have time for that we can merge this PR and I will make the appropriate modifications on my side, it will conflict with #2444 anyways. Otherwise you will have to wait a bit after we merge the other, test and last modifications are almost done.

sw005320:
> * I'm not sure what I prefer: casting or raising an error. We should at least warns users I think. @sw005320 do you have an opinion on that?

If casting makes it work, I think it's better to do it. Adding a warning is a nice idea.

ShigekiKarita:
I don't think the warning is needed because our ctc and pytorch/nvidia amp also cast implicitly
https://github.com/espnet/espnet/blob/c0d6029ccd3230d84f1d0f8d3413382304238475/espnet/nets/pytorch_backend/ctc.py#L106-L109
https://github.com/NVIDIA/apex/blob/93cabd5df060b3783228dd75941ad359be5d391d/apex/amp/lists/functional_overrides.py#L67

ShigekiKarita:
> What about others types? To be honest I never tested these options.

For warpctc, we already tested here https://github.com/espnet/espnet/blob/6f6452a3891d11b0b0306084dd46a464ddf8e25c/test/test_train_dtype.py#L69

ShigekiKarita (approved):

b-flo:
Thanks for the feedback @ShigekiKarita !

> I don't think the warning is needed because our ctc and pytorch/nvidia amp also cast implicitly

In general I prefer to warn the user when an option he requested for is not working but fine with me! 

> What about others types? To be honest I never tested these options.
> 
> > For warpctc, we already tested here 

I was wondering for `warp-transducer` as we don't test that option currently

GNroy:
Thanks for the feedback!
@b-flo @ShigekiKarita I think this PR can wait until #2444 is merged. After that, I can also test warp-rnnt for being able to handle float16.
Until then I'll try to add tests for transducer to test/test_train_dtype.py

tools/installers/install_warp-rnnt.sh
```diff
@@ -10,22 +10,9 @@ fi
 
 cuda_ver=$1
 
-if [ $cuda_ver != "10.0" ]; then
-    echo "warp-rnnt was only tested with CUDA_VERSION=10.0. Skipping install."
+if [ $cuda_ver != "10.0" ] || [ $cuda_ver != "10.1" ] || [ $cuda_ver != "10.2" ]; then
```
b-flo, line 13:
```suggestion
if ! [[ "$cuda_ver" =~ ^(10.0|10.1|10.2)$ ]]; then
```

I tested all cuda versions, I'm surprised you could install repo with `CUDA_VERSION=10.1` or `10.2`?

tools/installers/install_warp-rnnt.sh
```diff
@@ -10,22 +10,9 @@ fi
...
-fi
-
-rm -rf warp-rnnt
-git clone https://github.com/1ytic/warp-rnnt
-
-(
-    set -euo pipefail
-    cd warp-rnnt/pytorch_binding && python3 setup.py install
-)
+pip install warp_rnnt
```
b-flo, line 18: I would prefer not relying on pypi package. If needed we can quickly fix through PR or our own fork and not wait for pypi package new version.

tools/installers/install_warp-rnnt.sh
GNroy: I can't say I thoroughly tested it, but the warp-rnnt unit test worked. Though I installed warp-rnnt with pip, if it matters.

tools/installers/install_warp-rnnt.sh
GNroy: Ok, I'll revert the original installation back

tools/installers/install_warp-rnnt.sh
GNroy: No, it is `python -m warp_rnnt.test`

b-flo: Sorry I read too quickly!
Hum, every install (from repo or pypi) with a cuda version != 10.0 ended-up with an ImportError when I tested. Could you verify? If not, I'll do it tomorrow

GNroy: Welp, I have no ImportError in cuda 10.1 environment. I'm going to try this also with a clean miniconda environment to verify though.

b-flo: Sorry, I edited my message `ImportError` -> `error`! The error I encountered was:

```
  File "/home/fboye/espnet/tools/venv/lib/python3.7/site-packages/warp_rnnt-0.4.0-py3.7-linux-x86_64.egg/warp_rnnt/__init__.py", line 2, in <module>
    import warp_rnnt._C as core
ImportError: /home/fboye/espnet/tools/venv/lib/python3.7/site-packages/warp_rnnt-0.4.0-py3.7-linux-x86_64.egg/warp_rnnt/_C.cpython-37m-x86_64-linux-gnu.so: undefined symbol: cudaSetupArgument
```

GNroy: Just checked with clean cuda 10.2 torch 1.6 env. I am using Ubuntu 18.04.
```
import warp_rnnt._C as core
...
costs, grads = core.rnnt_loss(
...             xs.cuda(), ys.cuda(),
...             xn.cuda(), yn.cuda())

>>> costs
tensor([4.2742], device='cuda:0')
>>> grads
tensor([[[[ 0., -1.,  0.,  0.,  0.],
          [ 0.,  0., -1.,  0.,  0.],
          [-1.,  0.,  0.,  0.,  0.]]]], device='cuda:0')```

b-flo: Could you also try with torch==1.4 please? If it works, let's keep that condition (with the fix I proposed or change `||` to `&&`)!
I may have made a mistake when testing, good to know if it also works for other cases!

Thanks!

GNroy: Sorry, I pushed the fix just before reading your last comment.
torch 1.4 was my initial setup, where everything worked. I just used 1.6 for a wider check.

b-flo: OK, thank you!

GNroy:
@sw005320 @ShigekiKarita I suppose this PR is ready to be merged.

sw005320 (approved):
LGTM, thanks!

PR #2506, opened 2020-09-22
Unify arguments

hirofumi0810:
This PR removes duplicate lines for the argument parser in ASR/ST/MT tasks.


sw005320:
Thanks.
This change looks good, but I'm just concerning that this way would be a bit different from TTS or RNN-T.
But since both of them are very different from ASR/ST/MT while ASR/ST/MTare quite similar. So, this change should be OK.
@kan-bayashi, do you have some comments about it?

kan-bayashi (approved):
LGTM
Since TTS model has slightly different arguments, we can keep `e2e_transformer_tts.py` as is.

PR #2519, opened 2020-09-27
Update h5py version to avoid errors in Python3.8

shigabeev:
Old h5py doesn't run on python 3.8. New one seems [fully compatible](https://docs.h5py.org/en/latest/whatsnew/2.10.html). On my computer everything runs just fine, but I'm not a tester, so I can't be sure.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -37,7 +37,7 @@ requirements = {
         # File IO related
         "PyYAML>=5.1.2",
         "soundfile>=0.10.2",
-        "h5py==2.10.0",
+        "h5py>=2.10.0",
         "kaldiio>=2.17.0",
         # TTS related
         "inflect>=1.0.0",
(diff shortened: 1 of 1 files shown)


kamo-naoyuki:
Thanks. I think this is okay. I don't know why we fixed the version and we don't also need to fix it to v2.10.0. i.e. it should be `h5py>=2.10.0`.

> Old h5py doesn't run on python 3.8

 Our test passed in python3.8, so we couldn't detect it. Could you tell me what's error and when you meet?

setup.py
```diff
@@ -37,7 +37,7 @@
         # File IO related
         "PyYAML>=5.1.2",
         "soundfile>=0.10.2",
-        "h5py==2.9.0",
+        "h5py==2.10.0",
```
kan-bayashi, line 40:
```suggestion
        "h5py>=2.10.0",
```

PR #2521, opened 2020-09-27
Split asr.py into smaller modules.

ShigekiKarita:
espnet/asr/pytorch_backend/asr.py gets too large (1432 lines). Because it is hard to review and refactor, I decided to split it into smaller pieces (converter.py, enhance.py, recog_v1.py, train.py).

In addition, I also fixed new modules to pass flake8-docstring.


sw005320:
I'll read through this PR. One high-level comment is whether we should do the same refactoring for others.
Currently, most main task scripts (e.g., st.py, mt.py, lm.py) are aligned with the same style with `asr.py`, and ideally, we should do the same thing, but it would require a lot of work.
What do you think?

We can also ask @hirofumi0810 and @kan-bayashi about it.

ShigekiKarita:
I think at least, train() and decode() should be divided because they are
just entry points from espnet/bin and share nothing each other.

2020年9月29日(火) 1:01 Shinji Watanabe <notifications@github.com>:

ShigekiKarita:
> share nothing each other.

Except for task-specific Converter, maybe. Ideally, each task package will
have entry points (train.py, decode.py, etc), and local helpers
(converter.py, etc).

2020年9月29日(火) 1:15 Karita Shigeki <shigekikarita@gmail.com>:

kan-bayashi:
If we want to unify, I can do it for `tts.py`.
(But now I focus on ESPnet2, so I have not so high motivation actually☹️)

ShigekiKarita:
I'm ok if other tasks don't follow because they are smaller and do clearly one thing (tts, mt, st). However, asr.py is different. It is larger and does many things.

Especially, mt.py and st.py share the Updater and Evaluator in asr.py. Maybe, they should be in task-agnostic modules, e.g., espnet/utils/training/{evaluator,updater}.py?

sw005320:
> I'm ok if other tasks don't follow because they are smaller and do clearly one thing (tts, mt, st). However, asr.py is different. It is larger and does many things.

I see, and I'm also OK with your idea.
We may not have to unify everything.

ShigekiKarita:
> Especially, mt.py and st.py share the Updater and Evaluator in asr.py. Maybe, they should be in task-agnostic modules, e.g., espnet/utils/training/{evaluator,updater}.py?

Done (and resolved conflicts).

PR #2523, opened 2020-09-28
Conformer End-to-End Speech Translation

hirofumi0810:


sw005320:
Many thanks!
@pengchengguo, could you review this PR?

sw005320 (approved):
LGTM
- Once you get a result, please update the result and upload a model.
- You can make these configurations as default in the fisher_callhome_spanish task (change the symlink target to conformer one).

espnet/nets/pytorch_backend/e2e_st_conformer.py
```diff
@@ -0,0 +1,72 @@
+#!/usr/bin/env python3
```
ShigekiKarita, line 1: shebang is not needed for non executables

espnet/nets/pytorch_backend/e2e_st_conformer.py
```diff
@@ -0,0 +1,72 @@
...
+        """Construct an E2E object.
+
+        :param int idim: dimension of inputs
+        :param int odim: dimension of outputs
+        :param Namespace args: argument Namespace containing options
+        """
+        super().__init__(idim, odim, args, ignore_id)
+        if args.transformer_attn_dropout_rate is None:
+            args.transformer_attn_dropout_rate = args.dropout_rate
+        self.encoder = Encoder(
```
ShigekiKarita, line 55: How about make this encoder initialization in one common function like the `add_conformer_arguments`?

hirofumi0810: I think that is a general refactoring for all encoder-decoder models, so I'd like to consider it in another PR.

hirofumi0810:
@sw005320 I change the config link to the Conformer one.
Let me upload models in another PR because the Conformer models are still under tuning and maybe I will frequently change the config until then.

sw005320:
Many thanks for the impressive results!!!

> Let me upload models in another PR because the Conformer models are still under tuning and maybe I will frequently change the config until finishing tuning.

Sure, this sounds good to me!

PR #2533, opened 2020-09-29
[ESPnet2] Transducer

b-flo:
Hi,

This PR add vanilla RNN-T training + decoding (w/ all search algorithms supported) for ESPnet2. For now, it's quite straightforward and I duplicated and modified classes from ESPnet1: `RNNTDecoder`, `BeamSearchTransducerESPnet2` and `ErrorCalculatorTransESPnet2`.
I also added run script and config files in vivos recipe for testing purpose. The former will be removed in the future.

Concerning performances, I observed some degradations. I'll have to investigate, I may have made a mistake:

| dataset | cer / wer (espnet1) | cer / wer (espnet2) |
| - | - | - |
| dev | 17.4 / 39.7 | 18.9 / 45.1 |
| test | 18.4 / 38.9 | 21.1 / 45.8 |

@kamo-naoyuki I'm not sure what should be modified and future actions, feel free to assign me next tasks! 

P.S: I'll extend espnet2 test later to cover RNN-T when we have a proper v1.


espnet2/bin/asr_inference.py
```diff
@@ -14,6 +14,9 @@
 from typeguard import check_return_type
 from typing import List
 
+from espnet.nets.beam_search_transducer_espnet2 import BeamSearchTransducerESPnet2
+from espnet.nets.beam_search_transducer_espnet2 import Hypothesis as RNNTHypothesis
+
```
kamo-naoyuki, line 19: Should we use the name of `RNNT` in RNNTHypothesis? TransducerHypothesis is more general?

espnet2/asr/espnet_model.py
```diff
@@ -46,7 +49,7 @@ def __init__(
         encoder: AbsEncoder,
         decoder: AbsDecoder,
         ctc: CTC,
-        rnnt_decoder: None,
+        rnnt_decoder: Optional[AbsDecoder],
```
kamo-naoyuki, line 52: Should we use the name of `rnnt` in rnnt_decoder? transducer_decoder is more general?

espnet2/asr/decoder/rnnt_decoder.py
```diff
@@ -0,0 +1,331 @@
...
+
+from espnet.nets.beam_search_transducer_espnet2 import Hypothesis
+from espnet.nets.pytorch_backend.nets_utils import get_activation
+from espnet.nets.pytorch_backend.nets_utils import to_device
+
+from espnet2.asr.decoder.abs_decoder import AbsDecoder
+
+
+class RNNTDecoder(AbsDecoder):
+    """RNN-T Decoder module.
```
kamo-naoyuki, line 19: Can you consider RNNTDecoder into RNN module and joint module?
Otherwise, we need many many duplications only for transducer, RNNTDecoder, TransformerTDecoder, etc. This is one of the worst point of espnet1.

kamo-naoyuki (commented):
Thank you for quick PR!  I quickly reviewed and I think almost already good.

Next priority is creating a integration test for transducer in ci/integration_test.sh

sw005320:
Great @b-flo!

> I also added run script and config files in vivos recipe for testing purpose. The former will be removed in the future.

Do you mean you will remove `run2.sh` in the future?
I think it would be a good occasion to reconsider how we allow other ASR algorithms including this transducer and non-autoregressive ASR etc.
As you know, we avoided using a lot of `run*.sh` to easily maintain only the main `run.sh` per recipe and switching the algorithm by using the config files or regarding the other ASR algorithm as another recipe, e.g., `asr1_multi`.

TTS is in a similar situation and it deals with multiple algorithms by changing the config files. So, the transducer related methods can also be done similarly?

espnet/nets/beam_search_transducer_espnet2.py
```diff
@@ -0,0 +1,653 @@
...
+from dataclasses import dataclass
+from typeguard import check_argument_types
+from typing import Any
+from typing import Dict
+from typing import List
+from typing import Union
+
+from espnet.nets.lm_interface import LMInterface
+from espnet2.asr.decoder.abs_decoder import AbsDecoder
+
```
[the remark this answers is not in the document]
b-flo: We could with some unification, input/output args were not the same that's why I added a new one for espnet2. I took the opportunity to modify into a class.

espnet2/asr/decoder/rnnt_decoder.py
b-flo: Hum, I'm not sure I understand. You mean defining a general TransducerDecoder based on default `RNNDecoder` + `JointNetwork`? Sounds good to me but I'm not sure how to handle properly the different methods for RNN-T/RNN-T-Att

b-flo:
> Do you mean you will remove run2.sh in the future?

Yes, I just put it here temporary and it's also a note for me: `inference_asr_model=valid.loss.best.pth` needs to be set for transducer, maybe I should enforce automatically this parameter in that case.

> Should we use the name of rnnt in rnnt_decoder? transducer_decoder is more general?

@sw005320 @kamo-naoyuki I'm a bit conflicted on the naming. While RNN-T is not adapted given all different architectures we can use, I'm not sure I like "Transducer" or "Transducer-based ASR/model". It conflicts a bit with traditional hybrid systems relying on wFST in my opinion and can be confusing for some new users.
Do you have an opinion on that?

espnet2/asr/decoder/rnnt_decoder.py
kamo-naoyuki: Yes, exactly.

espnet/nets/e2e_asr_common.py
```diff
@@ -396,3 +397,159 @@ def calculate_wer(self, seqs_hat, seqs_true):
             word_ref_lens.append(len(ref_words))
 
         return float(sum(word_eds)) / sum(word_ref_lens)
+
+
+class ErrorCalculatorTransESPnet2(object):
+    """Calculate CER and WER for transducer models.
```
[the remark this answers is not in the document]
b-flo: I put it here because all ErrorCalculatorX were there but I think I'll unify `ErrorCalculatorTrans` and `ErrorCalculatorTransESPnet2`

espnet2/asr/decoder/rnnt_decoder.py
b-flo: OK, I'll do that! About the different methods between RNN-T/RNN-T-Att (`forward`, `score`, etc) what would be the correct approach in your opinion?

kamo-naoyuki: Thank you.

Sorry,  but I didn't deeply think how to do. Probably, we can and need to create multiple JointNetworks for RNN-T/RNN-T-Att respectively.
Maybe, you can produce better idea.

sw005320:
> > Should we use the name of rnnt in rnnt_decoder? transducer_decoder is more general?
> 
> @sw005320 @kamo-naoyuki I'm a bit conflicted on the naming. While RNN-T is not adapted given all different architectures we can use, I'm not sure I like "Transducer" or "Transducer-based ASR/model". It conflicts a bit with traditional hybrid systems relying on wFST in my opinion and can be confusing for some new users.
> Do you have an opinion on that?

I'm actually OK to call it "transducer_decoder" because traditional hybrid systems usually put "finite state" or whatever as a prefix. So, if we'll have fst decoder in the near future (by k2?), we can just call it "fst_decoder".

kamo-naoyuki:
>  It conflicts a bit with traditional hybrid systems relying on wFST in my opinion and can be confusing for some new users.

I'm not also sure. Actually, RNN-Transducder is widely used name, so perhaps RNNT is better to represent this architecture, as you said.  Both are good to me.

espnet2/asr/decoder/rnnt_decoder.py
b-flo: I made a new version, it could be improved but I guess it's a start.

espnet2/bin/asr_inference.py
b-flo: I've gone for Transducer for now!

espnet2/asr/espnet_model.py
b-flo: ditto

espnet/nets/beam_search_transducer_espnet2.py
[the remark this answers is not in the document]
b-flo: Done!
@ShigekiKarita I had to modify `espnet/asr/pytorch_backend/asr.py`, is that okay? I'm asking because of #2521

espnet/nets/e2e_asr_common.py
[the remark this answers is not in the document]
b-flo: I unified `ErrorCalculatorTrans` and `ErrorCalculatorTransESPnet2`, should I put the class in another file?

espnet/nets/beam_search_transducer_espnet2.py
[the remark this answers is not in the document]
kamo-naoyuki: Please keep as it is for espent1. It is not related to this PR.

espnet/nets/e2e_asr_common.py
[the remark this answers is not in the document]
kamo-naoyuki: No

kamo-naoyuki:
Please care about import order

```
(python official)

(third pirty)

(project modules)
```

```
# e.g.
import typing

import torch
import typeguard

import espnet
import espnet2
```

espnet2/asr/decoder/rnnt_decoder.py
kamo-naoyuki: Sorry, my explanation might be bad. Your modification is against my intention. 

I don't think much yesterday. I'll say about ideal at the next. I'm not sure whether it's not possible.

Again. We must not create duplications as much as possible. This is more important than you may expected for the future development and maintainability. 
I requested you doing it with not a little effort to refactor the current transducer structure drastically.

1. We don't need a general `TransducerDecoder`.
2. We must **share the same decoders** between att/ctc hybrid and transducer. 

About training:

```python
# ESPNetASRModel.forward()
decoder_out = decoder()
joint_network(enoder_out, decoder_out)
```

About scoring:

Currently, the scoring methods are quite different between att/ctc hybrid and transducer. I'm not sure whether we can unify them. If it's impossible, then RNNDecoder should have both methods.

```python
# espnet2.asr.decoders.rnn_decoder
class RNNDecoder:
    def score(self, ...):
    def score_transducer(self, ...):

class TransducerScorerAdapter:
     def __init__(self, decoder):
           self.decoder
     def score(self, ...)
           self.decodoer.score_transducer(self, ...):

```

What do you think?

espnet/nets/beam_search_transducer_espnet2.py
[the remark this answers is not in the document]
b-flo: Sorry, I thought it was your intention given first question...

If it's okay with you I'll keep it as if until PR is almost done and revert for espnet1 before merging. It was a bit complicated to unify `BeamSearchTransducer` and `ErrorCalculator` because I had to do it altogether alongside taking into account espnet1 and espnet2. As I refactored a lot of files at the same time, I couldn't properly group files / actions in commits so it will take me sometimes to revert without reverting almost everything. Sorry about that.

b-flo:
> Please care about import order
>
> > (python official)
> >
> > (third pirty)
> >
> > (project modules)

Sorry, I normally follow PEP8 style but I may have been lazy in some late changes! I will change that!

espnet2/asr/decoder/rnnt_decoder.py
b-flo: > Again. We must not create duplications as much as possible. This is more important than you may expected for the future development and maintainability.

Oh I totally understand that for development and maintanaibility.

So, If we are clear here: you want me to extend `RNNDecoder` instead of inheriting `RNNDecoder` in `TransducerDecoder` and use `JointNetwork` only in `_calc_transducer_loss` for training? I'm also fine with that but I need to modify `RNNDecoder` to handle some cases (no attention mainly).
Outside `JointNetwork` I'll move, I'm not sure it would be better for maintainability? With some small modifications in current implementation (calling parent methods in `init_state` and `forward` when attention is used), most code would just be moved as there are not many duplications. It would also be harder in terms of readability I think?
 
> About scoring:
> 
> Currently, the scoring methods are quite different between att/ctc hybrid and transducer. I'm not sure whether we can unify them. If it's impossible, then RNNDecoder should have both methods.

I'm not sure either, I couldn't do it in my first try.
Btw, I'm planning to replace both transducer training and decoding part by a torch lib for multiple purposes (optimizations, new features, speed, etc) so unification will take another form I guess...

espnet/nets/beam_search_transducer_espnet2.py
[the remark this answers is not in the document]
b-flo: Nevermind, I'll do another PR for beam search and error calculator in espnet1! The changes in this PR will be omitted after merging and updating master

b-flo:
Oups, sorry about last commit I mixed up branches. I'll wait for #2538 to be merged and I'll merge master into this PR to remove files related to espnet1.

b-flo:
@kamo-naoyuki I made the new version according to what we discussed previously:

- For now I only defined a transducer decoder as a `RNNDecoder` + `JointNetwork`. I went for the most minimal/straighforward implementation to easily spot difference. After we are set on a "template", I'll extend to other decoders (`RNNdecoder` -> `decoder_class`)
- I took some liberties for the architecture definition in config, not sure what you had in mind for that part?
- I added an additional parameter `use_output` to `RNNDecoder` to control output of `forward()` and `init_state()`
- Multi-task loss was extended with `loss_transducer` but in case of `transducer_weight=0.0`, main loss is defined as `loss = transducer_loss`
- I had to make a new `BeamSearchTransducer` and `ErrorCalculatorTransducer` for espnet2 but didn't know how to put files, I went for `espnet2/asr/transducer/`, is that okay?

P.S: I'll extend test later this day

kamo-naoyuki:
> * I had to make a new `BeamSearchTransducer` and `ErrorCalculatorTransducer` for espnet2 but didn't know how to put files, I went for `espnet2/asr/transducer/`, is that okay?

Yes.

espnet2/asr/espnet_model.py
```diff
@@ -46,8 +49,9 @@ def __init__(
         encoder: AbsEncoder,
         decoder: AbsDecoder,
         ctc: CTC,
-        rnnt_decoder: None,
+        transducer_decoder: Optional[Tuple[AbsDecoder, torch.nn.Module]],
```
kamo-naoyuki, line 52: Unreadable. Split it into two arguments: `transducer_decoder`, `joint_net`

espnet2/asr/transducer/beam_search_transducer.py
```diff
@@ -0,0 +1,679 @@
...
+
+        """
+        init_tensor = h.unsqueeze(0)
+        dec_state = self.decoder.init_state(init_tensor)
+
+        hyp = Hypothesis(score=0.0, yseq=[self.blank], dec_state=dec_state)
+
+        cache = {}
+
+        y, state, _ = self.decoder.score_transducer(hyp, cache, init_tensor)
```
kamo-naoyuki, line 149: `score_transducer` is confusing name because it doesn't score, but return the network output.

espnet2/asr/transducer/beam_search_transducer.py
```diff
@@ -0,0 +1,679 @@
...
+
+                top_k = ytu[1:].topk(beam_k, dim=-1)
+
+                ytu = (
+                    torch.cat((top_k[0], ytu[0:1])),
+                    torch.cat((top_k[1] + 1, blank_tensor)),
+                )
+
+                if self.lm:
+                    lm_state, lm_scores = self.lm.predict(max_hyp.lm_state, lm_tokens)
```
kamo-naoyuki, line 209: `predict` doesn't exist.

espnet2/bin/asr_inference.py
```diff
@@ -107,6 +111,19 @@ def __init__(
             token_list=token_list,
             pre_beam_score_key=None if ctc_weight == 1.0 else "full",
         )
+
```
kamo-naoyuki, line 114: - We need some flag to enable/disable transducer for multi task training.
- If constructing BeamSearchTransducer, stop to construct BeamSearch.

espnet2/asr/espnet_model.py
b-flo: OK!

espnet2/asr/transducer/beam_search_transducer.py
b-flo: Sorry, I didn't update LM part yet!

kamo-naoyuki: Maybe, can we use lm as it is? LM.score can be used?

espnet2/asr/transducer/beam_search_transducer.py
b-flo: Yes, I used `score` and `batch_score` as request in espnet1 to try matching Scorer, I followed the same style here.
I don't have a good idea for the naming here, maybe`one_step` and `one_batch_step`? Dunno...

espnet2/asr/transducer/beam_search_transducer.py
b-flo: I'm not familiar with the LM part in espnet2 sorry, I'll take a look and update!

espnet2/bin/asr_inference.py
b-flo: I didn't know what you wanted for decoding so I let it like that.
For now I think we should use `asr_model.rnnt_decoder` to control and construct either one of them based on existence. Is that ok?

espnet2/asr/espnet_model.py
b-flo: Out of curiosity, why do you think it's "unreadable"? I'm not too experienced with type hints yet so I'm trying to understand!

espnet2/asr/transducer/beam_search_transducer.py
kamo-naoyuki: `one_step` is not bad. Or just `step`?

b-flo: OK for `step_transducer` and `batch_step_transducer`!

espnet2/asr/espnet_model.py
kamo-naoyuki: This argument represents two independent objects. There are no reason to pack it into one and it's unexpected, normally.

espnet2/asr/transducer/beam_search_transducer.py
b-flo: We can use `score` and `batch_score` indeed, I just need to add a `zero_state` method to LM class.
Btw, I'm wondering two things about:

- The LM implementation in espnet2 is more suited for transducer as I don't need to create batch states manually nor extra manipulation to extract specific LM state. I'm wondering how efficient is it compared to implementation in espnet1?
- Is there a reason we didn't rely on list of batches (L, hid_units) in `RNNDecoder` instead of list of layers (B, hid_units)? I know it's direction against another but it would be easier/cleaner in transducer case

b-flo:
Hi,

I extended transducer from `RNNDecoder` to all decoders (`decoder_class`) except I disabled transformer-based transducer with `use_attention=True` (equivalent to original transformer with self and src attention). @kamo-naoyuki On that note, I'm not sure `use_attention` is a correct parameter name in case of transformer but at the same time I'm not sure how to name it.

What's remain is some clean-up/refactor and in-depth testing. While I checked I didn't break other implementations and added test, we should check I didn't harm performance. Maybe someone can help me on that part?

For the remaining items associated to transducer in espnet1 (tdnn, causal conv1d, some options, etc), I'll add them to espnet2 in another PR.

espnet2/asr/transducer/beam_search_transducer.py
kamo-naoyuki: > * Is there a reason we didn't rely on list of batches x (L, hid_units) in RNNDecoder instead of list of layers x (B, hid_units)? I know it's a direction against another but I think it would be easier/cleaner in transducer case

What part are you talking about?

b-flo: Sorry it was confusing: I was talking about how we deal with LM state during inference in espnet2 vs espnet1 and expected input/output (i.e.: predict in espnet1 vs score in espnet2) . I was wondering if I could follow the same direction for decoder states and also which one was the most efficient in terms of speed in my case.

But don't mind the second question, I'll do some testing and benchmarking!

stefan-falk:
Hi @b-flo :)

> dataset	cer / wer (espnet1)	cer / wer (espnet2)
> dev	17.4 / 39.7	18.9 / 45.1
> test	18.4 / 38.9	21.1 / 45.8

Just got a question to those results: What dataset was used here and what decoding method. Also, what are the hyper-parameters of the model that has been used there? 

My own model comes to around 18% CER on a German dataset but I struggle to reduce that.

I don't manage to improve those results nots even with a simple beam-search..

b-flo:
Hi,

> Just got a question to those results: What dataset was used here and what decoding method. Also, what are the hyper-parameters of the model that has been used there?

The dataset is VIVOS which is a small dataset (15h) I'm usually using to demonstrate/validate new features. I don't remember the decoding method but it would'nt change results too much (+/- 0.2 CER relative FWIR). For the training and decoding parameters, you can refer to configs and run scripts available in `egs/vivos/asr1/`, same parameters were used in espnet2.

stefan-falk:
@b-flo Thank you for the information! :)

> I don't remember the decoding method but it would'nt change results too much (+/- 0.2 CER relative FWIR)

Why would it change only that little? Shouldn't beam-search improve the outcome by some margin?

b-flo:
> Why would it change only that little? Shouldn't beam-search improve the outcome by some margin?

You mean compared to greedy search? Yes, it should improve performance by some margin. I was only talking about default beam search vs variant beam search!

stefan-falk:
@b-flo Ah, okay that makes sense. My problem at the moment is that beam-search does not really improve something (my own code, not espnet). The `default_beam_search` implementation is similar to the one from this PR but a beam-width or 5, 10, 20 etc is _worse_ than greedy decoding. If I increase the beam-size, the performance degrades so much that it does not make sense to evaluate that any further. I'm just not sure what that is so.. should I expect improvements from beam-search even with a small beam-width?

b-flo:
>  If I increase the beam-size, the performance degrades so much that it does not make sense to evaluate that any further.

A huge beam size can obviously degrades performance but it shouldn't be by a huge margin (with dataset sizes in mind of course).

> should I expect improvements from beam-search even with a small beam-width?

There are some cases where a very small beam size doesn't yield better performances but in general, yes it should improve performances.

stefan-falk:
@b-flo Any chance you can link me to a table which I can use for comparing my own results? I am not sure what I might be doing wrong here but I am also not so sure what I can expect from those approaches :/

b-flo:
> @b-flo Any chance you can link me to a table which I can use for comparing my own results?

Well, you can use [Voxforge ](https://github.com/espnet/espnet/blob/master/egs/voxforge/asr1/RESULTS.md) or [VIVOS](https://github.com/espnet/espnet/blob/master/egs/vivos/README.txt) as comparison I guess. There are both free also.
I'll update the later but performance differences between search algorithms stand.

stefan-falk:
@b-flo Thanks, that's useful! How do I read this table? https://github.com/espnet/espnet/blob/master/egs/voxforge/asr1/RESULTS.md#wer-1 Is **Err** the actual WER?

espnet2/asr/espnet_model.py
```diff
@@ -79,20 +85,39 @@ def __init__(
...
+                    self.joint_network,
+                    token_list,
+                    sym_space,
+                    sym_blank,
+                    report_cer,
+                    report_wer,
+                )
+
+            if not transducer_decoder or (
+                transducer_decoder and transducer_weight != 0.0
```
Emrys365, line 116: I guess the second condition should be `transducer_decoder and transducer_weight == 0.0`.

espnet2/asr/espnet_model.py
```diff
@@ -125,41 +150,64 @@ def forward(
...
-            loss_att, acc_att, cer_att, wer_att = None, None, None, None
-        else:
-            loss_att, acc_att, cer_att, wer_att = self._calc_att_loss(
+        loss_att, acc_att, cer_att, wer_att = None, None, None, None
+        loss_ctc, cer_ctc = None, None
+        loss_transducer, cer_transducer, wer_transducer = None, None, None
+
+        if not self.transducer_decoder or (
+            self.transducer_decoder and self.transducer_weight != 0.0
+        ):
```
Emrys365, line 159: ditto

espnet2/asr/espnet_model.py
b-flo: > I guess the second condition should be transducer_decoder and transducer_weight == 0.0.

I think the second condition is correct?

CER/WER computation for other models should only be enabled if 1) no transducer decoder is selected and 2) if a transducer decoder is selected and multi-task learning with transducer is enabled (`transducer_weight != 0.0`)

The difference between ESPnet1 and ESPnet2 is that transducer loss can be used in multi-task learning (current form is temporary though.)

Emrys365: Oh I see. I misunderstood the multi-task part.

sw005320:
@b-flo, could you explain to me the difference between https://github.com/espnet/espnet/blob/master/espnet/nets/beam_search_transducer.py and https://github.com/espnet/espnet/blob/f800a51a69e73bcda298e0299ec28418b4f1a808/espnet2/asr/transducer/beam_search_transducer.py? It seems that they are duplicated.

b-flo:
@sw005320 There are mostly the same, FWIR the differences are due to 1) the way LM is handled in ESPnet1 compared ESPnet2 and 2) how we define transducer decoder in ESPnet1 compared to ESPnet2.
As I was still updating ESPnet1 version, it was a bit difficult to consider changes in both versions so we decided to separate beam search for now.

It's not a priority I think but I'll try to unify the class when we're set on 2)

sw005320:
Thanks for the clarification.
I think it is fine for now. We also don't want to make one PR extremely complicated, and I agree with your plan for the beam search part.

stefan-falk:
@b-flo Hi! I'm back from my "vacation" and am going to continue working on my RNN-T implementation. My "last" remaining issue streaming-inference. I am focusing mainly on [[1]] but find it a bit hard to fully understand how they actually implement inference after having trained the model with the proposed masking method.

Anyway, referring to our chat from a few weeks ago, what is ESPnet's status in this regard and how can I follow your progress (or maybe even be helpful in some way)?

[1]: https://arxiv.org/pdf/2010.11395.pdf

b-flo:
Hi,

Well, I was away from ESPnet for a few weeks so I can't really say there was progress on my side... I came back today ;)
But in short:
 - Port to ESPnet2 need to be validated to initiate most of the work we discussed internally. I guess we will put a roadmap starting from that point.
 - @hirofumi0810 already started working on streaming implementation for its [neural_sp project](https://github.com/hirofumi0810/neural_sp). We may take inspiration from that depending on first item and other models (porting)
 - In the meantime I'm testing some minor things for general training and inference improvement (lstmcell vs lstm, warp-rnnt vs warp-transducer, auxiliary loss and additional criterion, etc).

stefan-falk:
@b-flo Hi :) That's the same for me. First day back at work and, typical for the first day, still a bit groggy. 😆 

I'd be interested in seeing/following your roadmap as soon as you have one :)

At the moment I am trying to figure out how decoding is supposed to work in the first place. This probably depends pretty much on the method that was used to train the model but in [Developing Real-Time Streaming Transformer Transducer for Speech Recognition on Large-Scale Dataset](https://arxiv.org/pdf/2010.11395.pdf) they are mentioning [Low Latency End-to-End Streaming Speech Recognition with a Scout Network](https://arxiv.org/pdf/2003.10369.pdf) in "_.. the streamable Transformer Seq2Seq is based on the chunk-wise trigger attention method, .._" as their method for inference-decoding. This method is quite more complex than what I thought it would be.

The way the mask establishes chunking, I assumed it would be sufficient to simply "slice" the encoder output and forward it to the label-decoder accordingly.

Do we know what approach @hirofumi0810 is trying out? I personally would like to succeed in implementing _any_ method at all in order to verify that the models I am producing right now can in fact be used for streaming. No matter how efficient as optimization is something for later I guess..

b-flo:
> Do we know what approach @hirofumi0810 is trying out? I personally would like to succeed in implementing any method at all in order to verify that the models I am producing right now can in fact be used for streaming. No matter how efficient as optimization is something for later I guess..

Sorry I didn't dig into the code yet, you can take a look yourself with the link I provided or wait for him to comment here when he has time! I know one of its main reference for attention-based models is [MoChA](https://arxiv.org/abs/1712.05382) though

stefan-falk:
I'll definitely take a look, thank you. :)

stefan-falk:
It turns out I misunderstood what MoChA is for. It appears this is for encoder-decoder based models but won't work in a Transducer - as far as I can tell.

It doesn't appear to be the way to go here but I had an interesting round-trip 😆 

Anyway, I'm back to where I ask myself how to make the T-T model streamable, or, how inference works for the [Streaming Transformer-Transducer](https://arxiv.org/pdf/2010.11395.pdf).

kingsword5566:
Hi
When will this feature be released? 
Which files that I would need to add if I want to try this for testing?

b-flo:
@kingsword5566 Sorry, it was put on stand-by because of internal discussions. Technically it would work after resolving conflicts if there were no major changes in ESPnet2.

I'll revisit the PR after Interspeech deadline, there were many modifications and features added for transducer in ESPnet1 so I will probably change a lot of things here.

csukuangfj:
> I'll revisit the PR after Interspeech deadline

Is there any progress with this PR?

b-flo:
@csukuangfj Sorry for the delay!

> Is there any progress with this PR?

No but I'll revisit this PR in the following weeks to propose a simple RNN-Transducer implementation. It will have minimal features compared to ESPnet1 version though.

Transducer port to ESPnet2 is a bit tricky to be honest. At least, if we want to 1) keep all ESPnet1 features related to Transducer, 2) add some new specific features and 3) modify/optimize some parts in regards to perf and decoding speed. Also, in current form, some implementations/modifications I had in mind may negatively impact other ASR models or parts.

That being said, we have a work in progress related to Transducer, ESPnet2, streaming and production purposes. I'm not sure on the final form (it may be a "standalone" ESPnet-Transducer) but I'm aiming mid january for the first working version.

b-flo:
@sw005320 I resolved conflicts but don't merge it until [WIP] is in the title! The implementation relies on ESPnet1 objects/classes but there was too many changes since last year.

sw005320 (commented):
Super thanks!
@ftshijt, can you also take a look at this?

1. `run2.sh` -> `run.sh`?
2. Can you add a result (README.md) and upload a model to the hugging face hub?

b-flo:
>  Can you add a result (README.md) and upload a model to the hugging face hub?

Sure, I'll train a model when the implementation is working. Which dataset should I use?
@ftshijt please wait a bit before reviewing, I'll notify you when it's ready.

>   run2.sh -> run.sh?

Ha yes, forgot about that. I'll modify it. For information, from previous message : 

>    Do you mean you will remove run2.sh in the future?
>
> > Yes, I just put it here temporary and it's also a note for me: inference_asr_model=valid.loss.best.pth needs to be set for transducer, maybe I should enforce automatically this parameter in that case.

espnet2/asr/espnet_model.py
```diff
@@ -93,20 +99,39 @@ def __init__(
             self.ctc = None
         else:
             self.ctc = ctc
-        self.rnnt_decoder = rnnt_decoder
+
+        self.transducer_decoder = transducer_decoder
+        self.joint_network = joint_network
+
+        self.criterion_transducer = TransLoss("warp-transducer", self.blank)
```
ftshijt, line 106: Hi, I think we currently do not have a default installation of warp transducer loss, probably we can add a condition here of whether to initial transducer loss here?

ftshijt (commented):
Excellent work! I like the idea of separating the joint network from rnnt decoder. I only have a concern over the warp-rnnt loss. In case that is from another repo with the potential of losing compatibility with later cuda/pytorch, I would suggest separating it out during the initialization, so it would cause problems to other modules' training even if the rnn-t is down.

b-flo:
@ftshijt The code is quite old and some things don't exist or make sense anymore.
Also, I'm currently updating the code and the design will be changed (decoder and joint network are still separated).

About loss module, we're relying on my fork of warp-transducer and I'll keep it updated (and add small features) until we make a proper one! I put a note in the [documentation](https://espnet.github.io/espnet/tutorial.html#transducer).

b-flo:
Hi,

@ftshijt OK, I quickly made a new version for ESPnet2, it seems to work properly.

Test with VIVOS : 

| dataset | CER / WER (espnet1) | CER / WER (old espnet2) | CER / WER (new espnet2) |
| - | - | - | - |
| dev | 17.4 / 40.7 | 18.9 / 45.1 | 17.0 / 42.4 |
| test | 16.8 / 38.3 | 21.1 / 45.8 | 19.0 / 44.0 |

TO DO : 

- [ ] Clean-up and some refactoring (there are some duplicates and unused/useless mechanisms).
- [ ] Check all Transducer-related features are correctly working.
- [ ] Perf analysis + optimization to match or get close to ESPnet1 results.
- [ ] Add tests

sw005320:
any update?

b-flo:
@sw005320 I was waiting for some feedback on the new design before the next items as I would need to revert commits if it's not OK. I can finish the TO DO anyways if you want.

csukuangfj:
Do you have results for the LibriSpeech dataset?

b-flo:
@csukuangfj Not with ESPnet2. With ESPnet1, we do, but some optimizations and tuning are needed (results are without LM): 

| Model | dev-clean | dev-other | test-clean | test-other |
| - | - | - | - | - |
| RNN-T | 4.0 | 11.2 | 4.2 | 11.4 |
| RNN-T + Aux | 3.8 | 10.8 | 4.0 | 10.9 |
| Conformer-T | 3.0 | 8.1 | 3.2 | 8.0 |
| Conformer-T + Aux | 2.7 | 6.7 | 2.9 | 6.8 |

csukuangfj:
@b-flo 
Thanks!  Could you also post the decoding method for the above WER?

b-flo:
Not sure on the beam size, I can't find the decoding config:

```
batch: 0
beam-size: 10 # or 5
search-type: default
```

We don't have a complete report for Librispeech but based on my previous experiments : performances would be similar with other algorithms and small parameters (one label + blank for time-synchronous approaches and `T_max={30-40}` for ALSD)

Anyway, for best WER/RTF the go-to is mAES, and ALSD in some offline cases.

espnet2/asr/espnet_model.py
```diff
@@ -81,32 +84,60 @@ def __init__(
...
-        # thanks Jeff Farris for pointing out the issue.
-        if ctc_weight == 1.0:
-            self.decoder = None
-        else:
+
+        self.use_transducer_decoder = joint_network is not None
+
+        self.error_calculator = None
+
+        if self.use_transducer_decoder:
```
ftshijt, line 92: I thought we would still have the option to apply transducer decoder and attention decoder at the same time, maybe we can just allow them to stand together instead of only allowing one at a time?

ftshijt (commented):
Looks cool from my side. Could you please explain a bit on why you separate the attention decoder and transducer decoder? I thought they might co-exist.

espnet2/asr/espnet_model.py
b-flo: I'm not sure I understand why we would want both decoder at the same time outside multi-decoding or some rescoring? Do you have an idea?
Previous implementation didn't consider this case, the unification was done to allow RNN-T decoder w/ attention. However, the performance gain is minimal and it would severely impact decoding speed.

ftshijt: Okay, I was at the beginning thought you would want to conduct some experiments with the setting of multi-decoder given the unification. Seems that I just misunderstood that. Then, everything looks cool! thanks

sw005320:
The current design is fine.
(It does not have so much interface change)
Can you finish this PR and work on your TODO in the other PR?

b-flo:
> Can you finish this PR and work on your TODO in the other PR?

No problem, I was talking about this [PR's TODO](https://github.com/espnet/espnet/pull/2533#issuecomment-996829732). I can address "Perf analysis + optimization to match or get close to ESPnet1 results." later though.

b-flo:
I'm confused by the test shell error. It constantly fails at the same point but I'm enable to reproduce the error locally.

sw005320:
It would mostly come from the download issue.
We need to rerun the CI...

b-flo:
@sw005320 I could trace the bug in the end, the error was on my side, sorry.

sw005320:
Thanks!
After the CI is passed, is it ready for merging?

b-flo:
It is but I'm bit worried by the performance degradation. I may have missed something.

sw005320:
Thanks, @b-flo!
Let's continue working on the performance part (here or other PRs).

sw005320:
> 
> Test with VIVOS :
> 
> dataset	CER / WER (espnet1)	CER / WER (old espnet2)	CER / WER (new espnet2)
> dev	17.4 / 40.7	18.9 / 45.1	17.2 / 41.2
> test	16.8 / 38.3	21.1 / 45.8	19.0 / 42.9
> _Edit: ~I made some mistake in the training config, currently re-running exp.~ Something weird is happening without projection layers, let me check._
> 
> TO DO :
> 
> * [x]  Clean-up and some refactoring (there are some duplicates and unused/useless mechanisms).
> * [x]  Check all Transducer-related features are correctly working.
> * [ ]  Perf analysis + optimization to match or get close to ESPnet1 results.
> * [x]  Add tests

@b-flo, could you share your config (or make a PR)?
We can also tune the result on our side.

b-flo:
Let's start with Voxforge instead, performance gap was more significant in my testing.

Training config (equivalent to ESPnet1):

````
# general
optim: adadelta
optim_conf:
    lr: 1.0
    eps: 1.0e-08
    rho: 0.95
init: chainer
batch_type: folded
batch_size: 24
max_epoch: 20
patience: 3

# criterion
val_scheduler_criterion:
    - valid
    - loss
early_stopping_criterion:
    - valid
    - loss
    - min
best_model_criterion:
-   - valid
    - loss
    - min
keep_nbest_models: 1

# MTL
model_conf:
    ctc_weight: 0.0
    report_cer: False
    report_wer: False

# network
encoder: vgg_rnn
encoder_conf:
    rnn_type: lstm
    bidirectional: True
    use_projection: False
    num_layers: 4
    hidden_size: 320
    output_size: 320
    dropout: 0.4
decoder: transducer
decoder_conf:
    rnn_type: lstm
    num_layers: 1
    hidden_size: 256
    dropout: 0.2
    dropout_embed: 0.2
joint_net_conf:
    joint_space_size: 256
````

Decoding config:

```
beam_size: 5
transducer_conf:
    search_type: default
    score_norm: True
```

Also, don't forget to path to set `inference_asr_model=valid.loss.best.pth` in run script.

PR #2536, opened 2020-09-30
Amharic ASR Demo Script and .flac Files

brianyan918:
This PR is for a demo of the Amharic E2E ASR demo trained on the Babel dataset.

To run the demo, I have added several files to `/egs2/babel/asr1/`
- `demo.sh` is the script to run inference on the demo audio files
- `demo/*` contains 5 .flac files, totaling ~500kB, and the corresponding .scp and ground truths
- `data/nlsym.txt` is required for the inference (but could be included in a different directory, eg demo or exp)

To host the ASR and LM .pth files, I have zipped the relevant portions of `exp` and hosted them on Zenodo

The Google Colab can be found here: [Link](https://colab.research.google.com/drive/1i0Cjy3bFRe-vUcqs7RcXMg62lT3SBR9n?usp=sharing)


sw005320:
Hi @brianyan918, thanks a lot!
It seems that these files are specific to the demonstration, and they should not have to be maintained by the espnet master.
Sorry for the confusion, but I think it would be better to be provided from your fork.

PR #2538, opened 2020-09-30
Refactor BeamSearchTransducer and ErrorCalculatorTrans

b-flo:
Extracted from #2533. In order to use the same `BeamSearchTransducer` and `ErrorCalculatorTrans` classes in espnet1 and espnet2 some refactoring needs to be done on espnet1 side. I also split joint network according to other PR.

Some notes though:

 - @ShigekiKarita `asr.py` is modified, I hope it's not an issue for #2521 
 - Uploaded models for VIVOS will be incompatible. To be honest I'm not motivated to run again all experiments... At least for now.
 - ~The PR is complete but I need to at least run some test to check I didn't break or impact performances with that changes.~ Done, there was a bug in NSC.

### DIFF
--- a/espnet/nets/beam_search_transducer.py
+++ b/espnet/nets/beam_search_transducer.py
@@ -102,7 +102,6 @@ class BeamSearchTransducer:
             nbest_hyps: N-best decoding results
 
         """
-        # if any(hasattr(self.decoder, attr) for attr in ["att", "att_list"]):
         if hasattr(self.decoder, "att_list"):
             self.decoder.att_list[0].reset()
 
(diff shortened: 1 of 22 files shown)


espnet/nets/beam_search_transducer.py
```diff
@@ -33,601 +31,634 @@ class Hypothesis:
...
+                    [hyp for hyp in kept_hyps if hyp.score > hyps_max],
+                    key=lambda x: x.score,
+                )
+                if len(kept_most_prob) >= beam:
+                    kept_hyps = kept_most_prob
+                    break
 
-    max_sym_exp = recog_args.max_sym_exp
-    nbest = recog_args.nbest
+        if self.score_norm:
```
ShigekiKarita, line 223: I wonder if all the algorithms support this option. How about moving this if/else block into a new method like `def sort_nbest(self, hyps)` and reuse it in other algorithms.

b-flo: Default beam search and NSC use score normalization but it's an option for the first one while it's enforced in the second!

b-flo:
It can be merged if it's okay for everyone! I guess that parts will be refactored one more time in the future based on modifications related to espnet2 port.

kamo-naoyuki (approved):
LGTM for my side

ShigekiKarita (approved):

PR #2545, opened 2020-10-02
Use all_gather_list to all_gather stats

kamo-naoyuki:
### Issue
If stats has different keys for each rank, the distributed training is failed.

```python
class ESPNetModel(AbsESPNetModel):
    def forward(self, ...):
        if rank == 0:
             stats = {"a": 4, "b": 5}
        else:
             stats = {"a": 4}
```

### This PR

Use `all_gather_list` instead all_reduce for each values. I copied all_gather_list from https://github.com/pytorch/fairseq/blob/0557ed8b0df90fe671bcb745f384ef7fd0386ab3/fairseq/distributed_utils.py#L279-L336

This function can gather any type of python object, but it has the size limitation for the object.

### DIFF
--- a/espnet2/train/distributed_utils.py
+++ b/espnet2/train/distributed_utils.py
@@ -373,56 +373,70 @@ def get_num_nodes(prior=None, launcher: str = None) -> Optional[int]:
         return int(os.environ.get("WORLD_SIZE", 1))
 
 
-def generate_buffer_for_all_gather_list(buffer_size: int = 16384):
-    world_size = torch.distributed.get_world_size()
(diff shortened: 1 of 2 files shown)


kamo-naoyuki:
I found that if backward paths are different for each ranks, the training will be hangups when loss.backward().

```python
class ESPNetModel(AbsESPNetModel):
    def forward(self, ...):
        if rank == 0:
             loss = loss1 + loss2
        else:
             loss = loss1
```

I concluded that this problem is inevitable and I don't think pytorch developers should fix it. This hangup may looks mysterious for many users, so I think it's worth to take care of it.

PR #2546, opened 2020-10-02
update the JSUT recipe with conformer

sw005320:
- update a result with conformer
- obtained 14.5% from 18.7% with conformer and specaug (speed perturbation was not effective, but I just included it)


sw005320:
@pengchengguo, please review this PR.

pengchengguo:
Hi @sw005320,
I am not familiar with this corpus. But I find Transformer's results include both 16k and 48k results, while the Conformer doesn't indicate it is 16k or 48k. Besides, the gain may come from RNNLM rescoring, model average, and Conformer.

sw005320:
Thanks! I'll add that description.

PR #2547, opened 2020-10-03
Install warpctc-pytorch wheel when torch version is 1.1 - 1.6

ysk24ok:
warpctc-pytorch 0.2.1 is released and the wheel can be installed when PyTorch >= 1.1 and <= 1.6.


sw005320:
Many thanks!

PR #2549, opened 2020-10-03
fix bug about att_ws in multi-enc case

lzm0706:
In multi-enc cases, att_w_list changes in each cycle, append method will make all elements in att_ws change synchronously, which leads to an incorrect attention-weights image.  So, we need to append a copy of att_w_list.


sw005320:
Thanks for using a multi encoder function!!
@ruizhilijhu, could you confirm it?

ruizhilijhu:
@sw005320 This is a bug in multi-encoder function. I also found this bug but has not pushed to espnet/master yet. I think it is legitimate to merge. 

@lzm0706 Thank you for pointing this out and using multi-encoder function. Feel free to reach out if you need any help.

lzm0706:
OK, thank you!

Issue #2526, opened 2020-09-28
CTC Segmentation Questions


s920128:
I try to modify tedlium2/align1  CTCsegment example to another example like egs/aishell
but report error as following
I have no idea to solve this index error.

```2020-09-28 01:37:54,391 (asr_align:208) INFO: Decoding device=cuda
2020-09-28 01:38:09,148 (asr_align:245) INFO: (1/14326) Aligning BAC009S0724W0121
2020-09-28 01:38:10,293 (ctc_segmentation:66) INFO: CTC segmentation of 19 chars to 4.20s audio (105 indices).
2020-09-28 01:38:10,296 (ctc_segmentation:140) WARNING: IndexError: Backtracking was not successful, the window size might be too small.
2020-09-28 01:38:10,296 (ctc_segmentation:145) WARNING: Increasing the window size to: 16000
2020-09-28 01:38:10,298 (ctc_segmentation:140) WARNING: IndexError: Backtracking was not successful, the window size might be too small.
2020-09-28 01:38:10,299 (ctc_segmentation:145) WARNING: Increasing the window size to: 32000
2020-09-28 01:38:10,301 (ctc_segmentation:140) WARNING: IndexError: Backtracking was not successful, the window size might be too small.
2020-09-28 01:38:10,301 (ctc_segmentation:145) WARNING: Increasing the window size to: 64000
2020-09-28 01:38:10,303 (ctc_segmentation:140) WARNING: IndexError: Backtracking was not successful, the window size might be too small.
2020-09-28 01:38:10,303 (ctc_segmentation:148) ERROR: Maximum window size reached.
2020-09-28 01:38:10,304 (ctc_segmentation:149) ERROR: Check data and character list!
Traceback (most recent call last):
  File "/mnt/nas2/*****/espnet/tools/venv/lib/python3.7/runpy.py", line 193, in _run_module_as_main
    "__main__", mod_spec)
  File "/mnt/nas2/*****/espnet/tools/venv/lib/python3.7/runpy.py", line 85, in _run_code
    exec(code, run_globals)
  File "/mnt/nas2/*****/espnet/espnet/bin/asr_align.py", line 277, in <module>
    main(sys.argv[1:])
  File "/mnt/nas2/*****/espnet/espnet/bin/asr_align.py", line 180, in main
    ctc_align(args, device)
  File "/mnt/nas2/*****/espnet/espnet/bin/asr_align.py", line 260, in ctc_align
    config, lpz, ground_truth_mat
[...]

PR #2566, opened 2020-10-10
CTC segmentation - package update

lumaku:
The ctc_segmentation python package was updated to version 1.1.0. with a fix for the bug discussed in #2526. I recommend to update to the new version.

After this update, the character list is passed via the configuration object, and the function `prepare_text()` does not need the dictionary as parameter anymore, so it was removed.

One other change: The parameter L used for generation of the segmentation score can now be changed via `asr_align.py` arguments.


sw005320:
Many thanks, @lumaku!

PR #2568, opened 2020-10-11
ESPnet2 recipe for TIMIT

sknadig:
This adds TIMIT for egs2
Results are similar to ESPnet1 

Tensorboard logs at https://tensorboard.dev/experiment/AtWXFPJfRl6j0L1pQ3bxrg/

I'm not sure if I'm following the guidelines in `asr.sh` for scoring. Please have a look and let me know if something needs to be changed.

### PER
|decode_asr_asr_model_valid.acc.ave/dev|400|15057|84.8|11.0|4.2|2.4|17.6|99.0|                                                                                                               
|decode_asr_asr_model_valid.acc.ave/test|192|7215|82.9|12.6|4.5|2.8|19.8|99.0|


kamo-naoyuki:
Thanks.

- Could you check the CI error and fix it? Linter test is failed now. 
- token_type=phn for `asr.sh` is exactly same as `char`, isn't it? We don't need it

sw005320 (commented):
Many thanks!
- `egs2/timit/asr1/local/path.sh` seems to be an empty file. Do you need this?
- Can you upload your model by following https://github.com/espnet/espnet_model_zoo ? The phoneme recognition pre-trained model is one of the most frequent requests for the pre-trained model.

sknadig:
I used the `egs2/TEMPLATE/asr1/setup.sh egs2/timit/asr1` to setup the recipe. `path.sh` is linked in this script.
Sure, I will upload the model.

@kamo-naoyuki I don't think the `token_type=char` will work for TIMIT as the `text` file needs to be split by the space character. I am also not sure if my implementation of scoring for `PER` is correct (it works but I don't know if we have some guidelines for this). Please let me know if the scoring function needs to be changed.
Do I need to change anything for this `codecov` CI error?

kamo-naoyuki:
Sorry, I said char by mistake. It's same as word, but asr.sh doesn't work for token_type=word, so please modify to work.

I forgot how espnet1 handle `sil` phone. 

Please ignore Codecov

kamo-naoyuki:
ESPnet1 replaces sil to space symbol.

https://github.com/espnet/espnet/blob/436cda87b05528356e0b40121383964841dfc6dc/utils/text2token.py#L129

This is equivalent to sil symbol because there are no special meanings for space characters, so I think we don't need to implement it.

Okay, so timit phone is same as word. 
Note that token_type=phone is reserved for future development.  It indicates phone level recognition using g2p for general task.

sknadig:
OK, Thanks. I will change the token_type to `word` and update the PR.
For the char level training and decoding, `token_type=char` should work without any changes.

kamo-naoyuki:
@sknadig I'll take over this PR, thank you for contribution.

sknadig:
@kamo-naoyuki Sorry I couldn't find time to update this. People go ahead.

PR #2571, opened 2020-10-16
[WIP][Feature] TMSV tts&lts recipe

Kuray107:
# Summary:
 ### In the TMSV recipe, we will do several experiments on this dataset, including Text-to-Speech and Lip-to-Speech.

 # Todo:
- [x] 1. Create a recipe for TTS (tts1), which is trained from scratch. Note that the data in TMSV is limited, so it will fail to generate the correct voice.
- [x] 2. Create a recipe for TTS (tts2), which uses the pre-trained model from COSPRO dataset. 
- [ ] 3. Create a recipe for Lip-to-Speech (lip1), which is trained from scratch.


kan-bayashi:
Thank you for your interesting PR!
Sorry for the late comment.

I have several requests:
- I'm not familiar with TMSV dataset. Could you provide the reference?
- We prefer lower case. Could you change TMSV -> tmsv?
- Please replace the copyright with your name in your created script.
- Please use a symlink in `tts2` to avoid creating duplicated scripts.
- Please update download script and remove unused one.

Kuray107:
Thanks for your great recommendation for my PR! I am still working on this PR now, so there might be some defects now, and I will try to fix them ASAP. 
 - The link to the TMSV dataset is here: https://bio-asplab.citi.sinica.edu.tw/Opensource.html#Dataset
 - For the download script, I would like to use the download_from_google_drive.sh to download the dataset, but currently the dataset is a google drive folder. I have recommended the dataset owner to change it into a compressed file, and I will fix the share_url in the data_download.sh once they accept my opinion.

PR #2574, opened 2020-10-17
[ESPnet1] Update VIVOS models

b-flo:
Updated models and model link as mentioned in #2538 
I also added another config + model with an architecture similar to https://arxiv.org/pdf/1911.12487v1.pdf where transformer block are replaced by conformer block and a RNN decoder is used. It seems to works properly but I need to rework the TDNN implementation (simple implementation with full context right now), training is really (really) slow.


sw005320 (commented):
Many thanks!
- Could you change the train/inference symlinks of this recipe to the latest one?

b-flo:
I didn't change the symlinks because the default training/inference configs are for the CTC model proposed by one of the VIVOS author (Hieu-Thi Luong). Should it be changed then?

sw005320:
I don't fully follow it, but if it is the best model configuration that the user should use, we should make it as default by changing the symlink. Otherwise, you can just keep it as it is.

sw005320:
Thanks!
Now I carefully checked the result.
I think this recipe requires a comparison between transducer and encoder-decoder, but we should do it later.

b-flo:
> I think this recipe requires a comparison between transducer and encoder-decoder, but we should do it later.

Yes, I did train models with CTC/Att previously to check performance range in comparison to transducer but didn't go further. VIVOS is mostly used to demonstrate the kind of architecture which can be used for transducer-based models.  Also, I didn't carefully tune transducer models, it could be improved I assume.

b-flo:
@sw005320 Do you want me to add Transformer model trained with CTC-Att to this PR btw? If not, it can be merged!

sw005320:
Thanks, @b-flo!
It would be great if you add that result in a later PR.

PR #2604, opened 2020-10-28
Conformer averaging model support for pytorch 1.6

siddalmia:
Averaging int tensors in Pytorch 1.6 gives a run time error - 

RuntimeError: Integer division of tensors using div or / is no longer supported, and in a future release div will perform true division as in Python 3. Use true_divide or floor_divide (// in Python) instead.

This PR will look for integer tensors and average accordingly.

### DIFF
--- a/utils/average_checkpoints.py
+++ b/utils/average_checkpoints.py
@@ -74,7 +74,10 @@ def main():
         # average
         for k in avg.keys():
             if avg[k] is not None:
-                avg[k] /= args.num
+                if avg[k].is_floating_point():
+                    avg[k] /= args.num
+                else:
+                    avg[k] //= args.num
 
         torch.save(avg, args.out)
 


utils/average_checkpoints.py
```diff
@@ -74,7 +74,10 @@ def main():
         # average
         for k in avg.keys():
             if avg[k] is not None:
-                avg[k] /= args.num
+                if avg[k].is_floating_point():
+                    avg[k] /= args.num
+                else:
+                    avg[k] //= args.num
```
sw005320, line 80: Probably casting is better?
I'm concerning the issue of "//" in https://en.wikibooks.org/wiki/Python_Programming/Operators

siddalmia: The int tensor in conformer seems to be something to do with kernel or positional information that's being saved. So I think might be the same value that's being averaged across checkpoints which case -- // shouldn't cause an issue.

sw005320 (approved):

sw005320:
Many thanks!

PR #2617, opened 2020-11-02
add librispeech conformer w/ speed perturbation + specaug

yuekaizhang:


sw005320 (commented):
Many thanks!
You can rename `run_sp.sh` as `run.sh`.
We don't want to maintain multiple `run_*.sh`

yuekaizhang:
> Many thanks!
> You can rename `run_sp.sh` as `run.sh`.
> We don't want to maintain multiple `run_*.sh`

Sure, I will fix that.

sw005320:
Many thanks!

PR #2628, opened 2020-11-04
Update WSJ result

kamo-naoyuki:
- Got better result than espnet1
- Conformer couldn't improve the result for this task.


sw005320:
It's interesting that now lm_weight is 1.2 (we emphasize more LM score than ASR score. I think it is true for the WSJ task).

kamo-naoyuki:
I added the lm_weight score in README.md.

I searched lm_weight between 0.5 - 3.0. 1.2 or 1.3 was the best.

PR #2637, opened 2020-11-05
Save model snapshot at every epoch even if save_interval_iters>0 - for model averaging

sknadig:
If `save_interval_iters>0`, the model snapshot is not saved at the end of the epoch.
This leads to an error when we try to average the n-best models after training as `model.ep.<epoch-num>` is not available, but the loss/accuracy entries for each epoch are available in `exp/<exp-name>/results/log`

This PR just removes the `else` condition for model snapshot so that along with `snapshot.iter.<iter-num>`, we also have `model.ep.<epoch-num>`.

### DIFF
--- a/espnet/asr/pytorch_backend/asr.py
+++ b/espnet/asr/pytorch_backend/asr.py
@@ -792,7 +792,7 @@ def train(args):
             torch_snapshot(filename="snapshot.iter.{.updater.iteration}"),
             trigger=(args.save_interval_iters, "iteration"),
         )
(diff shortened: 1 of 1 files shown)


sw005320:
Many thanks!
I think this change is reasonable.
We can always save the model in every epoch and additionally save it with specified iterations.

Could you fix the CI error by following https://github.com/espnet/espnet/pull/2637/checks?check_run_id=1357867185#step:8:10 ?

PR #2642, opened 2020-11-06
Update Voxforge with Conformer results

YosukeHiguchi:
Added config files and uploaded a pre-trained model for Conformer.


sw005320:
LGTM!
Many thanks!

Can you also port this result to espnet2?

PR #2654, opened 2020-11-08
lang was being used before being parsed for user input

siddalmia:
The run.sh script was broken because data_url, nbpe and datadir was being set before running utils/parse_options.sh


sw005320:
Many thanks, @siddalmia!

PR #2660, opened 2020-11-09
ASR template recipe: --srctexts -> --lm_train_text, --bpe_train_text

kamo-naoyuki:
- Bofore

`--srctext` specifies the common text corpus for lm training and bpe training. Token list is also derived from these texts.

- After

I splitted this option into --lm_train_text and --bpe_train_text because espnet1 doesn't use an external text corpus for bpe training while lm training uses it. I didn't know this difference,  sorry.
Also note that token list is derived from `lm_train_text`.

Now, this changing affects only librispeech in the existing espnet2 recipes. I already got better result using this new recipe, but still worse than espnet1. I'll retune librispeech configuration in this week.


sw005320 (approved):
LGTM!

PR #2661, opened 2020-11-09
CTC segmentation 1.4.0 feature updates

lumaku:
The `ctc_segmentation` python package was updated to v1.4.0.
This update brings some improvements and two new features to `asr_align.py`:

- Option for gratis transitions for blank - _very useful to skip larger unrelated segments within the audio file_.
- Option for inserting blanks between words; which was done in previous versions. This option is off by default since ctc_segmentation v1.4.0, as it does not work well together with gratis blank transitions. Depending on scenario, not adding spaces may bring a ~10% speedup (less tokens to be aligned) and in some cases slightly improved alignments.
- Blank token is now automatically set as the first entry of the character list; the old option to overwrite it from the char list is deprecated. This method is the default in almost all cases and therefore less error-prone.


lumaku:
The continuous integration test required an update to complete successfully, because of changes how the blank token is handled. However, the interface in `asr_align.py` was not changed. Thus, older versions of espnet since v0.9.2? are compatible with the newer version of the python package.

test/test_ctc_segmentation.py
```diff
@@ -21,10 +20,11 @@ def test_ctc_segmentation():
...
     """
     config = CtcSegmentationParameters()
-    config.min_window_size = 10
-    char_list = [config.blank, "a", "c", "d", "g", "o", "s", "t"]
+    config.min_window_size = 20
+    config.max_window_size = 50
+    char_list = ["•", "a", "c", "d", "g", "o", "s", "t"]
     text = ["catzz#\n", "dogs!!\n"]
-    lpz = torch.nn.functional.log_softmax(torch.rand(30, 8) * 10, dim=0).numpy()
+    # lpz = torch.nn.functional.log_softmax(torch.rand(30, 8) * 10, dim=0).numpy()
```
sw005320, line 27: why did you use pre-generated test vectors? 
If there are some reasons to use them, which is fine.
Then, please remove this line.

test/test_ctc_segmentation.py
```diff
@@ -61,10 +61,318 @@ def test_prepare_text():
     """
     config = CtcSegmentationParameters()
     text = ["catzz#\n", "dogs!!\n"]
-    char_list = [config.blank, "a", "c", "d", "g", "o", "s", "t"]
+    char_list = ["•", "a", "c", "d", "g", "o", "s", "t"]
```
sw005320, line 64: Why are you using "•"?

sw005320 (commented):
Many thanks, @lumaku!
I have a couple of questions.

lumaku:
I copied the [unit test from the repository](https://github.com/lumaku/ctc-segmentation/blob/master/tests/test_ctc_segmentation.py) as I only want to maintain one version. The window sizes and the test vector is chosen such that it tests the IndexError try-except case exactly once, which is not always the case with a random vector. This also dropped the pytorch dependency for the package and is ~1s faster without the additional pytorch import.

For the blank, I arbitrarily chose "•" as a single-character, rarely-used token. The unit test checks the shape of `ground_truth_mat`, which a longer token such as "\<blank>" would change. So it was either choosing a short token or adapting the line `assert gtm[1] == 1` in `test_prepare_text()`.

sw005320:
OK!
Thanks for your update, @lumaku!

PR #2672, opened 2020-11-11
Added unicode support for generated configs

Piteryo:
If your tokens are not English they are shown incorrectly in the generated config. Here is a little fix to generate proper yaml.


kamo-naoyuki:
This is not a problem except for the readability of the configuration file, but maybe better for many people. I'll merge.

kamo-naoyuki:
Thanks.

PR #2680, opened 2020-11-14
Add --init_param option

kamo-naoyuki:
I removed --pretrain_path and --pretrain_key options and implemented --init_param newly.

```bash
# Load all parameters
python -m espnet2.bin.asr_train --init_param model.pth
# Load only the parameters starting with "decoder"
python -m espnet2.bin.asr_train --init_param model.pth:decoder
# Load only the parameters starting with "decoder" and set it to model.decoder
python -m espnet2.bin.asr_train --init_param model.pth:decoder:decoder
# Set parameters to model.decoder
python -m espnet2.bin.asr_train --init_param decoder.pth::decoder
# Load all parameters excluding "decoder.embed"
python -m espnet2.bin.asr_train --init_param model.pth:::decoder.embed
# Load all parameters excluding "encoder" and "decoder.embed"
python -m espnet2.bin.asr_train --init_param model.pth:::encoder,decoder.embed
```

### DIFF
--- a/doc/espnet2_training_option.md
+++ b/doc/espnet2_training_option.md
@@ -108,6 +108,25 @@ Checkpoint includes the following states.
 - Reporter state
 - torch.cuda.amp state (from torch=1.6)
 
+## Transfer learning / Fine tuning using pretrained model
+
+Use `--init_param <file_path>:<src_key>:<dst_key>:<exclude_keys>`
+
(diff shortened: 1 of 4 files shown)


kan-bayashi (approved):
LGTM
Thank you for the great feature :)

PR #2681, opened 2020-11-15
Enable multi-thread processing of json files.

Peidong-Wang:
The data2json.sh script supports multi-thread processing but is not used in current recipe.

Added `--nj ${nj}` to data2json.sh in stage 2.


sw005320:
Thanks a lot, @Peidong-Wang!

PR #2703, opened 2020-11-20
Laborotv recipe

sw005320:
LaboroTVSpeech Japanese 2,000 hour ASR recipe (see https://laboro.ai/column/eg-laboro-tv-corpus-jp/)
- [x] add the tedx test set
  - The test set processing (https://github.com/laboroai/LaboroTVSpeech#tedxjp-10k-dataset) is pending due to some errors in https://github.com/laboroai/LaboroTVSpeech/issues/1
- [ ] add the result


sw005320:
I just merged this recipe. We can add some results in later PRs.

PR #2707, opened 2020-11-20
Wandb integration

kamo-naoyuki:


sw005320:
This is very cool!
I really wanted to use wandb.
Many thanks!

kamo-naoyuki:
Please give me feedback if you use. Due to the security policy of my company, http connection during training is very hard, so probably I wont't use this feature.

PR #2715, opened 2020-11-25
Added running as a service

MikeyBeez:
I updated the readme and added MyTTsService.py.  This allows espnet to run in a virtual environment separate from whatever else is running -- yet be available.

### DIFF
--- a/MyTTsService.py
+++ b/MyTTsService.py
@@ -1,0 +1,49 @@
+#!/usr/bin/env python3
+
+import socket
+s = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
+s.bind(("localhost", 9988))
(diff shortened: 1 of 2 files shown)


MikeyBeez:
I've never done a merge request before; so I probably did this wrong.  Sorry.  I hope the code helps.

sw005320:
Thanks for your contribution, @MikeyBeez!
We could support such functions in the main espnet, but I thought that we can have it in the Google colab (e.g., https://colab.research.google.com/github/espnet/notebook/blob/master/espnet2_tts_realtime_demo.ipynb) or Github Gist, and adding a link to the README.md.
Then, the user can play with this example more flexibly.

PR #2720, opened 2020-11-27
JSUT conformer recipe achieving 12.0/13.9 CER(%) for dev/eval1

hchung12:
This is a recipe tuned for JSUT domain. It achieves 12.0/13.9 CER(%) for dev/eval1 set using raw feature.
The model is trained using 4 GPUs and numel batch type.
You can download trained model in https://zenodo.org/record/4292742


kamo-naoyuki:
@hchung12 Thanks!

PR #2739, opened 2020-12-01
Update Commonvoice Recipe with Conformer Settings

ftshijt:
How about just pack the best model for each lang? I don't think we should provide all the models for every experiment.


egs/commonvoice/asr1/conf/tuning/train_conformer_standard.yaml
```diff
@@ -0,0 +1,48 @@
+# network architecture
```
sw005320, line 1: what does "standard" mean?
If this means Pencheng's standard setup, I think you can just say `train_conformer.yaml`

sw005320 (commented):
Yes, you can just pack the best model for each language.
Please prepare them and reflect them on `RESULTS.md` accoridngly.

ftshijt:
It's ready for merge.

sw005320:
why does https://github.com/espnet/espnet/blob/5db9f89fd7bc2d48e979c885fd5b29edabfcbc6f/egs/commonvoice/asr1/RESULTS.md#train_es_pytorch_conformer only have the test WER?

ftshijt:
> why does https://github.com/espnet/espnet/blob/5db9f89fd7bc2d48e979c885fd5b29edabfcbc6f/egs/commonvoice/asr1/RESULTS.md#train_es_pytorch_conformer only have the test WER?

It's weird. I found I made a key-interrupt at the time for this WER computing. I just reran it and added it into the RESULT.md

sw005320:
Many many thanks @ftshijt for your great contributions!

PR #2740, opened 2020-12-02
Move the dependencies coming from recipes

kamo-naoyuki:
Issue:
`pip install espnet` tries to install all dependencies including the requirements from recipes.
 
I created "recipe" command in setup.py and moved the dependencies.

- Minimal installation: `pip install espnet` 
- Inference only: `pip install espnet_model_zoo`
  - Note that `espnet_model_zoo` depends on `espnet`.
- Train using our recipe (=full installation): `pip install "espnet[recipe]"`

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -16,8 +16,6 @@ requirements = {
         "matplotlib==3.1.0",
         "pillow>=6.1.0",
         "editdistance==0.5.2",
-        "gdown",
-        "espnet_model_zoo",
         "ctc-segmentation>=1.4.0",
         "wandb",
         # DNN related packages are installed by Makefile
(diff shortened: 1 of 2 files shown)


setup.py
```diff
@@ -28,35 +26,35 @@
...
         "espnet_tts_frontend",
         # ASR frontend related
-        "museval>=0.2.1",
-        "pystoi>=0.2.2",
         "nara_wpe>=0.0.5",
         "torch_complex",
         "pytorch_wpe",
+    ],
+    "recipe": [
+        "espnet_model_zoo",
```
kan-bayashi, line 47: It is better to install model_zoo as a default since it can be used for the demo without the recipe.

kamo-naoyuki: The default modules are directly imported in `espnet` or `espnet2` module and `espnet_model_zoo` is not used.
I think we should decide it according to this definite rule to avoid confusion.

- Minimal installation: `pip install espnet` 
- Train using our recipe: `pip install "espnet[recipe]"`
- Inference only: `pip install espnet_model_zoo`
  - Note that `espnet_model_zoo` depends on `espnet`.

kan-bayashi: I see. Now I understand the point.

kamo-naoyuki:
Ideally, it's better to split the modules only for training: tensorboard, wandb, etc.

```sh
pip install "espnet[train]"
```

Now they are imported in main modules, so we can't split. 
We need to change them to dynamic importing in training modules, but it requires a little more refactoring. I won't do it for a while.

sw005320 (approved):
LGTM.
Thanks!

PR #2748, opened 2020-12-03
Change warp-ctc and warp-transducer to extra 

kamo-naoyuki:
`warp-ctc` and `warp-transducer` are likely to cause installation error, but many people doesn't need these library, so they should be optional.

### DIFF
--- a/ci/install.sh
+++ b/ci/install.sh
@@ -23,7 +23,7 @@ ${CXX:-g++} -v
     pip3 install pip==20.2.4
     make TH_VERSION="${TH_VERSION}"
 
-    make nkf.done moses.done mwerSegmenter.done pesq pyopenjtalk.done py3mmseg.done
+    make warp-ctc.done warp-transducer.done chainer_ctc.done nkf.done moses.done mwerSegmenter.done pesq pyopenjtalk.done py3mmseg.done
     rm -rf kaldi
 )
 . tools/activate_python.sh
--- a/doc/installation.md
+++ b/doc/installation.md
@@ -159,6 +159,8 @@ We also have [prebuilt Kaldi binaries](https://github.com/espnet/espnet/blob/mas
         ```sh
         $ cd <espnet-root>/tools
         $ ./setup_anaconda.sh [output-dir-name] [conda-env-name] [python-version]
+        # e.g.
+        $ ./setup_anaconda.sh anaconda espnet 3.8
         ```
         If `[output-dir-name]` is omitted, `venv` is generated.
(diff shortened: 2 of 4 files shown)


sw005320 (commented):
Can you add some installation documents when people try to use warp-ctc and warp-transducer options without installing them?
This would not be a good example, but the other optional tool like moses has 
the following error handles https://github.com/espnet/espnet/blob/dab2092bc9c8e184c48cc6e603037333bd97dcd1/egs2/how2/asr1/local/path.sh#L2-L7 so that people can easily install moses when they require it.

kamo-naoyuki:
Unlike tokenizer.perl, warp ctc and warp transducer can be used in all recipe changing the configuration , so we can't add such instruction in path.sh.

I have no idea. I just added an instruction in docs.

sw005320:
> Unlike tokenizer.perl, warp ctc and warp transducer can be used in all recipe changing the configuration , so we can't add such instruction in path.sh.
> 
> I have no idea. I just added an instruction in docs.

Hmm, then I don't want to approve this change, as it would affect many on-going projects.
Could you reconsider it?

sw005320:
How about making warp ctc and warp transducer installation as default but preparing the installation option without them, and adding instruction in README to use this installation option when it was failed?

kan-bayashi:
> How about making warp ctc and warp transducer installation as default but preparing the installation option without them, and adding instruction in README to use this installation option when it was failed?

I could not understand the advantage of the setting.
Since the default ASR and TTS recipes do not use these libraries, they should be optional?

kan-bayashi:
Another idea is to add the check of the installation when import the libraries.
```
try:
     import ...
except ImportError:
     raise ...
```

sw005320:
OK, if you don't agree with it, I think it is fine.
But please note that we have been working on a lot of ongoing projects and we use both of them as default internally in our config.

kan-bayashi:
After fixing the issue of PEP440, we can reconsider the default installation.

kan-bayashi:
> But please note that we have been working on a lot of ongoing projects and we use both of them as default internally in our config.

It that case, you can use `make all extra`.

sw005320:
> > But please note that we have been working on a lot of ongoing projects and we use both of them as default internally in our config.
> 
> It that case, you can use `make all extra`.

Thanks, year, I know...
Maybe, you can have more senses about it, and I just leave the decision to @kamo-naoyuki and @kan-bayashi.

kan-bayashi:
If these projects are merged, let us change the setting.

sw005320:
I see.
OK.

PR #2754, opened 2020-12-05
VCC2020 database

unilight:
The organizers of VCC2020 released the database, so I am modifying the recipes to automatically download the database. Also, since we have the ground truth files and the transcriptions of the evaluation files, we can calculate objective results like MCD, CER, and WER.

Also, I changed `utils/mcd_calculate.py` to detect voiced/unvoiced segments by power rather than f0, in order to eliminate the influence of f0 extraction errors.

- [x] vc1_task1
- [x] vc1_task2
- [x] voc1
- [x] update READMEs


kan-bayashi (approved):
Sorry for the late review.
LGTM.
If you can modify the numpy style docstring to google’s one, that is great.

unilight:
> Sorry for the late review.
> LGTM.
> If you can modify the numpy style docstring to google’s one, that is great.

Hmm.. I didn't touch python scrips though... or do you mean other files as well?

utils/mcd_calculate.py
```diff
@@ -65,6 +65,90 @@ def low_cut_filter(x, fs, cutoff=70):
     return lcf_x
 
 
+def spc2npow(spectrogram):
+    """Calculate normalized power sequence from spectrogram
+
+    Parameters
```
kan-bayashi, line 71: I mean this part.

kan-bayashi:
Maybe you copied from sprocket?
Then, let us keep as is.

PR #2765, opened 2020-12-09
Update models and results of Conformer.

pengchengguo:
Update models and results of Conformer on AIDATATANG, AURORA4, CHiME4 corpora.


sw005320:
LGTM!
Many thanks, @pengchengguo!

Issue #2595, opened 2020-10-23
Speech recognition from raw audio using Lightweight Sinc Convolutions


lumaku:
I am happy to share our work on speech recognition from raw audio data, presented on INTERSPEECH 2020:
https://arxiv.org/abs/2010.07597
In it, we describe Lightweight Sinc Convolutions (LSC), a small module that works well for classification directly on raw audio.

You can browse the code here:
https://github.com/lumaku/espnet/tree/lightweight_sinc_convolutions
The code includes the integration of LSC into RNN models and Transformer models, as well as the recipes for TEDlium2, TIMIT and Librispeech.

@sw005320  If it's OK for you and you want to include it in ESPnet, I can prepare a pull request in the next days.

sw005320:
That's very cool.
Yes, it would be great if you make a PR.
ASR from time-domain signals is one of the missing functions of espnet.
We could also combine some of our recent comforter efforts.

Fhrozen:
@lumaka Thank you for sharing an interesting new proposal.
I would like to ask about the SincNet part. Did it take long for training? Did you compare the time for training on SincNet with a model without it?

BTW, about the data preparation, you could use the dump_pcm that ESPnet already has implemented and then match it with the transformation layer.
You can have an example of this: 
https://github.com/Fhrozen/espnet/blob/291f32969ce3000d97a5bef5ce8cc040020fca8f/egs/fearless/wav_asr2/run.sh#L80-L141

Transformation layer:
https://github.com/Fhrozen/espnet/blob/tst-fs02/espnet/transform/wave.py

lumaku:
Yes, there is an increase in training time, but I did not yet measure by how much; it seemed moderate to me. As there are more layers, GPU memory consumption is increased, the batch size is reduced.

LSC has similarities to SincNet, but I'd like to emphasize that there are key differences between SincNet and LSC, mainly depthwise convolutions and log compression. LSC is optimized as lightweight input module to an end-to-end network.

My observation is that when the grouping in the depthwise convolutional layers is reduced - leading to a network that is more similar to SincNet - the complexity increases (memory consumption and training time). I think that the authors of the E2E-SincNet paper met the same problem and tried to mitigate using longer-time input context.

Also, @Fhrozen  thank you for the helpful suggestion on how to integrate this in the toolkit. I think that the dump_pcm is suitable and should be easy to port as it's similar to how we solved it.

[...]

PR #2768, opened 2020-12-09
Lightweight Sinc Convolutions for Espnet2

lumaku:
This pull request implements Lightweight Sinc convolutions for ESPnet 2.
https://arxiv.org/abs/2010.07597
See #2595 for more details and the implementation for ESPnet 1.

You can use sinc convolutions in your model, simply by adding the following lines to your yaml configuration file:

```yaml
frontend: sliding_window
frontend_conf:
    hop_length: 200
preencoder: sinc
```

Remarks:
As the input format is different to the default frontend, the step of calculating the global norm has to be re-done when switching from the default frontend, which is currently stage 9 in `asr.sh`. The default `hop_length` is set to the kaldi default, but depending on the task and model, ASR performance or memory consumption can be improved by finetuning this setting. This code was tested on RNN-based models.

#2595


kamo-naoyuki:
Could add test codes?

espnet2/asr/frontend/sinc.py
```diff
@@ -0,0 +1,229 @@
+#!/usr/bin/env python3
+# encoding: utf-8
```
kamo-naoyuki, line 2: `encoding: utf-8` is meaning less in python3, and it's not recommended by PEP

espnet2/asr/frontend/sinc.py
```diff
@@ -0,0 +1,229 @@
...
+        depthwise_kernel_size=9,
+        depthwise_stride=1,
+        depthwise_groups=None,
+        pointwise_groups=0,
+        dropout_probability=0.15,
+        avgpool=False,
+    ):
+        """Generate a block for lightweight Sinc convolutions.
+
+        :param in_channels:  Number of input channels
```
kamo-naoyuki, line 132: We are using google docstring style, please change to it.

https://sphinxcontrib-napoleon.readthedocs.io/en/latest/example_google.html

espnet2/asr/frontend/sinc.py
```diff
@@ -0,0 +1,229 @@
...
+    """Lightweight Sinc Convolutions.
+
+    Provide a frontend for raw audio input.
+    https://arxiv.org/abs/2010.07597
+    """
+
+    def __init__(
+        self,
+        fs: Union[int, str] = 16000,
+        in_channels=1,
```
kamo-naoyuki, line 28: please add type annotation for all functions.

espnet2/tasks/asr.py
```diff
@@ -53,7 +55,7 @@
 
 frontend_choices = ClassChoices(
     name="frontend",
-    classes=dict(default=DefaultFrontend),
+    classes=dict(default=DefaultFrontend, raw=SlidingWindow),
```
kamo-naoyuki, line 58: Why `raw`? 
Should be `sliding_window`

espnet2/asr/frontend/sinc.py
lumaku: I removed those lines.

espnet2/asr/frontend/sinc.py
lumaku: Changed!

espnet2/tasks/asr.py
lumaku: Changed to `sliding_window`.

espnet2/asr/frontend/sinc.py
lumaku: OK

espnet2/layers/sinc_conv.py
```diff
@@ -0,0 +1,181 @@
+#!/usr/bin/env python3
+#  2020, Technische Universität München;  Ludwig Kürzinger
+#  Apache 2.0  (http://www.apache.org/licenses/LICENSE-2.0)
+
+"""Sinc convolutions."""
+import torch
+from typing import Union
+
+CONSTANT_PI = 3.141592653589793
```
sw005320, line 9: I'm curious why you represent \pi as follows.
I just thought that you can use `math.pi`.

sw005320 (approved):
LGTM!
This is a very good example of learnable feature extraction!!!

Can you put the current result in https://github.com/espnet/espnet/blob/master/egs2/voxforge/asr1/README.md ?
I'm curious.

Also, are you planning to apply it to the other recipes as you did in your paper?
We could do it and investigate more in later PRs including the use of transformer/conformer ASR backend.

espnet2/utils/plot_sinc_filters.py
```diff
@@ -0,0 +1,260 @@
+#!/usr/bin/env python3
+#  2020, Technische Universität München; Nicolas Lindae, Ludwig Kürzinger
+#  Apache 2.0  (http://www.apache.org/licenses/LICENSE-2.0)
+
```
kamo-naoyuki, line 4: The path is not correct.
Move this script to pyscripts.

espnet2/utils/plot_sinc_filters.py
```diff
@@ -0,0 +1,260 @@
...
+            plt.plot(x, y)
+        img_path = os.path.join(args.out_folder, filename)
+        plt.savefig(img_path, bbox_inches="tight")
+        print("Plotted %s" % img_path)
+
+    plot_filters(range(len(F_mins)), "filters.%s" % args.filetype)
+    plot_filters(np.argsort(F_maxs - F_mins), "filters_len_sort.%s" % args.filetype)
+    plot_filters(np.argsort(F_mins), "filters_min_sort.%s" % args.filetype)
+
+    img_path = os.path.join(args.out_folder, "filtergraph.%s" % args.filetype)
```
kamo-naoyuki, line 253: Please use pathlib instead of os.path

espnet2/utils/plot_sinc_filters.py
```diff
@@ -0,0 +1,260 @@
...
+    filters = model["preencoder.filters.f"].detach().cpu().numpy()
+    f_mins = np.abs(filters[:, 0])
+    f_maxs = np.abs(filters[:, 0]) + np.abs(filters[:, 1] - filters[:, 0])
+    F_mins, F_maxs = f_mins * sample_rate, f_maxs * sample_rate
+    F_mins, F_maxs = np.round(F_mins).astype(np.int), np.round(F_maxs).astype(np.int)
+
+    plot_filter_kernels(filters, sample_rate, args)
+
+    def plot_filters(indices, filename):
+        x = np.linspace(0, np.max(F_maxs), np.max(F_maxs) + 1)
```
kamo-naoyuki, line 237: Please do not use internal function. 

Basically,  internal function shouldn't be used if it's not necessary

espnet2/utils/plot_sinc_filters.py
```diff
@@ -0,0 +1,260 @@
...
+    """Load the model, generate kernel and bandpass plots."""
+    parser = get_parser()
+    args = parser.parse_args(argv)
+    model_path = args.model_path
+    sample_rate = args.sample_rate
+
+    model = torch.load(model_path, map_location="cpu")
+    if "model" in model:  # snapshots vs. model.acc.best
+        model = model["model"]
+    filters = model["preencoder.filters.f"].detach().cpu().numpy()
```
kamo-naoyuki, line 228: Add type check for preencoder.

espnet2/layers/sinc_conv.py
```diff
@@ -0,0 +1,181 @@
...
+            in_channels: Number of input channels.
+            out_channels: Number of output channels.
+            kernel_size: Sinc filter kernel size (needs to be an odd number).
+            stride: See torch.nn.functional.conv1d.
+            padding: See torch.nn.functional.conv1d.
+            dilation: See torch.nn.functional.conv1d.
+            window_func: Window function on the filter, one of ["hamming", "none"].
+            fs (str, int, float): Sample rate of the input data
+        """
+        super(SincConv, self).__init__()
```
kamo-naoyuki, line 73: super().__int__()

This is python3 project, please care not to use python2 syntax.

espnet2/asr/frontend/sinc.py
```diff
@@ -0,0 +1,242 @@
...
+    ):
+        """Initialize the module.
+
+        Args:
+            fs: Sample rate.
+            in_channels: Number of input channels.
+            out_channels: Number of output channels (for each input channel).
+            activation_type: Choice of activation function.
+            dropout_type: Choice of dropout function.
+        """
```
kamo-naoyuki, line 41: please add  `assert check_argument_types()` for all __init__()

espnet2/tasks/asr.py
```diff
@@ -74,6 +76,15 @@
     default="utterance_mvn",
     optional=True,
 )
+preencoder_choices = ClassChoices(
+    name="preencoder",
+    classes=dict(
+        sinc=LightweightSincConvs,
+    ),
+    type_check=AbsFrontend,
```
kamo-naoyuki, line 84: This is not AbsFrontend.
Please define AbsPreEncoder

espnet2/layers/sinc_conv.py
lumaku: Correct!
In a previous version, we used `np.pi` and there is not yet a `torch.pi`.
My thought here was that setting the constant is faster than importing the numpy library only for that one number....
Also, one dependency less.

espnet2/utils/plot_sinc_filters.py
lumaku: moved `espnet2/utils/plot_sinc_filters.py`
to `egs2/TEMPLATE/asr1/pyscripts/utils/plot_sinc_filters.py`.
Is that the correct path?

espnet2/utils/plot_sinc_filters.py
lumaku: OK

espnet2/layers/sinc_conv.py
lumaku: Changed.

espnet2/asr/frontend/sinc.py
lumaku: Added

espnet2/tasks/asr.py
lumaku: OK

espnet2/utils/plot_sinc_filters.py
lumaku: I refactored the plotting functions.
There are then two nested functions `mel()` and `hz()` left. They depend on how parameter initialization is implemented. It is just easier to keep them nested instead of programming different versions.

espnet2/utils/plot_sinc_filters.py
lumaku: Added directly a type check for the filter values for more flexibility.

lumaku:
I will add recipes and results to the pull request later, after the code and those crazy CI tests are stable, or maybe in a separate pull request.

Fhrozen (commented):
LGTM, @lumaku Thank you for your collaboration.
About the test failed, it is because the code is consuming so much time (more than 2 secs).
I will recommend reducing the number of parameters, length of the sequence, or reduce the sampling frequency to make this faster.
Maybe using large stride will help.

espnet2/utils/plot_sinc_filters.py
kamo-naoyuki: Yes, please.

test/espnet2/utils/test_plot_sinc_filters.py
```diff
@@ -0,0 +1,44 @@
+import matplotlib
+import os
+import torch
+
+import egs2.TEMPLATE.asr1.pyscripts.utils.plot_sinc_filters as psf
+from espnet2.layers.sinc_conv import SincConv
```
kamo-naoyuki, line 6: We don't have importing system for pyscripts, so could you change this file to a script in `test_utils`?

espnet2/layers/sinc_conv.py
kamo-naoyuki: Please use math.pi.
Not faster, it's always same.

kamo-naoyuki:
> I refactored the plotting functions.
> There are then two nested functions mel() and hz() left. They depend on how parameter initialization is implemented. It is ?just easier to keep them nested instead of programming different versions.

I can't understand. What is "how parameter initialization is implemented"?

espnet2/utils/plot_sinc_filters.py
kamo-naoyuki: I can't allow such flexibility. It's d dangerous design.
Please add type check. We must add type check if new class is added.

kamo-naoyuki: Oh, sorry, you don't instantiated the model class, so we can't check the type.
Could you add checking the existence of the attribute?

egs2/TEMPLATE/asr1/pyscripts/utils/plot_sinc_filters.py
```diff
@@ -0,0 +1,318 @@
...
+
+def main(argv):
+    """Load the model, generate kernel and bandpass plots."""
+    parser = get_parser()
+    args = parser.parse_args(argv)
+    model_path = args.model_path
+    sample_rate = args.sample_rate
+
+    model = torch.load(model_path, map_location="cpu")
+    if "model" in model:  # snapshots vs. model.acc.best
```
kamo-naoyuki, line 277:
```suggestion
    if isinstance(model, dict) and "model" in model:  # snapshots vs. model.acc.best
```

Now this script doesn't work if it's state.

espnet2/asr/preencoder/sinc.py
```diff
@@ -0,0 +1,245 @@
...
+        self.choices_activation = {
+            "leakyrelu": torch.nn.LeakyReLU,
+            "relu": torch.nn.ReLU,
+        }
+
+        # initialization
+        self._create_sinc_convs()
+        self.init_sinc_convs()
+        # Sinc filters require custom initialization
+        self.espnet_initialization_fn = self.init_sinc_convs
```
kamo-naoyuki, line 69: This is a bad way. It's enexpected and there are no reason to set method dynamically.

Please change `espnet_initialization_fn` to a defined method and remove init_sinc_convs.

```
def espnet_initialization_fn
```

egs2/TEMPLATE/asr1/pyscripts/utils/plot_sinc_filters.py
kamo-naoyuki: Sorry, I missed, please ignore.

espnet2/asr/preencoder/sinc.py
```diff
@@ -0,0 +1,245 @@
...
+        input_frames = input.view(B * T, C_in, D_in)
+        output_frames = self.blocks.forward(input_frames)
+
+        # ---TRANSFORM: (B*T, C_out, D_out) -> (B, T, C_out*D_out)
+        _, C_out, D_out = output_frames.size()
+        output_frames = output_frames.view(B, T, C_out * D_out)
+        return output_frames, input_lengths  # no state in this layer
+
+    def get_odim(self, idim=400):
+        """Get output dimension by making one inference.
```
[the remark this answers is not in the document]
lumaku: Very useful for debugging, however, not necessary for training/inference. I will move this function to the testing script.

espnet2/layers/sinc_conv.py
lumaku: OK

lumaku:
@Fhrozen Thanks, that explains why the tests have a timeout. As the timeout happened during a matplotlib plot, I assumed that it had something to do with the plotting backend.......

@kamo-naoyuki Instead of using mel scale to initialize, one could use, say, any linear function. then the inverse `hz()` would look different. I'll out-group the mel scale to a separate module instead of using nested functions.

espnet2/utils/plot_sinc_filters.py
lumaku: Added the check for the key in the model dictionary.

lumaku:
@kamo-naoyuki @Fhrozen The code and tests are stable. Please review.

@kamo-naoyuki Regarding the nested functions `mel()` and `hz()`, I changed this to a separate module that maybe fits better as a namespace. See the MelScale and BarkScale modules. The Bark scale has wider filters at start, from the observation on the sinc filters is that they tend to get wider in training (see paper).

Also, I added results on voxforge - a relatively small corpus, and I observed that the net overfitted a little bit, but good enough for testing. I'll maybe add more recipes and results in a later pull request. Any recipe can be switched to Sinc convolutions by adding the above-described block to the yaml file.

kamo-naoyuki (approved):
LGTM. Many thanks for reflecting my comments! @lumaku

PR #2776, opened 2020-12-13
Update Russian open STT recipe for v1.01 of the dataset

akreal:
Current amount of training data is about 4700 hours.


sw005320:
Thanks for your catch up!

PR #2779, opened 2020-12-16
README_LightweightSincConvs.md

kamo-naoyuki:
I splitted the result file.


kamo-naoyuki:
@lumaku 
Could you write some description about pyscripts/utils/plot_sinc_filters.py in https://github.com/espnet/espnet/blob/master/egs2/voxforge/asr1/README_LightweightSincConvs.md ?
since we don't have any introduction about LightweightSincConvs.

PR #2782, opened 2020-12-16
Sinc Convolutions - add documentation for plot_sinc_filters.py

lumaku:
@kamo-naoyuki Added more documentation. Also, I changed the mode of `plot_sinc_filters.py` to executable...

The `README_LightweightSincConvs.md` in the voxforge recipe folder is maybe not the best place to document how to use the sinc convolutions module and the plotting tool, so I put some of the information into the module docstrings.

The plotted filters of the RNN network trained on voxforge look like this:
filtergraph

Comparison of initialized to learned filters looks like this:
(dashed lines for initialization values, blue line for learned filters)
filter_kernel_ensemble2

This PR is a follow-up to #2768.

### DIFF
--- a/egs2/TEMPLATE/asr1/pyscripts/utils/plot_sinc_filters.py
+++ b/egs2/TEMPLATE/asr1/pyscripts/utils/plot_sinc_filters.py
@@ -4,10 +4,11 @@
 
 """Visualize Sinc convolution filters.
 
-This program loads a pretrained Sinc convolution of an ESPnet2 ASR model and
-plots filters, as well as the bandpass frequencies.
-
-Plots are saved to the specified output directory.
+Description:
+    This program loads a pretrained Sinc convolution of an ESPnet2 ASR model and
+    plots filters, as well as the bandpass frequencies. The learned filter values
+    are automatically read out from a trained model file (`*.pth`). Plots are
+    saved to the specified output directory.
 """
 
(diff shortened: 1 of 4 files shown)


kamo-naoyuki:
I think because it's not used in `asr.sh`. Basically, I din't prepare introduction for each script because they are used in main recipe, but plot_sinc_filters.py is not used  in `asr.sh`, so  I think many users can't find this tool. We must put some introduction for this tool.

I'm not sure which place is the best, but README_LightweightSincConvs.md is not bad.

lumaku:
I see. I could write a document in the `docs` folder that shortly describes how to use the default frontend and the `sliding_window + sinc` frontend; and there we can include this information how to use the script. Would that make sense?

kamo-naoyuki:
Now, our docs is lack of description about each task, but general things is only written. 
According to the other task, 

e.g. Multi speaker TTS https://github.com/espnet/espnet/tree/master/egs2/vctk/tts1
e.g. Voice Conversion https://github.com/espnet/espnet/tree/master/egs/vcc20

I recommend you putting the description in README_LightweightSincConvs.md

(I don't think this is the ideal state, but for now, please follow this style)

lumaku:
OK - I added the relevant parts of the description.

kamo-naoyuki:
Thanks!

PR #2727, opened 2020-11-30
Transducer v4.1 (update #2444)

b-flo:
Hi,

This PR fixes/completes a bunch of things related to the last update for transducer in ESPnet1. Sorry it took so long, november was quite complicated.

- Add support for relative positional encoding in input layers (VGG2L, linear) and TDNN
- Fix input layer selection in custom model building
- Fix a bug in initializer method when RNN (enc) and Transformer (dec) are used.
- Fix some docs and variable names etc.
- Extend test coverage
- VIVOS: update configs + results for transformer-based and conformer-based transducer models
- VIVOS: add config + results for conformer MTL
- Voxforge: add config + updated results for conformer-rnn transducer
- Commonvoice (Czech): add config + results for conformer-rnn transducer
- Commonvoice (Welsh): add config + results for conformer-rnn transducer

P.S: It will be the last time I update transducer-related works in ESPnet1 (outside general maintenance), I'll swap completely to ESPnet2.

PR #2784, opened 2020-12-17
Minor fix rnnt

b-flo:
This is a follow-up to #2727 : 

- Rename `custom_(tt|ct).yaml` to `tdnn_(transformer|conformer)-rnn_transducer.yaml` and update README
- Remove  `warp-rnnt` from `check_install.py`
- Modified `warp-rnnt` install to automatically find cuda version (previously an argument was passed to the script from Makefile)
- Added installation instructions for Warp RNNT to `installation.md`
- Fixed a bug found by @yuekaizhang where an error is raised during inference if `trans-type: warp-rnnt` and no GPU is found. The error is raised when loading trained model, the initialization of transducer loss will fail due to the checking.
I'm not sure if it's the proper way to fix it, I couldn't rely on `self.training` because `self.eval()` is called after loading model.


doc/installation.md
```diff
@@ -239,9 +239,14 @@ e.g.
     ```sh
     cd <espnet-root>/tools
     . activate_python.sh
-    . ./setup_cuda_env.sh <cuda-root>  # e.g. <cuda-root> = /usr/local/cuda
     ./installers/install_warp-transducer.sh
```
kamo-naoyuki, line 242: This line is required

tools/check_install.py
```diff
@@ -101,17 +99,6 @@ def main(args):
             logging.warning("--> %s is not installed.\n###### Raw Error ######\n%s#######################" % (name, traceback.format_exc()))
             is_correct_installed_list.append(False)
 
-    # warp-rnnt was only tested and successfull with CUDA_VERSION=10.0
-    # however the library installation is optional ("warp-transducer" is used by default)
```
kamo-naoyuki, line 105: Why you removed

kamo-naoyuki:
Please split PRs from the next PR.  One PRs can only change one thing.
   (For now, it's okay)

doc/installation.md
```diff
@@ -239,9 +239,14 @@ e.g.
...
     ```
+- To install Warp RNNT
+    ```sh
+    cd <espnet-root>/tools
```
kamo-naoyuki, line 246: I think this is not necessary. RNNT is not worth to dare to write

doc/installation.md
b-flo: Hum, I can install both `warp-transducer` and `warp-rnnt` without executing `setup_cuda_env.sh`?

tools/check_install.py
b-flo: Because you made `warp-transducer `and` warp-rnnt `optional and `warp-transducer` was already removed from checking. Should I keep it?

doc/installation.md
b-flo: OK I'll remove it. However I'll revisit warp-transducer vs warp-rnnt, it seems I have some (small) differences compared to my initial benchmark!

doc/installation.md
kamo-naoyuki: It's required.

tools/check_install.py
kamo-naoyuki: Basically, I carefully wrote it. Please keep.

doc/installation.md
kamo-naoyuki: These tools are shown for examples. We don't need to show all.

tools/check_install.py
b-flo: Sorry I'm a bit confused with this one. I further modified to have the whole picture but I'll revert after I understood:

 1) `warp-transducer` (and also `warp-ctc`) checking was already removed before this PR  (lines ares commented)
 1) What I removed are only additions/modifications I introduced in #2444. It concers `warp-transducer` + `warp-rnnt` and some special cases for them: checking in case ESPnet was installed in CPU mode + manual appending to `library_list`(l.106 - l.113) and checking version through dpkg for both rnnt lib because they forgot adding `__version__` in some cases (l.134 - L.141).

kamo-naoyuki: This is temporary state. Please keep.

b-flo: OK, file reverted.

doc/installation.md
```diff
@@ -239,7 +239,7 @@ e.g.
     ```sh
     cd <espnet-root>/tools
     . activate_python.sh
-    . ./setup_cuda_env.sh <cuda-root>  # e.g. <cuda-root> = /usr/local/cuda
+    . ./setup_cuda_env.sh <cuda-root>
```
kamo-naoyuki, line 242: Would you revert as it is?

kamo-naoyuki (approved):
Thanks.

PR #2791, opened 2020-12-20
Add LibriTTS recipe for ESPnet2

kan-bayashi:
Once finished the training, I will upload the results.


sw005320:
Very coool!
Many thanks!

kan-bayashi:
GST model does not work well.
I will consider GST + X-vector in #2796.

PR #2793, opened 2020-12-21
ESPnet2 recipe for commonvoice

hchung12:
ESPnet2 recipe for commonvoice domain. 

1. Some sound files have two channels and they cause exception at stage 9. So, I modify the data_prep.pl
2. To keep the original asr.sh and process different languages in the same script, local/asr.sh is used and local/data.sh is slightly modified.
3. The model is trained using 4 GPUs and numel batch type.


kamo-naoyuki:
Thanks.

We can't add local/asr.sh as our policy.
Please explain the difference.

hchung12:
The difference with the original asr.sh is that the local/asr.sh makes language specific directory for token_listdir, asr_tag, lm_tag, asr_stats_dir, lm_stats_dir variables by appending $lang value is to each tag variables.

For example, 
if lang=cy in run.sh, the experiment directory is "exp/asr_train_asr_conformer5_raw_cy_bpe150_sp"
if lang=es in run.sh, the experiment directory is "exp/asr_train_asr_conformer5_raw_es_bpe150_sp"
,,,

kamo-naoyuki:
Why didn't you add it in common asr.sh?

hchung12:
I made local version just to keep the original name convention.
If it is okay to append $lang to every language related tags in asr.sh, I will modify the template asr.sh.

kamo-naoyuki:
It's okay

kamo-naoyuki (approved):

kamo-naoyuki:
Thanks! I'll merge.

TParcollet:
@hchung12 HI there, may I ask why you decided to go with the validated set and then remove the duplicate sentences rather than the train.tsv already provided ? What is the difference ?

hchung12:
@TParcollet, I just used data preparation script written by @sw005320. So, @sw005320 or @kamo-naoyuki seems more proper to answer your question.

PR #2794, opened 2020-12-21
Fix the compatibility of the pretrained ASR model

kan-bayashi:
`preencoder` is added but it does not exist in the old models.
This PR fixed the above issue.


kamo-naoyuki (approved):
Thanks, I didn't notice.

PR #2796, opened 2020-12-22
Add LibriTTS recipe in ESPnet2

kan-bayashi:
This PR adds LibriTTS recipe for ESPnet2-TTS.


kan-bayashi:
It seems that LibriTTS takes time to tune the hyperparameters.
I will split the PR.

Issue #2746, opened 2020-12-03
Problem with AMI EGS2 Recipe


picheny-nyu:
I am trying to run the egs2/ami recipe (for comparison with my MALACH run) and seem to have a problem with short utterances. I am a bit surprised because the length of the input segment in this case is 3200 samples which should be 20 frames, but the input length is reported as 5 frames (see attached). Do I need to set a minimum input segment length?

Thanks
Michael

```
2020-12-02 20:14:54,628 (beam_search:358) INFO: decoder input length: 5
2020-12-02 20:14:54,628 (beam_search:359) INFO: max output length: 5
2020-12-02 20:14:54,628 (beam_search:360) INFO: min output length: 0
2020-12-02 20:14:54,828 (batch_beam_search:321) INFO: adding <eos> in the last position in the loop
Traceback (most recent call last):
  File "/misc/vlgscratch4/PichenyGroup/picheny/anaconda3/envs/espnet-sep032020/lib/python3.7/runpy.py", line 193, in _run_module_as_main
    "__main__", mod_spec)
  File "/misc/vlgscratch4/PichenyGroup/picheny/anaconda3/envs/espnet-sep032020/lib/python3.7/runpy.py", line 85, in _run_code
    exec(code, run_globals)
  File "/misc/vlgscratch4/PichenyGroup/picheny/anaconda3/envs/espnet-sep032020/lib/python3.7/site-packages/espnet2/bin/asr_inference.py", line 447, in <module>
    main()
  File "/misc/vlgscratch4/PichenyGroup/picheny/anaconda3/envs/espnet-sep032020/lib/python3.7/site-packages/espnet2/bin/asr_inference.py", line 443, in main
    inference(**kwargs)
  File "/misc/vlgscratch4/PichenyGroup/picheny/anaconda3/envs/espnet-sep032020/lib/python3.7/site-packages/espnet2/bin/asr_inference.py", line 312, in inference
    results = speech2text(**batch)
  File "/misc/vlgscratch4/PichenyGroup/picheny/anaconda3/envs/espnet-sep032020/lib/python3.7/site-packages/torch/autograd/grad_mode.py", line 15, in decorate_context
    return func(*args, **kwargs)
  File "/misc/vlgscratch4/PichenyGroup/picheny/anaconda3/envs/espnet-sep032020/lib/python3.7/site-packages/espnet2/bin/asr_inference.py", line 194, in __call__
[...]

PR #2802, opened 2020-12-23
AMI Recipe (Short UTT checker)

ftshijt:
Add a checker to detect if the speech signal is too short for subsampling. Those utts will result in an empty hypothesis.

This PR is proposed as a solution to #2746


ftshijt:
The followings are the results I got using the default setting.


## asr_train_asr_raw_char_sp
### WER

|decode_asr_lm_lm_train_lm_char_valid.loss.ave_asr_model_valid.acc.ave/ihm_dev|13057|94802|65.8|26.0|8.2|6.0|40.2|71.3|
|decode_asr_lm_lm_train_lm_char_valid.loss.ave_asr_model_valid.acc.ave/ihm_eval|12643|89666|63.8|28.6|7.5|5.9|42.0|68.5|

### CER

|decode_asr_lm_lm_train_lm_char_valid.loss.ave_asr_model_valid.acc.ave/ihm_dev|13057|451641|80.7|8.8|10.5|5.7|25.1|71.3|
|decode_asr_lm_lm_train_lm_char_valid.loss.ave_asr_model_valid.acc.ave/ihm_eval|12643|432094|79.8|10.0|10.1|5.7|25.8|68.5|

espnet2/bin/asr_inference.py
```diff
@@ -309,7 +310,13 @@ def inference(
             batch = {k: v[0] for k, v in batch.items() if not k.endswith("_lengths")}
 
             # N-best list of (text, token, token_int, hyp_object)
-            results = speech2text(**batch)
+            try:
+                results = speech2text(**batch)
+            except TooShortUttError:
+                logging.info(
```
sw005320, line 316:
```suggestion
                logging.warning(
```

espnet2/bin/asr_inference.py
```diff
@@ -309,7 +310,13 @@ def inference(
...
+                    "Utterance is too short for subsampling, return empty results"
```
sw005320, line 317: Is it possible to add the following information?
1. utterance ID (if possible)
2. how many (subsampled?) frames of the utterance ID
3. how many (subsampled?) frames are required?

e.g, "Utterance [utterance ID] has XXX frames and is too short for subsampling (it needs more than YYY frames), return empty results"

sw005320 (commented):
Many thanks!
- Can you also put "try except" in the espnet1 case as well? (I think it may happen in espnet1, right?)

@kamo-naoyuki, could you also check it?
I'm actually not sure where we should put this "try except".

espnet2/bin/asr_inference.py
ftshijt: Sure, let me have a try on it

ftshijt:
I have written a code for espnet1 but seems to need some time for testing them. Maybe we could merge it first so that you could go on to the next stage.

kamo-naoyuki (approved):
I think this is not the best way, but acceptable.

We need same workaround for

- VGG-RNN encoder
- STFT in frontend

and etc. This is TODO, i think it's okay for now.  I'm not sure we really should always do it when we'll add new module.

PR #2807, opened 2020-12-23
Add libritts recipe for ESPnet2

kan-bayashi:
This PR adds LibriTTS recipe for ESPnet2.
(Sorry for the re-creation of PRs...)


kan-bayashi:
reduction factor = 1 is not working well.
The attention in decoding is very fragile.
We need to consider the data filtering or introduce a new model scheme.

## Good



## Bad

kan-bayashi:
By trimming the all silence to be fixed length, the decoding becomes much more stable than before.

sw005320:
> By trimming the all silence to be fixed length, the decoding becomes much more stable than before.

Even for GST?
Since an x-vector based approach is not so portable, I prefer GST.

kan-bayashi:
Not yet tried, but this cleaning is clearly effective for general seq2seq TTS models.
And I changed to pack the speaker averaged x-vextor, so the demo does not require kaldi for the closed speakers in the corpus.

kan-bayashi:
Now the on-going training seems to be reasonable and acceptable.
I will merge this PR.

Curiosci:
Please what does each spectrogram represent in the attention figure ?

Issue #2684, opened 2020-11-16
Why conformer relative postional encoding left shift have not masked.


jjjjohnson:
In Transformer-XL postional encoding implementation, we need to mask top right triangle to be zero. while the [code](https://github.com/espnet/espnet/blob/5d97a488c52cd303993064e147fb3a2be20337a5/espnet/nets/pytorch_backend/transformer/attention.py#L200) does not reflect that with default rel_shift(self, x, zero_triu=False). I cannot understand why.

Thanks for any hints!

sw005320:
Thanks for your comments!
This is because it is mainly used for self-attention in the encoder and we do not need to mask the top right triangle for auto-regressive models used in the decoder.
However, the current code is confusing. 
@pengchengguo, could you add some comments to the code?
Also, we may consider the option to handle auto-regressive models in the near future.

jjjjohnson:
Thanks! @sw005320 
I think the whole purpose of rel_shift is to left shift each row of x by (L - row_index), which results in the top right triangle being zero. Based on the Transformer-XL paper it reduces the computation from  O(N2) to O(N). I still can not understand why for self-attention in the encoder, the mask can be omitted...

sw005320:
Sorry, I was talking about different things.
@pengchengguo, could you respond to this?
I also have some difficulties in understanding this relative positional encoding part.
I want to understand more and also want to have some clarifications in the code.

pengchengguo:
No problem, I am checking the paper and code now, and will explain it soon.

pengchengguo:
Hi @jjjjohnson and @sw005320,

Thanks for your suggestion.

The current implementation does exist a bug, but it may not about the top right triangle.

**For the Transformer-xl paper**, they introduce a relative positional encoding `R_{i-j}`, which means the relative distance between a query vector at position `i` and a key vector at position `j`. For an LM related task, `i` should >= `j`, because we can't see future information when training an LM. So the range of `i-j` is from `0` to `L-1` and they force the top right triangle to be zero after `rel_shift`. The resulted term (b) in the paper is like (ignoring the `q^{T}W_{k, R}`):

```
R_{0}, 0, ... , 0
R_{1}, R_{0}, ... , 0
...
R_{L-1}, R_{L-2}, ... , R_{0}
```

[...]

PR #2816, opened 2020-12-28
Fix bugs of our Conformer implementation.

pengchengguo:
This PR trys to fix two bugs of our relative positional encoding:

1. Mentioned by Tomoki. In the previous implementation, we generate position encodings and reverse them in advance, which gives encodings like `5000, 4999, 4998, ..., 1, 0`. Giving an input sample of length `L`, its positional encoding will be generated by `pos_emb = self.pe[:, L]`, which means `5000, 4999, 4998, 5000-L` instead of `L, L-1, ..., 1, 0`.
2. Mentioned by https://github.com/espnet/espnet/issues/2684.

Currently, we use absolute values of `i-j` to solve the second bug, `i` means the position of query vector while `j` means the position of key vector. This means we use the same positional encoding for the historic and future information. Besides, for future streaming purpose, we also add an option to zero the top right triangle of the attention score matrix, which means we only have access to the historical information. Here are CER results on the Aishell1 corpus.

Model | Dev | Test
-- | -- | --
Conformer (Previous) | 4.6 | 5.1
Conformer (Current) | 4.7 | 5.1
Conformer_zerotriu (Current) | 4.7 | 5.1

Actually, the best way is regrading `i-j>0` and `i-j<0` as different information. After disucssing with Xuankai, we make a plan to do it. First, we will generate the positional encodings `self.pe1 = [1, 2, ..., 5000]` in the same way. Second, we can define a new positional encoding by `self.pe * -1` and get `self.pe2 = [-1, -2, ..., -5000]`. Then, we will concatnate these two vectors like `self.pe1 = [1, 2, ..., 4999, 5000, -5000, -4999, ..., -2, -1]`. If `i-j>0`, we will use information from position `1 to 5000`. If `i-j<0`, we can use information from position `-5000 to -1` by directly using `i-j` as index.

Many thanks if you could give me more suggestions.


danpovey:
Seems to me something like
  2 * abs(i-j) + (i >= j ? 0 : -1)
would be more natural.

kan-bayashi:
I'm afraid about the compatibility with the pretrained model.
Let me think the solution.

kan-bayashi:
Could you keep the previous version of modules with renaming and some notes?
e.g. `LegacyRelPositionalEmbedding`
Then, we can use them when loading the pretrained model trained in the old version.

pengchengguo:
Great idea! I will rename the old version and add some notes to distinguish those two implementations.

kan-bayashi:
I will add version info in `args` (#2841).
Then we can switch the type in `e2e_{st,asr}_conformer.py` 
e.g. 
```python
    def __init__(self, idim, odim, args, ignore_id=-1):
        if args.transformer_attn_dropout_rate is None:
            args.transformer_attn_dropout_rate = args.dropout_rate

        # check version
		if getattr(args. "version", None) is None:  # or <= 0.9.7?
			if args.transformer_encoder_selfattn_layer_type == "rel_selfattn":
				args.transformer_encoder_selfattn_layer_type = "legacy_rel_selfattn"
				# add some warning
				logging.warning("....")

...
```

kan-bayashi:
@pengchengguo @ShigekiKarita taught me how to deal with this kind of change.
`torch.nn.Module` has `_version` attribute.
https://github.com/pytorch/pytorch/blob/c51455a7bbeb71956023c49116f428097dc223af/torch/nn/modules/module.py#L253
You can make it more simple than the above solution.
Official example:
https://pytorch.org/docs/stable/_modules/torch/nn/modules/batchnorm.html#BatchNorm2d

kan-bayashi:
```python
import torch
class OldModule(torch.nn.Module):
    def __init__(self):
        super().__init__()
class NewModule(torch.nn.Module):
    _version = 2
    def __init__(self):
        super().__init__()
    def forward(self):
        if self._version < 2:
            return self.legacy_forward()
        else:
            print("new forward")
    def legacy_forward(self):
        print("legacy forward")
    def _load_from_state_dict(self, state_dict, prefix, local_metadata, strict,
                              missing_keys, unexpected_keys, error_msgs):
        version = local_metadata.get('version', None)
        if version is None or version < 2:
            self._version = 1
        super()._load_from_state_dict(
            state_dict, prefix, local_metadata, strict,
            missing_keys, unexpected_keys, error_msgs)
old_module = OldModule()
new_module = NewModule()
new_module()
new_module.load_state_dict(old_module.state_dict())
new_module()

```

```
>> new forward
>> legacy forward
```

espnet/nets/pytorch_backend/transformer/attention.py
```diff
@@ -142,24 +144,32 @@ def rel_shift(self, x, zero_triu=False):
...
+        idx = torch.arange(0, x.size(3), device=x.device)
+        k_idx, q_idx = idx.unsqueeze(0), idx.unsqueeze(1)
+        # for pos_idx[i, j], i means the postion of query, j means the position of key
+        # the range of i-j is from -(L-1) to L-1
+        # use the absolute value of i-j, so the range is from 0 to L-1
+        rel_pos_idx = torch.abs(k_idx - q_idx)
+        # original postional encodings are generated with reversed order
+        rel_pos_idx = x.size(3) - 1 - rel_pos_idx
+        rel_pos_idx = rel_pos_idx.repeat(x.size(0), x.size(1), 1, 1)
+        x_shift = torch.gather(x, dim=3, index=rel_pos_idx)
```
hirofumi0810, line 163:
```suggestion
        x_shift = torch.gather(x, dim=2, index=rel_pos_idx)
```
Let's think  a case of `qlen != klen`.

espnet/nets/pytorch_backend/transformer/attention.py
```diff
@@ -142,24 +144,32 @@ def rel_shift(self, x, zero_triu=False):
         """Compute relative positinal encoding.
 
         Args:
-            x (torch.Tensor): Input tensor (batch, time, size).
+            x (torch.Tensor): Input tensor (batch, head, time1, time2).
```
hirofumi0810, line 147: `time1` and `time2` are confusing. Please clarify which is query/key lengths.

espnet/nets/pytorch_backend/transformer/attention.py
```diff
@@ -142,24 +144,32 @@ def rel_shift(self, x, zero_triu=False):
...
-        x = x_padded[:, :, 1:].view_as(x)
+        idx = torch.arange(0, x.size(3), device=x.device)
+        k_idx, q_idx = idx.unsqueeze(0), idx.unsqueeze(1)
+        # for pos_idx[i, j], i means the postion of query, j means the position of key
+        # the range of i-j is from -(L-1) to L-1
+        # use the absolute value of i-j, so the range is from 0 to L-1
+        rel_pos_idx = torch.abs(k_idx - q_idx)
+        # original postional encodings are generated with reversed order
+        rel_pos_idx = x.size(3) - 1 - rel_pos_idx
+        rel_pos_idx = rel_pos_idx.repeat(x.size(0), x.size(1), 1, 1)
```
hirofumi0810, line 162:
```suggestion
        rel_pos_idx = rel_pos_idx.expand_as(x)
```

espnet/nets/pytorch_backend/transformer/attention.py
pengchengguo: Many thanks for the suggestions. I think `time1=query_len` and `time2=pos_emb_len`. Actually, the length of pos_emb is the same as query (use absolute distance) or `2*time1 - 1` (use positive and negative distance). I will clarify it.

pengchengguo:
Hi, everyone. I update a new implementation of the relative positional encodings of the Conformer model. Suppose `i` means the position of query vector and `j` means the position of key vector. Instead of using the absolute values of the relative distance (`abs(i-j)`), we use different encodings for the positive relative distance (`i>j`) and the negative relative distance (`i<j`). Similar ideas can also be found in https://arxiv.org/pdf/2005.09940.pdf.

There are also some points that should be mentioned:
- I add a simple version check to solve the compatibility problems with our pre-trained models as suggested by @kan-bayashi.  @kan-bayashi also mentioned a more professional solution, but sorry I failed to run it successfully. I will modify it after implementing it correctly.
- The class `RelPositionalEncoding` does not inherit to `PositionalEncoding`, since the function `self.extend_pe()` should generate both positive and negative encodings. Although we can add a parameter like `bidirectional=False` to the function `self.extend_pe()` of `PositionalEncoding` and make the codes more compact,  I am afraid it will be ambiguous and lead to some misunderstands.

kan-bayashi:
LGTM in terms of the design.
Could you fix the CI error?
It is better to merge soon.

pengchengguo:
Hi all,
I found there still exists a CI error in test-ubuntu18. I think it may not because of the code modifications.
Besides, here are some results on the Aishell1 corpus.

Model | Dev | Test
-- | -- | --
Conformer (Previous) | 4.6 | 5.1
Conformer (bidirectional pos_emb) | 4.6 | 5.1

The modified version shows the same results as the previous implementation. I think the motivation of positional encoding is to provide unique positional vectors to the model. Even though the previous implementation has some bugs, the vectors are still different at each position.

Now, I am training a Conformer model on the SWBD corpus with the bidirectional positional encodings. I will post results after finishing the training.

Many thanks if you could give me more suggestions.

chenjiasheng:
I've just trained a 12-blocks conformer encoder (encoder only, no decoder) on 60k hours ASR corpus, using espnet/master on 20210201.
The validation triphone state accuracy is 85.0%.
Loading this weights, if I set `matrix_bd` to 0 in function `RelPositionMultiHeadedAttention.forward()`, the accuracy drops to 75.5%; if  mask upper triangle of  `matrix_bd` , the accuracy drops to 78.1.

The accuracy drop is not so dramatic as I expected.
We can speculate that the buggy espnet RPE implementation is not working any positive effect at all.
If the bug were fixed, we would see a greate improvement on the accuracy.

pengchengguo:
Hi @chenjiasheng,

It is really an interesting result. Thanks so much for your information. Do you mean whether setting `matrix_bd` on current ESPnet or this PR's implementation, the accuracy results are all 75.5%?

I also test it on a well-trained Aishell-1 Conformer model. Sorry, I don't have a pure encoder model.  I try to zero the `matrix_bd` or zero the `pos_emb`. Actually, I think these two methods are the same. Here are some CER results (Sub/Del/Ins/Err) on the test set.

Model | no zero | zero `matrix_bd` | zero `pos_emb`
-- | -- | -- | --
Conformer (Previous) |  4.9 / 0.1 / 0.1 / 5.1 | 10.8 / 1.5 / 0.8 / 13.0 | 10.8 / 1.5 / 0.8 / 13.0
Conformer (bidirectional pos_emb) | 4.9 / 0.1 / 0.1 / 5.1 | 18.7 / 5.7 / 3.4 / 27.9 | not finished 

The CER drop is not so dramatic. I thought it would be very worse like ~80%.

pengchengguo:
Hi all,

I added an option `rel_pos_type` to switch between the legacy relative positional encoding and the latest one. The default setting is `rel_pos_type=latest`. People can change it to `legacy` when using our pre-trained models.

sw005320:
Thanks!
Just a confirmation. When we read the old pre-trained model with the legacy RPE, it uses the legacy RPE, right?

pengchengguo:
When loading old pre-trained models, we should manually set rel_pos_type=legacy in model.json (or we can update the model.json in Google Drive). Becasue the function fill_missing_args() in Transformer will always set the missing parameters to default value, we are not able to distinguish them automatically.  Another option is to change the default value to legacy. If we are moving from legacy to latest, I think regard latest as default is better. How do you think about it?

sw005320:
OK.
Then, let’s make legacy as default.
This would not change the performance in the normal setup, and this is safe.

sw005320:
Thanks, @pengchengguo!

sw005320:
@pengchengguo, could you add a unit test?

D-X-Y:
It seems that the new implementation costs a lot of memory. Do anyone investigate it?

pengchengguo:
> It seems that the new implementation costs a lot of memory. Do anyone investigate it?

Hi @D-X-Y,

Does the memory cost increase largely? Sorry, we didn't investigate it. We just conducted experiments (w/ legacy RPE or w/ latest RPE) on the same machine and it didn't have an OOM problem. 

The increase of memory cost is reasonable, since the tensors are kept the dimension of `(batch, head, len_q, d_k)` for legacy implement, while for the latest version, the relative positional encoding is extended to `2 * len_1 -1` and `matrix_bd` is `((batch, head, len_q, 2*len_q-1))`.

Do you run it in a computation restricted environment, like on-device? If so, maybe you could just use absolute positional encoding with Conformer.

D-X-Y:
@pengchengguo Thanks for your reply. I'm training a middle level model (like Conformer (M)) on 8-A100 machines. For the same batch size(bins), I can use 6-layers of the decoder w/ legacy vs. 1-layer decoder w/ lastest will raise OOM. I haven't done further analysis.

For, absolute positional encoding, would you mind pointing me where can I find them in ESPNet? and what is the performance difference between AbslutePositionalEncoding vs. the current latest for the ASR task in Librispeech?

pengchengguo:
Hi @D-X-Y,

Do you mean that both your experiments raise the OOM problem? Could you please use a smaller batch size, like 5, and observe whether the increasing trend of GPU memory? Besides, could you please show me your CUDA version and PyTorch version? 

Because one of my colleagues found the OOM problem when training using other toolkit w/ A100 + CUDA 450 (not sure) + torch 1.8.  His problem even occurs with very small batch size. After changing the CUDA version to 455 and the Pytorch version, he solved this problem. It seems the problem is caused by the machine environment.

D-X-Y:
Thanks for your suggestion.
I re-runned on another K80 GPU, it seems normal now...
For the default `train_asr_conformer7_n_fft512_hop_length256.yaml` from ESPNet, I changed the batch bins to 1000, and the memory cost is as follows:
- the `latest` model costs 3322MiB
- the `legacy` model costs 3310MiB

espnet/nets/pytorch_backend/transformer/attention.py
mungsoo: Hi, I'm confused why pos_emb len is the same as the query? What if query is short than key as illustrated in https://arxiv.org/pdf/1901.02860.pdf?

PR #2817, opened 2020-12-28
update the AMI result in espnet2

sw005320:
- fix the error handling when the utterance is too short
- update the AMI result (~30%) with our latest config, but it is still worse than the [espnet1 result](https://github.com/espnet/espnet/blob/master/egs/ami/asr1/RESULTS_ihm.md#transformer-with-large-encoder-with-speed-perturbation-based-data-augmentationepoch-100) (~25%) probably due to the word LM effect. We need some tuning.


sw005320:
This is a follow-up PR of #2802 and bug fix of #2746

PR #2826, opened 2021-01-02
Support cpu/gpu utterance batch decoding for Transformer ASR model

houwenxin:
Similar to e2e_asr, this is simple implementation of recognize_batch function in e2e_asr_transformer, which could serve as a temporary solution of Transformer gpu utterance batch decoding before the official support of api v2. It has been tested on wsj. 

During the debugging process, we notice that the way of subsampling masks will lead to some encoder output inconsistency between single-utterance and batch in Conv2dSubsampling due to the padding. (which might also slightly affect training...)

For example, currently the forward function returns x_mask[:, :, :-2:2][:, :, :-2:2] , if there is a 5-frame utterance batchfied with a longer one (7 frames), then the mask is [[True, True, True, True, True, False, False], [True, True, True, True, True, True, True]], after subsampling it becomes [[True, True, True], [True, True, True]], which is inconsistent with when we input a single 5-frame utterance (from [True, True, True, True, True] to [True, True] , 2 True in masks, as illustrated in the figure below).

Therefore, we change the mask subsampling to forms like x_mask[:, :, 2::2][:, :, 2::2] so that the subsampled mask will not be affected by padding in the end.


houwenxin:
图片

sw005320:
Many thanks!
Although the `recognize_batch` function is not based on beam search api v2, this implementation can be a good reference when we extend beam search api v2 with utterance batch decoding.
@ShigekiKarita, could you take a look at this PR?

ShigekiKarita:
Can you add tests comparing `recognize_batch` vs for-loop `recognize` results like this?
https://github.com/espnet/espnet/blob/b6f7e53ade7206c33bc9beac62cdc191a0c761be/test/test_batch_beam_search.py#L55-L59

houwenxin:
> 
> 
> Can you add tests comparing `recognize_batch` vs for-loop `recognize` results like this?
> 
> https://github.com/espnet/espnet/blob/b6f7e53ade7206c33bc9beac62cdc191a0c761be/test/test_batch_beam_search.py#L89

No problem. But since I have no previous experience in using pytest, could you give a guidance on how to test the testing functions locally?

ShigekiKarita:
Thanks. These are basic testing commands:
```
pip install -e ".[test]"
pytest test/test_xxx.py
```

See also https://github.com/espnet/espnet/blob/master/CONTRIBUTING.md#unit-testing

PR #2833, opened 2021-01-03
fix the incorrect $PATH setting in tools/extra_path.sh

jumon:
When I ran `egs/iwslt18/asr1/run.sh`, I got an error as shown below even after running `cd <my-espnet-path>/egs/iwslt18/asr1/../../../tools && make mwerSegmenter.done`.
```
Error: it seems that mwerSegmenter is not installed.
Error: please install mwerSegmenter as follows.
Error: cd <my-espnet-path>/espnet/egs/iwslt18/asr1/../../../tools && make mwerSegmenter.done
```
This was because `$PATH` variable was not set properly in `tools/extra_path.sh`.
 I fixed it.


kamo-naoyuki:
thanks!

PR #2837, opened 2021-01-04
Minor fixes in CSJ

YosukeHiguchi:
I made some minor changes in run.sh:
1. Set default `backend` to `pytorch`
2. Fixed a bug in model averaging (missing `average_opts` for averaging last N models)


sw005320:
Thanks for your fix!

PR #2840, opened 2021-01-05
Change version info location

kan-bayashi:
Currently, we get the information with pkg_resources but in this case we cannot update the version info when merging the latest changes.
This PR changes version info location and fixes the above issue.


kamo-naoyuki:
Strictly speaking, `setup.py` can't depend on the module to be installed, i.e. can't `import espnet`.

FYI,  three patterns are known to define `__version__`, as far as I know.

1.  Using `pkg_resources`: Current way
2.  Put an external text file, such as `version.txt`, and get the version from it:. e.g.  https://github.com/pytorch/pytorch/blob/d0369aabe1be99d4c6c64b85c338cf0dbb50972b/tools/generate_torch_version.py#L13-L27

    ```python
    # setupy.py
    with open("version.txt") as f:
        ...
    ```
3.  Dual writing the version text at `setup.py` and `espnet.__init__.py` both.

kan-bayashi:
I see. I think we may forgot to update in the case of 3.
I will do 3.

kan-bayashi:
@kamo-naoyuki Is it ok to place `version.txt` in `espnet/version.txt`?

kamo-naoyuki:
yes

kamo-naoyuki (approved):
LGTM

Probably, okay.  I'm not sure about side effects.

PR #2848, opened 2021-01-06
Update ci setting

kan-bayashi:
- Changed conditions not to perform duplicated CI when making branch in the original repo
- Removed duplicated CI in circle CI


.circleci/config.yml
```diff
@@ -9,8 +9,9 @@ workflows:
     jobs:
       - test-ubuntu18
```
ShigekiKarita, line 10: As ubuntu18 is also tested in Github Actions, how about disabling this too?

ShigekiKarita (approved):

kamo-naoyuki:
Could you update mergify.yml?

https://github.com/espnet/espnet/blob/master/.mergify.yml#L14-L15

.circleci/config.yml
kan-bayashi: I think this is for use_conda=true so I keep it as is.

kamo-naoyuki (approved):

kamo-naoyuki:
auto-merge doesn't work if mergify.yml is updated in the PR.

PR #2866, opened 2021-01-11
Option for GTN CTC mode

brianyan918:
This PR contains an implementation of CTC using FB's GTN package, enabling another ctc_type called 'gtnctc'. This mode was tested using Voxforge Italian: 

Model | CER | WER
-- | -- | --
Conformer (warpctc) | 8.5 | 30.0
Conformer (builtin) | 8.6 | 30.6
Conformer (gtnctc) | 8.5 | 30.0

* training curve shows spikes for builtin (red) and gtnctc (blue), but not for warpctc (orange)
image

* training times, from single 2080 GPU trials

Model | Train Time
-- | --
Conformer (warpctc) | 16h 4m
Conformer (builtin) | 17h 23m
Conformer (gtnctc) | 19h 54m


sw005320:
Added @awni
We are planning to support gtn: first as another CTC, then try to leverage the full gtn-based differentiable wFSTs in the near future.
We welcome any comments from you!

awni:
Hi! This looks great, I'm glad to see GTN is working. The run time in comparison to warp CTC and builtin (PyTorch?) is not bad, I assume the timings are dominated by something else in the pipeline. I will take a look at the implementation.

For what it's worth, we have seen these kind of spikes before in CTC training though I don't recall what caused them. They haven't been an issue for us though they are somewhat concerning..

espnet/nets/pytorch_backend/gtn_ctc.py
```diff
@@ -0,0 +1,84 @@
+import torch
+import gtn
+
+class GTNCTCLossFunction(torch.autograd.Function):
```
awni, line 4: It looks like this code is mostly copied from https://github.com/facebookresearch/gtn_applications/blob/master/utils.py#L251

I would recommend documenting as such for continuity purposes.

awni:
CC @vineelpratap who may recall some issues related to the spiking behaviour

CC @shubho who is working on a much faster GTN compose which should speed things up a lot here.

vineelpratap:
Regarding spiking behavior, is it possible that there could be samples which can trigger edge cases for CTC... Depending how the library treats them, the behavior can be different..

For example - `L = 0` or `T + R > L` where `L` - label length, `R` - label repeats, `T` - time frames

espnet/nets/pytorch_backend/gtn_ctc.py
brianyan918: Yes - it's documented now.

sw005320:
Many thanks, @brianyan918!
I just merge this PR.
But can you make the following PR?
- [ ] includes the installation (pip install) of gtn (in setup.py?)
- [ ] add a test
- [ ] please confirm that it would work in both espnet1 and espnet2
- [ ] add documentation about the CTC options in https://espnet.github.io/espnet/tutorial.html#ctc-attention-and-hybrid-ctc-attention

sw005320:
Also, k2 once you fix the installation issue.

sw005320:
@awni and @vineelpratap, thanks for the valuable comments!
As I said, we'll add further examples of gtn.
Stay tuned!

brianyan918:
@sw005320 got it. Thanks!

PR #2877, opened 2021-01-15
Minor fix of laborotv recipe

hfujihara:
Updated the way laborotv recipe handles the test set TEDxJP-10K following updates of https://github.com/laboroai/LaboroTVSpeech and https://github.com/laboroai/TEDxJP-10K


kan-bayashi:
Thank you for your kind modification!

PR #2887, opened 2021-01-19
[ESPnet1] Refactoring RNNT

b-flo:
This PR removes RNN decoder w/ att for ESPnet1, switch LSTMcell to LSTM and refactor beam search according to these changes. As I splitted a bigger PR into smaller ones, there must be some typos left in this one which will be fixed in next PRs.

Also:
1) I'm going to make once again incompatible the pre-trained transducer-based models. I don't know when I'll update them, it will depend on future additions.
2) Decoder pre-initialization w/ pre-trained LM won't work, it will be updated in a later PR.

PR #2898, opened 2021-01-25
[ESPnet1] Renaming

b-flo:
Follow up to #2887: define custom type instead of using transformer to use customizable architecture w/ transformer, conformer, tdnn, etc. I think it's also better for a bunch of things (future addition, composition w/ multiple types, general naming, etc). I also updated configs, methods and documentation.


sw005320:
LGTM!
Thanks, @b-flo!

PR #2910, opened 2021-01-28
Fix model averaging

b-flo:
Follow #2898, fix checking for model averaging in run scripts (`transformer` -> `custom`).

Edit: don't mind second commit, I dunno why it was added here but it was already covered in #2887


sw005320:
Thanks, @b-flo!

PR #2901, opened 2021-01-26
Refactor Must-C recipe

hirofumi0810:
- Added the original dev set (not trimmed) to automatic evaluation in stage-5 + some refactoring.
- Enabled GPU decoding in stage-5.

PR #2911, opened 2021-01-28
Refactor Mboshi-French corpus

hirofumi0810:
Same refactoring as #2901 + add Conformer results.
Conformer is effective for very low-resource ST task.


sw005320:
The result is really impressive!
Thanks!

PR #2916, opened 2021-01-29
Update IWSLT18 (ST-TED) ASR recipe

hirofumi0810:
- Changed to Transformer
- Added SpecAugment
- Same refactoring as #2901 
- Added patch for mwerSegmenter, which is written by `python2`


hirofumi0810:
@sw005320 This PR is ready.

utils/average_checkpoints.py
```diff
@@ -115,7 +121,13 @@ def get_parser():
         default="",
         type=str,
         nargs="?",
-        choices=["acc", "bleu", "cer_ctc", "loss", "perplexity"],
+        choices=["acc", "bleu", "cer", "cer_ctc", "loss", "perplexity"],
+    )
+    parser.add_argument(
```
sw005320, line 126: What are the intentions of this option?

hirofumi0810: @sw005320 Sorry. In the initial experiments, I found `cer_ctc` option gave better performance on this corpus. However, it turned out that `acc` option is better than it after parameter tuning. Although `cet_ctc` does not give the best performance, I just want to keep it as an option.

sw005320: OK!

utils/average_checkpoints.py
```diff
@@ -115,7 +121,13 @@ def get_parser():
...
+        "--max-epoch",
+        default=10000000,
+        type=int,
+        nargs="?",
```
sw005320, line 130: How about this option?
How is this option used?

hirofumi0810: @sw005320 I usually use this option to check overfitting and compare multiple models epoch by epoch. If you don't prefer it, I'll remove it. But I'd like to add it in another PR.

sw005320: Thanks!
We can keep this option then.

sw005320:
I just merged!
Thanks, @hirofumi0810!

Issue #2897, opened 2021-01-23
Question about the common voice results


wenjie-p:
Hi, I have run the egs/commonvoice/asr1 code using the default config on the German (v5.1-2020-06-22) dataset. The training config is conf/tuning/train_transformer.yaml, I am glad that you have provide the pretrained model and the results. However, I found neither of them is similar to my exp results:

1. The WER of the transformer results on the German test set is 19% as shown in the RESULTS.md, but I got a much lower WER after several(<100) epochs.
2. Then I decide to decode using the pretrained [model](https://drive.google.com/file/d/1IYqKPhaiQ0iy2kPxHej7Q0A3-_DOpC0U/view?usp=sharing), but got a WER of 84.2%.

So my question is:

1. How do you train these models for each language, in the fashion of monolingual or multilingual? FYI, my model was trained in the monolingual fashion.
2. If the model was trained in the monolingual fashion, why the results are so different?

ftshijt:
We already solved the issue via emails. But I would still like to explain the issue here to other curious people.

> How do you train these models for each language, in the fashion of monolingual or multilingual? FYI, my model was trained in the monolingual fashion.

The model published is trained in the monolingual fashion.

> If the model was trained in the monolingual fashion, why the results are so different?
The major difference between RESULTS.md and your experiments is that the result we publish on common-voice did not apply spec-augmentation and speech perturbation. However, in the `run.sh`, we default set those augmentation methods to be on.

> Then I decide to decode using the pretrained model, but got a WER of 84.2%.

We should not observe such large WER, I will go over this model link this week.

wenjie-p:
Hi Jiatong,

My apology for forgetting to mention about this issue in the email. Yes the problem was solved through the email and thanks for the great toolkit you guys have built! If you want any information (e.g. the pretrained models performance details), I would be very happy to help with!

wenjie-p:
Update:

I found that training the `de` German part will cause `nan loss`, during the training of `espnet`, however, the loss is pretty normal. Thus I think `espnet` have done some magic to handle this, after going through the code, I found `espnet` will not update the model if the loss is `nan` (shown in `espnet/espnet/asr/pytorch_backend/asr.py`). The `git log` shows that such functionality was added to `espnet` in Fri Apr 17 22:38:30 2020 with commit id `7cb8a4f4c0dd689ac619d65d86f11acc0f157e0c`.

[...]

PR #2927, opened 2021-02-02
Update transformer training results on common vioce dataset

wenjie-p:
Update the transformer training results of German in part of the `Common Voice` dataset, which has been discussed in #2897


wenjie-p:
@ftshijt Hi Jiatong, it seems that some checks failed, what should I do to fix such issue?

ftshijt:
You don't have to worry about the code-cov for documentation.

Many thanks, the results look very cool!

PR #2932, opened 2021-02-03
Minor refactoring

b-flo:
- Minor clean-up and optimizations for `default_beam_search`, `nsc_beam_search` and decoder methods related to inference
- Split decoder and joint network (mainly for auxiliary task)
- Fix transformer/decoder caching and `batch_score` method
- Fix typo in a VIVOS config (`transformer` -> `custom`)
- Make standalone initialization for transducer models
- Migrate `ErrorCalculatorTransducer` to its own file
- Set higher timeout for multi-GPU test with transducer models (@ShigekiKarita: initial contribution guide mentions `@pytest.mark.timeout(sec)` however from my understanding it won't set local timeout during test execution? Instead `@pytest.mark.execution_timeout(sec)` should be used, is that correct?)

On another note, I'm facing a (known?) issue with RNN layers, multi-GPUs and DataParrallel. Basically, using a RNN-transducer for example:
-  If I'm using torch <=1.4 without flattening parameters, the following warning will be displayed at every step: 
```
/opt/conda/conda-bld/pytorch_1591914880026/work/aten/src/ATen/native/cudnn/RNN.cpp:1269: UserWarning: RNN module weights are not part of single contiguous chunk of memory. This means they need to be compacted at every call, possibly greatly increasing memory usage. To compact weights again call flatten_parameters().
```
Until it breaks with the following error:
```
RuntimeError: set_storage is not allowed on a Tensor created from .data or .detach().
If your intent is to change the metadata of a Tensor (such as sizes / strides / storage / storage_offset)
without autograd tracking the change, remove the .data / .detach() call and wrap the change in a `with torch.no_grad():` block.
```
-  If I'm using torch <=1.4 with flattening parameters, it will work without warning however there seems to be an underlying issue causing degradation (I will provide the link to the related discussion in pytorch when I retrieve it)
- If I'm using torch >=1.4, the error from first bullet will be displayed directly and linked to the RNN encoder because of the flattening parts. If I switch to a custom encoder and keep RNN decoder (w/ LSTM layers and no flattening), a warning will be displayed at the end of the training but it won't cause an error or degradation.

I saw @Emrys365 comment [on the issue](https://github.com/pytorch/pytorch/issues/21108#issuecomment-552948320) some time ago but it seems nobody complained since then so I'm not sure if the issue is on my side..?


sw005320:
> Split decoder and joint network (mainly for auxiliary task)

I like this change.
The design looks better with this.

b-flo:
> > Split decoder and joint network (mainly for auxiliary task)
> 
> I like this change.
> The design looks better with this.

Yes, I agree.
Also, in regard to design, for auxiliary task I need to use a modified (RNN) `Encoder` to access intermediate layer outputs. Because I don't want to overload other models I made a new class for transducer models based on `rnn/encoders.py` I modified for the task. Is that okay for you?

ShigekiKarita:
> Set higher timeout for multi-GPU test with transducer models (@ShigekiKarita: initial contribution guide mentions @pytest.mark.timeout(sec) however from my understanding it won't set local timeout during test execution? Instead @pytest.mark.execution_timeout(sec) should be used, is that correct?)

Oh you are correct. I did so in fact. 
https://github.com/espnet/espnet/blob/dab2092bc9c8e184c48cc6e603037333bd97dcd1/test/espnet2/bin/test_tts_inference.py#L57
Can you update CONTRIBUTING.md?

sw005320:
> I saw @Emrys365 comment on the issue some time ago but it seems nobody complained since then so I'm not sure if the issue is on my side..?

@Emrys365, any comment?

Emrys365:
> I saw @Emrys365 comment [on the issue](https://github.com/pytorch/pytorch/issues/21108#issuecomment-552948320) some time ago but it seems nobody complained since then so I'm not sure if the issue is on my side..?

Yes, I was stuck by this problem at that time. Later I found a solution that solved it for `torch.nn.DataParallel` [here](https://github.com/pytorch/pytorch/pull/21197), but also noticed it didn't work for `torch.nn.parallel.data_parallel`.

Are you having this problem with `torch.nn.DataParallel` or `torch.nn.parallel.data_parallel`?

b-flo:
Thank you for your answer @Emrys365 !

> Are you having this problem with torch.nn.DataParallel or torch.nn.parallel.data_parallel?

If I'm not mistaking `torch.nn.parallel.data_parallel` is used in ESPnet1? I rarely use multi-GPUs and only found the issue when testing transducer decoder after switching from LSTMCell to LSTM. However, I met the same error with current master when upgrading my pytorch version. For example with the following env:

```
## Environments
- date: `Sat Feb  6 14:23:39 CET 2021`
- python version: `3.7.6 (default, Jan  8 2020, 19:59:22)  [GCC 7.3.0]`
- espnet version: `espnet 0.9.7`
- chainer version: `chainer 6.0.0`
- pytorch version: `pytorch 1.6.0`
- Git hash: `744d76bf8fbbf845e8f219a9b9c2eb0dc3f711fe`
  - Commit date: `Sat Feb 6 10:30:12 2021 +0100`
```

And a configuration with a RNN encoder, I end up with the following error:

```
...
  File "/home/fboye/espnet/espnet/nets/pytorch_backend/rnn/encoders.py", line 72, in forward
    rnn.flatten_parameters()
  File "/home/fboye/espnet/tools/venv/lib/python3.7/site-packages/torch/nn/modules/rnn.py", line 151, in flatten_parameters
    self.batch_first, bool(self.bidirectional))
RuntimeError: set_storage is not allowed on a Tensor created from .data or .detach().
If your intent is to change the metadata of a Tensor (such as sizes / strides / storage / storage_offset)
without autograd tracking the change, remove the .data / .detach() call and wrap the change in a `with torch.no_grad():` block.
```

Emrys365:
> > Are you having this problem with torch.nn.DataParallel or torch.nn.parallel.data_parallel?
> 
> If I'm not mistaking `torch.nn.parallel.data_parallel` is used in ESPnet1? I rarely use multi-GPUs and only found the issue when testing transducer decoder after switching from LSTMCell to LSTM. However, I met the same error with current master when upgrading my pytorch version. For example with the following env:
> 
> ...
> 
> And a configuration with a RNN encoder, I end up with the following error:
> 
> ...

Honestly, I couldn't tell what is the cause. Could you find a minimal example that can reproduce the error you are observing?

b-flo:
> Honestly, I couldn't tell what is the cause. Could you find a minimal example that can reproduce the error you are observing?

In the env I described, the code snippet you provided [here](https://github.com/pytorch/pytorch/issues/21108#issuecomment-552948320) would produce the same error!

b-flo:
I guess it's ready for review and merging. We can investigate and address issue related to flatten_parameters in a later PR, as it doesn't seem to be an issue for other users?

sw005320:
Thanks!
I just merged it.

desh2608:
> Thank you for your answer @Emrys365 !
> 
> > Are you having this problem with torch.nn.DataParallel or torch.nn.parallel.data_parallel?
> 
> If I'm not mistaking `torch.nn.parallel.data_parallel` is used in ESPnet1? I rarely use multi-GPUs and only found the issue when testing transducer decoder after switching from LSTMCell to LSTM. However, I met the same error with current master when upgrading my pytorch version. For example with the following env:
> 
> ```
> ## Environments
> - date: `Sat Feb  6 14:23:39 CET 2021`
> - python version: `3.7.6 (default, Jan  8 2020, 19:59:22)  [GCC 7.3.0]`
> - espnet version: `espnet 0.9.7`
> - chainer version: `chainer 6.0.0`
> - pytorch version: `pytorch 1.6.0`
> - Git hash: `744d76bf8fbbf845e8f219a9b9c2eb0dc3f711fe`
>   - Commit date: `Sat Feb 6 10:30:12 2021 +0100`
> ```
> 
> And a configuration with a RNN encoder, I end up with the following error:
> 
> ```
> ...
>   File "/home/fboye/espnet/espnet/nets/pytorch_backend/rnn/encoders.py", line 72, in forward
>     rnn.flatten_parameters()
>   File "/home/fboye/espnet/tools/venv/lib/python3.7/site-packages/torch/nn/modules/rnn.py", line 151, in flatten_parameters
>     self.batch_first, bool(self.bidirectional))
> RuntimeError: set_storage is not allowed on a Tensor created from .data or .detach().
> If your intent is to change the metadata of a Tensor (such as sizes / strides / storage / storage_offset)
> without autograd tracking the change, remove the .data / .detach() call and wrap the change in a `with torch.no_grad():` block.
> ```

@b-flo I am facing this same problem when trying multi-GPU training with LSTM-based Transducer. Did you manage to find a workaround?

b-flo:
To be honest, no, I rarely use multi-GPUs these days. However, I'll investigate the issue in a few days, because `flatten_parameters`    is blocking my next item (quantization).
What is your pytorch version? It may be solved with pytorch 1.{8, 9}+, that's one of the things I need to check.

desh2608:
I'm working with PyTorch 1.9.0 but still facing this issue. Please let me know if you manage to solve it. I'll use single GPU training in the meantime.

PR #2934, opened 2021-02-03
IWSLT'21 ASR recipe

hirofumi0810:
Baseline ASR recipe
- Concatenation of `TEDLIUM2`, `Must-C`, `ST-TED`, and `Librispeech`
- Large model is used by default
- Additional text on `TEDLIUM2` and `Librispeech` is used for LM training
- Offlimit talks are removed


ftshijt:
Everything looks great! Just wondering if you will share the results?

hirofumi0810:
@ftshijt I don't have any results now. I'll add later because it'll take more than a week.

ftshijt:
> @ftshijt I don't have any results now. I'll add later because it'll take more than a week.

Sounds cool. I will merge it as it is already in a good shape. You could start a new PR for the results. Many thanks!

PR #2936, opened 2021-02-03
minor fix of li52 and nsc recipes

sw005320:

PR #2937, opened 2021-02-04
added nsc results

sw005320:
- follow up PR of #2936


pzelasko:
Cool, I saw a similar number with the hybrid system. BTW that reminded me to make a PR in Kaldi 😉

PR #2938, opened 2021-02-04
Open-li52(add language id scoring & text case align for test set)

ftshijt:


sw005320 (commented):
Thanks!
It seems that the normalization issue was only in the test set and I do not seem to re-train the model :)

Will you also change `asr.sh` to accept `score_lang_id`?
Another option would be to add this function in `local/score.sh`
Then, you may not have to change `asr.sh` according to https://github.com/espnet/espnet/blob/master/egs2/TEMPLATE/asr1/asr.sh#L1278
This is a bit unique function for multilingual ASR and I recommend this way.

ftshijt:
> Thanks!
> It seems that the normalization issue was only in the test set and I do not seem to re-train the model :)
> 
> Will you also change `asr.sh` to accept `score_lang_id`?
> Another option would be to add this function in `local/score.sh`
> Then, you may not have to change `asr.sh` according to https://github.com/espnet/espnet/blob/master/egs2/TEMPLATE/asr1/asr.sh#L1278
> This is a bit unique function for multilingual ASR and I recommend this way.

Cool, seems that I forgot to update the related change in `asr.sh`. Currently, I'm updating the `asr.sh`. I will follow your advice on making it local. Thanks!

sw005320:
How about making `score_lang_id=true` as default and do not touch any change in `asr.sh`?
It is not ideal to modify the `asr.sh` template only for this purpose.

sw005320:
Another possibility is to make an option for `local/score.sh` in `asr.sh` like `--local_data_opts`.
This would be better.

ftshijt:
> Another possibility is to make an option for `local/score.sh` in `asr.sh` like `--local_data_opts`.
> This would be better.

Sounds cool~ I just made a change according to this option.

egs2/open_li52/asr1/run.sh
```diff
@@ -48,27 +48,28 @@ full_set="${high_resource_test_set} ${mid_resource_test_set} ${low_resource_test
 test_set=${full_set}
```
sw005320, line 48: Some changes in the file are your local setup.
You should revert them.

ftshijt: Oh! Missed that. Reflected.

sw005320:
Thanks!
Do you have some numbers?
I'm thinking of the models I shared with you as our initial pre-trained models.
I could also do it in my environment.

ftshijt:
> Thanks!
> Do you have some numbers?
> I'm thinking of the models I shared with you as our initial pre-trained models.
> I could also do it in my environment.

I just test the system with a very small subset. Could you do the decoding in your environment if possible?

PR #2939, opened 2021-02-04
Speaker diarization implementation in ESPnet

ftshijt:
This is a very initial PR for the speech diarization in ESPnet2. The method basically follows the basic EEND in PyTorch.

We use simulated mini_librispeech as the initial recipe workspace.
 
Following are the action item
- [x] Basic Model structure
- [x] Training (sequence-based)
- [x] Recipe setup
- [x] Training (chunk-based)
- [x] Inference
- [x] Scoring

Current issues:

1. [Solved] Where to frame the spk labels (from rttm to spk-label frame sequence). Currently, I frame the spk label when loading the rttm, which requires us to decide the hop_length at the beginning of the loader (e.g., [dataset](https://github.com/ftshijt/espnet/blob/diarization/espnet2/train/dataset.py#L293)). Seems not a good way.
2. [Solved] current chunk_iter_factory need training data to be the same size (see [chunk_iter_factory](https://github.com/espnet/espnet/blob/master/espnet2/iterators/chunk_iter_factory.py#L127)), this cannot hold if we feed the raw-speech and framed spk labels. Maybe add a new iter_factory for these cases?
3. [Solved] Since the encoder for diarization can be regarded as the same for asr, I'm now directly using the encoder from [asr encoder](https://github.com/ftshijt/espnet/blob/diarization/espnet2/diar/espnet_model.py#L13) (default not with conv-subsampling) 

Any comments, suggestions, and help are welcome!


sw005320:
Great job, @ftshijt!

@kamo-naoyuki, could you give us some advice about these issues?

sw005320:
Two high-level comments
- Please appropriately add some credits if you take it from EEND's sources and respect their work ;)
- For this particular project, can you make a new file with the MIT license? (Just add a `# Licensed under the MIT license.` in the header) if it is not complicated?

sw005320:
> 3. Since the encoder for diarization can be regarded as the same for asr, I'm now directly using the encoder from [asr encoder](https://github.com/ftshijt/espnet/blob/diarization/espnet2/diar/espnet_model.py#L13) (default not with conv-subsampling)

I think this is good (or I would say that this is what we want. Then, we can easily reuse other ASR encoders like conformer). 
Also, just for the future possibility. If we extend it for EEND-EDA or SC-EEND, an encoder-decoder design would be more appropriate. I think you already considered it and this is just a note.

ftshijt:
> Two high-level comments
> 
> * Please appropriately add some credits if you take it from EEND's sources and respect their work ;)
> * For this particular project, can you make a new file with the MIT license? (Just add a `# Licensed under the MIT license.` in the header) if it is not complicated?

Thanks for your comments, I will reflect on these soon.

espnet2/diar/frontend/default.py
```diff
@@ -0,0 +1,115 @@
+# Licensed under the MIT license.
+
+import copy
```
kamo-naoyuki, line 3: Is there a reason to create diar/frontend instead of asr/frontend?

espnet2/bin/diar_inference.py
```diff
@@ -0,0 +1,239 @@
...
+                / (enh.pow(2).sum(dim=1) * ref.pow(2).sum(dim=1) + EPS)
+            ).clamp(min=EPS, max=1 - EPS),
+        }[criterion]
+
+        _, perm = self.enh_model._permutation_loss(ref_wavs, enh_wavs, loss_func)
+        return perm
+
+
+def humanfriendly_or_none(value: str):
+    if value in ("none", "None", "NONE"):
```
kamo-naoyuki, line 237: Could you use it?

https://github.com/espnet/espnet/blob/2cfecaf70a2d2826e43db9a2aa34305c6628c5a5/espnet2/utils/types.py#L75

sw005320:
@ftshijt, speech diarization —> speaker diarization

espnet2/fileio/rttm.py
```diff
@@ -0,0 +1,119 @@
...
+    assert check_argument_types()
+    spk_index = 0
+    data = {}
+    spk_dict = {}
+    with Path(path).open("r", encoding="utf-8") as f:
+        for linenum, line in enumerate(f, 1):
+            sps = re.split(" +", line.rstrip())
+
+            # RTTM format must have exactly 9 fields
+            assert len(sps) == 9 and path
```
kamo-naoyuki, line 38: Could you always add a message after assertion check to show the status for debugging?

```python
assert len(sps) == 9 and path, len(sps)
```

What `and path` is?

espnet2/fileio/rttm.py
```diff
@@ -0,0 +1,119 @@
...
+            label_type, utt_id, channel, start, duration, _, _, spk_id, _ = sps
+
+            # Only support speaker label now
+            assert label_type == "SPEAKER"
+
+            if spk_id not in spk_dict.keys():
+                spk_dict[spk_id] = spk_index
+                spk_index += 1
+            data[utt_id] = data.get(utt_id, []) + [
+                (spk_id, float(start), float(start) + float(duration))
```
kamo-naoyuki, line 48: What happens if there are overlap speakers?

kamo-naoyuki: Sorry , i missed. Please ignore

kamo-naoyuki:
>1. Where to frame the spk labels (from rttm to spk-label frame sequence). Currently, I frame the spk label when loading the rttm, which requires us to decide the hop_length at the beginning of the loader (e.g., [dataset](https://github.com/ftshijt/espnet/blob/diarization/espnet2/train/dataset.py#L293)). Seems not a good way.
>2. current chunk_iter_factory need training data to be the same size (see [chunk_iter_factory](https://github.com/espnet/espnet/blob/master/espnet2/iterators/chunk_iter_factory.py#L127)), this cannot hold if we feed the raw-speech and framed spk labels. Maybe add a new iter_factory for these cases?

Why don't you keep the duration information in time domain? Then you can use chunk_iter_factory as it is. You covert time domain information to frame region in espnet model.

espnet2/train/preprocessor.py
```diff
@@ -377,3 +387,103 @@ def __call__(
                 data[text_n] = np.array(text_ints, dtype=np.int64)
         assert check_return_type(data)
         return data
+
+
```
kamo-naoyuki, line 391: CommonPreprocessor can work with only inputting speech. I think SpeechPreprocessor is not necessary.

If you need new preprocessor, could you move the preprocessor class to espnet2/diar/preprocessor.py? (I think this other task should also do it.)

espnet2/diar/decoder/linear_decoder.py
```diff
@@ -0,0 +1,28 @@
+# Licensed under the MIT license.
```
kamo-naoyuki, line 1: Is there some reasons to use MIT license?

If you'll apply MIT license for diar task, you must put LICENSE file and also write clarify it in README.md.

espnet2/diar/frontend/default.py
ftshijt: My apology! This is something to be fixed. Thanks for pointing it out.

espnet2/diar/decoder/linear_decoder.py
ftshijt: > Two high-level comments
> 
> * Please appropriately add some credits if you take it from EEND's sources and respect their work ;)
> * For this particular project, can you make a new file with the MIT license? (Just add a `# Licensed under the MIT license.` in the header) if it is not complicated?

I was following @sw005320 's comment. This task is developed based on https://github.com/hitachi-speech/EEND , where they used the MIT license. I will clarify that later.

espnet2/train/preprocessor.py
ftshijt: Ohh, I didn't aware of that. Thanks for reminding me! Then I think I will keep CommonPreprocessor for now.

ftshijt:
> > 1. Where to frame the spk labels (from rttm to spk-label frame sequence). Currently, I frame the spk label when loading the rttm, which requires us to decide the hop_length at the beginning of the loader (e.g., [dataset](https://github.com/ftshijt/espnet/blob/diarization/espnet2/train/dataset.py#L293)). Seems not a good way.
> > 2. current chunk_iter_factory need training data to be the same size (see [chunk_iter_factory](https://github.com/espnet/espnet/blob/master/espnet2/iterators/chunk_iter_factory.py#L127)), this cannot hold if we feed the raw-speech and framed spk labels. Maybe add a new iter_factory for these cases?
> 
> Why don't you keep the duration information in time domain? Then you can use chunk_iter_factory as it is. You covert time domain information to frame region in espnet model.

This is a very good suggestion! Many thanks!

espnet2/fileio/rttm.py
ftshijt: Just mean to show the files in the wrong format. I will attach the message letter.

sw005320:
@ftshijt, can you finish this PR once you make this prototype work from training to inference to some extent?
I'm sure that this requires a couple of iterations, but we can make it with the following PRs.
Several people are interested in this PR, and they may give us some comments with their trials once it is merged.

ftshijt:
> @ftshijt, can you finish this PR once you make this prototype work from training to inference to some extent?
> I'm sure that this requires a couple of iterations, but we can make it with the following PRs.
> Several people are interested in this PR, and they may give us some comments with their trials once it is merged.

Sure, the inference is 90% ready, but I still need some time to refactor current comments. I expect to have this PR be into a runnable shape by this Sunday.

sw005320:
Sounds good.
Once we finish the first PR, we can distribute our efforts to the others who are interested in contributing to this project.

espnet2/diar/decoder/linear_decoder.py
kamo-naoyuki: @sw005320 @ftshijt 
MIT license doesn't forbid to change the LICENSE from the original. Regardless of changing or keeping the original LICENSE, only you have to do is clarifying the original copyright and pointing the URL of MIT LICENSE or writing the full text of the LICENSE. We should use same LICENSE as espnet, i.e. apache, if there are not your personal reason to use MIT LICENSE.

espnet2/diar/espnet_model.py
```diff
@@ -0,0 +1,311 @@
+#!/usr/bin/env python3
```
kamo-naoyuki, line 1: ¥Minor comment

Could you remove shebang for not executable scripts?  Shebang is originally used when executing a script, though shebang exists in some scripts now, maybe, to represent python? This might be a style, but I didn't do so in espnet2, so please remove.

ftshijt:
I'm not sure about the CI logs for travis-ci. Any suggestions or comments?

Other than this, I think the current PR is kind of ready (maybe we could start a new branch on this first). There are many many TODOs (e.g., unit-tests, recipes, architecture support, joint-train with other modules) where I will summarize soon in an issue.

kamo-naoyuki (approved):
Thanks! LGTM for my side

utils/make_rttm.py
```diff
@@ -0,0 +1,47 @@
...
+# Copyright 2019 Hitachi, Ltd. (Yusuke Fujita)
+#           2021 Johns Hopkins University (Jiatong Shi)
+# Apache 2.0  (http://www.apache.org/licenses/LICENSE-2.0)
+
+import argparse
+from espnet2.fileio.npy_scp import NpyScpReader
+import logging
+import numpy as np
+from scipy.signal import medfilt
+
```
kamo-naoyuki, line 12: Probably, the CI error comes from this file.

We automatically generate the usage of command line tools in our documentation, and to generate it,  the script need to have get_parser() function. e.g. https://github.com/espnet/espnet/blob/master/utils/compute-fbank-feats.py

\ I don't know detail.

utils/score_der.sh
```diff
@@ -0,0 +1,41 @@
...
+infer_scp=$2    # inference scp from diar_inference
+gt_label=$3     # ground truth rttm
+
+echo "Scoring at ${scoring_dir}"
+mkdir -p $scoring_dir || exit 1;
+
+for med in 1 11; do
+    for th in 0.3 0.4 0.5 0.6 0.7; do
+        make_rttm.py --median=$med --threshold=$th \
+            --frame_shift=${frame_shift} --subsampling=${subsampling} --sampling_rate=${fs} \
```
kamo-naoyuki, line 28: Sorry, I missed. 
We don't use `utils` way any more. This is one of the bad point of espnet1 because the script doesn't show the real path of the script, it makes hard to find script paths.

Please use `scripts` and `pyscripts` under the template directory instead.

alumae:
The EEND recipe uses 10x subsampling, which is also reflected in the paper ASRU 2019 paper ("To deal with a long audio sequence in our neural networks, we subsampled the concatenated features by a factor of ten."). If I understand correctly, the Espnet model doesn't use subsampling. Can anybody comment?

ftshijt:
Yeah, you are right @alumae . We currently do not use subsampling in the codebase. As the subsampling initially was applied to allow decoding long recordings, it won't be that necessary when we do chunk-based training. However, for the purpose of reducing computations, it would be some help. We would like to support the features in the next round (at least allow users could change the subsampling factor on their own). Thanks for pointing it out.

alumae:
Thanks a lot for the fast answer.
We are trying to get Callhome recipe working, hence my questions: the EEND recipe (https://github.com/hitachi-speech/EEND/blob/master/egs/callhome/v1/conf/train.yaml) uses 7-frame splicing, right? I think it's important, as the spliced features contain much more speaker-characteristic information than a single feature, maiking the job for the transformer layer much easier. Any idea how to implement this the best way in Espnet?

sw005320:
@ftshijt and @YushiUeda, can you consider including this?
I think it is necessary even for our usages.

PR #2940, opened 2021-02-04
Update IWSLT'19 recipe

hirofumi0810:
Follow-up of #2934. `iwslt19` follows `iwlst21`'s structure.
The major modification is to change `data_code` injection from suffix to prefix.


sw005320 (approved):
Thanks, @hirofumi0810!

PR #2941, opened 2021-02-04
[WIP] support k2 ctc

yaguanghu:
This is an initial PR for k2 ctc loss support in ESPnet. I'm running running experiments on Aishell datasets, and the loss seems normal.
I will also verify it in espnet2.

### DIFF
--- a/espnet/nets/pytorch_backend/ctc.py
+++ b/espnet/nets/pytorch_backend/ctc.py
@@ -54,7 +54,8 @@ class CTC(torch.nn.Module):
             self.ctc_loss = K2CTCLoss(odim, reduction=reduction_type)
         else:
             raise ValueError(
-                'ctc_type must be "builtin" or "warpctc": {}'.format(self.ctc_type)
+                'ctc_type must be "builtin", "warpctc", "gtnctc" or "k2ctc", '
+                "but it is: {}".format(self.ctc_type)
             )
 
         self.ignore_id = -1
--- a/espnet/nets/pytorch_backend/ctc.py
+++ b/espnet/nets/pytorch_backend/ctc.py
@@ -19,7 +19,12 @@ class CTC(torch.nn.Module):
     :param bool reduce: reduce the CTC loss into a scalar
     """
 
-    def __init__(self, odim, eprojs, dropout_rate, ctc_type="warpctc", reduce=True):
+    def __init__(self,
+                 odim,
+                 eprojs,
+                 dropout_rate,
+                 ctc_type="warpctc",
+                 reduce=True):
         super().__init__()
         self.dropout_rate = dropout_rate
         self.loss = None
@@ -34,7 +39,8 @@ class CTC(torch.nn.Module):
         #     else "builtin"
         # )
         if ctc_type != self.ctc_type:
-            logging.warning(f"CTC was set to {self.ctc_type} due to PyTorch version.")
+            logging.warning(
+                f"CTC was set to {self.ctc_type} due to PyTorch version.")
         if self.ctc_type == "builtin":
             reduction_type = "sum" if reduce else "none"
             self.ctc_loss = torch.nn.CTCLoss(reduction=reduction_type)
@@ -50,11 +56,12 @@ class CTC(torch.nn.Module):
             logging.info(f"CTC type is {self.ctc_type}")
             reduction_type = "sum" if reduce else "none"
             from espnet.nets.pytorch_backend.k2_ctc import K2CTCLoss
-            self.ctc_loss = K2CTCLoss(odim, reduction=reduction_type)
+            self.ctc_loss = K2CTCLoss(odim,
+                                      reduction=reduction_type)
         else:
             raise ValueError(
-                'ctc_type must be "builtin" or "warpctc": {}'.format(self.ctc_type)
-            )
+                'ctc_type must be "builtin" or "warpctc": {}'.format(
+                    self.ctc_type))
 
         self.ignore_id = -1
         self.reduce = reduce
@@ -99,7 +106,8 @@ class CTC(torch.nn.Module):
 
         self.loss = None
         hlens = torch.from_numpy(np.fromiter(hlens, dtype=np.int32))
-        olens = torch.from_numpy(np.fromiter((x.size(0) for x in ys), dtype=np.int32))
+        olens = torch.from_numpy(
+            np.fromiter((x.size(0) for x in ys), dtype=np.int32))
 
         # zero padding for hs
         ys_hat = self.ctc_lo(F.dropout(hs_pad, p=self.dropout_rate))
@@ -108,16 +116,10 @@ class CTC(torch.nn.Module):
         ys_true = torch.cat(ys).cpu().int()  # batch x olen
 
         # get length info
-        logging.info(
-            self.__class__.__name__
-            + " input lengths:  "
-            + "".join(str(hlens).split("\n"))
-        )
-        logging.info(
-            self.__class__.__name__
-            + " output lengths: "
-            + "".join(str(olens).split("\n"))
-        )
+        logging.info(self.__class__.__name__ + " input lengths:  " +
+                     "".join(str(hlens).split("\n")))
+        logging.info(self.__class__.__name__ + " output lengths: " +
+                     "".join(str(olens).split("\n")))
 
         # get ctc loss
         # expected shape of seqLength x batchSize x alphabet_size
@@ -134,9 +136,8 @@ class CTC(torch.nn.Module):
         if self.ctc_type == "gtnctc":
             # keep as list for gtn
             ys_true = ys
-        self.loss = to_device(hs_pad, self.loss_fn(ys_hat, ys_true, hlens, olens)).to(
-            dtype=dtype
-        )
+        self.loss = to_device(hs_pad, self.loss_fn(ys_hat, ys_true, hlens,
+                                                   olens)).to(dtype=dtype)
         if self.reduce:
             # NOTE: sum() is needed to keep consistency
             # since warpctc return as tensor w/ shape (1,)
@@ -198,10 +199,10 @@ class CTC(torch.nn.Module):
 
         y_int = interpolate_blank(y, blank_id)
 
-        logdelta = np.zeros((lpz.size(0), len(y_int))) - 100000000000.0  # log of zero
-        state_path = (
-            np.zeros((lpz.size(0), len(y_int)), dtype=np.int16) - 1
-        )  # state path
+        logdelta = np.zeros(
+            (lpz.size(0), len(y_int))) - 100000000000.0  # log of zero
+        state_path = (np.zeros(
+            (lpz.size(0), len(y_int)), dtype=np.int16) - 1)  # state path
 
         logdelta[0, 0] = lpz[0][y_int[0]]
         logdelta[0, 1] = lpz[0][y_int[1]]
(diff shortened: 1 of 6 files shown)


sw005320:
Added @brianyan918 

@yaguanghu, many thanks!
What I'm concerning is how we establish robust k2 installation in the espnet environment.
warpctc had several issues and we make it in the current form.
It would be great if you help us in this regard.

espnet/nets/pytorch_backend/ctc.py
```diff
@@ -25,13 +25,14 @@ def __init__(self, odim, eprojs, dropout_rate, ctc_type="warpctc", reduce=True):
...
-        self.ctc_type = (
-            ctc_type
-            if LooseVersion(torch.__version__) < LooseVersion("1.7.0")
-            else "builtin"
-        )
+        # self.ctc_type = (
+        #     ctc_type
+        #     if LooseVersion(torch.__version__) < LooseVersion("1.7.0")
+        #     else "builtin"
+        # )
```
sw005320, line 35: I think this part should be refined.
This part was made since warpctc does not support pytorch 1.7.0
gtn and k2 should work with Pytorch >= 1.7.0 while they would not work with some older Pytorch versions.
It would be great if you refine this condition and adding gtn and k2 unit test so that it could pass the CI check (we're checking PyTorch 1.0.1, 1.1.0, 1.2.0, 1.3.1, 1.4.0, 1.5.1, 1.6.0, and 1.7.1)

yaguanghu: I will try to refine this logic.

yaguanghu:
> Added @brianyan918
> 
> @yaguanghu, many thanks!
> What I'm concerning is how we establish robust k2 installation in the espnet environment.
> warpctc had several issues and we make it in the current form.
> It would be great if you help us in this regard.

K2 team members are trying to improve the installation. Hopefully we just need to run `pip install k2`.

danpovey:
Yes, should just be installable with pip, but the version to be installed depends on the PyTorch version.
I believe we support 1.6 and 1.7 (not sure the precise version numbers within that).
@csukuangfj should be able to comment.

csukuangfj:
k2 can be installed by just running
```bash
pip install k2
```
or
```bash
pip install --pre k2 # the latest version on PyPI, not necessarily the same with the latest master
```

k2 installed with pip supports:
- Python 3.6, 3.7, 3.8
- CUDA 10.1
- PyTorch 1.7.1

---

If k2 is installed from source, it is known to work with
- Python 3.6, 3.7, 3.8 (the minimum required version is 3.6)
- CUDA 10.1, 10.2, 11.0 (other CUDA versions may also work, but have not been tested)
- PyTorch 1.6.x, 1.7.x (1.6.0, 1.7.0, 1.7.1 are known to work; other versions may also work, but have not been tested)

We have a detailed documentation about the installation of k2. Please see
https://k2.readthedocs.io/en/latest/installation.html

espnet/nets/pytorch_backend/k2_ctc.py
```diff
@@ -0,0 +1,150 @@
...
+            else:
+                arcs += f'{i} {j} {tokens[j]} {tokens[j]} 0.0\n'
+        arcs += f'{i} {final_state} -1 -1 0.0\n'
+    arcs += f'{final_state}'
+    ans = k2.Fsa.from_str(arcs)
+    ans = k2.arc_sort(ans)
+    return ans
+
+
+class CtcTrainingGraphCompiler(object):
```
csukuangfj, line 89: We have another version of building graphs where all operations are moved to CUDA.
Please see 
https://github.com/k2-fsa/snowfall/blob/2e679045be6fc251e9fbb5fff614a5e35104136f/snowfall/training/mmi_mbr_graph.py#L109

yaguanghu: There may be problems with multi-gpu training if we build graphs with CUDA and put graphs in a specific GPU memories. I haven't figured out how to deal with this problem, but I don't think it's a problem because of the speed of building graphs is quite faster and the size of CTC graph is small enough.

danpovey: mm I didn't like setting the gpu_id explicitly either.  I feel like we should be getting it from somewhere else.

yaguanghu: Anyway, I have implemented the CUDA graphs building method. But it will not be used until I figure out how to work with multi-gpu training.

sw005320:
Please fix the CI error.
Check https://github.com/espnet/espnet/blob/master/CONTRIBUTING.md if you want to test the CI test locally.

espnet/nets/pytorch_backend/ctc.py
```diff
@@ -45,6 +46,12 @@ def __init__(self, odim, eprojs, dropout_rate, ctc_type="warpctc", reduce=True):
...
             self.ctc_loss = GTNCTCLossFunction.apply
+        elif self.ctc_type == "k2ctc":
+            logging.info(f"CTC type is {self.ctc_type}")
+            reduction_type = "sum" if reduce else "none"
+            from espnet.nets.pytorch_backend.k2_ctc import K2CTCLoss
+
+            self.ctc_loss = K2CTCLoss(odim, reduction=reduction_type)
         else:
             raise ValueError(
                 'ctc_type must be "builtin" or "warpctc": {}'.format(self.ctc_type)
```
csukuangfj, line 57: I think this error message has to be updated to mention `gtnctc` and `k2ctc`.

test/espnet2/asr/test_k2.py
```diff
@@ -0,0 +1,47 @@
...
+    pytorch_loss = pytroch_ctc_loss(
+        pytorch_log_probs, target, input_lengths, target_lengths
+    )
+    pytorch_loss.backward()
+
+    devices = [torch.device("cpu")]
+    if torch.cuda.is_available():
+        devices.append(torch.device("cuda"))
+    for device in devices:
+        k2_ctc_loss = K2CTCLoss(C)
```
csukuangfj, line 35: K2CTCLoss supports two extra arguments: `reduction` and `device`. It would be great if there are tests covering these two arguments too.

espnet/nets/pytorch_backend/k2_ctc.py
```diff
@@ -0,0 +1,168 @@
...
+            logging.error(
+                f"k2 ctc loss reduction type: {reduction} is not supported yet,"
+                "change it to be 'sum'"
+            )
+            self.reduction = "sum"
+        self.graph_compiler = CtcTrainingGraphCompiler(odim, device=device)
+
+    def forward(
+        self,
+        log_probs: torch.Tensor,
```
csukuangfj, line 36: Can you also add documentation to these arguments?

espnet/nets/pytorch_backend/k2_ctc.py
```diff
@@ -0,0 +1,168 @@
...
+
+        decoding_graph = self.graph_compiler.compile(targets, target_lengths)
+        if self.device != log_probs.device:
+            decoding_graph = decoding_graph.to(log_probs.device)
+
+        target_graph = k2.intersect_dense(decoding_graph, dense_fsa_vec, 100.0)
+        tot_scores = target_graph.get_tot_scores(
+            log_semiring=True, use_double_scores=False
+        )
+        if self.reduction == "none":
```
csukuangfj, line 63: What are the differences between `none` and `sum`?
Did you forget to use an early `return` inside the `if` statement?

espnet/nets/pytorch_backend/k2_ctc.py
```diff
@@ -0,0 +1,168 @@
...
+    ans = k2.Fsa.from_str(arcs)
+    ans = k2.arc_sort(ans)
+    return ans
+
+
+class CtcTrainingGraphCompiler(object):
+    def __init__(self, odim: int, device: torch.device):
+        """
+        Args:
+        odim:
```
csukuangfj, line 107: The indentation of the docstring for the argument is not correct.

espnet/nets/pytorch_backend/k2_ctc.py
```diff
@@ -0,0 +1,168 @@
...
+        decoding_graph = k2.intersect(
+            self.ctc_topo_inv, label_graph, treat_epsilons_specially=False
+        )
+        decoding_graph = k2.connect(decoding_graph.invert_())
+        return decoding_graph
+
+
+def get_tot_objf_and_num_frames(
+    tot_scores: torch.Tensor, frames_per_seq: torch.Tensor
+) -> Tuple[float, int, int]:
```
csukuangfj, line 147: The type hints for the return values are not correct. The three values are tensors, not scalars.

espnet/nets/pytorch_backend/k2_ctc.py
```diff
@@ -0,0 +1,168 @@
...
+
+        dense_fsa_vec = k2.DenseFsaVec(log_probs, supervision_segments)
+
+        decoding_graph = self.graph_compiler.compile(targets, target_lengths)
+        if self.device != log_probs.device:
+            decoding_graph = decoding_graph.to(log_probs.device)
+
+        target_graph = k2.intersect_dense(decoding_graph, dense_fsa_vec, 100.0)
+        tot_scores = target_graph.get_tot_scores(
+            log_semiring=True, use_double_scores=False
```
csukuangfj, line 61: I would recommend setting `use_double_scores` to True.

csukuangfj (changes requested):

espnet/nets/pytorch_backend/k2_ctc.py
```diff
@@ -0,0 +1,168 @@
...
+            [
+                self.compile_one_and_cache(
+                    texts[texts_end_index[i] : texts_end_index[i + 1]]
+                )
+                for i in range(texts_lengths.shape[0] - 1)
+            ]
+        )
+
+        # make sure the gradient is not accumulated
+        decoding_graphs.requires_grad_(False)
```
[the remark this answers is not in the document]
yaguanghu: `intersect with ctc_topo` is already done in `compile_one_and_cache` function.

espnet/nets/pytorch_backend/k2_ctc.py
```diff
@@ -0,0 +1,168 @@
...
+        if self.device != log_probs.device:
+            decoding_graph = decoding_graph.to(log_probs.device)
+
+        target_graph = k2.intersect_dense(decoding_graph, dense_fsa_vec, 100.0)
+        tot_scores = target_graph.get_tot_scores(
+            log_semiring=True, use_double_scores=False
+        )
+        if self.reduction == "none":
+            tot_scores = tot_scores / target_lengths.to(log_probs.device)
+        (tot_score, tot_frames, all_frames) = get_tot_objf_and_num_frames(
```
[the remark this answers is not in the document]
yaguanghu: I think this is only for debugging, but anyway it's not used in current code.

espnet/nets/pytorch_backend/k2_ctc.py
yaguanghu: These two different types are just different loss normalization method. There is no logic problem.

espnet/nets/pytorch_backend/k2_ctc.py
```diff
@@ -0,0 +1,168 @@
...
+        return decoding_graphs
+
+    @lru_cache(maxsize=100000)
+    def compile_one_and_cache(self, text: torch.Tensor) -> k2.Fsa:
+        label_graph = k2.linear_fsa(text.tolist(), self.device)
+        label_graph = k2.add_epsilon_self_loops(label_graph)
+        decoding_graph = k2.intersect(
+            self.ctc_topo_inv, label_graph, treat_epsilons_specially=False
+        )
+        decoding_graph = k2.connect(decoding_graph.invert_())
```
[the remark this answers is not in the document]
yaguanghu: I tested it on the CUDA device, it seemed woking? But this operation is not necessary, I'll remove it.

espnet/nets/pytorch_backend/k2_ctc.py
```diff
@@ -0,0 +1,211 @@
...
+            Tuple or tensor of size (N). It represent lengths of the targets.
+        """
+        log_probs = log_probs.permute(
+            1, 0, 2
+        )  # now log_probs is [N, T, C]  batchSize x seqLength x alphabet_size
+        supervision_segments = torch.stack(
+            (
+                torch.tensor(range(input_lengths.shape[0])),
+                torch.zeros(input_lengths.shape[0]),
+                input_lengths,
```
csukuangfj, line 76: Are elements of `input_lengths` sorted in descending order?

If not, I would recommend you to sort them. Please refer to
https://github.com/k2-fsa/snowfall/blob/bbcf1ec1185179b9b502287c51b46fbc2accb8d0/egs/librispeech/asr/simple_v1/ctc_train.py#L80-L81

```
    indices = torch.argsort(supervision_segments[:, 2], descending=True)
    supervision_segments = supervision_segments[indices]
```

danpovey: If this constraint turns out to be a hassle, we could perhaps introduce into the C++ a recursion, that checks whether they are sorted by length, and if not, sorts them by length, does the intersection and then indexes them to get them back in the correct order.

PR #2942, opened 2021-02-04
RTF calculation

hirofumi0810:
Added RTF calculation, mainly for methods which requires latency constraint such as non-autoregressive models.

### DIFF
--- a/egs/mini_an4/st1/run.sh
+++ b/egs/mini_an4/st1/run.sh
@@ -242,6 +242,11 @@ if [ ${stage} -le 5 ] && [ ${stop_stage} -ge 5 ]; then
         decode_dir=decode_${ttask}_$(basename ${decode_config%.*})
         feat_trans_dir=${dumpdir}/${ttask}/delta${do_delta}
 
+        # reset log for RTF calculation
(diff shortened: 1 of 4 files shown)


sw005320:
LGTM!
Could you show me how the output log looks like?

hirofumi0810:
@sw005320 

> Could you show me how the output log looks like?

The output is like this:
```
Total audio duration: 10270.870 [sec]
Total decoding time: 27688.130 [sec]
RTF: 2.696
Latency: 4356.219 [ms/sentence]
```

Currently, ST models + Transformer ASR/MT models are supported.

sw005320:
Looks good!
How about calling it from https://github.com/espnet/espnet/blob/master/utils/score_sclite.sh or https://github.com/espnet/espnet/blob/master/utils/score_sclite.sh and make them as default?
Of course, we can first merge this PR, and later If you have time, you can also consider it.

hirofumi0810:
@sw005320 I tried to adopt such a design yesterday, but it makes the scoring script complicated.
Also, we need to reset log files before decoding for a case where `nj` is changed. 
So it is difficult to embed it to the scoring script.

sw005320:
I see.
We can then keep the current style.

PR #2950, opened 2021-02-08
Fix local/data_prep.sh for egs2/mls/asr1 recipe

sharockys:
Missing `--target` option for `local/data_prep.py`, the script will have no output because of it. 
Fixed data directories for `utils/fix_data_dir.sh`, changed to output of `data_prep.py`.

### DIFF
--- a/egs2/mls/asr1/local/data.sh
+++ b/egs2/mls/asr1/local/data.sh
@@ -71,10 +71,11 @@ fi
 if [ ${stage} -le 1 ] && [ ${stop_stage} -ge 1 ]; then
     log "stage 1: Preparing Data for MLS"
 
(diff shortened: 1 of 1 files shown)


ftshijt:
Sorry that I did not fully understand your revision. The `--target` is default set to `data` and it will result in `data/mls_${lang}_${subdir}`.

sharockys:
> Sorry that I did not fully understand your revision. The `--target` is default set to `data` and it will result in `data/mls_${lang}_${subdir}`.

Hello, thank you for checking this. 
Sorry I misunderstood the design.
I was trying to load the downloaded data with ${MLS} path and put the output at the same place. 
I should have used symbolic link to data instead. 
Thank you for the clarification. 
I closed it.

PR #2951, opened 2021-02-09
Auxiliary task

b-flo:
This PR adds auxiliary tasks (https://arxiv.org/pdf/2011.03109.pdf) and multi-task learning (https://arxiv.org/pdf/2011.00771.pdf) for transducer models. 

To do:
  - [x] Auxiliary RNN-T criterion
  - [x] Auxiliary symmetric KL divergence criterion
  - [x]  Multi-task learning w/ CTC
  - [x]  ~~Multi-task learning w/ LM~~ Multi-task learning w/ CE
  - [x] Units tests
  - [ ] Validation draft and optimisation
  - [ ] Results

Notes:
 - I won't cover CD graphemic state prediction (https://arxiv.org/pdf/2011.03109.pdf) but I want to revisit a similar approach with k2 in a later PR
 - Instead of using the two-step training described in https://arxiv.org/pdf/2011.00771.pdf for multi-task with LM classifier, I'm using a one-step training with multi-task CE. In my opinion the two-step method is too labyrinthine but it may be due to my implementation, I'll contact the authors.
 - New `RNNEncoder` was added with support to access intermediate layer outputs. I didn't want to overload the original class so I made a new one but I can unify them if you don't want duplication.
 - Moved arguments from `e2e_asr_transducer.py` to `transducer/arguments.py` as first file was too overloaded.
 - All modules related to auxiliary task and multi-task are not loaded during inference (so it won't impact decoding speed).


b-flo:
FYI, I did some early experiments with auxiliary CTC and CE on Voxforge (I didn't try to tune each contribution thought) :

| Model | dev set | test set |
| - | - | - |
| RNN CTC-Att | 12.9 / N.A | 12.6 / N.A |
| RNN-T | 13.1 / 44.3 | 12.4 / 43.4 |
| + aux CTC |  12.8 / 44.6 | 12.1 / 43.1 |
| + aux CE | 12.9 / 43.9 | 12.0 / 42.1 |

I also tested a Conformer-RNN-T architecture with a deeper decoder which previously resulted in degraded performances:

| Model | dev set | test set |
| - | - | - |
| Conformer CTC-Att | 8.7 / 32.6 | 8.2 / 30.5 |
| Conformer-RNN-T + aux CTC + aux CE | 8.8 / 31.9 | 7.9 / 30.1 |

b-flo:
It can be reviewed now!

sw005320:
I want to merge this PR as soon as possible and I just check a quick/shallow reviewer.
The PR looks good, but I'm concerning this

> New RNNEncoder was added with support to access intermediate layer outputs. I didn't want to overload the original class so I made a new one but I can unify them if you don't want duplication.

We need some clear descriptions of why and where it is different from https://github.com/espnet/espnet/blob/master/espnet/nets/pytorch_backend/rnn/encoders.py.

How about moving this one under https://github.com/espnet/espnet/blob/master/espnet/nets/pytorch_backend/rnn/ ?
We may also rename it from rnn_encoder.py to rnn_state.py (or other names that clearly show the access of intermediate layer outputs).
Then, if we clearly describe the difference between them (in the header of the files), that is less confused.

b-flo:
I modified the main header to explain differences with default RNN classes.

> How about moving this one under https://github.com/espnet/espnet/blob/master/espnet/nets/pytorch_backend/rnn/ ?

Hum, I'm not sure as it's only intended to be used with transducer models and it may be clearer to have all related classes in the same place. However it's only my opinion, I can rename file and move it to `rnn/`!

sw005320:
Probably you're right, but I was just confused.
Very similar files are located in very different places with ununified names.
Also, I thought the state access can be also useful for the encoder-decoder architecture in the future.
But it's up to you. I don't want to take additional time due to these ambiguous discussions.

b-flo:
OK, let's keep it like that for now. If it's useful for other models I'll do modifications in the future according to your review!

sw005320:
Auto-merge did not work, so I merged it manually.
Anyway, thanks!

PR #2963, opened 2021-02-12
Must-C v2 recipe

hirofumi0810:
Added Must-C ASR/ST/MT recipes. The difference from the original one is the style of yaml files.


ftshijt:
Many thanks! The new recipe looks good. Given that most of the codes are the same, could you please tell me why you would like to add a v2 instead of directly adding a new option in the original one?

hirofumi0810:
> Many thanks! The new recipe looks good. Given that most of the codes are the same, could you please tell me why you would like to add a v2 instead of directly adding a new option in the original one?



> Many thanks! The new recipe looks good. Given that most of the codes are the same, could you please tell me why you would like to add a v2 instead of directly adding a new option in the original one?

The v2 is a superset of the original one. Also, audio files are different (quality was improved).
For simplification, I decided to make a different directory after the discussion with @sw005320.

ftshijt:
> > Many thanks! The new recipe looks good. Given that most of the codes are the same, could you please tell me why you would like to add a v2 instead of directly adding a new option in the original one?
> 
> The v2 is a superset of the original one. Also, audio files are different (quality was improved).
> For simplification, I decided to make a different directory after the discussion with @sw005320.

Thanks for your clarification! Then, my suggestion would be to add some explanation in https://github.com/espnet/espnet/blob/master/egs/README.md

hirofumi0810:
@ftshijt Got it. I added some descriptions.

hirofumi0810:
@ftshijt This PR is ready for merging.

hirofumi0810:
@ftshijt Wait, I have to fix downloading.

ftshijt:
Many thanks! Looks great to me. Merged.

PR #2966, opened 2021-02-15
Update espnet2 librispeech result

kamo-naoyuki:
Slightly worse than espnet1.


kamo-naoyuki:
I restricted the pytorch version to >=1.6.0 for the configuration validation in ci test because `use_amp=true` can't be used now.

PR #2969, opened 2021-02-15
Auto Resampling to 16khz for pretrained models

siddalmia:
This is a simple change which will resample the audio to the default settings of ESPnet models.


sw005320:
Thanks, @siddalmia!!!

PR #2971, opened 2021-02-15
Move import BeamSearchTransducer to a conditional

siddalmia:
In order to avoid some compatibility issues of espnet dockers - https://github.com/espnet/espnet/issues/2970

PR #2973, opened 2021-02-15
Fix docker error caused by BeamSearchTransducer

### LINKED ISSUE #2970: Old Docker Version failing recog_wav.sh due to a recent commit to support RNN-T decoding
The BeamSearchTransducer function in https://github.com/espnet/espnet/blob/master/espnet/asr/pytorch_backend/asr.py#L38

causes old docker versions to fail which do not have espnet-2 installed. 

@b-flo for the time being would it be okay to move this to - https://github.com/espnet/espnet/blob/master/espnet/asr/pytorch_backend/asr.py#L991

It seems like it's being used only for the RNN-T decoding.

Thanks
Sid

b-flo:
Follow-up of #2971. Closes #2970.

It can be merged after @siddalmia gives the go-ahead!

### DIFF
--- a/espnet/nets/beam_search_transducer.py
+++ b/espnet/nets/beam_search_transducer.py
@@ -1,11 +1,6 @@
 """Search algorithms for transducer models."""
 
-from dataclasses import dataclass
-from typing import Any
-from typing import Dict
 from typing import List
-from typing import Optional
-from typing import Tuple
 from typing import Union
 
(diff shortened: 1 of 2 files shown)


siddalmia:
Thanks I checked with the docker and it works  :)

sw005320:
Thanks, @b-flo!

PR #2977, opened 2021-02-16
swbd run.sh is including dev data in the training set

brianyan918:
This is a bug fix to make sure that the swbd train_set, train_nodup_trim does not include the dev set. The bug was introduced when speed perturbation was added in recently. 

@sw005320 can you check this fix?


sw005320 (approved):
Thanks! for the bug fix!

PR #2981, opened 2021-02-18
bug fixed for streaming transformer ASR

eml914:
I found an error in decoding using entire encoded features instead of blocks.
Also, I modified configuration part for the decoder to be more versatile.
Added test for the configuration accordingly.


sw005320:
Thanks, @eml914!

PR #2988, opened 2021-02-18
builtin ctc modification

siddalmia:
I modified the existing CTC builtin code as the CuDNN CTC code (set with option "cudnnctc"). 

Added the pytorch CUDA CTC as the "builtin" option, which has the option of "zero_infinity". This takes care of "poorly aligned" utterances. The "builtin" option also accepts input larger than 256 frames. (https://github.com/pytorch/pytorch/pull/21244)

### DIFF
--- a/doc/espnet2_distributed.md
+++ b/doc/espnet2_distributed.md
@@ -31,7 +31,7 @@ If you meet some errors with distributed mode, please try single gpu mode or mul
 We supports sharded training provided by [fairscale](https://github.com/facebookresearch/fairscale)
 
 ```bash
-% python -m espnet2.bin.asr_train --ngpu 4 --multiprocessing_distributed true --sharded_ddp true
+% python -m espnet2.bin.asr_train --ngpu 4 --multiprocessing_distributed true --ddp_sharded true
 ```
 
 Note that the other features of fairscale are not supported now.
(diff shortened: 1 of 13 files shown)


sw005320:
Could you fix the formatting issue in https://github.com/espnet/espnet/pull/2988/checks?check_run_id=1930293715#step:8:9 ?

I think it's better to have a unit test. Actually, gtn also does not have a unit test.
So, it would be great if @siddalmia and/or @brianyan918 add a unit test in the later PR.

siddalmia:
@sw005320 could you check now? Apologies about that.

siddalmia:
@sw005320 @brianyan918 apologies for the many notifications that you received.

sw005320:
No problem, but it seems that the test detects some errors in https://github.com/espnet/espnet/pull/2988/checks?check_run_id=1938011807#step:8:418
Maybe some PyTorch versions do not support `zero_infinity=True`?
So, we need to make this argument available depending on the specific PyTorch version.
You can find a couple of sample codes in https://github.com/espnet/espnet/search?q=LooseVersion&type=code

siddalmia:
@sw005320 it should work now. I apologize about the messy commits. I think I used my master branch to make this PR, which caused all this issue.

siddalmia:
Let me close this PR and reopen with the fix.

sw005320:
Sure. Maybe this is easier.

PR #2994, opened 2021-02-19
Fix sharded training for multiple nodes

kamo-naoyuki:
- I fixed to invoke `consolidate_state_dict()` for multiple nodes case
- I reverted --ddp_shared to --sharded_ddp. I changed my mind

Multiple nodes training can't work now because fairscale has a bug.

### DIFF
--- a/doc/espnet2_distributed.md
+++ b/doc/espnet2_distributed.md
@@ -31,7 +31,7 @@ If you meet some errors with distributed mode, please try single gpu mode or mul
 We supports sharded training provided by [fairscale](https://github.com/facebookresearch/fairscale)
 
 ```bash
-% python -m espnet2.bin.asr_train --ngpu 4 --multiprocessing_distributed true --ddp_sharded true
+% python -m espnet2.bin.asr_train --ngpu 4 --multiprocessing_distributed true --sharded_ddp true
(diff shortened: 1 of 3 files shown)


kamo-naoyuki:
Hmm, I did benchmark test of sharded, but I couldn't find any improvement...

kamo-naoyuki:
Okay, I confirmed some improvement of sharded. We should use this feature actively because there are no disadvantage. I'll change sharded to default if  fairscale becomes more stable.

- Environment
   - v100x8 on single node
   - Disable cudnn because cudnn8.0 doesn't work with amp + v100 now?
   -  Using librispeech: https://github.com/espnet/espnet/blob/master/egs2/librispeech/asr1/conf/tuning/train_asr_conformer6_n_fft512_hop_length256.yaml

- Speed comparison with same  batch_bins=210000000

|amp|sharded|Train time per iteration|
|---|---|---|
|no|no|1.46|
|no|yes|1.285|
|yes|no|1.082|
|yes|yes|0.878|

- Maximum batch_bins with same environment

|amp|sharded|Batch bins|
|---|---|---|
|no|no|210000000|
|no|yes|230000000|
|yes|no|370000000|

PR #2996, opened 2021-02-19
Print total params and trainable params.

siddalmia:
Adding a small snippet of code that prints the number of model params and the trainable params along with the percentage trainable (borrowed from https://github.com/pytorch/fairseq/pull/469) in both espnet and espnet-2.

For espnet-2, I just added total params and the percentage to the model_summary.py as suggested by @simpleoier. 

I have tested it for ASR task in ESPnet and it works and thanks @simpleoier for testing it on ESPnet-2!

### DIFF
--- a/espnet/asr/pytorch_backend/asr.py
+++ b/espnet/asr/pytorch_backend/asr.py
@@ -499,6 +499,16 @@ def train(args):
     else:
         model_params = model.parameters()
 
+    logging.warning(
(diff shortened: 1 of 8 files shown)


simpleoier:
I tested it on ESPnet2. It works fine!

sw005320:
This looks very useful.
Thanks, @siddalmia!

PR #2997, opened 2021-02-20
bug fix for Wav2Vec encoder

simpleoier:
1. Bug fix: using deepcopy() to copy the parameters of pretrained wav2vec encoder
2. Adding new freeze_finetune_updates parameter not to update the wav2vec encoder for a few iterations

### DIFF
--- a/espnet2/asr/encoder/wav2vec2_encoder.py
+++ b/espnet2/asr/encoder/wav2vec2_encoder.py
@@ -3,6 +3,7 @@
 
 """Encoder definition."""
 import contextlib
+import copy
 import logging
 import os
 from typing import Optional
@@ -40,6 +41,7 @@ class FairSeqWav2Vec2Encoder(AbsEncoder):
         normalize_before: bool = True,
         freeze_w2v: bool = True,
         finetune_last_n_layers: int = 0,
+        freeze_finetune_updates: int = 0,
     ):
         assert check_argument_types()
         super().__init__()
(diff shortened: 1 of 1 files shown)


sw005320:
1. Can you add a test?

Also, an example with some documentations is really helpful, e.g., 
1. General descriptions in https://github.com/espnet/espnet#asr-automatic-speech-recognition
2. Detailed explanations in https://espnet.github.io/espnet/tutorial.html#how-to-use-finetuning

W2V2 shows great interests and if you add them, the function would be widely used.

simpleoier:
Hi @sw005320 , thank you for your comment. I will add something to the documentation. However, I made a test script for wav2vec encoder. But then I realized that `FairSeq` is not installed by default. So this may fail in the test.
```python
# ./text/espnet2/asr/encoder/test_wav2vec2_encoder.py
@pytest.mark.parametrize(
    "w2v_url", ["https://dl.fbaipublicfiles.com/fairseq/wav2vec/wav2vec_small_960h.pt"]
)
@pytest.mark.parametrize("freeze_w2v", [True, False])
@pytest.mark.parametrize("finetune_last_n_layers", [0, 1])
def test_Encoder_forward_backward(w2v_url, freeze_w2v, finetune_last_n_layers):
    encoder = FairSeqWav2Vec2Encoder(
        w2v_url=w2v_url,
        freeze_w2v=freeze_w2v,
        finetune_last_n_layers=finetune_last_n_layers,
    )
    x = torch.randn(2, 16000, requires_grad=True)  # 1s speech with sampling rate 16KHz
    x_lens = torch.LongTensor([16000, 15000])
    y, _, _ = encoder(x, x_lens)
    y.sum().backward()


def test_Encoder_output_size():
    encoder = FairSeqWav2Vec2Encoder(
        w2v_url="https://dl.fbaipublicfiles.com/fairseq/wav2vec/wav2vec_small_960h.pt",
        output_size=256,
    )
    assert encoder.output_size() == 256


def test_Encoder_invalid_type():
    with pytest.raises(ValueError):
        FairSeqWav2Vec2Encoder(w2v_url="")
```

sw005320:
Year, this would be an annoying issue and happens to the other tools that we did not provide the default installation.
We can leave it for now, but please work on the documentation in the follow-up PRs.

PR #2999, opened 2021-02-20
Install warp-ctc using pip>=21.0

ysk24ok:
fixes https://github.com/espnet/warp-ctc/issues/35

Now that warp-ctc wheels are served in its [release page](https://github.com/espnet/warp-ctc/releases/tag/v0.2.2) instead of pypi.org, PEP440 restriction can be avoided.


kan-bayashi:
Thanks!

PR #3002, opened 2021-02-20
Adding Attention Weights to ASR Beam Search Decoding

roshansh-cmu:
1. Adding support for providing a fixed maxlen parameter that overrides maxlen = maxlenratio * hlen if provided. This is useful for certain ASR tasks and SLU tasks with fixed length of outputs rather than setting maxlenratio and minlenratio.

2. Modifications to beam_search.py and batch_beam_search.py to support handling of attention as an additional parameter in the definition of Hypothesis object.

This PR was tested with both ESPnet1 and ESPnet2 transformer decoding using the v2 API.
This should still allow beam search to work for CTC only decoding- but this has not been tested yet.

Welcome any feedback 

Thanks

### DIFF
--- a/espnet/asr/pytorch_backend/recog.py
+++ b/espnet/asr/pytorch_backend/recog.py
@@ -2,7 +2,7 @@
 
 import json
 import logging
-
+import os
 import torch
 
 from espnet.asr.asr_utils import add_results_to_json
@@ -17,6 +17,8 @@ from espnet.nets.scorer_interface import BatchScorerInterface
 from espnet.nets.scorers.length_bonus import LengthBonus
 from espnet.utils.deterministic_utils import set_deterministic_pytorch
 from espnet.utils.io_utils import LoadInputsAndTargets
+
+from espnet2.bin.asr_inference import plot_attentions
 
 
 def recog_v2(args):
@@ -128,6 +130,11 @@ def recog_v2(args):
     with open(args.recog_json, "rb") as f:
         js = json.load(f)["utts"]
     new_js = {}
+
+    ## Attention Plot Directory
+    att_dir = os.path.join(os.path.dirname(args.result_label), "att_wts")
+    if not os.path.isdir(att_dir):
+        os.mkdir(att_dir)
     with torch.no_grad():
         for idx, name in enumerate(js.keys(), 1):
             logging.info("(%d/%d) decoding " + name, idx, len(js.keys()))
@@ -143,6 +150,19 @@ def recog_v2(args):
             new_js[name] = add_results_to_json(
                 js[name], nbest_hyps, train_args.char_list
             )
+            ## Save Attention Plots
+            attentions = nbest_hyps[0]["att_wt"]
+            print([v.shape for k, v in attentions.items()])
+            print([v.shape for k, v in attentions.items()])
+
+            tokens = new_js[name]["output"][0]["rec_token"].split(" ")
+            if attentions != dict():
+                plot_attentions(
+                    name,
+                    attentions,
+                    att_dir,
+                    ytokens=tokens,
+                )
 
     with open(args.result_label, "wb") as f:
         f.write(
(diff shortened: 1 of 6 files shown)


sw005320:
Thanks @roshansh-cmu! 
This is really useful and we'll have a lot of applications based on this.

- Could you fix the conflict?
- Could you add a unit test by just referring to some other beam search tests?
- Was the decoding speed changed? I guess not, but I just want to make sure.

roshansh-cmu:
I will fix the conflict. 
I will also try writing the unit test
Decoding should take around the same time except that plotting is slow, but I will work on getting exact numbers.
I am waiting on any code review feedback so I can work on all of it together. Thanks.

sw005320:
@ShigekiKarita, could you make time to check it?
I think this looks like straightforward extensions within your beam search functions.

espnet/asr/pytorch_backend/recog.py
```diff
@@ -18,6 +18,8 @@
 from espnet.utils.deterministic_utils import set_deterministic_pytorch
 from espnet.utils.io_utils import LoadInputsAndTargets
 
+from espnet2.bin.asr_inference import plot_attentions
```
b-flo, line 21: Is the reference to espnet2 not going to be an issue for the docker?

roshansh-cmu: It might be- I plan to move the plotting function elsewhere- maybe espnet/nets/asr_utils.py ?

b-flo: I'm not sure, @Fhrozen What do you think?

Fhrozen: Line 21 will be a problem for docker (egs). I would recommend to avoid calling from espnet2 -> espnet1.
in the case of moving to espnet/nets/asr_utils, then it should not be a problem.

roshansh-cmu: Thanks- will do

roshansh-cmu:
Maybe it will help if I finish all the code before the code review. Working on it now, and will make another PR by Monday.

sw005320:
I think the change is very reasonable.
So, if you reflect on our comments, I can just merge this PR.

espnet2/asr/decoder/transformer_decoder.py
```diff
@@ -184,11 +184,14 @@ def score(self, ys, state, x):
...
-        return logp.squeeze(0), state
+        ## Get Attention Weights
+        attentions = self.get_attention_weights()
+
+        return logp.squeeze(0), state, attentions
 
     def batch_score(
         self, ys: torch.Tensor, states: List[Any], xs: torch.Tensor
-    ) -> Tuple[torch.Tensor, List[Any]]:
+    ) -> Tuple[torch.Tensor, List[Any], dict]:
```
ShigekiKarita, line 194: Maybe we should update ScorerInterface.batch_score like this return type to support attention weights decoding in all the models. OR if you have no plan to do that, it is OK to leave this small inconsistency.

espnet/nets/pytorch_backend/transformer/decoder.py
```diff
@@ -367,4 +369,26 @@ def batch_score(
 
         # transpose state of [layer, batch] into [batch, layer]
         state_list = [[states[i][b] for i in range(n_layers)] for b in range(n_batch)]
-        return logp, state_list
+
+        ## Get Attention Weights
+        attentions = self.get_attention_weights()
+        return logp, state_list, attentions
```
ShigekiKarita, line 375: Can you update the return value type of this function?

ShigekiKarita (commented):
Sorry for late review. LGTM overall except for minor typing problems.

espnet2/asr/decoder/transformer_decoder.py
ShigekiKarita: I think the best approach is a new NamedTuple like beam_search.Hypothesis:
```python
# in espnet.nets.scorer_interface

class BatchScore(NamedTuple):
    score: torch.Tensor
    states: List[Any] = []
    attention: dict = {}

class BatchScorerInterface:
    def batch_score(
        self, ys: torch.Tensor, states: List[Any], xs: torch.Tensor
    ) -> BatchScore:
```

PR #3004, opened 2021-02-21
Fix arguments in dynamic and lightweight conv

hirofumi0810:
kernel size definition is done outside the sub-module.


sw005320:
LGTM!
Thanks a lot!
If it is ready, you can just merge it.

PR #3014, opened 2021-02-25
Supporting multiple optimizers in the default trainer

kamo-naoyuki:
- I removed trainer.num_optimizer. Now you need to only change Task.num_optimizer and task.build_optimizer() if you need multiple optimizers.

```python
class Task(AbsTask):
    num_optimizer: int = 2

    @classmethod
    def build_optimizers(cls, args, model):
        ...
        return optimizers 
```

- I added a default behaviour for multiple optimizers case. You still need to modify trainer class if some tweak  is required, but I think this is enough useful.

```python
# default behaviour
for optimizer in optimizers:
     optimizer.step()
```


espnet2/train/trainer.py
```diff
@@ -391,7 +381,26 @@ def train_one_epoch(
...
+                        if "optim_idx" in retval:
+                            optim_idx = retval["optim_idx"]
+                        else:
+                            optim_idx = None
+
+                    #   b. tuple or list type
+                    else:
+                        loss, stats, weight = retval
+                        optim_idx = None
+
```
kamo-naoyuki, line 403: I also added a system to select a optimizer to be used from the model for each step:

```python
class YourModel:
    def forward(self, ...):
        ...
        # The 0th optimizer will be called
        return {"loss": loss, "stats": stats, "weight": weight, "optim_idx": 0}
```

PR #3024, opened 2021-02-27
Refactor layer selection in Transformer

hirofumi0810:
Follow Conformer's layer definition.


sw005320:
Thanks!
Does it keep the compatibility (can we read an old model?)?
If so, we can merge this.

hirofumi0810:
@sw005320 This change does not affect any compatibility. Just for readability.

sw005320:
OK, thanks!

PR #3026, opened 2021-02-28
Add documentation for CTC decoding

YosukeHiguchi:
I added some descriptions for CTC decoding in doc/tutorial.md. I also made a minor change in README.md.


sw005320:
Can you clarify more about how to specify the best path decoding with the command line example? (I guess you may need to specify the beam size and API?)

YosukeHiguchi:
I noticed that the part which I'm currently editing in the tutorial document is little out-dated. Arguments such as `mtlalpha` or `ctc_weight` should be specified in config files instead of passing those to run.sh. I'll try to update the whole part and include specific example for CTC decoding.

sw005320:
> I noticed that the part which I'm currently editing in the tutorial document is little out-dated. Arguments such as `mtlalpha` or `ctc_weight` should be specified in config files instead of passing those to run.sh. I'll try to update the whole part and include specific example for CTC decoding.

This is really helpful.
Thanks a lot!

sw005320:
Thanks!

PR #3036, opened 2021-03-02
modification of freezing option for Wav2Vec encoder, add documents

simpleoier:
1. Remove the `extract_feature.sh` which overwrote the fairseq encoder forward function.
2. Make freezing parameter of wav2vec encoder consistent with ESPnet2 style. But the drawback is too long list of frozen parameters in configuration file.
3. Add documents for wav2vec2 model and update the tutorial for freezing option.


espnet2/asr/encoder/wav2vec2_encoder.py
```diff
@@ -60,15 +55,9 @@ def __init__(
 
         self.w2v_model_path = download_w2v(w2v_url, w2v_dir_path)
```
kamo-naoyuki, line 56: Hi, I found a small problem.
If ngpu >=2, 2 or more training processes run in parallel, so download_w2v() also run in parallel. 

This is not fatal because we can avoid it if once successfully downloading is finished, so for now, please keep in mind.

One of the solutions is using a file lock, but I'm not sure.

simpleoier: You are right. I noticed this at the beginning. Another solution is to have a small script to download the model before Step 9: collect stats.

kamo-naoyuki: Oh, I forgot `Step 9: collect stats`.  Now this downloading is done at step9? This problem is likely to happen, I think it's better to solve this issue now.

File lock solution can be a simple recipe, but stats jobs are stopped when downloading without calculation, and these jobs occupy the queue of scheduler wastefully for this time. File-lock also depends on OS functions, but basically we should avoid OS-depending functions.

Small script solution needs an extra step in asr.sh. I feel reading config file is a little bit tricky way. 

I think file lock solution is a little better. This model file is not so large and the waiting time is acceptable, maybe. 

Have you used filelock in the pask? I think https://pypi.org/project/filelock/ is good.
(It's also okay if you ask me to tackle this problem)

simpleoier: Thanks for the suggestion.
Now it prevents downloading parallelly. How does it look now?
BTW, it seems that python filelock library [does not delete the lock automatically](https://stackoverflow.com/questions/58098634/why-does-the-python-filelock-library-delete-lockfiles-on-windows-but-not-unix) on UNIX like systems.

kamo-naoyuki: Great! Thanks! Maybe, the following is not needed.

```
        if os.path.exists(model_path):
            return model_path
```

> BTW, it seems that python filelock library does not delete the lock automatically on UNIX like systems.

It's okay. I think this is inevitable.

PR #3048, opened 2021-03-05
Fix+update RNN encoder

b-flo:
Only apply to RNN class (RNNP is not impacted):
 - Fix case when eprojs != eunits
 - Add projection for output layer
 - Add projection for auxiliary output layers


sw005320:
Thanks, @b-flo for your quick fix!

PR #3055, opened 2021-03-07
[docs] [minor] fix a broken link

IlnarSelimcan:
Also, support for ubuntu 16.04 seems to have been dropped, and justly so.


kamo-naoyuki:
Thanks!

PR #3071, opened 2021-03-11
Update IWSLT18 MT recipe

hirofumi0810:
- Simplified data preparation pipeline
- Used original segmentation instead of downloading from Google Drive (both files are exactly the same)
- Supported Transformer configuration
- Fixed a bug when using BPE
- Added `tst2018` and `tst2019` test sets (references are available now!)


ftshijt:
Looks great! Just interested: will gpu decoding be faster than cpu for ST? I found it is slower than cpu for my environment on asr.

hirofumi0810:
@ftshijt GPU decoding can significantly accelerate inference speed in the ST task (< 300ms per utterance). 
FYI: https://arxiv.org/abs/2010.13047

PR #3081, opened 2021-03-14
Add joint-memory-reduction option

b-flo:
This PR add `joint-memory-reduction` which apply "Efficient encoder and prediction output combination" (https://www.microsoft.com/en-us/research/uploads/prod/2019/10/RNNT.pdf) during training. It works as intended but I observed a slight performance degradation I need to investigate before merging.

Note that some operations will be removed in the future when we add our own torch lib for RNNT computation, I added it now because memory reduction may be strongly needed with large model using auxiliary RNN-T criterion.


tbright17:
Hi there, not sure why you use for loop over batch for memory reduction? Is this tradeoff between speed and memory?

Thanks

b-flo:
Hi,

> Hi there, not sure why you use for loop over batch for memory reduction? Is this tradeoff between speed and memory?

It is, it shouldn't impact training speed too much though. Please check section 3.1 from the paper I linked.
Btw, note that it's a temporary version as I'm currently reworking most code related to training (loss computation and training scheme) and auxiliary tasks.

tbright17:
Thanks for the update. Are you going to implement section 3.2? It could
involve cuda-level programming.

b-flo:
It is on my to do list but I can't give an estimated time yet.

maxwellzh:
Hi @b-flo , I'm also working on this impl.
I have implemented the section 3.1 with CUDA binding and make better perf.
Do you have any idea about the section 3.2? I tried to implement it, although memory is reduced, the computation is slower.

refers: https://github.com/1ytic/warp-rnnt/pull/26

b-flo:
Hi,

> I have implemented the section 3.1 with CUDA binding and make better perf.

Nice! Incorporating the approach into our `warp-transducer` fork was the next step but it's further down my TODO list right now. 

> Do you have any idea about the section 3.2? I tried to implement it, although memory is reduced, the computation is slower.

I'm not sure about what you're asking? From what I understand, computation would be slightly slower with same settings but the memory reduction allows us to use larger batches, resulting in a faster training speed.

maxwellzh:
> I'm not sure about what you're asking? From what I understand, computation would be slightly slower with same settings but the memory reduction allows us to use larger batches, resulting in a faster training speed.

As you may know, currently we use `gather=True` in rnnt loss calculation. So the memory footprint of loss cal is `(N, T, U, 2)` instead of `(N, T, U, V)`. In my benchmarking, you can see the memory reduction isn't very significant.

Theoretically, even though with the same batch size, fusion with softmax should be faster than the native impl. But my impl of softmax fusion seems poorer :(
So I wonder if there is any efficient way to make the impl.

benchmark of warp-rnnt-gather-compact and warp-rnnt-fusion. tested on RTX 3090 with 24GB memory
```
gather + compact
T=150   U=20    V=5000  N=1     time=2.93       memory=243                                                                                                                                                             
T=150   U=20    V=5000  N=16    time=13.70      memory=2768                                                                                                                                                            
T=150   U=20    V=5000  N=32    time=23.09      memory=4993                                                                                                                                                            
T=150   U=20    V=5000  N=64    time=69.36      memory=9671                                                                                                                                                            
T=150   U=20    V=5000  N=128   time=198.37     memory=19286                                                                                                                                                           

T=400   U=100   V=1024  N=1     time=6.00       memory=?                                                                                                                                                          
T=400   U=100   V=1024  N=16    time=73.29      memory=7165                                                                                                                                                            
T=400   U=100   V=1024  N=32    time=114.88     memory=13580                                                                                                                                                           
T=400   U=100   V=1024  N=64    OOM

fusion
T=150   U=20    V=5000  N=1     time=2.50       memory=182
T=150   U=20    V=5000  N=16    time=15.89      memory=2244
T=150   U=20    V=5000  N=32    time=30.34      memory=4284
T=150   U=20    V=5000  N=64    time=59.26      memory=8464
T=150   U=20    V=5000  N=128   time=110.91     memory=16469

T=400   U=100   V=1024  N=1     time=6.63       memory=?
T=400   U=100   V=1024  N=16    time=47.49      memory=6061
T=400   U=100   V=1024  N=32    time=164.36     memory=11885
T=400   U=100   V=1024  N=64    time=329.29     memory=23412
T=400   U=100   V=1024  N=128   OOM
```

b-flo:
Thank you for the benchmark! It seems speed/memory consumption trade-off is all over the place between the implementation :smile:

I'll take a look at your implementation when I have time. Unfortunately, I need to leave for two days and I have some other works to finish before diving into your PR. I may start mid-end next week I think

Issue #3018, opened 2021-02-26
Audio-to-text alignment with trained Espnet2 asr model


qdenisq:
Hi, can anyone tell how one does audio-to-text alignment using Espnet2? I can see there is `asr_align.py`  in Espnet and was curious if Espnet2 provides a similar interface. Thank you

sw005320:
Very good point!
Yes, right now, we only support it in espnet1.
@lumaku, are you interested in extending your CTC alignment to espnet2?
As you know, espnet2 is more portable and has better API, and your CTC alignment tool becomes more powerful in these regards.

lumaku:
Yes, I can do that.
A small pre-trained model for English speech (librispeech/wsj), similar to wsj.transformer_small.v1 in espnet1, would be very helpful for testing. Does espnet2 currently have a similar model?

kamo-naoyuki:
All pretrained models can be found in https://github.com/espnet/espnet_model_zoo

qdenisq:
@kamo-naoyuki @sw005320 I have one more question aside from the actual CTC alignment script and was hoping you could advise me on this. I have a Conformer trained on LIbrispeech with BPE vocabulary (size=5000) according to the recipe from [https://zenodo.org/record/4276519](https://zenodo.org/record/4276519). After taking argmax on ctc output I get the following:

**Argmax result**
```
[   0    0    0    7    0    0    0    0    0    0    0    0   83    0
    0    0    0    0    0    0    0    0  729    0    0    0    0    0
    0    0    0    0 3415    0    0    0    0    0    0    0    0    0
    0    0    0    0    0    0    0]
```
, where 
```
{
 0: '<blank>',
 7: '_A',
 83: '▁LITTLE',
 729:  '▁GREEN',
 3415: '▁FROG'
}
```
Thus,

**Tokenized transcript**
`['▁A', '▁LITTLE', '▁GREEN', '▁FROG']`

I checked the indexes of the non-zero tokens and timestamps in the audio recording.  It seems these indexes match with the timestamps of the beginnings of the tokens in the recording. It means that one could easily retrieve the starting timestamps, but it is absolutely unclear to me how to get the endings of tokens.

Do I understand it correctly that it is not really feasible to retrieve the token lengths from ctc output since only the starting frame is encoded with the token id and the rest is filled with `<blank>` until the next token is detected?

lumaku:
[...]

Issue #2592, opened 2020-10-22
Can CTC align with token


s920128:
If I use spm_train and spm_encode, i.e. sentencepiece, in ESPNET.
The dictionary is subword units based or tokens.
Can I use ctc_segmentation directly (like tedlium2 example)?
It seems to be possibly raise "Audio is shorter than text!" error, where number of tokens is less than frame but  len(ground_truth) > lpz.shape[0]. (because len(ground_truth) is length of string)

lumaku:
CTC segmentation also works with tokens, as long as the input string can be partitioned into these tokens with regular string comparisons.

Your issue sounds like there are long tokens in the dictionary. The error "Audio is shorter than text" appears when the `len(ground_truth)` in `prepare_text()` is longer than the number of frames.

I suspect that that there are longer tokens in your dictionary, such as "\<space\>" instead of just " ", which increase the length of ground truth. Could you show the dictionary (`config.char_list`)?

s920128:
Yes , there are longer tokens in my dictionary.
```
char_list = [ "<blank>",
            "<unk>",
            "'",
            "0",
            "1",
            "10",
            "14",
            "15",
            "2",
            "3",
            "35",
            "4",
            "5",
            "50",
            "6",
            "7",
            "9",
            "<v-noise>",
            "[UNK]",
            "_",
            "a",
            "aa",
            "ab",
            "abandon",
            "abandoned",
            "abdomen",
            "abilities",
            "ability",
            "able",
            "about",
            "above",
            "absent",
            "absolute",
            "absorbed",
            ... ,
            "▁a",
            "▁aa",
            "▁aad",
            "▁ab",
            "▁aba",
            "▁abb",
            "▁abi",
            "▁ability",
            "▁able",
            "▁ables",
            "▁ably",
            "▁abo",
            "▁ac",
             etc.]
``` 
This is the information of the utterance, which raise "Audio is shorter than text!".
INFO: CTC segmentation of 52 chars to 1.84s audio (46 indices).
```
In data_json: 
    shape": [11, 12309],
    "text": "所 以 i am very surprised that i passed everything",
    "token": "所 以 i am very surprise ▁d that i passed everything",
    "tokenid": "10653 9722 2628 213 5925 5396 6785 5558 2628 3986 1911"
```

lumaku:
[...]

PR #3087, opened 2021-03-15
CTC Segmentation for ESPnet 2

lumaku:
This pull request brings CTC segmentation to ESPnet 2. It includes a python module `CTCSegmentation` that can be used in a similar fashion as the `Speech2Text` module of ESPnet2. Using the example file included in ESPnet, to align utterances with the new `CTCSegmentation` module:

```python
# load model
from espnet_model_zoo.downloader import ModelDownloader
d = ModelDownloader(cachedir="./modelcache")
wsjmodel = d.download_and_unpack("kamo-naoyuki/wsj")
# load audio
import soundfile
speech, rate = soundfile.read("./test_utils/ctc_align_test.wav")
# CTC segmentation
from espnet2.bin.asr_align import CTCSegmentation
aligner = CTCSegmentation( **wsjmodel , fs=rate )
text = """
utt1 THE SALE OF THE HOTELS
utt2 IS PART OF HOLIDAY'S STRATEGY
utt3 TO SELL OFF ASSETS
utt4 AND CONCENTRATE ON PROPERTY MANAGEMENT
"""
segments = aligner(speech, text)
print(segments)
# utt1 utt 0.26 1.73 -0.0154 THE SALE OF THE HOTELS
# utt2 utt 1.73 3.19 -0.7674 IS PART OF HOLIDAY'S STRATEGY
# utt3 utt 3.19 4.20 -0.7433 TO SELL OFF ASSETS
# utt4 utt 4.20 6.10 -0.4899 AND CONCENTRATE ON PROPERTY MANAGEMENT
```

This also works with fragments of the text.
```python
aligner.set_config( gratis_blank=True, kaldi_style_text=False )
text = ["SALE OF THE HOTELS", "PROPERTY MANAGEMENT"]
segments = aligner(speech, text)
print(segments)
# utt_0000 utt 0.37 1.72 -2.0651 SALE OF THE HOTELS
# utt_0001 utt 4.70 6.10 -5.0566 PROPERTY MANAGEMENT
```

The interface of the script `asr_align.py` is simple:
```
asr_align.py  --asr_train_config config.yaml --asr_model_file model.pth --audio audio.wav --text text.txt --output segments
```

`README.md` includes a description how to use these.

* *Note about input and output for the script interface*: The script has a simple input interface that only supports aligning a single audio file. The text converter is currently not in use. Support for multiple audio files and parallel multiprocessing, using the iterative dataloader, may be added later.

* *Note about tokenized text*: The current implementation does not support tokenized text. This may be added later, if needed. I already included the function for tokenized text in the library, as discussed in #2592.

* *Note about parallel processing*: The two steps, getting-lpz and segmenting, are separated into two functions, such that later, if needed, these two steps can be parallelized (and distributed over GPUs) using a multiprocessing queue. The `CTCSegmentationResult` object can be used to pass any intermediate lpz data and config. I'd expect a roughly 2x speedup with a single GPU-process, but also a heavily increased memory consumption. An implementation should take into account that for multiprocessing, data passed through queue is stored in shared memory - which can get large for longer audio files, in the range of several GB.

Thank you @kamo-naoyuki for the productive discussion about the interface.

#3018


lumaku:
A note on timing:
To determine the time stamps of the segments, the module needs to know the duration of one CTC index. In ESPnet 1, this was done from two variables: the frame duration and the subsampling factor of the encoder. My first approach for Espnet 2 uses a single inference to determine how many audio samples correspond to one CTC index. But, not all architectures are the same, i.e. some discard trailing audio data while others do not. The difference might be only 0.3% (~10ms on the short example file), but can already cause a significant time offset for longer audio files.

Because of this, it's better to infer the duration at inference directly as `len(speech) / len(ctc_output) * 1/fs` and give this as duration parameter to the algorithm.
This is new in the ctc_segmentation package version `1.6.0` - I recommend to update.
I also included the corresponding change in `asr_align.py` of ESPnet 1.

kan-bayashi:
@lumaku Sorry, to pass the CI, I temporary fixed ctc-segmentation version in #3136.

lumaku:
@kan-bayashi Sorry! The new changes are backwards compability for ESPnet 1, but I forgot about the "old" CTC segmentation test file ` test/test_ctc_segmentation.py`. We could remove it, as this is a duplicate of the test file in the CTC segmentation repository...

kan-bayashi:
@lumaku I see! Could you remove that file?

espnet2/bin/asr_align.py
```diff
@@ -0,0 +1,675 @@
...
+        if self.kaldi_style_text:
+            utt_ids = [utt.split(" ", 1)[0] for utt in text]
+            text = [utt.split(" ", 1)[1] for utt in text]
+        return utt_ids, text
+
+    def get_segments(self, text, lpz, timing_cfg=None):
+        """Obtain segments for given utterance texts and CTC log posteriors."""
+        config = self.config
+        # text is needed in the form of a list.
+        utt_ids, text = self._text_preprocess(text)
```
kamo-naoyuki, line 401: The preprocess must be done by preprocessor class. Could you refer the following ?https://github.com/espnet/espnet/blob/967deef41e579af031ba5d96774f2ec74d250d89/espnet2/bin/tts_inference.py#L141

kamo-naoyuki (commented):
Sorry for delaying.

espnet2/bin/asr_align.py
kamo-naoyuki: Sorry, I'm not sure  what text format you used here.

ESPnet uses the following text format as here:

```
uttid1 hello world
uttid2 abc der
```

What does "tokenized_text" mean here?

kamo-naoyuki: In my understanding, you are converting a raw text to integer ids: e.g `hello world` -> `3 10`. If so,
 we must use preprocess_fn. It's not allowed to refer `asr_model.token_list` directly because no one knows how to convert a  raw text to integer ids except for the preprocessor class.

Then, the option of "tokenized_text" is not needed because the input type is always raw text in espnet2.

lumaku: Thank you for the hint! `tokenized_text` was a workaround, it's not needed anymore when we have `preprocess_fn`.

So the input to the module shall be:
```python
text = """uttid1 hello world
uttid2 abc def
"""
```
Then, my module would split the utterances like this:
```python
text = ["hello world", "abc def"]
utt_ids = ["uttid1", "uttid2"]
```
I could then use `preprocess_fn` from `ASRTask.build_preprocess_fn` to obtain the token IDs, maybe like this:
```python
text = [preprocess_fn(utt) for utt in text]
# [np.array([ 3, 10]), np.array([ 17, 19])]
```
Will this work? I'll have to rewrite some functions and test, takes a few days...

kamo-naoyuki: > Will this work? I'll have to rewrite some functions and test, takes a few days...

Yes, thanks!

lumaku: The model now uses `preprocess_fn` and the kaldi style text format by default.

lumaku: The alignment result is a bit different when we use the `preprocess_fn`. The preprocessing method we used in espnet 1 scans for partial tokens, which is better for models with long tokens in their dictionaries. Therefore, I additionally included an option to handle the input text as in espnet 1 (with `text_converter="classic"`)...

The `asr_model.token_list` is used to generate debug information. For example, the `segments.state_list` of the `ctc_align_test.wav`:

```python
"".join(segments.state_list)
# "εεεεεεεεεεεεεεεε<blank>THEε<space>SεALEε<space>OεF<space>THE<space>HεOTεεEεLεεεεεSε<blank>εεεεIS<space>εPAεRT<space>εOF<space>HεOLεIεDAY'S<space>εSTRεATεEεGεYε<blank>εεεTO<space>SεELεL<space>OFεF<space>εAεSSεεεETεSε<blank>ANεD<space>εCOεNεCEεNεTRεATEε<space>ONε<space>PεROPεERεTYε<space>εMεANεAεGEεMEεNεT<blank>"
```

kamo-naoyuki: > The preprocessing method we used in espnet 1 scans for partial tokens, which is better for models with long tokens in their dictionaries. 

Sorry, I couldn't understand totally. Could you explain a bit more?

lumaku: Let's say we have a text `text = ["cat"]` and a dictionary that includes the word cat as well as its parts: `char_list = ["•", "UNK", "a", "c", "t", "cat"]`.
The "tokenize" method that uses the `preprocess_fn` will produce:

```python
text = "cat"
token_list = preprocess_fn("<dummy>", {"text": text})["text"]
token_list
# [array([5])]
ground_truth_mat, utt_begin_indices = prepare_token_list(config, text)
ground_truth_mat
# array([[-1],
#        [ 0],
#        [ 5],
#        [ 0]])
```

The previous method scans for tokens:
https://github.com/lumaku/ctc-segmentation/blob/596e20f23dc9a42780b42afeb736a47ace80152f/ctc_segmentation/ctc_segmentation.py#L272-L279

This will produce:

```python
text = ["cat"]
char_list = ["•", "UNK", "a", "c", "t", "cat"]
ground_truth_mat, utt_begin_indices = prepare_text(config, text, char_list)
# array([[-1, -1, -1],
#        [ 0, -1, -1],
#        [ 3, -1, -1],
#        [ 2, -1, -1],
#        [ 4, -1,  5],
#        [ 0, -1, -1]])
```

You can see here that the partial characters are detected (3,2,4), as well as the full "cat" token (5).
This is done to have a better time resolution for the alignment.

lumaku: Another point: The `preprocess_fn` converts unknown chars into the "\<unk>" token.
This may fill up the expected character sequence with "\<unk>" tokens, but these unknown characters should rather be ignored.
Is there an option for `preprocess_fn` to ignore those characters? - or maybe is there already a function to filter out these OOV parts separately?

lumaku: One question: Is the "\<unk>" token included in every espnet 2 dictionary?
https://github.com/espnet/espnet/blob/dab2092bc9c8e184c48cc6e603037333bd97dcd1/espnet2/text/token_id_converter.py#L15

If so, it would be easy to filter the tokenized sequence based on this specific token.

qdenisq:
Hi @lumaku, thanks for the integration with espnet2. I checked the `espnet2/bin/asr_align.py` and got a suspicion that there could be a bug in computing logprobs `lpz = self.ctc.log_softmax(enc.unsqueeze(0)).detach()`. Correct ime if I am wrong, but it seems that `ctc.log_softmax` expects the input shape to be [B, T, E] and after `enc.unsqueeze(0)` in my tests it becomes 4d: [1, 1, 157, 512]. So we should probably just pass `enc` as is.

Cheers

espnet2/bin/asr_align.py
kamo-naoyuki: > One question: Is the "<unk>" token included in every espnet 2 dictionary?

Yes. 

> char_list = ["•", "UNK", "a", "c", "t", "cat"].

I'm not sure. The token list should be char level or bpe symbols. What do you mean here?

lumaku:
Good catch @qdenisq ! This bug had a significant impact on the result, now the score is really better.....

espnet2/bin/asr_align.py
lumaku: OK, I included a filter for the "\<unk>" token.

lumaku: > I'm not sure. The token list should be char level or bpe symbols. What do you mean here?

For example, the char list of a model includes "a", "al", "as", "▁", "▁a", "▁about", "▁actually" ... when configured correctly, the above mentioned method finds parts of the text corresponding to the tokens.

lumaku: <details><summary>Example with a bpe 500 model char list from tedlium 2</summary><div>

```python

from ctc_segmentation import CtcSegmentationParameters
from ctc_segmentation import prepare_text

char_list = [ "<unk>", "'", "a", "ab", "able", "ace", "ach", "ack",
"act","ad","ag", "age", "ain", "al", "alk", "all", "ally", "am",
"ame", "an","and", "ans", "ant", "ap", "ar", "ard", "are",
"art","as", "ase","ass", "ast", "at", "ate", "ated", "ater", "ation",
"ations", "ause","ay", "b", "ber", "ble", "c", "ce", "cent",
"ces","ch", "ci", "ck","co", "ct", "d", "de", "du", "e", "ear",
"ect", "ed", "een", "el","ell", "em", "en", "ence", "ens",
"ent","enty", "ep", "er", "ere","ers", "es", "ess", "est", "et", "f",
"fe", "ff", "g", "ge", "gh","ght", "h", "her", "hing", "ht",
"i","ia", "ial", "ib", "ic","ical", "ice", "ich", "ict", "id", "ide",
"ie", "ies", "if","iff","ig", "ight", "ign", "il", "ild", "ill","im",
"in", "ind","ine", "ing", "ink", "int", "ion", "ions", "ip",
"ir","ire","is","ish", "ist", "it", "ite", "ith", "itt", "ittle",
"ity", "iv","ive", "ix", "iz", "j", "k", "ke", "king", "l",
"ld","le","ll","ly", "m", "ment", "ms", "n", "nd", "nder", "nt", "o",
"od","ody", "og", "ol", "olog", "om", "ome", "on",
"one","ong","oo","ood", "ook", "op", "or", "ore", "orm", "ort",
"ory", "os","ose", "ot", "other", "ou", "ould", "ound",
"ount","our","ous","ousand", "out", "ow", "own", "p", "ph", "ple",
"pp", "pt","q", "qu", "r", "ra", "rain", "re", "reat", "red",
"ree","res","ro","rou", "rough", "round", "ru", "ry", "s", "se",
"sel", "so","st","t", "ter", "th", "ther", "ty", "u", "ually",
"ud","ue", "ul","ult", "um", "un", "und", "ur", "ure", "us", "use",
"ust", "ut","v","ve", "vel", "ven", "ver", "very", "ves", "ving","w",
"way","x", "y", "z", "ăť", "ō", "▁", "▁a", "▁ab",
"▁about","▁ac","▁act","▁actually", "▁ad", "▁af", "▁ag", "▁al",
"▁all", "▁also","▁am", "▁an", "▁and", "▁any", "▁ar",
"▁are","▁around", "▁as","▁at","▁b", "▁back", "▁be", "▁bec",
"▁because", "▁been", "▁being","▁bet", "▁bl", "▁br", "▁bu",
"▁but","▁by", "▁c", "▁call","▁can","▁ch", "▁chan", "▁cl", "▁co",
"▁com", "▁comm", "▁comp","▁con", "▁cont", "▁could", "▁d",
"▁day","▁de", "▁des","▁did","▁diff", "▁differe", "▁different",
"▁dis", "▁do", "▁does","▁don", "▁down", "▁e", "▁en",
"▁even","▁every", "▁ex", "▁exp","▁f","▁fe", "▁fir", "▁first",
"▁five", "▁for", "▁fr", "▁from", "▁g","▁get", "▁go",
"▁going","▁good", "▁got", "▁h", "▁ha","▁had","▁happ", "▁has",
"▁have", "▁he", "▁her", "▁here", "▁his","▁how", "▁hum",
"▁hundred","▁i", "▁ide", "▁if", "▁im", "▁imp","▁in","▁ind", "▁int",
"▁inter", "▁into", "▁is", "▁it", "▁j", "▁just","▁k","▁kind",
"▁kn","▁know", "▁l", "▁le", "▁let", "▁li", "▁life","▁like",
"▁little", "▁lo", "▁look", "▁lot", "▁m",
"▁ma","▁make","▁man","▁many", "▁may", "▁me", "▁mo", "▁more",
"▁most","▁mu", "▁much", "▁my", "▁n", "▁ne", "▁need", "▁new",
"▁no","▁not","▁now", "▁o", "▁of", "▁on", "▁one","▁only", "▁or",
"▁other", "▁our","▁out", "▁over", "▁p", "▁part", "▁pe",
"▁peop","▁people","▁per","▁ph", "▁pl", "▁po", "▁pr", "▁pre", "▁pro",
"▁put", "▁qu", "▁r","▁re", "▁real", "▁really", "▁res",
"▁right","▁ro", "▁s","▁sa","▁said", "▁say", "▁sc", "▁se", "▁see",
"▁sh", "▁she", "▁show", "▁so","▁som", "▁some", "▁somet","▁something",
"▁sp","▁spe", "▁st","▁start", "▁su", "▁sy", "▁t", "▁ta",
"▁take","▁talk", "▁te","▁th","▁than", "▁that", "▁the", "▁their",
"▁them", "▁then", "▁there","▁these", "▁they", "▁thing",
"▁things","▁think", "▁this","▁those","▁thousand", "▁three",
"▁through", "▁tim", "▁time", "▁to", "▁tr","▁tw", "▁two", "▁u",
"▁un","▁under", "▁up", "▁us","▁v", "▁very","▁w", "▁want", "▁was",
"▁way", "▁we", "▁well", "▁were","▁wh","▁what", "▁when",
"▁where","▁which", "▁who", "▁why", "▁will","▁with", "▁wor", "▁work",
"▁world", "▁would", "▁y","▁year","▁years", "▁you", "▁your"]

text = ["I ▁really ▁like ▁technology", "The quick brown fox jumps over the lazy dog.", "unknown chars äüößß-!$ "]
config = CtcSegmentationParameters()
config.char_list = char_list
ground_truth_mat, utt_begin_indices = prepare_text(config, text)

# ground_truth_mat
# array([[ -1,  -1,  -1,  -1,  -1,  -1,  -1,  -1,  -1,  -1],
#        [  0,  -1,  -1,  -1,  -1,  -1,  -1,  -1,  -1,  -1],
#        [244,  -1,  -1,  -1,  -1,  -1,  -1,  -1,  -1,  -1],
#        [190, 410,  -1,  -1,  -1,  -1,  -1,  -1,  -1,  -1],
#        [ 55, 193, 411,  -1,  -1,  -1,  -1,  -1,  -1,  -1],
#        [  2,  -1,  -1,  -1,  -1,  -1,  -1,  -1,  -1,  -1],
#        [137,  13,  -1,  -1, 412,  -1,  -1,  -1,  -1,  -1],
#        [137, 140,  15,  -1,  -1,  -1,  -1,  -1,  -1,  -1],
#        [240, 141,  -1,  16,  -1,  -1, 413,  -1,  -1,  -1],
#        [244,  -1,  -1,  -1,  -1,  -1,  -1,  -1,  -1,  -1],
#        [137, 356,  -1,  -1,  -1,  -1,  -1,  -1,  -1,  -1],
#        [ 87,  -1, 359,  -1,  -1,  -1,  -1,  -1,  -1,  -1],
#        [134,  -1,  -1,  -1,  -1,  -1,  -1,  -1,  -1,  -1],
#        [ 55, 135,  -1,  -1, 361,  -1,  -1,  -1,  -1,  -1],
#        [244,  -1,  -1,  -1,  -1,  -1,  -1,  -1,  -1,  -1],
#        [209, 438,  -1,  -1,  -1,  -1,  -1,  -1,  -1,  -1],
#        [ 55,  -1, 442,  -1,  -1,  -1,  -1,  -1,  -1,  -1],
#        [ 43,  -1,  -1,  -1,  -1,  -1,  -1,  -1,  -1,  -1],
#        [ 83,  47,  -1,  -1,  -1,  -1,  -1,  -1,  -1,  -1],
#        [145,  -1,  -1,  -1,  -1,  -1,  -1,  -1,  -1,  -1],
#        [149,  -1,  -1,  -1,  -1,  -1,  -1,  -1,  -1,  -1],
#        [137, 153,  -1,  -1,  -1,  -1,  -1,  -1,  -1,  -1],
#        [149,  -1,  -1,  -1,  -1,  -1,  -1,  -1,  -1,  -1],
#        [ 79, 152,  -1, 154,  -1,  -1,  -1,  -1,  -1,  -1],
#        [240,  -1,  -1,  -1,  -1,  -1,  -1,  -1,  -1,  -1],
#        [  0,  -1,  -1,  -1,  -1,  -1,  -1,  -1,  -1,  -1],
#        [ 83,  -1,  -1,  -1,  -1,  -1,  -1,  -1,  -1,  -1],
#        [ 55,  -1,  -1,  -1,  -1,  -1,  -1,  -1,  -1,  -1],
#        [188,  -1,  -1,  -1,  -1,  -1,  -1,  -1,  -1,  -1],
#        [214, 189,  -1,  -1,  -1,  -1,  -1,  -1,  -1,  -1],
#        [ 87,  -1,  -1,  -1,  -1,  -1,  -1,  -1,  -1,  -1],
#        [ 43,  91,  -1,  -1,  -1,  -1,  -1,  -1,  -1,  -1],
#        [134,  49,  -1,  -1,  -1,  -1,  -1,  -1,  -1,  -1],
#        [ 40,  -1,  -1,  -1,  -1,  -1,  -1,  -1,  -1,  -1],
#        [190,  -1,  -1,  -1,  -1,  -1,  -1,  -1,  -1,  -1],
#        [149, 198,  -1,  -1,  -1,  -1,  -1,  -1,  -1,  -1],
#        [237, 181,  -1,  -1,  -1,  -1,  -1,  -1,  -1,  -1],
#        [145,  -1, 182,  -1,  -1,  -1,  -1,  -1,  -1,  -1],
#        [ 76,  -1,  -1,  -1,  -1,  -1,  -1,  -1,  -1,  -1],
#        [149,  -1,  -1,  -1,  -1,  -1,  -1,  -1,  -1,  -1],
#        [239,  -1,  -1,  -1,  -1,  -1,  -1,  -1,  -1,  -1],
#        [133,  -1,  -1,  -1,  -1,  -1,  -1,  -1,  -1,  -1],
#        [214,  -1,  -1,  -1,  -1,  -1,  -1,  -1,  -1,  -1],
#        [142, 220,  -1,  -1,  -1,  -1,  -1,  -1,  -1,  -1],
#        [183,  -1,  -1,  -1,  -1,  -1,  -1,  -1,  -1,  -1],
#        [204,  -1,  -1,  -1,  -1,  -1,  -1,  -1,  -1,  -1],
#        [149, 207,  -1,  -1,  -1,  -1,  -1,  -1,  -1,  -1],
#        [229,  -1,  -1,  -1,  -1,  -1,  -1,  -1,  -1,  -1],
#        [ 55, 230,  -1,  -1,  -1,  -1,  -1,  -1,  -1,  -1],
#        [190,  69, 233,  -1,  -1,  -1,  -1,  -1,  -1,  -1],
#        [209,  -1,  -1,  -1,  -1,  -1,  -1,  -1,  -1,  -1],
#        [ 83, 211,  -1,  -1,  -1,  -1,  -1,  -1,  -1,  -1],
#        [ 55,  -1,  -1,  -1,  -1,  -1,  -1,  -1,  -1,  -1],
#        [137,  60,  -1,  -1,  -1,  -1,  -1,  -1,  -1,  -1],
#        [  2,  -1,  -1,  -1,  -1,  -1,  -1,  -1,  -1,  -1],
#        [241,  -1,  -1,  -1,  -1,  -1,  -1,  -1,  -1,  -1],
#        [240,  -1,  -1,  -1,  -1,  -1,  -1,  -1,  -1,  -1],
#        [ 52,  -1,  -1,  -1,  -1,  -1,  -1,  -1,  -1,  -1],
#        [149,  -1,  -1,  -1,  -1,  -1,  -1,  -1,  -1,  -1],
#        [ 79, 152,  -1,  -1,  -1,  -1,  -1,  -1,  -1,  -1],
#        [  0,  -1,  -1,  -1,  -1,  -1,  -1,  -1,  -1,  -1],
#        [214,  -1,  -1,  -1,  -1,  -1,  -1,  -1,  -1,  -1],
#        [145, 221,  -1,  -1,  -1,  -1,  -1,  -1,  -1,  -1],
#        [134,  -1,  -1,  -1,  -1,  -1,  -1,  -1,  -1,  -1],
#        [145,  -1,  -1,  -1,  -1,  -1,  -1,  -1,  -1,  -1],
#        [149,  -1,  -1,  -1,  -1,  -1,  -1,  -1,  -1,  -1],
#        [237, 181,  -1,  -1,  -1,  -1,  -1,  -1,  -1,  -1],
#        [145,  -1, 182,  -1,  -1,  -1,  -1,  -1,  -1,  -1],
#        [ 43,  -1,  -1,  -1,  -1,  -1,  -1,  -1,  -1,  -1],
#        [ 83,  47,  -1,  -1,  -1,  -1,  -1,  -1,  -1,  -1],
#        [  2,  -1,  -1,  -1,  -1,  -1,  -1,  -1,  -1,  -1],
#        [190,  24,  -1,  -1,  -1,  -1,  -1,  -1,  -1,  -1],
#        [204,  -1,  -1,  -1,  -1,  -1,  -1,  -1,  -1,  -1],
#        [  0,  -1,  -1,  -1,  -1,  -1,  -1,  -1,  -1,  -1]])
```

</div></details>

lumaku: In the example, parts of the word "▁really" were separated into the token ids: [244, 410, 411, 412, 413]
This corresponds to `['▁', '▁r', '▁re', '▁real', '▁really']`

The CTC segmentation algorithm then iterates over these tokens in the ground truth, calculates the transition probabilities for each token from `lpz` and takes the token combination with the highest combined transition probability.

kamo-naoyuki: Thanks. I understood.

Then, we need to use some parts of the preprocessor:

https://github.com/espnet/espnet/blob/854afb8c0aa506a2aadc48c6259f64824c506c57/espnet2/train/preprocessor.py#L305-L308

```python
# e.g.
text = preprocress_fn. text_cleaner(text)
tokens = preprocress_fn.tokenizer.text2tokens(text)
(...create grand truth matrix using token_list)
```

At least, we must invoke `text_cleaner` for cleaning since you didn't apply it.

lumaku: Yes. The text cleaner is very helpful. Pre-trained ASR models usually don't have text cleaners - what would be the right way to add new text cleaners? Should I add a `text_cleaner: Collection[str] = None,` parameter to the `CTCSegmentation` object?

kamo-naoyuki: Do you need text_cleaner for CTCSegmentation now? Currently it's not used for all asr recipe now as you said.

lumaku: OK - For now, let's keep the already included cleaners. It's still possible to add a parameter for new cleaners later...

Text cleaning is certainly an essential step. That's what we did with the spacy framework in the [CTC segmentation paper](https://arxiv.org/abs/2007.09127), e.g., changing numbers to words, changing orthography, etc.

It is more important for establishing a good ground truth than for the actual alignment. Therefore, I would rather expect that text cleaning and text normalizing happens as a separate step before doing the segmentation. Doing text normalization as a separate step allows to manually check whether the result if acceptable.

lumaku:
Note on parallel processing:
The ASR inference and the CTC segmentation are two separate steps of the process; both can be parallelized:
1. ASR inference to get the log CTC posteriors can be done in batches, and is practical for short audio files with roughly the same length. But here, this is not always the case. It would bring less improvements for longer audio files.
2. CTC segmentation can be parallelized with `multiprocessing`. Aligning one file can be separated as one task. The number of tasks is limited by the amount of RAM and CPU threads available on the computer.

Aligning multiple files in parallel already works in the the command line interface.
The current script interface (argparse) does not yet support multiple input files.
Here is a dummy example that aligns our example audio file x-times in parallel:


```python
import soundfile
from espnet_model_zoo.downloader import ModelDownloader
from espnet2.bin.asr_align import CTCSegmentation
from torch.multiprocessing import Process, Queue

ngpu=0
NUMBER_OF_PROCESSES = 4
NUMBER_OF_AUDIO_FILES = 20


def align_worker(in_queue, out_queue):
    for task in iter(in_queue.get, 'STOP'):
        result = CTCSegmentation.get_segments(task)
        out_queue.put(result)


def test_multiprocessing_alignment():
    d = ModelDownloader(cachedir="./modelcache")
    wsjmodel = d.download_and_unpack("kamo-naoyuki/wsj")
    aligner = CTCSegmentation( **wsjmodel , fs=16000, samples_to_frames_ratio=512, ngpu=ngpu )

    # Create queues
    task_queue = Queue()
    done_queue = Queue()
    task_dict = {}

    # Start worker processes
    for i in range(NUMBER_OF_PROCESSES):
        Process(target=align_worker, args=(task_queue, done_queue)).start()

    # infer the audio to obtain lpz, and put task on queue
    for audio_file_nr in range(NUMBER_OF_AUDIO_FILES):
        # use a unique name for each audio file
        name = f"audio_{audio_file_nr}"
        # prepare text and lpz
        speech, rate = soundfile.read("./test_utils/ctc_align_test.wav")
        lpz = aligner.get_lpz(speech)
        text = "utt1 THE SALE OF THE HOTELS IS PART OF HOLIDAY'S STRATEGY TO SELL OFF ASSETS AND CONCENTRATE ON PROPERTY MANAGEMENT"
        task = aligner.prepare_segmentation_task(text, lpz, name, speech.shape[0])
        # put the task on the queue
        task_queue.put(task)
        # remember the task for re-assignment
        task_dict[task.name] = task

    # read the task results (segments) and re-assign to tasks
    for i in range(NUMBER_OF_AUDIO_FILES):
        task_result = done_queue.get()
        # to find the result that corresponds to a certain task,
        # use name as identifier; name must be unique
        name = task_result["name"]
        task_dict[name].set(**task_result)
        print(str(task_dict[name]), end="")
        # task no longer needed, as results was already printed
        del task_dict[name]

    # Tell child processes to stop
    for i in range(NUMBER_OF_PROCESSES):
        task_queue.put('STOP')
```

Start it with `test_multiprocessing_alignment()`.

espnet2/bin/asr_align.py
kamo-naoyuki: Sorry for late comment.

Could you change the code to my idea and remove the classic part?
Sorry, I'm not sure I could tell you my intention in the past comment.


```python
# e.g.
text = preprocress_fn. text_cleaner(text)
tokens = preprocress_fn.tokenizer.text2tokens(text)
(...create grand truth matrix using token_list)
```

lumaku: Sorry, I misunderstood your comment because I thought that `preprocess_fn.text_cleaner()` was already part of `preprocess_fn()`.
Changed it!

And also, for the classic method, true, it is used only in certain ASR recipes.
It is useful in those situations with long tokens that can be broken down in partial tokens.
I removed the option and left instructions in the form of a docstring for advanced users.

kamo-naoyuki: Sorry, I meant please remove not classic one and merge into one, classic only.
Then, I think , you need character/token lists instead of numpy.ndarray, so I thought we can't use `preprocess_fn()` because it converts it to ndarray.   We need to use `preprocress_fn. text_cleaner` and `preprocress_fn. text2tokens` in preprocress_fn.

kamo-naoyuki: Also note that preprocess_fn includes text_cleaner(): https://github.com/espnet/espnet/blob/fee8f1755e91d01bbefae4904630c3653fb1a379/espnet2/train/preprocessor.py#L304-L308

lumaku: Ah - now I understand better.
It's a little bit complicated - using the "classic" method has a disadvantage and an advantage.

The advantage is that it detects partial tokens.
Howeever, I think that this has also one disadvantage:
To explain this, I need to go into the details of how the `prepare_text` function is implemented.

Your proposed idea is great and correct, it helps the "classic" to better find tokens because of the preprocessing,
the implementation would maybe look like this:
```python
text = [self.preprocess_fn.text_cleaner(utt) for utt in text]
token_list = [" ".join(self.preprocress_fn.tokenizer.text2tokens(utt)) for utt in text]
token_list = [utt.replace("<unk>", "") for utt in token_list]
ground_truth_mat, utt_begin_indices = prepare_text(config, token_list)
```
The output would be something like `This <space> sequence <space> has <space> tokens <space> as <space> spaces.`
With your proposed method, all tokens will be detected correctly.

But when we compare the string-length of the ground truth:
```python
len("This sequence has tokens as spaces.")
# 35
len("This<space>sequence<space>has<space>tokens<space>as<space>spaces.")
# 56
```
As you can see the `token` sequence is longer.
The `prepare_text` function generates a ground truth matrix with dimensions `[length of the ground truth, length of longest token]`.
With this, in some cases, it happens that the prepared `ground_truth_mat` is longer than time steps in lpz (`lpz.shape[0]`)
and this causes an error in `ctc_segmentation`.
https://github.com/lumaku/ctc-segmentation/blob/905d6cc066cd7380b9d023050f5aa9b895eace0c/ctc_segmentation/ctc_segmentation.py#L142-L143
I would estimate, on average one token activation every 4 lpz indices, therefore, this error is not unprobable.

As the "easier" solution, I set the ESPnet 2 "tokenize" as default.
I argue that the "tokenize" method also prevents one error specific to the method how espnet 2 tokenizes.

I would argue: In 70% of use cases, the regular `tokenize` method is the best.
In the other 30% of cases, with the long tokens, then it would be be better to use the "classic" method.
Therefore, both methods have a reason.

You probably will prefer only one method to reduce the maintenance overhead.
For this, I recommend the "tokenize" method...

lumaku: The "tokenize" method is more efficient and the easiest:
* For models based on latin / cyrillic script that only contains the main characters, `["a", "b", ...]`, the "tokenize" method is already.
* For a Japanese or Chinese ASR model with e.g. ~3000 tokens Kanji / Hanzi

For models that contain long tokens, such as `_really`, the "classic" method improves the accuracy of the alignments, but with a greater complexity for implementation and computation...

kamo-naoyuki: > It's a little bit complicated - using the "classic" method has a disadvantage and an advantage.

Oh, I see.

> You probably will prefer only one method to reduce the maintenance overhead.

No, I misunderstood that classic one is better. My intension is using preprocess_fn for tokenization, so I think we should have two methods .

lumaku: Yes, I agree. I reverted to the commit with both methods and added documentation on `text_converter`. Thanks!

kamo-naoyuki: Please use the follwing in classic. We need to use text2tokens. My intention is here. External tools can't know how to tokenize the text.

```
text = [self.preprocess_fn.text_cleaner(utt) for utt in text]
token_list = [" ".join(self.preprocress_fn.tokenizer.text2tokens(utt)) for utt in text]
token_list = [utt.replace("<unk>", "") for utt in token_list]
ground_truth_mat, utt_begin_indices = prepare_text(config, token_list)
```

lumaku: I want to prevent the issue with too long ground truth sequences caused by long tokens, primarily by the space token `<space>`.
I think that a possible solution is to replace all instances of this token by space, i.e. to change `<space>` to " " at the tokenizer (`aligner.preprocess_fn.tokenizer.space_symbol`). Would this work?

kamo-naoyuki: Sorry, I didn't understand yet. How do you handle the problem in the current classic way? What is the difference?

lumaku: In the Espnet 1, the space token was just " " or it was ignored.
The difference is the length of the ground truth, see the example above: See `len("This sequence has tokens as spaces.")` versus `len("This<space>sequence<space>has<space>tokens<space>as<space>spaces.")`.

kamo-naoyuki: Why not be it done now as same? I think this situation is completely same.

lumaku: There is a small difference in how the `prepare_text` handles text.
But it's not that important. Sorry for the confusion.
The latest commit does not handle the space token differently.
Please check - If you are OK with this, then we can merge

kamo-naoyuki: If you are okay, it's okay for me, but I couldn't understand why don't you do as same? It seems easy.

lumaku: Ha! You are right, why make it easy if we can make it complicated :wink: ?
I am OK with the code as it is.

lumaku: The CI test failed - it seems because of an unknown error. Can we restart it?

kamo-naoyuki:
Okay, then, let's merge this PR. Thank you! Sorry for waiting long time due to my delaying.

PR #3110, opened 2021-03-25
Update loss report

b-flo:
- Update loss report for transducer models and include all auxiliaries losses
- Fix `mtl_mode` with custom transducer model (forgot to update when I replaced `transformer|conformer` type to `custom`)

For transducer model reporting in `asr.py`, I went for a straightforward solution (I didn't have too much time to think about it), if anyone have a proper solution feel free to ask for modifications!


sw005320:
LGTM!
This is very useful!
Thanks, @b-flo!

PR #3140, opened 2021-04-07
Add IWSLT21 MT recipe

hirofumi0810:
Added new recipe for IWSLT21 MT (Must-C v1+v2, ST-TED, WMT20).


ftshijt:
LGTM! Many thanks.

PR #3142, opened 2021-04-08
Fix tokenization on TEDLIUM2 in IWSLT21 ASR recipe

hirofumi0810:
Added tokenization with moses for TEDLIUM2 in `stage-0`.

### DIFF
--- a/egs/iwslt21/asr1/run.sh
+++ b/egs/iwslt21/asr1/run.sh
@@ -126,7 +126,7 @@ if [ ${stage} -le 0 ] && [ ${stop_stage} -ge 0 ]; then
     # additionally we copy text to text.${case}
     for x in tr_${data_code} dt_${data_code} et_${data_code}_dev et_${data_code}_test; do
         for case in tc lc lc.rm; do
-            cp data/${x}/text data/${x}/text.${case}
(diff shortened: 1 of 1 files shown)


hirofumi0810:
@sw005320 @ftshijt This PR is ready.

egs/iwslt21/asr1/run.sh
```diff
@@ -126,7 +126,7 @@ if [ ${stage} -le 0 ] && [ ${stop_stage} -ge 0 ]; then
     # additionally we copy text to text.${case}
     for x in tr_${data_code} dt_${data_code} et_${data_code}_dev et_${data_code}_test; do
         for case in tc lc lc.rm; do
-            cp data/${x}/text data/${x}/text.${case}
+            paste -d " " <(cut -d " " -f 1 data/${x}/text) <(cut -d " " -f 2- data/${x}/text | lowercase.perl | tokenizer.perl -l en -q) > data/${x}/text.${case}
```
ftshijt, line 129: it seems that we also do the `lowercase` for case `tc`, is it intentional?

ftshijt: I understand it's for asr training, so this might not be that important though.

hirofumi0810: @ftshijt Because TEDLIUM2 is already lowercased, I think it's fine.

ftshijt: Cool~

ftshijt:
Merged! Thanks for the update.

PR #3144, opened 2021-04-09
Ensemble for ST

ftshijt:
Add a simple wrapper for ST models

### DIFF
--- a/espnet/bin/st_trans.py
+++ b/espnet/bin/st_trans.py
@@ -89,7 +89,11 @@ def get_parser():
     )
     # model (parameter) related
     parser.add_argument(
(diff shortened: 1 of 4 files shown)


ftshijt:
The current version basically follows fairseq's methods:

- ensemble models should use the same units (e.g., char or bpe)
- models should have a uniform interface (e.g., encode(), decode_one_step() ...)
- all the models are with equal weights (using logsumexp)


simply add `--model ${model1} --model ${model2}` would be fine for st-translate

ftshijt:
Current ensembling does not support ST with basic RNN, because of the interface issue. I would consider redesigning it in the future, so will close the PR

sw005320:
I agree.

PR #3145, opened 2021-04-09
IWSLT21 punctuation restoration recipe

hirofumi0810:


ftshijt:
LGTM! Will merge it soon~

We may not consider it now, but maybe after this period, we could explain a bit about the workflow in README to this eg. If considering pre-training, the process would have many steps.

In addition, I'm curious about if the mt settings would perform well on this punctuation recovering, is that too large a network for the task?

PR #3174, opened 2021-04-25
fix receipe bug for swbd

yuekaizhang:
espnet1 swbd recipe has several minor bugs

For example, in data/train/text 
after this command `sed -i 's/\._/ /g; s/\.//g; s/them_1/them/g' data/${train_set}/text`
_sp0.9xxxx this is a t.e.xt --> sp09xxxx this is a text_ would change the utterance name and cause utternance name mismatch with other scp files.


sw005320:
Thanks a lot!

sw005320:
@yuekaizhang, I don't know why, but after I merge your PR, I got the following error
https://github.com/espnet/espnet/runs/2442097675?check_suite_focus=true#step:8:18
Is it related to your change or some others?

It is strange for me because your PR passed the test...

yuekaizhang:
I only changed the swbd run.sh. It looks like some others' caused this problem.

sw005320:
Yes, it was fixed by #3180.
Sorry about it.

PR #3187, opened 2021-04-30
add s3prl frontend

simpleoier:
Add [s3prl](https://github.com/s3prl/s3prl) frontend to extract unsupervised learning representations, in addition to usual spectrum features.
Still developing training configs.


sw005320:
How about starting from CTC like s3prl projects?

sw005320:
Can you fix https://github.com/espnet/espnet/pull/3187/checks?check_run_id=2495233203#step:8:11?

simpleoier:
> Can you fix https://github.com/espnet/espnet/pull/3187/checks?check_run_id=2495233203#step:8:11?

Yes, I will do it soon.

tools/installers/install_s3prl.sh
```diff
@@ -0,0 +1,67 @@
...
+    sed -i -e "s/torch==/torch>=/" \
+        -e "s/torchaudio==/torchaudio>=/" \
+        -e "s/numba==/numba>=/" \
+        -e "s/librosa==/librosa>=/" \
+        -e "s/tqdm==/tqdm>=/" \
+        requirements.txt
+    pip install -r requirements.txt
+
+else
+    echo "[WARNING] fairseq is not prepared for pytorch<1.7.0, python<3.6 now"
```
kan-bayashi, line 65:
```suggestion
    echo "[WARNING] s3prl is not prepared for pytorch<1.7.0, python<3.6 now"
```

egs2/librispeech/asr1/local/path.sh
```diff
@@ -0,0 +1,3 @@
+if [ -d ${MAIN_ROOT}/tools/s3prl ]; then
```
kan-bayashi, line 1: Maybe updating `tools/extra_path.sh` is better?

espnet2/asr/frontend/s3prl_frontend.py
```diff
@@ -0,0 +1,105 @@
...
+
+        # Saving CUDA Memory
+        del feats
+
+        return input_feats, feats_lens
+
+    def _compute_stft(
+        self, input: torch.Tensor, input_lengths: torch.Tensor
+    ) -> torch.Tensor:
+        raise NotImplementedError
```
kan-bayashi, line 105: Is this part needed?

espnet2/asr/frontend/s3prl_frontend.py
```diff
@@ -0,0 +1,105 @@
...
+        self, input: torch.Tensor, input_lengths: torch.Tensor
+    ) -> Tuple[torch.Tensor, torch.Tensor]:
+        wavs = [wav[: input_lengths[i]] for i, wav in enumerate(input)]
+        if self.upstream_trainable:
+            feats = self.upstream(wavs)
+        else:
+            with torch.no_grad():
+                feats = self.upstream(wavs)
+        input_feats = pad_list(feats, 0.0)
+        feats_lens = torch.tensor([f.shape[0] for f in input_feats], dtype=torch.long)
```
kan-bayashi, line 92: `input_feats` seems padded so the length will be the same. Is it correct?

tools/installers/install_s3prl.sh
```diff
@@ -0,0 +1,67 @@
...
+EOF
+)
+echo "cuda_version=${cuda_version}"
+
+if "${torch_17_plus}" && "${python_36_plus}"; then
+
+    rm -rf s3prl
+
+    # S3PRL Commit id when making this PR: `commit a6d2ae11c644c024cd096b3dc362f44dc91a51e5`
+    git clone https://github.com/s3prl/s3prl.git
```
kan-bayashi, line 51: git clone https://github.com/s3prl/s3prl.git -b superb-interspeech2021

kan-bayashi: It is better to add TODO comment to use tagged version?

tools/installers/install_s3prl.sh
simpleoier: Thanks for the correction!

tools/installers/install_s3prl.sh
simpleoier: Good suggestion. Will do it.

espnet2/asr/frontend/s3prl_frontend.py
simpleoier: You are correct. It should be the `feats` rather than `input_feats`. I guess it was auto-filled by IDE. Sorry for this.

egs2/librispeech/asr1/local/path.sh
simpleoier: Good idea! Will do it.

lumaku:
@simpleoier I currently prepare distribution packages, and found that there might be duplicate dependencies. Espnet depends on `kaldiio`, while s3prl depends on `kaldi_io`. Can you check this?
`kaldi_io`: https://github.com/vesis84/kaldi-io-for-python
`kaldiio`: https://github.com/nttcslab-sp/kaldiio

simpleoier:
@lumaku  I'm just now working on this PR. Thanks for the comment. Although I didn't notice this conflict, I will check it before I push my modifications.

simpleoier:
@sw005320 @kan-bayashi Can you look at this PR again? I'm preparing another PR for the configuration and results on WSJ.

sw005320 (commented):
LGTM.
I basically want you to add some tests.
- can you add Conv2dSubsampling2 test? Maybe in https://github.com/espnet/espnet/blob/08feae5bb93fa8f6dcba66760c8617a4b5e39d70/test/espnet2/asr/encoder/test_conformer_encoder.py#L8 ?
- is it possible to add S3prlFrontend test? Maybe it is difficult to do it as we need to install fairseq, but once it is made, our implementation becomes robust I believe
- can you add a unit test of `preencoder/linear.py`?

espnet/nets/pytorch_backend/transformer/subsampling.py
```diff
@@ -98,6 +100,65 @@ def __getitem__(self, key):
         return self.out[key]
 
 
+class Conv2dSubsampling2(torch.nn.Module):
```
Fhrozen, line 103: Given that there are multiples Conv2Sub, I think that it could be possible to set a one single class and when calling change the kernel or stride parameters.

espnet2/asr/frontend/s3prl_frontend.py
```diff
@@ -0,0 +1,120 @@
+from argparse import Namespace
```
Fhrozen, line 1: The name of this file could be just `s3prl` because it is already inside the frontend folder.

espnet2/asr/preencoder/linear.py
```diff
@@ -0,0 +1,38 @@
+#!/usr/bin/env python3
+#  2021, Carnegie Mellon University;  Xuankai Chang
+#  Apache 2.0  (http://www.apache.org/licenses/LICENSE-2.0)
+
+"""Sinc convolutions for raw audio input."""
```
Fhrozen, line 5: Is this sinc or linear?

simpleoier:
Thanks @sw005320 for your comments. I'll add the tests for Conv2d2 and linear preencoder. Regarding to S3prlFrontend, because s3prl and fairseq are optional to install, it may be difficult. If we can install s3prl as default, an easy test for `log_stft` as upstream probably can be used.

sw005320:
> Thanks @sw005320 for your comments. I'll add the tests for Conv2d2 and linear preencoder. Regarding to S3prlFrontend, because s3prl and fairseq are optional to install, it may be difficult. If we can install s3prl as default, an easy test for `log_stft` as upstream probably can be used.

Year, I totally understand it. This is very difficult to include such huge libraries in the test. For now, we can leave this, but we can think of it in the future.

sw005320:
I think it’s in a good shape.

@pengchengguo, could you quickly take a look at this review?
Then, I’ll merge this PR.

tools/installers/install_s3prl.sh
```diff
@@ -0,0 +1,56 @@
...
+
+if "${torch_17_plus}" && "${python_36_plus}"; then
+
+    rm -rf s3prl
+
+    # S3PRL Commit id when making this PR: `commit a3318ca43e9cdf0b33cc934b91ed37bdbf883abb`
+    git clone https://github.com/s3prl/s3prl.git
+
+else
+    echo "[WARNING] s3prl is not prepared for pytorch<1.7.0, python<3.6 now"
```
Fhrozen, line 54: Is this package only be downloaded? why cannot be installed ignoring the requirements `--ignore-requires-python` option? so then you can avoid adding the export line at `tools/extra_path.sh`. I observed that `kaldi_io` is used for preprocessing specific data, such as timit or voxceleb: https://github.com/s3prl/s3prl/blob/master/preprocess/timit2ark.py or https://github.com/s3prl/s3prl/blob/master/preprocess/ark2voxceleb.py

espnet2/asr/preencoder/linear.py
simpleoier: Thanks! Will fix it.

espnet2/asr/frontend/s3prl_frontend.py
simpleoier: Thanks! Will fix it!

espnet/nets/pytorch_backend/transformer/subsampling.py
simpleoier: Good idea. I think it's time to think about it.

sw005320: I agree. This is a very cool idea.

tools/installers/install_s3prl.sh
simpleoier: Just download it could work. I am not sure which one is better, only download it or install it.

sw005320: In this case, if it works, my preference is only download.
If it’s working, we can also include it in the test since it would be quick.

sw005320: Same for fairseq.
But probably only download would not work for fariseq…

simpleoier: > In this case, if it works, my preference is only download.
> If it’s working, we can also include it in the test since it would be quick.

Just to confirm, do we still make s3prl optional?

sw005320: Yes, I think it (keeping s3prl to be an optional package but adding a test script) would be a good idea.
I think you can do it in a later PR.
Can you keep this as an action item?

simpleoier: Sure. I'll add it in the next PR, together with WSJ recipes and results.

espnet/nets/pytorch_backend/transformer/subsampling.py
sw005320: @Fhrozen, is it possible for you to do it as a separate PR?
@simpleoier is preparing a follow up PR and we want to merge this PR as soon as possible.

sw005320:
The PR looks good and I’ll merge it after it passes the CI test.
We can keep on working on several items that we discussed here in a follow up PR.

sw005320:
Another important item is to add this great function to our main README.md in the later PR.
We should advertise this!!!

pengchengguo:
Hi @simpleoier, how about using ‘frontend_conf[name] is startswith “wav2vec”’ for this line https://github.com/espnet/espnet/pull/3187/files#diff-0315b63228f74c3c13237696e0b62c774c47ff95a14688bda4b314e39be40ef8R50. Maybe it will be a little concise.

sw005320:
I just merged it. Thanks @simpleoier!!!
Also, thanks @pengchengguo for your review. You can keep on reviewing this PR.
I believe @simpleoier will incorporate it in the next PR.

PR #3197, opened 2021-05-06
[WIP][ESPnet2] Mask-CTC

YosukeHiguchi:
This PR adds [Mask-CTC](https://arxiv.org/abs/2005.08700)-based non-autoregressive ASR model to ESPnet2. I only added the training part and will work on the inference soon.

I found it difficult to implement the model by simply replacing the ctc-attention model's `decoder` with masked LM. Instead, I defined a new model class `MaskCTCModel` in maskctc_model.py, which inherits some common methods defined in `ESPnetASRModel`.  With this change, one can specify which ASR model architecture to use in a training configuration. Could you give me some advice if there's a better way for the implementation, or am I doing OK?


sw005320:
I’m OK about your design but other people would have different opinions about this kind of design discussions.
Could you fix https://github.com/espnet/espnet/pull/3197/checks?check_run_id=2516383613#step:8:10 ?

sw005320:
@YosukeHiguchi, let's try to finish this PR.
Could you first fix the CI error?

@hirofumi0810, could you review this PR?

sw005320:
Can you add some tests?

sw005320:
Can you add a test, e.g., under https://github.com/espnet/espnet/tree/master/test/espnet2/asr/decoder?
Also, after it is ready, can you remove [WIP] from the issue title?

sw005320:
@YosukeHiguchi, can you restart this?
I think this would be a very cool function.
We can show the effectiveness of mask-ctc with various demonstrations in espnet2.

sw005320:
Given the conversation today, you can continue this PR (or make another PR) for the mask CTC development in espnet.
I think the current design (having a different model from `espnet_model.py`) sounds good to me.

PR #3199, opened 2021-05-06
[WIP] GTN CTC bug fix, unit test, and installer

brianyan918:
This PR is WIP - I am testing the new implementation of CTC loss on Voxforge IT.

In a previous PR (https://github.com/espnet/espnet/pull/2866) I added a 'gtnctc' option for computing CTC loss using GTN differentiable WFSTs. This PR is a follow-up containing: 1) a bug fix 2) unit tests, and 3) installer.

1) Bug fix for improper handling of padding in log prob tensor.
The differentiable WFST based CTC loss function processes samples in the batch in parallel, meaning we index a tensor of size (T, C) from the batched tensor of size (B, T, C) representing log probs. Problem is that this includes padding in T that is not ignored. The fix for this is to slice the (T, C) tensor according to the unpadded input lengths: https://github.com/brianyan918/espnet-ml/blob/8c5b2414cb5351aa3584d8178e66034a8b555d89/espnet/nets/pytorch_backend/gtn_ctc.py#L59

2) Unit tests for CTC losses.
Following the style of the previous unit tests comparing warp and torch ctc to chainer ctc, I added tests for gtn ctc. We also added a cudnn ctc loss (https://github.com/espnet/espnet/pull/3001) and I added a unit test for this too.

3) Installer for GTN package.
Note GTN's requirement on Python version (>=3.5) isn't checked in the installer since it is looser than ESPnet's own requirement (>=3.6.1)

### DIFF
--- a/tools/installers/install_gtn.sh
+++ b/tools/installers/install_gtn.sh
@@ -1,0 +1,19 @@
+#!/usr/bin/env bash
+set -euo pipefail
+
+
+if [ $# != 0 ]; then
+    echo "Usage: $0"
+    exit 1;
+fi
+
+# Install pyopenjtalk
+if [ ! -e gtn.done ]; then
+    (
+        set -euo pipefail
+        python3 -m pip install gtn
+    )
+    touch gtn.done
+else
+    echo "gtn is already installed"
+fi
--- a/tools/installers/install_gtn.sh
+++ b/tools/installers/install_gtn.sh
@@ -7,7 +7,7 @@ if [ $# != 0 ]; then
     exit 1;
 fi
 
-# Install pyopenjtalk
+# Install gtn
 if [ ! -e gtn.done ]; then
     (
         set -euo pipefail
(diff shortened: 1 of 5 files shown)


tools/installers/install_gtn.sh
```diff
@@ -0,0 +1,19 @@
+#!/usr/bin/env bash
+set -euo pipefail
+
+
+if [ $# != 0 ]; then
+    echo "Usage: $0"
+    exit 1;
+fi
+
+# Install pyopenjtalk
```
sw005320, line 10:
```suggestion
# Install gtn
```

sw005320 (approved):
Thanks! LGTM.
Will you add a Voxforge IT recipe?
You can make another PR about it to avoid making a PR too large.

brianyan918:
> Will you add a Voxforge IT recipe?
> You can make another PR about it to avoid making a PR too large.

OK - Then I will add it in another PR after I have finished testing it.

sw005320:
Thanks, @brianyan918!
The PR looks good and I just merged it.
Please continue to work on the recipe.

PR #3208, opened 2021-05-10
Fix a bug about variable spelling errors

lzm0706:
The variable subsampling_factors in class CustomConverterMulEnc was  incorrectly  written as subsamping_factors , resulting in inconsistency.


sw005320:
Thanks a lot!

PR #3217, opened 2021-05-15
Transducer v5

b-flo:
This PR is refactor+update of the code related to transducer model and its features. The number of changed lines may be big but there are a lot of "cosmetic" changes. I'll update all configs, results and models in the next PR and new ones will be added after that for Aishell1, Commonvoice, Librispeech and Voxforge-IT.

**General**

- Use consistent naming for local variables, parameters, structures, etc.
- Add typing to all classes and methods.
- Remove dead(?) code related to tensor definition such as data type. I think some part were unintentionally removed in the past, I will properly add it back if needed. 
- Update classes' and methods' documentation.

**Architecture**

- Return auxiliary encoder time lengths for auxiliary transducer criterion in `RNNEncoder` and `CustomEncoder`.

**Tasks / Losses**

- Add `TransducerTasks` encapsulating all transducer tasks and related modules + refactor `e2e_asr_transducer`.
- Use symmetric KL divergence instead of Jensen-Shannon divergence for L4. The first one is originally proposed in https://arxiv.org/abs/2011.03109.
- Add exception for symmetric KL divergence loss when inputs have different time lengths.
- Modify reduction for all tasks output ("mean" -> "sum" / B or "sum" / (B x T x U x L) for symm KL)

**Training**

- Remove scheme to automatically find most dominant dimension for Noam scheduler and rely on newly added `noam-adim` argument instead. Noam can now be used with RNN architecture.
- Use beam search with `beamsize=2`for CER/WER reporting instead of "constrained" greedy search. We trade some training time for more precision in reporting.

**Decoding**

- Add `temperature` for softmax to all search algorithms.
- Add modified Adaptive Expansion Search (mAES). It's basically a NSCv2 were AES (https://ieeexplore.ieee.org/document/9250505) is modified to handle N-steps expansion instead of relying on the proposed 2-steps expansion.

**Transfer learning**

- Added compatibility for RNN encoder weights transfer. It couldn't work before because I made a new RNN class to access auxiliary encoder layer weights and there was some slight differences (naming + weights shape).
- Improved reporting when there is a mismatch between specified pre-trained model module(s) and new model modules()
- Raise error and stop training for failed cases instead of raising warning and carrying on.

**Installation**

- ~~Modify `warp-transducer` installation to support CUDA >10.2.~~ I decided to rely on my updated fork of warp-transducer instead of doing on-the-fly patching for the moment. Support for CUDA 11.1 and Ampere GPU was added.
- Remove support for `warp-rnnt`.

I'll stick to only one RNN-T loss computation module until we replace it by our own. On that note, I'm also testing a numba implementation based on SpeechBrain's one. There is a slight training speed improvement compared to `warp-transducer` and I kinda like the integration but I think a C++ implementation is still preferred.

**Test**

- Remove non used variables and parameters in `test_e2e_asr_transducer.py` and `test_custom_transducer.py`.
- Fix some test for auxiliary tasks in `test_e2e_asr_transducer.py`.
- Fix and improve test for transfer learning in `test_asr_init.py`
- Fix transducer configs for integration test.
- Add integration test for transducer with auxiliary tasks.

**Documentation**

- Add mention of auxiliary tasks in `README.md` + refactor transducer section.
- Add documentation for auxiliary tasks in `tutorial.md` + refactor transducer section.


jinggaizi:
why remove warp-rnnt

b-flo:
Hi,

> why remove warp-rnnt

Because, ultimately, we want to only support one RNNT loss computation module and current one will be replaced by our own in a near future. We kept `warp-transducer` because both CPU and GPU are supported, and from my experience and given feedback there was no clear advantage in using `warp-rnnt`.

espnet/nets/pytorch_backend/transducer/arguments.py
```diff
@@ -201,33 +202,32 @@ def add_custom_decoder_arguments(group):
     return group
 
 
-def add_custom_training_arguments(group):
+def add_custom_training_arguments(group: _ArgumentGroup) -> _ArgumentGroup:
     """Define arguments for training with Custom architecture."""
     group.add_argument(
-        "--transformer-warmup-steps",
```
sw005320, line 207: Please accept the old one and add some warning as deprecate options.

espnet/nets/pytorch_backend/transducer/arguments.py
```diff
@@ -201,33 +202,32 @@ def add_custom_decoder_arguments(group):
...
+        "--optimizer-warmup-steps",
         default=25000,
         type=int,
         help="Optimizer warmup steps",
     )
     group.add_argument(
-        "--transformer-lr",
```
sw005320, line 213: ditto

sw005320 (commented):
Many thanks, @b-flo!
I have several comments and questions.
-  espnet/nets/beam_search_transducer.py: My understanding of the major changes of this code is to just rename the variables. I think it would be fine but they are not so necessary. If they are very different from the other beam search functions and may cause some confusion, we may use the original ones. I don't have particular examples for this, but please double-check them from this perspective.
- espnet/nets/pytorch_backend/e2e_asr_transducer.py: same comments for the above. Some of the variable name changes do not seem necessary and break the consistency with the others.

After I reviewed the codes, my overall comments are above. I'm concerning that some naming rules seem to be different from the other files (e.g., xs, xs_pad, ys). Your naming would be better but it loses the consistency. Also, these changes produce unnecessary large file changes.
Could you minimize these changes?

Note that I fully agree with better naming for variables. I think the long-term goal would be to unify to use more intuitive variable names like what you proposed for all our sources, but this can be done later.

sw005320:
@kan-bayashi, could you also take a look at the code?
I'm concerning that some of these changes might affect the other parts (I could not fully check them because this PR has many changes due to the variable name change).

b-flo:
Thank you for you review @sw005320 ! I'll minimize the changes as requested but I would like to point out a few things for the sake of discussion:

- Old models won't be compatible anyway because of some inner changes such as newly added `TransducerTasks`. 
- Transducer in ESPnet1 is self-contained outside shared training utilities/scheme, proposed changes won't impact other models or utilities.
- While I initially kept `xs`, `xs_pad`, etc, I eventually switched because some things were confusing without further changes:
    - We also use `x` and `y` for some lambda/inner operations not implicating `xs`, `ys`, etc.
    - In some part, we (obviously) re-use the variable but naming isn't self-explanatory (e.g.: `xs_pad` containing unpadded values)
    - Generic naming (+ doc and typing) for model/architecture-independant layers, modules or operation is easier for re-use.

I do think we can keep proposed naming without confusing users used to the "old" naming or other models. However, I do agree I should keep old naming for shared parameters and we should update it later for all models altogether! If that's okay for you, I will revert `transformer-*` parameters, `yseq` attribute and other shared ones.

Note that I do not have a strong opinion on the naming in itself. I just want to make it easier for new users or potential contributors to understand. Also, I do know it's one of the motivation for ESPnet2 and I shouldn't bother but porting some features or models will take time and at the end I suppose re-used code from ESPnet1 will be refactored.

sw005320:
Yes, I fully agree with your comments.
My primary concern is a couple of shared ones.
Also, the internal variables must be OK especially when they are specific for transducer but it would be great if you consider some consistency across the other beam search variable names.

From a maintenance perspective, what I usually do is, if some changes happen that will affect the overall beam search parts, I'll search the entire code with the variable name as a query.
If the variable naming rules are changed, I would overlook the code that we have to deal with.
And as you mentioned, we can avoid it and get the benefit of better renaming if we apply it to all our codes consistently.

Also, your renaming of `transformer-*` is good to me but I just want to have some backward compatibility.
So, if you accept the old naming with a deprecation warning, that is perfect.
This is actually renamed in espnet2 and it is very reasonable.

So, overall, 

> If that's okay for you, I will revert transformer-* parameters, yseq attribute and other shared ones.

Sure, that's reasonable.
But you can keep your new names for `transformer-* parameters` if you care about the backward compatibility.

b-flo:
Got it, sorry I thought you meant "keeping the old models compatible" when talking about backward compatibility which wouldn't work even by keeping deprecated parameters. I'll add back parameters with a deprecation notice!

> Also, the internal variables must be OK especially when they are specific for transducer but it would be great if you consider some consistency across the other beam search variable names.

For information, I switched `yseq` to `label_seq` to emphasise the difference compared to other models. While using `y` naming for CTC and Att as output sequence or decoder output sequence is normal, it's a bit confusing to use `yseq` for transducer models because the label sequence is constructed based on the encoder and decoder combination, where the final representation is referred as `z`. All in all, `zseq` may be better defined for transducer while `label_seq` is more generic and could be used for all models.

espnet/asr/asr_utils.py
```diff
@@ -831,7 +831,10 @@ def parse_hypothesis(hyp, char_list):
 
     """
     # remove sos and get results
-    tokenid_as_list = list(map(int, hyp["yseq"][1:]))
+    if "label_seq" in hyp:
```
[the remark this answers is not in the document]
b-flo: My apologies, I forgot to answer. Please see my last github message for explanation

sw005320:
@b-flo,
I'm still concerning several naming changes that I discussed and waiting for your update on reverting several important names, but I also don't want to make it pending for a long time.
If you don't plan to update this PR, I'll merge it.

b-flo:
> I'm still concerning several naming changes that I discussed and waiting for your update on reverting several important names, but I also don't want to make it pending for a long time.

Sorry, I thought the issue was resolved with my last changes outside `label_seq` we were still discussing. Could you point out the important names which still bother you? I'll update asap.

sw005320:
Is it possible to split this PR into the main implementations and the renaming PRs?
Sorry that this PR mixes everything, and it is difficult to check each variable one by one.

espnet/asr/pytorch_backend/asr.py
```diff
@@ -514,16 +514,23 @@ def train(args):
     elif args.opt == "noam":
         from espnet.nets.pytorch_backend.transformer.optimizer import get_std_opt
 
-        # For transformer-transducer, adim declaration is within the block definition.
-        # Thus, we need retrieve the most dominant value (d_hidden) for Noam scheduler.
-        if hasattr(args, "enc_block_arch") or hasattr(args, "dec_block_arch"):
-            adim = model.most_dom_dim
+        if "transducer" in mtl_mode:
```
sw005320, line 517: Can you make these to be minimum changes?
I see that the name of `args.transformer_***` does not make sense, but the inconsistent change (only apply this change in the transducer mode) becomes more inconsistent.

b-flo: My apologies, I'm a bit confused. Do you want me to revert that part?

> Remove scheme to automatically find most dominant dimension for Noam scheduler and rely on newly added noam-adim argument instead. Noam can now be used with RNN architecture.

For `args.transformer_***`, I added the old parameters with a deprecation notice as requested: https://github.com/espnet/espnet/pull/3217/commits/38dc2bc9236abc9973bd7fd085d5f7bb4189aea4
In relation to the part you pointed out, both `adim` and new `noam-adim` are supported.

b-flo:
> Is it possible to split this PR into the main implementations and the renaming PRs?

I can try, sure.

espnet/asr/pytorch_backend/asr.py
sw005320: Yes, I noticed it, but I feel that keeping the different naming does not look good to me and the deprecation notice is not enough.

b-flo:
Hi,

Sorry for the delay, I'm back.
@sw005320 It will be a bit difficult to revert all changes related to naming, if possible could you please give me a list of names   you want reverted (or pinpoint part of code you want changed)? Sorry, I've re-read previous messages and I'm still confused on what I can change or should be replaced, how I should handle parameters changes, etc.
Btw, I think we shouldn't really care about compatibility changes in the end. This version shoud be seen as a proper TransducerV2 and I'll update all recipes and models in the next PR anyways.

sw005320:
Thanks a lot!

Actually, the PR is getting bigger and bigger, and it is very difficult for me to check all of them.
So, as I said, how about splitting this PR?
We'll check them one by one.
Then, naming issues would also be discussed in a separate PR with various cares.

Sorry to ask for additional efforts, but I feel that this is more efficient and can be merged more easily.

sw005320:
If you cannot split this PR, please let me know.
I'll ask other people for this review.

b-flo:
> If you cannot split this PR, please let me know.
> I'll ask other people for this review.

Let's do that, thanks. I'm trying to split the PR anyways but it will take some time.

sw005320:
@pengchengguo and @Fhrozen, could you review this pr? I'm really caring about the compatibility.

b-flo:
OK, it's now complete. Now onto the renaming issue.

As it was pointed out before, the remaining issue is with shared variable names and parameters. After comparing to previous versions and other models, it concerns the following:

 - Local variables in `e2e_asr_transducer.py` (e.g.: `xs`, `ys`, `xs_pad`, `ys_pad`, etc).
 - An attribute in hypothesis dict, used in `asr_init.py` (e.g.: `label_seq` instead of  yseq`).
 - Model training parameters (e.g.:`transformer_warmup_steps` and`transformer_lr`. Here, `adim` is not concerned because it was unused before for transducer).

@sw005320 Is that enough changes for you? It won't really affect the size of the PR but it will address your concerns.
If it's not enough, I'll put this PR in standby until I revert changes related to naming or decide on something else.

sw005320:
Great @b-flo!
I'll check it.

Meanwhile, it would be great if @Fhrozen @pengchengguo review this PR.

pengchengguo:
> Great @b-flo!
> I'll check it.
> 
> Meanwhile, it would be great if @Fhrozen @pengchengguo review this PR.

Sorry, I had some projects these days. I will start the review tomorrow.

sw005320 (commented):
It's OK to change the files under espnet/nets/pytorch_backend/transducer/ but, could you minimize the changes in the common files?

I'll also check the documentation later.

espnet/asr/asr_utils.py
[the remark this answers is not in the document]
sw005320: Please add the difference here as a comment.
This looks very confusing.

sw005320:
@Fhrozen and @pengchengguo, please respond to @b-flo's comments.

espnet/asr/asr_utils.py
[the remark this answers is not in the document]
b-flo: It's reverted.

b-flo:
I reverted changes related to typing in `espnet/asr/pytorch_backend/asr_init.py ` and changed `label_seq` to `yseq` in related files.

> It's OK to change the styles or local variables in the files under espnet/nets/pytorch_backend/transducer/, but could you minimize the changes in the other files that are not related to the transducer implementation?

Done. The only modifications not completely related to transducer are in `espnet/asr/pytorch_backend/asr_init.py`, which change error reporting during transfer learning for all model (thoroughly neeeded).

doc/tutorial.md
```diff
@@ -195,24 +195,25 @@ minlenratio: 0.3
...
-To switch to transducer mode, the following should be set in the training config:
+ESPnet supports models trained with transducer loss, aka transducer models. To train such model, the following should be set in the training config:
 
 ```
 criterion: loss
 model-module: "espnet.nets.pytorch_backend.e2e_asr_transducer:E2E"
 ```
 
+### Architecture
+
```
sw005320, line 208: Can you emphasize that the customizable architecture is only supported in the transducer mode and why this is important for the transducer mode?

doc/tutorial.md
```diff
@@ -195,24 +195,25 @@ minlenratio: 0.3
...
 
-The architecture specification is separated for the encoder and decoder parts, and defined by the user through, respectively, `etype` and `dtype` in training config. If `custom` is specified for either, a customizable architecture will be used for the corresponding part, otherwise a RNN-based architecture will be selected.
+The architecture specification is separated for the encoder and decoder part, and defined by the user through, respectively, `etype` and `dtype` in training config. If `custom` is specified for either, a customizable architecture will be used for the corresponding part, otherwise a RNN-based architecture will be selected.
 
-While defining a RNN architecture is done in an usual manner (similarly to CTC, Att and MTL) with global parameters, a customizable architecture definition for transducer is different:
+While defining a RNN architecture is done in an usual manner (similarly to CTC, attention and MTL) with global parameters, a customizable architecture definition for transducer is different:
 1) Each blocks (or layers) for both network part should be specified individually through `enc-block-arch` or/and `dec-block-arch`:
```
sw005320, line 217:
```suggestion
1) Each block (or layer) for both network parts should be specified individually through `enc-block-arch` or/and `dec-block-arch`:
```

doc/tutorial.md
```diff
@@ -233,37 +234,41 @@ While defining a RNN architecture is done in an usual manner (similarly to CTC,
...
-          idim: input dimension
-          odim: output dimension
-          kernel_size: size of convolving kernel
-          stride: stride of the convolution
-          dilation: spacing between the kernel points
+          idim: [Input dimension. (int)]
+          odim: [Output dimension. (int)]
+          kernel_size: [Size of convolving kernel. (int)]
+          stride: [Stride of the convolution. (int)]
+          dilation: [Spacing between the kernel points. (int)]
```
sw005320, line 271: no dropout in conv1d?

doc/tutorial.md
```diff
@@ -282,47 +287,93 @@ While defining a RNN architecture is done in an usual manner (similarly to CTC,
                   att-dropout-rate: 0.4
         dec-block-repeat: 2
 
-For more information about the customizable architecture, please refer to [vivos config examples](https://github.com/espnet/espnet/tree/master/egs/vivos/asr1/conf/tuning/transducer) which cover all cases.
+For more information about the customizable architecture, please refer to [vivos config examples](https://github.com/espnet/espnet/tree/master/egs/vivos/asr1/conf/tuning/transducer) which cover most cases.
+
+### Augmented training
+
+We also supports augmented transducer model training with various auxiliary tasks, such as: CTC loss, LM loss (label-smoothing), auxiliary transducer loss and symmetric KL divergence loss.
```
sw005320, line 294:
```suggestion
We also support augmented transducer model training with various auxiliary tasks, such as: CTC loss, LM loss (label-smoothing), auxiliary transducer loss, and symmetric KL divergence loss.
```

doc/tutorial.md
```diff
@@ -282,47 +287,93 @@ While defining a RNN architecture is done in an usual manner (similarly to CTC,
...
+- There are several differences between MTL and transducer training/decoding options. The users should refer to `espnet/espnet/nets/pytorch_backend/e2e_asr_transducer.py` for an overview and `espnet/espnet/nets/pytorch_backend/transducer/arguments` for all possible arguments.
+- Memory usage during training can be reduced with minor training time augmentation using efficient encoder and decoder output combination [Li et al. (2019)](https://arxiv.org/pdf/1909.12415.pdf). To enable the feature, `joint-memory-reduction` can be set to True in training config.
 - RNN-decoder pre-initialization using a LM is supported. The LM state dict keys (`predictor.*`) will be matched to AM state dict keys (`dec.*`).
 - Transformer-decoder pre-initialization using a transformer LM is not supported yet.
-- Transformer and conformer blocks within the same architecture part (i.e: encoder) is not supported yet.
-- Customizable architecture is a in-progress work and will be eventually extended to RNN. Please report any encountered error or usage issue.
+- Mixing transformer and conformer blocks within the same architecture part (i.e.: encoder) is not supported.
```
sw005320, line 377: The following descriptions are common about the transducer and CTC/attention modes, but it looks like it is under the transducer. Please adjust the level of the header appropriately

doc/tutorial.md
```diff
@@ -195,24 +195,25 @@ minlenratio: 0.3
 - The pure attention mode requires to set the maximum and minimum hypothesis length (`--maxlenratio` and `--minlenratio`), appropriately. In general, if you have more insertion errors, you can decrease the `maxlenratio` value, while if you have more deletion errors you can increase the `minlenratio` value. Note that the optimum values depend on the ratio of the input frame and output label lengths, which is changed for each language and each BPE unit.
 - About the effectiveness of hybrid CTC/attention during training and recognition, see [2] and [3]. For example, hybrid CTC/attention is not sensitive to the above maximum and minimum hypothesis heuristics. 
 
-### Transducer
+## Transducer
```
sw005320, line 198: Transducer description should be the same header level as CTC/attention.
Please use `###`.
Also, the following headers should be `####`.

sw005320 (commented):
I checked the document and left some comments.
Can you throw the modified part to Grammarly?
There are a couple of grammatical errors or typos that Grammarly can easily fix

doc/tutorial.md
b-flo: Is that okay for you?

doc/tutorial.md
sw005320: We can keep it as an action item and will work on it in a follow-up PR.

doc/tutorial.md
sw005320: LGTM!

sw005320 (approved):
I read through your changes.
As you mentioned, the most changes are naming and typing and they look OK.
I also checked new functions and they seem to be also OK.

If it is ready, you can merge it by yourself (or I can do it for you).

espnet/nets/beam_search_transducer.py
```diff
@@ -129,71 +161,117 @@ def sort_nbest(
...
+                    logp = torch.log_softmax(
+                        self.joint_network(enc_out_t, hyp_i.dec_out[-1])
+                        / self.softmax_temperature,
+                        dim=-1,
+                    )
+
+                    curr_score = hyp_i.score + float(logp[hyp_j.yseq[pref_id]])
+
+                    for k in range(pref_id, (curr_id - 1)):
+                        logp = torch.log_softmax(
```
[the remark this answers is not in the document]
b-flo: This is on my TODO list, I'll make a PR to reach 100% coverage on Transducer-related files after the next PR (recipes update). If it's okay with you, you can merge current PR, otherwise, I'll add them later in the day.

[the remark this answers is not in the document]
sw005320: We have several internal implementations, but they are espnet2.
This would be the discussion point.
But anyway, the plan looks good to me!

sw005320:
Thanks a lot!
We still have action items, but let’s work on it.

PR #3220, opened 2021-05-16
Fix a bug of adding noise to the training data.

pengchengguo:
Hi all,

When adding noise to the training data, I found the probability of adding noise is determined by `self.rir_apply_prob`, instead of `self.noise_apply_prob`. Maybe this is a mistake.


kamo-naoyuki:
Oh, thanks! I think you are right!

Issue #3232, opened 2021-05-21
espnet2 ctc segmentation


abnerLing:
I've tried testing out the CTC segmentation example with ESPnet2 and I get the same error of "No such file or directory: 'data/nlsyms.txt'."
This occurs when using the below script and also when running the exact same python code in the example.
`espnet2/bin/asr_align.py  --asr_train_config config.yaml --asr_model_file model.pth --audio audio.wav --text text.txt --output segments`

`
Traceback (most recent call last):
  File "espnet2/bin/asr_align.py", line 827, in <module>
    main()
  File "espnet2/bin/asr_align.py", line 823, in main
    ctc_align(**kwargs)
  File "espnet2/bin/asr_align.py", line 632, in ctc_align
    aligner = CTCSegmentation(**model, **kwargs)
  File "espnet2/bin/asr_align.py", line 239, in __init__
    self.preprocess_fn = ASRTask.build_preprocess_fn(asr_train_args, False)
  File "/home/abner/work/espnet/espnet2/tasks/asr.py", line 293, in build_preprocess_fn
    retval = CommonPreprocessor(
  File "/home/abner/work/espnet/espnet2/train/preprocessor.py", line 161, in __init__
    self.tokenizer = build_tokenizer(
  File "/home/abner/work/espnet/espnet2/text/build_tokenizer.py", line 46, in build_tokenizer
    return CharTokenizer(
  File "/home/abner/work/espnet/espnet2/text/char_tokenizer.py", line 24, in __init__
    with non_linguistic_symbols.open("r", encoding="utf-8") as f:
  File "/home/abner/anaconda3/envs/es/lib/python3.8/pathlib.py", line 1222, in open
    return io.open(self, mode, buffering, encoding, errors, newline,
  File "/home/abner/anaconda3/envs/es/lib/python3.8/pathlib.py", line 1078, in _opener
    return self._accessor.open(self, flags, mode)
FileNotFoundError: [Errno 2] No such file or directory: 'data/nlsyms.txt'
`

**Basic environments:**
 - OS information: Ubuntu 20.04
 - python version: 3.8
 - espnet version: 0.9.9
 - Git hash [a9169489a5bc570ebbfe6f343aadca343dfeb08b]
 - Commit date Fri May 21 20:25:28 2021 +0900
 - pytorch version: 1.8.0

abnerLing:
Well

lumaku:
I can confirm this issue.
[...]

PR #3234, opened 2021-05-21
Fix missing nlsyms.txt for pretrained models

lumaku:
It seems that for some pretrained models, the text cleaner tries to load `non_linguistic_symbols` from a file that doesn't exist. I did  short description in #3232  and tried to solve this by catching the file open error.

@kamo-naoyuki 
You understand the tokenizer better. Could you look into it if this solution is OK?


kamo-naoyuki:
I see.
For quick workaround, it's okay, but originally we should upload nlsyms.txt at the following step:

https://github.com/espnet/espnet/blob/fe551ff9370dbb95507d74701c139743b180fa59/egs2/TEMPLATE/asr1/asr.sh#L1314-L1320

and we also add a test for ctc segmentation in model zoo: https://github.com/espnet/espnet_model_zoo/blob/master/ci/test_model.py. I'll do them.

lumaku:
Cool. Thank you!

PR #3240, opened 2021-05-26
Transducer configs & results for AISHELL-1

yusshino:
This PR adds config files & results of transducer models for AISHELL-1.

Specifically, three configurations (and their results) are provided in this PR:

- RNN-Transducer
- Conformer-Transducer
- Conformer-Transducer with auxiliary task

Note that `run.sh` was modified to disable `--api v2` if the model is transducer, because it seems transducer doesn't support the API version2; therefore ngram model (if given) is ignored in Stage 5 (decoding).

### DIFF
--- a/egs/aishell/asr1/RESULTS.md
+++ b/egs/aishell/asr1/RESULTS.md
@@ -1,3 +1,64 @@
+# Conformer-Transducer with auxiliary task (CTC weight = 0.5)
+
+## Environments
+- Same as RNN-Transducer (see below)
+
+## Config files
+- preprocess config: `conf/specaug.yaml`
+- train config: `conf/tuning/transducer/train_conformer-rnn_transducer_aux_ngpu4.yaml`
+- lm config: `-` (LM was not used)
+- decode config: `conf/tuning/transducer/decode_default.yaml`
+- ngpu: `4`
+
+## Results (CER)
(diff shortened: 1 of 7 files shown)


b-flo:
Thank you @yusshino !
If I remember correctly, we now also have a RNN-T w/ auxiliary tasks which showed improvement?

> Note that run.sh was modified to disable --api v2 if the model is transducer, because it seems transducer doesn't support the API version2; therefore ngram model (if given) is ignored in Stage 5 (decoding).

Yes, unfortunately we couldn't make an unified API v2 so a separate decoding interface was made for transducer. N-gram features was not made compatible with transducer models. May be worth adding compatibility though.

sw005320 (approved):
LGTM!
Thanks, @yusshino!
If @b-flo is OK and the CI is passed, we can merge this PR.

yusshino:
Thank you both for comments!

@b-flo
Yes, I have it. I'll add the config and results for RNN-T with an auxiliary task. (And yes, you're right, we've observed some improvement over pure RNN-T by adding an auxiliary task.)

Also thanks for explanation on the API! It's helpful to clarify my understanding. (I didn't use any LM for this experiments anyway.)

yusshino:
Added the config & results!

b-flo (approved):
LGTM! Thanks @yusshino !

PR #3249, opened 2021-05-29
fix the required number of arguments

AshrafMahdhi:
I see that the required number of arguments are 4 instead of 3


sw005320:
Thanks, @AshrafMahdhi!

AshrafMahdhi:
so welcome @sw005320!

PR #3252, opened 2021-05-31
Fix some bugs in xml2stm.py

AshrafMahdhi:
if I run xml2stm.py to generate STM and CTM files from XML file I get this error :
```
  File "xml2stm.py", line 133, in <module>
    main(args)
  File "xml2stm.py", line 98, in main
    stm(data)
  File "xml2stm.py", line 72, in stm
    "{} 1 UNKNOWN {:.02f} {:.02f} ".format(data["id"], e.startTime, e.endTime)
  File "/home/jupyter-achraf.mahdhi/.conda/envs/espnet/lib/python3.7/codecs.py", line 378, in write
    self.stream.write(data)
TypeError: write() argument must be str, not bytes
```
so I fixed this error on the following script and it works for me now!

### DIFF
--- a/egs/mgb2/asr1/local/xml2stm.py
+++ b/egs/mgb2/asr1/local/xml2stm.py
@@ -8,7 +8,6 @@
 
 __author__ = "Yifan Zhang (yzhang@qf.org.qa)"
 
-import codecs
 import sys
 from xml.dom import minidom
 
--- a/egs/mgb2/asr1/local/xml2stm.py
+++ b/egs/mgb2/asr1/local/xml2stm.py
@@ -66,30 +66,30 @@ def loadXml(xmlFileName, opts):
 
 
 def stm(data):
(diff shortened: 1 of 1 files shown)


AshrafMahdhi:
@sw005320 i think that the old script is for python 2.7 (availble in kaldi recipe) amd not for python 3

AmirHussein96:
> @sw005320 @AmirHussein96 i think that the old script is for python 2.7 (available in the Kaldi recipe) and not for python 3

Hello @AshrafMahdhi, yes you are right the old code is compatible with python 2.7. Thank you for fixing it. Let me know if you found any other script that is written in python 2.7.

AmirHussein96 (approved):

sw005320:
Thanks, @AshrafMahdhi!

AshrafMahdhi:
@AmirHussein96 So far that's all I have! I will let you know if found other bugs :) 
@sw005320 so welcome :)!

PR #3264, opened 2021-06-02
Fix typo in asr.sh (espnet2) that might cause bug

tjysdsg:
In stage 12 of [asr.sh](https://github.com/espnet/espnet/blob/master/egs2/TEMPLATE/asr1/asr.sh#L1178):

```bash
if [ "${token_type}" = pnh ]; then
    log "Error: Not implemented for token_type=phn"
    exit 1
fi
```

it should be `phn`


kamo-naoyuki:
Thanks!

PR #3278, opened 2021-06-08
Update W&B integration

AyushExel:
## This update introduces the following changes:
* All the metrics have their own sub-sections. It is easier to navigate across large number of metrics as they are automatically assigned their sections like - `train` , `valid`, 'attn` etc.
* Users can provide `--entity` argument, which is useful for building shared dashboards for teams and benchmarking
* Users can set `--model_log_interval` to log the checkpoint as W&B artifact after every `n` epochs. This fits into the larger benchmark submissions that we discussed.
 
Idea for next iteration:
**Support for dataset Logging:** We can log datasets as W&B Tables to view/query/filter them inside the browser itself! I was trying to find the most optimal way to iterate through the entire dataset inside the Trainer class itself. @sw005320 suggested that @kamo-naoyuki would be the right person to discuss this with.

### Live updated dashboards and previews

[Example updated dashboard](https://wandb.ai/cayush/EspNet_new/runs/28l14wyc?workspace=user-cayush)
[Artifacts panel](https://wandb.ai/cayush/EspNet_new/artifacts/model/model_28l14wyc/b4b67d3285db2e92c9e4)
Screenshot (113)
Screenshot (115)

cc: @sw005320


AyushExel:
I think the CI tests are failing because `use_wandb` is set to true and if API key is not set by calling `wandb login` it throws an error on initialization. I'll try to revert this and see if tests pass

espnet2/train/reporter.py
```diff
@@ -559,9 +572,23 @@ def wandb_log(self, epoch: int = None, commit: bool = True):
...
                     continue
-                d[f"{key1}_{key2}_epoch"] = self.stats[epoch][key1][key2]
+                key = f"{key1}_{key2}_epoch"
+                d[wandb_get_prefix(key) + key] = self.stats[epoch][key1][key2]
         d["epoch"] = epoch
-        wandb.log(d, commit=commit)
+        wandb.log(d)
+
+    def wandb_log_image(self, wandb_image_dict):
+        wandb.log(wandb_image_dict)
```
kamo-naoyuki, line 581: Thank you for contributing espnet!

`wandb_log_image` and `wandb_log_model_artifact` are not need to  be methods since they doesn't refer any fields of reporter.
Instead, it's okay to invoke wandb.log and wandb.log_artifact in trainer code directly.

espnet2/tasks/abs_task.py
```diff
@@ -1275,6 +1292,8 @@ def main_worker(cls, args: argparse.Namespace):
                 distributed_option=distributed_option,
             )
 
+            wandb.finish() if wandb.run else None
```
kamo-naoyuki, line 1295: Sorry, could you explain your intention of this line?  I'm not so familiar with wandb.

espnet2/train/reporter.py
```diff
@@ -559,9 +572,23 @@ def wandb_log(self, epoch: int = None, commit: bool = True):
...
+
+    def wandb_log_model_artifact(
+        self, model_path: str, aliases: List[str], metadata: Dict
+    ):
+        print("ArtifactLogger")
```
kamo-naoyuki, line 586: Could you use logging instead of print?

espnet2/train/reporter.py
AyushExel: Sure. I'll make these changes

espnet2/tasks/abs_task.py
AyushExel: `wandb.finish()` marks the run as finished.

espnet2/train/reporter.py
AyushExel: Ohh thanks for the catch. This is a typo. Will remove

espnet2/tasks/abs_task.py
kamo-naoyuki: Okay, please change it to if-block.

```python
if wandb.run:
    wandb.finish()
```

espnet2/tasks/abs_task.py
```diff
@@ -577,6 +577,24 @@ class ArgumentDefaultsRawTextHelpFormatter(
...
+            help="Specify wandb entity",
+        )
+        group.add_argument(
+            "--wandb_name",
+            type=str,
+            default=None,
+            help="Specify wandb run name",
+        )
+        group.add_argument(
+            "--model_log_interval",
```
kamo-naoyuki, line 593: Could you change --model_log_interval to --wandb_--model_log_interval? I'd like to unify these options to --wandb_*.

AyushExel: okay. makes sense. I'll change it to `wandb_model_log_interval`

espnet2/tasks/abs_task.py
AyushExel: okay

kamo-naoyuki (approved):
Thank you a lot!

AyushExel:
The CI test on my system is failing because `kenlm` is not found. I tried installing is manually but it isn't compiling

espnet2/tasks/abs_task.py
```diff
@@ -562,7 +562,7 @@ class ArgumentDefaultsRawTextHelpFormatter(
         group.add_argument(
             "--use_wandb",
             type=str2bool,
-            default=False,
+            default=True,
```
kamo-naoyuki, line 565: Sorry, I didn't realize the default is True.
I'll change this to False because wandb causes an exception when no http connections exist.

kamo-naoyuki: Sorry, my bad. It seems to work, so there are no problems.

kamo-naoyuki: Hmm, sorry, after all, it doesn't seem to work when there are  http connection error, but the case must work for use case.
Is it correct? If so, I'll change the default to False.

AyushExel: Yes wandb will print http error if it is not able to connect but that will not stop the training. If you decide to set `use_wandb = False` by default, please include this line in the readme so that users know how to enable it.
```
Once installed, run `wandb login` to enable tracking runs using W&B. To enable W&B tracking, set `use_wandb` to True
```

kamo-naoyuki: In my environment, it tries to connect, probably until some timeout? and it doesn't raise error.  I'll change the default to False.

Issue #3237, opened 2021-05-24
The grad norm is nan when set use_amp=True


jinggaizi:
i try to use mix precision to training and set use_amp: true in conf/train_asr_conformer.yaml in Aishell recipe .
however , it's just print " WARNING: The grad norm is nan. Skipping updating the model" in the log
has anyone  trained？

andrusenkoau:
I have the same problem...(

andrusenkoau:
Сan this be due to the fact that the calculation of the loss function is located in "with autocast()"? If we got big loss value on initial step, this value will out the range of float16. It is just my assumption.

sw005320:
Thanks for discussing the issue.
We actually tried it but at least I had the same issue, and I could not find a solution.
I really appreciate your comment about it.
If someone also gives us some suggestions about it, that is great.

jaesong:
The NaN may be a non-issue or an issue depending on how frequent it occurs.

Gradient scaling (`loss <- loss * scale`) is used to prevent overflow/underflow. But it requires finding a good scale. Thus, `torch.cuda.amp.GradScaler` initially sets scale to some good number (`2**16`), and increases it (if there has been no NaNs for certain iterations, e.g. 2000), or decreases it (if NaN is found). Thus, NaN occasionally occurs as GradScaler tries to grow the scale.

However, if you see excessive NaN logs, the network is suffering numerical issues, especially during forward pass.
fp16 has limited range (approximately -65000 ~ 65000) and the model output grows as the training goes.
Note that anomaly detection may not detect the issue correctly if NaN is already in forward values.
Weight decay may help to prevent output growth. I've heard using a smaller LR is also helpful but I haven't tried it.

It is good to monitor `scaler.get_scale()`. If the scale becomes 0.0 due to consecutive NaNs, the training should be terminated.

sw005320:
#3282 may solve the issue.
Can someone test it?
I'm in an ICASSP mode...

andrusenkoau:
Great, #3282 works for me. I got 25% training speed up due to the usage of "use_amp=true". Thank you very much for the quick solution.

PR #3282, opened 2021-06-09
Fix a bug in trainer.py

sw005320:
https://github.com/espnet/espnet/issues/3237 can be fixed by this bug.
Thanks, Jeff Farris for pointing out the bug.


kamo-naoyuki:
Oh, this is a fatal bug from 460d068. 
Your code is incorrect. This is not a reason of #3237. 
I'll fix quickly.

Issue #3170, opened 2021-04-24
The grad norm is nan in ESPnet2 ASR task


lin-nerd:
I use ESPnet2 in ASR task, config is conf/train_asr_conformer.yaml

- python version: `3.8.5 (default, Sep  4 2020, 07:30:14)  [GCC 7.3.0]`
- espnet version: `espnet 0.9.9`
- pytorch version: `pytorch 1.4.0`

log is:

[user] 2021-04-23 19:22:52,420 (trainer:581) WARNING: The grad norm is nan. Skipping updating the model.
[user] 2021-04-23 19:22:53,113 (trainer:581) WARNING: The grad norm is nan. Skipping updating the model.
[user] 2021-04-23 19:22:53,694 (trainer:581) WARNING: The grad norm is nan. Skipping updating the model.
[user] 2021-04-23 19:22:54,316 (trainer:581) WARNING: The grad norm is nan. Skipping updating the model.
[user] 2021-04-23 19:22:54,997 (trainer:581) WARNING: The grad norm is nan. Skipping updating the model.
[user] 2021-04-23 19:22:55,597 (trainer:581) WARNING: The grad norm is nan. Skipping updating the model.
[user] 2021-04-23 19:22:56,229 (trainer:581) WARNING: The grad norm is nan. Skipping updating the model.
[user] 2021-04-23 19:22:56,245 (trainer:636) INFO: 2epoch:train:12009-12640batch: iter_time=3.163e-04, forward_time=0.110, loss=65.129, loss_att=124.079, loss_ctc=144.685, acc=0.065, backward_time=0.086, optim0_lr0=1.485e-05, train_time=0.624
[user] 2021-04-23 19:22:56,858 (trainer:581) WARNING: The grad norm is nan. Skipping updating the model.
[user] 2021-04-23 19:22:57,476 (trainer:581) WARNING: The grad norm is nan. Skipping updating the model.
[...]

PR #3283, opened 2021-06-10
Bug fix of accum_grad when grad-nan

kamo-naoyuki:
Now optimizer.zero_grad() is not invoked when the gradient is nan and accum_grad>1.
This is a fatal bug of #3014 and probably causes different result from previous version... Maybe #3170 also comes from here.

### DIFF
--- a/espnet2/train/trainer.py
+++ b/espnet2/train/trainer.py
@@ -634,7 +634,10 @@ class Trainer:
                                 optimizer.step()
                             if isinstance(scheduler, AbsBatchStepScheduler):
                                 scheduler.step()
-                            optimizer.zero_grad()
+                for iopt, optimizer in enumerate(optimizers):
+                    if optim_idx is not None and iopt != optim_idx:
+                        continue
+                    optimizer.zero_grad()
 
                 # Register lr and train/load time[sec/step],
                 # where step refers to accum_grad * mini-batch


shincling:
Hi Kamo, I got one more question about the behavior of the Grad_nan related problem.  

If the grad-nan happens and there is accum_grad>1 in config, 
then the skip updating behavior will **_skip the specific batch (where nan happans)_** or **_the whole set of batches in this round of accum_grad_**?

I checked the trainer.py, but I'm not fully clear about this.

kamo-naoyuki:
Hi, now, we selected `the whole set of batches in this round of accum_grad` because it is more similar behaviour to the real batch size without `accum_grad`.

Perhaps `skip the specific batch (where nan happans` is better, but actually, I don't know how to  implement it.

```
# iter1
loss.backward()  # grad = grad_iter1
# iter2
loss.backward()  # grad = grad_iter1 + grad_iter2
# iter3
loss.backward()  # grad = grad_iter1 + grad_iter2 + grad_iter3
...
```

Every time `loss.backward()`, gradients are accumulated, but I don't know how to get each gradient of each iteration before accumulated using PyTorch API, except for the first iteration.

shincling:
> Hi, now, we selected `the whole set of batches in this round of accum_grad` because it is more similar behaviour to the real batch size without `accum_grad`.
> 
> Perhaps `skip the specific batch (where nan happans` is better, but actually, I don't know how to implement it.
> 
> ```
> # iter1
> loss.backward()  # grad = grad_iter1
> # iter2
> loss.backward()  # grad = grad_iter1 + grad_iter2
> # iter3
> loss.backward()  # grad = grad_iter1 + grad_iter2 + grad_iter3
> ...
> ```
> 
> Every time `loss.backward()`, gradients are accumulated, but I don't know how to get each gradient of each iteration before accumulated using PyTorch API, except for the first iteration.

Thanks, Kamo. That's quite helpful.

PR #3285, opened 2021-06-10
cache frequently accessed file handlers during data splitting to speed it up

lizezheng:
cache frequently accessed file handlers during data splitting to speed it up

### DIFF
--- a/espnet2/bin/split_scps.py
+++ b/espnet2/bin/split_scps.py
@@ -1,3 +1,4 @@
+#!/usr/bin/env python3
 import argparse
 from collections import Counter
 from itertools import zip_longest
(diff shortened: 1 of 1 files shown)


kamo-naoyuki:
This is one of the my todos, thank you!
 
 I think it's better to have an option to give the upper limit of cache file descriptors, maybe 100-300 is reasonable.

lizezheng:
> This is one of the my todos, thank you!
> 
> I think it's better to have an option to give the upper limit of cache file descriptors, maybe 100-300 is reasonable.

yeah, a upper limit is preferred, cause we need to take care of the maximum file handlers system allows

Issue #3290, opened 2021-06-15
espnet2 aishell recipe


200987299:
Dear developers, I am tring to write a script for aishell2 from aishell1 script, but when it comes to training process, I always get NaN error. (grad is NaN, skiping update model). It happened for all transformer recipes in the conf/tuning/ folder. Does anyone have any  idea how to solve this problem?

thanks!

sw005320:
How often does it happen?
If it does not happen so often, we can ignore it.

200987299:
> How often does it happen?
> If it does not happen so often, we can ignore it.

Dear Professor Watanabe,

It happens every time for all transformer recipes. I am tuning the learning rate hoping to get a better result. I am also trying the rnn recipes, let you know later.

Bests
D

sw005320:
It should not happen so often. I’m curious about your config files.
Maybe, it is a good time to make an aishell2 recipe in espnet2.
Is it OK for you to make a PR about your current one, then do some debugging together through PR? 
Or I may ask some of my colleagues to do it.
(@pengchengguo and @yuekaizhang, I remember that either of you touched aishell2. Do you think you can make an espnet2 recipe?)

200987299:
> It should not happen so often. I’m curious about your config files.
> Maybe, it is a good time to make an aishell2 recipe in espnet2.
> Is it OK for you to make a PR about your current one, then do some debugging together through PR?
> Or I may ask some of my colleagues to do it.
> (@pengchengguo and @yuekaizhang, I remember that either of you touched aishell2. Do you think you can make an espnet2 recipe?)

I would love to participate. I copied 99% of the code from espnet2/aishell. I am starting to think that the error happened because I didn't filter out un-clean data from the dataset. I will post my code a bit later after I check it again. I am also love to work with @pengchengguo and @yuekaizhang, thanks for the help and input.

sw005320:
Very cool.
Actually, debugging through PR or issues would work.
So, if you’re ready, please make a PR.
We may find an issue quickly.

kamo-naoyuki:
Recently we found a fatal bug, #3283, and this bug was existing for a long time.  This is my fault, sorry.  If you didn't update espnet, please do it, and I recommend you using `--ctc_conf ignore_nan_grad= true`

200987299:
[...]

PR #3292, opened 2021-06-16
Update ctc.py

200987299:
setting the --ctc_conf ignore_nan_grad= true like @pengchengguo and @sw005320 suggested at #3290


sw005320:
Thanks a lot, @200987299!

Issue #3300, opened 2021-06-18
Upgrade pytorch version to 1.9.0 and drop old ones (< 1.3)


ShigekiKarita:
Now we support too many pytorch versions 1.0.1 - 1.8.1. I think it is high time to drop old ones when we support new one, 1.9.0. I would like to discuss which versions to remove. 

At this moment, I and @sw005320 consider dynamic quantization (#3210) is an impactful feature. To make it enabled by default, we should drop the incompatible versions < 1.3. https://github.com/pytorch/pytorch/releases/tag/v1.3.0

TODO: volunteers are welcomed
- [x] update CI to add 1.9.0 and drop ver < 1.3 (#3299)
- [x] clean up espnet1 code dispatching old pytorch < 1.3.0 (#3301)
- [x] clean up espnet2 code dispatching old pytorch < 1.3.0 (#3303)
- [x] major update espnet to **v0.10.0**

```
$ cd espnet; grep -r version
lm/chainer_backend/lm.py:    # display chainer version
lm/chainer_backend/lm.py:    logging.info("chainer version = " + chainer.__version__)
lm/lm_utils.py:        # Floating point version of epoch.
lm/lm_utils.py:            # guess previous_epoch_detail for older version
lm/pytorch_backend/lm.py:    # display torch version
lm/pytorch_backend/lm.py:    logging.info("torch version = " + torch.__version__)
asr/chainer_backend/asr.py:    # display chainer version
asr/chainer_backend/asr.py:    logging.info("chainer version = " + chainer.__version__)
asr/chainer_backend/asr.py:    # display chainer version
asr/chainer_backend/asr.py:    logging.info("chainer version = " + chainer.__version__)
asr/pytorch_backend/asr.py:if sys.version_info[0] == 2:
asr/pytorch_backend/recog.py:        # NOTE: for a compatibility with less than 0.5.0 version models
vc/pytorch_backend/vc.py:        # perform padding and conversion to tensor
tts/pytorch_backend/tts.py:        # perform padding and conversion to tensor
st/pytorch_backend/st.py:if sys.version_info[0] == 2:
nets/pytorch_backend/rnn/decoders.py:from distutils.version import LooseVersion
nets/pytorch_backend/rnn/decoders.py:        if LooseVersion(torch.__version__) < LooseVersion("1.0"):
nets/pytorch_backend/rnn/attentions.py:    This attention is an extended version of location aware attention.
[...]

PR #3303, opened 2021-06-19
Delete codes for old pytorch in espnet2.

ShigekiKarita:
See #3300


kamo-naoyuki (approved):
Thank you @ShigekiKarita

PR #3306, opened 2021-06-20
Update version to 0.10.0

ShigekiKarita:
See #3300


sw005320 (approved):
Great!
@kan-bayashi, can you also make a release note?

PR #3311, opened 2021-06-21
Recipe of JTubeSpeech

sw005320:
We are making a Japanese YouTube Speech corpus for ASR, SV, and TTS.

TODO and NOTES
- For now, we need to specify the speech and corresponding segmentation in `local/data.sh`. This setting will be moved to `db.sh`, once we fix the database.
- Currently, we used a CTC threshold -0.3 to prune the wrong files
- We split the valid and training sets, but probably, we should also make an eval set.
- The current validation CER is 16.5%, which is not bad, but we'll change the validation set 
- The current scoring considers the space, and we need to remove them from scoring, but I don't have a good idea of how to do it.
```
id: (0ly1drmmvsc-0ly1drmmvsc_0047)
Scores: (#C #S #D #I) 24 4 4 1
REF:  や ア イ ン シ ュ タ イ ン と い っ た 著 名 人 <space> に も 接 触 し ******* 神 に 対 す る 自 身 の 理 解
HYP:  や *** *** *** *** 愛 し た い と い っ た 著 名 人 <space> に も 接 触 し <SPACE> 神 に 対 す る 自 身 の 理 解
Eval:     D   D   D   D   S   S   S   S                                                           I
```


Alex-Songs:
@sw005320 hello，where can I get the list of filtered jtubespeech training sets? Thank you.

PR #3316, opened 2021-06-23
Allow override default 'speed_control_alpha' parameter

### LINKED ISSUE #3256: Speed alpha parameter in sythesize method for FastSpeech2 TTS
**Now:** 

the *speed_control_alpha* parameter is configurable in Text2Speech class constructor. https://github.com/espnet/espnet/blob/39032a3011ecaa7b4fc0bc60979e3933835d2aea/espnet2/bin/tts_inference.py#L63

---

**Problem**: 

I'm successfully serving FastSpeech2 model using the Text2Speech class. The model is loaded into GPU memory once and serving many incoming web requests. I need implement a possibility to slower or speed up a speech. The current implementation would require recreate Text2Speech with different *speed_control_alpha*.

---

**Suggestion**: 

Add *speed_control_alpha* into the synthesize method: https://github.com/espnet/espnet/blob/39032a3011ecaa7b4fc0bc60979e3933835d2aea/espnet2/bin/tts_inference.py#L127 
If not provided here it could use the default one from the constructor, otherwise use the provided one.

---

I could provide a pull-request if you see a suggestion useful.

airenas:
It adds a possibility to synthesize speech with various speed values without reloading a model.

Resolves #3256.

### DIFF
--- a/espnet2/bin/tts_inference.py
+++ b/espnet2/bin/tts_inference.py
@@ -130,7 +130,7 @@ class Text2Speech:
         speech: Union[torch.Tensor, np.ndarray] = None,
         durations: Union[torch.Tensor, np.ndarray] = None,
         spembs: Union[torch.Tensor, np.ndarray] = None,
-        speed_control_alpha: float = None,
+        speed_control_alpha: Optional[float] = None,
     ):
         assert check_argument_types()
 
--- a/espnet2/bin/tts_inference.py
+++ b/espnet2/bin/tts_inference.py
@@ -149,14 +149,14 @@ class Text2Speech:
             batch["spembs"] = spembs
 
         cfg = self.decode_config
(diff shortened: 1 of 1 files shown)


espnet2/bin/tts_inference.py
```diff
@@ -130,6 +130,7 @@ def __call__(
         speech: Union[torch.Tensor, np.ndarray] = None,
         durations: Union[torch.Tensor, np.ndarray] = None,
         spembs: Union[torch.Tensor, np.ndarray] = None,
+        speed_control_alpha: float = None,
```
kan-bayashi, line 133:
```suggestion
        speed_control_alpha: Optional[float] = None,
```

kan-bayashi (commented):
Thank you for your good suggestion!

PR #3319, opened 2021-06-24
fix a bug in OneCycleLR and CyclicLR

sw005320:
OneCycleLR and CyclicLR schedulers should be per iteration, not per epoch.
Thanks, Jeff Farris for pointing out the issue!


kamo-naoyuki:
Hmm, this bug comes from #3303. I didn't noticed.
CosineAnnealingWarmRestarts is also per iteration.

ShigekiKarita:
thanks for fixing my bug!

PR #3324, opened 2021-06-26
Fix a type error of swbd data preparation.

pengchengguo:
Fix a type error of SWBD data preparation and also change the command `find` to `find -L` in case the corpus directory is a soft link.


sw005320:
Thanks, @pengchengguo!

Issue #3337, opened 2021-07-06
Commonvoice fails due to issue in Pillow 8.3.0


blokje:
**Describe the bug**
Pillow 8.3.0 introduces and issue which affect (at least) the commonvoice egs

During some investigation I found the following github issue with the same problem and suggested solution. It seems that Pillow 8.3.0 is triggering this issue.

- https://github.com/tensorflow/tensorflow/issues/46840
- https://github.com/python-pillow/Pillow/issues/5571

Installing any other version than 8.3.0 solved the problem

`pip install "pillow!=8.3.0"`

**Basic environments:**
 - OS information: `Linux gpu03 5.4.0-1051-azure #53~18.04.1-Ubuntu SMP Fri Jun 18 22:32:58 UTC 2021 x86_64 x86_64 x86_64 GNU/Linux`
- python version: `3.6.9 (default, Jan 26 2021, 15:33:00)  [GCC 8.4.0]`
- espnet version: `espnet 0.10.0`
- pytorch version: `pytorch 1.8.0`

**Environments from `torch.utils.collect_env`:**
Collecting environment information...
PyTorch version: 1.8.0
Is debug build: False
CUDA used to build PyTorch: 10.2
ROCM used to build PyTorch: N/A

OS: Ubuntu 18.04.5 LTS (x86_64)
GCC version: (Ubuntu 7.5.0-3ubuntu1~18.04) 7.5.0
Clang version: Could not collect
CMake version: version 3.10.2

Python version: 3.6 (64-bit runtime)
Is CUDA available: True
CUDA runtime version: 9.1.85
GPU models and configuration: GPU 0: Tesla K80
Nvidia driver version: 470.42.01
cuDNN version: Probably one of the following:
/usr/lib/x86_64-linux-gnu/libcudnn.so.8.2.1
/usr/lib/x86_64-linux-gnu/libcudnn_adv_infer.so.8.2.1
/usr/lib/x86_64-linux-gnu/libcudnn_adv_train.so.8.2.1
/usr/lib/x86_64-linux-gnu/libcudnn_cnn_infer.so.8.2.1
/usr/lib/x86_64-linux-gnu/libcudnn_cnn_train.so.8.2.1
/usr/lib/x86_64-linux-gnu/libcudnn_ops_infer.so.8.2.1
/usr/lib/x86_64-linux-gnu/libcudnn_ops_train.so.8.2.1
HIP runtime version: N/A
MIOpen runtime version: N/A

[...]

PR #3338, opened 2021-07-06
avoid to use Pillow 8.3.0

sw005320:
pointed by @blokje in #3337


sw005320:
It seems that the issue is fixed in 8.3.1 and now the default installation becomes normal.
I'll close this issue.

PR #3349, opened 2021-07-13
fix path error in local/score.sh in swbd

wonkyuml:
this should fix path error when scoring ctm


sw005320:
Thanks, @wonkyuml!
@roshansh-cmu, could you check this?

roshansh-cmu (approved):
Thanks for the fix @wonkyuml
LGTM- good to merge

PR #3355, opened 2021-07-15
Add PyTorch Builtin Complex Support in the Speech Enhancement Task

Emrys365:
If the PyTorch version is larger than or equal to 1.8.0, we can use either torch_complex or the builtin complex implementation, by setting the `use_builtin_complex` flag in `espnet2/enh/encoder/stft_encoder.py`.

> BTW, the [pytorch_wpe](https://pypi.org/project/pytorch-wpe/) package should also be updated by someone to be compatible to the builtin complex implementation in PyTorch 1.8.0+.


espnet2/enh/decoder/stft_decoder.py
```diff
@@ -1,9 +1,12 @@
+from distutils.version import LooseVersion
 import torch
 from torch_complex.tensor import ComplexTensor
 
 from espnet2.enh.decoder.abs_decoder import AbsDecoder
 from espnet2.layers.stft import Stft
 
+is_torch_1_8_plus = LooseVersion(torch.__version__) >= LooseVersion("1.8.0")
```
mthrok, line 8: Looking at the number of times this pattern shows up in the PR, putting this into some utility module and reusing across the implementations would give a better maintenance experience.

For example, I would recommend to use 1.9 as switch instead of 1.8, but with the current approach, we need to replace it repeatedly.

espnet2/enh/layers/beamformer.py
```diff
@@ -34,39 +102,50 @@ def get_rtf(
...
+            noise covariance matrix (..., F, C, C)
         reference_vector (torch.Tensor or int): (..., C) or scalar
         iterations (int): number of iterations in power method
         use_torch_solver (bool): Whether to use `solve` instead of `inverse`
     Returns:
-        rtf (ComplexTensor): (..., F, C, 1)
+        rtf (torch.complex64/ComplexTensor): (..., F, C, 1)
     """
+    if isinstance(psd_speech, ComplexTensor):
+        complex_wrapper = FC
```
mthrok, line 116: I think calling this `warpper` is a bit confusing as it is also used in `espnet2/enh/espnet_model.py` to note something different.

I would use something like `module` or `impl`.

espnet2/enh/layers/beamformer.py
```diff
@@ -9,19 +10,86 @@
 from torch_complex.tensor import ComplexTensor
 
 EPS = torch.finfo(torch.double).eps
+is_torch_1_8_plus = LooseVersion(torch.__version__) >= LooseVersion("1.8.0")
 
 
-def complex_norm(c: ComplexTensor) -> torch.Tensor:
+def is_complex(c):
```
mthrok, line 16: I think these functions can live in a utility module with `is_torch_1_8_plus` logic.

mthrok (commented):
I had a quick review. I see many occurrences of the following pattern and wonder we can use some generic version and reuse the logic. At a glance the expressions in `if~else~` are different so not sure of a simple way. 🧐

```python
        if isinstance(data, ComplexTensor):
            ...
        elif is_torch_1_8_plus and torch.is_complex(data):
            ...
        else:
            raise ValueError(...)
```

espnet2/enh/layers/beamformer.py
Emrys365: Thanks, I have renamed them to keep consistency.

espnet2/enh/layers/beamformer.py
Emrys365: Thanks, I have added the condition and move the related functions to a new script `complex_utils.py`.

espnet2/enh/espnet_model.py
```diff
@@ -7,17 +8,19 @@
...
 from espnet2.enh.decoder.abs_decoder import AbsDecoder
 from espnet2.enh.encoder.abs_encoder import AbsEncoder
 from espnet2.enh.encoder.conv_encoder import ConvEncoder
+from espnet2.enh.layers.complex_utils import is_complex
 from espnet2.enh.separator.abs_separator import AbsSeparator
 from espnet2.torch_utils.device_funcs import force_gatherable
 from espnet2.train.abs_espnet_model import AbsESPnetModel
 
 
+is_torch_1_3_plus = LooseVersion(torch.__version__) >= LooseVersion("1.3.0")
```
ShigekiKarita, line 22: Don't worry about this because we already dropped pytorch < 1.3.0. https://github.com/espnet/espnet/releases/tag/v.0.10.0

Emrys365: Thanks, I didn't notice this. I will remove these checks then.

Emrys365:
BTW, some of the CI tests fails because the pytorch_wpe code still does not support the PyTorch builtin complex implementation. Where can I make a PR to update it?

sw005320:
How about asking torchaudio to support WPE, in addition to maintain our own WPE?
It is very good to have our own WPE implementations under espnet and we should keep this direction, but the reference implementation of WPE by torchaudio helps us a lot.

This will also make pytorch complex implementations more solid, since the use and stability of them can be well checked by torch and torchaudio teams.
We can safely use them then.

@mthrok, what do you think?

nateanl:
That's a good idea. I'm planning to add [MVDR and GEV beamformer](https://github.com/pytorch/audio/issues/1656) to torchaudio. WPE or WPD beamformer is also good to add! If it's okay, I can work with Wangyou together on this since he has rich experiences on the beamforming methods 🙂@Emrys365

Emrys365:
> That's a good idea. I'm planning to add [MVDR and GEV beamformer](https://github.com/pytorch/audio/issues/1656) to torchaudio. WPE or WPD beamformer is also good to add! If it's okay, I can work with Wangyou together on this since he has rich experiences on the beamforming methods 🙂@Emrys365 

Sure! I would be happy to help.

test/espnet2/enh/test_espnet_model.py
```diff
@@ -15,13 +15,20 @@
 from espnet2.enh.separator.transformer_separator import TransformerSeparator
 
 is_torch_1_2_plus = LooseVersion(torch.__version__) >= LooseVersion("1.2.0")
```
sw005320, line 17: You can drop this. 
We don't support torch 1.3.1>

sw005320 (commented):
LGTM.
I have a couple of general comments for the test.
If it is difficult or taking a lot of time in the test, you can ignore them.

- could you try to increase the coverage of the unit test more? There are some cases that the test is not covered, e.g., `use_torch_solver = False` etc.
- is there a unit test to check that the built-in complex and torch_complex get a similar result? This will make this PR safe.

test/espnet2/enh/test_espnet_model.py
Emrys365: OK, I'll remove it.

sw005320:
Sorry for pending it for a while.
The PR looks very good to me.
If you're ready, I can merge this PR (or you can do it by yourself!)

Emrys365:
> Sorry for pending it for a while.
> The PR looks very good to me.
> If you're ready, I can merge this PR (or you can do it by yourself!)

Sure, let me merge it.

mthrok:
> How about asking torchaudio to support WPE, in addition to maintain our own WPE?
> It is very good to have our own WPE implementations under espnet and we should keep this direction, but the reference implementation of WPE by torchaudio helps us a lot.
> 
> This will also make pytorch complex implementations more solid, since the use and stability of them can be well checked by torch and torchaudio teams.
> We can safely use them then.
> 
> @mthrok, what do you think?

@sw005320 @nateanl sorry I forgot to respond, but yes, we can add it in torchaudio.

PR #3389, opened 2021-07-30
model_summary'output

triper1022:
I am wondering whether the last line(70) uses print() is more straightforward.


sw005320:
Thanks for your PR.
I think keeping the message as an object is good since we can call it from `logging.info` and so on.
We can keep it as it is.

PR #3394, opened 2021-08-01
Add JMD recipe

takenori-y:
I added JMD TTS recipe and tested it on Kumamoto dialect. The input of Tacotron2 is phoneme. It works well. I tried to use Hiragana as the input, but it does not work. This may be due to the lack of training data (1300 sentences).

I will make segments file for Osaka dialect in a few days.


kan-bayashi:
Thanks! Could you add egs2/jmd/tts1/README.md like egs2/jsut/tts1?
Adding the result part can be later.

takenori-y:
Thank you for your response. I added RADME.md file.

kan-bayashi:
I forgot to mention about `local/path.sh`.
Could you copy https://github.com/espnet/espnet/blob/master/egs2/jsut/tts1/local/path.sh ?

takenori-y:
`egs2/jsss/tts1/local/path.sh` file is also empty, so I updated it. Is it ok?

kan-bayashi:
Oh thanks! That is fine!

kan-bayashi (approved):
LGTM

PR #3395, opened 2021-08-01
Add initial Switchboard Dialogue Act classification recipe

akreal:
The numbers of data samples here are slightly lower than in our paper (93742/8143/2379 vs. 97756/8591/2507) because we used already prepared data and there were no data preparation script. Hopefully the new data preparation script from this recipe can serve as a more accessible baseline.


ftshijt (commented):
Please also update `db.sh` with SWBD_DA.

btw, do we really need to host the train-dev-test split in the repo, can we get it in the ldc or somewhere else?

akreal:
> Please also update `db.sh` with SWBD_DA.

Will do.

> btw, do we really need to host the train-dev-test split in the repo, can we get it in the ldc or somewhere else?

The original source for the splits is this GitHub repo: https://github.com/Franck-Dernoncourt/naacl2016
I copied the splits in case that repo will move or disappear, but maybe it's fine to fetch them from there in `data.sh`. What do you think?

sw005320:
> > btw, do we really need to host the train-dev-test split in the repo, can we get it in the ldc or somewhere else?
> 
> The original source for the splits is this GitHub repo: https://github.com/Franck-Dernoncourt/naacl2016
> I copied the splits in case that repo will move or disappear, but maybe it's fine to fetch them from there in `data.sh`. What do you think?

@ftshijt, this is a good point, but I'm OK with @akreal's solution, as it is easier and makes our repo more self-consistent.

ftshijt:
> > > btw, do we really need to host the train-dev-test split in the repo, can we get it in the ldc or somewhere else?
> > 
> > 
> > The original source for the splits is this GitHub repo: https://github.com/Franck-Dernoncourt/naacl2016
> > I copied the splits in case that repo will move or disappear, but maybe it's fine to fetch them from there in `data.sh`. What do you think?
> 
> @ftshijt, this is a good point, but I'm OK with @akreal's solution, as it is easier and makes our repo more self-consistent.

Cool! Then I think current PR is ready to merge. Thanks for your effort @akreal !

akreal:
Thanks for your comments, @ftshijt and @sw005320 !
I have not yet added `SWBD_DA` to `db.sh`, but I'll do it in the follow up PR (currently `SWBD_DA` is not used).

PR #3410, opened 2021-08-05
Add NLU post-encoder using Hugging Face Transformers

akreal:
Please let me know what you think about the naming and overall API design.

Currently I've tried it with BERT only, other models may require special handling depending on differences in their internal APIs.


sw005320:
@simpleoier, could you review it?
Especially, I want your experience of using the external model (w2v2 in your case) in espnet2 for this review.

espnet2/asr/postencoder/transformers.py
```diff
@@ -0,0 +1,57 @@
...
+            from transformers import AutoModel
+        except Exception as e:
+            print("Error: Transformers is not properly installed.")
+            print(
+                "Please install Transformers: cd ${MAIN_ROOT}/tools && make transformers.done"
+            )
+            raise e
+
+        model = AutoModel.from_pretrained(model_name_or_path)
+        self.transformer = model.encoder
```
simpleoier, line 37: Did you make sure that the parameters of `self.transformer` wouldn't be messed up by the [initialization](https://github.com/espnet/espnet/blob/2a9e8ad0b95fe0bababa88dec53fc8ce6ba55059/espnet2/tasks/asr.py#L456)?

simpleoier (commented):
This PR looks good to me in general. I currently have two questions:
1. Did you make sure that the initialization step won't mess up with the parameters of pretrained models?
2. Can you also fix the format of results in this [README](https://github.com/espnet/espnet/tree/master/egs2/swbd_da/asr1)?

espnet2/asr/postencoder/transformers.py
akreal: They were, fixed now.

akreal:
Thank you for the review!

> Did you make sure that the initialization step won't mess up with the parameters of pretrained models?

Thanks for pointing this out, the parameters actually were overwritten when `init` is set. It is fixed now.

> Can you also fix the format of results in this README?

Fixed.

tools/installers/install_transformers.sh
```diff
@@ -0,0 +1,19 @@
...
+if [ $# != 0 ]; then
+    echo "Usage: $0"
+    exit 1;
+fi
+
+# Install Hugging Face Transformers
+if [ ! -e transformers.done ]; then
+    (
+        set -euo pipefail
+        python3 -m pip install transformers
```
simpleoier, line 14: Maybe it is better to put the version info of transformers here. What do you think?

egs2/swbd_da/asr1/conf/tuning/train_asr_transformer_postencoder.yaml
```diff
@@ -0,0 +1,34 @@
...
+    dropout_rate: 0.1
+    positional_dropout_rate: 0.1
+    attention_dropout_rate: 0.0
+    input_layer: conv2d # encoder architecture type
+    normalize_before: true
+
+# postencoder
+postencoder: transformers
+postencoder_conf:
+    model_name_or_path: "bert-base-cased"
```
sw005320, line 18: What other models can we use?
Can we basically pick up a model from https://huggingface.co/models?filter=transformers?
If so, it would be better to add such information as a comment here.

egs2/swbd_da/asr1/conf/tuning/train_asr_transformer_postencoder.yaml
```diff
@@ -0,0 +1,34 @@
...
+decoder: transformer
+decoder_conf:
+    attention_heads: 4
+    linear_units: 2048
+    num_blocks: 6
+    dropout_rate: 0.1
+    positional_dropout_rate: 0.1
+    self_attention_dropout_rate: 0.0
+    src_attention_dropout_rate: 0.0
+
```
siddhu001, line 30: I would recommend using Adam optimiser instead of default Adagrad with learning rate of the order of 1e-5. Also warmup scheduler will be very helpful for effective finetuning.

tools/installers/install_transformers.sh
akreal: Any fairly recent `transformers` should work, because the code relies on the `AutoModel` class that was introduced about [2 years ago](https://github.com/huggingface/transformers/releases/tag/1.1.0). Although I'll add a comment that it's tested with 4.9.1.

egs2/swbd_da/asr1/conf/tuning/train_asr_transformer_postencoder.yaml
akreal: I've tried more models and added a comment about it. However, it required some special handling, so maybe this PR needs another review.

tools/installers/install_transformers.sh
akreal: Now after adding special handing for some models that relies on different configuration parameters probably it's a good idea. So I added `>4.9.0`.

egs2/swbd_da/asr1/conf/tuning/train_asr_transformer_postencoder.yaml
akreal: Thanks, I'll make another PR with better configurations. This configuration file is just to show how to enable NLU.

Alex-Songs:
Very nice work, is there a paper describing your post-encoder? thanks! @akreal

akreal:
> Very nice work, is there a paper describing your post-encoder?

Thank you! There is no paper for this particular configuration yet, but it is very similar to this older paper: http://www.interspeech2020.org/uploadfile/pdf/Mon-2-9-4.pdf. The main difference is that it's normal supervised training now, instead of Teacher-Student learning.

akreal:
@sw005320 @simpleoier, this PR should be ready for the review and merge.

tools/Makefile
```diff
@@ -57,10 +57,10 @@ all: kaldi showenv python conda_packages.done sctk.done sph2pipe.done check_inst
 
 ifneq ($(strip $(CHAINER_VERSION)),)
 python: activate_python.sh espnet.done pytorch.done chainer.done
-extra: warp-ctc.done warp-transducer.done chainer_ctc.done nkf.done moses.done mwerSegmenter.done pesq kenlm.done pyopenjtalk.done py3mmseg.done beamformit.done fairseq.done s3prl.done k2.done
+extra: warp-ctc.done warp-transducer.done chainer_ctc.done nkf.done moses.done mwerSegmenter.done pesq kenlm.done pyopenjtalk.done py3mmseg.done beamformit.done fairseq.done s3prl.done k2.done transformers.done
```
sw005320, line 60: `transformers` sounds a bit too general for me.
hf_transformers?
Please come up with a better name.

sw005320 (commented):
Many thanks, @akreal!!!
I have several requests.
- Can you add a unit test to cover postencoder/transformers.py like https://github.com/espnet/espnet/tree/master/test/espnet2/asr/preencoder?
- I see that you put the pre-trained models after the encoder (post-encoder). Does this mean that you perform NLU processing based on CTC? Another possibility is to put it after the decoder (post-decoder) and perform NLU processing based on the encoder-decoder. The current option seems to be fine, but I just want to confirm it.

Alex-Songs:
Thanks! @akreal

akreal:
> Can you add a unit test to cover postencoder/transformers.py

Done.

>I see that you put the pre-trained models after the encoder (post-encoder). Does this mean that you perform NLU processing based on CTC?

Post-encoder takes hidden representations from the encoder and outputs hidden representations that can be taken by either CTC or decoder. So the network looks like this:

```
Encoder-->Post-encoder--->Decoder
                       \->CTC
```
Current configuration uses decoder only, no CTC.

In the future the decoder can be also taken from a pre-trained model (e.g. BART) in order to improve generative NLU or translation.

tools/Makefile
akreal: `transformers` requirement is now added to the top-level `setup.py` in order to pass the test. I've renamed other related things to "Hugging Face Transformers", hopefully it's not too long.

sw005320:
Thanks for the explanation.
This is a great function!

PR #3422, opened 2021-08-11
Update Sentencepiece and add python 3.9 to CI

shirayu:
This pull request fixes the test for sentencepiece and remove Sentencepiece version limitation.
Because [Sentencepiece supports Python 3.9 since v0.19.5](https://github.com/google/sentencepiece/issues/591#issuecomment-757425951), it enables python 3.9 support.

The reason for the test failures by increasing the version of Sentencepiece was the improper specification of options in the test.
(Insertion of an invalid duplicated option ``--user_defined_symbols``)


sw005320:
Thanks a lot, @shirayu!

PR #3425, opened 2021-08-11
Update SWBD DA configuration

akreal:
- Extend `maxlenratio` to specify a constant output length limit and set it to 1 for SWBD DA.
- Use Conformer, SpecAug, speed perturbation for SWBD DA.


egs2/swbd_da/asr1/conf/tuning/train_asr_conformer.yaml
```diff
@@ -0,0 +1,71 @@
+batch_type: numel
+batch_bins: 7000000
+accum_grad: 1
+max_epoch: 35
```
siddhu001, line 4: Can you try increasing epochs? Maybe adding train/val loss curve will give more intuition if increasing number of epochs will help.

espnet/nets/beam_search.py
```diff
@@ -352,6 +354,8 @@ def forward(
         # set length bounds
         if maxlenratio == 0:
             maxlen = x.shape[0]
+        elif maxlenratio < 0:
```
ftshijt, line 357: Could you also consider adding a unit test for this modification?

akreal: Yes, sure. Is it fine now, or it should be done another way?

ftshijt: LGTM! Thank you.

sw005320:
I think it would be good to have more documentation about the fixed-length output option.
https://github.com/espnet/espnet/blob/1f04ac9d6a12321f18e4ef61f840dff368425a70/doc/tutorial.md#ctc-attention-and-hybrid-ctcattention would not be the best place, but at least it discusses `maxlenratio` and please add this usage here.

Ideally, we want to split the descriptions about training and inference in https://github.com/espnet/espnet/blob/1f04ac9d6a12321f18e4ef61f840dff368425a70/doc/tutorial.md#ctc-attention-and-hybrid-ctcattention and discuss the conventional beam search with maxlenratio etc., and the special case with the fixed-length output option for the classification task.

akreal:
> I think it would be good to have more documentation about the fixed-length output option.
Alright, I'll take a look at it.

PR #3439, opened 2021-08-13
Add vocoder modules related to VITS 

kan-bayashi:
Subset of #3423 

This PR adds vocoder related modules, which used for VITS.
In the next PR, I will add more modules.

PR #3436, opened 2021-08-13
Add new trainer for GAN-based training

kan-bayashi:
Subset of #3423

This PR adds new trainer for GAN-based training.
To do this, a new abstract class `AbsGANESPnetModel` is added.
To use this trainer, the model must be accept `forward_generator=True` as an argument,
which switch the loss calculation of generator and discriminator.

```python
        >>> from espnet2.tasks.abs_task import AbsTask
        >>> class YourESPnetModel(AbsGANESPnetModel):
        ...     def forward(self, input, input_lengths, forward_generator=True):
        ...         ...
        ...         if forward_generator:
        ...             # return loss for the generator
        ...             # optim idx 0 indicates generator optimizer
        ...             return dict(loss=loss, stats=stats, weight=weight, optim_idx=0)
        ...         else:
        ...             # return loss for the discriminator
        ...             # optim idx 1 indicates discriminator optimizer
        ...             return dict(loss=loss, stats=stats, weight=weight, optim_idx=1)
        >>> class YourTask(AbsTask):
        ...     @classmethod
        ...     def build_model(cls, args: argparse.Namespace) -> YourESPnetModel:
```

PR #3423, opened 2021-08-11
[Not for merge] Add VITS for ESPnet2-TTS

kan-bayashi:
This PR adds VITS.

## TODO

- [x] model implementation
  - [x] wavenet
  - [x] hifigan discriminator
  - [x] hifigan generator
  - [x] posterior encoder
  - [x] text encoder
  - [x] stochastic duration predictor
  - [x] flow
  - [x] vits generator
  - [x] mel loss
  - [x] generator adversarial loss
  - [x] discriminator adversarial loss
  - [x] feat match loss
  - [x] kl loss
  - [x] linear spectrogram frontend
- [ ] model test
  - [x] overall test
  - [ ] module-wise test  
- [x] task implementation
  - [x] abs GAN model 
  - [x] abs GAN-TTS model
  - [x] GANTrainer
  - [x] GAN-TTS task 
- [ ] recipe
  - [x] modify `tts.sh` to accept the switch of the task
  - [ ] support `sids` input -> another PR?
- [x] additional items
  - [x] selectable discriminator
  - [x] cache outputs in training

PR #3449, opened 2021-08-13
Add GAN-TTS task with VITS

kan-bayashi:
Subset of #3423 

Merged after the following PRs:
1. #3436 
2. #3437 
3. #3438 
4. #3439 
5. #3448
6. #3497

This PR adds GAN-TTS task and VITS-based model example.


kan-bayashi:
@r9y9 Sorry for many requests of review...
Could you check `vits.py`?
If you can provide comments about the GAN trainer design, that is great:)

r9y9:
No problem! I'll do it today.

espnet2/gan_tts/vits/vits.py
```diff
@@ -0,0 +1,566 @@
...
+        stats = dict(
+            generator_loss=loss.item(),
+            generator_mel_loss=mel_loss.item(),
+            generator_kl_loss=kl_loss.item(),
+            generator_dur_loss=dur_loss.item(),
+            generator_adv_loss=adv_loss.item(),
+            generator_feat_match_loss=feat_match_loss.item(),
+        )
+
+        loss, stats, weight = force_gatherable((loss, stats, batch_size), loss.device)
```
r9y9, line 409: Question: is `weight` the same as `batch_size`? Shouldn't we use the same variable name for the input and output of `force_gatherable `?

espnet2/gan_tts/vits/vits.py
```diff
@@ -0,0 +1,566 @@
...
+            self._cache = outs
+
+        # parse outputs
+        speech_hat_, _, _, start_idxs, *_ = outs
+        speech_ = self.generator.get_segments(
+            x=speech,
+            start_idxs=start_idxs * self.generator.upsample_factor,
+            segment_size=self.generator.segment_size * self.generator.upsample_factor,
+        )
+
```
r9y9, line 484: Question: Are the segments here the same as those in `_forward_generator`? I am concerning if it's okay to use different segments in `_forward_discrminator` and `_forward_generator`.

espnet2/gan_tts/vits/vits.py
```diff
@@ -0,0 +1,566 @@
...
+        # calculate losses
+        real_loss, fake_loss = self.discriminator_adv_loss(p_hat, p)
+        loss = real_loss + fake_loss
+
+        stats = dict(
+            discriminator_loss=loss.item(),
+            discriminator_real_loss=real_loss.item(),
+            discriminator_fake_loss=fake_loss.item(),
+        )
+        loss, stats, weight = force_gatherable((loss, stats, batch_size), loss.device)
```
r9y9, line 498: Same comment as above.

r9y9 (commented):
Comments added for vits.py!

espnet2/gan_tts/vits/vits.py
kan-bayashi: The segments should be the same since the current trainer loops of generator and discriminator with the same batch.
https://github.com/kan-bayashi/espnet/blob/0c2d99e625cb722bd1e05414611917087c8d3394/espnet2/train/gan_trainer.py#L153-L173

espnet2/gan_tts/vits/vits.py
kan-bayashi: `weight` is the kind of special name used in the trainer.
https://github.com/kan-bayashi/espnet/blob/0c2d99e625cb722bd1e05414611917087c8d3394/espnet2/train/gan_trainer.py#L185

PR #3477, opened 2021-08-19
Support joint training of text2mel and vocoder

kan-bayashi:
Should be merged after #3449

This PR enables us to train text2mel and vocoder jointly with random windowed discriminator.


kan-bayashi:
I will re-create PR.

PR #3480, opened 2021-08-19
Predict ASR transcript along with Intent for SLU

siddhu001:


egs2/fsc/asr1/README.md
```diff
@@ -25,8 +25,18 @@
...
 - Pretrainded model: https://zenodo.org/record/5171544
 
-|dataset|Snt|S.Err|
+|dataset|Snt|Intent Classification (%)|
 |---|---|---|
 |inference_asr_model_valid.acc.ave_5best/test|3793|98.8|
 |inference_asr_model_valid.acc.ave_5best/valid|3118|93.6|
 
+
+## Using Transformer based encoder-decoder and decoding sentence with spectral augmentation and predicting transcript along with intent
```
sw005320, line 34: Can you also add an ASR result (just as a record)?

egs2/fsc/asr1/run.sh
```diff
@@ -9,7 +9,7 @@ train_set="train"
 valid_set="valid"
 test_sets="test valid"
 
-asr_config=conf/train_asr.yaml
+asr_config=conf/tuning/train_asr_transformer_adam_specaug.yaml
```
sw005320, line 12: If this config scores the best, you can change a symlink of `conf/train_asr.yaml` to specify `conf/tuning/train_asr_transformer_adam_specaug.yaml`, and make it as default.

egs2/fsc/asr1/README.md
siddhu001: I have added the ASR results but they are over the entire text i.e. over the intent concatenated with transcript. Should I add another scoring script to compute these metrics only over transcript (i.e. after excluding the first word)?

egs2/fsc/asr1/run.sh
siddhu001: Good point! I have made that change.

egs2/fsc/asr1/README.md
sw005320: I think having ASR results without the intent is better.

PR #3481, opened 2021-08-19
Gradio Web Demo

AK391:
easy to use web demo on huggingface spaces using gradio

### DIFF
--- a/README.md
+++ b/README.md
@@ -76,6 +76,7 @@ and also follows [Kaldi](http://kaldi-asr.org/) style data processing, feature e
 
 Demonstration
 - Real-time ASR demo with ESPnet2  [![Open In Colab](https://colab.research.google.com/assets/colab-badge.svg)](https://colab.research.google.com/github/espnet/notebook/blob/master/espnet2_asr_realtime_demo.ipynb)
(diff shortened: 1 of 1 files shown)


sw005320:
Thanks!
Can you tell me the benefit of Gradio in this case?

AK391:
@sw005320 Hi, it is a web demo that can be used without installing anything in the browser, will use the multilingual model to do ASR from a audio file and output the text all in the browser, easy to use for users

sw005320:
Thanks for the clarification!
- I'm checking https://huggingface.co/spaces/akhaliq/espnet2_asr and https://huggingface.co/julien-c/mini_an4_asr_train_raw_bpe_valid and we could also have some web interface to support the demo from the browser. What would be the additional benefits?
- When we need to add more models, how to do it?

AK391:
@sw005320 

- the app is here for the first link https://huggingface.co/spaces/akhaliq/espnet2_asr/blob/main/app.py, this can be edited to update the app, the benefit would be having a easy to use interface to test out the model and also learn from user feedback to check model robustness
- for adding more models let me know, I can update the app

AK391:
@sw005320 yes, https://huggingface.co/spaces/akhaliq/espnet2_asr/blob/main/app.py currently it is using the multilingual model but I can add optional for other models

sw005320:
I see. 
Thanks a lot!
I think this is very cool.
Year, we have too many ASR models but it would be great if you add several models, including librispeech, laborotv, zeroth_korean, wsj, csj, etc.

PR #3483, opened 2021-08-20
Add cython code for VITS

kan-bayashi:
This PR adds cython code of maximum path calculation.
I'm wondering which is better:
- keep the cython setup as an optional (current style)
- integrate with root `setup.py` as a default 

For setup.py, I referred this one:
https://github.com/lumaku/ctc-segmentation/blob/master/setup.py


r9y9 (approved):
LGTM

> keep the cython setup as an optional (current style)

This looks better to me. I suppose ASR people will not probably use this module in most use cases.

madjxatw:
`core.pyx` is not available in espnet package from pypi, do i have to manually download and compile it?

PR #3484, opened 2021-08-20
Add ASR multitask with CTC loss

siddhu001:


sw005320:
@simpleoier, could I ask you to review this PR?
I think this is a very good extension, but the way of the extension would be used for many purposes (various structured/multimodal data) and we need some careful reviews.

simpleoier:
@siddhu001 Can you make the `*_transcript` as an optional arguments? Then it will be more consistent with the previous pipeline to avoid the incompatibility with previous versions. If so, most of previous data / stats can be reused.

PR #3485, opened 2021-08-20
Add Swahili ASR recipe

akreal:


egs2/TEMPLATE/asr1/db.sh
```diff
@@ -99,6 +100,8 @@ CMU_INDIC=downloads
 JKAC=
 MUCS_SUBTASK1=downloads
 MUCS_SUBTASK2=downloads
+GAMAYUN=downloads
+IWSLT21LR=downloads/iwslt21
```
ftshijt, line 104: Please also add an entry for JHU domain

egs2/swahili/asr1/local/valid-iwslt.txt
```diff
@@ -0,0 +1,869 @@
+iwslt_1-828fade7fed32c93944ace4965f81a48__1573474880.3137 Julia anasema Tommy atakula chakula cha mchana na yeye.
```
ftshijt, line 1: Can we load this somewhere else instead of host it in the repo?

egs2/swahili/asr1/run.sh
```diff
@@ -0,0 +1,33 @@
...
+test_sets="test_iwslt_swa test_iwslt_swc test_iwslt_swa_raw test_iwslt_swc_raw"
+
+asr_config=conf/train_asr_conformer.yaml
+lm_config=conf/train_lm_transformer.yaml
+inference_config=conf/decode_asr.yaml
+
+./asr.sh \
+    --lang sw \
+    --ngpu 2 \
+    --nbpe 100 \
```
ftshijt, line 19: Definite OK if your system is set as so. I'm just interesting if the bpe size is a bit small to the data (I remember the combination of all the data should be around 100 hours (correct me if I'm wrong)

ftshijt (commented):
Thanks for your effort! Is it possible to host or process on the fly with `egs2/swahili/asr1/local/num2words-swc.txt` and `egs2/swahili/asr1/local/num2words-swa.txt` in somewhere else. Seems that the file is too large to host in the repo.

egs2/swahili/asr1/run.sh
akreal: That's right, the data is 60 hours.

WER seems to increase when the BPE size is increased:

|nbpe|swa|swc|
|----|---|---|
|100| 27.7| 26.8|
|150| 26.8| 28.0|
|200| 28.6| 26.9|
|500| 28.6| 27.7|

ftshijt: Interesting, thanks for your inforamtion!

egs2/TEMPLATE/asr1/db.sh
akreal: Done.

egs2/swahili/asr1/local/valid-iwslt.txt
akreal: Moved to zenodo.

akreal:
Thank you for the review!

> Thanks for your effort! Is it possible to host or process on the fly with `egs2/swahili/asr1/local/num2words-swc.txt` and `egs2/swahili/asr1/local/num2words-swa.txt` in somewhere else. Seems that the file is too large to host in the repo.

Moved to zenodo.

ftshijt:
Very cool! I will merge it after you fix the ci part.

akreal:
CI fails because the config validation tries to load the pre-trained model referenced in the configuration. I've tried to fix this by skipping the model load for `dry_run` and `collect_stats` modes.

ftshijt:
LGTM! thanks for your effort.

sw005320:
Sorry, I just have one request. The recipe name `swahili` sounds a bit too general.
Can we change the recipe name a bit more specific?

akreal:
> Sorry, I just have one request. The recipe name `swahili` sounds a bit too general.
> Can we change the recipe name a bit more specific?

This recipe combines several Swahili datasets (ALFFA, BABEL, Gamayun, IWSLT21), so it is sort of general. Naming all datasets would be too long. Kaldi uses names like `multi_en` and `mutli_cn` for such recipes. Would it be better to name this recipe `multi_sw`?

ftshijt:
`multi_sw` seems OK. I would suggest using `iwslt_low_resource` as the recipe are for that purpose? This also aligns with `egs`

PR #3490, opened 2021-08-21
Support SID and LID inputs for TTS models

kan-bayashi:
This PR supports SID and LID inputs.
And I changed the directory format to be more consistent.


espnet2/tts/fastspeech/fastspeech.py
```diff
@@ -123,6 +65,7 @@ def __init__(
         postnet_layers: int = 5,
         postnet_chans: int = 512,
         postnet_filts: int = 5,
+        postnet_dropout_rate: float = 0.5,
```
r9y9, line 68: Regarding the change of arguments order: this may break user code. Assuming this class is used everywhere with kwargs (not args), this change would be okay.

r9y9 (commented):
Please check one my concern. If that's not a problem, LGTM.

espnet2/tts/fastspeech/fastspeech.py
r9y9: The same applies to Tacotron2 and Transformer.

kan-bayashi: In espnet2, all model arguments are dealt as kwargs. Therefore is should be no problem.

r9y9 (approved):

PR #3492, opened 2021-08-21
Add the GigaSpeech normalization and fix the WER

chaisz19:


sw005320:
Can you merge the master?
We seem to have a CI error related to the k2 installation, which is fixed in the master repo.

PR #3501, opened 2021-08-21
Support joint training of text2mel and vocoder

kan-bayashi:
This PR enables us to train text2mel and vocoder modules jointly.


espnet2/gan_tts/joint/joint_text2wav.py
```diff
@@ -0,0 +1,641 @@
...
+        )
+        segments = self.get_segments(x, start_idxs, segment_size)
+        return segments, start_idxs
+
+    def get_segments(
+        self,
+        x: torch.Tensor,
+        start_idxs: torch.Tensor,
+        segment_size: int,
+    ) -> torch.Tensor:
```
r9y9, line 600: Dup of https://github.com/espnet/espnet/blob/f03101557753517ebac8c432f0793d97d68fa5f0/espnet2/gan_tts/vits/generator.py#L406-L453?

Maybe we can move these instance methods to functions.

r9y9 (commented):
Added a comment. Other than that, LGTM.

espnet2/gan_tts/joint/joint_text2wav.py
kan-bayashi: Made it as a separated function.

PR #3505, opened 2021-08-22
aishell3 dataset preparation

actboy:
1. data/train_phn/wav.scp.tmp may be typo
2. set data_aishell3.tgz file size


ftshijt:
Many thanks!

PR #3509, opened 2021-08-23
Update SIWIS README

takenori-y:
I uploaded the pretrained models trained on SIWIS corpus. Do you need the pretrained fastspeech models?


kan-bayashi:
> Do you need the pretrained fastspeech models?

If you have a free gpu, that is great!

kan-bayashi (approved):
LGTM, thanks!

PR #3457, opened 2021-08-16
Fix minor bugs in ESPnet2 diarization recipe

YushiUeda:
This PR is for fixing the following minor bugs.
1. Stage numbers in egs2/TEMPLATE/diar1/diar.sh
2. Directory paths in egs2/mini_librispeech/diar1/local/data.sh

PR #3524, opened 2021-08-24
Modify ESPnet2 diarization recipe

YushiUeda:
This PR is a continuation of https://github.com/espnet/espnet/pull/3457.
2 further modifications have been made.

1. `train_diar.yaml`
Training with `init: chainer` didn't work properly, so it is modified to `init: xavier_uniform`. Some other hyper-parameters are modified as well for achieving better performance.

2. `label_aggregation.py`
Zero padding seemed to affect the label aggregation in the start and end frames. (For these frames, speaker labels tend to be `0`, since half of the samples are padded with `0`). To avoid this, zero-padded samples are replaced with the copy of actual input samples within the frames.

### DIFF
--- a/egs2/mini_librispeech/diar1/conf/train_diar.yaml
+++ b/egs2/mini_librispeech/diar1/conf/train_diar.yaml
@@ -26,10 +26,10 @@ batch_size: 16
 optim: adam
 accum_grad: 2
 grad_clip: 5
-patience: 0
+patience: 3
 max_epoch: 100
 optim_conf:
-    lr: 1.0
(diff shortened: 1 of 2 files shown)


sw005320:
Please fix the CI error https://github.com/espnet/espnet/pull/3524/checks?check_run_id=3416315966#step:8:7

sw005320:
will you add a result and upload a model?

YushiUeda:
@sw005320 thank you for the comments!

> will you add a result and upload a model?

I will work on it shortly.

PR #3535, opened 2021-08-30
Support `from_pretrained` for ASR and ENH

kan-bayashi:
This PR supports `from_pretrained` function for ASR and ENH.

```python
[ins] In [1]: from espnet2.bin.asr_inference import Speech2Text

[ins] In [2]: asr = Speech2Text.from_pretrained("Shinji Watanabe/laborotv_asr_train_asr_conformer2_latest33_raw_char_sp_valid.acc.ave")

[ins] In [3]: from espnet2.bin.enh_inference import SeparateSpeech

[ins] In [4]: ss = SeparateSpeech.from_pretrained("Chenda Li/wsj0_2mix_enh_train_enh_rnn_tf_raw_valid.si_snr.ave")
```

### DIFF
--- a/espnet2/bin/enh_inference.py
+++ b/espnet2/bin/enh_inference.py
@@ -3,6 +3,7 @@ import argparse
 import logging
 from pathlib import Path
 import sys
+from typing import Any
 from typing import List
 from typing import Optional
 from typing import Sequence
@@ -43,7 +44,7 @@ class SeparateSpeech:
 
     def __init__(
         self,
-        enh_train_config: Union[Path, str],
+        enh_train_config: Union[Path, str] = None,
         enh_model_file: Union[Path, str] = None,
         segment_size: Optional[float] = None,
         hop_size: Optional[float] = None,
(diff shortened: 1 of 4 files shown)


espnet2/bin/enh_inference.py
```diff
@@ -253,6 +254,38 @@ def cal_permumation(self, ref_wavs, enh_wavs, criterion="si_snr"):
...
+
+            except ImportError:
+                logging.error(
+                    "`espnet_model_zoo` is not installed. "
+                    "Please install via `pip install -U espnet_model_zoo`."
+                )
+                raise
+            d = ModelDownloader()
+            train_config, model_file = d.download_and_unpack(enh_model_tag).values()
+            # FIXME: The arugment name is mismatched, it is better to rename.
```
kan-bayashi, line 284: @Emrys365 Here the arguments from downloader are different, which are defined in https://github.com/espnet/espnet/blob/fb75313bd1ddf027e075ea7197448c929b176f5d/espnet2/bin/pack.py#L24-L26
Maybe it is better to keep the same.

kan-bayashi:
I also updated README in espnet_model_zoo.
https://github.com/espnet/espnet_model_zoo/pull/46

sw005320 (approved):
LGTM.
@YushiUeda, can you also work on the same thing for diarization?

YushiUeda:
> @YushiUeda, can you also work on the same thing for diarization?
Sure! I will work on it right away.

ftshijt (commented):
Looks cool from my side. Just two comments for discussion:

- I was using `init_param` for pre-training in the past, do you think we should somehow unify the interface with both?
- Could you please also add instances for some egs? people may not know we have provided this good function

kan-bayashi:
> I was using init_param for pre-training in the past, do you think we should somehow unify the interface with both?

I think adding `from_pretrained` for each module is a little bit complicated and current `--init_param` option can deal with complex loading such as subset of parameters, excluding come parameters, and so on. Therefore, `from_pretrained` should be for only inference interface now.

> Could you please also add instances for some egs? people may not know we have provided this good function

For now I updated espnet_model_zoo instruction.
https://github.com/espnet/espnet_model_zoo/blob/5c20277fedab9377a421a587ab1fa2b10fd195df/README.md
I will update the usage in README in another PR.

ftshijt (approved):
LGTM! Many thanks

espnet2/bin/enh_inference.py
Emrys365: Sorry for the mismatch. We should have kept consistency with other tasks. Could you remove the prefix "enh_" from all argument names in enh_inference.py?

kan-bayashi: I will do it in the next PR.

PR #3537, opened 2021-08-30
Support `from_pretrain` for DIAR

YushiUeda:
This PR is related to https://github.com/espnet/espnet/pull/3535.
This is to support `from_pretrain` also for DIAR.

### DIFF
--- a/espnet2/bin/diar_inference.py
+++ b/espnet2/bin/diar_inference.py
@@ -4,6 +4,7 @@ import argparse
 import logging
 from pathlib import Path
 import sys
+from typing import Any
 from typing import List
 from typing import Optional
 from typing import Sequence
@@ -41,7 +42,7 @@ class DiarizeSpeech:
 
     def __init__(
         self,
-        diar_train_config: Union[Path, str],
+        diar_train_config: Union[Path, str] = None,
         diar_model_file: Union[Path, str] = None,
         segment_size: Optional[float] = None,
         normalize_segment_scale: bool = False,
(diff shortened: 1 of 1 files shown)


espnet2/bin/diar_inference.py
```diff
@@ -159,6 +160,36 @@ def __call__(
...
+                from espnet_model_zoo.downloader import ModelDownloader
+
+            except ImportError:
+                logging.error(
+                    "`espnet_model_zoo` is not installed. "
+                    "Please install via `pip install -U espnet_model_zoo`."
+                )
+                raise
+            d = ModelDownloader()
+            kwargs.update(**d.download_and_unpack(model_tag))
```
kan-bayashi, line 189: The name of returned arguments are different.
https://github.com/espnet/espnet/blob/fb75313bd1ddf027e075ea7197448c929b176f5d/espnet2/bin/pack.py#L29-L32
Please change the names or add workaround solution.
https://github.com/kan-bayashi/espnet/blob/7f42a925b82dc722df7d93f08e07a1a62df390b3/espnet2/bin/enh_inference.py#L283-L285

espnet2/bin/diar_inference.py
```diff
@@ -294,8 +330,16 @@ def get_parser():
...
+    group.add_argument(
+        "--diar_train_config",
+        type=str,
+        help="Diarization training configuration",
+    )
+    group.add_argument(
+        "--diar_model_file",
+        type=str,
+        help="Diarization model parameter file",
+    )
```
kan-bayashi, line 342: `model_tag` option is not added.

espnet2/bin/diar_inference.py
YushiUeda: Thanks for the review. I corrected it as in `enh_inference.py`.

espnet2/bin/diar_inference.py
YushiUeda: added `model_tag`

espnet2/bin/diar_inference.py
kan-bayashi: Thanks!

PR #3544, opened 2021-09-01
Avoid tokenization for reference during evaluation

hirofumi0810:
In current evaluation, we detokenize "tokenized reference" for evaluation. Here, OOV tokens in references are ignored.
This PR recovers such OOV tokens in the references. But, in my experience, this does not change results so much (within 0.05 BLEU).
The same issue still exists in the ASR task.

### DIFF
--- a/egs/fisher_callhome_spanish/st1/local/score_bleu.sh
+++ b/egs/fisher_callhome_spanish/st1/local/score_bleu.sh
@@ -45,13 +45,13 @@ if [ ! -z ${set} ] && [ -f ${dir}/data_ref1.json ]; then
 fi
 
 if [ ! -z ${bpemodel} ]; then
-    spm_decode --model=${bpemodel} --input_format=piece < ${dir}/ref.trn | sed -e "s/▁/ /g" > ${dir}/ref.wrd.trn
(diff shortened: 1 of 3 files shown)


sw005320:
Thanks

> The same issue still exists in the ASR task.

I see.
I just tried to understand it.
Since the evaluation of the OOV characters is skipped, this may change the denominator (basically, it skips the OOV character)? 
Does it happen when we have OOV characters like Mandarin/Japanese?

hirofumi0810:
Because we remove OOV tokens, the denominator (#words) might decrease. I think this would be severe when using large BPE units.
> Since the evaluation of the OOV characters is skipped, this may change the denominator (basically, it skips the OOV character)?


Yes. It is more severe than English.
> Does it happen when we have OOV characters like Mandarin/Japanese?

sw005320:
Thanks!
Does it happen in English?
Even when we use a large BPE size, I thought that we can always tokenize/detokenize a sentence if there is no OOV character since the BPE token also includes a character.
I thought that it always guarantees to tokenize/detokenize any sentences.

sw005320:
> But most corpora have noisy non-English characters (e.g. TED).

Got it!

hirofumi0810:
@sw005320 This PR is ready.

PR #3557, opened 2021-09-04
Add new Japanese G2P module

kan-bayashi:
This PR add a new Japanese G2P for TTS based on [Prosodic features control by symbols as input of sequence-to-sequence acoustic modeling for neural TTS](https://doi.org/10.1587/transinf.2020EDP7104) with @r9y9's tweaks.


kan-bayashi:
I made a branch from wrong root...
Re-create PR.

PR #3569, opened 2021-09-07
Add ASR portable evaluation script

kan-bayashi:
This PR adds a portable ASR script, which can be used for other tasks (e.g., tts).

### DIFF
--- a/egs2/TEMPLATE/asr1/scripts/utils/evaluate_asr.sh
+++ b/egs2/TEMPLATE/asr1/scripts/utils/evaluate_asr.sh
@@ -28,8 +28,8 @@ SECONDS=0
 # General configuration
 stage=1
 stop_stage=1
-nj=1
-gpu_inference=true
+nj=8
+gpu_inference=false
 
 # Model related configuration
 model_tag=""
@@ -124,18 +124,22 @@ else
 fi
 
 if [ ${stage} -le 1 ] && [ ${stop_stage} -ge 1 ]; then
-    _opts=
+    _opts=()
     if [ -n "${inference_config}" ]; then
-        _opts+="--config ${inference_config} "
+        _opts+=("--config")
+        _opts+=("${inference_config}")
     fi
     if [ -n "${asr_model_file}" ]; then
-        _opts+="--asr_model_file ${asr_model_file} "
+        _opts+=("--asr_model_file")
+        _opts+=("${asr_model_file}")
     fi
     if [ -n "${lm_file}" ]; then
-        _opts+="--lm_file ${lm_file} "
+        _opts+=("--lm_file")
+        _opts+=("${lm_file}")
     fi
     if [ -n "${model_tag}" ]; then
-        _opts+="--model_tag ${model_tag} "
+        _opts+=("--model_tag")
+        _opts+=("${model_tag}")
     fi
 
     logdir="${outdir}/logdir"
@@ -160,7 +164,7 @@ if [ ${stage} -le 1 ] && [ ${stop_stage} -ge 1 ]; then
             --data_path_and_name_and_type "${wavscp},speech,sound" \
             --key_file "${logdir}"/keys.JOB.scp \
             --output_dir "${logdir}"/output.JOB \
-            ${_opts} ${inference_args}
+            "${_opts[@]}" ${inference_args}
 
     # 3. Concatenates the output files from each jobs
     for f in token token_int score gt_text; do
--- a/egs2/TEMPLATE/asr1/scripts/utils/evaluate_asr.sh
+++ b/egs2/TEMPLATE/asr1/scripts/utils/evaluate_asr.sh
@@ -27,9 +27,10 @@ SECONDS=0
 
 # General configuration
 stage=1
-stop_stage=1
+stop_stage=2
 nj=8
 gpu_inference=false
+fs=16000
 
 # Model related configuration
 model_tag=""
@@ -55,6 +56,7 @@ Options:
     --stop_stage     # Processes is stopped at the specified stage (default="${stop_stage}").
     --nj             # Number of parallel jobs (default="${nj}").
     --gpu_inference  # Whether to use gpu in the inference (default="${gpu_inference}").
+    --fs             # Sampling rate for ASR model inputs (default="${fs}").
 
     # Model related configuration
     --model_tag       # Model tag or url available in espnet_model_zoo (default="${model_tag}")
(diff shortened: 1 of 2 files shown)


egs2/TEMPLATE/asr1/scripts/utils/evaluate_asr.sh
```diff
@@ -0,0 +1,266 @@
...
+        _opts+=("--asr_model_file")
+        _opts+=("${asr_model_file}")
+    fi
+    if [ -n "${lm_file}" ]; then
+        _opts+=("--lm_file")
+        _opts+=("${lm_file}")
+    fi
+    if [ -n "${model_tag}" ]; then
+        # FIXME: workaround until fixing filelock
+        espnet_model_zoo_download --unpack true "${model_tag}"
```
Emrys365, line 142: Would you consider the support for using the configuration packed inside the model for inference and decoding? e.g. the language model, BPE configs.
> In [egs2/TEMPLATE/asr1/asr.sh](https://github.com/espnet/espnet/blob/master/egs2/TEMPLATE/asr1/asr.sh#L1350-L1374), it allows packing `lm_file`, `bpemodel`, `nlsyms_txt` and other files along with `asr_model_file` into the model.

I used to try something like the following to parse the BPE config from the downloaded model:
```bash
log "Downloading ${asr_model} for decoding and evaluation"
asr_exp="${download_dir}/${asr_model// /_}"
mkdir -p "${asr_exp}"

# If the model already exists, you can skip downloading
espnet_model_zoo_download --unpack true "${asr_model}" > "${asr_exp}/config.txt"

# Get the path of each file
_asr_model_file=$(<"${asr_exp}/config.txt" sed -e "s/.*'asr_model_file': '\([^']*\)'.*$/\1/")
_asr_train_config=$(<"${asr_exp}/config.txt" sed -e "s/.*'asr_train_config': '\([^']*\)'.*$/\1/")

# Get token_type
token_type=$(<"${_asr_train_config}" grep -Eo "^token_type: (\w+)$" | awk '{print $2}')
if [ "${token_type}" = "bpe" ]; then
    bpemodel=$(<"${_asr_train_config}" grep -Eo "^bpemodel: .*$" | awk '{print $2}' | sed -e 's/^"\(.*\)"/\1/' -e "s/^'\(.*\)'/\1/")
    if [ -z "${bpemodel}" ]; then
        log "Invalid model: token_type=bpe, but 'bpemodel' is not defined in '${_asr_train_config}'"
        exit 1
    elif [ ! -f "${bpemodel}" ]; then
        log "No such file for 'bpemodel': ${bpemodel}"
        exit 1
    fi
    # nbpe_mode=$(echo "${bpemodel}" | grep -Po "(?<=/bpe_)\w+(?=/bpe\.model$)")
    # nbpe=$(echo "${nbpe_mode}" | grep -Po "\d+")
    # bpemode=$(echo "${nbpe_mode}" | grep -Po "[a-zA-Z]+")
else
    bpemodel=none
fi

# Create symbolic links
ln -sf "${_asr_model_file}" "${asr_exp}"
ln -sf "${_asr_train_config}" "${asr_exp}"
inference_asr_model=$(basename "${_asr_model_file}")

if [ "$(<${asr_exp}/config.txt grep -c lm_file)" -gt 0 ]; then
    _lm_file=$(<"${asr_exp}/config.txt" sed -e "s/.*'lm_file': '\([^']*\)'.*$/\1/")
    _lm_train_config=$(<"${asr_exp}/config.txt" sed -e "s/.*'lm_train_config': '\([^']*\)'.*$/\1/")

    lm_exp="${expdir}/${download_model// /_}/lm"
    mkdir -p "${lm_exp}"

    ln -sf "${_lm_file}" "${lm_exp}"
    ln -sf "${_lm_train_config}" "${lm_exp}"
    inference_lm=$(basename "${_lm_file}")
fi
```

egs2/TEMPLATE/asr1/scripts/utils/evaluate_asr.sh
```diff
@@ -0,0 +1,266 @@
...
+    # 3. Concatenates the output files from each jobs
+    for f in token token_int score gt_text; do
+        for i in $(seq "${_nj}"); do
+            cat "${logdir}/output.${i}/1best_recog/${f}"
+        done | LC_ALL=C sort -k1 >"${outdir}/${f}"
+    done
+fi
+
+# 4. Scoring
+if [ ${stage} -le 2 ] && [ ${stop_stage} -ge 2 ]; then
```
sw005320, line 180: This scoring part seems to be too simple and could not use `local/score.sh` (e.g., in swbd). 
But I think we should not make it complicated and this looks fine to me.
This is just a note.

sw005320 (commented):
It's good for me.
- In general, I'm concerning that the change in `asr.sh` cannot be reflected, but it would be difficult, and we just care it manually.
- Can we also add a test of this script in https://github.com/espnet/espnet/tree/master/test_utils? This will mitigate the above issue.

egs2/TEMPLATE/asr1/scripts/utils/evaluate_asr.sh
kan-bayashi: If we use model_tag, the model will be loaded through `from_pretrained` function.
https://github.com/espnet/espnet/blob/bd23667ed1fb34e62d2a2c89030001840feeb61f/espnet2/bin/asr_inference.py#L262-L275
So the lm or bpe setting will be loaded automatically.
And this download part is needed to prevent from simultaneous download when using parallel `asr_inference` jobs (basically, this can prevent with `filelock` but now espnet_model_zoo's file lock does not work well).

egs2/TEMPLATE/asr1/scripts/utils/evaluate_asr.sh
```diff
@@ -0,0 +1,266 @@
...
+    ${_cmd} --gpu "${_ngpu}" JOB=1:"${_nj}" "${logdir}"/asr_inference.JOB.log \
+        python3 -m espnet2.bin.asr_inference \
+            --ngpu "${_ngpu}" \
+            --data_path_and_name_and_type "${wavscp},speech,sound" \
+            --key_file "${logdir}"/keys.JOB.scp \
+            --output_dir "${logdir}"/output.JOB \
+            "${_opts[@]}" ${inference_args}
+
+    # 3. Concatenates the output files from each jobs
+    for f in token token_int score gt_text; do
```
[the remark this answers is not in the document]
kan-bayashi: Sorry this is a typo caused by replacing.
I will fix it.

kan-bayashi:
- I added test code.
- I added format wav stage since asr_inference does not check the sampling rate.

egs2/TEMPLATE/asr1/scripts/utils/evaluate_asr.sh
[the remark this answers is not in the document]
kan-bayashi: fixed.

egs2/TEMPLATE/asr1/scripts/utils/evaluate_asr.sh
Emrys365: I see. It very cool to have the support already!

sw005320 (approved):

PR #3584, opened 2021-09-13
ESPnet2 iemocap recipe

YushiUeda:
This PR is to add ESPnet2 recipe for iemocap database.


ftshijt:
LGTM! Many thanks. I would recommend uploading the model to `espnet_model_zoo` as well

egs2/iemocap/asr1/local/lowercase.pl
```diff
@@ -0,0 +1,24 @@
...
+#
+# This file is part of moses.  Its use is licensed under the GNU Lesser General
+# Public License version 2.1 or, at your option, any later version.
+# Modified for espnet2 iemocap recipe
+# Yushi Ueda, Carnegie Mellon University (2021)
+
+use warnings;
+use strict;
+
+while (@ARGV) {
```
ftshijt, line 11: We have `moses` installation available at espnet, maybe we could use that directly?

egs2/iemocap/asr1/local/normalize_punctuation.pl
```diff
@@ -0,0 +1,117 @@
+#!/usr/bin/env perl
+#
+# This file is part of moses.  Its use is licensed under the GNU Lesser General
```
ftshijt, line 3: Ditto

egs2/iemocap/asr1/local/score.py
```diff
@@ -0,0 +1,88 @@
+#!/usr/bin/env python
+
+# Copyright 2021  Siddhant Arora
```
ftshijt, line 3: If it is the same as other recipes, you could add a softlink instead of copy it.

egs2/iemocap/asr1/local/lowercase.pl
YushiUeda: Thanks, but I slightly modified the code to fit to this recipe, so we couldn't use moses directly in this case.

egs2/iemocap/asr1/local/normalize_punctuation.pl
YushiUeda: Ditto

egs2/iemocap/asr1/local/score.py
YushiUeda: This one is also modified for this recipe. (most of the parts are taken from https://github.com/espnet/espnet/blob/master/egs2/fsc/asr1/local/score.py)

ftshijt:
Sounds cool! I will merge the PR. Thanks for your effort @YushiUeda

ftshijt (approved):

YushiUeda:
@ftshijt Thanks so much for your review!

PR #3586, opened 2021-09-14
add korean_jaso tokenizer and korean_cleaner

windtoker:
Add korean_jaso tokenizer and korean_cleaner
- supporting kss recipe
- unlike g2pk, only splitting into graphemes


kan-bayashi:
Thank you for your great PR!
Is it possible to integrate jaso_tokenizer in phoneme_tokenizer?
I think we can use the same interface.

windtoker:
Yes. I think its poosible. 
I will try to merge Jaso into phonemizer.py and then check if it works.
Then I will send a new PR.

PR #3588, opened 2021-09-15
Add korean_jaso tokenizer and korean_cleaner

windtoker:
- supporting kss recipe
- korean_jaso tokenizer : support splitting of korean sentence to jaso graphemes
- korean_cleaner : basic conversion of english alphabets and numbers to korean pronounciation
- integration to phoneme_tokenizer for code reusability

### DIFF
--- a/espnet2/text/cleaner.py
+++ b/espnet2/text/cleaner.py
@@ -8,7 +8,11 @@ try:
     from vietnamese_cleaner import vietnamese_cleaners
 except ImportError:
     vietnamese_cleaners = None
-
+    
+try:
+    from .korean_cleaner import Korean_cleaner
+except ImportError:
+    Korean_cleaner = None
 
 class TextCleaner:
     """Text cleaner.
@@ -40,6 +44,10 @@ class TextCleaner:
                 if vietnamese_cleaners is None:
                     raise RuntimeError("Please install underthesea")
                 text = vietnamese_cleaners.vietnamese_cleaner(text)
+            elif t == 'korean_cleaner':
+                if Korean_cleaner is None:
+                    raise RuntimeError("Please install underthesea")
+                text = Korean_cleaner.normalize_text(text)
             else:
                 raise RuntimeError(f"Not supported: type={t}")
 
--- a/espnet2/text/korean_cleaner.py
+++ b/espnet2/text/korean_cleaner.py
@@ -1,0 +1,77 @@
+
+## Referenced from https://github.com/hccho2/Tacotron-Wavenet-Vocoder-Korean
+
+import re
+
+class Korean_cleaner:
+    @classmethod
+    def _normalize_numbers(cls, text):
+        number_to_kor = {
+            '0': '영',
+            '1': '일',
+            '2': '이',
+            '3': '삼',
+            '4': '사',
+            '5': '오',
+            '6': '육',
+            '7': '칠',
+            '8': '팔',
+            '9': '구',
+        }
+        new_text = "".join(number_to_kor[char] if char in number_to_kor.keys() else char for char in text)
+        return new_text
+
+    @classmethod
+    def _normalize_english_text(cls, text):
+        upper_alphabet_to_kor = {
+            'A': '에이',
+            'B': '비',
+            'C': '씨',
+            'D': '디',
+            'E': '이',
+            'F': '에프',
+            'G': '지',
+            'H': '에이치',
+            'I': '아이',
+            'J': '제이',
+            'K': '케이',
+            'L': '엘',
+            'M': '엠',
+            'N': '엔',
+            'O': '오',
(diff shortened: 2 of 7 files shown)


espnet2/text/korean_cleaner.py
```diff
@@ -0,0 +1,77 @@
...
+        # stage 2 : normalize english text
+        text = cls._normalize_english_text(text)
+        return text
+
+if __name__ == '__main__':
+    sentence = "내가 ABC 마트에 가는길은 너무나 험난해. 1분 1초도 아까워"
+    sentence = "지난 mbC 프로그램 봤어? 12세 이상 관람가인데 이상해"
+
+    new_sentence = Korean_cleaner.normalize_text(sentence)
+    print(new_sentence)
```
kan-bayashi, line 77: Could you move this test to the following test code?
https://github.com/espnet/espnet/blob/master/test/espnet2/text/test_cleaner.py

espnet2/text/cleaner.py
```diff
@@ -8,7 +8,11 @@
     from vietnamese_cleaner import vietnamese_cleaners
 except ImportError:
     vietnamese_cleaners = None
-
+    
+try:
+    from .korean_cleaner import Korean_cleaner
+except ImportError:
+    Korean_cleaner = None
```
kan-bayashi, line 15: Since we have no dependency for korean cleaner, jsut using `from ... import ...` is OK without try-except.

espnet2/text/korean_cleaner.py
```diff
@@ -0,0 +1,77 @@
+
+## Referenced from https://github.com/hccho2/Tacotron-Wavenet-Vocoder-Korean
+
+import re
+
+class Korean_cleaner:
```
kan-bayashi, line 6: I like CamelCase for class name.
```suggestion
class KoreanCleaner:
```

espnet2/text/cleaner.py
```diff
@@ -40,6 +44,10 @@ def __call__(self, text: str) -> str:
                 if vietnamese_cleaners is None:
                     raise RuntimeError("Please install underthesea")
                 text = vietnamese_cleaners.vietnamese_cleaner(text)
+            elif t == 'korean_cleaner':
+                if Korean_cleaner is None:
+                    raise RuntimeError("Please install underthesea")
```
kan-bayashi, line 49: Ditto.
We do not need to check here.

espnet2/text/cleaner.py
kan-bayashi: `vietnamese_cleaners` requires additional dependency so we use try-except.

kan-bayashi (changes requested):
Thank you for your modification! Looks nice :)
Could you also add the g2p test?
https://github.com/espnet/espnet/blob/master/test/espnet2/text/test_phoneme_tokenizer.py

espnet2/text/korean_cleaner.py
windtoker: Sure I can. 

Btw I have a question regarding space_symbol. "PhonemeTokenizer" class takes "space_symbol" argument but not sure where it changed " " to space_symbol defined. tested on korean gp2k for token list generation and " " was put in token list instead of <space>.  I converted " " to "<space>" in "Jaso" class I created as an workaround. Because I don't know how other g2p modules work, if you think its okay, I can add a line that converts " " to space_symbol in PhonemeTokenizer.text2tokens.

kan-bayashi: Oh you are right. Space symbol is not used in the current setting, so " " is used as is. Let me think it.

windtoker: Okay. I will hold it till its determined then.

kan-bayashi: Sorry for the late reply.
I think replacing with `<space>` in your class should be fine (also using " " as is also fine). So keeping current style is OK.
The modification of `PhonemeTokenizer.text2tokens` might affect the compatibility of the pretrained model, so please keep it as is for now.

kan-bayashi: Sorry for the late reply.
I think replacing with `<space>` in your class should be fine (also using " " as is also fine). So keeping current style is OK.
The modification of `PhonemeTokenizer.text2tokens` might affect the compatibility of the pretrained model, so please keep it as is for now.

windtoker: okay! thank you. then ill proceed the remaining process.

windtoker:
Here I updated the feedbacks you requested.

Please check if there are any missed part. I'd happy to reflect it.

windtoker:
Sorry for the late reply! I resolved the issues and it seems all checks are passed. Btw Great job for the paper submission!

PR #3590, opened 2021-09-15
Update pret-trained model for TTS recipes

ftshijt:


kan-bayashi (approved):
Thank you for updating!

PR #3591, opened 2021-09-16
FastEmit support

b-flo:
Hi,

This PR adds support for FastEmit (https://arxiv.org/abs/2010.11148) through `fastemit-lambda` parameter. To make it work, `warp-transducer` (forked version) should be re-installed as the functionality is only supported with `espnet_v1.1` branch instead of `patch` branch previously.

Let me do some testing before merging.

P.S: Out-of-scope but I also re-ordered parameters in TransducerTasks and added missing docstrings.


sw005320:
Great job, @b-flo!
LGTM.
My only suggestion is to advertise this new function in our main readme.

After you fix the CI error, I'll merge this PR.
If you have some example configs and results, please share them in the later PR.

sw005320 (approved):

b-flo:
> After you fix the CI error, I'll merge this PR.

It may be better to do some testing before merging. There shouldn't be any issues with my changes in `warp-transducer` but better be sure as it would be critical.

b-flo:
@sw005320 I'm merging the PR, FastEmit addition won't harm the default setup (`fastemit_lambda = 0.0`) so we're good to go.

PR #3597, opened 2021-09-18
Add a new english G2P

kan-bayashi:
Add a new English G2P based on VITS implementation.
(espeak + word boundary + split each phone into single char)


gafsd:
Will this have better pronunciation?

kan-bayashi:
Maybe but very slow.

PR #3599, opened 2021-09-18
Split test_integration.sh

ShigekiKarita:
I'd like to split test_integration.sh into three components
1. espnet/ and egs/
2. espnet2/ and egs2/
3. utils/

This is because the large one script is annoying especially when new changes are limited to each area. And we can easily know which components are broken or time consuming in CI.


kan-bayashi (approved):
Cool! LGTM

PR #3603, opened 2021-09-20
Add codecov config and flags.

ShigekiKarita:
I think our codecov with the default config is too strict for PR checking, i.e. negative coverage% always fails. So I'd like to add `threshold: 1%` to allow reasonable small coverage decrease < 1% (e.g. #3584). Because codecov always fails, nobody cares if the coverage changed 80% -> 50% recently (see #3600).

In this PR, I also tried https://docs.codecov.com/docs/flags to see coverage %s in detail.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -35,6 +35,9 @@ jobs:
             chainer-verssion: 6.0.0
             use-conda: false
     steps:
+      - uses: styfle/cancel-workflow-action@0.9.1
+        with:
+          access_token: ${{ github.token }}
       - uses: actions/checkout@master
       - uses: actions/cache@v1
         with:
@@ -69,8 +72,10 @@ jobs:
       - uses: codecov/codecov-action@v2
         with:
           flags: unittests
-      - name: reset coverage
-        run: rm .coverage
+      - name: coverage erase
+        run: |
+          source tools/activate_python.sh
+          coverage erase
 
       - name: install kaldi
         run: |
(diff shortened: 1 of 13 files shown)


ShigekiKarita:
I also fixed `ci scripts` to use `coverage run --append` and `coverage combine` instead of direct `python` call and `ln -s ${cwd}/.coverage .`, respectively.

.github/workflows/ci.yaml
```diff
@@ -59,19 +59,47 @@ jobs:
...
+
       - name: test utils
         run: ./ci/test_utils.sh
+      - uses: codecov/codecov-action@v2
+        with:
+          flags: test_utils
+      - name: coverage erase
+        run: |
+          source tools/activate_python.sh
+          coverage erase
```
kan-bayashi, line 89: Could you tell me what is happened with this command?

.github/workflows/debian9.yml
```diff
@@ -47,4 +47,3 @@ jobs:
         run: ./ci/test_integration_espnet1.sh
       - name: test espnet2 integration
         run: ./ci/test_integration_espnet2.sh
```
kan-bayashi, line 49: Do we need to use `codecov-action` for debian and centos test?

.github/workflows/ci.yaml
ShigekiKarita: As far as I know, it just deletes `$(pwd).coverage.*` (a mysql DB file). I think the command is better than `rm` because it is wrapped by the official command. https://coverage.readthedocs.io/en/coverage-5.5/cmd.html#cmd-erase

ShigekiKarita: And if we don't erase the file, the sub coverage per flags will be broken (overwrapped).

.github/workflows/debian9.yml
ShigekiKarita: Well, I'm still thinking about that. We might get more coverage if there are difference (but no). On the other hand, there are potential risks to collapse coverage if we accidentally have inconsistent codecov usage (e.g. flags, coverage erase) in the additional files. So I decided to remove codecov from other yml.

sw005320 (approved):
Thanks, @ShigekiKarita!

kan-bayashi (approved):

PR #3613, opened 2021-09-22
[WIP] HuggingFace upload support

Fhrozen:
This PR updates the code to upload models to HuggingFace:

- [ ] ESPnet
- [x] ESPnet2 ASR
- [ ] ESPnet2 others

On `TEMPLATE/asr.sh `, I tried to modify the minimum (using the previously compressed file and uncompress it on the huggingface git folder.

You can find a test model at: https://huggingface.co/Fhrozen/test_an4

@sw005320, let me know if this modification suits the upload or if it will be better to change the `espne2.bin.pack` code.
Also, I added a new stage (16) for this purpose, if the zenodo support will be deprecated, then I can move stage 16 into 15 and remove zenodo support.

### DIFF
--- a/egs2/TEMPLATE/asr1/asr.sh
+++ b/egs2/TEMPLATE/asr1/asr.sh
@@ -29,6 +29,7 @@ skip_data_prep=false # Skip data preparation stages.
 skip_train=false     # Skip training stages.
 skip_eval=false      # Skip decoding and evaluation stages.
 skip_upload=true     # Skip packing and uploading stages.
+skip_upload_hf=true  # Skip uploading to hugging face stages.
 ngpu=1               # The number of gpus ("0" uses cpu, otherwise use gpu).
 num_nodes=1          # The number of nodes.
 nj=32                # The number of parallel jobs.
@@ -91,6 +92,9 @@ asr_args=      # Arguments for asr model training, e.g., "--max_epoch 10".
                # Note that it will overwrite args in asr config.
 feats_normalize=global_mvn # Normalizaton layer type.
 num_splits_asr=1           # Number of splitting for lm corpus.
+
+# Upload model related
+hf_repo=
 
 # Decoding related
 use_k2=false      # Whether to use k2 based decoder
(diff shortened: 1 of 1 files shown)


egs2/TEMPLATE/asr1/asr.sh
```diff
@@ -1342,38 +1346,37 @@ fi
...
+        --asr_model_file "${asr_exp}"/"${inference_asr_model}" \
+        ${_opts} \
+        --option "${asr_exp}"/RESULTS.md \
+        --option "${asr_exp}"/RESULTS.md \
+        --option "${asr_exp}"/images \
+        --outpath "${packed_model}"
+fi
 
-
+if ! "${skip_upload}"; then
```
simpleoier, line 1379: Thanks for the PR.
Just a small suggestion. Would it be better to make the structure in this way?
```bash
upload_target="huggingface"  # default as huggingface
...

if ! "${skip_upload}"; then
    if [ ${stage} -le 14 ] && [ ${stop_stage} -ge 14 ]; then
        # Pack the models
    fi

    if [ ${stage} -le 15 ] && [ ${stop_stage} -ge 15 ]; then
        if [ ${upload_target} = "zenodo" ]; then
            # Upload to Zenodo
        else if [ ${upload_target} = "huggingface" ]; then
            # Upload to HF
        else
            echo "Please choose the upload destination from [zenodo, huggingface]"
        fi
    fi
fi
```

sw005320: I actually like the current @Fhrozen's solution, since we can easily switch zenodo or HF hub based on the stage.
We can also upload both zenodo/HF in the current form.
Also, in the future, we may completely move to HF hub (which I really want to based on the issue a number of people encountered in zenodo), then, we can just remove a zenodo stage.
Either is fine, but this is my preference.

sw005320:
Thanks!
Can you also upload the related documents?
I found that you may need to add some documents in 
https://github.com/espnet/espnet_model_zoo#huggingface
https://github.com/espnet/espnet/blob/08feae5bb93fa8f6dcba66760c8617a4b5e39d70/doc/espnet2_tutorial.md#start-from-a-specified-stage-and-stop-at-a-specified-stage

Basically, check https://github.com/espnet/espnet/search?q=skip_upload

PR #3616, opened 2021-09-22
Fix quantization

b-flo:
This PR fixes dynamic quantization for Transducer (and also update arguments doc). There are currently two issues that ends-up with an `AttributeError` or `RuntimeError` : 

1) With `quantize-config: [torch.nn.LSTM]` and any LSTM-based encoder
2) With `quantize-config: [torch.nn.Linear]` and most search algorithms (greedy, default, NSC, and mAES)

The first issue is due to `flatten_parameters` attribute. Looking at the [torch doc](https://pytorch.org/docs/stable/_modules/torch/nn/quantized/dynamic/modules/rnn.html#LSTM), it seems `torch.nn.quantized.dynamic.LSTM` doesn't have the attribute compared to `torch.nn.quantized.LSTM`. I suppose other models' encoders are impacted by this?
The second issue is due to passing tensors with dim() < 2 to `torch.nn.quantized.dynamic.Linear`. Typically, when we rely on non-batch beam (greedy or default) or during prefix search (default, NSC and mAES).

The fixes are straightforward, It's a bit ugly. Feel free to propose something else!
Also, I guess I added too many tests, some may be useless. I'll check that out.

### DIFF
--- a/espnet/asr/pytorch_backend/asr.py
+++ b/espnet/asr/pytorch_backend/asr.py
@@ -986,10 +986,7 @@ def recog(args):
         q_config = {torch.nn.Linear}
 
     if args.quantize_asr_model:
-        assert (
-            "transducer" not in train_args.model_module
-        ), "Quantization of transducer model is not supported yet."
-        logging.info("Use quantized asr model for decoding")
+        logging.info("Use a quantized ASR model for decoding.")
 
         dtype = getattr(torch, args.quantize_dtype)
         model = torch.quantization.quantize_dynamic(model, q_config, dtype=dtype)
@@ -1099,6 +1096,7 @@ def recog(args):
             score_norm=args.score_norm,
             softmax_temperature=args.softmax_temperature,
             nbest=args.nbest,
+            quantization=args.quantize_asr_model,
         )
 
     if args.batchsize == 0:
--- a/test/test_e2e_asr_transducer.py
+++ b/test/test_e2e_asr_transducer.py
@@ -391,3 +391,64 @@ def test_invalid_aux_transducer_loss_enc_layers():
 
     with pytest.raises(ValueError):
         E2E(idim, odim, train_args)
+
+
+@pytest.mark.parametrize(
+    "train_dic",
+    [
+        {},
+        {"etype": "vggblstm"},
+    ],
+)
+@pytest.mark.parametrize(
+    "recog_dic",
+    [
+        {},
+        {"beam_size": 2, "search_type": "default"},
+        {"beam_size": 2, "search_type": "alsd"},
+        {"beam_size": 2, "search_type": "tsd"},
+        {"beam_size": 2, "search_type": "nsc"},
+        {"beam_size": 2, "search_type": "maes"},
+    ],
+)
+@pytest.mark.parametrize(
+    "quantize_dic",
+    [
+        {"mod": {torch.nn.Linear}, "dtype": torch.qint8},
+        {"mod": {torch.nn.Linear}, "dtype": torch.float16},
+        {"mod": {torch.nn.LSTM}, "dtype": torch.qint8},
+        {"mod": {torch.nn.LSTM}, "dtype": torch.float16},
+        {"mod": {torch.nn.Linear, torch.nn.LSTM}, "dtype": torch.qint8},
+        {"mod": {torch.nn.Linear, torch.nn.LSTM}, "dtype": torch.float16},
+    ],
+)
+def test_dynamic_quantization(train_dic, recog_dic, quantize_dic):
+    idim, odim, ilens, olens = get_default_scope_inputs()
+
+    train_args = get_default_train_args(**train_dic)
+    recog_args = get_default_recog_args(**recog_dic)
+
+    model = E2E(idim, odim, train_args)
+    model = torch.quantization.quantize_dynamic(
+        model, quantize_dic["mod"], dtype=quantize_dic["dtype"]
+    )
+
+    beam_search = BeamSearchTransducer(
+        decoder=model.dec,
+        joint_network=model.transducer_tasks.joint_network,
+        beam_size=recog_args.beam_size,
+        lm=recog_args.rnnlm,
+        lm_weight=recog_args.lm_weight,
+        search_type=recog_args.search_type,
+        max_sym_exp=recog_args.max_sym_exp,
+        u_max=recog_args.u_max,
+        nstep=recog_args.nstep,
+        prefix_alpha=recog_args.prefix_alpha,
+        score_norm=recog_args.score_norm_transducer,
+        quantization=True,
+    )
+
+    with torch.no_grad():
+        in_data = np.random.randn(20, idim)
+
+        model.recognize(in_data, beam_search)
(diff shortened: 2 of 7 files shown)


b-flo:
Hum, it seems there are additional issues : 

- With `quantize-config: [torch.nn.LSTM]` and torch < 1.4
- With `quantize-dtype: float16` and torch < 1.5

The first issue is related to using a packed sequence with quantized LSTM (see code notes). For the second one, I'm not sure, I couldn't find anything useful as a reference.

From what I gather, all mentioned issues are not checked in the integration tests or unit tests. I suppose, outside the one related to Transducer `JointNetwork`, others ASR models should be impacted by these issues.

@sw005320 @ShigekiKarita Related topic: should we plan to bump the torch requirements for ESPnet1's next major version (if there is one)? I'm afraid it will become difficult to add some features and maintain all torch versions (1.3.1 - 1.9+) without overloading/impacting some parts of the current implementation.

ShigekiKarita:
> With quantize-dtype: float16 and torch < 1.5

> @sw005320 @ShigekiKarita Related topic: should we plan to bump the torch requirements for ESPnet1's next major version

Sounds good. Pytorch 1.5 was released [one year ago](https://pytorch.org/blog/pytorch-1-dot-5-released-with-new-and-updated-apis/). I believe many people already use it.

espnet/asr/pytorch_backend/asr.py
```diff
@@ -986,12 +987,32 @@ def recog(args):
...
+            and "lstm" in train_args.etype
+            and torch.nn.LSTM in q_config
+        ):
+            raise ValueError(
+                "Quantized LSTM in ESPnet is only supported with torch 1.4+."
+            )
+
+        # Dunno why but weight_observer from dynamic quantized module must have
+        # dtype=torch.qint8 with torch < 1.5 although dtype=torch.float16 is supported.
+        if args.quantize_dtype == "float16" and LooseVersion(
```
ShigekiKarita, line 1005: Can you test this new option in https://github.com/espnet/espnet/blob/master/ci/test_integration_espnet1.sh?

espnet/asr/pytorch_backend/asr.py
```diff
@@ -986,12 +987,32 @@ def recog(args):
...
+            torch.__version__
+        ) < LooseVersion("1.5.0"):
+            logging.warning(
```
ShigekiKarita, line 1008: Can you make this `ValueError` instead of warning because it is critical when someone compares performance? Of course, if you update the requirement to 1.5, you should remove this block.

ShigekiKarita (approved):

espnet/asr/pytorch_backend/asr.py
b-flo: Sorry, I'm a bit confused. If I add an integration test for `quantize-dtype: float16`, CI will fail because `ValueError `will be raised?
Ultimately, I would need to add a unit test but we don't directly test `recog(...)` from `asr.py`...

klknn: I see. So let's go pytorch 1.5+ for the integration test? @sw005320 

I agree that the unittest is very useful, and be careful with pytest.timeout.

klknn:
Because we implemented several new features in --api=v2, can you also update https://github.com/espnet/espnet/blob/master/espnet/asr/pytorch_backend/recog.py

ShigekiKarita:
(Sorry I used my different account in the previous reply)

Can you update https://github.com/espnet/espnet/blob/master/espnet/asr/pytorch_backend/recog.py#L22

espnet/asr/pytorch_backend/asr.py
ShigekiKarita: > CI will fail because ValueError will be raised?

Shall we omit pytorch 1.5- in the CI? @sw005320 

> unittest

Sounds good, and be careful with pytest.timeout.

b-flo:
> Can you also update recog_v2 https://github.com/espnet/espnet/blob/master/espnet/asr/pytorch_backend/recog.py#L22?

Sure, I'll do!
Btw, what about the first issue (with `quantize-config: [torch.nn.LSTM]` and any LSTM-based encoder)? If I'm not mistaken other models are impacted and I didn't fix it RNN/RNNP encoders outside Transducer model. I'm not too fan of skipping flatten_parameters() in these cases but I'm also not sure we can fix it differently.

espnet/asr/pytorch_backend/asr.py
b-flo: > Sounds good, and be careful with pytest.timeout

Oh btw, how can I be sure which value I should assign to `pytest.mark.execution_timeout(...)`? I mean, I'm setting it up according to my local settings but CI rely on different hardware.

test/test_e2e_asr_transducer.py
```diff
@@ -391,3 +395,86 @@ def test_invalid_aux_transducer_loss_enc_layers():
...
+    ):
+        # In recognize(...) from asr.py we raise ValueError however
+        # AssertionError is originaly raised by torch.
+        with pytest.raises(AssertionError):
+            model = torch.quantization.quantize_dynamic(
+                model,
+                quantize_dic["mod"],
+                dtype=quantize_dic["dtype"],
+            )
+        pytest.skip("Skip rest of the test after checking AssertionError")
```
b-flo, line 449: A `pytest.skip` is needed here otherwise next `pytest.raises(...)` may fail because `quantize_dynamic` was not applied.
Note that I'm fully aware this is bad practice to do that however this is only a temporary measure, it will be updated after we bump torch version.

sw005320:
> > With quantize-dtype: float16 and torch < 1.5
> 
> > @sw005320 @ShigekiKarita Related topic: should we plan to bump the torch requirements for ESPnet1's next major version
> 
> Sounds good. Pytorch 1.5 was released [one year ago](https://pytorch.org/blog/pytorch-1-dot-5-released-with-new-and-updated-apis/). I believe many people already use it.

I agree.
Make Pytorch >=1.5 as default should be no problem.

espnet/asr/pytorch_backend/asr.py
ShigekiKarita: Because it is a kinda integration test, you can set a large value enough to pass CI. For example, 50 sec is safe:
https://github.com/espnet/espnet/blob/08feae5bb93fa8f6dcba66760c8617a4b5e39d70/test/espnet2/tasks/test_abs_task.py#L108-L109

b-flo:
OK, I added the checks to API v2. I guess we can keep it until we bump torch version. However : 

- I didn't test nor add checks for quantization in others implementations (LM, TTS, ST, etc). Depending on the PR for torch version, I'll add them in the meantime.
- The first issue (with quantize-config: [torch.nn.LSTM] and any LSTM-based encoder) happen even with to torch 1.9.  We should maybe think about some other fix/implementation: 1) I'm not fan of overloading some computation parts with if-else, and 2) we could remove calls to `flatten_parameters()` and it would also fix some issues we have with multi-GPUs (cc @Desh in case the conversation go somewhere). Obviously, it will impact training time.

sw005320:
Thanks, @b-flo!

PR #3619, opened 2021-09-24
Add snips hubert feature training

yuekaizhang:
Test accuracy for intent classification is 89.2% using hubert feature. 
Screen Shot 2021-09-24 at 5 31 54 PM
. 

Would try Xuankai's hubert asr model perhaps in another PR.


sw005320:
Thanks, @yuekaizhang!
LGTM.

@siddhu001, could you also check it?
We can merge this version anyway if there is no measure issue.

siddhu001:
We need to compute F1 instead of accuracy to compare with SOTA results. That would require some changes in local/score.sh

yuekaizhang:
> Great work, @yuekaizhang!
> it is great that we are very near SOTA accuracy i.e. around 91.72
> Screenshot 2021-09-24 at 5 47 40 PM
> 
> (P.S. : The paper says F1 but I think it is computing accuracy since micro F1=accuracy for classification task)
> 
> @yuekaizhang can you take a look here -https://github.com/snipsco/snips-nlu-metrics just to see if our evaluation metric is in sync with one used in paper.

I have checked the above link. They used Macro-F1 which average all F1 scores rather Micro-F1. Let me re-calculate the Macro-F1.

siddhu001:
It might make sense to use their metric compute script - https://github.com/snipsco/snips-nlu-metrics so as to ensure that there is no discrepancy

yuekaizhang:
> It might make sense to use their metric compute script - https://github.com/snipsco/snips-nlu-metrics so as to ensure that there is no discrepancy

It would be a little tricky to use or import their metrics. However, it would make sense to utilize their metric for the whole slu project. I directly implement one according to their code, only for intent classification.

Train using hubert feature: Macro-f1 test set 87.4
Fine-tune using hubert pretrained ASR: Macro-f1 test set 89.1

PR #3621, opened 2021-09-24
EEND-EDA model for diarization task

YushiUeda:
This PR is for adding encoder-decoder based attractor calculation for end-to-end neural diarization (EEND-EDA) which estimates the number of speakers along with speaker diarization (see https://www.isca-speech.org/archive/interspeech_2020/horiguchi20_interspeech.html).
This PR also includes configuration files and preliminary experimental results using mini librispeech dataset.


espnet2/tasks/diar.py
```diff
@@ -74,6 +82,15 @@
     type_check=AbsDecoder,
     default="linear",
 )
+attractor_choices = ClassChoices(
+    "attractor",
+    classes=dict(
+        rnn=RnnAttractor,
```
ftshijt, line 88: Please also add a `None` case for models without attractor

espnet2/tasks/diar.py
```diff
@@ -74,6 +82,15 @@
...
+        rnn_nograd=RnnAttractorNoGrad,
```
ftshijt, line 89: I did not quite understand where this will be used. Could you show me how it is functioning? If you just want to freeze the params when training sometimes, you can just forward with `no_grad()`

egs2/mini_librispeech/diar1/conf/train_diar_eda.yaml
```diff
@@ -0,0 +1,54 @@
+model_conf:
```
ftshijt, line 1: Please add more information to the models (e.g., related papers). EDA and SA are familiar around us but not for all.

ftshijt (commented):
Well done! It's already in a good shape now. I just have some minor suggestions.

espnet2/tasks/diar.py
YushiUeda: It seems that we don't explicitly add `None` case in the choices (see https://github.com/espnet/espnet/blob/master/espnet2/tasks/asr.py). Instead, I modified `default="rnn"` to `default=None`.

espnet2/tasks/diar.py
YushiUeda: `rnn_nograd` corresponds to equations (23) and (24) in https://arxiv.org/pdf/2106.10654.pdf. My intension of adding this option is to easily use it for future experiments. However, if there is a better way to do this, I will be glad to fix it!

egs2/mini_librispeech/diar1/conf/train_diar_eda.yaml
YushiUeda: Added information in `train_diar_eda.yaml` and `espnet2/diar/espnet_model.py`.

YushiUeda:
@ftshijt thanks so much for you review and helpful comments!

espnet2/tasks/diar.py
ftshijt: Yeah, that's what I'm expecting! Thanks

espnet2/tasks/diar.py
ftshijt: I think you can add an option to the `RnnAttractor` class to enable no_grad by https://pytorch.org/docs/stable/generated/torch.no_grad.html Because at least for the structure, it's exact the same right?

egs2/mini_librispeech/diar1/conf/train_diar_eda.yaml
ftshijt: Very cool! thanks

espnet2/tasks/diar.py
YushiUeda: Thanks for the suggestion! I totally agree. Added an option to `RnnAttractor`.

YushiUeda:
Is this error something that I can fix on my side? https://github.com/espnet/espnet/pull/3621/checks?check_run_id=3738366149
It doesn't seem to be related to the codes per se.

ftshijt:
LGTM! You may add some test functions for your implementation in the following PR.

PR #3635, opened 2021-10-01
Change installation exception for latest pytorch version to error

lumaku:
The current `setup.py` raises a `NotImplementedError` exception when the newest Pytorch version is detected.

I usually have the recent Pytorch version on my system and install Espnet from a system package, and with that, `setup.py` throws an exception, even if just running `python setup.py --version`. Espnet seems to run fine on the newest Pytorch version, I am not aware of any compability problems (However, I only tested it for ASR).

I thus propose to change the exception to an error message.

@sw005320 @kamo-naoyuki 

The system package can be found here:
https://aur.archlinux.org/packages/python-espnet-git/

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -3,6 +3,7 @@
 """ESPnet setup script."""
 
 import os
+import sys
 
 from distutils.version import LooseVersion
 from setuptools import find_packages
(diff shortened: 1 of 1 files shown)


kamo-naoyuki:
Thanks.

Usually, I do some actions when  pytorch is updated: See #3122

kamo-naoyuki:
I'll do them.

lumaku:
I see. Thanks!

Regarding possible compatible versions: I successfully tested Espnet 2 on my machine with Pytorch 1.9.1 and CUDA 11.4....

From my point of view, the system package does not use the Python environment from `tools/`. Thus, in this context, to be compatible with future software versions, it would only be necessary that `setup.py` accepts newer Pytorch versions.

PR #3636, opened 2021-10-01
Change setup.py for pytorch1.9.1

kamo-naoyuki:


kamo-naoyuki:
K2 wasn't updated for pytorch1.9.1 yet, so in this PR, I updated setup.py only.

I'll do update our ci if k2 is updated.

PR #3648, opened 2021-10-11
Update TTS results

kan-bayashi:


gafsd:
Do you have comparison wav for new g2p?

kan-bayashi:
Here you are.
https://drive.google.com/drive/folders/1acGIz4mV-GkKltaz1jxcYBsKSTYaIgfQ?usp=sharing

PR #3653, opened 2021-10-11
Add Hubert pretrained ASR in FSC SLU

siddhu001:


simpleoier:
The error comes from the configuration files you specified. In the default config, you use s3prl frontend.
The test environment is for `linter_and_test (ubuntu-18.04, 3.7, 1.5.1, 6.0.0, 1.2.dev20210726+cpu, false)`
However, s3prl is installed when `pytorch > 1.7`. Thus, the s3prl error comes.
Can you follow this to fix the error @siddhu001 ?

PR #3655, opened 2021-10-12
fix CTCPrefixScore init of r according alg2

zh794390558:
fix CTCPrefixScore init of r according alg2, line 6

image

### DIFF
--- a/espnet/nets/ctc_prefix_score.py
+++ b/espnet/nets/ctc_prefix_score.py
@@ -319,7 +319,7 @@ class CTCPrefixScore(object):
             r[0, 0] = xs[0]
             r[0, 1] = self.logzero
         else:
-            r[output_length - 1] = self.logzero
+            r[0] = self.logzero
 
         # prepare forward probabilities for the last label
         r_sum = self.xp.logaddexp(


sw005320:
Thanks!
@takaaki-hori, could you check this?

takaaki-hori:
Hi, thanks for checking. Although the code does not exactly follow Algorithm 2, we don't have to change it because we can assume r_t(h)=0 for t < |h| in CTC forward computation (Note: we assume here that index t starts with 0).
The purpose of this difference is to reduce the number of for-loops https://github.com/espnet/espnet/blob/2dab44dbb013517f96a490090e75ede8c957f809/espnet/nets/ctc_prefix_score.py#L340
where we start to accumulate r_t(h) from t=|h| and iterate r_t(h) = (r_{t-1}(h) + ...) to T-1, avoiding accumulating zeros for t=1~|h|-1.
Thus, we need to set r_{|h|-1}(h) = 0, i.e., `r[output_length-1] = logzero`, for initialization.
This is just for reducing the computation. Maybe it is better to add some comments to the code for clarification.

PR #3668, opened 2021-10-16
Do stft with librosa when neither MKL nor CUDA is available.

CTinRay:
Executing `torch.stft()` throws error `fft: ATen not compiled with MKL support` on ARM devices (Raspberry Pi 4). To circumvent this problem, librosa can be used when neither MKL nor CUDA is available.

### DIFF
--- a/espnet2/layers/stft.py
+++ b/espnet2/layers/stft.py
@@ -106,6 +106,12 @@ class Stft(torch.nn.Module, InversibleInterface):
                 stft_kwargs["return_complex"] = False
             output = torch.stft(input, **stft_kwargs)
         else:
+            if self.training:
+                raise NotImplementedError(
+                    "stft is implemented with librosa on this device, which does not "
+                    "support the training mode."
+                )
+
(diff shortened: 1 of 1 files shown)


sw005320:
Thanks!
Can you fix the CI error?
https://github.com/espnet/espnet/pull/3668/checks?check_run_id=3925833665#step:8:9

kamo-naoyuki:
> It passed the test, but I am not sure whether it covers all the use case of torch.stft in espnet.

Thanks, I think it's enough.

> Thanks for the good suggestion! Should I raise a NotImplementedError?

Please use `NotImplementedError` .

Emrys365:
> @Emrys365, could I ask you to make sure that this change is OK?
> Can you review this PR?

Yes, I actually reviewed it. It looks good to me, and as I know, torch.stft is designed to be compatible with librosa.stft.

kamo-naoyuki:
Thanks. @CTinRay  I'll merge.

PR #3685, opened 2021-10-22
CATSLU-MAPS: Added recipe

SujaySKumar:


ftshijt:
Hi Sujay, could you follow the instructions in CI tests? Seems that you have modified the `slurm.conf` which throws an error

SujaySKumar:
I had removed the `default gpu=0` option according to a previous code review comment. I will add it back

ftshijt:
Thanks!

PR #3703, opened 2021-10-26
Bug fixed for averaging models

ranchlai:
The current way of model averaging has two problems: 
1. It is unable to produce averaging models until the end of training, i.e., until reaching the maximum epochs. This is a problem when people want to do some testing before training ends (e.g., when training takes much longer than expected)
2. Setting --keep_nbest_models 10 works ok,   but setting --keep_nbest_models 3, 5, 10 works the same as  --keep_nbest_models 10.   Multiple nbest numbers should be supported as seen from current design, but it fails to save 3best, 5best models. 

This PR fixed the above two problems.


kamo-naoyuki:
> The current way of model averaging has two problems:
> 
> 1. It is unable to produce averaging models until the end of training, i.e., until reaching the maximum epochs. This is a problem when people want to do some testing before training ends (e.g., when training takes much longer than expected)

@sw005320 @kan-bayashi What do you think? I think this is useful, but the read-write cost for every epoch increases.


> 2. Setting --keep_nbest_models 10 works ok,   but setting --keep_nbest_models 3, 5, 10 works the same as  --keep_nbest_models 10.   Multiple nbest numbers should be supported as seen from current design, but it fails to save 3best, 5best models.
> 

Oh, thanks, this is a bug. Maybe, I mistakenly changed trainer.py in the past.

kan-bayashi:
This is very useful but I think it is better to add the option to skip the creation of  averaged model since some tts model performs more than hundreds of epochs.

ranchlai:
> This is very useful but I think it is better to add the option to skip the creation of averaged model since some tts models perform more than hundreds of epochs.

Thank you for the review.

I'd to add a new parameter, e.g., --nbest_averaging_interval 3 means performing averaging at every 3 epochs, and 0(default) means no intermedia averaging will occur.  At the end of the training,  model averaging is perform regardless of this interval value (unless keep_nbest_models is set to 1, meaning no actual averaging). 

If ok, I will update the PR soon. Thanks!

kamo-naoyuki:
> I'd to add a new parameter, e.g., --nbest_averaging_interval 3 means performing averaging at every 3 epochs, and 0(default) means no intermedia averaging will occur. At the end of the training, model averaging is performed regardless of this interval value (unless keep_nbest_models is set to 1, meaning no actual averaging).
> 

Thanks! This is completely as same as I thought!

sw005320:
Sorry for my late response.
Yes, adding an option is a good idea.
Thanks, @ranchlai

ranchlai:
> Sorry for my late response. Yes, adding an option is a good idea. Thanks, @ranchlai

I have already updated the PR, but the CI failed at early stage, please help~~

kamo-naoyuki:
Sorry for bothering you. Please merge the current master to fix the CI error,

kamo-naoyuki:
LGTM, thank you for contributing!

PR #3707, opened 2021-10-27
Refactor setup.py

kamo-naoyuki:
I removed the following block from setup.py, 

https://github.com/espnet/espnet/blob/5c21f63e45e0961a5d817017c282b0cafd68a3aa/setup.py#L88-L130

because it's a dirty way and I created three scripts instead.

- intall_fairscale.sh
- install_torchaudio.sh
- install_torch_optimizer.sh

( We can't append these modules to the espnet requirements directly because the minimum version requirement of pytorch is higher than our requirement, i.e. pytorch>=1.3)

I also added `pip install "espnet[all]"` to install these optional modules( `all` is prepared for users only using inference mode,  so it's not used in Makefile, but using the above scripts in Makefile). `pip install "espnet[all]"` also installs `espnet_model_zoo`.

Note that I also added `pytorch>=1.3.0` to the requirements of espnet, i.e. `pip install espnet` also tries to install pytorch if it's not installed.  Maybe this is more convenient for non developer people.


kamo-naoyuki:
I'll refactor further more. Please wait pytorch1.10

PR #3722, opened 2021-10-30
installers/install_torch.sh

kamo-naoyuki:
- I created installers/install_torch.sh
- I removed installers/install_torchaudio.sh and check_pytorch_cuda_compatibility.py. They are integrated in  installers/install_torch.sh


kan-bayashi:
Memo: Wait for fixing
https://github.com/pytorch/audio/issues/1948

PR #3739, opened 2021-11-02
Slurp entity classification

siddhu001:


siddhu001:
@ftshijt let me know your comments on this. I will upload the model to zenodo soon. 

One main question that I had is that I borrow most of evaluation code from external repository (https://github.com/pswietojanski/slurp/tree/master/scripts/evaluation). How should I acknowledge this?

sw005320:
> One main question that I had is that I borrow most of evaluation code from external repository (https://github.com/pswietojanski/slurp/tree/master/scripts/evaluation). How should I acknowledge this?

This is a good question. We try to acknowledge it as much as possible.
First, we can put their link, citation, and acknowledgment on the recipe readme.
Then, you can put some comments in the source code as well and clarify how you borrow the code.

ftshijt:
Your implementation is fine. I'm just curious if it's possible we directly `git clone` the repo for this recipe. In that case, we could always sync to their latest update (if any)

siddhu001:
@ftshijt We can not git clone because I had to make some small change based on my tokenisation of transcripts but as @swatanabe mentioned I am going to acknowledge it everywhere I use their code.

ftshijt:
> @ftshijt We can not git clone because I had to make some small change based on my tokenisation of transcripts but as @swatanabe mentioned I am going to acknowledge it everywhere I use their code.

I see, that's fine as well. Please fix the CI, then I will merge it

PR #3742, opened 2021-11-03
Fix SWBD Data Prep Bug

brianyan918:
I found that stage 7 (LM training) fails in the current recipe due to some duplicates in the Fisher training data. This added line should fix that problem.


brianyan918:
The CI failed with some installation error. Looks like it tried to install torch 1.10.

sw005320:
> The CI failed with some installation error. Looks like it tried to install torch 1.10.

We'll fix this issue later. So, please ignore this. If the other CI check is passed, it must be no problem.

sw005320:
Thanks, @brianyan918!

PR #3746, opened 2021-11-04
Add Multilingual TEDx recipe

hirofumi0810:
https://arxiv.org/pdf/2102.01757.pdf


ftshijt:
LGTM! I do not have other comments. Many thanks

PR #3751, opened 2021-11-05
Add option to tile up s3prl features

ftshijt:
Given that some upstream models in s3prl have 20ms shift, CTC cannot be working for some specific languages because they have a relatively long transcription to the audio. In case of that, we could tile up the upstream features so as to match the fbank precision with 10ms.


simpleoier:
LGTM!

PR #3764, opened 2021-11-09
Fix shell check error

kamo-naoyuki:


kamo-naoyuki:
I'll merge this PR and I'll also fix the error from fairscale.

Issue #3781, opened 2021-11-15
Index error in Transducer beam search ALSD


maxwellzh:
**Describe the bug**
The index of `(t, u)` is wrong. Since indices of `u` and `i` are beginning with  0, the `t` should be `t = i - u`. In current impl, `t` is beginning with 1, and the encoder output at `t=0` is never used.
https://github.com/espnet/espnet/blob/3c30eab699d8c3a0c61331e3cac38050345527de/espnet/nets/beam_search_transducer.py#L467-L469

b-flo:
Hi,

Seems right to me, thanks for catching that! Could you make a PR, please? If not, I will do it at the end of the week.

maxwellzh:
@b-flo I have make the PR. Only one line changes. It should be easy to review.

b-flo:
Fixed by #3794.

PR #3794, opened 2021-11-16
fix rnn-t ALSD beam search index bug

maxwellzh:
Fix bug in #3781 . @b-flo


b-flo (approved):

b-flo:
Thanks for the fix and PR!

PR #3810, opened 2021-11-22
Some minor fixes for commonvoice recipe

ftshijt:


egs2/commonvoice/asr1/conf/tuning/train_asr_transformer.yaml
```diff
@@ -48,4 +48,4 @@ best_model_criterion:
    - max
 keep_nbest_models: 10
 
-init: chainer
+init: xavier_uniform
```
sw005320, line 51: Maybe you can leave some comments that this is very important?

sw005320 (approved):

egs2/commonvoice/asr1/conf/tuning/train_asr_transformer.yaml
ftshijt: Sounds good

PR #3819, opened 2021-11-28
Update MISP challenge ASR baseline and add AVSR baseline

neillu23:
For AVSR, we use feats_type=extracted for the video features, and we have the wav.scp.
Generating dummy wav.scp from cmvn.scp would cause id mismatch error for my case because the keys in wav.scp/feats.scp/cmvn.scp are not the same levels(sentences/segments/speakers).
Can we add an option that uses the extracted mode but skip L546 in asr.sh?
https://github.com/espnet/espnet/blob/master/egs2/TEMPLATE/asr1/asr.sh#L545-L546


egs2/misp2021/avsr1/extractor/LICENSE
```diff
@@ -0,0 +1,33 @@
+License
```
sw005320, line 1: Did you get these files from some repo?
If so, 
1. please add such information
2. how about including `git clone` of this repo in the data preparation stage instead of including the copy of them?

sw005320 (changes requested):
- the performance does not look great. WER looks like it's not working. Can you tell me why? Maybe, we need to improve this.
- `egs2/misp2021/avsr1/extractor` part seems to come from the other repo. These are huge files and I recommend you not to include them. Instead, please download them in the data preparation stage.

neillu23:
Thanks for your comments! I will check the redundant lines.
The high WER comes from the calculation ways. Most of the sentences are considered one word and calculated WER by sentence level. For example, this sentence only has two characters errors, but the WER is 100%.
CER:
 
WER:

How do we calculate the correct one, or could we remove the WER?

The extractor is from another repo, but Hang said there are some issues in the original repo, and the maintainer didn't reply to his email. So he decided to copy this in his repo. So maybe we can clone[ MISP 2021 baseline repo](https://github.com/mispchallenge/misp2021_baseline) and use it?

ftshijt:
> Thanks for your comments! I will check the redundant lines. The high WER comes from the calculation ways. Most of the sentences are considered one word and calculated WER by sentence level. For example, this sentence only has two characters errors, but the WER is 100%. CER:  WER:  How do we calculate the correct one, or could we remove the WER?
> 
> The extractor is from another repo, but Hang said there are some issues in the original repo, and the maintainer didn't reply to his email. So he decided to copy this in his repo. So maybe we can clone[ MISP 2021 baseline repo](https://github.com/mispchallenge/misp2021_baseline) and use it?

Understand! It's indeed meaningless to calculate WER for Mandarin corpus, how about we just remove it?

ftshijt (commented):
Looks good to me in general. But could you try to follow Shinji's suggestion of using `git clone` instead of copying the code from other repo, so that we can always follow their latest implementation without modifying our base code? In addition, there are many unnecessary files related to copying.

neillu23:
I have removed the extractor in our codebase and the redundant results. I clone the [MISP 2021 baseline repo](https://github.com/mispchallenge/misp2021_baseline) instead and only keep its extractor.

ftshijt:
Everything looks cool! Last request: could you add a brief intro about the recipe (how to use it, or the pipelines)? Cause it's quite different from other asr recipes given its complex preparation.

sw005320 (approved):
LGTM!

sw005320:
Thanks a lot!

neillu23:
Thank you! I will add some introduction to the recipe.

PR #3827, opened 2021-11-30
fix readme

seastar105:
fix different abbreviation of korean dataset kss
ko -> kr


ftshijt:
Thanks for your update!

PR #3839, opened 2021-12-03
update s3prl frontend w.r.t. recent modification in s3prl interface

simpleoier:
Update: Setting `layerdrop` before `Featurizer initialization`.
The error is from this [part](https://github.com/s3prl/s3prl/blob/3854f33867f3136ab97f5389578f4355ef6c660e/s3prl/upstream/interfaces.py#L214).


sw005320:
Thanks!

PR #3856, opened 2021-12-09
Implement self-conditioned CTC

komatta-san:
This implements self-conditioned CTC for Transformer and Conformer encoders.
https://arxiv.org/abs/2104.02724

### DIFF
--- a/espnet/nets/pytorch_backend/conformer/encoder.py
+++ b/espnet/nets/pytorch_backend/conformer/encoder.py
@@ -94,7 +94,7 @@ class Encoder(torch.nn.Module):
         stochastic_depth_rate=0.0,
         intermediate_layers=None,
         ctc_softmax=None,
-        odim=None,
+        conditioning_layer_dim=None,
     ):
         """Construct an Encoder object."""
         super(Encoder, self).__init__()
@@ -233,7 +233,7 @@ class Encoder(torch.nn.Module):
         self.use_conditioning = True if ctc_softmax is not None else False
         if self.use_conditioning:
             self.ctc_softmax = ctc_softmax
-            self.conditioning_layer = torch.nn.Linear(odim, attention_dim)
+            self.conditioning_layer = torch.nn.Linear(conditioning_layer_dim, attention_dim)
 
     def forward(self, xs, masks):
         """Encode input sequence.
--- a/espnet/nets/pytorch_backend/conformer/encoder.py
+++ b/espnet/nets/pytorch_backend/conformer/encoder.py
@@ -71,6 +71,7 @@ class Encoder(torch.nn.Module):
     def __init__(
         self,
         idim,
+        odim,
         attention_dim=256,
         attention_heads=4,
         linear_units=2048,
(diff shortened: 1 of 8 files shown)


sw005320:
This https://github.com/espnet/espnet/runs/4467881330?check_suite_focus=true#step:8:36 does not seem to be related to this PR.
@Emrys365, do you have any idea?

Emrys365:
I think it is related to the new positional argument `odim` added to the transformer encoder: https://github.com/espnet/espnet/blob/3b96776cb75afcdc3f7ae0ff0f227ce156f88271/espnet/nets/pytorch_backend/transformer/encoder.py#L84-L88

The TransformerSeparator is dependent on the base transformer implementation. So it should also be modified to match the new interface.

sw005320:
Oh, yes.
You're right.
@komatta-san, can you fix it?

komatta-san:
I modified `odim` to be a named argument and moved it to the last position.

espnet/nets/pytorch_backend/conformer/encoder.py
```diff
@@ -93,6 +93,8 @@ def __init__(
         padding_idx=-1,
         stochastic_depth_rate=0.0,
         intermediate_layers=None,
+        ctc_softmax=None,
+        odim=None,
```
pengchengguo, line 97: I prefer to use `conditioning_layer_dim` instead of `odim` here. Although the output dimension of CTC is vocabulary size (`odim`), using another name like `conditioning_layer_dim` may reduce ambiguity. It is okay to keep `odim`.

pengchengguo (commented):
Thanks, @komatta-san.
Most of the codes are good to me except for some minor suggestions. Besides, please add some CI tests for the new components.

espnet/nets/pytorch_backend/conformer/encoder.py
```diff
@@ -261,10 +267,22 @@ def forward(self, xs, masks):
...
+                    if self.normalize_before:
+                        encoder_output = self.after_norm(encoder_output)
+
                     intermediate_outputs.append(encoder_output)
 
+                    if self.use_conditioning:
+                        intermediate_result = self.ctc_softmax(encoder_output)
+
+                        if isinstance(xs, tuple):
+                            x, pos_emb = xs[0], xs[1]
```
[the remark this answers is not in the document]
komatta-san: I was trying to write it simply as well. But tuple `xs[0]` is immutable, so it can not be written like your suggestion.

espnet/nets/pytorch_backend/conformer/encoder.py
komatta-san: I will change it to `conditioning_layer_dim`.

komatta-san:
@pengchengguo 
Thank you for your suggestion. I modified the name of `odim` to `conditioning_layer_dim`.
For tests, I am not sure how to write it correctly, but I have added lines to  `test/test_e2e_asr_{transformer,conformer}.py`.

sw005320:
Thanks, @komatta-san!

PR #3864, opened 2021-12-15
fix the sort order in select_k_expansions()

freewym:
Just come across this function and think it is probably supposed to sort in descending order of logprob.

Correct me if it is wrong


b-flo (approved):
You're right, good catch! We can merge it.

(Note for me: it shouldn't change results for most models but it's better to double-check each mAES config parameters.)

PR #3869, opened 2021-12-15
Fix cache for Transducer search strategies + doc

### LINKED ISSUE #3863: Incorrect usage of the `cache` in the method `score()` for transducer beam search
See the code below
https://github.com/espnet/espnet/blob/ae635955ff79efde9c4df007bb8324fea386c169/espnet/nets/pytorch_backend/transducer/rnn_decoder.py#L180-L183

It concatenates all the integers in `yseq` and uses it as a key into the `cache`.

The issue is that it places no separators during the concatenation. The consequence is that different `yseq` may map to the same key. For instance, both `[1, 11, 1, 2]` and `[1, 1, 1, 12]` map to the same key `11112`.
```python3
>>> ''.join(list(map(str, [1, 11, 1, 2])))
'11112'
>>> ''.join(list(map(str, [1, 1, 1, 12])))
'11112'
```

b-flo:
Closes #3863 + minor fix/improve of Transducer-related doc.

### DIFF
--- a/README.md
+++ b/README.md
@@ -59,15 +59,17 @@ and also follows [Kaldi](http://kaldi-asr.org/) style data processing, feature e
     - VGG2L (RNN/custom encoder) and Conv2D (custom encoder) bottlenecks.
   - Search algorithms:
     - Greedy search constrained to one emission by timestep.
-    - Default beam search algorithm without prefix search.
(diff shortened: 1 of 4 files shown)


b-flo:
Manually merging, I wanted to verify something before that.

Issue #3803, opened 2021-11-18
An error occurred in the ASR task using the streaming conformer config on ESPnet2


lin-nerd:
- python=3.8.5
- torch=1.10.0+cu102
- torch cuda=10.2
- espnet=0.10.5a1

I want to use streaming conformer config for my dataset of the ASR task, but this error occurs after the first epoch training.
The log is:
[user] 2021-11-17 21:07:07,236 (trainer:668) INFO: 1epoch:train:26767-28253batch: iter_time=7.281e-05, forward_time=0.133, loss=150.549, loss_att=116.794, loss_ctc=229.310, acc=0.325, backward_time=0.169, optim_step_time=0.031, optim0_lr0=4.585e-04, train_time=0.425
[user] 2021-11-17 21:17:35,754 (trainer:668) INFO: 1epoch:train:28254-29740batch: iter_time=2.378e-04, forward_time=0.133, loss=142.284, loss_att=110.287, loss_ctc=216.944, acc=0.327, backward_time=0.166, optim_step_time=0.031, optim0_lr0=4.833e-04, train_time=0.422
Traceback (most recent call last):
  File "/home/user/espnet/tools/anaconda/envs/espnet/lib/python3.8/runpy.py", line 194, in _run_module_as_main
    return _run_code(code, main_globals, None,
  File "/home/user/espnet/tools/anaconda/envs/espnet/lib/python3.8/runpy.py", line 87, in _run_code
    exec(code, run_globals)
  File "/home/user/espnet/espnet2/bin/asr_train.py", line 23, in <module>
    main()
  File "/home/user/espnet/espnet2/bin/asr_train.py", line 19, in main
    ASRTask.main(cmd=cmd)
  File "/home/user/espnet/espnet2/tasks/abs_task.py", line 1013, in main
    cls.main_worker(args)
  File "/home/user/espnet/espnet2/tasks/abs_task.py", line 1305, in main_worker
    cls.trainer.run(
  File "/home/user/espnet/espnet2/train/trainer.py", line 293, in run
    cls.validate_one_epoch(
  File "/home/user/espnet/tools/anaconda/envs/espnet/lib/python3.8/site-packages/torch/autograd/grad_mode.py", line 28, in decorate_context
    return func(*args, **kwargs)
  File "/home/user/espnet/espnet2/train/trainer.py", line 711, in validate_one_epoch
    retval = model(**batch)
[...]

PR #3880, opened 2021-12-19
Fixing issues about streaming Transformer/Conformer training

D-Keqi:
This PR is to fix the streaming Transformer/Conformer training issues mentioned by https://github.com/espnet/espnet/issues/3803


espnet2/asr/encoder/contextual_block_conformer_encoder.py
```diff
@@ -208,17 +208,19 @@ def forward(
         ilens: torch.Tensor,
         prev_states: torch.Tensor = None,
         is_final=True,
+        infer=False,
```
sw005320, line 211: Can you come up with a better variable name?
`infer` is a bit ambiguous.
`infer_mode`?

espnet2/asr/encoder/contextual_block_conformer_encoder.py
```diff
@@ -208,17 +208,19 @@ def forward(
...
         Args:
             xs_pad: input tensor (B, L, D)
             ilens: input length (B)
             prev_states: Not to be used now.
+            infer: whether to be used for inference.
         Returns:
             position embedded tensor and mask
         """
-        if self.training or xs_pad.size(0) > 1:
+        if self.training or not infer:
```
lumaku, line 223: @D-Keqi 
You introduced here an additional parameter that indicates training mode.
Usually, the pythonic way of signalling that the module is used for evaluation and not for training is by calling [`module.eval()`](https://pytorch.org/docs/stable/generated/torch.nn.Module.html#torch.nn.Module.eval). This then sets `self.training` to False.
I I read correctly, `eval()` was already called in `espnet2/bin/asr_inference_streaming.py`.

D-Keqi: Thanks! During training and validation, the train_forward should be used; and during decoding, the forward_infer should be used. The additional parameter "infer" is actually employed to distinguish between validation and inference.

espnet2/asr/encoder/contextual_block_conformer_encoder.py
D-Keqi: OK! I think "infer_mode" is good.

espnet2/asr/encoder/contextual_block_conformer_encoder.py
sw005320: Are there other solutions or can we add more precise comments?
This logic is complicated, and it would have a risk of introducing further errors.

Also, can you make a unit/integration test of this mode?
Otherwise, this part would not be stable.

D-Keqi: > Are there other solutions or can we add more precise comments? This logic is complicated, and it would have a risk of introducing further errors.
> 
> Also, can you make a unit/integration test of this mode? Otherwise, this part would not be stable.

Before making a PR, I always make a test to make sure the code works (at least for me). But it seems that previous implementations do not work for everyone. Inspired by https://github.com/espnet/espnet/issues/3803, I have fixed another issue and I think this version will work. And I have added more comments for clarity.

sw005320: Can you make a test to pass this mode?
You can find an example in 
https://github.com/espnet/espnet/tree/master/test/espnet2
or
https://github.com/espnet/espnet/blob/master/ci/test_integration_espnet2.sh

Your local test is important, but this does not cover everyone's usage.

D-Keqi: > Can you make a test to pass this mode? You can find an example in https://github.com/espnet/espnet/tree/master/test/espnet2 or https://github.com/espnet/espnet/blob/master/ci/test_integration_espnet2.sh
> 
> Your local test is important, but this does not cover everyone's usage.

Just let you know that this mode has passed the integration test, and according to other people's responses https://github.com/espnet/espnet/issues/3803#issuecomment-1000020642, I think this version is ok.

sw005320:
OK, thanks!

PR #3896, opened 2021-12-22
Hubert/Wav2vec2 deep feature loss and STOI loss for SE

neillu23:


espnet2/enh/loss/criterions/time_domain.py
```diff
@@ -38,11 +178,7 @@ def __init__(self, filter_length=512):
     def name(self) -> str:
         return "ci_sdr_loss"
 
-    def forward(
-        self,
-        ref: torch.Tensor,
-        inf: torch.Tensor,
-    ) -> torch.Tensor:
+    def forward(self, ref: torch.Tensor, inf: torch.Tensor,) -> torch.Tensor:
```
Emrys365, line 181: The last comma `,` is not necessary for a single-line function signature.

espnet2/enh/loss/criterions/time_domain.py
```diff
@@ -60,11 +196,7 @@ def __init__(self, eps=EPS):
     def name(self) -> str:
         return "snr_loss"
 
-    def forward(
-        self,
-        ref: torch.Tensor,
-        inf: torch.Tensor,
-    ) -> torch.Tensor:
+    def forward(self, ref: torch.Tensor, inf: torch.Tensor,) -> torch.Tensor:
```
Emrys365, line 199: ditto

espnet2/enh/loss/criterions/time_domain.py
```diff
@@ -85,11 +217,7 @@ def __init__(self, eps=EPS):
     def name(self) -> str:
         return "si_snr_loss"
 
-    def forward(
-        self,
-        ref: torch.Tensor,
-        inf: torch.Tensor,
-    ) -> torch.Tensor:
+    def forward(self, ref: torch.Tensor, inf: torch.Tensor,) -> torch.Tensor:
```
Emrys365, line 220: ditto

espnet2/enh/loss/criterions/time_domain.py
```diff
@@ -13,6 +14,145 @@ class TimeDomainLoss(AbsEnhLoss, ABC):
...
+
+    def __init__(self, sr=16000):
+        super().__init__()
+        self.loss_stoi = NegSTOILoss(sr)
+
+    @property
+    def name(self) -> str:
+        return "stoi_loss"
+
+    def forward(self, ref: torch.Tensor, inf: torch.Tensor,) -> torch.Tensor:
```
Emrys365, line 48: The last comma , is not necessary for a single-line function signature.

Emrys365 (commented):
BTW, I think it would be too heavy to include the parameters pretrained Wav2Vec2 or HuBERT model in the checkpoint or the final model, if they will not be updated.
This will not only occupy too much disk space, but also make it difficult to recognize the amount of parameters of the original SE model.

Do you have any ideas on how to work around this issue?

> My current hack is to wrap the pretrained model in `HubertDFLoss` or `Wav2vec2DFLoss` in a list, e.g.
> ```python
> self.w2v2 = [Wav2Vec2ForCTC.from_pretrained("facebook/wav2vec2-base-960h")]
> ```
> Since nn.Module will not be recognized as a part of parameters if it is contained in a list, it will not be stored in the checkpoint.
> But you will have to cast the module to the proper device manually in `forward`, and access it by taking the first element of the list.

neillu23:
@Emrys365 Thank you so much! I have changed those codes according to your comments.
I have put pretrained models in the list, but I also found that the model information only contained the SE model in the previous implementation.
螢幕快照 2021-12-28 下午4 49 45

Emrys365 (approved):
LGTM now.

ftshijt:

PR #3902, opened 2021-12-23
Add SpecAug2 which supports variable maximum width in time masking

pyf98:
In the original `SpecAug`, the maximum width for time masking is fixed (e.g. 40) regardless of the input sequence length. In some papers, the maximum width is dependent on the sequence length: `max_width = max_width_ratio * seq_len`. For example, Google's conformer uses 10 time masks with max_width_ratio=0.05. In this `SpecAug2` module, a new argument `time_mask_width_ratio_range` can be used to realize it.

Note that currently I'm not sure if it can improve performance, so I just add an option here for future exploration.


kamo-naoyuki:
Thank you for contributing. I think this option would be useful.

I'd like to merge `time_mask_width_ratio_range` and `time_mask_width_range` into a class.

```python
class SpecAug:
   def __init__(
        self,
        time_mask_width_range=None,
        time_mask_width_ratio_range=None,
        ...
    ):
        if time_mask_width_range is not None and time_mask_width_ratio_range is not None:
              raise RuntimeError('Either one of "time_mask_width_range" or  "time_mask_width_ratio_range" can be used')
```

sw005320:
@pyf98, can you follow @kamo-naoyuki's suggestions?

pyf98:
Sure, I'll modify it

pyf98:
Hi, I merged the new class into `SpecAug`. Now this module accepts two arguments for time mask width: `time_mask_width_range` and `time_mask_width_ratio_range`. Their default values are both `None`. If `apply_time_mask` is `True`, only one of them should be `None` and the other should have a valid value. The test is updated accordingly.

Could you please double check it? Thank you.

kamo-naoyuki:
Thanks! LGTM. I'll merge

PR #3904, opened 2021-12-23
add eps in  beam-forming reference channel selection

LiChenda:
Add an EPS in amplitude calculatation for complex tensor to improve numerical stability.
Otherwise, the gradient of zero square root may be NaN.


Emrys365 (approved):
LGTM!

sw005320:
Thanks!

PR #3906, opened 2021-12-23
bugfix lang2lid in tts.sh

imdanboy:
lang2lid is not generated correctly.
utt2lid should be replaced with lang2lid.


kan-bayashi (approved):
This is my mistake. Thank you for your kind fixing.

PR #3913, opened 2021-12-24
Update an4 config

pyf98:
The transformer configuration of AN4 is updated in ESPnet2.
- Reduced `warmup_steps`
- Added speed perturbation


sw005320:
Did it improve the performance?
If so, can you update `README.md`?

pyf98:
Hi, I added a file `README.md` in this recipe. Originally there was no result file in `egs2/an4`. Compared with ESPnet1 (`egs/an4`), the result is better.

sw005320:
Thanks a lot, @pyf98!

PR #3914, opened 2021-12-25
Update Grabo (SLU recipe)

pyf98:
This PR updates the SLU recipe: Grabo.

- Fix issues when the data directory is not default (`download`)
- Update `run.sh` to call `local/score.sh` inside `asr.sh`
- Update `result.md` to include model links and emphasize the randomness during data split
  - I intended to make the train/dev/test split deterministic, but the randomness of `os.listdir()` was ignored. Thus, the split can be different in different machines.
  - Since this dataset is very small, it might be better to run multiple experiments with different data splits and calculate average scores.


sw005320:
Thanks a lot!
As I mentioned in the other PR, it would be great if you upload a model to the HF hub.

pyf98:
Thank you. I will export the models to huggingface next.

sw005320:
I just elaborated the explanation about the HF hub https://github.com/espnet/espnet_model_zoo#huggingface
Please try it and let me know if you have some suggestions for the improvements of the documentations.

PR #3916, opened 2021-12-25
Refactor Phonemizer-based G2P

kan-bayashi:
Current phonemizer-based G2P is unstable and too slow due to the bad usage of the phonemizer library.
This PR fixed this issue. 
- Old: Construct the instance at every g2p processing
- New: Construct the instance in `__init__` and call at every g2p processing

Now phonemizer-based on-the-fly G2P works reasonably.


kan-bayashi:
Phonemizer version is fixed until fixing the issue https://github.com/bootphon/phonemizer/issues/104

PR #3929, opened 2021-12-28
updated the results of Slue voxceleb

siddhu001:


siddhu001:
The test set is blind so we will have to submit prediction result to a website (which the authors are preparing) to get test results. The paper does not report dev results so the results are not comparable with one reported in paper.

sw005320:
Cool!
Can you merge the latest master?
We fix the CI issue.

siddhu001:
@sw005320 I think this is ready to be merged!

sw005320:
OK, thanks, @siddhu001!

PR #3930, opened 2021-12-28
fixing rtf estimation using power method

popcornell:
I think the current rtf estimation with power method has a bug: see https://arxiv.org/pdf/2011.15003.pdf section 3.3

### DIFF
--- a/espnet2/enh/layers/beamformer.py
+++ b/espnet2/enh/layers/beamformer.py
@@ -86,8 +86,9 @@ def get_rtf(
         rtf (torch.complex64/ComplexTensor): (..., F, C, 1)
 
     References:
-        [1] Boeddeker, Christoph, et al. "Convolutive transfer function invariant sdr training criteria for multi-channel reverberant speech separation.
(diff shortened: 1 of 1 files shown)


Emrys365:
Actually, I intended to implement it in this way (for numerical stability), and it's mathematically equivalent to the description in Section 3.3 of [[Convolutive transfer function invariant SDR training criteria for multi-channel reverberant speech separation]](https://arxiv.org/pdf/2011.15003.pdf).
> This implementation is initially from Christoph. And yes, I was also confused at the beginning. But after some discussion with him, I got the point. I hope the figure below explains the correspondence between the formulas and the current code.

rtf

For the two formulas in the orange section, you can see that the matrices 𝐑_n and 𝐑_n⁻¹ cancel, and that's why in the code I only write `matmul(psd_speech, rtf)`.

Also note that the last formula is not reflected in the `get_rtf` function, but done in the functions where the resulting RTF is used, e.g. [get_mvdr_vector_with_rtf](https://github.com/espnet/espnet/blob/a6e310354d425e0d3bcbc5f7e681101ea788447e/espnet2/enh/layers/beamformer.py#L216-L218).

popcornell:
Makes sense,  I totally missed that. Closing.

PR #3932, opened 2021-12-28
DPT-FSNet: Dual-path Transformer Based Fusion Network for SE

neillu23:
Add DPT-FSNet according to https://arxiv.org/abs/2104.13002. 
This implementation has not been tested in the espnet repo.

### DIFF
--- a/egs2/vctk_noisy/enh1/conf/tuning/train_enh_dptfsnet.yaml
+++ b/egs2/vctk_noisy/enh1/conf/tuning/train_enh_dptfsnet.yaml
@@ -1,0 +1,59 @@
+optim: adam
+init: xavier_uniform
+max_epoch: 100
+batch_type: folded
+batch_size: 4 # batch_size 16 can be trained on 4 RTX 2080ti
+iterator_type: chunk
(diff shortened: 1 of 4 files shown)


egs2/vctk_noisy/enh1/conf/tuning/train_enh_dptfsnet.yaml
```diff
@@ -0,0 +1,59 @@
...
+    stride: 1
+decoder: conv
+decoder_conf:
+    channel: 64
+    kernel_size: 2
+    stride: 1
+separator: dptfsnet
+separator_conf:
+    num_spk: 1
+    layer: 6
```
Emrys365, line 43: I tested locally on a single GTX 1080ti GPU. The input is of shape (2, 32000).
And I found that it could only fit into the GPU memory when layer=2.
It consumes around 10367 MiB GPU memory for the separator forward process (with gradient).
I am not sure whether the current config file can be used directly on a RTX 2080ti GPU (as stated).

ftshijt:

PR #3943, opened 2022-01-05
espnet2 asr inference bugfix for transducer

jeon30c:
In transducer-based system, we generally do not utilize the EOS symbol because the system is monotonic and has hard alignment for frame data.
But asr_inference eliminates the first and the last symbol which corresponds to SOS and EOS, respectively. That is correct only for AED based ASR. So this PR is for transducer-based ASR.

I trained conformer-transducer using ESPNET2 on librispeech recipe for 100 epochs and the last epoch is utilized for tests.
Before bugfix, WERs for test-clean/test-other were 7.7 and 12.3. After fixing it, WERs drop to 2.9 and 7.2, respectively.


b-flo (approved):
Thanks for the fix!
TransHypothesis' `yseq` is always a list so the following is also suitable (I added a comment for clarity):

```
# remove sos/eos and get results
if isinstance(hyp.yseq, list):
    # no eos token in Transducer hyps; sos token is blank.
    _eos = None if self.asr_model.use_transducer_decoder else -1
    token_int = hyp.yseq[1:_eos]
else:
    token_int = hyp.yseq[1:-1].tolist()
```

Either fixes are fine with me. @sw005320 Feel free to merge or ask for changes.

csukuangfj:
How do different types of errors change before and after the fix, i.e, insertions, deletions, and substitutions?

jeon30c:
@csukuangfj 
> How do different types of errors change before and after the fix, i.e, insertions, deletions, and substitutions?

Before this fix, the last token is missing, which results in either an insertion error (when last token is a complete word) or a substitution error (when last token is a subword). The errors can be corrected.

ftshijt:
Many thanks for your contribution. Can you merge the latest commit in master branch to activate the fixed CI tests?

csukuangfj:
> Before this fix, the last token is missing, which results in either an insertion error 

I am surprised that that error pattern has not been discovered for quite a long time before this PR.

sw005320:
> > Before this fix, the last token is missing, which results in either an insertion error
> 
> I am surprised that that error pattern has not been discovered for quite a long time before this PR.

espnet2 RNN-T PR is very recently made and still in the development mode. It's not a long time.
espnet1 RNN-T is robustly working.

b-flo:
As Shinji said, ESPnet1 is still the go-to for Transducer models. I quickly reworked ESPnet2 implementation because it was needed for some members but I didn't extensively test it. Sorry about that.

Also, there are some performance degradation compared to ESPnet1 from what I observed and a lot of features are missing in this version.
I didn't have time in previous months but we will accelerate work related to Transducer models and a new/updated version is in the pipeline.

csukuangfj:
I see. Thanks!

sw005320:
Hi @jeon30c, could you merge the current master?
We fix the CI issue.

jeon30c:
@sw005320 For me, it says "merging is blocked. The base branch restricts merging to authorized users. You're not authorized to merge this pull request".

csukuangfj:
I think what @sw005320 means is that you merge the master locally into your own branch `asr_transducer_inference` and then push it to GitHub.

You can do
```bash
cd /path/to/espnet
git checkout asr_transducer_inference
git remote add espnet https://github.com/espnet/espnet
git fetch espnet
git merge espnet/master
git push
```

jeon30c:
@csukuangfj I just did what you told me. It merged the current espnet/master to my forked git repository. Is that right?

csukuangfj:
Yes, I believe so.

sw005320:
Yes, that’s what I meant to say.
Thanks @jeon30c and @csukuangfj

sw005320:
Many thanks, @jeon30c!!!

sw005320:
@jeon30c, could you share your config files (training/decoding)?
We want to reproduce your result.

jeon30c:
@sw005320 Yes of course.
```
# Trained with Tesla V100-SXM2(32GB) x 8 GPUs. It takes about 1.5 days.
batch_type: numel
batch_bins: 20000000
accum_grad: 2
max_epoch: 100
patience: none
init: none
best_model_criterion:
-   - valid
    - acc
    - max
keep_nbest_models: 10

model_conf:
    ctc_weight: 0.0
    report_cer: False
    report_wer: False

encoder: conformer
encoder_conf:
    output_size: 512
    attention_heads: 8
    linear_units: 2048
    num_blocks: 12
    dropout_rate: 0.1
    positional_dropout_rate: 0.1
    attention_dropout_rate: 0.1
    input_layer: conv2d
    normalize_before: true
    macaron_style: true
    pos_enc_layer_type: "rel_pos"
    selfattention_layer_type: "rel_selfattn"
    activation_type: "swish"
    use_cnn_module:  true
    cnn_module_kernel: 31

decoder: transducer
decoder_conf:
    rnn_type: lstm
    num_layers: 1
    hidden_size: 512
    dropout: 0.1
    dropout_embed: 0.1

joint_net_conf:
    joint_space_size: 640
# model_conf:
#     ctc_weight: 0.3
#     lsm_weight: 0.1
#     length_normalized_loss: false

optim: adam
optim_conf:
    lr: 0.0015
scheduler: warmuplr
scheduler_conf:
    warmup_steps: 25000

frontend_conf:
  n_fft: 512
  hop_length: 160 

specaug: specaug
specaug_conf:
    apply_time_warp: true
    time_warp_window: 5
    time_warp_mode: bicubic
    apply_freq_mask: true
    freq_mask_width_range:
    - 0
    - 30
    num_freq_mask: 2
    apply_time_mask: true
    time_mask_width_range:
    - 0
    - 40
    num_time_mask: 2
```
```
beam_size: 10
transducer_conf:
    search_type: default
    score_norm: True
```
The setting are for basical bench-marking to the previous experiments.
Currently, I think that the results of espnet2-transducer ASR are fairly good. So I am moving my projects to espnet2 now. Thanks for the implementations.

sw005320:
Super thanks!!!

sw005320:
@jeon30c, could I ask you to make a PR with this config and upload a model?
Several people are interested in your results!

b-flo:
@sw005320 If it's okay with you, it would be better to wait for a bit as I'll open a PR in a few days with the new version!
We may work on ESPnet2 recipes/models from there.

sw005320:
OK, but it must be independent, right?
Do you mean that you'd touch the librispeech recipe?

b-flo:
> OK, but it must be independent, right?

It will be independent but if everything is OK, the current version will be removed.

> Do you mean that you'd touch the librispeech recipe?

Well, it would be great if someone can re-train a model with @jeon30c config and the new version. Performance should be on par or better than currently (If there was indeed a degradation in performance).

sw005320:
OK, and I hope it would not degrade the performance as it is already working with the nice performance as @jeon30c reported.

jeon30c:
@sw005320 @b-flo I am also looking forward to applying a new version. By the way, you said that you observed some performance degradation in case of espnet2-transducer compared to espnet1-transducer even after the bug was fixed. So I am curious about what the WER numbers were especially when no LM is applied. Could you tell me about the comparison?

b-flo:
> So I am curious about what the WER numbers were especially when no LM is applied. Could you tell me about the comparison?

I was talking about my initial experiments with other datasets (mainly VIVOS, Voxforge IT and a custom dataset) from where I couldn't reproduce ESPnet1 results. A difference in performance is expected as some parts behave differently between ESpnet1 and ESPnet2 (initialization, definition, etc), but I couldn't reach the best results even by tuning. With comparable training and decoding, I end up with almost a 4% CER difference in some cases.
Because I didn't extensively test the ESPnet2 implementation, I suspect I missed something.

Librispeech model is another story. We could propose an ESPnet1 model with proper performance thanks to @sw005320 efforts but there is room for improvement. If you have some resources for what, it would be great to work on an ESPnet2 recipe when the new implementation with all features is ready.

Stasiche:
Hello! Do you have a checkpoint with trained RNN-T conformer?

JingbiaoMei:
I was wonderining if you are using asr.py or asr_transformer.py? As asr_transformer does not have the "encoder: conformer" argument

PR #3953, opened 2022-01-10
Add test codes for diarization

YushiUeda:
This is a follow-up PR of https://github.com/espnet/espnet/pull/3900.


ftshijt:
LGTM. Many thanks! it's very helpful

PR #3957, opened 2022-01-11
Diarization: fix for convolutional input layer in the encoder

alumae:
If the dirization encoder uses subsampling (e.g. conv2d8 as input_layer), the sequence lengths of pred and spk_labels will be slightly different (even when using the correct value as `hop_length` in the label aggregator). This happens because the stride and padding is done slightly differently in label aggregator and convolution modules. This fix allows the predictions to be slightly shorter than speaker labels. Works with conv2d and conv2d8 according to my tests.

I am not sure if this is the best fix to this problem, I am just providing this as a starting point and I am open for comments.
BTW, EEND solves this by just reducing the labels and predictions to the length that is equal to the shortest, using zip (see https://github.com/hitachi-speech/EEND/blob/ddf4df4da55a493f550b6c45689b2d60a716c06d/eend/chainer_backend/models.py#L65).


espnet2/diar/espnet_model.py
```diff
@@ -125,6 +125,14 @@ def forward(
             spk_labels, spk_labels_lengths
         )
 
+        # If encoder uses conv* as input_layer (i.e., subsampling), 
+        # the sequence length of 'pred' might me slighly less than the
```
ftshijt, line 129: `me` seems to be a typo?

alumae:
Merged from master, fixed typo, applied black.

ftshijt:
> Merged from master, fixed typo, applied black.

Looks very cool! will merge it after passing the CI, many thanks

PR #3960, opened 2022-01-11
Fix the numpy version for CI test

ftshijt:
The CI is failed due to the recent update of numpy version. The incompatible issues include both matplotlib and cython. For matplotlib, the problem could be that our version is a bit old. However, for cython, the issue is just brought up with no further fix yet. Therefore, we propose a temporary fix by using numpy 1.20.3 instead of 1.22.0


ftshijt:
will combine the commit later

PR #3968, opened 2022-01-15
Add ESPnet2 ASR recipe for the SEAME Mandarin-English code-switching corpus

vectominist:
I add an espnet2 recipe for the [SEAME](https://catalog.ldc.upenn.edu/LDC2015S04) Mandarin-English code-switching corpus.
Trained ASR and LM are uploaded to Zenodo and Huggingface.

The conformer ASR performance is slightly different from our AR baseline [ASRU 2021 paper](https://arxiv.org/abs/2104.02258) because the new recipe processes data more similar to the [official document](https://github.com/zengzp0912/SEAME-dev-set).
Still, only a conformer seq2seq model + transformer LM achieved similar performance with the previous SOTA:
| method | devman | devsge |
| --- | --- | --- |
| [Zhou et al. 2020](https://arxiv.org/abs/2006.10414) | 16.7 | 23.1 |
| ESPnet2 Conformer | 16.5 | 23.2 |


egs2/seame/asr1/run.sh
```diff
@@ -0,0 +1,51 @@
...
+
+train_set="train"
+valid_set="valid"
+test_sets="devman devsge"
+
+asr_config=conf/tuning/train_asr_conformer.yaml
+lm_config=conf/tuning/train_lm_transformer.yaml
+inference_config=conf/decode_asr.yaml
+
+if [ ${stage} -eq 1 ] || [ ! -f "data/train/token.man.2" ]; then
```
ftshijt, line 19: Maybe just add a check for data/train/token.man.2 here, then through an error is stage is not 1? We would prefer to have the data preprocessing in the main asr.sh

vectominist: Just fixed, please take a look. Thanks!

ftshijt:
The recipe generally looks cool! I will merge it after you fix the CI issues

egs2/seame/asr1/conf/decode_asr.yaml
```diff
@@ -0,0 +1,3 @@
+lm_weight: 0.2
+ctc_weight: 0.4
+beam_size: 30
```
sw005320, line 3: Is it after tuning?
Beam size is a bit larger than usual (20) (it’s OK if this is after some tuning)

egs2/seame/asr1/local/preprocess.py
```diff
@@ -0,0 +1,628 @@
+#!/usr/bin/env python3
+# -*- encoding: utf8 -*-
+
+import re
```
sw005320, line 4: I just want to make sure whether this preprocess is common and the same with the others (e.g., the other reports), since the different preprocessing would cause the different results.
If so (or if not), please leave some comments in the header of this file.

sw005320 (commented):
Looks very good to me.
Thanks @vectominist!
I have some questions.

sw005320:
@brianyan918, can you also quickly review this PR?

egs2/seame/asr1/conf/decode_asr.yaml
vectominist: Yes, the weights and beam size are tuned. I can make the beam size 10 (only 0.1% absolute WER increase in both eval sets).

egs2/seame/asr1/local/preprocess.py
vectominist: Just added notes to this file. Please take a look. Thanks!

egs2/seame/asr1/conf/decode_asr.yaml
sw005320: ok!
either is fine!

egs2/seame/asr1/local/preprocess.py
sw005320: LGTM!

brianyan918:
> @brianyan918, can you also quickly review this PR?

@vectominist this looks great. I have one suggestion regarding scoring. Should we add a script to compute the English WER and Mandarin CER separately as well? Correct me if I'm wrong, but I didn't see this option in this PR.

I think we can merge this PR and add that script in another subsequent PR.

sw005320:
@vectominist, thanks a lot!

> Should we add a script to compute the English WER and Mandarin CER separately as well? Correct me if I'm wrong, but I didn't see this option in this PR.

This is a great suggestion. @vectominist, can you consider it?

vectominist:
Sure! I will create a new PR with separately scoring Mandarin CER and English WER.

ftshijt:
I have a question for separate scoring, how do you perform separated wer when code-switching? Do we need to consider the other language's place-holding, when doing soring? @brianyan918

vectominist:
As we did in our [ASRU paper](https://arxiv.org/abs/2104.02258), we first separate English and Mandarin transcriptions, then respectively compute WER and CER.
I think this is a standard way of computing the error rates in code-switching ASR tasks.
BTW, I just created a PR.

PR #3972, opened 2022-01-18
Bugfix for ST scoring

ftshijt:
This PR serves for the scoring bugfix.

Current results could reach similar performances to egs1 on transformer with pre-trained ASR.

I will finish this PR after adding support for multiple references setting


sw005320 (approved):
LTTM!

PR #3976, opened 2022-01-19
Scoring Mandarin / English separately for the SEAME corpus

vectominist:
I add a `local/score.sh` to separately scoring Mandarin CER and English WER in the `egs2/seame` recipe.
New results are also added in `README.md`.


sw005320:
LGTM!
Thanks, @vectominist!

PR #3981, opened 2022-01-21
Faster diar1/pyscripts/utils/convert_rttm.py 

alumae:
Another small change to the diarization training recipe.

The current implementation of egs2/TEMPLATE/diar1/pyscripts/utils/convert_rttm.py reads all diarization training files serially temporarily into memory, just to get their length. This takes time (could be like an hour) when the training dataset is large. This change fixes it so that the audio file length will be derived from the header, and the whole audio file reading is not needed. It should be around 3 times faster but of course depends on many factors.


sw005320:
Very cool, @alumae!
Many thanks!

PR #3984, opened 2022-01-22
[Code cleaning] Speech translation (merging into master)

ftshijt:
This is the initial PR for speech translation. Currently, we could achieve comparable results to espnet1 on transformer-asr-pretraining and multitask learning. The BLEU for fisher_test is 48.6; The BLEU for callhome test is 19.6 (a bit better to similar setting)


ftshijt:
This PR is mostly for code-cleaning purpose

PR #3987, opened 2022-01-22
Speech translation framework (merging into master)

ftshijt:
This is the initial PR for speech translation. Currently, we could achieve comparable results to espnet1 on transformer-asr-pretraining and multitask learning. The BLEU for fisher_test is 48.6; The BLEU for callhome test is 19.6 (a bit better to similar setting)


ftshijt:
The PR is ready to merge

brianyan918:
I did not check it fully, but I think since this is mostly recipe we can safely merge it

PR #4002, opened 2022-01-27
Bugfix for espnet2 ngram

yaochie:
Train the ngram model using the whole utterance instead of just the first token.


sw005320:
Thanks a lot, @yaochie!

PR #4017, opened 2022-01-31
Puebla-Nahuatl speech translation recipe for egs2

ftshijt:


ftshijt:
Initially, I thought the import CI issue comes from a newer fairseq version. However, it seems not :-(

Issue #3948, opened 2022-01-08
AttributeError: module 'warprnnt_pytorch.warp_rnnt' has no attribute 'gpu_rnnt'


ily6:
cuda: 11.3,  pytorch: 1.10.1, python: 3.8.5
I trained a conformer transducer model using aishell recipe, but got this as follows, could anyone give me a hand?
2022-01-08 19:48:09,317 (asr_train:563) WARNING: Skip DEBUG/INFO messages
2022-01-08 19:48:17,344 (asr:493) WARNING: num. model params: 32,575,561 (num. trained: 32,575,561 (100.0%))
2022-01-08 19:48:23,164 (io_utils:57) WARNING: [Experimental feature] Some preprocessing will be done for the mini-batch creation using Transformation(
    0: TimeWarp(max_time_warp=5, inplace=True, mode=PIL)
    1: FreqMask(F=30, n_mask=2, replace_with_zero=False, inplace=True)
    2: TimeMask(T=40, n_mask=2, replace_with_zero=False, inplace=True))
2022-01-08 19:48:23,167 (io_utils:57) WARNING: [Experimental feature] Some preprocessing will be done for the mini-batch creation using Transformation(
    0: TimeWarp(max_time_warp=5, inplace=True, mode=PIL)
    1: FreqMask(F=30, n_mask=2, replace_with_zero=False, inplace=True)
    2: TimeMask(T=40, n_mask=2, replace_with_zero=False, inplace=True))
^[[JException in main training loop: module 'warprnnt_pytorch.warp_rnnt' has no attribute 'gpu_rnnt'
Traceback (most recent call last):
  File "/home/lli/anaconda3/lib/python3.8/site-packages/chainer/training/trainer.py", line 316, in run
    update()
  File "/mnt/data3/lli/espnet/espnet/asr/pytorch_backend/asr.py", line 231, in update
    self.update_core()
  File "/mnt/data3/lli/espnet/espnet/asr/pytorch_backend/asr.py", line 195, in update_core
    data_parallel(self.model, x, range(self.ngpu)).mean() / self.accum_grad
  File "/home/lli/anaconda3/lib/python3.8/site-packages/torch/nn/parallel/data_parallel.py", line 228, in data_parallel
    return module(*inputs[0], **module_kwargs[0])
  File "/home/lli/anaconda3/lib/python3.8/site-packages/torch/nn/modules/module.py", line 1102, in _call_impl
    return forward_call(*input, **kwargs)
[...]

PR #3697, opened 2021-10-25
Refactor custom Transducer build

b-flo:
**Edit:** I put too much work into making a proper custom model building and the current version would introduce a lot of changes to ESPnet1. Holding onto that, I will stick to modifying the following for now : 

- Add `Conv1d` class for custom encoder
- Add BN-ReLU supports to `CausalConv1d`
- Add arguments for encoder input layer dropout rate and pos. enc. dropout rate
- Small refactor of custom model-building to be more modular and more comprehensible.
- _Unrelated_: fixed some default arguments (mainly aux. tasks)

-------------------------------

This is a follow-up of #3217 I forgot to push : 

- Add `ConvEncoderLayer` class for custom encoder which supports 1D and 2D conv (https://arxiv.org/pdf/2008.05750.pdf).
- Remove TDNN class. I'll add a specific class when we have a proper implementation (TDNN-F is pending).
- Add missing layers and options for `CausalConv1d`
- Added `none` input layer for pos. encoding only.
- Extend conv-related test in `test/test_custom_transducer.py`
- Unrelated: fixed some default arguments (mainly aux. tasks)

~~I was under the impression I didn't finish that task but it seems so? I'll double-check, I may have forgotten something.~~
I forgot to fix and refactor some things related to custom model building.

PR #4020, opened 2022-02-02
Fix bug for Conformer-T

YosukeHiguchi:
This PR fixes the bug in Conformer-T reported in #3948. The problem was that each module in a conformer block was sharing the same parameters (pointing to the same memory address) across other blocks.


b-flo (approved):
Thanks a lot @YosukeHiguchi! I'm OK with the changes, the issue was discussed off the list.

For reference : 

- Some unexpected behavior was reported in #3948 (and by email) resulting in 99% ER during decoding with the Conformer-T model.
- @yusshino re-run Aishell-1 Conformer-T experiments and observed that: 1) More than 50% of the model parameters were missing and 2) The issue could be "avoided" by tuning/skipping model averaging (obviously ER still degraded compared to previous versions)
- We found out that the issue began appearing in v.0.10.4. The only possible culprit is the block building refactor in #3697.
- @YosukeHiguchi pinpointed the issue: some parameters are initialized outside of the lambda function in the Conformer block, instead of inside in previous versions. During block duplication, the references will be copied instead of creating new objects.

After verification, other blocks should not be impacted by my (faulty) refactor from v.0.10.4.

yusshino:
Great!

ftshijt:

YosukeHiguchi:
I've tested training a model on librispeech-100 and confirmed the bug is now fixed. This PR is ready to merge.

b-flo:
I'm merging, thanks again @YosukeHiguchi!

PR #4029, opened 2022-02-03
added learnable fourier features

popcornell:
reference original PR that I have closed by error #3926


sw005320:
Couple of errors (due to https://github.com/espnet/espnet/runs/5061410072?check_suite_focus=true#step:8:499 ??). Please fix them.

popcornell:
Some are due to old version of Pytorch e.g. GELU not available in 1.3.1. Should I fall back to ReLU for that one ? Or add custom GELU activation ?

sw005320:
We will discuss the drop of the support of 1.3.1 and let you know it as soon as possible.

sw005320:
We decided to drop 1.3.1. Can you merge the master again?

sw005320:
Hmm, we may need to skip the half-precision test for this function?
https://github.com/espnet/espnet/runs/5130447612?check_suite_focus=true#step:8:296

popcornell:
Sure. BTW I have torch 1.8.0 on local and all tests are fine. Half-precision is important but I guess only really significant for latest torch releases. Removed.

sw005320:
I just merge this PR!
Many thanks @popcornell!

sw005320:
We may also include the usage of this new function, later :)

PR #4032, opened 2022-02-04
Espnet2 transducer v2

b-flo:
Hi,

This PR is a draft for the new version of Transducer models in ESPnet2, separated from the main ASR task (CTC+Att). It's working but please note that :

- This is not the final version, it's only to open discussion. If needed, I have some alternatives / other versions.
- Some parts or features are removed compared to the previous version. It can be easily added but I would like to add them one by one with careful testing or feedback.
- This draft may contain minor issues and typos, useless code, etc. Feel free to point out any weird/wrong parts.

Performance should be on par or better than previously. I also found out what caused the performance degradation for the Voxforge model (mainly due to initialization, and some small training differences). It may be worth extending the investigation though!
@jeon30c Would it be possible for you to re-train a Librispeech model with this version to compare performance, please?
@sw005320 Do you know if we have other models to compare? I'm not sure who already used the first version. 

Also, after we are set on the task and model definition, I would like to at least make the encoder and decoder fully customizable (similar to the custom model in ESPnet1). Mainly, the changes would be :
	- Add unified Encoder containing PreEncoder (bottlenecks/input blocks) + BodyEncoder (supporting main nets/blocks + some bridge blocks)
	- Same for Decoder.
	- Refactor the BeamSearch / Scorer part to reflect changes and optimize for ESPnet2.

After that :

- [ ] Add tests
- [ ] Add documentation


b-flo:
Thank you very much! Btw, the following should be changed in comparison to your previous training : 

- Remove `init: none` in training config
- Add `--asr_transducer true` to `asr.sh` parameters in `run.sh`

Outside of that (and bugs), everything should work as intended!

jeon30c:
@b-flo
screenshot_loss
I observed that the loss was not well decreasing compared to the previous version.
As the basic functionality is same, I think it should be almost same to the previous training as
screenshot_loss_prev

b-flo:
> I observed that the loss was not well decreasing compared to the previous version.
> As the basic functionality is same, I think it should be almost same to the previous training as ...

The implementation is equivalent except for the model initialization which rely on the ESPnet1 one here. If you comment l.435 in [espnet2/tasks/asr_transducer.py](https://github.com/espnet/espnet/pull/4032/files#diff-de1e4903db4f5b993b92542ec847ff975dc68328d552260494a7b4aacfc763d0), it should be equivalent to your previous run. Could you also test, please?

Interestingly enough, in some other experiments (Voxforge and a custom dataset), I couldn't replicate previous performances without applying this initialization.

jeon30c:
@b-flo
스크린샷 2022-02-14 오후 4 51 44
This is very interesting. As you said, I commented out the init part and retrained a model, that showed very similar training procedure as the above graph. The init function seems to have great effect on performance. 

After finishing the training completely, I will post again the final WER results.

b-flo:
Initialization is one of the parts I want to address alongside training techniques. I'll work on that after adding custom architectures, for now, I'll add back the `init` option with support for "espnet1 initialization".

jeon30c:
Without init, the performance is on par with espnet2 transducer v1. For test-clean and test-other, WERs are 3.1, 7.2, respectively.

b-flo:
Thank you very much for the report!
I'll add the custom encoder and decoder to have more flexibility in terms of architecture and we can tune from there.

b-flo:
Hi,

Sorry for the delay, the last commit is a new proposition for Transducer models in ESPnet2.
Some mistakes or rough edges may still exist as I'm dealing with several versions at the same time, but I tested most features/setups in this version. It worked as intended.

For the changes compared to previous version (outside minor refactoring, renaming, etc):

## Encoder

Unique encoder type. Similar to ESPnet1 Transducer custom model except we also support RNN here.

### x Changes

The encoder architecture is defined through three arguments:

  - `general_params` (optional, Dict): General params for the encoder. Contains only conformer-related parameters for now.
  - `input_block` (optional, Dict): Input block parameters. Supported types: VGG, Conv2d (with sub. factor = {2, 4, 6})
  - `body_block` (List[Dict]): Body block parameters. Supported types: Conv1d, Conformer, RNN

### x Notes

 - Each block has its own parameters (same as ESPnet1).
 - The output layer of each block is removed if a RNN or Conv1d block follows. It's similar to ESPnet1 but it kind of complicate setting correct parameters for the input block and first body block. I'll revisit it.
- I think an unique customizable input block may be better?

## Decoder

Similar to the previous version.

### x Changes

  - **General**: added `dim_embedding` parameter + some refactor for inference.
  - `batch_score(...)`: Removed caching scheme.

### x Notes

  - We could use a N-layers LSTM instead of a stack of 1-layer LSTM but applying dropout to RNN's output seems quite effective.
  - If we want to add `TransformerDecoder`, I also have an implementation for an unique decoder type like for the encoder. It can support Transformer and Conv1d layers.

## Beam search

Similar to the previous version.

### x Changes

  - **NSC**: Removed.
  - **mAES**: Removed prefix search.
  - **mAES**: Fixed `select_top_k_candidates` loop taking a long time with a big vocab size. It also improve decoding time in general.
  - **General**: moved some utility methods within `BeamSearchTransducer` + general minor refactoring due to new changes.


### x Notes

  - I'm considering also dropping **TSD**.
  - The behaviour of `ALSD` with `StatelessDecoder` is kind of... Unexpected? I suspect an issue due to memory sharing but I couldn't find anything yet.

b-flo:
This PR is mostly done, what's missing is 1) the documentation (it will be done by tomorrow) and 2) the remaining unit tests to reach 100% coverage (it can be done later). I'm also validating current performance on Librispeech-100h, the training will finish in 1,5 days.

@sw005320 I'm not sure who I should assign for the review? I suppose the priority for merging is to at least ensure I didn't make any mistakes when separating the Transducer task/model from the main ASR task.

b-flo:
Performance seems okay with the new version? Results for Conformer-T trained with Libri-100:

|dev_clean|2703|54402|94.0|5.3|0.7|0.6|6.6|58.2|
|dev_other|2864|50948|84.7|13.2|2.1|1.7|17.0|79.0|
|test_clean|2620|52576|93.7|5.5|0.8|0.7|7.0|59.3|
|test_other|2939|52343|84.0|13.5|2.4|1.7|17.6|81.5|

It's not tuned because I don't have resources for that right now but I'll use this model as baseline to work on the general optimizations I talked previously (init, training schemes, etc).

doc/espnet2_tutorial.md
```diff
@@ -340,3 +340,197 @@ To enable online decoding, the argument `--use_streaming true` should be added t
...
+### Multi-task learning
+
+We also support multi-task learning with two auxiliary tasks: CTC and cross-entropy w/ label smoothing option (called LM loss here). The auxiliary tasks contributes to the overal task defined as:
+
+**L_tot = (λ_trans x L_trans) + (λ_auxCTC x L_auxCTC) + (λ_auxLM x L_auxLM)**
+
+where the losses (L_*) are respectively, in order: The Transducer loss, the CTC loss and the LM loss. Lambda values define their respective contribution to the total loss. Each task can be parametrized using the following options, passed to `model_conf`:
+
+    model_conf:
+      transducer-loss-weight: Weight of the Transducer loss (float, default = 1.0)
```
funboarder13920, line 503: typo: transducer_weight
suggestion: add the yaml config you are using to the librispeech conf folder and reference it ?

b-flo: > typo: transducer-loss-weight -> transducer_weight

Thanks.

> suggestion: add the yaml config you are using to the librispeech conf folder and reference it ?

It's the plan but it will be updated in another PR. Libri work won't be included here for clarity.

b-flo:
Hi,

> I was able to try the rnn-t architecture. Conformer encoder seems to work well.
> However, I was not able to make rnn encoder converge/overfit neither on librispeech_100 nor on a very small dataset. I tried a few different learning rates.
> Were you able to make an rnn encoder architecture converge ?

Yes, I previously trained a RNN-T on VIVOS and Libri-100 based on our ESPnet1 configs. I don't think I broke something since then? What's your configuration?

> Can you explain why you chose to diverge from/not use the AbsEncoder and AbsDecoder espnet interface, in the transducer encoder/decoder classes ? 

There are a lot of reasons but ultimately it come down to allowing more freedom in terms of development for concurrent works (Transducer vs. Joint CTC-Att) without worrying about impacting other models or tasks in any way.

We could easily unify everything under the same ASR task and model but in my opinion, the separation make it easier for everyone if we want to do add or handle model-specific features and optimizations. Or if we want to make some drastic design choices like the customizable encoder architecture. It was proposed in order to ease reproduction of some Transducer architectures such as [Huang et al., 2020](https://arxiv.org/pdf/2008.05750.pdf) or [Weng et al., 2019](https://arxiv.org/pdf/1911.12487.pdf) but it's not something needed nor requested for other models.

In the end, there are debatable pros and cons either you unify or separate tasks/models in any form. It depends on the objective. I took the same route as in ESPnet1 but I'm totally open on discussing or implementing other designs.

> I am under the impression that non-transducer encoders/decoders could work out of the box with your implementation but that the interface change prevents using them.

I'm not sure what "your implementation" refer to here?

> Each block would have to be implemented twice: for attention based models and for rnn-t models

Note that some encoder classes (such as Conformer) were duplicated because I did some temporary modification/simplification. Like in ESPnet1, we could simply re-use already existing definitions/classes.
Also, adding or sharing new block introduced for other ASR models should be rather straightforward in my opinion.

funboarder13920:
I finally manage to have a rnn encoder model converge on a small dataset.


> I'm not sure what "your implementation" refer to here?

I thought that the forward method of transducer encoders were taking (sequences, masks) as input when regular encoders were taking (sequences, sequence_lengths). Both are using (sequences, sequence_lengths), so using non-transducer encoders should work out of the box.

b-flo:
> I thought that the forward method of transducer encoders were taking (sequences, masks) as input when regular encoders were taking (sequences, sequence_lengths). Both are using (sequences, sequence_lengths), so using non-transducer encoders should work out of the box.

Small note: because we handle all encoder block types within the same architecture, "masks" may refer to different things whether we're dealing with RNN (seq. lengths) or Transformer/Conformer (mask).

Current implementation won't work out of the box for other models because some changes were made in relation to the single customizable architecture. Also, some encoder parts were slightly modified/simplified either because we either dropped support for some features/parameters, laid groundwork for future work or removed parts not useful for Transducer model.

Again, we could unify most parts and the current implementation in master is like that. The separation is motivated by what I talked previously. Also, duplicated parts may seem counterproductive but, outside previous arguments, we can at least control behavior and what's needed in regards to a single task.

sw005320:
@ftshijt, @pyf98, and @pengchengguo, can we start to check this very cool PR?
Actually, the high-level discussion was already done by @b-flo and me (e.g., making it separate from the main ASR task).
With this, we can allow transducer implementations to have a different style based on a block.
I think this is cool, but at the same time, the block-based approaches may have some more unification and distinction with the asr task.
I would like you to review this PR and also would like to collect your opinions about it.

b-flo:
I'm currently doing some experiments on activation functions (Mish, E-Swish, Smish, etc) and some seems to work well for Conformer and JointNetwork. I'm adding them here because it's kind of related to the design discussion but sorry, I won't add further things for the review.

Note on that addition: Currently, we have `get_activation()` in `espnet/nets/pytorch_backend/nets_utils.py` and `Swish()` in `espnet/nets/pytorch_backend/conformer/swish.py`. Instead of adding one or multiple files in conformer directory, I prefer proposing a refactor with a more general design: by adding an `activation.py` with all act. functions and support for more options in `get_activation.py`. I made a standalone version for Transducer because I was unsure on the (back-)compatibility part for other models but it can be unified easily.

funboarder13920 (commented):
Hello,

I started using your code for some experiments. I made a few comments regarding the difference of interface between rnn-t and non rnn-t decoders' methods that makes navigating between different architectures' of espnet a bit confusing.

Best

espnet2/asr/transducer/decoder/abs_decoder.py
```diff
@@ -0,0 +1,117 @@
...
+            dec_out: Decoder output sequences. (B, T, D_dec)
+
+        """
+        raise NotImplementedError
+
+    @abstractmethod
+    def score(
+        self,
+        label: torch.Tensor,
+        label_sequence: int,
```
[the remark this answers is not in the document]
b-flo: Good catch, thanks!

espnet2/asr/transducer/beam_search_transducer.py
```diff
@@ -279,7 +276,7 @@ def default_beam_search(self, enc_out: torch.Tensor) -> List[Hypothesis]:
 
                 if self.use_lm:
                     lm_scores, lm_state = self.lm.score(
-                        lm_tokens, max_hyp.lm_state, None
+                        label[0], max_hyp.lm_state, None
```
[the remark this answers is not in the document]
b-flo: label[0] is a tensor and `max_hyp.yseq[-1]` is an integer. For the naming part, see my comment below.

espnet2/asr/transducer/beam_search_transducer.py
```diff
@@ -260,7 +252,12 @@ def default_beam_search(self, enc_out: torch.Tensor) -> List[Hypothesis]:
                 max_hyp = max(hyps, key=lambda x: x.score)
                 hyps.remove(max_hyp)
 
-                dec_out, state, lm_tokens = self.decoder.score(max_hyp, cache)
+                label = torch.full(
+                    (1, 1), max_hyp.yseq[-1], dtype=torch.long, device=enc_out.device
+                )
+                dec_out, state = self.decoder.score(
```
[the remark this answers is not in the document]
funboarder13920: Thank you for your reply.
I find difficult that I need to learn a new interface and new namings despite being inside espnet and the concepts being similar.

Also, the choice of `score`  naming and interface makes the decoder impossible to inherit the ScorerInterface, unfortunately.

I am not saying that to annoy you, I just thought that some comments about the issues I encountered using your PR and the change I had to make could help the espnet project.

Best,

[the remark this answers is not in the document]
b-flo: > I am not saying that to annoy you, I just thought that some comments about the issues I encountered using your PR and the change I had to make could help the espnet project.

No worries, that's the point of an open review. Feedback is appreciated and debating is better than accepting PRs without questions!

> I find difficult that I need to learn a new interface and new namings despite being inside espnet and the concepts being similar.

If we talk about naming, ESPnet1 Transducer users will also be confused if we revert to original meaning in that case. But again, I'm not against using any naming!
About the similarity in terms of concepts, I'm not sure I understand what you mean? Similar design is a thing but some parts (decoder, decoding process, multi-task loss, etc) are tailor-made for Transducer models so the difference is intended.

> Also, the choice of score naming and interface makes the decoder impossible to inherit the ScorerInterface, unfortunately.

We did try to unify the process, in both ESPnet1 and ESPnet2.
`ScorerInterface` was made with CTC/Att decoding in mind and is not suited for Transducer models, at least without some refactoring to keep resources usage consistent (speed, memory, etc) and also make sense for everyone. We're open to proposition though!

pengchengguo (commented):
Thanks, @b-flo, it is a cool PR!!!
I think using separated task and model definitions for the Transducer model is a good idea, which makes the code more clear and compact.
All the implementations are good to me. A slight suggestion is that some codes of the encoder blocks (in `espnet2/asr/transducer/encoder`) are a little redundant since most of them are also implemented in `espnet/pytorch_backend/{transducer, transformer}`. Is it possible to merge them in future work?

b-flo:
Thank you for looking at this PR @pengchengguo !

>  Is it possible to merge them in future work?

Yes but first, for the sake of the discussion, let me try to summarize why I duplicated each encoder block right now:

- `Conv2dSubsamplingX` and `VGG2L`: Because the "custom" encoder in ESPnet2 also covers RNN compared to ESPnet1, we need to handle two things in these versions: output block definition and different "mask" (`mask` from Transformer/Conformer and `seq_len` from RNN). For that, I made a unified `Conv2dSubsamplingX` to avoid too many duplications. Later, I want to replace all classes by a unique `InputBlock` with a fully customizable definition.
- `Conv1d`: Same as input blocks (except for the last sentence).
- `RNN` and `RNNP`: I reworked/simplified the old versions but that is not the reason for the duplication. I could rely on [the rnn layers](https://github.com/espnet/espnet/blob/master/espnet/nets/pytorch_backend/transducer/rnn_encoder.py) for Transducer in ESPnet1 but I'm reworking two parts (future PRs): 1) how we handle "intermediate encoder output" for auxiliary Transducer loss and auxiliary Symm-KL loss, and 2) how we handle things related streaming, caches, etc.
- `Conformer`: There are no present reasons to duplicate this one outside simplification and future work preparation. However, I need to match other blocks' definitions and behavior for now.

As I said previously, duplication is also my concern but I'm less concerned by small duplications than impacting other models in any way (i.e.: compatibility, behavior, adding overhead in terms of code or resources, etc).
In the end, if we want to do a proper unification, I think we should think of a general refactor taking into account all models but also the use of each part in terms of future works and objectives... But I may be mistaken or lack experience. Feel free to propose any changes!

sw005320:
@b-flo, I discussed several people, including people actually using both the current and these new implementations.
Overall, I have the following suggestions.

1. we want to keep our current implementation since it could have tight integration with various existing encoders. Please revert them. I think it is no problem for us to have two types of implementations. Don't worry if it is difficult for you to maintain both implementations; I'll ask other people to maintain the current one. You can just focus on the new one.
2. About your new implementation. My suggestion (given the discussions with @kan-bayashi) is that we may refer to `espnet2/gan_tts`. Actually, `tts` and `gan_tts` require different tasks, and @kan-bayashi split them. You can also follow the same way, making your transducer more independent and self-consistent ;). For example, you can have `espnet/asr_transducer` instead of putting some of them in `espnet/asr`
3. given this separate task, @kan-bayashi suggests that we can change two asr modes in the recipe by following https://github.com/espnet/espnet/blob/ccdc59c3d8c265c6711740185b508221c9126e4c/egs2/TEMPLATE/tts1/tts.sh#L88

b-flo:
Hi,

 1. Let me check if I can revert because I nuked several things. I'm kind of puzzled by the proposed solution though... I'll 
 2. I took inspiration on TTS for the split, if I understand, you only want me to move the directory `asr/transducer` to `asr_transducer`, right?
 3. OK, what we are doing currently is almost the same (in asr template, we use a `asr_transducer` flag to set the proper binaries)

sw005320:
> 2. I took inspiration on TTS for the split, if I understand, you only want me to move the directory `asr/transducer` to `asr_transducer`, right?

Yes!

b-flo:
> Let me check if I can revert because I nuked several things. I'm kind of puzzled by the proposed solution though...

The previous version was composed of the following files:

```bash
├── asr
│ ...
├── transducer
│   ├── beam_search_transducer.py
│   ├── error_calculator.py
│   ├── joint_network.py
│   ├── transducer_decoder.py
│   └── utils.py
```
I can revert everything but it will lead to many duplications. If that's okay, I think we can modify the following to also ease the maintenance of the two versions:

- `joint_network.py `-> remove file and depend on `asr_transducer/joint_network.py`
- `transducer_decoder.py` -> move file to `asr/decoder/`
- `utils.py` -> remove the file and depend on `asr_transducer/utils.py` (the current file contains one method whereas the new version contains multiple ones)

I'm not sure what to do with `beam_search_transducer.py`, there are too many changes with the new version to avoid duplication. Maybe just move the old one to `asr/`?
Subsequently, I'm not sure what to do with `error_calculator.py`. For a unique class, I think the easiest thing is to pass the beam search object to `ErrorCalculator(...)` and change typing related to the decoder in the file to `torch.nn.Module`?

sw005320:
@b-flo, I'm asking @pyf98 to review this PR, but he is busy these days.
So, please wait for a while.

While we wait for his review, I'll list a couple of high-level comments.
- use our naming conventions for functions and variables: Please use the same or similar names as many as possible.
- minimize the configuration changes from the current ones. (it is similar to the above comment, but it is particularly important)
- minimize duplication of the core network model codes (e.g., conformer, lstm, transformer, etc.)

b-flo:
> While we wait for his review, I'll list a couple of high-level comments.
> ....

We're talking about the v1, right? If so, I reverted to what it was before this PR, no worries! No changes outside some files moved and a `JointNetwork(...)` I'm sharing between versions (I had to rename one variable `vocab_size` -> `dim_vocab` because of that).

sw005320:
No, I'm talking about this PR.
The following is a good example

image

`encoder_output_size` is used in many ASR codes.
Please do so.

chintu619:
Sorry in advance for the lengthy comments below.

Last week I started training the librispeech recipe with this PR. I used the auxiliary CTC loss and small weight decay from your Librispeech-100 training config. The model training went well, without any issues. I uploaded the pretrained model, training images etc. [here](https://huggingface.co/chaitu619/chai_librispeech_asr_train_transducer_v2_raw_en_bpe5000_sp). Currently I just uploaded this model to my personal HF hub. After the PR is merged, I can upload it to HF/espnet hub. 

Note: the pretrained model that I uploaded for transducer v1 as part of #4327, is actually from the above model training. Basically, I copied the pretrained model weights from v2 to v1. I ensured that both models were making identical computations, which explains why the dev/test scores are also identical. However, this process was a bit challenging because of the changes in names of encoder layers from v1 to v2. To maintain consistency with other ASR configs, would it be possible to retain the same layer names as v1? This will facilitate the use of other pretrained models using: `--init_param model.pth:encoder:encoder`.

I noticed that the `bias` of the linear decoder layer in the JointNetwork is set to `False` in v2 [here](https://github.com/espnet/espnet/pull/4032/files#diff-6cfe6d81ea7afb74ecf6d3f68885b20b509e281be1f196635168145b0835acb2R33). So when I copied the weights from v2 to v1, I set this bias to zero. When this PR is merged, I will update the pretrained model in #4327 by disabling this bias.

b-flo:
> I noticed that the bias of the linear decoder layer in the JointNetwork is set to False in v2 [here](https://github.com/espnet/espnet/pull/4032/files#diff-6cfe6d81ea7afb74ecf6d3f68885b20b509e281be1f196635168145b0835acb2R33).

Yes, that's intended! I don't recall the full discussion but we came to the conclusion with Hirofumi and Mingkun (warp-transducer author) that the bias in the decoder linear projection was redundant information. However, it shouldn't make a difference in practice.

espnet2/asr_transducer/beam_search_transducer.py
```diff
@@ -0,0 +1,638 @@
...
+            self.search_algorithm = self.time_sync_decoding
+        elif search_type == "alsd":
+            assert u_max >= 0, "u_max should be a positive integer, a portion of max_T."
+            self.u_max = u_max
+
+            self.search_algorithm = self.align_length_sync_decoding
+        elif search_type == "maes":
+            assert self.dim_vocab >= beam_size + expansion_beta, (
+                "beam_size (%d) + expansion_beta (%d) "
+                " should be smaller or equal to vocab size (%d)."
```
csukuangfj, line 103:
```suggestion
                " should be smaller than or equal to vocab size (%d)."
```

espnet2/asr_transducer/beam_search_transducer.py
```diff
@@ -0,0 +1,638 @@
...
+        """
+        final = []
+
+        for hyp in hyps:
+            seq_final = [f.yseq for f in final if f.yseq]
+
+            if hyp.yseq in seq_final:
+                seq_pos = seq_final.index(hyp.yseq)
+
+                final[seq_pos].score = np.logaddexp(final[seq_pos].score, hyp.score)
```
csukuangfj, line 179: Have you tried `max` instead of `log-sum-exp`?

espnet2/asr_transducer/beam_search_transducer.py
```diff
@@ -0,0 +1,638 @@
...
+            final: Recombined hypotheses.
+
+        """
+        final = []
+
+        for hyp in hyps:
+            seq_final = [f.yseq for f in final if f.yseq]
+
+            if hyp.yseq in seq_final:
+                seq_pos = seq_final.index(hyp.yseq)
```
csukuangfj, line 177: I am worried about the efficiency of this function.
Can we convert `yseq` to a string, store it in a dict, and use the dict for lookup?

espnet2/asr_transducer/beam_search_transducer.py
```diff
@@ -0,0 +1,638 @@
...
+                )
+                dec_out, state = self.decoder.score(
+                    label, max_hyp.yseq, max_hyp.dec_state, cache
+                )
+
+                logp = torch.log_softmax(
+                    self.joint_network(enc_out[t : t + 1, :], dec_out),
+                    dim=-1,
+                ).squeeze(0)
+                top_k = logp[1:].topk(beam_k, dim=-1)
```
csukuangfj, line 267: If you hard code blank id to 0, why do you need to use `self.blank_id` to store the blank id?

csukuangfj (changes requested):

doc/espnet2_tutorial.md
```diff
@@ -340,3 +340,213 @@ To enable online decoding, the argument `--use_streaming true` should be added t
 ### FAQ
 1. Issue about `'NoneType' object has no attribute 'max'` during training: Please make sure you employ `forward_train` function during traininig, check more details [here](https://github.com/espnet/espnet/issues/3803).
 3. I successfully trained the model, but encountered the above issue during decoding: You may forget to specify `--use_streaming true` to select streaming inference.
+
+## Transducer ASR
+
+> ***Important***: If you encounter any issue related to Transducer loss, please open an issue in [our fork of warp-transducer](https://github.com/b-flo/warp-transducer).
```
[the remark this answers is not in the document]
b-flo: Thanks a lot for pointing this out btw! I enable issues.

espnet2/asr_transducer/beam_search_transducer.py
```diff
@@ -0,0 +1,638 @@
...
+    yseq: List[int]
+    dec_state: Optional[Tuple[torch.Tensor, Optional[torch.Tensor]]]
+    lm_state: Optional[Union[Dict[str, Any], List[Any]]] = None
+
+
+@dataclass
+class ExtendedHypothesis(Hypothesis):
+    """Extended hypothesis definition for NSC beam search and mAES."""
+
+    dec_out: torch.Tensor = None
```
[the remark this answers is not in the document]
b-flo: It can't be None but because I'm extending `Hypothesis`, I could not set mandatory parameters after optional ones (`lm_state` from `Hypothesis` in that case).
I'm not sure what's the proper way here?

espnet2/asr_transducer/beam_search_transducer.py
b-flo: I did try max but it yields slightly worse performance in my experiments. What's the current conclusion on your side?

espnet2/asr_transducer/beam_search_transducer.py
b-flo: Btw, do you mean something like this?

```python3
    def recombine_hyps(
        self, hyps: List[Union[Hypothesis, ExtendedHypothesis]]
    ) -> List[Union[Hypothesis, ExtendedHypothesis]]:
        """Recombine hypotheses with same label ID sequence.

        Args:
            hyps: Hypotheses.

        Returns:
            final: Recombined hypotheses.

        """ 
        final = {}
                                                                                
        for hyp in hyps:
            str_yseq = "".join(map(str, hyp.yseq)
                      
            if str_yseq in final:
                final[str_yseq].score = np.logaddexp(final[str_yseq].score, hyp.score)
            else:
                final[str_yseq] = hyp
       
        return list(final.values())
```

Or do you mea, more generally, `yseq` should be defined as string instead of list of int?

b-flo: Sigh, I did the same mistake in the past and you already pointed that out in #3863. Thanks!

espnet2/asr_transducer/beam_search_transducer.py
csukuangfj: See https://github.com/k2-fsa/icefall/pull/420
and
https://github.com/k2-fsa/icefall/pull/277

log-add also performs better than max in our experiments

PR #3997, opened 2022-01-27
Add librispeech-100h recipe

YosukeHiguchi:
This PR adds librispeech-100h (train-clean-100) recipe for ESPnet1. This recipe can be used for evaluating models in a limited resource setting and is useful for conducting preliminary experiments of the full 960-hour set.

I'm planning to add the following models:

- [x] CTC
- [x] CTC/Attention
- [ ] Transducer

The training/decoding configurations are basically following [this paper](https://arxiv.org/pdf/2110.05249.pdf).

Overall results (w/o an LM and beam-search decoding):
|  |  dev_clean  |  dev_other  |  test_clean  |  test_other  |
| :----: | :----: | :----: | :----: | :----: |
| CTC |  7.4  |  20.2  |  7.7  |  20.6  |
| CTC/Attention |  7.2  |  18.5  |  7.3  |  19.3  |
| Transducer |  ***  |  ***  |  ***  |  ***  |

PR #4036, opened 2022-02-06
Update egs/librispeech_100

YosukeHiguchi:
This is a follow-up PR of #3997. I added results including Conformer-T and uploaded models.


sw005320:
Thanks, @YosukeHiguchi!!!

PR #4042, opened 2022-02-08
Update the baseline results for LibriSpeech 100h in ESPnet2

pyf98:
Hi, here is the baseline result for LibriSpeech 100h in ESPnet2. The encoder is Conformer. We do not use an LM. The BPE size is 5000. Automatic mixed precision (AMP) is enabled to speed up training.


sw005320:
Thanks!
We have a CI issue in #4043, which is fixed now.
So, could you merge the master?

sw005320:
Thanks, @pyf98!

PR #4043, opened 2022-02-08
fix ci for librosa update

ftshijt:


sw005320:
Thanks for your quick fix, @ftshijt!

PR #4052, opened 2022-02-10
Add installer for ParallelWaveGAN

ftshijt:
I'm adding some requirements on python and torch versions, cause I saw that there is an issue when doing torch 1.6 

Meanwhile, I'm adding `tools` into the environmental path so as to directly use binaries (e.g., jq) If you have any good suggestion, I'm happy to revise them.


tools/installers/install_parallel-wavegan.sh
```diff
@@ -0,0 +1,61 @@
+#!/usr/bin/env bash
```
kan-bayashi, line 1: The parallel-wavegan repo supports python 3.6+ and pytorch 1.4+.
We do not need to check the version here since espnet assumes python3.7+ and pytorch 1.4+.

kan-bayashi: And `jq` command is used for only recipe, in our usage we may not need it.

kan-bayashi: So we may make it much simpler.
```bash
#!/usr/bin/env bash

set -euo pipefail

if [ $# != 0 ]; then
    echo "Usage: $0"
    exit 1;
fi

rm -rf ParallelWaveGAN

# ParallelWaveGAN  Commit id when making this PR: `commit 5bef5a0610e5e0d6153b601bbf91c78582260c8f`
git clone https://github.com/kan-bayashi/ParallelWaveGAN.git
cd ParallelWaveGAN
pip install -e .
cd ..
```

Or further simplify like `pip install parallel_wavgan==0.5.4`

tools/extra_path.sh
```diff
@@ -27,3 +28,4 @@ export PATH="${TOOL_DIR}"/festival/bin:"${PATH:-}"
 export LD_LIBRARY_PATH="${TOOL_DIR}"/lib:"${TOOL_DIR}"/lib64:"${LD_LIBRARY_PATH:-}"
 export LD_LIBRARY_PATH="${TOOL_DIR}"/espeak-ng/lib:"${LD_LIBRARY_PATH:-}"
 export PYTHONPATH="${TOOL_DIR}"/s3prl:"${PYTHONPATH:-}"
+export PYTHONPATH="${TOOL_DIR}"/ParallelWaveGAN:"${PYTHONPATH:-}"
```
kan-bayashi, line 31: This is not needed since parallel_wavegan is installed as python library.

tools/extra_path.sh
```diff
@@ -13,6 +13,7 @@ else
     TOOL_DIR="$(pwd)"
 fi
 
+export PATH="${TOOL_DIR}":"${PATH:-}"
```
kan-bayashi, line 16: If we do not use jq, we may not need it.

ftshijt: I see, understand

ftshijt:
Make sense, thanks for the commetns. I just revised it to be simpler.

tools/installers/install_parallel-wavegan.sh
```diff
@@ -0,0 +1,10 @@
...
+
+set -euo pipefail
+
+if [ $# != 0 ]; then
+    echo "Usage: $0"
+    exit 1;
+fi
+
+pip install parallel_wavgan==0.5.4
```
kan-bayashi, line 10:
```suggestion
pip install parallel_wavegan==0.5.4
```

kan-bayashi (approved):

ftshijt:
Seems that the CI issues come from the timeout when downloading from google drive. We probably could either increase the timeout limit directly to pwg or maybe add an argument to `download_pretrained_model`? How do you think @kan-bayashi

kan-bayashi:
Oh sorry, I forgot about the test, which uses pretrained models.
I designed the test code to be running in local environment, so I did not care about the time.
I will modify the code.

kan-bayashi:
@ftshijt I modified the test in #4061. Could you merge it?

ftshijt:
Sure thing, thanks for your quick fix

sw005320:
FYI, @kan-bayashi, it is still >2.0s https://github.com/espnet/espnet/runs/5158915826?check_suite_focus=true#step:8:334
If it happens often, we may need to reduce the time more.

kan-bayashi:
Sorry, I will fix again.

kan-bayashi:
@ftshijt Again, could you merge the current master? (sorry for bothering you...)

ftshijt:
> @ftshijt Again, could you merge the current master? (sorry for bothering you...)

Done! Thanks for your quick fix

ftshijt:
Seems to be a precision error for the test? @kan-bayashi

kan-bayashi:
I miss it. I will fix again...

ftshijt:
Sorry, I made some commits from a wrong branch, will refactor it

PR #4058, opened 2022-02-10
Create decode_rnnt_conformer.yaml

sw005320:
From https://github.com/espnet/espnet/pull/3943#issuecomment-1015024576


b-flo:
@jeon30c I forgot to ask in the previous discussion: did you try other search algorithms?

jeon30c:
@b-flo No. I didn't. 
We are using our own beam search algorithm and developing a totally different type of algorithm. 
So, for baseline, the default beam search is used. 
We are in process to decide whether we can contribute our algorithms to espnet or not. I hope I can contribute to that part after the decision.

PR #4067, opened 2022-02-12
Add an optional suffix to the averaged model file name

pyf98:
Hi, this PR adds a suffix to the averaged model file name, if `nbest_averaging_interval` is used.

After every `nbest_averaging_interval` epochs, an averaged checkpoint will be saved with a suffix `till{xx}epoch.pth`. This is a unique name, so the previously saved model files will not be overwritten.

Note that the suffix is not added to the final averaged model after all epochs.

### DIFF
--- a/espnet2/main_funcs/average_nbest_models.py
+++ b/espnet2/main_funcs/average_nbest_models.py
@@ -1,5 +1,6 @@
 import logging
 from pathlib import Path
+from typing import Optional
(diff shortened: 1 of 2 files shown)


sw005320:
LGTM.
Thanks, @pyf98!
I'll merge this PR.

@kamo-naoyuki, it would be great if you give us some comments about this PR.

PR #4071, opened 2022-02-13
Restricted Self Attention for E2E Speech Summarization

roshansh-cmu:
Based on #3992


roshansh-cmu:
The issue appears to be related to PR #1491 where black requires imports to be in two line format for that line, while flake8 doesn't allow that. 
@kamo-naoyuki - do you have any suggestions to fix this ? Thanks!

sw005320:
Are all comments in the previous PR fixed?
I’ll also do second pass.

roshansh-cmu:
All comments were addressed bar 2 I think:
1. Testing for longformer Self-attention module
2. Testing for evaluation script

I have code for both- 
What is the order of testing, i.e., does the CI check install all optional and required modules before doing utils test and module test ?

roshansh-cmu:
Should work now- @sw005320 can you check and merge ?
I am unable to add tests because the longformer package is optional currently. But I will open a new PR and add that later

egs2/how2/sum1/local/run_asr.sh
```diff
@@ -0,0 +1,39 @@
+#!/usr/bin/env bash
```
sw005320, line 1: it would be better to move this to `egs2/how2/asr1/run.sh`.
You can do it in your follow up PR, later.

espnet2/asr/encoder/conformer_encoder.py
```diff
@@ -97,14 +105,17 @@ def __init__(
         macaron_style: bool = False,
         rel_pos_type: str = "legacy",
         pos_enc_layer_type: str = "rel_pos",
-        selfattention_layer_type: str = "rel_selfattn",
+        selfattention_layer_type: str = "legacy_rel_selfattn",
```
sw005320, line 108: ???
I think it's better not to change it.
Can you tell me why you need to change it?

sw005320 (commented):
- I think it is better to have `espnet2/asr/encoder/longformer_encoder.py`, instead of adding it to `espnet2/asr/encoder/conformer_encoder.py`. In this way, we can add other variants of transformer encoders in espnet by following the same manner. You can also inherit a conformer class to this.
- is this only applied to conformer (in this case, we may name it longconformer?)? can we also have an original transformer?

espnet2/asr/encoder/conformer_encoder.py
roshansh-cmu: Sorry this might have been a typo- I don't need to change it

roshansh-cmu:
So essentially, the change we are making is in 3 parts:

1.  Adding a different kind of self-attention layer that can be accessed by type "lf_selfattn"
2. Making sure that in the encoder, we pad the audio once to make sure that the signal length is a multiple of the window size
3. In the decoder, making sure that we account for this padding in the source target attention

I have my numbers with a conformed encoder and transformer decoder, but there is no reason why we can't apply this to an encode-decoder combination as long as 2 and 3 are handled

egs2/how2/sum1/local/run_asr.sh
roshansh-cmu: One issue might be that the current recipe is for How2-300h, while my run_asr operates on How2-2000h

sw005320:
I think we do not need to have that in the decoder.
Only having that in the encoder is fine.
Sorry for the confusion (although it would be great if we have that in the decoder, we can consider it later).

egs2/how2/sum1/local/run_asr.sh
sw005320: how about making a new directory then?
how2: 300h
how2_2000h: 2000h

roshansh-cmu: Will do in a future PR

espnet2/asr/decoder/transformer_decoder.py
```diff
@@ -128,6 +128,12 @@ def forward(
         memory_mask = (~make_pad_mask(hlens, maxlen=memory.size(1)))[:, None, :].to(
             memory.device
         )
+        # Padding for Longformer
```
roshansh-cmu, line 131: @sw005320 - This is the change in the decoder, i.e., while creating the mask for the encoded features, make-pad-mask uses the longest input length, but the longest length may be more than that because of the attention padding in the encoder. 
So for any decoder- transformer ,transducer etc, this is the only change to use longformer self-attention in the encoder

sw005320: ok

roshansh-cmu:
The CI test appears to be failing because of the recent ESPNet 2 enhancement for pytorch 1.9.1- do you have suggestions on how I can fix it ?

sw005320:
We're fixing it. Please ignore this.

roshansh-cmu:
Do let me know if anything else is required to complete the PR merge. Thanks

sw005320:
Sorry, I believe it is ready but we have a CI issue.

sw005320:
We fixed a CI issue.
Can you merge the master?

sw005320:
Again, is it possible to make `espnet2/asr/encoder/longformer_encoder.py` instead of adding longformer to `espnet2/asr/encoder/conformer_encoder.py`? `espnet2/asr/encoder/conformer_encoder.py` is a very important function and I don't want to put a lot of "if" and make it complicated.

roshansh-cmu:
I guess I can change create a longformer_encoder.py, but the only change will be in the init function, and padding in forward- if this is acceptable to you, I can create a duplicate and have the changed __init__ only in ``longformer_encoder.py``. The base architecture will still be a conformer (macron-style feedforward, convolution and self-attention), but with longformer attention instead of standard multi-head attention.

egs2/how2_2000h/asr1/local/score.sh
```diff
@@ -0,0 +1,25 @@
+#!/usr/bin/env bash
```
sw005320, line 1: Is it needed?

roshansh-cmu:
@sw005320 - I think all checks have passed now. Please merge this-thanks!

sw005320:
Many thanks!
This is a really valuable PR in various aspects.
@pyf98, maybe it’s time for you to add various transformer variants by following this PR ;)

sw005320:
Can you add a description to our main README.md?
Add the summarization recipe support in https://github.com/espnet/espnet#kaldi-style-complete-recipe and longformer supports in https://github.com/espnet/espnet#asr-automatic-speech-recognition

sw005320:
To advertise your great efforts?

roshansh-cmu:
Will do that in another PR shortly
Would you be interested in a Colab demo for the summarization task ? 
I have pretrained models that use raw audio on huggingface

PR #4078, opened 2022-02-16
Fix scripts related to speechbrain xvector extraction

seastar105:


kan-bayashi (approved):
Thank you for your kind fixing!

PR #4084, opened 2022-02-17
[ESPnet2] Intermediate/Self-conditioned CTC

YosukeHiguchi:
This PR adds implementations for Intermediate/Self-conditioned CTC (Inter/SC-CTC) on espnet2, along with a strong baseline configuration for training a pure Conformer-CTC model. I'm still running some experiments to verify that the implementation is accurate and after that I will add the results.

### DIFF
--- a/egs2/librispeech_100/asr1/conf/tuning/decode_ctc_bs1.yaml
+++ b/egs2/librispeech_100/asr1/conf/tuning/decode_ctc_bs1.yaml
@@ -1,0 +1,3 @@
+lm_weight: 0.0
+ctc_weight: 1.0
+beam_size: 1
--- a/egs2/librispeech_100/asr1/conf/tuning/train_conformer_ctc.yaml
+++ b/egs2/librispeech_100/asr1/conf/tuning/train_conformer_ctc.yaml
@@ -1,0 +1,61 @@
+batch_type: numel
+batch_bins: 2000000
+accum_grad: 16
+max_epoch: 60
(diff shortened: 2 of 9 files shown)


sw005320:
@brianyan918, can I ask you to review this PR?
Let's help @YosukeHiguchi and merge this PR ASAP

brianyan918:
I believe this implementation is well done. I agree with the way to pass the pointer to the ctc module from forward() instead of during init. I still need to fix that way in espnet1, because it would cause issues when using multi-gpu.

YosukeHiguchi:
First results on LibriSpeech-100h (w/o an LM and beam-search decoding):
|  |  dev_clean  |  dev_other  |  test_clean  |  test_other  |
| :----: | :----: | :----: | :----: | :----: |
| CTC |  7.6  |  21.2  |  7.9  |  21.4  |
| InterCTC |  6.8  |  20.0  |  7.1  |  20.3  |
| SC-CTC |  6.8  |  20.2  |  7.2  |  20.5  | 

Overall, the results are really good except the SC-CTC results are slightly worse than InterCTC's, which contradicts our observations on espnet1 (https://arxiv.org/pdf/2110.05249.pdf). I will look over my implementation and tune the models.

sw005320:
Can you add a test?

sw005320:
Thanks, @YosukeHiguchi!
It would be great if you keep on working on improving self-condition CTC and adding tests.

PR #4089, opened 2022-02-18
Add transformer results for LibriSpeech 100h

pyf98:
Hi, here is the Transformer baseline for LibriSpeech 100h. The Transformer encoder has 18 layers. As expected, the performance of Transformer is worse than that of Conformer.


sw005320:
Thanks, @pyf98!

PR #4090, opened 2022-02-19
Fix argument name asr_config in ESPnet2 doc

pyf98:
Hi, in ESPnet2, I think the argument name for asr training config should be `--asr_config` instead of `--asr_train_config`. I modified it in the documentation. I also saw a typo in the same file: `upstream` vs `upsteam`.


ftshijt:
Looks cool to me, thanks for the update

PR #4091, opened 2022-02-20
Update perturb_data_dir_speed.sh

AmirHussein96:
When I tried speed perturbation on my local data combined from several sources there was a difference in 2 segments vs utt2spk, which caused an issue. I had to use "fix_data_dir.sh" to fix this so I thought it better to add it here as well.


sw005320:
Thanks, @AmirHussein96!
This is a good idea!

Issue #4095, opened 2022-02-21
Hubert model training issues


arunbaby0:
I am trying to train a model with 500 hrs of custom data set using the config https://github.com/espnet/espnet/blob/master/egs2/librispeech/asr1/conf/tuning/train_asr_conformer7_hubert_ll60k_large.yaml.

There is a warning which is consistently showing while training:
`INFO: 1epoch:train:36631-42735batch: iter_time=1.299e-04, forward_time=0.113, loss_ctc=86.964, loss_att=58.959, acc=0.212, loss=66.878, backward_time=0.125, optim_step_time=0.053, optim0_lr0=0.001, train_time=0.408
[7cc1303a881b:0/8] 2022-02-21 11:42:05,031 (trainer:617) WARNING: The grad norm is nan. Skipping updating the model.
[7cc1303a881b:0/8] 2022-02-21 11:44:13,348 (trainer:617) WARNING: The grad norm is nan. Skipping updating the model.
[7cc1303a881b:0/8] 2022-02-21 11:44:54,398 (ctc:74) WARNING: 1/2 samples got nan grad. These were ignored for CTC loss.
[7cc1303a881b:0/8] 2022-02-21 11:45:47,820 (trainer:617) WARNING: The grad norm is nan. Skipping updating the model.
[7cc1303a881b:0/8] 2022-02-21 11:46:53,100 (trainer:617) WARNING: The grad norm is nan. Skipping updating the model.
[7cc1303a881b:0/8] 2022-02-21 11:47:18,511 (ctc:74) WARNING: 1/2 samples got nan grad. These were ignored for CTC loss.
[7cc1303a881b:0/8] 2022-02-21 11:51:51,166 (trainer:617) WARNING: The grad norm is nan. Skipping updating the model.
[7cc1303a881b:0/8] 2022-02-21 11:53:14,947 (ctc:74) WARNING: 1/2 samples got nan grad. These were ignored for CTC loss.
[7cc1303a881b:0/8] 2022-02-21 11:54:56,897 (trainer:617) WARNING: The grad norm is nan. Skipping updating the model.
[7cc1303a881b:0/8] 2022-02-21 11:55:38,768 (trainer:617) WARNING: The grad norm is nan. Skipping updating the model.
[7cc1303a881b:0/8] 2022-02-21 11:58:58,129 (trainer:617) WARNING: The grad norm is nan. Skipping updating the model.
[...]

PR #4098, opened 2022-02-21
fix errors in configs of librispeech ssl frontends

simpleoier:
related to #4095


sw005320:
Thanks for the quick fix!

PR #4116, opened 2022-02-27
Update the example PyTorch version in Installation doc

pyf98:
Hi, I updated the example PyTorch and CUDA version in the installation doc to avoid confusion. Originally, the PyTorch version was 1.3.1, but according to https://github.com/espnet/espnet/pull/4038, it is no longer supported.


sw005320:
Thanks!

PR #4122, opened 2022-03-02
set gtn version

brianyan918:
Looks like pip has a new version of gtn==0.0.1 released on 2/28. Perhaps that's the reason for the CI problem, so I'm trying to set this version.


sw005320:
super thanks!
We still did not pass one condition (https://github.com/espnet/espnet/runs/5390324073?check_suite_focus=true#step:8:281)
Could you check it?

PR #4125, opened 2022-03-02
Mandarin ISO id should be CMN instead of ZHO

xinjli:
Mandarin ISO id should be CMN instead of ZHO
- ZHO is the macro language id including many Chinese dialects, 
- CMN is the individual language id corresponding to Mandarin Chinese.

See this reference link: https://iso639-3.sil.org/code/zho


ftshijt:
Wow, it's cool to know the difference, many thanks @xinjli

PR #4130, opened 2022-03-04
Add iFaSNet for enhancement/separation tasks.

LiChenda:
This is an implemntation of [iFaSNet](https://arxiv.org/abs/1910.14104).
The code is based on Yi Luo's [implementation](https://github.com/yluo42/TAC/blob/master/iFaSNet.py).
I also add a corresponding configuration file for CHiME4 recipe. A pretrained model on CHiME4 will be uploaded to hugginface in near future.

### DIFF
--- a/egs2/chime4/enh1/conf/tuning/train_enh_dprnntac_ifasnet.yaml
+++ b/egs2/chime4/enh1/conf/tuning/train_enh_dprnntac_ifasnet.yaml
@@ -2,7 +2,7 @@ optim: adam
 init: xavier_uniform
 max_epoch: 100
 batch_type: folded
-batch_size: 16 # batch_size 16 can be trained on 4 RTX 2080ti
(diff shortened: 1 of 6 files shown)


Emrys365:
Could you also add the original FasNet model?
They are based on the same codebase, and it would be convenient to have both models in ESPnet.

sw005320:
LGTM.
@Emrys365, if you're OK, you can marge it.

@LiChenda, it would be great if you prepare a pre-trained model and result later.

sw005320:
LGTM.
@Emrys365, if you're OK, you can marge it.

@LiChenda, it would be great if you prepare a pre-trained model and result later.

LiChenda:
Thanks for the comments! As @Emrys365 suggested, I also add the origin fasnet. The pretrained model will be uploaded soon.

Emrys365 (approved):
Thanks a lot, @LiChenda

PR #4134, opened 2022-03-04
ESPnet2 swbd_sentiment recipe

YushiUeda:
This PR adds ESPnet2 swdb_sentiment recipe.


egs2/swbd_sentiment/asr1/conf/decode_asr.yaml
```diff
@@ -0,0 +1 @@
+beam_size: 10
```
ftshijt, line 1: Do we have a sequence of output? If not, `beam_size` to 1 would be fine (just as other slu tasks)

egs2/swbd_sentiment/asr1/conf/tuning/train_asr_conformer_wav2vec2.yaml
```diff
@@ -0,0 +1,89 @@
+# network architecture
```
ftshijt, line 1: The wav2vec2 and hubert confs are almost the same, can you combine them by adding comments instead of a new one? As for initial PR, we would recommend to simplify the config.

ftshijt (commented):
Many thanks! A few minor comments as attached

ftshijt:
LGTM! Will merge it after CI passed. Thanks!

YushiUeda:
@ftshijt thanks for the review!

PR #4137, opened 2022-03-06
fix bug where asr multitasking uses st decoder

brianyan918:


ftshijt:
Thanks Brian for the fixing!

PR #4143, opened 2022-03-07
Fix joint tokenization in st.sh

pyf98:
Hi, in `st.sh`, the source and target texts should be merged before tokenization, if `token_joint` is `True`.

This modification is based on the same stage in `mt.sh`. Please check if it's correct.

### DIFF
--- a/egs2/TEMPLATE/st1/st.sh
+++ b/egs2/TEMPLATE/st1/st.sh
@@ -296,18 +296,8 @@ fi
 # Extra files for translation process
 utt_extra_files="text.${src_case}.${src_lang} text.${tgt_case}.${tgt_lang}"
 # Use the same text as ST for bpe training if not specified.
-if "${token_joint}"; then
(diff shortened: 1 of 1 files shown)


ftshijt:
Looks cool to me, I will merge it after CI passed. Many thanks!

siddalmia (changes requested):
https://github.com/espnet/espnet/blob/3ae0dcc9442981274a98ab4cd375e381e4e2b491/egs2/TEMPLATE/st1/st.sh#L299-L307

This will no longer be needed.

pyf98:
> https://github.com/espnet/espnet/blob/3ae0dcc9442981274a98ab4cd375e381e4e2b491/egs2/TEMPLATE/st1/st.sh#L299-L307
> 
> This will no longer be needed.

Thanks for pointing it out. I removed these lines. Could you review it again?

siddalmia:
Looks good!

PR #4148, opened 2022-03-07
Speech translation ensemble

ftshijt:


siddalmia:
Can we add an option of doing ensembling with an MT model? Do you think that would be useful? @brianyan918 like ensemble cascade and an e2e model.

ftshijt:
> It has some dsing files.

I notice that, will update

ftshijt:
> Can we add an option of doing ensembling with an MT model? Do you think that would be useful? @brianyan918 like ensemble cascade and an e2e model.



> Can we add an option of doing ensembling with an MT model? Do you think that would be useful? @brianyan918 like ensemble cascade and an e2e model.

We are discussing the probability, there are some concerns in the design stage, but hopefully, we can handle it

PR #4149, opened 2022-03-07
fix bug in mt/st templates for having separate token lists

brianyan918:


siddalmia:
Looks good! Good catch.

brianyan918:
I added a corresponding change to st/espnet_model.py. These bug fixes are tested and I think that now separate src and target vocabs is fully supported.

siddalmia:
@brianyan918 Good catch of this subtle bug! 

One comment - can you rename `show_st_result.sh` to `toshow_translation_result.sh` so that it works for both mt and st. Following the PR - https://github.com/espnet/espnet/pull/4146

ftshijt:
Cool! It looks cool to me, there is one case that needs to concerned is the sacrebleu has a different layout for 1.5.1 and 2.0, maybe we should also fix the change there?

egs2/TEMPLATE/asr1/scripts/utils/show_st_result.sh
```diff
@@ -0,0 +1,65 @@
+#!/usr/bin/env bash
```
ftshijt, line 1: We can just remove the file

sw005320:
Can you merge the latest master?
After that, I'll merge this PR.

sw005320:
Thanks, @brianyan918!

PR #4152, opened 2022-03-08
Add new config and support for computing WER in SLUE-VoxCeleb

siddhu001:


pyf98:
Thanks for the PR! I have one comment on the design of `local/score.sh`. Currently the `inference_tag` is somehow fixed because the default value is used. If we specify `decode_asr.yaml` or a different `inference_tag`, the scoring cannot be executed.

siddhu001:
It can. Run-

./local/score.sh <asr-exp-dir> <valid_inference_folder> <test_inference_folder>

Here in the valid inference folder you can specify the decode asr's inference folder (e.g. decode_asr_asr_model_valid.acc.ave_10best/devel/). Let me know if you have any further questions

pyf98:
Thank you for the reply. Yeah, we can manually run the scoring script after decoding. That's what I used to do. Now I just modify the default path and don't specify a new name.

I think this can be a TODO item in our future development. Since all the information are available in the `run.sh` or `asr.sh`, the script should automatically determine the decoding folder name and show the SLU results in `RESULTS.md`.

sw005320:
Good suggestion, @pyf98 
@siddhu001, can you add a TODO to `run.sh` or `local/score.sh`?

siddhu001:
That makes lot of sense @pyf98.

@sw005320 , Sure I will add a to-do to local/score.sh for this!

sw005320:
LGTM.
@siddhu001, is it ready to merge?

siddhu001:
Yeah It is ready to merge! I updated some results on Readme.

sw005320:
Thanks, @siddhu001!
Sorry for my late response.

chintu619:
> That makes lot of sense @pyf98.
> 
> @sw005320 , Sure I will add a to-do to local/score.sh for this!

I had this similar issue when running `score.sh`. I am not sure about the original intent of line 1376 in `asr.sh` [here](https://github.com/espnet/espnet/blob/master/egs2/TEMPLATE/asr1/asr.sh#L1376).

Correct me if I'm wrong. The inference (stage 12) and scoring (stage 13) only make changes in `${asr_exp}/${inference_tag}`. So I expected the scoring command to be: 
`[ -f local/score.sh ] && local/score.sh ${local_score_opts} "${asr_exp}/${inference_tag}"`, instead of 
`[ -f local/score.sh ] && local/score.sh ${local_score_opts} "${asr_exp}"`
This modification will ensure that we only have to specify the names of validation and test splits in `local_score_opts`

PR #4155, opened 2022-03-08
Update README.md for How2 2000h ASR,SUM

roshansh-cmu:
Minor update in README for How2 ASR and SUM based on #4150 #4071


sw005320:
Thanks, @roshansh-cmu!

PR #4158, opened 2022-03-11
[WIP] [ESPnet2] Mask-CTC

YosukeHiguchi:
This PR adds [Mask-CTC](https://arxiv.org/abs/2005.08700)-based non-autoregressive ASR model to ESPnet2.


YosukeHiguchi:
@sw005320 Are these CI errors related to my commits?

YosukeHiguchi:
I think the implementation is ready. I'll also add WSJ results along with baseline CTC results.

egs2/wsj/asr1/conf/tuning/train_asr_transformer_maskctc.yaml
```diff
@@ -0,0 +1,65 @@
+batch_type: folded
```
sw005320, line 1: Can you add some remarks (comments) on which parts are different from the original training config?
Obviously decoder, right? Other parts?

espnet2/tasks/asr.py
```diff
@@ -232,12 +247,6 @@ def add_task_arguments(cls, parser: argparse.ArgumentParser):
             default=None,
             help="The keyword arguments for joint network class.",
         )
-        group.add_argument(
-            "--model_conf",
```
sw005320, line 236: Sorry I cannot understand this part.
Why this option is removed?

sw005320 (commented):
LGTM.
Some minor comments

espnet2/tasks/asr.py
YosukeHiguchi: I've defined `model_choices` in L112 and added it to `class_choices_list` in L187. With this change, arguments from --model and --model_conf should be loaded by the `add_arguments` method in L330.

sw005320: Thanks. How is the help message (`help="The keyword arguments for model class."`) handled?

YosukeHiguchi: It will be handled in https://github.com/espnet/espnet/blob/master/espnet2/train/class_choices.py#L94

sw005320: thanks!

sw005320:
Thanks a lot.
It would be great if you add some results, models, and comments to the config files.

PR #4162, opened 2022-03-13
Add new conformer config with hop length 160 for LibriSpeech 960h

pyf98:
Hi, I am sharing a new Conformer config for LibriSpeech 960h. The `hop_length` in `frontend_conf` is reduced from 256 to 160. I will update the results once the decoding finishes. Currently, the WERs on dev/test-other sets are lower without an LM.


pyf98:
There is an error related to zenodo time out:
```
requests.exceptions.ConnectionError: HTTPSConnectionPool(host='zenodo.org', port=443): Read timed out.
```

Issue #4105, opened 2022-02-24
nGram based decoding


arunbaby0:
when trying to decode using the n-gram in the latest version, 0.10.6, the decoding doesn't take the ngram properly, however, this was working fine with the 0.10.4 version. This is the kenlm based n-gram decode.
Any recent changes are done in the codebase affects the n-gram based decode.

The following is the args passed to the asr_inference_tool which is set to espnet2.bin.asr_inference: 
` --config /opt/espnet2/egs2/test/asr1/conf/decode_asr.yaml --ngram_file exp/ngram/model/5gram.bin`

Also, is it better to use k2 based decoding for this? Any specific advantage for each of these methods

sw005320:
Thanks for your report.
Could you let me know what kind of issues appear?

Also, yes, k2-CTC is a very good option.
You can refer to https://github.com/espnet/espnet/tree/master/egs2/librispeech/asr1#ctc-decoding-with-nbest-rescoring-from-decoder-and-lm-using-k2

arunbaby0:
while decoding the ngram model is not taken into account in the latest version it looks like.
the following is the decode config used:
`ctc_weight: 0.4
ngram_weight: 0.8
beam_size: 60`

With the new code(0.10.6) base the log looks like:
`2022-02-22 14:14:46,892 (beam_search:362) INFO: decoder input length: 23
2022-02-22 14:14:46,892 (beam_search:363) INFO: max output length: 23
2022-02-22 14:14:46,892 (beam_search:364) INFO: min output length: 0
2022-02-22 14:14:49,398 (beam_search:376) INFO: end detected at 18
2022-02-22 14:14:49,399 (beam_search:400) INFO:  -5.02 * 0.6 =  -3.01 for decoder
2022-02-22 14:14:49,399 (beam_search:400) INFO:  -3.10 * 0.4 =  -1.24 for ctc
2022-02-22 14:14:49,399 (beam_search:403) INFO: total log probability: -4.26
2022-02-22 14:14:49,399 (beam_search:404) INFO: normalized log probability: -0.35
2022-02-22 14:14:49,399 (beam_search:405) INFO: total number of ended hypotheses: 279`


and for old(0.10.4):
`2022-02-23 06:01:02,856 (beam_search:364) INFO: min output length: 0
2022-02-23 06:01:12,571 (batch_beam_search:321) INFO: adding <eos> in the last position in the loop
2022-02-23 06:01:12,574 (beam_search:376) INFO: end detected at 8
[...]

PR #4168, opened 2022-03-17
Fix n-gram decoding

sw005320:
https://github.com/espnet/espnet/issues/4105


sw005320:
According to @arunbaby0, this will fix the issue.
But I want @tanghaoyu to double-check it to make sure.

PR #4171, opened 2022-03-19
Fix notebook in the official doc.

ShigekiKarita:
I just noticed our gh-page doc does NOT contain notebooks https://espnet.github.io/espnet/
This is because our gh action forget submodules. Moreover, the submodule was **not updated for two years**.

For simplicity, we should clone the latest snapshot and generate rst during build. See https://karita.xyz/espnet/ for the updated doc

image


sw005320:
Thanks a lot, @ShigekiKarita!
This is a great idea!

PR #4196, opened 2022-03-24
bug fixes and efficient train/dev split in data prep of Microsoft Indian Languages recipe

chintu619:


sw005320:
I'm not sure, but your PR does not go through the CI check...

chintu619:
Sorry to break the CI for others as well :( 
I may have found the bug. Could you please check again? I think previously the code was trying to create a null directory

xinjli:
This is just a summary I mentioned in our Slack channel

The CI fails here because it fails to pass the black format check

```
if ! black --check ${modules}; then
    printf 'Please apply:\n    $ black %s\n' "${modules}"
    exit 1
fi
```

The format error required by black seems to be difficult to catch by youself, for example, it requires space to be inserted before and after operators

```
-with open(train_datadir+'text', 'w') as f:
+with open(train_datadir + "text", "w") as f:
```

and

```
-valid_totaldur = 2*60*60 # (in seconds) 2 hours taken for validation split
+valid_totaldur = 2 * 60 * 60  # (in seconds) 2 hours taken for validation split
```

We need to make sure everyone is running black --check before submitting their pull request.

In the checking CI script, it might be better to modify it to output more details using --diff flag to see why it fails 
```
if ! black --check ${modules} --diff; then
    printf 'Please apply:\n    $ black %s\n' "${modules}"
    exit 1
fi
```

If it makes sense, I will submit a PR to update this CI script.

sw005320:
Thanks, @xinjli for the investigation!
Actually, it happens the first time in our thousands of PRs.
So, I think we may not need it, but your suggestion of "We need to make sure everyone is running black --check before submitting their pull request." would make sense to me.

Let me try to establish it.

PR #4206, opened 2022-03-25
fix a bug in Mandarin pypinyin_g2p_phone

WeiGodHorse:
I found that when using the latest version of pypinyin (0.46.0), pypinyin_g2p_phone for Mandarin will drop the finals. This phenomenon does not exist in pypinyin (0.43.0). I fixed this bug by modifying espnet2/text/phone_tokenizer.py.  Also, I found that the previous code would treat a single tone as a phoneme, which didn't make sense, so I fixed that as well. Espnet is a very powerful repository. Hope my suggestion helps the repository to be more robust on Mandarin, thanks!


WeiGodHorse:
> * Can you merge the latest master branch?
> * Can you fix the CI error https://github.com/espnet/espnet/runs/5693967837?check_suite_focus=true#step:8:9?

I cloned the latest code and only changed two lines of code in one file, but there were still some CI error. I'm not quite sure why this error occurs. Could you please tell me how to troubleshoot this error?

sw005320:
Check https://digitaldrummerj.me/git-sync-fork-to-master/

sw005320:
Can you fix the CI error?
https://github.com/espnet/espnet/runs/6292366451?check_suite_focus=true#step:8:10

WeiGodHorse:
> Can you fix the CI error? https://github.com/espnet/espnet/runs/6292366451?check_suite_focus=true#step:8:10
Hi, I have formatted the code with black. I didn't understand where the problem was before, it turned out to be a formatting problem. Thank you very much for your patient help.

sw005320:
There is another CI error
https://github.com/espnet/espnet/runs/6320224007?check_suite_focus=true#step:8:787
Is it expected?
If so, we may also change the test part.

WeiGodHorse:
> There is another CI error https://github.com/espnet/espnet/runs/6320224007?check_suite_focus=true#step:8:787 Is it expected? If so, we may also change the test part.


Oh, my previous revision lost the punctuation, I've fixed this mistake now.

kan-bayashi (commented):
Sorry for the late review. It may break the compatibility of the pretrained models.
How do you think, the effect is serious or not?

WeiGodHorse:
> Sorry for the late review. It may break the compatibility of the pretrained models. How do you think, the effect is serious or not?

I don't think the effect on the pre-trained model will be great, as only individual phoneme markers change, such as the "un" phoneme in the old version of pypinyin that I found changed to "uen". I think it only needs to replace the corresponding tokens in the token list without affecting the pre-training model. Or more simply, the bug occurred because the code is not compatible with the new version of pypinyin, and specifying to use the previous version of pypinyin can also solve the problem.

kan-bayashi:
I see. Thank you for your explanation!

Issue #4202, opened 2022-03-25
Error in espnet2 ASR streaming inference step


jhlee9010:
Hello,
I think there is a bug in ASR streaming inference step.

In 'apply_frontend' function, some fixed values are used instead of parameters which used for training. 
[https://github.com/espnet/espnet/blob/master/espnet2/bin/asr_inference_streaming.py#L203-L208](url)
I think it should be change like this below.
`            n_frames = (speech.size(0) - self.n_fft) // self.hop_length
            n_residual = (speech.size(0) - self.n_fft) % self.hop_length
            speech_to_process = speech.narrow(0, 0, self.n_fft + n_frames * self.hop_length)
            waveform_buffer = speech.narrow(
                0, speech.size(0) - self.n_fft - n_residual, self.n_fft + n_residual
            ).clone()
`
Also, fixed value used in [https://github.com/espnet/espnet/blob/master/espnet2/bin/asr_inference_streaming.py#L232-L237](url).
It would be better to be changed according to the self.n_fft used for training.

Is it makes sense??

sw005320:
Thanks a lot for your report.
We're checking it. @D-Keqi, @eml914

jhlee9010:
I created a PR for this issue. (https://github.com/espnet/espnet/pull/4216)

PR #4216, opened 2022-03-29
Bugfix in streaming inference

jhlee9010:
Related to issues #4202 (https://github.com/espnet/espnet/issues/4202)


sw005320:
Thanks, @jhlee9010!
Can you fix the CI issue?
https://github.com/espnet/espnet/runs/5740058274?check_suite_focus=true#step:8:9

@D-Keqi, can you review this PR?

D-Keqi:
> Thanks, @jhlee9010! Can you fix the CI issue? https://github.com/espnet/espnet/runs/5740058274?check_suite_focus=true#step:8:9
> 
> @D-Keqi, can you review this PR?

Sure!

D-Keqi (commented):
Hi! @jhlee9010  Thanks for your work! The fixed value should definitely be solved. 
But I think this PR may come into 2 issues.
The first is that 'Speech2TextStreaming' object has no attribute 'n_fft' actually in current implementation. And the "n_fft" actually should be "win_length", although the "win_length" is always set to the same value as the "n_fft".
The second is that I think the "n_fft" is actually should be "n_fft - hop_length" or "win_length - hop_length".

jhlee9010:
@D-Keqi Thank you for the review. 
1. I forgot to add some lines in initialization function. I'll fix it.
   "win_length" is right, is there any way to get "win_length" value? "win_length" is not exist in .yaml file.
2. Which line of "n_fft" should be "n_fft - hop_length"?
    Totally? or just in trimming part?

D-Keqi:
> @D-Keqi Thank you for the review.
> 
> 1. I forgot to add some lines in initialization function. I'll fix it.
>    Also, "win_length" is right. Is there any good way to get "win_length" value? "win_length" is not exist in ".yaml" file.
> 2. "n_fft" should be "n_fft - hop_length" means this line?
>    https://github.com/espnet/espnet/pull/4216/files#diff-a09b1920d44b5255b5125bf8c5fd82107c71559044b8f75c7f7caed3a3042ac4R237
>    Then it might be changed like this `feats = feats.narrow(1, 2, feats.size(1) - (n_fft - hop_length)//self.hop_length - 2)`
1. Yes, win_length is always not specified in the yaml, maybe you can use a conditional statement. like if win_length is None, win_length = n_fft.
2. https://github.com/espnet/espnet/pull/4216/files#diff-a09b1920d44b5255b5125bf8c5fd82107c71559044b8f75c7f7caed3a3042ac4R237 yes, but except this, I think all your "n_fft" should be "n_fft - hop_length". I think the default value of n_fft is 512 and the one for hop_length is 128, so the 384 actually corresponds to "n_fft - hop_length". And maybe you can have a test before PR to make sure it works if this is available for you.

D-Keqi:
> Sorry for bothering, I might misunderstood something. Therefore, I want to clarify about it especially trimming part, before I create new commit.
> 
> Originally, `feats = feats.narrow(1, 2, feats.size(1) - 4)` was used, these trimming variables seem to be derived from n_fft/2 padding on both sides in 'Stft' function. So it should be changed as `feats = feats.narrow(1, (self.win_length//2)//self.hop_length, feats.size(1) - (self.win_length-self.hop_length)//self.hop_length - 1)` Is it right?
> 
> It seemed like the attached file was going well in 'espnet2_streaming_asr_demo.ipynb' and my model(n_fft=512, hop_size=256). [asr_inference_streaming.py.zip](https://github.com/espnet/espnet/files/8389831/asr_inference_streaming.py.zip)

Thanks for your hard work! I think it may should be `feats = feats.narrow(1, math.ceil(math.ceil(self.win_length/self.hop_length)/2), feats.size(1) - math.ceil(self.win_length/self.hop_length)`

The reason I think should be related with the simulated chunk setting https://github.com/espnet/espnet/blob/master/espnet2/bin/asr_inference_streaming.py#L419 and the waveform_buffer inheritance mechanism https://github.com/espnet/espnet/blob/master/espnet2/bin/asr_inference_streaming.py#L206

jhlee9010:
Thank you! I revised it with your help.

jhlee9010:
> Thanks, @jhlee9010! Can you fix the CI issue? https://github.com/espnet/espnet/runs/5740058274?check_suite_focus=true#step:8:9
> 
> @D-Keqi, can you review this PR?

I hope this commit fix the CI issue

espnetUser:
@jhlee9010 and @D-Keqi: I am also working on espnet2 asr streaming model and tried your fix for espnet2/bin/asr_inference_streaming.py in my setup because I noticed that the hardcoded win_length and hop_length values do not fit my own Conformer model config:
```
win_length: 256
n_fft: 256
hop_length = 128
```
With your fix I noticed that for some audio files I see this error:
```
File "espnet/tools/anaconda/envs/espnet/lib/python3.8/site-packages/torch/nn/functional.py", line 4179, in _pad
return torch._C._nn.reflection_pad1d(input, pad)
RuntimeError: Argument #4: Padding size should be less than the corresponding input dimension, but got: padding (128, 128) at dimension 2 of input [1, 1, 128]
```
In order to make sure the input before padding is always larger than pad size (n_fft/2) I modified the waveform buffer object in your fix from
```
waveform_buffer = speech.narrow(
    0,
    speech.size(0) - (self.win_length - self.hop_length) - n_residual,
    (self.win_length - self.hop_length) + n_residual,
).clone()
```
to
```
waveform_buffer = speech.narrow(
    0,
    speech.size(0) - (self.n_fft // 2) - n_residual - 1,
    (self.n_fft // 2) + n_residual + 1,
).clone()
```
With that change the error disappeared for my setup and all audio files got decoded.

Could you please check if that change makes sense or if there is a better solution?

D-Keqi:
> @jhlee9010 and @D-Keqi: I am also working on espnet2 asr streaming model and tried your fix for espnet2/bin/asr_inference_streaming.py in my setup because I noticed that the hardcoded win_length and hop_length values do not fit my own Conformer model config:
> 
> ```
> win_length: 256
> n_fft: 256
> hop_length = 128
> ```
> 
> With your fix I noticed that for some audio files I see this error:
> 
> ```
> File "espnet/tools/anaconda/envs/espnet/lib/python3.8/site-packages/torch/nn/functional.py", line 4179, in _pad
> return torch._C._nn.reflection_pad1d(input, pad)
> RuntimeError: Argument #4: Padding size should be less than the corresponding input dimension, but got: padding (128, 128) at dimension 2 of input [1, 1, 128]
> ```
> 
> In order to make sure the input before padding is always larger than pad size (n_fft/2) I modified the waveform buffer object in your fix from
> 
> ```
> waveform_buffer = speech.narrow(
>     0,
>     speech.size(0) - (self.win_length - self.hop_length) - n_residual,
>     (self.win_length - self.hop_length) + n_residual,
> ).clone()
> ```
> 
> to
> 
> ```
> waveform_buffer = speech.narrow(
>     0,
>     speech.size(0) - (self.n_fft // 2) - n_residual - 1,
>     (self.n_fft // 2) + n_residual + 1,
> ).clone()
> ```
> 
> With that change the error disappeared for my setup and all audio files got decoded.
> 
> Could you please check if that change makes sense or if there is a better solution?

I think this may not make sense, can you try decoding with larger 'sim_chunk_length'? https://github.com/espnet/espnet/blob/master/espnet2/bin/asr_inference_streaming.py#L493 I think this may help to solve your issue.

D-Keqi:
@jhlee9010 Hi, sorry, as for the line 274 https://github.com/espnet/espnet/pull/4216/files#diff-a09b1920d44b5255b5125bf8c5fd82107c71559044b8f75c7f7caed3a3042ac4R274 can you kindly change the 'math.ceil(self.win_length/self.hop_length)' to '2*math.ceil(math.ceil(self.win_length/self.hop_length)/2)'? I think this is more stable and makes sense, what do you think?

espnetUser:
> I think this may not make sense, can you try decoding with larger 'sim_chunk_length'? 
> https://github.com/espnet/espnet/blob/master/espnet2/bin/asr_inference_streaming.py#L493 
> I think this may help to solve your issue.

I tried with increasing sim_chunk_length from 512 to 1024 and get the same padding error:

sim_chunk_length: 512
```
2022-04-07 08:49:42,214 (asr_inference_streaming:248) DEBUG: speech_lengths: tensor([640])
2022-04-07 08:49:42,217 (asr_inference_streaming:248) DEBUG: speech_lengths: tensor([128])
...
    return torch._C._nn.reflection_pad1d(input, pad)
RuntimeError: Argument #4: Padding size should be less than the corresponding input dimension, but got: padding (128, 128) at dimension 2 of input [1, 1, 128]
```

sim_chunk_length: 1024
```
2022-04-07 08:48:01,389 (asr_inference_streaming:248) DEBUG: speech_lengths: tensor([1152])
2022-04-07 08:48:01,394 (asr_inference_streaming:248) DEBUG: speech_lengths: tensor([128])
...
    return torch._C._nn.reflection_pad1d(input, pad)
RuntimeError: Argument #4: Padding size should be less than the corresponding input dimension, but got: padding (128, 128) at dimension 2 of input [1, 1, 128]
```

How does the code guarantee that input to stft function is always larger than padding size (n_fft/2)?

D-Keqi:
> @jhlee9010 and @D-Keqi: I am also working on espnet2 asr streaming model and tried your fix for espnet2/bin/asr_inference_streaming.py in my setup because I noticed that the hardcoded win_length and hop_length values do not fit my own Conformer model config:
> 
> ```
> win_length: 256
> n_fft: 256
> hop_length = 128
> ```
> 
> With your fix I noticed that for some audio files I see this error:
> 
> ```
> File "espnet/tools/anaconda/envs/espnet/lib/python3.8/site-packages/torch/nn/functional.py", line 4179, in _pad
> return torch._C._nn.reflection_pad1d(input, pad)
> RuntimeError: Argument #4: Padding size should be less than the corresponding input dimension, but got: padding (128, 128) at dimension 2 of input [1, 1, 128]
> ```
> 
> In order to make sure the input before padding is always larger than pad size (n_fft/2) I modified the waveform buffer object in your fix from
> 
> ```
> waveform_buffer = speech.narrow(
>     0,
>     speech.size(0) - (self.win_length - self.hop_length) - n_residual,
>     (self.win_length - self.hop_length) + n_residual,
> ).clone()
> ```
> 
> to
> 
> ```
> waveform_buffer = speech.narrow(
>     0,
>     speech.size(0) - (self.n_fft // 2) - n_residual - 1,
>     (self.n_fft // 2) + n_residual + 1,
> ).clone()
> ```
> 
> With that change the error disappeared for my setup and all audio files got decoded.
> 
> Could you please check if that change makes sense or if there is a better solution?

Considering your configuration of win_length and hop_length, it seems that it is the extra "1" that fixes your issue?

espnetUser:
I did some more debugging and think for my audio file the problem is this:

```
len(speech) = 67584
sim_chunk_length = 512
len(speech) // sim_chunk_length = 132 
```
So in this case len(speech) is exactly a multiple of sim_chunk_length and the final speech2text call 

```
 results = speech2text(
     speech[(i + 1) * sim_chunk_length : len(speech)], is_final=True
 )
```
will be with with an empty 'speech' object as index slice "(i + 1) * sim_chunk_length : len(speech)" will be "(131 + 1) * 512 : 67584 = 67584:67584". So this means the next input will be just the previous waveform buffer:
```
speech = torch.cat([buf, speech], dim=0)
```
which happens to be 128 samples. My n_fft size is 256 so padding size (n_fft/2) is also 128 which causes the error I posted above.

D-Keqi:
> I did some more debugging and think for my audio file the problem is this:
> 
> ```
> len(speech) = 67584
> sim_chunk_length = 512
> len(speech) // sim_chunk_length = 132 
> ```
> 
> So in this case len(speech) is exactly a multiple of sim_chunk_length and the final speech2text call
> 
> ```
>  results = speech2text(
>      speech[(i + 1) * sim_chunk_length : len(speech)], is_final=True
>  )
> ```
> 
> will be with with an empty 'speech' object as index slice "(i + 1) * sim_chunk_length : len(speech)" will be "(131 + 1) * 512 : 67584 = 67584:67584". So this means the next input will be just the previous waveform buffer:
> 
> ```
> speech = torch.cat([buf, speech], dim=0)
> ```
> 
> which happens to be 128 samples. My n_fft size is 256 so padding size (n_fft/2) is also 128 which causes the error I posted above.

I see, thanks for all your hard work!

jhlee9010:
> math.ceil(self.win_length/self.hop_length)

Sure!

jhlee9010:
> @jhlee9010 Hi, sorry, as for the line 274 https://github.com/espnet/espnet/pull/4216/files#diff-a09b1920d44b5255b5125bf8c5fd82107c71559044b8f75c7f7caed3a3042ac4R274 can you kindly change the 'math.ceil(self.win_length/self.hop_length)' to '2*math.ceil(math.ceil(self.win_length/self.hop_length)/2)'? I think this is more stable and makes sense, what do you think?

Sure!

espnetUser:
@D-Keqi and @jhlee9010, I worked a bit further on the two (edge case) errors I am seeing for my setup when 
```
1. sim_chunk_length <= n_fft

    1.a) sim_chunk_length < (n_fft // 2):
    IndexError: Dimension out of range (expected to be in range of [-64, 63], but got -128)

    1.b) (n_fft // 2) <= sim_chunk_length <= n_fft:
    RuntimeError: Argument #4: Padding size should be less than the corresponding input dimension, but got: padding (128, 128) at dimension 2 of input [1, 1, 128]

2. len(speech_to_process) == (n_fft // 2), (see my example above)

    RuntimeError: Argument #4: Padding size should be less than the corresponding input dimension, but got: padding (128, 128) at dimension 2 of input [1, 1, 128]
```
Based on @jhlee9010 last version (https://github.com/espnet/espnet/pull/4216/commits/98a65a943a0419de698b05e1f6e895a0e4c7b427) I added some further changes to address both of the above points. Basic idea is to keep filling the buffer until sufficient data samples have been accumulated (to avoid issue 1.) and padding the final chunk of data if necessary (to avoid issue 2.).

This revised version [asr_inference_streaming.zip](https://github.com/espnet/espnet/files/8490310/asr_inference_streaming.zip) was able to run with sim_chunk_lengths of 32, 64, 512, 1024 and even just 1 sample without errors obtaining nearly identical WER results with my Conformer setup (n_fft = 256, hoplength=128).

Hopefully, this could be a better solution than my initial fix and I would appreciate hearing your thoughts on it.

Thanks!

PS: Appologies for hijacking this thread, please let me know if I should open a new issue on this subject!

D-Keqi:
> @D-Keqi and @jhlee9010, I worked a bit further on the two (edge case) errors I am seeing for my setup when
> 
> ```
> 1. sim_chunk_length <= n_fft
> 
>     1.a) sim_chunk_length < (n_fft // 2):
>     IndexError: Dimension out of range (expected to be in range of [-64, 63], but got -128)
> 
>     1.b) (n_fft // 2) <= sim_chunk_length <= n_fft:
>     RuntimeError: Argument #4: Padding size should be less than the corresponding input dimension, but got: padding (128, 128) at dimension 2 of input [1, 1, 128]
> 
> 2. len(speech_to_process) == (n_fft // 2), (see my example above)
> 
>     RuntimeError: Argument #4: Padding size should be less than the corresponding input dimension, but got: padding (128, 128) at dimension 2 of input [1, 1, 128]
> ```
> 
> Based on @jhlee9010 last version ([98a65a9](https://github.com/espnet/espnet/commit/98a65a943a0419de698b05e1f6e895a0e4c7b427)) I added some further changes to address both of the above points. Basic idea is to keep filling the buffer until sufficient data samples have been accumulated (to avoid issue 1.) and padding the final chunk of data if necessary (to avoid issue 2.).
> 
> This revised version [asr_inference_streaming.zip](https://github.com/espnet/espnet/files/8490310/asr_inference_streaming.zip) was able to run with sim_chunk_lengths of 32, 64, 512, 1024 and even just 1 sample without errors obtaining nearly identical WER results with my Conformer setup (n_fft = 256, hoplength=128).
> 
> Hopefully, this could be a better solution than my initial fix and I would appreciate hearing your thoughts on it.
> 
> Thanks!
> 
> PS: Appologies for hijacking this thread, please let me know if I should open a new issue on this subject!

Thanks! It looks good. As for "obtaining nearly identical WER results", can you explain this in more detail？What results did you get respectively. @espnetUser

jhlee9010:
@espnetUser It seems nice. If you don't mind, I'd like to add your revision for another commit.

D-Keqi:
> @espnetUser It seems nice. If you don't mind, I'd like to add your revision for another commit. @D-Keqi Is there anything I have to do for those failed checks?

Maybe merging the latest ESPnet will be good. But before adding @espnetUser 's code, I hope we can make sure how much of a performance difference it makes. It will be very cool if you can explain it a little bit. @espnetUser

sw005320:
Thanks for the valuable discussion.
Can we also add a unit test to check @espnetUser's codes are working by providing a specific length (or random length) of input and checking that `feats.shape` is expected or not?

espnetUser:
Thanks all for your feedback. I will prepare a list of results for different values of sim_chunk_lengths and get back to you.

espnetUser:
As I can't share results and models for my own data set I decided to provide results for tedlium2 task which should also help with reproducability on your side. 

Using the default decoder config (conf/decode_asr_streaming.yaml) all tests were run from egs2/tedlium2/asr1/ folder with this call:
```
./run_streaming.sh --stage 12 --skip_data_prep true --skip_train true --download_model D-Keqi/espnet_asr_train_asr_streaming_transformer_raw_en_bpe500_sp_valid.acc.ave
```
across a range of values for sim_chunk_length and different versions v0, v1 and v2 of espnet2/bin/asr_inference_streaming.py file.

**Results:**
image

**Legend:**

v0 - current version: [asr_inference_streaming.v0.py.zip](https://github.com/espnet/espnet/files/8539610/asr_inference_streaming.v0.py.zip)
v1 - [jhlee9010](https://github.com/jhlee9010): [asr_inference_streaming.v1.py.zip](https://github.com/espnet/espnet/files/8539613/asr_inference_streaming.v1.py.zip)
v2 - [espnetUser](https://github.com/espnetUser): [asr_inference_streaming.v2.py.zip](https://github.com/espnet/espnet/files/8539616/asr_inference_streaming.v2.py.zip)

error1:
```IndexError: Dimension out of range (expected to be in range of [-1, 0], but got -384)```
error2:
```RuntimeError: start (128) + length (384) exceeds dimension size (256).```
error3:
```RuntimeError: Argument #4: Padding size should be less than the corresponding input dimension, but got: padding (256, 256) at dimension 2 of input [1, 1, 256]```

The last error type error3 I have seen only for v1 with my own data set for the rare edge case when:

```len(speech_to_process) == n_fft // 2``` 
and 
```hop_length == n_fft // 2```

Please find attached a modified version ([asr_inference_streaming.v1a.py.zip](https://github.com/espnet/espnet/files/8539703/asr_inference_streaming.v1a.py.zip)) of v1 that simulates this condition in case you would like to reproduce this behaviour. Perhaps it is possible to also add a unit test case for this edge case.

> As for "obtaining nearly identical WER results", can you explain this in more detail？

By that I meant I got identical results for the aggregated WER over all speakers for the different sim_chunk_length settings but on individual speaker level they are slightly different (for example check the attached v2 result logs for sim_chunk_length of 32 & 1024)
[score_wer_result.32.txt](https://github.com/espnet/espnet/files/8539795/score_wer_result.32.txt)
[score_wer_result.1024.txt](https://github.com/espnet/espnet/files/8539796/score_wer_result.1024.txt)

Hope that helps!

sw005320:
Super thanks, @espnetUser!

@jhlee9010, could I ask you to add a unit test to
https://github.com/espnet/espnet/blob/master/test/espnet2/bin/test_asr_inference.py ?
We may prepare some lengths of waveforms with a random vector, and pre-compute various values (e.g., `n_frames`, `n_residual`, `speech_to_process`, `waveform_buffer`, and the shape of `feats`) with several configurations (`n_fft`, `hop_length`, and `sim_chunk_length`), especially for the edge case that @espnetUser mentioned.

If you're not very sure, probably, @D-Keqi and I could add a unit test for you.

jhlee9010:
> Super thanks, @espnetUser!
> 
> @jhlee9010, could I ask you to add a unit test to https://github.com/espnet/espnet/blob/master/test/espnet2/bin/test_asr_inference.py ? We may prepare some lengths of waveforms with a random vector, and pre-compute various values (e.g., `n_frames`, `n_residual`, `speech_to_process`, `waveform_buffer`, and the shape of `feats`) with several configurations (`n_fft`, `hop_length`, and `sim_chunk_length`), especially for the edge case that @espnetUser mentioned.
> 
> If you're not very sure, probably, @D-Keqi and I could add a unit test for you.

@sw005320 I'm not pretty sure that how to create and test for multiple asr_config_file_streaming settings with pytest. I'd appreciate it if you add a unit test.
@espnetUser Thank you for all the tests. I've tested and created another commit with your revision.

PR #4227, opened 2022-04-01
fix issues about longformer package upgrades pytorch package

simpleoier:
Longformer depends on pytorch-lightning, which requires torch>=1.8.*. [link](https://github.com/PyTorchLightning/pytorch-lightning/blob/3e808b8a7002f13970c9d6f449fb2bb1042b9b98/requirements.txt#L4)

### DIFF
--- a/egs2/zh_openslr38/asr1/local/check_train_test_duplicate.py
+++ b/egs2/zh_openslr38/asr1/local/check_train_test_duplicate.py
@@ -32,7 +32,7 @@ for test_name in ("test", "dev"):
     print(count, "duplicates in", test_name)
 
     # if input("continue? [y/n]") == 'y':
-    # remove all instances of duplicate uttids in these files: spk2utt, text, utt2spk, wav.scp
+    # remove all instances of duplicate uttids in: spk2utt, text, utt2spk, wav.scp
     with open(f"data/{test_name}/spk2utt", "r") as f:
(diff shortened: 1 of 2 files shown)


setup.py
```diff
@@ -91,6 +91,7 @@
         "black",
     ],
     "doc": [
+        "Jinja<=3.0.3",
```
sw005320, line 94: We can leave the sphinx issue for now.
If the CI passes without the document build, we can merge this PR.

simpleoier: OK. Then I think this PR is almost ready. A few minutes to finish the rest tests. However, the sphinx issue needs some more time to get fixed. Or we can just hope sphinx could fix [this issue](https://github.com/sphinx-doc/sphinx/issues/10291) soon.

sw005320: I was expecting it would be fixed soon, but if it does not in the next week, I may ask you or @xinjli to do it.

sw005320:
Thanks, @simpleoier, for taking care of it.
Also, can you ask some existing PRs to merge this change and rerun the CI check?
Rerunning the CI check is very important (that was my mistake...)

simpleoier:
OK, I'll go through the existing PRs.

setup.py
xinjli: yeah, I can help with this if you want to fix it

PR #4250, opened 2022-04-10
Add option to use norm. feats over denorm.

G-Thor:
I added a parameter upon initialization of the Text2Speech object, which, when set to True will cause the vocoder model to receive normalized features rather than denormalized, when both are present in acoustic model output. The default value of this parameter is False so this should not cause any changes elsewhere.

### DIFF
--- a/espnet2/bin/tts_inference.py
+++ b/espnet2/bin/tts_inference.py
@@ -92,6 +92,7 @@ class Text2Speech:
         device: str = "cpu",
         seed: int = 777,
         always_fix_seed: bool = False,
+        prefer_normalized_feats: bool = False,
     ):
         """Initialize Text2Speech module."""
         assert check_argument_types()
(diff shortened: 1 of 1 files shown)


espnet2/bin/tts_inference.py
```diff
@@ -209,7 +211,9 @@ def __call__(
 
         # apply vocoder (mel-to-wav)
         if self.vocoder is not None:
-            if output_dict.get("feat_gen_denorm") is not None:
+            if self.prefer_normalized_feats and output_dict.get("feat_gen") is not None:
+                input_feat = output_dict["feat_gen"]
+            elif output_dict.get("feat_gen_denorm") is not None:
                 input_feat = output_dict["feat_gen_denorm"]
             else:
                 input_feat = output_dict["feat_gen"]
```
kan-bayashi, line 219: Simplified if-else block.
```suggestion
            if self.prefer_normalized_feats or output_dict.get("feat_gen_denorm") is None:
                input_feat = output_dict["feat_gen"]
            else:
                input_feat = output_dict["feat_gen_denorm"]
```

kan-bayashi (approved):
LGTM
Thank you for your contribution!

kan-bayashi:
I wil merge this PR since the CI errors are not related to this PR and these are fixed in master.

PR #4256, opened 2022-04-11
Import fast_bss_eval and update some time-domain losses for enh task

LiChenda:
In this PR, I import the [fast_bss_eval](https://github.com/fakufaku/fast_bss_eval) package, and replace the `si-snr` loss with it's more stable implementation.
Also, `sdr` loss is added with fast_bss_eval.


espnet2/enh/loss/criterions/time_domain.py
```diff
@@ -1,8 +1,12 @@
 from abc import ABC
+import logging
+
 
 import ci_sdr
```
Emrys365, line 5: I think we can remove CI-SDR now because it is equivalent to the added SDRLoss.

espnet2/enh/loss/criterions/time_domain.py
```diff
@@ -76,10 +80,76 @@ def forward(
...
+        ref: Tensor, (..., n_samples)
+            reference signal
+        est: Tensor (..., n_samples)
+            estimated signal
+
+        Returns
+        -------
+        loss: (...,)
+            the SDR loss (negative sdr)
+        """
```
Emrys365, line 121: It is better to follow Google's docstring convention as mentioned here: https://github.com/espnet/espnet/blob/master/CONTRIBUTING.md#41-python

espnet2/enh/loss/criterions/time_domain.py
```diff
@@ -76,10 +80,76 @@ def forward(
...
-    def __init__(self, eps=EPS):
+    """SI-SNR (SI-SDR) loss
+
+    A more stable SI-SNR loss with clamp from `fast_bss_eval`.
+    clamp_db: float
+        clamp the output value in  [-clamp_db, clamp_db]
+    zero_mean: bool
+        When set to True, the mean of all signals is subtracted prior.
+    eps: float
+        Deprecated. Keeped for compatibility.
```
Emrys365, line 144: Ditto

espnet2/enh/loss/criterions/time_domain.py
```diff
@@ -88,35 +158,39 @@ def name(self) -> str:
...
+
+        ref: Tensor, (..., n_samples)
+            reference signal
+        est: Tensor (..., n_samples)
+            estimated signal
+
+        Returns
+        -------
+        loss: (...,)
+            the SI-SDR loss (negative si-sdr)
```
Emrys365, line 173: Ditto

Emrys365 (approved):
LGTM in general.

Emrys365:
Hi @fakufaku, could you take a look at the CI error (https://github.com/espnet/espnet/runs/5985253746?check_suite_focus=true#step:8:416)?
I believe it is related the PyTorch compatibility issue in [fast_bss_eval](https://github.com/fakufaku/fast_bss_eval).

fakufaku:
@Emrys365 
```
>       X = torch.fft.rfft(x, n=n_fft, dim=-1)
E       AttributeError: 'builtin_function_or_method' object has no attribute 'rfft'
```
Yes, I think the `torch.fft` sub-module was added quite late to torch. I can check if I could fix this to work on older torch too.

fakufaku:
What is the oldest torch version you want to support ?

Emrys365:
> What is the oldest torch version you want to support ? The `torch.fft.rfft` is available since torch 1.7.0 I think.

Thanks! According to https://github.com/espnet/espnet#espnet-end-to-end-speech-processing-toolkit, currently espnet supports PyTorch 1.4.0 to 1.10.1.

fakufaku:
Ouch, ancient ;)

Is it allowed to restrict some parts of the code to run only on recent versions ?

Emrys365:
> Ouch, ancient ;)
> 
> Is it allowed to restrict some parts of the code to run only on recent versions ?

Yeah, 1.4.0 is a bit old. 😂

Do you mean restrict some code in espnet? We can do the PyTorch version check in the loss module. But I think it would be complicated to use two different implementations across different versions.

I did a quick search. How about importing rfft according to the PyTorch version?
```python
from distutils.version import LooseVersion
import torch

is_torch_1_7_plus = LooseVersion(torch.__version__) >= LooseVersion("1.7.0")

...


if is_torch_1_7_plus:
    import torch.fft.rfft as rfft
else:
    import torch.rfft as rfft
```

<details><summary>Below is the docstring I found for torch 1.3.0:</summary><div>

```
rfft(...)
    rfft(input, signal_ndim, normalized=False, onesided=True) -> Tensor

    Real-to-complex Discrete Fourier Transform

    This method computes the real-to-complex discrete Fourier transform. It is
    mathematically equivalent with :func:`~torch.fft` with differences only in
    formats of the input and output.

    This method supports 1D, 2D and 3D real-to-complex transforms, indicated
    by :attr:`signal_ndim`. :attr:`input` must be a tensor with at least
    ``signal_ndim`` dimensions with optionally arbitrary number of leading batch
    dimensions. If :attr:`normalized` is set to ``True``, this normalizes the result
    by dividing it with :math:`\sqrt{\prod_{i=1}^K N_i}` so that the operator is
    unitary, where :math:`N_i` is the size of signal dimension :math:`i`.

    The real-to-complex Fourier transform results follow conjugate symmetry:

    .. math::
        X[\omega_1, \dots, \omega_d] = X^*[N_1 - \omega_1, \dots, N_d - \omega_d],

    where the index arithmetic is computed modulus the size of the corresponding
    dimension, :math:`\ ^*` is the conjugate operator, and
    :math:`d` = :attr:`signal_ndim`. :attr:`onesided` flag controls whether to avoid
    redundancy in the output results. If set to ``True`` (default), the output will
    not be full complex result of shape :math:`(*, 2)`, where :math:`*` is the shape
    of :attr:`input`, but instead the last dimension will be halfed as of size
    :math:`\lfloor \frac{N_d}{2} \rfloor + 1`.

    The inverse of this function is :func:`~torch.irfft`.

    .. note::
        For CUDA tensors, an LRU cache is used for cuFFT plans to speed up
        repeatedly running FFT methods on tensors of same geometry with same
        configuration. See :ref:`cufft-plan-cache` for more details on how to
        monitor and control the cache.

    .. warning::
        For CPU tensors, this method is currently only available with MKL. Use
        :func:`torch.backends.mkl.is_available` to check if MKL is installed.

    Arguments:
        input (Tensor): the input tensor of at least :attr:`signal_ndim` dimensions
        signal_ndim (int): the number of dimensions in each signal.
            :attr:`signal_ndim` can only be 1, 2 or 3
        normalized (bool, optional): controls whether to return normalized results.
            Default: ``False``
        onesided (bool, optional): controls whether to return half of results to
            avoid redundancy. Default: ``True``

    Returns:
        Tensor: A tensor containing the real-to-complex Fourier transform result

    Example::

        >>> x = torch.randn(5, 5)
        >>> torch.rfft(x, 2).shape
        torch.Size([5, 3, 2])
        >>> torch.rfft(x, 2, onesided=False).shape
        torch.Size([5, 5, 2])
```

</div></details>

fakufaku:
The problem is that pytorch does not have good support for complex number in these versions, so a reimplementation of the freq-domain multiplication is needed.
`torch.rfft` returns a real tensor of size (..., 2) I think.

Emrys365:
> The problem is that pytorch does not have good support for complex number in these versions, so a reimplementation of the freq-domain multiplication is needed. `torch.rfft` returns a real tensor of size (..., 2) I think.

Oh I see. It would take a lot of efforts to handle the complex tensor issue.

BTW, I would like to know whether all loss functions have dependence on the native complex support. If only some of them rely on it, I think we can do the switch on espnet side.

espnet2/enh/loss/criterions/time_domain.py
LiChenda: Done.

LiChenda:
Thanks @Emrys365 and @fakufaku for the discussion, I just updated the docstring and added a version check in the espnet side.

fakufaku:
Hum, we also need to check for `torch.linalg.solve` that has only been made available recently...
```
 if use_cg_iter is None:
                # regular matrix solver
                R_mat = toeplitz(acf)
>               sol = torch.linalg.solve(R_mat, xcorr)
E               AttributeError: module 'torch.linalg' has no attribute 'solve'
```

LiChenda:
> Hum, we also need to check for `torch.linalg.solve` that has only been made available recently...
> 
> ```
>  if use_cg_iter is None:
>                 # regular matrix solver
>                 R_mat = toeplitz(acf)
> >               sol = torch.linalg.solve(R_mat, xcorr)
> E               AttributeError: module 'torch.linalg' has no attribute 'solve'
> ```

Thanks for pointing it out! I search the torch doc and they begin to support it since 1.8.0. Just updated the version check.

fakufaku:
@LiChenda I have patched `fast_bss_eval` (new version 0.1.3) to be backward compatible with torch down to 1.4. For 1.7 and below, it uses `torch_complex` for complex operations.

LiChenda:
> @LiChenda I have patched `fast_bss_eval` (new version 0.1.3) to be backward compatible with torch down to 1.4. For 1.7 and below, it uses `torch_complex` for complex operations.

Cool!  So I can remove the version check in the espnet side. Thank you @fakufaku !

sw005320:
Does this go through a test?

fakufaku:
On the fast_bss_eval side, the package is regression tested back to torch 1.4.

Emrys365:
It seems all clear!

sw005320:
Thanks, @LiChenda, @fakufaku, and @Emrys365!

PR #4265, opened 2022-04-11
update install_pesq.sh

LiChenda:
`tools/installers/install_pesq.sh` sometimes fails in the CI check due to the network issue.
So I host the source package in my github [repo](https://github.com/LiChenda/itu_pesq). 
Then the URL shoud be more stable.


sw005320:
Thanks, @LiChenda!

PR #4277, opened 2022-04-14
Fix aishell3 data preparation script

LanceaKing:


sw005320:
@LanceaKing, thanks a lot.
Could you tell me how this PR fixes the issue with an example?

ftshijt:
I can somehow understand the intention of this fix. If we set as `bool` in argparse, it would be set as true as long as the argument is presented during the function call (even if it is false). The external_g2p is set as True as default, so the external_g2p will always be set as True.

ftshijt (approved):
LGTM! Many thanks

LanceaKing:
> I can somehow understand the intention of this fix. If we set as `bool` in argparse, it would be set as true as long as the argument is presented during the function call (even if it is false). The external_g2p is set as True as default, so the external_g2p will always be set as True.

As @ftshijt said, the argument external_g2p in the previous code is always True unless passing an empty string. Also, the previous code would add spaces between Chinese characters, which is not correct.

PR #4295, opened 2022-04-20
Github Action Trigger Docker Build

Fhrozen:
Minor fix for trigger git action for docker. It should be launched 6 days ago, but it was skipped.
I assume that the issues is because comparing true with 'true'


sw005320:
Thanks, @Fhrozen!

Issue #4296, opened 2022-04-20
Two times sqrt in utterance_mvn.py


boeddeker:
https://github.com/espnet/espnet/blob/c7881c215e019324257620bc1dd83cec898ba6a9/espnet2/layers/utterance_mvn.py#L73-L80

I was checking what `utterance_mvn` does and recognized that `sqrt` was used two times instead of just once to convert the variance to the standard deviation in `espnet2`.

Emrys365:
Thanks @boeddeker, this is indeed a bug. I will fix it immediately.

PR #4304, opened 2022-04-21
Fix a bug in utterance_mvn

Emrys365:
This PR fix the bug mentioned in https://github.com/espnet/espnet/issues/4296.


sw005320:
Thanks, @Emrys365!

PR #4315, opened 2022-04-25
Harpervalley recipe

YushiUeda:
This PR adds harpervalley intent classification recipe as part of ESPnet-SLU.


ftshijt (approved):
Many thanks! The recipe looks perfect to me.

PR #4321, opened 2022-04-26
Add EnhPreprocessor for Speech Enhancement

Emrys365:
This PR adds a general preprocessor for adding noise and reverberation on the fly.

Below is an example usage of the newly added EnhPreprocessor in `egs2/wsj0_2mix/enh1`:

* run.sh
    > The added `--extra_wav_list` argument specifies the name of additional audio scp files to dump in stage 3, which will be used in EnhPreprocessor.
    > See [egs2/TEMPLATE/enh1/enh.sh#L359-L368](https://github.com/espnet/espnet/pull/4321/files#diff-5c188ea48af07a88f6ecde41a5edaf4ab4e356a168d40120d3f75d51f1179efeR359-R368) for more details.

    ```bash
    #!/usr/bin/env bash
    # Set bash to 'debug' mode, it will exit on :
    # -e 'error', -u 'undefined variable', -o ... 'error in pipeline', -x 'print commands',
    set -e
    set -u
    set -o pipefail

    min_or_max=min # "min" or "max". This is to determine how the mixtures are generated in local/data.sh.
    sample_rate=8k


    train_set="tr_${min_or_max}_${sample_rate}"
    valid_set="cv_${min_or_max}_${sample_rate}"
    test_sets="tt_${min_or_max}_${sample_rate} "

    ./enh.sh \
        --audio_format wav \
        --train_set "${train_set}" \
        --valid_set "${valid_set}" \
        --test_sets "${test_sets}" \
        --fs "${sample_rate}" \
        --lang en \
        --ngpu 1 \
        --use_preprocessor true \
        --extra_wav_list "rirs.scp noises.scp" \
        --local_data_opts "--sample_rate ${sample_rate} --min_or_max ${min_or_max}" \
        --enh_config conf/tuning/train_enh_dprnn_tasnet_with_preprocessor.yaml \
        "$@"
    ```

* conf/tuning/train_enh_dprnn_tasnet_with_preprocessor.yaml

    > <details><summary>expand</summary><div>
    >
    > ```yaml
    > optim: adam
    > init: xavier_uniform
    > max_epoch: 150
    > batch_type: folded
    > batch_size: 4
    > iterator_type: chunk
    > chunk_length: 32000
    > num_workers: 4
    > optim_conf:
    >     lr: 1.0e-03
    >     eps: 1.0e-08
    >     weight_decay: 0
    > patience: 4
    > val_scheduler_criterion:
    > - valid
    > - loss
    > best_model_criterion:
    > -   - valid
    >     - si_snr
    >     - max
    > -   - valid
    >     - loss
    >     - min
    > keep_nbest_models: 1
    > scheduler: reducelronplateau
    > scheduler_conf:
    >     mode: min
    >     factor: 0.7
    >     patience: 1
    > 
    > # preprocessor config
    > rir_scp: dump/raw/tr_min_8k/rirs.scp
    > rir_apply_prob: 1.0
    > noise_scp: dump/raw/tr_min_8k/noises.scp
    > noise_apply_prob: 1.0
    > noise_db_range: "0_30"
    > use_reverberant_ref: true
    > num_spk: 2
    > num_noise_type: 1
    > sample_rate: 8000
    > force_single_channel: true
    > 
    > encoder: conv
    > encoder_conf:
    >     channel: 64
    >     kernel_size: 2
    >     stride: 1
    > decoder: conv
    > decoder_conf:
    >     channel: 64
    >     kernel_size: 2
    >     stride: 1
    > separator: dprnn
    > separator_conf:
    >     num_spk: 2
    >     layer: 6
    >     rnn_type: lstm
    >     bidirectional: True  # this is for the inter-block rnn
    >     nonlinear: relu
    >     unit: 128
    >     segment_size: 250
    >     dropout: 0.1
    >     nonlinear: relu
    > 
    > # A list for criterions
    > # The overlall loss in the multi-task learning will be:
    > # loss = weight_1 * loss_1 + ... + weight_N * loss_N
    > # The default `weight` for each sub-loss is 1.0
    > criterions:
    >   # The first criterion
    >   - name: si_snr
    >     conf:
    >       eps: 1.0e-7
    >     wrapper: pit
    >     wrapper_conf:
    >       weight: 1.0
    >       independent_perm: True
    > ```
    >
    > </div></details>


Emrys365:
I modified the DataLoader initialization to set an epoch-dependent random seed for each worker, while the previous default Dataloader tends to use the same random seeds at each epoch.

sw005320:
@popcornell, could I ask you to review this PR?

sw005320:
@Emrys365, is it possible to make a unit test to cover this?
If it is difficult, we may consider having an integration test.

Emrys365:
I was thinking about it, but I could not find an example unit test for the preprocessor.
Maybe an integration test is easier, but it will need some noise and RIR signals for loading.

sw005320:
I think it is no problem to upload such files for the integration test (or create it with some files randomly and regard them as RIRs or noises).

popcornell:
@sw005320 @Emrys365  Looks like one of you need to invite me to do the review.

sw005320:
> @sw005320 @Emrys365 Looks like one of you need to invite me to do the review.

I think you can just go through the code and leave the comments.
Please let me know if it would not work.

popcornell:
If it is fine for you two, I'll do it this way then !

espnet2/tasks/enh.py
```diff
@@ -187,7 +189,74 @@ def add_task_arguments(cls, parser: argparse.ArgumentParser):
...
+            help="The file path of noise scp file.",
+        )
+        group.add_argument(
+            "--noise_apply_prob",
+            type=float,
+            default=1.0,
+            help="The probability applying Noise adding.",
+        )
+        group.add_argument(
+            "--noise_db_range",
```
popcornell, line 225: isn't it better to use SNR ? Because if speech material level varies a lot, then for some utterance you can be unlucky and end up with very low (or very high) SNR. If you sample instead the SNR you have more control.

espnet2/tasks/enh.py
```diff
@@ -187,7 +189,74 @@ def add_task_arguments(cls, parser: argparse.ArgumentParser):
             "--use_preprocessor",
             type=str2bool,
             default=False,
-            help="Apply preprocessing to data or not",
+            help="Whether to apply preprocessing to data or not",
+        )
+        group.add_argument(
+            "--speech_volume_normalize",
```
popcornell, line 195: Wouldn't be useful to add also an option for varying dynamically the range of level for speech ? 
In dynamic mixing you want often to do that too.

espnet2/train/preprocessor.py
```diff
@@ -489,3 +493,203 @@ def _text_process(
...
+        if isinstance(signal, tuple):
+            return tuple(self._ensure_2d(sig) for sig in signal)
+        elif isinstance(signal, list):
+            return [self._ensure_2d(sig) for sig in signal]
+        else:
+            # (Nmic, Time)
+            return signal[None, :] if signal.ndim == 1 else signal.T
+
+    def _get_early_signal(self, speech, rir):
+        predelay = 50  # milliseconds
```
popcornell, line 558: Correct way to deal with misalignment when you don't know the anechoic RIR ! I like this.

espnet2/tasks/enh.py
Emrys365: I think it is actually SNR (see [here](https://github.com/espnet/espnet/pull/4321/files#diff-bf4e9d668125c9efceb4d3a18b2f2df71d95f59596d0b442cb5aee2d9d270b05R260-R266)), although the name is a bit confusing.

Emrys365: I just kept the argument name as it was introduced in ASR preprocessors.

espnet2/tasks/enh.py
Emrys365: Do you mean randomly sampling from a range rather than a fixed value? It sounds reasonable to me.
We can just do that like for `noise_db_range`. @kamo-naoyuki What do you think?

espnet2/train/preprocessor.py
Emrys365: Oh yes. I forgot to do it! Thanks for reminding me of that.

espnet2/tasks/enh.py
popcornell: Ok thanks, makes sense ! maybe specify it in the help.

espnet2/tasks/enh.py
popcornell: Yes that would be good. There are many instances where you want to reduce for example the overall level of the speech utterances e.g. to simulate far field speech (correctly you re scale back after reverberation but sometime you want to simulate the low gain of speech captured by distant mics).

Emrys365:
> I changed ci tests to check the import order by [isort](https://github.com/PyCQA/isort) #4372, so please also apply isort after resolving the conflicts. Sorry for bothering you.

Sure, no problem.

espnet2/tasks/enh.py
Emrys365: I just added support of it (only for EnhPreprocessor).

Emrys365:
Hi @popcornell, could you review again? I made some changes to resolve all comments.

espnet2/tasks/enh.py
popcornell: Isn't it better to scale the energy in dB instead of the peak value ? As I see now

espnet2/train/preprocessor.py
```diff
@@ -203,85 +203,89 @@ def __init__(
...
+                    noise = f.read(dtype=np.float64, always_2d=True)
+                elif f.frames < nsamples:
+                    offset = np.random.randint(0, nsamples - f.frames)
+                    # noise: (Time, Nmic)
+                    noise = f.read(dtype=np.float64, always_2d=True)
+                    # Repeat noise
+                    noise = np.pad(
+                        noise,
+                        [(offset, nsamples - f.frames - offset), (0, 0)],
+                        mode="wrap",
```
popcornell, line 242: This is nice but could lead to funny noise clips if the noise is very short. Maybe raise a warning ? (but could be too verbose as it is in the dataloader).

Emrys365: Thanks for the comment. I have added support for it.

sw005320:
Thanks, @Emrys365!

PR #4243, opened 2022-04-07
Add streaming ST/SLU

D-Keqi:
This PR is to add streaming ST/SLU.

PR #4323, opened 2022-04-27
fix the enh_s2t_task argument in espnet2/bin/st_inference.py

simpleoier:
This is related to #4243


sw005320:
Thanks for the quick fix!

PR #4329, opened 2022-04-30
Wandb Minor Fix for Model Resume

roshansh-cmu:
To enable model resume within Weights and Biases without starting a new run, I updated the resume option to read from args.resume. 

https://docs.wandb.ai/guides/track/advanced/resuming


sw005320:
Thanks, @roshansh-cmu!

PR #4333, opened 2022-05-02
Added tamil SLU recipe

SujaySKumar:


sw005320:
@siddhu001, can you review this PR?

egs2/tamil/asr1/local/score.py
```diff
@@ -0,0 +1,103 @@
+#!/usr/bin/env bash
```
siddhu001, line 1: Hi @SujaySKumar , can you instead use the common scoring file https://github.com/espnet/espnet/blob/master/egs2/TEMPLATE/asr1/pyscripts/utils/score_intent.py?

siddhu001:
Thanks @SujaySKumar for making this PR. Please try to handle all my questions and suggestions.

sw005320:
@SujaySKumar, reminder.
Please follow @siddhu001’s suggestions

sw005320:
There are a bit too many config files.
If all have some results, it is fine (but add the results).
Otherwise, you can just keep a few important ones.

egs2/tamil/asr1/local/score.py
SujaySKumar: Done

sw005320:
- Please add a brief description of this recipe in https://github.com/espnet/espnet/blob/master/egs2/README.md
- The recipe name is not appropriate (too generic). Please come up with an appropriate name by following the other recipe name

SujaySKumar:
> * Please add a brief description of this recipe in https://github.com/espnet/espnet/blob/master/egs2/README.md
> * The recipe name is not appropriate (too generic). Please come up with an appropriate name by following the other recipe name

@sw005320 Will do!

SujaySKumar:
The `test_frontend_init` test seems to be failing with the following error

```
>   from fairseq.modules import EMAModule, EMAModuleConfig
[7830](https://github.com/espnet/espnet/runs/6549515166?check_suite_focus=true#step:8:7831)
E   ImportError: cannot import name 'EMAModule' from 'fairseq.modules'
```
 
I don't think it is related to the current recipe. Should we restart the CI again?

PR #4343, opened 2022-05-07
Fix a bug in stats aggregation when PITSolver is used

Emrys365:
This PR fixes the bug that stats cannot be correctly counted when permutation invariant training (PIT) is applied.


sw005320:
Thanks, @Emrys365!

PR #4344, opened 2022-05-08
Add Adapter Support to S3PRL front end

karthik19967829:
This PR adds support to Adapters to S3PRL front-end by referring to https://arxiv.org/pdf/2202.03218.pdf .


kamo-naoyuki:
I changed ci tests to check the import order by [isort](https://github.com/PyCQA/isort) #4372, so please also apply isort after resolving the conflicts. Sorry for bothering you.

karthik19967829:
@ftshijt please review this PR

karthik19967829:
@sw005320 Professor , I have cleaned up to some extent , there are few more shared files to be fixed , but it would be great if we can start discussion here on the design choice to make the below pipeline to be supported via config file 

Screen Shot 2022-05-23 at 8 09 53 AM

sw005320:
@karthik19967829, this PR is not ready for review. There are unnecessary files. You also changed unnecessary lines/files too much. You also have to fix the conflict. I strongly recommend you check our https://github.com/espnet/espnet/blob/master/CONTRIBUTING.md.

ftshijt:
Could you first clean the code for review? It makes it difficult to follow your changes as you are copying the `asr.sh` instead of use symlinks. Also, there are some temporary changes, which I'm not sure where to check.

ftshijt:
For your design question, I think the current framework is already capable to achieve the second one with some specific configurations. For example, setting ctc_weight to 1.0 would automatically skip the decoder part. Setting a lightweight linear encoder could form a CTC-decoder as the second work.

Depends on the case, but I don't think we need to refactor the whole framework or design a new one. Only adding some extra modules would be fine.

karthik19967829:
@ftshijt Also as we need to add Adapters to Hubert and WavLM . @jwrh is working on adding to s3prl toolkit ,and then we can just import it into ESPnet and add a recipe with above config

PR #4370, opened 2022-05-17
Update .mergify.yml

kamo-naoyuki:


kan-bayashi:
Timeout issue should be solved in #4369

PR #4381, opened 2022-05-20
Add support of test-only criterions after each epoch

Emrys365:
This PR enables specifying whether some criterions in the ENH task will only be calculated during inference (after each epoch).
This can be now achieved by adding an additional argument `only_for_test` in the model config file:
```yaml
criterions: 
  # The first criterion
  - name: si_snr 
    conf:
      eps: 1.0e-7
      only_for_test: True
    # the wrapper for the current criterion
    # for single-talker case, we simplely use fixed_order wrapper
    wrapper: fixed_order
    wrapper_conf:
      weight: 1.0
```


espnet2/enh/loss/criterions/abs_loss.py
```diff
@@ -13,6 +13,12 @@ class AbsEnhLoss(torch.nn.Module, ABC):
     def name(self) -> str:
         return NotImplementedError
 
+    # This property specifies whether the criterion will only
+    # be evaluated during the inference stage
+    @property
+    def only_for_test(self) -> bool:
+        return NotImplementedError
```
LiChenda, line 20: Given that most losses are used in both training and testing, maybe we can set defalut `False` here?

LiChenda (commented):
LGTM me!

sw005320:
Thanks a lot, @Emrys365!

PR #4382, opened 2022-05-20
Calculate RTF and latency in espnet2

espnetUser:
Make utils/calculate_rtf.py work with espnet2 inference scripts


espnetUser:
ESPNET2 streaming recipe: espnet/egs2/tedlium2/asr1
```
$ ./run_streaming.sh --stage 12 --skip_data_prep true --skip_train true --download_model D-Keqi/espnet_asr_train_asr_streaming_transformer_raw_en_bpe500_sp_valid.acc.ave
```
Output:
```
$ cat exp/D-Keqi/espnet_asr_train_asr_streaming_transformer_raw_en_bpe500_sp_valid.acc.ave/decode_asr_streaming_lm_lm_train_lm_en_bpe500_valid.loss.ave_asr_model_valid.acc.ave/test/logdir/calculate_rtf.log
Total audio duration: 155148.950 [sec]
Total decoding time: 249780.999 [sec]
RTF: 1.610
Latency: 1308.773 [ms/sentence]
```
ESPNET2 non-streaming recipe: espnet/egs2/librispeech/asr1:
```
$ ./run.sh --stage 12  --use_streaming false --skip_data_prep true --skip_train true --download_model byan/librispeech_asr_train_asr_conformer_raw_bpe_batch_bins30000000_accum_grad3_optim_conflr0.001_sp
```
Output:
```
$ cat exp/byan/librispeech_asr_train_asr_conformer_raw_bpe_batch_bins30000000_accum_grad3_optim_conflr0.001_sp/decode_asr_lm_lm_train_lm_transformer2_en_bpe5000_valid.loss.ave_asr_model_valid.acc.ave/test_clean/logdir/calculate_rtf.log
Total audio duration: 321903.500 [sec]
Total decoding time: 3783131.865 [sec]
RTF: 11.752
Latency: 87467.212 [ms/sentence]
```

espnetUser:
- added new time marker ("received final input") in asr_inference log in batch_beam_search_online.py for streaming mode 
- modified the latency calculation in calculate_rtf.py for streaming mode but not sure yet if that is correct way to measure latency this way (needs check!)
- validated changes with espnet2 (tedlium streaming and librispeech non-streaming) but not espnet1

espnet2/bin/asr_inference.py
```diff
@@ -470,13 +487,19 @@ def inference(
             assert len(keys) == _bs, f"{len(keys)} != {_bs}"
             batch = {k: v[0] for k, v in batch.items() if not k.endswith("_lengths")}
 
-            # N-best list of (text, token, token_int, hyp_object)
+            # log number of 10ms input frames needed for RTF calculations
```
sw005320, line 490: Why did you fix 10ms?
It can be changed depending on the hop size.

espnet2/bin/asr_inference_streaming.py
```diff
@@ -456,6 +473,10 @@ def inference(
             batch = {k: v[0] for k, v in batch.items() if not k.endswith("_lengths")}
             assert len(batch.keys()) == 1
 
+            # log number of 10ms input frames needed for RTF calculations
+            num_frames = int(len(batch["speech"]) / sample_rate * 1000 / 10)
```
sw005320, line 477: Ditto

sw005320 (commented):
Sorry for my late review.

sw005320:
Please also fix the CI errors in https://github.com/espnet/espnet/runs/6528144146?check_suite_focus=true#step:7:348

espnet2/bin/asr_inference.py
espnetUser: Yeah, this is not nice. I fixed the logging message to report the input lengths in terms of number of 10ms frames because I think the already existing calculate_rtf.py assumes a frame shift of 10ms. Let me explain below in more detail.

In https://github.com/espnet/espnet/blob/master/utils/calculate_rtf.py#L51 the audio duration in seconds is calculated like this:
```
audio_sec += sum(audio_durations) / 100  # [sec]
```
I understood that audio_durations is actually the accumulated number of input frames as retrieved from the logfile:
```
if "INFO: input lengths" in x:
   audio_durations += [int(x.split("input lengths: ")[1])]
```
So for the above conversion of audio_durations to audio_sec I figured this implicitly assumes a 10ms frame shift:
```
sum(audio_durations) * 10 / 1000  # [sec] = sum(audio_durations) / 100  # [sec]
```
Is my thinking correct or did I miss something here?

Maybe a better solution could be to log the correct number of input frames (based on model's hop_size) and supply a frame_shift/hop_size argument to calculate_rtf.py?

sw005320: > Is my thinking correct or did I miss something here?

You're right!

Yeah, espnet1 fixed to use 10ms while espnet2 can change it (and many recipes are also using 8ms, now)...

> Maybe a better solution could be to log the correct number of input frames (based on model's hop_size) and supply a frame_shift/hop_size argument to calculate_rtf.py?

Sounds good.

espnetUser:
@sw005320: Thanks for reviewing this. I thought a bit more about the streaming model and I think my first shot at measuring the latency in that case is not correct because:

- If the model is faster than real-time (RTF < 1) I think the encoder would need to wait some time before the next samples would be available for processing. That is currently not accounted for I think.
- For streaming case the new marker in logfile ("received final input") is logged after receiving the final chunk of data. But that chunk can have different lengths depending how close it is to end of file:
```
chunk_results = speech2text(
         speech[(i + 1) * sim_chunk_length : len(speech)], is_final=True                     
)
```
- So that biases the latency measured as the time between "end-of-speech" and "end-of-decoding".

If you have a good idea how to measure latency in streaming case please let me know!

sw005320:
I think this requires some discussions as there are several measures.
@eml914, do you have some suggestions?

From a repository management perspective, my suggestion is to provide an offline (non-streaming) RTF measure part, first within this PR. Then, after the discussion and fixing our strategy, we may have a follow-up PR to extend it to the streaming case.

espnetUser:
> From a repository management perspective, my suggestion is to provide an offline (non-streaming) RTF measure part, first within this PR. Then, after the discussion and fixing our strategy, we may have a follow-up PR to extend it to the streaming case.

Yes, makes sense. I felt the same way and already reverted back the changes for streaming mode and reworked the code for non-streaming RTF/Latency measurement. Will update the PR soon.

eml914:
Apporogies for my belated reply, and I agree with you that it's better first to do for offline and then for streaming later.
RTF does not matter because it is simply computational time per once second input.  But latency is a little bit tricky and I usually measure latencies with 10ms sleep in each input frames so that I can reproduce streaming input situation.  But it takes a lot of time to evaluate....

utils/calculate_rtf.py
```diff
@@ -20,6 +20,36 @@ def get_parser():
...
+        type=str,
+        default="decode",
+        choices=["decode", "asr_inference"],
+        help="name of logfile, e.g., 'decode' (espnet1) and "
+        "'asr_inference' (espnet2)",
+    )
+    parser.add_argument(
+        "--input-shift",
+        type=float,
+        default=10.0,
```
sw005320, line 34: It is good to set a default value, but it can be a source of some confusion.
I recommend adding a logging message that this script uses the 10ms shift as default when this option is not specified.

espnetUser:
As agreed this is a new version that handles just the non-streaming case:

- Added several options to script calculate_rtf.py to handle differences between espnet1 and espnet2:
```
usage: calculate_rtf.py [-h] [--log-dir LOG_DIR]
                        [--log-name {decode,asr_inference}]
                        [--input-shift INPUT_SHIFT]
                        [--start-times-marker {input lengths,speech length}]
                        [--end-times-marker {prediction,best hypo}]

calculate real time factor (RTF)

optional arguments:
  -h, --help            show this help message and exit
  --log-dir LOG_DIR     path to logging directory
  --log-name {decode,asr_inference}
                        name of logfile, e.g., 'decode' (espnet1) and
                        'asr_inference' (espnet2)
  --input-shift INPUT_SHIFT
                        shift of inputs in milliseconds
  --start-times-marker {input lengths,speech length}
                        String marking start of decoding in logfile, e.g.,
                        'input lengths' (espnet1) and 'speech length'
                        (espnet2)
  --end-times-marker {prediction,best hypo}
                        String marking end of decoding in logfile, e.g.,
                        'prediction' (espnet1) and 'best hypo' (espnet2)
```

- For espnet2 I found it easier to just log the number of speech samples and use the sample shift (1/fs) rather than log the number of feature frames. As the frame shift is configurable in espnet2 and there are different frontend configs/feature types it makes it a bit more difficult to find out the correct feature frame shift configuration used. However, this might be confusing and IMO it would be nicer to log the input in identical way for both espnet1 & 2. 

- For now the rtf & latency measurement is only run when "asr_inference_tool == "espnet2.bin.asr_inference". 
k2, maskctc and streaming modes are still left to do.
  
- Please note I also made a change in calculate_rtf.py as I think in its current form the script miscalculates the total audio duration: 

image

Without the above change I believe the next line will count audio durations of previous log files multiple times:
```
audio_sec += sum(audio_durations) * args.input_shift / 1000  # [sec]
```
because ```audio_durations``` is not reset after moving to next logfile.

Example output for librispeech test run:
```
./run.sh --stage 12  --use_streaming false --skip_data_prep true --skip_train true --download_model byan/librispeech_asr_train_asr_conformer_raw_bpe_batch_bins30000000_accum_grad3_optim_conflr0.001_sp
```
Output for test_clean subset:
```
# ../../../utils/calculate_rtf.py --log-dir exp/byan/librispeech_asr_train_asr_conformer_raw_bpe_batch_bins30000000_accum_grad3_optim_conflr0.001_sp/decode_as
r_lm_lm_train_lm_transformer2_en_bpe5000_valid.loss.ave_asr_model_valid.acc.ave/test_clean/logdir --log-name asr_inference --input-shift 0.0625 --start-times-
marker "speech length" --end-times-marker "best hypo"
Total audio duration: 19452.481 [sec]
Total decoding time: 137762.231 [sec]
RTF: 7.082
Latency: 52581.004 [ms/sentence]
```

Thanks!

utils/calculate_rtf.py
espnetUser: Thanks. I will add a logging message that clarifies about the frame shift and default setting.

sw005320:
Sorry for my late response.
Thanks!

sw005320:
I think this is a great function.
Can you add the usage in our documentation?
After the streaming ASR section? 
https://espnet.github.io/espnet/espnet2_tutorial.html#streaming-asr

espnetUser:
> I think this is a great function.
Can you add the usage in our documentation?
After the streaming ASR section?
https://espnet.github.io/espnet/espnet2_tutorial.html#streaming-asr

@sw005320: Yes, I will create a new Section "Real-Time-Factor and Latency" after streaming ASR section and add the usage + comments for how to use the script for non-streaming case (espnet1 & 2) to the documentation.

Sorry, I forgot to confirm with you before the merge if the ```../../../utils/calculate_rtf.py``` should be run automatically after each decoding call in Stage 12 in ```asr.sh``` as is currently done in this PR (see changes in ```asr.sh```) or if it should be better kept as a standalone script to be run as needed?

sw005320:
> @sw005320: Yes, I will create a new Section "Real-Time-Factor and Latency" after streaming ASR section and add the usage + comments for how to use the script for non-streaming case (espnet1 & 2) to the documentation.

Thanks, this is very cool.

> Sorry, I forgot to confirm with you before the merge if the `../../../utils/calculate_rtf.py` should be run automatically after each decoding call in Stage 12 in `asr.sh` as is currently done in this PR (see changes in `asr.sh`) or if it should be better kept as a standalone script to be run as needed?

Yes, attaching it to stage 12 is very helpful.
In this case, the script should be stable, and I also would like you to add a test script to https://github.com/espnet/espnet/tree/master/test_utils

espnetUser:
@sw005320: Please see https://github.com/espnet/espnet/pull/4560 for usage documentation and added test script.

PR #4399, opened 2022-05-27
fix s3prl install by using legacy version. Temporal solution.

simpleoier:
This is a temporal solution by using a legacy version. The new features like data2vec is not supported. I will test the compatibility of new fairseq and s3prl versions.


simpleoier:
@sw005320 This fix is ready to be merged.

sw005320:
OK!
Thanks for your quick fix!

Issue #4401, opened 2022-05-27
MFCC extraction in stage 2 of tts recipe fails with sox pipe in wav.scp


G-Thor:
When extracting MFCCs in the TTS recipe, you run into problems if you trim the durations of audio files with sox pipes as part of your preprocessing.
This happens since the MFCC extraction datadir is copied from the original `data/${dset}` in the referenced line, rather than `${data_feats}${_suf}/${dset}`, which references trimmed files (dumped) instead.

https://github.com/espnet/espnet/blob/7698194a242d4da9c87142addd3e59ac406cfa89/egs2/TEMPLATE/tts1/tts.sh#L361

Should be a quick and easy fix.

PR #4402, opened 2022-05-27
Fix for issue #4401

G-Thor:
This implements my suggested fix for issue #4401


kan-bayashi (approved):
Thank you for your suggestion!
I agree with your suggestion, it is better to use trimmed one.

PR #4364, opened 2022-05-13
add e2e tts model: JETS

imdanboy:
Hi, thanks for your great works; espnet

I experimented the e2e tts model (JETS) recently following 
paper: https://arxiv.org/abs/2203.16852
code: https://github.com/imdanboy/jets
and get some reasonable results on ljspeech, kss dataset.

Most part of the code is based on fastspeech2 and hifi-gan of espnet. It would be honor if it can be implemented in espnet.
I think an alignment module, related to an text-speech alignment learning, would be helpful at least to simplify the training of non-autoregressive tts model such as fastspeech.

Thanks.

PR #4406, opened 2022-05-29
add pre-trained model jets in the recipe of ljspeech, kss

imdanboy:
Hi, I pre-trained jets recently and updated the recipe README following #4364 
Thanks :)


kan-bayashi (approved):
Cool! Thank you for uploading the pretrained model!

PR #4415, opened 2022-06-02
Zero_infinity option for ctc loss

kamo-naoyuki:
When I implemented `ignore_nan_grad` for  ctc modules of espnet2, pytorch-ctc didn't have the zero_infinity option.

 `ignore_nan_grad` is not so efficient and looks complicated impelemetation, so I changed it to use  the `zero_infinity` option. (ESPnet1 already used it.)

Note that this CTC class has both  `ignore_nan_grad` and `zero_infinity`  options and `ignore_nan_grad`  is an alias of `zero_infinity`. I left  `ignore_nan_grad` for backward compatiblity.


sw005320 (approved):
LGTM.
Thanks, @kamo-naoyuki

Issue #4410, opened 2022-05-31
tensorboard graphs


nellorebhanuteja:
In the current setup when I type 
```
tensorboard --logdir <log_dir>
```

i can visualize model training progress.
Can i also generate tensorboard graphs?
It will help me understand the architecture of underlying model

kamo-naoyuki:
No, it's not supported. I'll consider it.

nellorebhanuteja:
@kamo-naoyuki 

thanks for the update
I am trying to run tacotron2 training on LJSpeech
I am still not able to see graphs

kamo-naoyuki:
Tensorboard graph is not always added because it depends on torch script.  If it's failed, there are no ways to see it.

> Also the training has failed

Oh... I'll check it.

nellorebhanuteja:
I also noticed the GPU usage is going to 100 % and stagnating there

kamo-naoyuki:
> Also the training has failed


I fixed it.

kamo-naoyuki:
Once again, tensorboard graph is not always added. If it's failed, it is impossible to add the graph.

PR #4418, opened 2022-06-03
Support tensorboard graph

kamo-naoyuki:
#4410

I changed espnet2-trainer to add the tensorboard graph of the model.

- The graph is added at the first iteration for each epoch in the training.
- Internally, The tensorboard graph depends on `torch script`, thus the graph
creation is failed in some situations.
   - If the graph creation is failed, the related errors are ignored and training goes ahead.
  
@kan-bayashi I'm not sure this change doesn't have any side effects. Could you also check?


espnet2/train/trainer.py
```diff
@@ -513,6 +514,24 @@ def train_one_epoch(
                 all_steps_are_invalid = False
                 continue
 
+            if iiter == 1 and summary_writer is not None:
```
kan-bayashi, line 517: This may try to add graph in every epoch, right?
If so, is it needed?

kamo-naoyuki: Yes.

- If the graph is added only when 0epoch, the graph is not added when resuming from middle epochs.
- In this method, we can't judge whether it is the first epoch when resuming. We need to modify trainer.py a little bit.
- I think the graph creation is acceptable if it is only done once for each epoch.

kan-bayashi: OK. I understand your point.

kan-bayashi (approved):
LGTM

espnet2/train/trainer.py
kamo-naoyuki: If you found it time-consuming after merging, please tell me it.

PR #4427, opened 2022-06-06
Fix an4 URL

pyf98:
Hi, this PR adds a backup URL for AN4.

If the original website doesn't respond, the following URL should work:
```bash
wget https://huggingface.co/datasets/espnet/an4/resolve/main/an4_sphere.tar.gz
```


sw005320:
Thanks, @pyf98!

Issue #4377, opened 2022-05-19
Multi-GPU performance improvement for ASR in ESPnet1 by DDP


lazykyama:
**Describe the bug**

Dear ESPnet team,

This issue is a proposal to improve the multi-GPU training performance for **v1** ASR.
I'm filing this because I didn't find any other issues and/or PRs except for https://github.com/espnet/espnet/issues/3583 as far as I checked. (Sorry in advance if I overlooked any other similar issues/PRs and/or should discuss on https://github.com/espnet/espnet/issues/3583.)
I have already made a basic DDP implementation for a verification as mentioned later, and can send a PR.
Here, I would like to discuss with you if this feature is required at first.
After that, also would like to talk about if the current modifications are acceptable, etc.

As discussed in https://github.com/espnet/espnet/issues/3583, it looks like ASR training script in v1 has a large bottleneck in multi-GPU case.
As far as I investigated, it's caused by thread-based multi-GPU implementation.
According to a profiling result by Nsight Systems, GPU computation occasionally stopped even though actual training codes don't contain any blockers, and GPU has enough capability to continue calculation.
Also, during some GPUs stopped, it looks like only one GPU could run simultaneously, and the GPU that could run was determined in a round-robin fashion.
Therefore, I guess that this is caused by GIL.

To verify if my guess is correct, I made several changes for ASR v1 training codes to support PyTorch's DDP.
As a result, 8GPU case can achieve x2 faster compared to non-DDP version (=current implementation).
Performance data is below.
Note that 8GPU case includes several times validations because 2500 iters mean around 10 epochs in my setting, and only `train-clean-100` dataset is used for this measurement to avoid too long training time.

| num of GPUs | num of iters | current impl [sec.] | DDP [sec.] | 
|:--:|:--:|:--:|:--:|
| 1 | 2000 | 1270.37 | 1299.95 | 
| 8 | 2500 | 5327.21 | 2470.36 | 

And, throughput roughly calculated from the result above is below. (bs=32 is assumed.)

| num of GPUs | current impl [samples / sec.] | DDP [samples / sec.] | DDP vs current impl |
|:--:|:--:|:--:|:--:|
| 1 | 201.5160937 | 196.9306512 | x0.98 |
| 8 | 480.5517335 | 1036.286209 | x2.16 |

**Basic environments:**

I did an early investigation and PoC implementation on DGX A100 (80GB memory version) and within a NGC's PyTorch docker container.
Detailed information is below.

[...]

PR #4430, opened 2022-06-07
Add DDP support for v1 ASR training.

lazykyama:
This PR is related to #4377. As I described at a comment on the issue, this PR includes:

- adding a new script to launch multiple processes on each GPU like `pytorch's torch.distributed.launch`,
- adding a new option, `--use-ddp`, to `asr_train.py` to switch DDP and standard `data_parallel` when specifying multiple GPUs,
- changing evaluation schemes to distributed version,
- replacing a way to split dataset with pytorch's `DistributedSampler`,
- suppressing saving a checkpoint, intermediate info, plots, etc from non-main processes.

Please let me know if we need to discuss more about the changes.

### DIFF
--- a/espnet/distributed/pytorch_backend/launch.py
+++ b/espnet/distributed/pytorch_backend/launch.py
@@ -1,0 +1,167 @@
+#
+# SPDX-FileCopyrightText: Copyright (c) 2022 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+#
+# This code uses an official implementation of
+# distributed data parallel launcher as just a reference.
+# https://github.com/pytorch/pytorch/blob/v1.8.2/torch/distributed/launch.py
+#
+# One main difference is this code focuses on
+# launching simple function with given arguments.
+#
+
+
+import multiprocessing
+import os
+import signal
+import time
+
+
+_signalno_name_map = {
+    s.value: s.name for s in signal.valid_signals()
+    if isinstance(s, signal.Signals)
+}
+
+
+class WorkerError(multiprocessing.ProcessError):
+    def __init__(self, *, msg, exitcode, worker_id):
+        super(WorkerError, self).__init__(msg)
+        self._exitcode = exitcode
+        self._worker_id = worker_id
+
+    def __str__(self):
+        return (
+            f"worker[{self._worker_id}] failed with "
+            f"exitcode={self._exitcode}"
+        )
+
+    @property
+    def exitcode(self):
+        return self._exitcode
+
+    @property
+    def worker_id(self):
+        return self._worker_id
+
+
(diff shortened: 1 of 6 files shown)


espnet/distributed/pytorch_backend/launch.py
```diff
@@ -0,0 +1,167 @@
...
+            else:
+                p.kill()
+        except:  # noqa: E722
+            # NOTE: Ignore any exception happens during killing a process
+            # because this intends to send kill signal to *all* processes.
+            pass
+
+
+def launch(func, args, nprocs, master_addr="localhost", master_port=29500):
+    """
```
kamo-naoyuki, line 90: Is the port number 29500 a hard coding? 

In espnet2, we select a free port number automatically.

https://github.com/espnet/espnet/blob/43ce0c69fb32961235534b348700dc6c74ad5792/espnet2/train/distributed_utils.py#L187-L198

kamo-naoyuki (approved):
@sw005320 I understood this implementation and looks okay without my comment. However, I'm not sure this code works without problems. I recommend you running some asr recipes to check it.

espnet/distributed/pytorch_backend/launch.py
lazykyama: @kamo-naoyuki Thank you for the reviewing. The reason why I hard coded a port number here is that the original pytorch distributed also uses a hard coded value, and I thought library should have a default value. But, as you mentioned, a mechanism which automatically finds a free port is better for users. Let me update my PR.

lazykyama: @kamo-naoyuki Sorry for my late action. I have updated this hard coded value. Default value is changed to `None` and it will be filled with automatically detected port number at the beginning of this function if `master_port` is not specified. Could you please check again if this modification makes sense?

kamo-naoyuki: LGTM, thanks

kamo-naoyuki:
Could you fix the errors of our CI? Please apply black to the code.

lazykyama:
@kamo-naoyuki  Ok, I'm applying the formatter to the codes and will push these changes. BTW, as far as I checked, it looks like several files I haven't touched were reformatted by black. Should I include such changes too? Or, I must include only what I changed?

I simply ran a command below.

```
cd /path/to/workdir/espnet/
black espnet espnet2 test utils setup.py egs*/*/*/local egs2/TEMPLATE/*/pyscripts tools/*.py ci/*.py
```

kamo-naoyuki:
>Or, I must include only what I changed?

Yes, please.

kamo-naoyuki:
@sw005320 Please merge this PR if it's good for you

sw005320:
Thanks, @lazykyama and @kamo-naoyuki!

PR #4436, opened 2022-06-08
Add notes on batch size and num of GPUs in ESPnet2 documentation

pyf98:
We have noticed that many users encountered this issue when using multiple GPUs on ESPnet2. Based on various investigations, ESPnet2 actually has good multi-GPU scalability, if the batch size is properly set.

Note that the behavior of batch size in ESPnet2 during multi-GPU training is different from that in ESPnet1. **In ESPnet2, the total batch size is not changed regardless of the number of GPUs.** Therefore, you need to manually increase the batch size if you increase the number of GPUs. Please refer to this [doc](https://espnet.github.io/espnet/espnet2_training_option.html#the-relation-between-mini-batch-size-and-number-of-gpus) for more information.

### DIFF
--- a/doc/espnet2_distributed.md
+++ b/doc/espnet2_distributed.md
@@ -14,6 +14,9 @@ ESPnet2 provides some kinds of data-parallel distributed training.
 
 ## Examples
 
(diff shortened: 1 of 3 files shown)


sw005320:
Thanks, @pyf98!
I believe this will mitigate the confusion regarding the multi-GPU usage!

PR #4441, opened 2022-06-13
Update language codes in AmericasNLP 2022 baseline

akreal:
This is to match the codes used last year, hopefully that's fine.


sw005320:
Thanks, @akreal!

PR #4445, opened 2022-06-14
Espnet multimodal PR

chorongi:
Made fix in installers and change in AVSR model for convergence / unimodal usage.


sw005320:
@chorongi, it seems that your change in the Makefile (`tools/Makefile`) causes the issue, especially for the numpy related.
Please check it.
Also, move the multimodal tool installation to the extra part in Makefile.

kamo-naoyuki:
I ran black with cloning your PR on my local machine, and it reformated the codes, just like the error message on the CI tests. 

Please recheck your black version and run black with --verbose option to check not to use any configuration files for black.

```
black --version
black --verbose ...
```

chorongi:
@kamo-naoyuki  Thank you so much for the help! I think there was an issue with my format script using black. 
I really appreciate your help. Have a wonderful day!!

karthik19967829:
@chorongi can you add the TODO tasks here I will take care of this PR from here ?

Issue #4439, opened 2022-06-11
[HELP] JETS Finetuning


seastar105:
Hi, I'm trying to finetune JETS model trained on kss.

after, 200epochs model makes quite audible sound now, but attention plots are screwed up. I'm confused. is it trained well..?

![image](https://user-images.githubusercontent.com/30820469/173175102-d7bc3b39-3a59-4fb2-a028-44309294d126.png)

![image](https://user-images.githubusercontent.com/30820469/173175123-0f231702-d9a1-40d4-9638-ae26c3f4338f.png)

Here's finetuned model [link](https://drive.google.com/file/d/1uJogjx_w4ub3qeJ9mPtNDBKwDwbnrxnp/view?usp=sharing).

and, wandb run [link](https://wandb.ai/seastar105/JETS-finetune-low-lr-long-epoch/runs/2jkkvf2q?workspace=user-seastar105) (from 97 epoch to 200 epoch)

@imdanboy

imdanboy:
Your attention plot seems to be came from self-attention not alignment attention. So i am not sure how self 
-attention plot looks like if it is trained well.

Instead, did you expect diagonal alignment attention plot between text and mel-spectrogram?
If so, plotting alignment attention plot yourself and checking its diagonality may be helpful.
Alignment attention matrix can be obtained from `torch.exp(log_p_attn)` at
https://github.com/espnet/espnet/blob/master/espnet2/gan_tts/jets/generator.py#L580

seastar105:
Thank you. I'll check torch.exp(log_p_attn)

seastar105:
![image](https://user-images.githubusercontent.com/30820469/174423447-a984c400-a1db-44c0-8131-7f04ddedd51a.png)

I could find text and mel-spectrogram attention plots have diagonality, thank you @imdanboy

PR #4446, opened 2022-06-14
Add JETS AlignmentModule in calculate_all_attentions.py

seastar105:
JETS model has alignment module but has no attention plot now. See Issue #4439 

So, added AlignmentModule in attention

below is attention plot with ljspeech recipe. I ran recipe only one step with pretrained model for attention plots. also, you can check more plots [here](https://wandb.ai/seastar105/jet_ljspeech_plot/runs/2xapg9t1?workspace=user-seastar105)

image


kan-bayashi (commented):
This is really useful! After fixing CI, I will merge it.

Issue #4465, opened 2022-06-24
ModuleNotFoundError: No module named 'espnet2.diar.layers' in Colab SE demo


cycentum:
When running the Colab notebook of SE demo, I encountered the error 

> ModuleNotFoundError: No module named 'espnet2.diar.layers'

at “Enhance your own pre-recordings” cell.

I found the error is caused by espnet/espnet2/diar and espnet/espnet2/layer not properly installed.

In my local environment, I could fix the error by adding __init__.py in these directories.

Can I open a pull request to fix this bug?

**To Reproduce**
Run cells sequentially from the top in ESPnet SSE Demonstration.ipynb

Emrys365:
Yes, I can reproduce this bug. @cycentum It would be great if you could make a PR to fix it!

PR #4470, opened 2022-06-25
Adding __init__.py to espnet2/diar/layers and espnet2/diar/separator

cycentum:
Because espnet2/diar/layers and espnet2/diar/separator did not contain __init__.py, py files in these directories are not properly installed via pip.
In this commit I added __init__.py to these directories.

This commit should fix the bug reported in the issue _ModuleNotFoundError: No module named 'espnet2.diar.layers' in Colab SE demo_ https://github.com/espnet/espnet/issues/4465


Emrys365:
LGTM. Thanks @cycentum !

Could you format the added files with `black` to resolve the CI test errors?

PR #4476, opened 2022-06-28
Add support for calculating losses on noise and dereverberated signals

Emrys365:
This PR restores the support for calculating losses on noise and dereverberated signals.
The function has bee removed since the commit https://github.com/espnet/espnet/commit/2caec34f979b54410c9c9785812ecdf3d09ed6b8 due to the major refactorization of [espnet2/enh/espnet_model.py](https://github.com/espnet/espnet/blob/master/espnet2/enh/espnet_model.py).
This function is now re-implemented by adding two new attributes to the criterion module: `is_noise_loss` and `is_dereverb_loss`. They can also be easily configured in the yaml file.

Also, most speech enhancement/separation models provided in [espnet2/enh/separator/](https://github.com/espnet/espnet/blob/master/espnet2/enh/separator) have been modified to support predicting the noise signal in addition to the speech signals, i.e., the attribute `predict_noise` is added as an indicator.

### DIFF
--- a/egs2/TEMPLATE/enh1/README.md
+++ b/egs2/TEMPLATE/enh1/README.md
@@ -41,7 +41,7 @@ In `egs2/TEMPLATE/enh1/enh.sh`, 13 stages are included.
 This stage is similar to stage 1 in [asr.sh](https://github.com/espnet/espnet/blob/master/egs2/TEMPLATE/asr1/asr.sh).
 
 #### Stage 2: Speech perturbation
-Speech perturbation is widely used in ASR task, but rarely in speech enhancement. Some of our initial experiments have shown that speech perturbation works on `wsj0_2mix`. We are conducting more experiments to make sure if it works.
+Speech perturbation is widely used in the ASR task, but rarely in speech enhancement. Some of our initial experiments have shown that speech perturbation works on `wsj0_2mix`. We are conducting more experiments to make sure if it works.
 The speech perturbation procedure is almost the same as that in ASR, we have copied `scripts/utils/perturb_data_dir_speed.sh` to `scripts/utils/perturb_enh_data_dir_speed.sh` and made some minor modifications to support the speech perturbation for more scp files rather than `wav.scp` only.
 
 #### Stage 3: Format wav.scp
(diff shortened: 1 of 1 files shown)


Emrys365:
@LiChenda Could you take a look at the changes?

LiChenda:
> @LiChenda Could you take a look at the changes?

Sure!

espnet2/enh/loss/criterions/time_domain.py
```diff
@@ -20,10 +20,37 @@ def name(self) -> str:
...
+        is_dereverb_loss=False,
+    ):
         super().__init__()
-        self._name = name
+        # only used during validation
         self._only_for_test = only_for_test
+        # only used to calculate the noise-related loss
+        self._is_noise_loss = is_noise_loss
+        # only used to calculate the dereverberation-related loss
+        self._is_dereverb_loss = is_dereverb_loss
```
LiChenda, line 44: I saw you added `only_for_test`, `is_noise_loss ` and `is_dereverb_loss ` for both time-domain loss and frequency-domain loss.  Adding these attributes to `AbsEnhLoss(torch.nn.Module, ABC)`  might be a more brief way. The current approach is good. It's just a suggestion.

espnet2/enh/separator/dc_crn_separator.py
```diff
@@ -84,11 +86,13 @@ def __init__(
         super().__init__()
 
         self._num_spk = num_spk
+        self.predict_noise = predict_noise
         self.mode = mode
         if mode not in ("mapping", "masking"):
             raise ValueError("mode=%s is not supported" % mode)
         self.ref_channel = ref_channel
 
+        num_outputs = self.num_spk + 1 if self.predict_noise else self.num_spk
```
LiChenda, line 95: Given that all separators has the new attributes `num_outputs ` and `predict_noise `, maybe we should consider adding it to the super class `AbsSeparator`.

LiChenda (approved):
LGTM! Great Job! I just post some small suggestions!

espnet2/enh/separator/dc_crn_separator.py
Emrys365: Not all models can easily support `predict_noise`, such as DPCL and DANet. So I prefer to add support for this function only when it needs minor modifications. Of course, we would recommend newly implemented models to try to support this function.

espnet2/enh/loss/criterions/time_domain.py
Emrys365: My initial thought is that some losses may not be designed for noise signals, like the DPCL loss. So I didn't add these properties to `AbsEnhLoss` directly.
If you think it is better to inherit from `AbsEnhLoss`, I would do that.

sw005320:
Thanks a lot!

PR #4479, opened 2022-06-29
Offline/Online (standalone) ESPnet2 Transducer

b-flo:
Hi,

This PR is a re-do of #4032 with streaming capabilities based on WeNet [chunk-by-chunk approaches](https://arxiv.org/pdf/2012.05481.pdf) and Icefall implementations.

The custom encoder architecture was kept here but limited to conv1d and conformer blocks. The idea is to support other *-former architecture (branchformer, enformer, k2-conformer, longformer, etc) as blocks to make a custom X-former architecture for offline and streaming ASR. I implemented and tested most of them already but it'll be added in next PRs.

In regards to the reviews in #4032 and change requests:
 - Naming: There may be minor differences but it should be consistent with other models now!
 - Duplication: We should discuss each duplicated module individually in this version. Most of them were duplicated in preparation to future additions but a few may just be for my convenience and may be removed/merged.

**NOTE**: Everything should work but this PR is a rebase of previous PR with stitched elements from different work branches, it may contains bugs or mistakes. Feel free to correct or point-out any suspicious parts please!

**TO DO:**

- [x] Add tests for online ASR (dummy ones for this PR)
- [x] Refine docs
- [x] Add missing references

@csukuangfj I would be glad if you or other Icefall members could take a look at the PR!
Also, if you could point-out any missing references to your work/implementation, it would be great! Because we've gone full circle (ESPnet -> WeNet -> Icefall -> ESPnet ...) on some parts, I'm a bit confused on the proper references...


espnet2/asr_transducer/decoder/stateless_decoder.py
```diff
@@ -0,0 +1,145 @@
...
+        self.blank_id = embed_pad
+
+        self.device = next(self.parameters()).device
+        self.score_cache = {}
+
+    def forward(
+        self,
+        labels: torch.Tensor,
+        states: Optional[Tuple[torch.Tensor, Optional[torch.Tensor]]] = None,
+    ) -> Tuple[torch.Tensor, None]:
```
csukuangfj, line 49: The type hint for the return value does not match the actual return value and hence it is misleading.

espnet2/asr_transducer/decoder/stateless_decoder.py
```diff
@@ -0,0 +1,145 @@
...
+        """Encode source label sequences.
+
+        Args:
+            labels: Label ID sequences. (B, L)
+            states: Decoder hidden states. None
+
+        Returns:
+            dec_out: Decoder output sequences. (B, U, D_dec)
+            states: Decoder hidden states. None
```
csukuangfj, line 58: It returns only a single value.

espnet2/asr_transducer/decoder/stateless_decoder.py
```diff
@@ -0,0 +1,145 @@
...
+
+        Returns:
+            dec_out: Decoder output sequence. (1, D_emb)
+            state: Decoder hidden states. None
+
+        """
+        str_labels = "_".join(map(str, label_sequence))
+
+        if str_labels in self.score_cache:
+            dec_embed, state = self.score_cache[str_labels]
```
csukuangfj, line 86: Since `state` is always `None`, I think there is no need to store it into the cache.

csukuangfj (changes requested):

espnet2/asr_transducer/decoder/stateless_decoder.py
b-flo: What a timing, I was fixing a bunch of docstrings and type hints! Sorry for that!
Thanks!

espnet2/asr_transducer/decoder/stateless_decoder.py
b-flo: Right! We do not need to ensure cache values are Tuple in this version! Thanks!

espnet2/asr_transducer/decoder/stateless_decoder.py
```diff
@@ -0,0 +1,144 @@
...
+        else:
+            dec_embed = self.embed(label)
+
+            self.score_cache[str_labels] = dec_embed
+
+        return dec_embed[0], None
+
+    def batch_score(
+        self,
+        hyps: Union[List[Hypothesis], List[ExtendedHypothesis]],
```
csukuangfj, line 95: `ExtendedHypothesis` is a subclass of `Hypothesis`, so it is safe to use
```python3
hyps: List[Hypothesis]
```
here. No need to use 
`Union[List[Hypothesis], List[ExtendedHypothesis]]`

espnet2/asr_transducer/decoder/stateless_decoder.py
```diff
@@ -0,0 +1,144 @@
...
+        self.device = device
+
+    def init_state(self, batch_size: int) -> None:
+        """Initialize decoder states.
+
+        Args:
+            batch_size: Batch size.
+
+        Returns:
+            : Initial decoder hidden states. None
```
csukuangfj, line 128:
```suggestion
            Initial decoder hidden states. None
```

espnet2/asr_transducer/decoder/rnn_decoder.py
```diff
@@ -0,0 +1,262 @@
...
+        hidden_size: int = 256,
+        rnn_type: str = "lstm",
+        num_layers: int = 1,
+        dropout_rate: float = 0.0,
+        embed_dropout_rate: float = 0.0,
+        embed_pad: int = 0,
+    ) -> None:
+        assert check_argument_types()
+
+        if rnn_type not in {"lstm", "gru"}:
```
csukuangfj, line 40:
```suggestion
        if rnn_type not in ("lstm", "gru"):
```
A tuple may be more efficient than a set here.

espnet2/asr_transducer/decoder/rnn_decoder.py
```diff
@@ -0,0 +1,262 @@
...
+
+    def forward(
+        self,
+        labels: torch.Tensor,
+        states: Tuple[torch.Tensor, Optional[torch.Tensor]] = None,
+    ) -> torch.Tensor:
+        """Encode source label sequences.
+
+        Args:
+            labels: Label ID sequences. (B, L)
```
csukuangfj, line 80: It has two inputs, but only one is documented.

espnet2/asr_transducer/decoder/rnn_decoder.py
```diff
@@ -0,0 +1,262 @@
...
+
+        self.blank_id = embed_pad
+
+        self.device = next(self.parameters()).device
+        self.score_cache = {}
+
+    def forward(
+        self,
+        labels: torch.Tensor,
+        states: Tuple[torch.Tensor, Optional[torch.Tensor]] = None,
```
csukuangfj, line 75: If it can be `None`, please use `Optional[]` here.

espnet2/asr_transducer/decoder/rnn_decoder.py
```diff
@@ -0,0 +1,262 @@
...
+        """One-step forward hypothesis.
+
+        Args:
+            label: Previous label. (1, 1)
+            label_sequence: Current label sequence.
+            dec_state: Previous decoder hidden states.
+                         ((N, 1, D_dec), (N, 1, D_dec) or None)
+
+        Returns:
+            dec_out: Decoder output sequence. (1, D_dec) or None
```
csukuangfj, line 143: In which case will `dec_out` be `None`?

espnet2/asr_transducer/decoder/stateless_decoder.py
csukuangfj: ah, ok.

espnet2/asr_transducer/decoder/rnn_decoder.py
b-flo: None in this version! Thanks

espnet2/asr_transducer/decoder/rnn_decoder.py
```diff
@@ -0,0 +1,264 @@
...
+
+        """
+        return (
+            states[0][:, idx : idx + 1, :],
+            states[1][:, idx : idx + 1, :] if self.dtype == "lstm" else None,
+        )
+
+    def create_batch_states(
+        self,
+        states: Tuple[torch.Tensor, Optional[torch.Tensor]],
```
csukuangfj, line 246: Looks like this input is never used.

espnet2/asr_transducer/beam_search_transducer.py
```diff
@@ -0,0 +1,711 @@
...
+
+class BeamSearchTransducer:
+    """Beam search implementation for Transducer."""
+
+    def __init__(
+        self,
+        decoder: AbsDecoder,
+        joint_network: JointNetwork,
+        beam_size: int,
+        lm: torch.nn.Module = None,
```
csukuangfj, line 55:
```suggestion
        lm: Optional[torch.nn.Module] = None,
```

espnet2/asr_transducer/beam_search_transducer.py
```diff
@@ -0,0 +1,711 @@
...
+        topk_idx: torch.Tensor,
+        topk_logp: torch.Tensor,
+    ) -> List[ExtendedHypothesis]:
+        """Return K hypotheses candidates for expansion from a list of hypothesis.
+
+        K candidates are selected according to the extended hypotheses probabilities
+        and a prune-by-value method. Where K is equal to beam_size + beta.
+
+        Args:
+            hyps: Hypotheses.
```
csukuangfj, line 234: Names here don't match the input argument names.

espnet2/asr_transducer/beam_search_transducer.py
```diff
@@ -0,0 +1,711 @@
...
+                    label,
+                    max_hyp.yseq,
+                    max_hyp.dec_state,
+                )
+
+                logp = torch.log_softmax(
+                    self.joint_network(enc_out[t : t + 1, :], dec_out),
+                    dim=-1,
+                ).squeeze(0)
+                top_k = logp[1:].topk(beam_k, dim=-1)
```
csukuangfj, line 333: Does it hard-code the blank_id to 0?

I see that there is a field `blank_id` in `self` and I assume that `blank_id` can be any value between 0 and `vocab_size - 1`.

espnet2/asr_transducer/beam_search_transducer.py
```diff
@@ -0,0 +1,711 @@
...
+                [self.sos] + ([self.blank_id] * (max_len - len(h))) + h[1:]
+                for h in hyps_seq
+            ],
+            device=self.decoder.device,
+        )
+
+    def default_beam_search(self, enc_out: torch.Tensor) -> List[Hypothesis]:
+        """Beam search implementation.
+
+        Modified from https://arxiv.org/pdf/1211.3711.pdf
```
csukuangfj, line 286: I suggest documenting that prefix search is omitted in this implementation.

espnet2/asr_transducer/beam_search_transducer.py
```diff
@@ -0,0 +1,711 @@
+"""Search algorithms for Transducer models."""
```
csukuangfj, line 1: Is there a plan to support decoding in batches?

I see that implementation in this file supports only
decoding a single utterance at a time.

b-flo: There was a plan to make a unified `ScorerInterface` and `BatchScorerInterface` with other models but we faced some difficulties. Supporting batch decoding is on my to do, I'll revisit it after everything related to this version is sorted!

b-flo:
> It is a shame that right_context might also degrade performances in Icefall, you can see that we set the default value to 0. I did not find the reason either, right_context sometimes helps to decrease the WERs sometimes not (You can see my results here https://github.com/k2-fsa/icefall/pull/380#issuecomment-1160934534), we suspected it is because of the training decoding mismatching.

Yes, the report you linked is one of my reference! Taking Libri-100 as example and  `right_context > 0` here, the WER degradation is quite extreme on `*_clean` sets (up to +60%) while the WER degradation is equivalent to what your reported for the `*_other sets. I couldn't find a case where it helped yet.
Funnily enough, by removing some trimming related to `right_context`, I could slightly improve performance.

> we suspected it is because of the training decoding mismatching.

What mismatch are you referring to (static vs dynamic, embedding, etc)?
I'm doing some experiments right now to try limiting the degradation when using smaller decoding chunks. I think the dynamic chunk training procedure has some rooms for improvement.

> [edit]: As for the left_context, I think using larger values does help to improve performance, will use more memory though.

Yes, large values tend to slightly improve performance. I'm wondering if it's worth the extra memory?

pkufool:
> Funnily enough, by removing some trimming related to right_context, I could slightly improve performance.

What is the trimming, I don't  remember we need trimming for right_context.

b-flo:
Found a bug in raw context computation I introduced during a refactor. Bigger values should be expected for `chunk_size` and `left_context` because of this.

The performance variation given `chunk_size` and `left_context` values seems equivalent to what was reported in Icefall now.

b-flo:
I included support for the simplified attention score computation and `BasicNorm` from K2 in the default Conformer implementation. I won't add the other modules from their reworked model here but I'll support a `k2Conformer` in later PRs, alongside other X-former.

@csukuangfj I referenced the pull requests here because PEP8 won't allow longer links in docstrings (referencing commit, file or method won't work). Feel free to propose changes if there are better ways!

espnet2/asr_transducer/encoder/modules/positional_encoding.py
```diff
@@ -0,0 +1,91 @@
...
+            x: Input sequences. (B, T, ?)
+            left_context: Number of frames in left context.
+
+        """
+        time1 = x.size(1) + left_context
+
+        if self.pe is not None:
+            if self.pe.size(1) >= time1 * 2 - 1:
+                if self.pe.dtype != x.dtype or self.pe.device != x.device:
+                    self.pe = self.pe.to(device=x.device, dtype=x.dtype)
```
b-flo, line 47: I'm not sure if there are cases where `self.pe` and inputs have different `device` or `dytpe` properties?

b-flo: Yes, I'm dumb.

sw005320:
While waiting for @pengchengguo and @pyf98's input, I started to check the interface part. Overall, the PR looks good to me from a compatibility perspective :)

@b-flo, I'm curious about how we handle various network architectures.
For example, if we want to import a network architecture in the asr task (say branchformer), do we need to write it for the asr_transducer task from scratch, or is there a way to call (or convert) it?

b-flo:
> @b-flo, I'm curious about how we handle various network architectures.
> For example, if we want to import a network architecture in the asr task (say branchformer), do we need to write it for the asr_transducer task from scratch, or is there a way to call (or convert) it?

Yes, we handle both online and offline cases so you will need to write a new block! However, it's really easy to add one with this design and there are (a lot) less duplications compared to let's say, adding a `ContextualBlockBranchformerEncoder` (+ modules). Also, because we want efficiency here, most options and related code are removed (we don't usually use them anyway).

I already proposed in slack but I can add the `Branchformer` block in this PR to give you an example and revert after that.

espnet2/asr_transducer/espnet_transducer_model.py
```diff
@@ -0,0 +1,466 @@
...
+
+        Return:
+            loss_transducer: Transducer loss value.
+            cer_transducer: Character error rate for Transducer.
+            wer_transducer: Word Error Rate for Transducer.
+
+        """
+        if self.criterion_transducer is None:
+            try:
+                from warprnnt_pytorch import RNNTLoss
```
pengchengguo, line 349: Are you planing to implement `torchaudio.functional.rnnt_loss` in the furture?

espnet2/asr_transducer/error_calculator.py
```diff
@@ -0,0 +1,168 @@
...
+            : Sentence-level WER score.
+
+        """
+        cer, wer = None, None
+
+        batchsize = int(encoder_out.size(0))
+
+        encoder_out = encoder_out.to(next(self.decoder.parameters()).device)
+
+        batch_nbest = [self.beam_search(encoder_out[b]) for b in range(batchsize)]
```
pengchengguo, line 74: Sorry, I don't fully understand here. Do we need beam search to calculate Acc/CER during the training?
Besides, it is possible to combine this class w/ `espnet/net2/e2e_asr_common.py:ErrorCaculator()`.

espnet2/asr_transducer/encoder/modules/positional_encoding.py
```diff
@@ -0,0 +1,91 @@
+"""Positional encoding modules."""
```
pengchengguo, line 1: This class can also be moved to a general directory since ASR Conformer also uses relative positional encoding.
Besides, I am not sure we should make ESPnet2 more close-loop or use definitions in `espnet/nets/pytorch_backend/transformer/embedding.py`.

pengchengguo (approved):
Thanks, @b-flo, it is an excellent PR. The implementations are ok for me.

espnet2/asr_transducer/activation.py
```diff
@@ -0,0 +1,212 @@
+"""Activation functions for Transducer."""
```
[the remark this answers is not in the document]
b-flo: Yes we can!
However, I think it's out of scope for this PR as I would need to do some modification to other tasks. This one is quite different from the general [get_activation(...)](https://github.com/espnet/espnet/blob/fae3afc4039cd22200cd0eda3a753715292fcb23/espnet/nets/pytorch_backend/nets_utils.py#L490)

espnet2/bin/asr_transducer_inference.py
```diff
@@ -0,0 +1,779 @@
...
+                    _end = 0
+
+                    for i in range(_steps):
+                        _end = (i + 1) * speech2text._raw_ctx
+
+                        speech2text(
+                            speech[i * speech2text._raw_ctx : _end], is_final=False
+                        )
+
+                    final_hyps = speech2text(speech[_end : len(speech)], is_final=True)
```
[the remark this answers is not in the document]
b-flo: I don't mind, it would just mean splitting `speech2text.__call___`!
Currently, if `streaming: False` (default), batch of inputs will be passed and the following will be executed: 

```python
            self.asr_model.encoder.dynamic_chunk_training = False

            enc_out, _ = self.asr_model.encoder(feats, feats_length)

            nbest_hyps = self.beam_search(enc_out[0])
```

Same as offline, `foward(...)` methods will be called instead of `chunk_forward(...)`

espnet2/asr_transducer/error_calculator.py
b-flo: > Sorry, I don't fully understand here. Do we need beam search to calculate Acc/CER during the training?

If CER/WER report is enabled, yes. If you're asking why we-re using beam search instead of greedy search, it's because of the unrestricted expansion. It's better for monitoring IMO.

espnet2/asr_transducer/encoder/modules/positional_encoding.py
b-flo: > This class can also be moved to a general directory since ASR Conformer also uses relative positional encoding.

This one is mostly equivalent to `StreamPositionalEncoding(...)`, yes!
If it's okay I would like to keep it despite the duplication as one of my next PR will further modify the class. If you think there are no major revision to warrant a duplication, I'll extend the class from `embedding.py`!

> Besides, I am not sure whether to make ESPnet2 more close-loop or use definitions in espnet/nets/pytorch_backend/transformer/embedding.py.

This is only my opinion but I would rather have duplicated code in different versions than depending on classes from previous version. It can bring a lot of dependencies (in terms of design, convention, etc) and it become difficult in terms of maintenance. Also, the code is sometimes overloaded due to adding support for new thing over time or because some things were implemented without a "global" purpose at time.

espnet2/asr_transducer/beam_search_transducer.py
```diff
@@ -0,0 +1,703 @@
...
+                )
+            )
+
+        return k_expansions
+
+    def create_lm_batch_inputs(self, hyps_seq: List[List[int]]) -> torch.Tensor:
+        """Make batch of inputs with left padding for LM scoring.
+
+        Args:
+            hyps_seqs: Hypothesis sequences.
```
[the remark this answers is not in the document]
b-flo: Good eyes :smile:

espnet2/asr_transducer/encoder/modules/positional_encoding.py
csukuangfj: I think we can remove this if statement and the `to()` statement inside it.

b-flo: I thought so too but I had a mismatch at runtime. I need to do a complete trace because there are unnecessary cast or device pining along the way.

espnet2/asr_transducer/initialize.py
```diff
@@ -0,0 +1,75 @@
+"""Initialize ESPnet2 Transducer model modules."""
```
pyf98, line 1: What is the difference between the initialization of Transducer and that of the other models (i.e., `torch_utils/initialize.py`)? This one does not support `espnet_initialization_fn`?

doc/espnet2_tutorial.md
```diff
@@ -346,3 +346,229 @@ To enable online decoding, the argument `--use_streaming true` should be added t
...
+
+```sh
+asr.sh --asr-transducer true [...]
+```
+
+For Transducer loss computation during training, we rely on a fork of `warp-transducer`. The installation procedure is described [here](https://espnet.github.io/espnet/installation.html#step-3-optional-custom-tool-installation).
+
+**Note:** If you encounter any error related to this tool, please open an issue in ESPnet instead of the [original repository](https://github.com/HawkAaron/warp-transducer/issues).
+
+**Note 2:** We made available FastEmit regularization [[Yu et al., 2021]](https://arxiv.org/pdf/2010.11148) during loss computation. To enable it, `fastemit-lambda` need to be set in `model_conf`:
```
pyf98, line 372: Is it `fastemit-lambda` or `fastemit_lambda`?

b-flo: Good catch, thanks!

espnet2/bin/asr_transducer_inference.py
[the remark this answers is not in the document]
b-flo: For non streaming I used `__call__` and added a `streaming_decode(...)` method! Is that OK like this?

doc/espnet2_tutorial.md
```diff
@@ -346,3 +346,243 @@ To enable online decoding, the argument `--use_streaming true` should be added t
...
+For the user, it means two things. First, some features or modules may not be supported depending on the version used. Second, the usage of some common ASR features or modules may differ between the models. In addition, some core modules (e.g.: `preencoder` or `postencoder`) may be missing in the standalone version until futher testing.
+
+***The following sections of this tutorial are dedicated to the introduction of the version under asr_transducer***. In that regards, the user should keep in mind that most features described here are not available in the first version.
+
+### General usage
+
+To enable Transducer model training or decoding in your experiments, the following option should be supplied to `asr.sh` in your `run.sh`:
+
+```sh
+asr.sh --asr-transducer true [...]
```
pyf98, line 365: Do we need to set `--asr_task asr_transducer`?

espnet2/asr_transducer/initialize.py
b-flo: Thanks for raising this issue! This file is basically useless right now because 1) initialization is `None` in most experiments and we don't support `RNN` blocks now nor "special" init for Transducer.

I kept it because we really need to revisit initialization and I also need to adapt it for custom module. Do you want me to remove it and initialization option (or raise an issue if it's not `None`)?

doc/espnet2_tutorial.md
b-flo: Good catch, forgot to update!

espnet2/asr_transducer/initialize.py
pyf98: Thanks for the answer! Yeah, I think it's reasonable to keep it since we may need customized initialization in the future. For now, it might be better to raise an issue if it's not `None`, otherwise some users can get confused.

pyf98:
Thanks for the great PR! I didn't look into the algorithm itself, but I made a few comments about the `doc` and `init` just now.

I think it is already well organized. I especially like the flexible design of the encoder which supports different hyper-parameters for different blocks (if my understanding is correct) instead of sharing the same config across all encoder blocks.

espnet2/asr_transducer/initialize.py
b-flo: Kept the file but I added a comment at the top + removed call and import of `initialize` in model init

b-flo:
Thanks a lot @pyf98 and @pengchengguo 

>  I especially like the flexible design of the encoder which supports different hyper-parameters for different blocks (if my understanding is correct) instead of sharing the same config across all encoder blocks.

Your understanding is correct! You can also mix blocks if you want (well in next PRs)!  I'll add some ensemble methods and revisit auxiliary losses with intermediate representations for that.

doc/espnet2_tutorial.md
```diff
@@ -346,3 +346,243 @@ To enable online decoding, the argument `--use_streaming true` should be added t
...
+    right_context: Number of frames in the right context. (int, default = 0)
+
+For each parameter, the number of frames is defined AFTER subsampling, meaning the input chunk will be bigger than the one provided. The input size is determined by the frontend and the input block's subsampling, given `chunk_size + right_context` defining the decoding window.
+
+***Note:*** Because the training part does not consider the right context, relying on `right_context` during decoding may result in a mismatch and performance degration.
+***Note 2:*** All search algorithms but ALSD are available with chunk-by-chunk inference.
+
+### FAQ
+
+#### How to add a new block type to the custom encoder?
```
b-flo, line 578: @sw005320 I added a FAQ with a doc for new block type addition. Is this good enough?
Also, is there anything else I should add in the FAQ? @pengchengguo , @pyf98 ?

pyf98: It looks good to me.

danpovey:
> I included support for the simplified attention score computation and `BasicNorm` from K2 in the default Conformer implementation. I won't add the other modules from their reworked model here but I'll support a `k2Conformer` in later PRs, alongside other X-former.
> 
> @csukuangfj I referenced the pull requests here because PEP8 won't allow longer links in docstrings (referencing commit, file or method won't work). Feel free to propose changes if there are better ways!

Just FYI, one of the changes I made in our Conformer was to remove the normalization from the individual modules inside the conformer layer.  I only expect this to work well if you are using the ScaledLinear/ScaledConv1d modules, which learn a scaling factor for each weight and bias.  Otherwise it has no way to learn the appropriate scale on each sub-module except for scaling the whole weight matrix, which is difficult for SGD to learn.  BasicNorm would not be expected to be a good solution for the normalization for the individual modules, because it does not support an overall scale on the output.

Also, I am working on (still tuning) an optimization method that will learn the parameter scales as part of the optimizer, without requiring the individual scales for weights and biases, so I expect to eventually remove the ScaledLinear and ScaledConv1d (in newer directories), but the recipe will depend on properties of the optimizer.

b-flo:
Thanks a lot for the explanation @danpovey !!
To be honest, I didn't expect `BasicNorm` to be a good replacement candidate for the reason you gave and it was removed/added back multiple times. However, because it was found appropriate (i.e.: same performance at cheapest cost) in some setups, I decided to keep it in the end.

I'm reworking the normalization module definition, I'll add some warning and explanations to the class doc. I'm also testing [`AdaNorm`](https://arxiv.org/pdf/1911.07013.pdf) right now.

> Also, I am working on (still tuning) an optimization method that will learn the parameter scales as part of the optimizer, without requiring the individual scales for weights and biases, so I expect to eventually remove the ScaledLinear and ScaledConv1d (in newer directories), but the recipe will depend on properties of the optimizer.

Thanks for the update, I'll keep an eye on the development!

espnet2/bin/asr_transducer_inference.py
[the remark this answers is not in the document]
pengchengguo: Sure! I think it is ok.

espnet2/asr_transducer/error_calculator.py
pengchengguo: I see, thanks for the explanation.

espnet2/asr_transducer/encoder/modules/positional_encoding.py
pengchengguo: It is okay for me to keep the current setups :)

espnet2/asr_transducer/initialize.py
b-flo: In fact, I'll remove the initialization part for now! I'm not too fan of keeping dead code even if it should be temporary.

doc/espnet2_tutorial.md
```diff
@@ -346,3 +346,243 @@ To enable online decoding, the argument `--use_streaming true` should be added t
...
+    model_conf:
+      transducer_weight: Weight of the Transducer loss (float, default = 1.0)
+      auxiliary_ctc_weight: Weight of the CTC loss. (float, default = 0.0)
+      auxiliary_ctc_dropout_rate: Dropout rate for the CTC loss inputs. (float, default = 0.0)
+      auxiliary_lm_loss_weight: Weight of the LM loss. (float, default = 0.2)
+      auxiliary_lm_loss_smoothing: Smoothing rate for LM loss. If > 0, label smoothing is enabled. (float, default = 0.0)
+
+**Note:** We do not support other auxiliary tasks in ESPnet2 yet.
+
+### Inference
```
pyf98, line 522: In this section, the hyper-parameter names have hyphens `-` instead of underscores `_`. It would be better to make them consistent.

pyf98:
I have got two questions.
1. Does it support GPU inference?
2. Does it support automatic mixed precision training with `use_amp: true`?

I'm increasing the model size to 120M and extending the number of epochs to 60.

csukuangfj:
> For LibriSpeech, I'm increasing the nonstreaming model size to 120M and extending the number of epochs to 60.

Does the model need to be so large and does it need to be trained for so many epochs?

We are using a model with about 80 M parameters and training it for 30 epochs on the LibriSpeech dataset in [icefall](https://github.com/k2-fsa/icefall)

b-flo:
> Does it support GPU inference?

It does! If it's not, that's a bug on my part.

> Does it support automatic mixed precision training with use_amp: true?
 
Sorry, not yet, I need to update `warp-transducer` for that. Let me finish some other things first and I'll work on it (I'll take a look this weekend)!

> Does the model need to be so large and does it need to be trained for so many epochs?
> We are using a model with about 80 M parameters and training it for 30 epochs on the LibriSpeech dataset in [icefall](https://github.com/k2-fsa/icefall)

Good question.
We usually (at least from my experience) prioritize performance in terms of xER over anything else in ESPnet for ASR. We sometimes add many extra epochs or rely on really large architecture to squeeze some .x%.

That being said:

1) I do think 120M parameters is too much. I don't mind for the offline version, but we should be careful about the number of parameters in the online model.
2) 60 epochs is OK in comparison to other model training in ESPnet. However, we really need to improve training in terms of stability and efficiency. That will be the focus after next PR, alongside initialization.

b-flo:
Small update:

1) I removed the parts related to initialization as we don't use them in our experiments. It'll be reworked in a later PR.
2) I refactored the normalization module for future additions/works. I also added `RMSNorm` and `ScaleNorm` but it was not extensively tested (I also tried `AdaNorm` but found it difficult to converge).

ypeng-asapp:
> > For LibriSpeech, I'm increasing the nonstreaming model size to 120M and extending the number of epochs to 60.
> 
> Does the model need to be so large and does it need to be trained for so many epochs?
> 
> We are using a model with about 80 M parameters and training it for 30 epochs on the LibriSpeech dataset in [icefall](https://github.com/k2-fsa/icefall)

Thanks for the info! I don't have a specific goal. I just try to match the original Conformer-Transducer Large config and see how it performs.

pyf98:
> > For LibriSpeech, I'm increasing the nonstreaming model size to 120M and extending the number of epochs to 60.
> 
> Does the model need to be so large and does it need to be trained for so many epochs?
> 
> We are using a model with about 80 M parameters and training it for 30 epochs on the LibriSpeech dataset in [icefall](https://github.com/k2-fsa/icefall)

Thanks for the info! I just try to match the original Conformer-Transducer Large config and see how it performs. This would be an interesting investigation. I don't know if 30 epochs is sufficient for Transducer, but at least it is not for our joint CTC/Attention.

b-flo:
After discussion, I'm merging this PR! It was a long road for this one, thanks to everyone for your help :tada:

Now onto the next items!

PR #4480, opened 2022-06-30
raise more useful error in espnet2/asr/frontend/s3prl.py if s3prl is not installed

popcornell:


espnet2/asr/frontend/s3prl.py
```diff
@@ -63,7 +63,8 @@ def _get_upstream(self, frontend_conf):
             if p.endswith("s3prl"):
                 s3prl_path = p
                 break
-        assert s3prl_path is not None
+        assert s3prl_path is not None, "Cannot find s3prl in your PYTHONPATH, " \
+                               "you can install it via pip install s3prl"
```
ftshijt, line 67: Maybe you can refer the installer in tools?

popcornell: you are right. `s3prl_path` should be `${YOUR_PATH2ESPNet}/espnet/tools/s3prl`. 
And the code above can find it rightly when you source `path.sh` in an ESPNet recipe. 
However it cannot do it when you just pip install espnet-model-zoo and try to use the pretrained model without espnet or s3prl. We need to devise a method to point out to a s3prl git cloned directory (pip package won't work I tried).

popcornell: This means that espnet model zoo models that rely on s3prl front-ends are broken outside recipes

sw005320: Yes, you're right. I think we need to deal with this more smart way for easy inference.
@simpleoier, can you consider it?

At this moment, we can just add meaningful error messages through an installer and `path.sh`
@popcornell and @ftshijt, can you do it?

popcornell: I ll handle the error right now

popcornell: The code now works without installing espnet (only model-zoo) but you need to git clone s3prl locally and install it from the repo. I guess it is a reasonable compromise.

sw005320 (approved):

PR #4490, opened 2022-07-06
fix the corrupted pretrained model

wentaoxandry:
The pretrained model was corrupted. fixed the download link


sw005320:
Thanks for your quick fix, @wentaoxandry!

Issue #4498, opened 2022-07-08
char_tokenizer use fixed space_symbol "<space>" instead of `self.space_symbol`


xiabingquan:
**Describe the bug**
A bug about the [char_tokenizer](https://github.com/espnet/espnet/blob/master/espnet2/text/char_tokenizer.py).
In the class `CharTokenizer`, the space_symbol is defined by `self.space_symbol`. However, in the method `text2tokens`t, it says

```python
                t = line[0]
                if t == " ":
                    t = "<space>"
                tokens.append(t)
                line = line[1:]
```
It means that the space symbol is fixed to be `<space>` instead of `self.space_symbol`

It should be

```python
                t = line[0]
                if t == " ":
                    t = self.sapce_symbol
                tokens.append(t)
                line = line[1:]
```

In this way, the settings of `sapce_symbol` can be in line with the method `tokens2text`:
```python
tokens = [t if t != self.space_symbol else " " for t in tokens]
```

kamo-naoyuki:
Thanks. I think you are completely correct. Can you create a PR to fix it?

xiabingquan:
You're welcome. I'll create a PR now : )


| |
夏炳权
|
|
***@***.***
|
On 7/8/2022 16:36，Naoyuki ***@***.***> wrote：

Thanks. I think you are completely correct. Can you create a PR to fix it?

xiabingquan:
Fixed in this [PR](https://github.com/espnet/espnet/pull/4499)

PR #4499, opened 2022-07-08
Update char_tokenizer.py

xiabingquan:
Fix this [issue](https://github.com/espnet/espnet/issues/4498)

In the method `text2tokens` of `CharTokenizer`. the `space_symbol` is set to `"<space>"` and can not be changed, but the right value should be `self.space_symbol`.


kamo-naoyuki (approved):
LGTM, thanks

PR #4501, opened 2022-07-10
Aishell-2 GPU info

jctian98:
This PR is to add the GPU info in the training config files. 
For both hybrid CTC/Attention frameworks and transducer frameworks, I'm using 8 V100 GPUs and the training will take around 4 days


b-flo (approved):

b-flo:
Thanks a lot @jctian98 !

@sw005320 It was discussed in the past but we need to update the `README.md` structure (and related files) to add some information about GPUs used (number and models), average VRAM and peaks, training and decoding time, etc. I'll make a draft when I have time.

jctian98:
It seems the CI check fails with the info `espnet/distributed/pytorch_backend/launch.py:226: Failed`. As I only add some comments in this PR, I'm a bit confused why this happens

b-flo:
There was a timeout, nothing to do with your PR itself. I'll manually merge.

PR #4517, opened 2022-07-17
Fix librispeech default path to signify auto download 

karthik19967829:
As Librispeech recipe automatically downloads the dataset, adding downloads against the path


ftshijt:
Sorry, I do not fully understand why the change is necessary. In my view, Librispeech is relatively large to download, it would be better for users to add their own preferences of the directory in `db.sh` for memory safety.

karthik19967829:
Yes that's one view point . I just wanted to indicate a way to indicate that it downloads. As the db.sh header says that all datasets with downloads has auto download

sw005320:
If we strictly follow the rule, @karthik19967829 is right.
I'll merge this PR.
Thanks, @karthik19967829!

PR #4312, opened 2022-04-24
Fix mAES with big vocab size

b-flo:
Fixes #4303. To merge after #4032 to avoid dealing with conflicts (beam search interface is currently shared by ESPnet1 and ESPnet2).

PR #4520, opened 2022-07-20
Beamsearch for transducer ASR, that is, maes needs to be modified

jeon30c:
The beamsearch method, named MAES based on the paper https://ieeexplore.ieee.org/abstract/document/9250505 implemented in espnet has two problems.

1) Actual beam size does not vary (if the beamsize=20, the actual beam size will be 20 after saturation).
2) The algorithm has several copied of a certain beam meaning some beams share the same yseq, score, state.

That is due to not implementing duplication check of fig 2. in the original paper.

After adding duplication check method, the problems could be solved.

### DIFF
--- a/espnet/nets/beam_search_transducer.py
+++ b/espnet/nets/beam_search_transducer.py
@@ -779,6 +779,8 @@ class BeamSearchTransducer:
             beam_enc_out = enc_out_t.unsqueeze(0)
 
             list_b = []
+            duplication_check = [hyp.yseq for hyp in hyps]
+            
             for n in range(self.nstep):
                 beam_dec_out = torch.stack([h.dec_out[-1] for h in hyps])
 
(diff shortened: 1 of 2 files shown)


b-flo:
Hi @jeon30c,

We are aware of two issues in ESPnet1/ESPnet2 mAES: 1) duplication happens during search and big vocab takes a long time (due to current topk selection).

It's already fixed [in the new version](https://github.com/espnet/espnet/pull/4479) and other versions will be updated after the merge. Sorry about that, it was to avoid too many conflicts.

Btw, for duplication check, we're using ALSD's `recombine_hyps` between each timestep instead.

jeon30c:
@b-flo It's cool. Then I am closing this PR.

b-flo:
Sorry again, I'll try to fix it ASAP in other versions..
Btw, feel free to comment on the version I linked (PR #4479) if you see any issue or improvement!

sw005320:
@jeon30c and @b-flo,
Could you continue this PR?
Many people are using the current RNN-T implementation, and this PR benefits them.
We want to keep the compatibility and maintain both for a while (like we're doing for espnet1 and espnet2).

b-flo:
> We want to keep the compatibility and maintain both for a while (like we're doing for espnet1 and espnet2).

Hum, yes but I wanted to do that after the new version is merged to avoid conflicts. From here, I'll add the changes in the other versions.
Btw, there is already PR #4312 which fixes the issue with big vocab. I'll extend it for duplication (which is not critical).

sw005320:
No, this would not have large conflicts with your PRs.
This PR is in espent1 and espnet2/asr/transducer

The compatibility in the current master branch is important for management.
It would be great if you understood it.

b-flo:
> The compatibility in the current master branch is important for management.
> It would be great if you understood it.

I do understand but given that 1) PR #4312 is open since April and went unnoticed and 2) only one person raised the issue, I thought it could wait before re-working the other versions.
Anyway, I'll add a duplication check in other PR and merge it.

sw005320:
OK, thanks for your understanding.

PR #4530, opened 2022-07-24
Refactoring 'is_prefix' function

jhlee9010:
I found that recognition time become slower as the utterance length is long in streaming inference for transducer model.
In my case, utterance length was more than 10 minutes.
By refactoring 'is_prefix' function, I could save some time.
Because we already checked front part of prefix sequences in previous steps, it's more efficient to check from the end of prefix sequences.


b-flo:
Hi,

Thanks for the optimization! 

> In my case, utterance length was more than 10 minutes.

There is no reason not to approve the PR but could you give a small benchmark with different lengths, please?
Usually, we feed chunks/audios of less than 2-3minutes so 10 minutes is somewhat an extreme case.

jhlee9010:
@b-flo Sure!
Actually, I tested streaming inference of contextual conformer transducer with in-house Korean dataset.
Before refactoring: rtf 0.09 for 4min length utterance, rtf 0.24 for 22min length utterance.
After refactoring: rtf 0.09 for 4min length utterance, rtf 0.19 for 22min length utterance.

Now, I started to train English model with Librispeeh for re-verification. When it finished, I'll share the results.

b-flo:
> Actually, I tested streaming inference of contextual conformer transducer with in-house Korean dataset.
> Before refactoring: rtf 0.09 for 4min length utterance, rtf 0.24 for 22min length utterance.
> After refactoring: rtf 0.09 for 4min length utterance, rtf 0.19 for 22min length utterance.

Sorry, I'm a bit confused here. How did you compute the RTF here?

> Now, I started to train English model with Librispeeh for re-verification. When it finished, I'll share the results.

Could you please use Librispeech-100 instead? I'm doing some experiments on it with another streaming Transducer and it would make a good comparison.

jhlee9010:
@b-flo  
> > Sorry, I'm a bit confused here. How did you compute the RTF here?
> Calculation method is 'RTF = inference time/utterance length' and gpu inference was used. 
> > Could you please use Librispeech-100 instead? I'm doing some experiments on it with another streaming Transducer and it would make a good comparison.
> Sure!

b-flo:
> Calculation method is 'RTF = inference time/utterance length' and gpu inference was used.

Sorry, I may have badly asked. I know the RTF formula but I was asking how it was computed in case of the streaming process. It seems rather low here (not that I complain! haha)

> Could you please use Librispeech-100 instead? I'm doing some experiments on it with another streaming Transducer and it would make a good comparison.

> > Sure!

Thanks a lot! I'll merge this PR after CI pass.
Could you open a new PR to add Streaming Transducer for Librispeech-100 please? We don't have a recipe for it so it would make a great addition!
We can help you tune it if needed!

Issue #4543, opened 2022-07-27
Multiple TracerWarning on every epoch


iamanigeeit:
I am running FastSpeech2 and every epoch gives me this:

```
/home/perry/PycharmProjects/espnet/espnet/nets/pytorch_backend/nets_utils.py:154: TracerWarning: Converting a tensor to a Python list might cause the trace to be incorrect. We can't record the data flow of Python values, so this value will be treated as a constant in the future. This means that the trace might not generalize to other inputs!
  lengths = lengths.long().tolist()
/home/perry/PycharmProjects/espnet/espnet/nets/pytorch_backend/nets_utils.py:172: TracerWarning: Converting a tensor to a Python boolean might cause the trace to be incorrect. We can't record the data flow of Python values, so this value will be treated as a constant in the future. This means that the trace might not generalize to other inputs!
  assert xs.size(0) == bs, (xs.size(0), bs)
/home/perry/PycharmProjects/espnet/espnet2/tts/feats_extract/log_mel_fbank.py:92: TracerWarning: Converting a tensor to a Python boolean might cause the trace to be incorrect. We can't record the data flow of Python values, so this value will be treated as a constant in the future. This means that the trace might not generalize to other inputs!
  assert input_stft.shape[-1] == 2, input_stft.shape
/home/perry/PycharmProjects/espnet/espnet2/tts/fastspeech2/fastspeech2.py:533: TracerWarning: Iterating over a tensor might cause the trace to be incorrect. Passing a tensor of different shape won't change the number of iterations executed (and might lead to errors or silently give incorrect results).
  for i, l in enumerate(text_lengths):
/home/perry/PycharmProjects/espnet/espnet/nets/pytorch_backend/transformer/embedding.py:61: TracerWarning: Converting a tensor to a Python boolean might cause the trace to be incorrect. We can't record the data flow of Python values, so this value will be treated as a constant in the future. This means that the trace might not generalize to other inputs!
  if self.pe.size(1) >= x.size(1):
/home/perry/PycharmProjects/espnet/espnet/nets/pytorch_backend/transformer/attention.py:81: TracerWarning: torch.tensor results are registered as constants in the trace. You can safely ignore this warning if you use this function to create tensors out of constant variables that would be the same every time you call this function. In any other case, this might cause the trace to be incorrect.
  numpy.finfo(torch.tensor(0, dtype=scores.dtype).numpy().dtype).min
[...]

PR #4551, opened 2022-08-01
Optionize graph creation

kan-bayashi:
This PR optionizes the graph creation in tensorboard.
The graph in tensorboard is informative but it requires extra gpu memory and show a lot of tracerwarning (See #4543).
Therefore, it should be optionized and be False as a default.

### DIFF
--- a/espnet2/tasks/abs_task.py
+++ b/espnet2/tasks/abs_task.py
@@ -572,7 +572,7 @@ class AbsTask(ABC):
             help="Enable tensorboard logging",
         )
         group.add_argument(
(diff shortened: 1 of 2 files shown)


kamo-naoyuki:
LGTM

iamanigeeit:
Sorry, just to add: it doesn't use extra GPU memory, but it uses CPU memory (linearly increase every epoch until i hit out-of-memory)

kan-bayashi:
@iamanigeeit I see. Thank you for kind explanation.

PR #4563, opened 2022-08-10
[not-to-merge] ESPnet2 integration tests

b-flo:
Hi,

This PR is to continue the discussion on the integration and unit tests for ESPnet2.

As @ShigekiKarita and @kan-bayashi said, there are some straightforward ways to reduce execution time, such as: 1) defining default configs with modified parameters for time-dependent model/tasks, and 2) disable or move some operations related to sub-task, notably when we test in loop.
With that, I could reduce ASR tests exec. time by about 60%.

However, I don’t think it'll help in the long run. Outside of parameters selection, the current overload is mainly due to two things:

- Redundant (or unnecessary due to overlap) testing.
- Validation of the config files.

For first point, I think a major refactor may be needed given the execution time and current code coverage. We should carefully define what should be covered in integration tests and unit tests and how (there are a lot of dummy tests, mine included).

For the second point, I’m not sure but we could maybe execute specific tests depending on some tags? Let's say, before each release or when a PR modify an existing recipe?

### DIFF
--- a/ci/test_integration_espnet2.sh
+++ b/ci/test_integration_espnet2.sh
@@ -11,48 +11,58 @@ cwd=$(pwd)
 #### Make sure chainer-independent ####
 python3 -m pip uninstall -y chainer
 
+# default config paths (for each task).
+train_conf=conf/train.yaml
+decode_conf=conf/decode.yaml
+lm_conf=conf/lm.yaml
+
 # [ESPnet2] test asr recipe
-cd ./egs2/mini_an4/asr1
 echo "==== [ESPnet2] ASR ==="
-./run.sh --stage 1 --stop-stage 1
+cd ./egs2/mini_an4/asr1
+
 feats_types="raw fbank_pitch"
 token_types="bpe char"
+asr_add_args="--nbpe 22 --ngram-num 1 --word-vocab-size 5 --compute_rtf false"
+
+./run.sh --stage 1 --stop-stage 1
+
 for t in ${feats_types}; do
     ./run.sh --stage 2 --stop-stage 4 --feats-type "${t}" --python "${python}"
 done
 for t in ${token_types}; do
-    ./run.sh --stage 5 --stop-stage 5 --token-type "${t}" --python "${python}"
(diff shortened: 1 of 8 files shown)


egs2/mini_an4/asr1/run.sh
```diff
@@ -9,5 +9,5 @@ set -o pipefail
     --lang en \
     --train_set train_nodev \
     --valid_set train_dev \
-    --test_sets "train_dev test test_seg" \
+    --test_sets "test test_seg" \
```
b-flo, line 12: I guess we don't need to use all sets for integration tests? It may reduce decoding for some tests (with LM notably).

We can directly modify the recipe if it's only used in integration tests or overwrite the option in integration tests scripts.

ci/test_integration_espnet2.sh
```diff
@@ -11,48 +11,58 @@ cwd=$(pwd)
...
-
-    echo "==== use_k2, num_paths == nll_batch_size, feats_type=raw, token_types=bpe, model_conf.extract_feats_in_collect_stats=False, normalize=utt_mvn ==="
-    ./run.sh --num_paths 20 --nll_batch_size 20 --use_k2 true --ngpu 0 --stage 12 --stop-stage 13 --skip-upload false --feats-type "raw" --token-type "bpe" \
-       --feats_normalize "utterance_mvn" --lm-args "--max_epoch=1" --python "${python}" \
-       --asr-args "--model_conf extract_feats_in_collect_stats=false --max_epoch=1"
-fi
+    decode_k2_conf=conf/decode_k2.yaml
+    asr_add_args+=" --use_k2 true --num_paths 5 --nll_batch_size 2 --use_nbest_rescoring false"
 
+    echo "=== use_k2, num_paths > nll_batch_size, feats_type=raw, token_types=char, model_conf.extract_feats_in_collect_stats=False, normalize=utt_mvn ==="
```
b-flo, line 60: Btw, do we need to test `num_paths > nll_batch_size` and `num_paths == nll_batch_size` here?

Also, sorry, I forgot k2 tests were not executed on my current machine (due to no compatible build available for my setup). I will fix it!

PR #4566, opened 2022-08-11
Support enh_s2t joint training on multi-speaker data

Emrys365:
As discussed with @simpleoier and @YoshikiMas, I made this PR to add support for joint training of Enh and ASR tasks on multi-speaker data. (The compatibility with other S2T tasks is preserved as before.)

Most changes are made in two scripts:
  * `egs2/TEMPLATE/enh_asr1/enh_asr.sh`
    * I replace all `text` data files with speaker-related ones (`text_spk1`, `text_spk2`, etc.)
    * I added support for loading noise and dereverberation reference signals during training.
    * The ASR evaluation stage is also modified accordingly to support multi-speaker data.

  * `espnet2/enh/espnet_enh_s2t_model.py`
    * I replace all `text` data files with speaker-related ones (`text_spk1`, `text_spk2`, etc.)
    * Permutation invariant training related code is added.
    * NOTE: the multi-condition training related code should be rewritten to utilize the information from `utt2category`.

Note:
  * Currently, the `bypass_enh_prob` flag cannot be used for multi-speaker cases, as the multi-condition training function is not well designed for now. We can add support for this after the multi-condition training part is refactored.

TODO:
- [x] The inference-related code should also be modified accordingly.

I also added a recipe in `egs2/wsj0_2mix_spatialized/enh_asr1` for training [MIMO-Speech](https://arxiv.org/abs/1910.06522) style models.


espnet2/enh/espnet_enh_s2t_model.py
```diff
@@ -329,6 +401,97 @@ def nll(
 
     batchify_nll = ESPnetASRModel.batchify_nll
 
+    def asr_pit_loss(self, speech, speech_lengths, text, text_lengths):
```
simpleoier, line 404: I have a rough idea about implementing all the pit loss within a PIT class. Maybe we can implement this later. In this PIT class, it can compute the pit loss for all criterions and determine the optimal permutation, can reorder the data, etc. The major target is to keep the code minimal and clean.

egs2/wsj0_2mix_spatialized/enh_asr1/conf/tuning/train_enh_asr_beamformer_fbank_transformer.yaml
```diff
@@ -0,0 +1,131 @@
...
+    bdropout_rate: 0.0
+    shared_power: False
+enh_criterions:
+  # The first criterion
+  - name: si_snr
+    conf:
+        eps: 1.0e-7
+    # the wrapper for the current criterion
+    # for single-talker case, we simplely use fixed_order wrapper
+    wrapper: fixed_order
```
simpleoier, line 64: This should be PIT, right? We should also explicitly indicate whether the permutation is determined by enh_loss or ctc_loss. Ignore this comment if there is already such indication.

egs2/chime4/enh_asr1/run.sh
```diff
@@ -41,5 +41,5 @@ word_vocab_size=65000
     --train_set "${train_set}" \
     --valid_set "${valid_set}" \
     --test_sets "${test_sets}" \
-    --bpe_train_text "data/${train_set}/text" \
-    --lm_train_text "data/${train_set}/text data/local/other_text/text" "$@"
+    --bpe_train_text "data/${train_set}/text_spk1" \
```
simpleoier, line 44: I think we can keep `text` here to be the same as chime4/asr1/run.sh. But either is fine.

simpleoier (commented):
Thanks @Emrys365 . If the training experiment works, it would be good.

espnet2/enh/espnet_enh_s2t_model.py
Emrys365: Sure.  Indeed, this part can be wrapped as a standalone module. We can add support for pit later, and for now we can always use `enh_model` to determine the permutation.

egs2/wsj0_2mix_spatialized/enh_asr1/conf/tuning/train_enh_asr_beamformer_fbank_transformer.yaml
Emrys365: You are right. I will handle this issue.

egs2/chime4/enh_asr1/run.sh
Emrys365: Yes, I made a symbolic link in `local/data.sh` for chime4. Here I just want to explicitly use `text_spk1` to reminder users. Because for multi-speaker data, the use of `text` could cause errors.

egs2/wsj0_2mix_spatialized/enh_asr1/run.sh
```diff
@@ -0,0 +1,48 @@
...
+
+
+sample_rate=16k     # 8k or 16k
+
+
+train_set=tr_spatialized_anechoic_multich
+valid_set=cv_spatialized_anechoic_multich
+test_sets="tt_spatialized_anechoic_multich"
+
+# train_set=tr_spatialized_reverb_multich
```
sw005320, line 16: add some comments (or remove them).

espnet2/asr/espnet_model.py
```diff
@@ -558,3 +558,36 @@ def _calc_transducer_loss(
             )
 
         return loss_transducer, cer_transducer, wer_transducer
+
+    def _calc_batch_ctc_loss(
```
sw005320, line 562: It this function included in the unit test?

sw005320 (commented):
Thanks a lot!
The overall design looks good to me.

espnet2/enh/espnet_enh_s2t_model.py
YoshikiMas: Related to another comment from @simpleoier, can we use the permutation defined by `asr_pit_loss` for the enhancement loss?

Emrys365: > can we use the permutation defined by asr_pit_loss for the enhancement loss?

This is possible. But the implementation might be complex. We can consider adding this function later.

espnet2/asr/espnet_model.py
Emrys365: Yes, it is implicitly covered by `test_enh_asr_model_2spk` in `test/espnet2/enh/test_espnet_enh_s2t_model.py` when `calc_enh_loss` is True.

egs2/wsj0_2mix_spatialized/enh_asr1/run.sh
Emrys365: Done

espnet2/enh/espnet_enh_s2t_model.py
YoshikiMas: I see, we can do it later.

Emrys365:
Hi @simpleoier and @YoshikiMas, could you review the new changes?

Emrys365:
> Thanks @Emrys365 . If the training experiment works, it would be good.

I've verified the training, inference, and scoring stages can be run successfully.

egs2/TEMPLATE/enh_asr1/scripts/utils/eval_perm_free_error.py
```diff
@@ -0,0 +1,232 @@
...
+    min_scores, hyp_perms = [], []
+    for idx, error_rate in enumerate(all_error_rates):
+        row_idx, col_idx = linear_sum_assignment(error_rate)
+
+        hyp_perms.append(col_idx)
+        min_scores.append(np.sum(all_scores[idx, row_idx, col_idx], axis=0))
+
+    min_scores = np.stack(min_scores)
+
+    # Print results
```
YoshikiMas, line 80: Is it better to leave the following comment-outed lines?

YoshikiMas (commented):
I left a minor question.

egs2/TEMPLATE/enh_asr1/scripts/utils/eval_perm_free_error.py
Emrys365: I think so.
This script was written by @simpleoier. Do you think we can remove the commented part?

Emrys365:
One point that needs discussion is that:
  * is it better to add a scaling factor for `loss_enh` when both Enh and ASR losses are used?

Because Enh and ASR losses may have different treatment for the sequence length dimension (average or sum), the resultant loss scale can be very different and impedes the joint training.

YoshikiMas:
I think `weight` in `wrapper_conf` for `enh_criterions` can adjust the ratio between two losses.
And, can `length_normalized_loss` help reducing such a discrepancy further? I didn't try it yet.

Emrys365:
> I think `weight` in `wrapper_conf` for `enh_criterions` can adjust the ratio between two losses. And, can `length_normalized_loss` help reducing such a discrepancy further? I didn't try it yet.

Oh yes, `enh_criterions` can be used to scale `loss_enh`.

`length_normalized_loss` should also be helpful, but we may have to care about all ASR losses if this flag is used, because it only impacts `loss_att`.

egs2/TEMPLATE/enh_asr1/scripts/utils/eval_perm_free_error.py
sw005320: We can just remove them.

YoshikiMas:
> length_normalized_loss should also be helpful, but we may have to care about all ASR losses if this flag is used, because it only impacts loss_att.

I see. If the training works with the single scale parameter (`weight`), the current one is good enough for me as an initial implementation.

espnet2/enh/espnet_enh_s2t_model.py
```diff
@@ -66,19 +77,72 @@ def forward(
...
+                ...
+            For other tasks:
+                text: (Batch, Length) default None just to keep the argument order
+                text_lengths: (Batch,)
+                    default None for the same reason as speech_lengths
         """
-        if text_lengths is not None:
-            assert text_lengths.dim() == 1, text_lengths.shape
-        if speech_lengths is not None and text_lengths is not None:
+        if "text" in kwargs:
```
simpleoier, line 92: Can we make it compact here? What do you think about the following:
```python
if any(key.startswith("text_spk") for key in kwargs.keys()):  # to check if there are text input using "text_spk*"
    filtered_keys = filter(...). # filter the keys of text_spk*
    num_refs = len(filtered_keys)
    text_ref_keys = [f"text_spk{i}" for i in range(num_refs)]
else:
    text_ref_keys = ["text"]

text = []
for key in text_ref_keys:
    text.append(kwargs.get(key, None))
    text_ref_lengths.append(kwargs.get(key + "lengths", None))

# remember to check the number of speakers is consistant with enhancement model
assert len(text) == self.enh_model.num_spk
```

egs2/TEMPLATE/enh_asr1/scripts/utils/eval_perm_free_error.py
simpleoier: Yes, removing them is fine.

espnet2/enh/espnet_enh_s2t_model.py
Emrys365: For the legacy case (`text` is used), the variables `text` and `text_ref_lengths` should be a tensor instead of a list. I feel it is a bit complicated to process both cases with unified code.

Emrys365: For the legacy case (`text` is used), the variables `text` and `text_ref_lengths` should be a tensor instead of a list. I feel it is a bit complicated to process both cases with unified code.

egs2/TEMPLATE/enh_asr1/scripts/utils/eval_perm_free_error.py
Emrys365: OK, I will do that.

Emrys365:
> If the training works with the single scale parameter (`weight`), the current one is good enough for me as an initial implementation.

I think you can work with it now. Below are my training curves based on `conf/tuning/train_enh_asr_beamformer_fbank_transformer.yaml`:
mtl_enh_asr

It took 4 days to finish 50 epochs with a single TITAN RTX (24 GB).

> I am now training the language model. After that, I can obtain the decoding performance.

sw005320:
Hi @Emrys365, Great PR!
I think we can marge this PR now, and then we can do some tuning.

I have one major comment that we can consider in the future.
We may think of the multi-speaker case as new ASR or new SE-ASR joint model instead of hacking the enh_s2t model (like we make asr_model and enh_s2t_model separately).
In the future, we should also think of adding a target speaker model or recursive speaker model, and this separated models (or we can even make a new task) could be more flexible.

Emrys365:
> Hi @Emrys365, Great PR! I think we can marge this PR now, and then we can do some tuning.
> 
> I have one major comment that we can consider in the future. We may think of the multi-speaker case as new ASR or new SE-ASR joint model instead of hacking the enh_s2t model (like we make asr_model and enh_s2t_model separately). In the future, we should also think of adding a target speaker model or recursive speaker model, and this separated models (or we can even make a new task) could be more flexible.

@sw005320 Thanks for your comments! I totally agree that it is better to split the single-speaker and multi-speaker functions into different scripts.

For the target speaker extraction function, I have an espnet2-based implementation at hand, but it needs more work to tidy up. Anyway, we may need to carefully discuss the overall design some time later.

Emrys365:
I just updated the experimental results in `egs2/wsj0_2mix_spatialized/enh_asr1`.

PR #4582, opened 2022-08-18
Offline/Online Branchformer Transducer

b-flo:
Hi,

This PR add Branchformer block with offline and streaming capabilities to the new Transducer version.
Btw, I need to redesign the units tests in the future, it'll become a mess if I pile up block tests. It's OK for now because the other block types aren't a priority.

Also, as follow up to #4479:

- Set the conv mod normalization module for Conformer to `BatchNorm1d`. Subsequently, the type was removed from `get_normalization(...)`.
- Fix some main and Conformer parameters in the ESPnet2 tutorial doc.
- Moved the dynamic chunk training parameters description from Encoder section to Streaming section.

### DIFF
--- a/espnet2/asr_transducer/encoder/modules/convolution.py
+++ b/espnet2/asr_transducer/encoder/modules/convolution.py
@@ -9,8 +9,8 @@ class ConformerConvolution(torch.nn.Module):
     """ConformerConvolution module definition.
 
     Args:
-        channels: The number of channels of conv layers.
-        kernel_size: Kernel size of conv layers.
+        channels: The number of channels.
+        kernel_size: Size of the convolving kernel.
         activation: Type of activation function.
         norm_class: Normalization module class.
         norm_args: Normalization module arguments.
@@ -112,8 +112,8 @@ class ConvolutionalSpatialGatingUnit(torch.nn.Module):
     """Convolutional Spatial Gating Unit module definition.
 
     Args:
-        size: Initial size to determine number of channels for convolution.
-        kernel_size: Kernel size of conv layers.
+        size: Initial size to determine the number of channels.
+        kernel_size: Size of the convolving kernel.
         activation: Type of activation function.
         norm_class: Normalization module class.
         norm_args: Normalization module arguments.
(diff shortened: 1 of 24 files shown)


b-flo:
@pyf98 Could you review this PR when you have time please? I can't assign you it seems.

pyf98:
Thanks @b-flo , I'll read this PR.

b-flo:
Do not mind the changes from the last commit, it's only cosmetic (docs + typos). I needed to squeeze changes from ext. review into this branch to do some reorganization/rebase.

sw005320:
@pyf98, can you review it?

espnet2/asr_transducer/encoder/modules/convolution.py
```diff
@@ -107,3 +104,95 @@ def forward(
...
+        dropout_rate: Dropout rate.
+        causal: Whether to use causal convolution (set to True if streaming).
+
+    """
+
+    def __init__(
+        self,
+        size: int,
+        kernel_size: int,
+        activation: torch.nn.Module = torch.nn.ReLU(),
```
pyf98, line 127: For CSGU, I recommend to set the `activation` as `Identity()`. Based on previous experiments, a nonlinear activation does not improve the performance.

pyf98:
Thanks @b-flo! This PR looks good to me.

I haven't tested the code though. We may test Branchformer or its variant in the Transducer framework later.

espnet2/asr_transducer/encoder/modules/convolution.py
b-flo: Removed activation argument, I manually set `Identity()` for now.

b-flo:
I think I'll split `main_params` config into dedicated shared configs in the future (e.g.: for attention, norm, activation, etc). It'll be easier to set proper parameters while keeping flexibility and code clean. It'll be better for delegating some parts to each block type  for example.
Not a priority right now though.

> I haven't tested the code though. We may test Branchformer (or its variant) in the Transducer framework later.

I trained an offline and online model on Librispeech-100. I didn't have time/resources to tune it so I couldn't reach Conformer perf but it seemed to work correctly.

b-flo:
@sw005320 I'll merge this PR, is that OK for you?

sw005320:
LGTM!
Thanks, @b-flo!

PR #4593, opened 2022-08-23
[WIP] Update to fit the recent update in s3prl.

simpleoier:
This PR is to update the S3PRL frontend, because s3prl has some updates recently.
Reference: https://s3prl.github.io/s3prl/tutorial/upstream_collection.html


sw005320:
Cool!
@leo19941227, can you also review this code and give us some suggestions?

leo19941227:
Hi @simpleoier and @sw005320 ,

Sure, I am working on this

espnet2/asr/frontend/s3prl.py
```diff
@@ -36,75 +21,37 @@ def __init__(
         download_dir: str = None,
         multilayer_feature: bool = False,
```
leo19941227, line 22: It seems this argument is presented but not used in this class.

espnet2/asr/frontend/s3prl.py
```diff
@@ -36,75 +21,37 @@ def __init__(
...
         super().__init__()
+
         if isinstance(fs, str):
             fs = humanfriendly.parse_size(fs)
+        if fs != 16000:
+            logging.warning("All the upstream models in S3PRL now only support 16 kHz audio.")
 
+        # TODO(simpleoier): to remove download_dir and update related configs.
         if download_dir is not None:
             torch.hub.set_dir(download_dir)
```
leo19941227, line 42: Now we don't download checkpoint to the directory defined by torch.hub. We download to [./cache/s3prl/download](https://github.com/s3prl/s3prl/blob/20b0664a5f88de8a2c16c871b8e866731b7a9d19/s3prl/util/download.py#L18-L28). We also have an API (despite not yet on the documentation page, `s3prl.util.download.set_dir`) to change this directory if you wish to change it!

espnet2/asr/frontend/s3prl.py
```diff
@@ -36,75 +21,37 @@ def __init__(
...
-        else:
-            feature_selection = "last_hidden_state"
-        s3prl_featurizer = Featurizer(
-            upstream=s3prl_upstream,
-            feature_selection=feature_selection,
-            upstream_device="cpu",
-        )
-
-        return s3prl_upstream, s3prl_featurizer
+        self.hop_length = self.upstream.downsample_rates
```
leo19941227, line 53: This `self.upstream.downsample_rates` return a list (number of layers) of downsampling rates (e.g. [160, 160, 160, ...]) to explicitly defining every layer's downsampling rate. Also, `self.upstream.hidden_sizes` is similar (e.g. [768, 768, 768, ...]). This is designed to cope with future upstream with different internal strides or hidden sizes. Currently all upstream have the same stride/hidden_size for all layers, so you can simply take the first one.

leo19941227: Since Featurizer is used to reduce multiple hidden states into a single hidden state (single stride and hidden size), how about I add an `downsample_rate` attribute in Featurizer, so you know the dimension and the stride after the reduction?

leo19941227: Also, would it be better to explicitly make this attribute an public API like `self.hop_length()`? (to align with `self.output_size()`). This is just a random thought, feel free to adopt it or not.

espnet2/asr/frontend/s3prl.py
simpleoier: Yes, I found that in the s3prl code. I will delete this and the content in line 40-42 soon. But since this is a tentative PR and many existing configs still use download_dir, it's easier to pass the test and let others test the s3prl frontend in the meantime.

espnet2/asr/frontend/s3prl.py
simpleoier: Thanks for letting me know this. I think it is good to add downsample_rate in s3prl featurizer. Also, this hop_length variable seems not be used. I don't know why we defined here before. I'm trying to figure that out.

leo19941227: It seems to appear in `fused.py`, but I only went through that file very briefly.

leo19941227: Cool I am updating this new property to S3PRL master

espnet2/asr/frontend/s3prl.py
leo19941227: Oh, then maybe you can change `torch.hub.set_dir(download_dir)` to `s3prl.util.download.set_dir(download_dir)`?

simpleoier: OK, let me change to this. Thanks!

espnet2/asr/frontend/s3prl.py
simpleoier: Right, will delete this!

espnet2/asr/frontend/s3prl.py
```diff
@@ -36,75 +21,37 @@ def __init__(
...
+        self.tile_factor = 1
```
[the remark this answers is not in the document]
simpleoier: Any idea how I can get this tile_factor from upstream models? Maybe `self.tile_factor = getattr(self.upstream, "tile_factor", 1)`?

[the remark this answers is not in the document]
leo19941227: I think this is kind of an ESPNet feature, to easily align different representation to an assigned stride. This is not a property for a given upstream, but a post-processing step to tile an upstream's output representation by a factor of 2, or 3. I quote this part since I am not sure whether there are some recipe or pre-trained model depending on using tile_factor > 1.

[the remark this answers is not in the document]
simpleoier: OK, I found that. And I updated the code.

espnet2/asr/frontend/s3prl.py
leo19941227: Oh, I thought you might leave this argument to keep the backward compatibility. So `multilayer_feature=True` will pass the output of S3PRLUpstream (the entire `all_hidden_states`) into Featurizer for a weighted-sum, and `multilayer_feature=False` will simply feed the last item of `all_hidden_states` (`all_hidden_states[-1:]`) into Featurizer. Feel free to delete this argument if you can change this API directly.

PR #4596, opened 2022-08-24
Add improved results on FLEURS dataset

wanchichen:
Updated the FLEURS recipe to include improved results for multilingual and English ASR
Add a script to score language identification accuracy for FLEURS
Add a sym link for the default score_lang_id script to pyscripts

Uploaded Models:
https://huggingface.co/espnet/wanchichen_fleurs_multilingual_asr_hubert_frontend
https://huggingface.co/espnet/wanchichen_fleurs_english_asr_wav2vec_frontend


egs2/TEMPLATE/asr1/pyscripts/utils/score_lang_id.py
```diff
@@ -0,0 +1,60 @@
...
+# -*- coding: utf-8 -*-
+
+# Copyright 2021 Johns Hopkins University (Jiatong Shi)
+#  Apache 2.0  (http://www.apache.org/licenses/LICENSE-2.0)
+
+import argparse
+import codecs
+import sys
+
+
```
wanchichen, line 11: This should have been a sym link. I think creating a sym link in windows just copies the file over instead of creating an actual link. Will use a unix system to fix.

egs2/fleurs/asr1/README.md
```diff
@@ -1,5 +1,70 @@
...
+- date: `Sun Aug 21 15:18:30 EDT 2022`
+- python version: `3.8.6 (default, Dec 17 2020, 16:57:01)  [GCC 10.2.0]`
+- espnet version: `espnet 202205`
+- pytorch version: `pytorch 1.8.1+cu102`
+- Git hash: `45e8cb9173a072f85ee7a7ccbcae06af7c5c484a`
+  - Commit date: `Wed Jun 1 14:21:14 2022 +0900`
+- Pretrained Model: https://huggingface.co/espnet/wanchichen_fleurs_multilingual_asr_hubert_frontend
+
+### WER
+
```
sw005320, line 14: How about adding the language recognition accuracy here?

sw005320 (commented):
We usually use a symlink for cont/train_asr.yaml points to the best config.
Can you follow this?

Also, if we have to change the config between EN and multilingual setups, it is better to leave that comment in `run.sh`

egs2/fleurs/asr1/run.sh
```diff
@@ -46,14 +47,15 @@ if [[ "all" == *"${lang}"* ]]; then
       --valid_set "${train_dev}" \
       --test_sets "${test_set}" \
       --bpe_train_text "data/${train_set}/text" \
-      --lm_train_text "data/${train_set}/text" "$@" \
-      --local_score_opts "--score_lang_id true" "$@" 
+      --lm_train_text "data/${train_set}/text" \
+      --local_score_opts "--score_lang_id true --decode_folder decode_asr_asr_model_valid.acc.best" "$@"
```
ftshijt, line 51: I would not recommend to fix the decoder_folder here, since it can be automatically get from the scripts, can we use that instead? Otherwise, users may need multiple changes if they change any tags in the template.

wanchichen: That was how I originally planned to write it, but i would need to modify `asr.sh` so that it passes the decode folder to score.sh as argument. If that is ok with you, I can do that instead.

I just wasn't sure if that was a good change since we would start passing the decode folder as an argument for all asr recipes that otherwise would not use it.

egs2/fleurs/asr1/README.md
wanchichen: Since the best configs are different between multilingual and monolingual, which one would you like me to symlink to  `conf/train_asr.yaml` ?

sw005320: How about making English one as default?
Also, put both config information in `run.sh`
Then, we should not have confusions, I believe.

egs2/fleurs/asr1/run.sh
ftshijt: How about following the `open_li52` recipe, which just test all the possible decoding results? As in general the scoring is very fast, it won't be taking a long time?

sw005320:
Thanks a lot, @wanchichen!

PR #4598, opened 2022-08-24
Unfix numpy version

kamo-naoyuki:


kamo-naoyuki:
FYI:
Sometimes, we got the following error from numpy:

```
ValueError: numpy.ndarray size changed, may indicate binary incompatibility. Expected [  ] from C header, got [  ] from PyObject
```

This is caused by the incompatible NumPy version with the binary, such as `ctc_segmentation`;  For example, if you compiled `ctc_segmentation`  with `numpy=1.22.0`, and installed `numpy=1.21.0` after it, you could get the error.


The Numpy documentation (https://numpy.org/doc/stable/user/depending_on_numpy.html#understanding-numpy-s-versioning-and-api-abi-stability) says that

```
binaries compiled against a given version of NumPy will still run correctly with newer NumPy versions, but not with older versions.
```

Therefore, if we use the latest numpy always,  this issue doesn't happen. However, unfortunately, this is difficult because dependencies sometimes require a specific numpy version. In the current dependencies, Only Numba requires it, i.e. the latest Numba requires numpy<1.23.

Even though some modules require it, if we installed all modules by `pip install espnet`, all dependencies are resolved at this time, thus this error doesn't happen. However, our Makefile installs `pytorch` before installing `espnet`, and the latest numpy, which might be incompatible with our requirements, is also installed with pytorch.

Thus, we inserted `pip install numpy<=1.21.4` in the Makefile, but as I wrote, basically, we should use the latest numpy as possible. For this reason, I changed it to install `numba` instead of `numpy` in this PR.

kamo-naoyuki:
So far, I've written a strategy for espnet.

By the way, this error can be avoided by the developer of the module, and I think, originally the developer should care about it. I mean that, if the wheel file which is distributed in pypi is created with the old numpy, this error can be avoided.

However, there are no ways to upload the wheel file for a future Python  version, which doesn't exist now. Therefore, if some guy tries to install it with the new python, the module is installed using tgz file, i.e. the module is compiled at this time.
　
To avoid this problem completely, I think, it's better to prepare `pyproject.toml` and specify the numpy version for build time.
For example, `scipy` specifies the oldest numpy for each environment in the build system:  https://github.com/scipy/scipy/blob/main/pyproject.toml

Fortunately, the developers of scipy maintain `oldest-supported-numpy`, https://github.com/scipy/oldest-supported-numpy, it's easy to do this action.

In my understanding, this is the best practice, but this is not well known.

tools/Makefile
```diff
@@ -88,21 +88,29 @@ packaging.done: activate_python.sh
 	. ./activate_python.sh && python3 -m pip install packaging
 	touch packaging.done
 
-pytorch.done: activate_python.sh packaging.done
+numba.done: activate_python.sh
+ifeq ($(strip $(USE_CONDA)),)
+	. ./activate_python.sh && python3 -m pip install numba
+else
+	. ./activate_python.sh && conda install -y numba
```
pyf98, line 95: It seems that `conda` cannot automatically install a very new version of `numpy`. However, `fairscale` requires `numpy >= 1.22.0`, which will cause an incompatible issue eventually.

```
numba 0.55.1 requires numpy<1.22,>=1.18, but you have numpy 1.23.3 which is incompatible.
Successfully installed fairscale-0.4.9 numpy-1.23.3
```

Instead, `python3 -m pip install numba` seems to work well. It will install `numpy 1.23`.

kamo-naoyuki: I got it. Thank you.

PR #4617, opened 2022-09-03
Muskit installation

A-Quarter-Mile:
Hi @ftshijt, here are the new updates:

1. muskit install & check under tools/.
2. segmentation from XML in stage 1 (core codes only). In this stage, we keep both prep_segment from manually labeled file or from musicXML. Sub xmls will be generated under xml_dump/. Additionally, structured info(position) will be stored in xmlnote for further use.


tools/installers/install_muskit.sh
```diff
@@ -0,0 +1,46 @@
...
+
+if [ $# != 0 ]; then
+    echo "Usage: $0"
+    exit 1;
+fi
+
+rm -rf muskit.done
+
+# Install ParallelWaveGAN
+rm -rf ParallelWaveGAN
```
ftshijt, line 12: Probably better to check if there is already a parallel-wavegan installed (if already installed, print a message and skip removing it, we can always suggest to remove, but prefer not to remove without asking)? It might accidentally remove useful things ahead

egs2/TEMPLATE/svs1/pyscripts/utils/prep_segments_from_xml.py
```diff
@@ -0,0 +1,179 @@
...
+    )
+    parser.add_argument("scp", type=str, help="scp folder")
+    parser.add_argument(
+        "threshold", type=int, help="threshold for silence identification."
+    )
+    parser.add_argument(
+        "--silence", action="append", help="silence_phone", default=["pau"]
+    )
+    parser.add_argument(
+        "--xml_dump", type=str, default="xml_dump", help="xml dump directory")
```
ftshijt, line 89: we may simply follow the espnet dump directory convention. But you can use the current way for now, we can easily change it later

egs2/TEMPLATE/svs1/pyscripts/utils/prep_segments_from_xml.py
```diff
@@ -0,0 +1,179 @@
...
+        if self.start == self.end:
+            self.label_id = label_id
+            self.start = start
+            self.end = end
+        else:
+            self.end = end
+
+
+class SegInfo(object):
+    def __init__(self):
```
ftshijt, line 28: How about giving some examples to the split process?

ftshijt (commented):
Good start! some minor comments.

ftshijt:
For segmentation, we can just keep what we have discussed

egs2/TEMPLATE/svs1/pyscripts/utils/prep_segments_from_xml.py
ftshijt, line 28:
```suggestion
    def __init__(self):
    # TODO(Yuning): add examples
```

egs2/TEMPLATE/svs1/pyscripts/utils/prep_segments_from_xml.py
ftshijt, line 89:
```suggestion
        "--xml_dump", type=str, default="xml_dump", help="xml dump directory")
        # TODO (Yuning): update xml prepare scheme
```

PR #4621, opened 2022-09-05
recipe config update

ftshijt:
batch_bins does not work for folded batching, need to set to batch_size to be working


ftshijt:
Since it is a very minor update to recipe, I will directly merge it

PR #4627, opened 2022-09-08
Add python=3.10 to CI

kamo-naoyuki:


kamo-naoyuki:
@kan-bayashi Espnet depends on `nnmnkwii`(https://github.com/espnet/espnet/blob/master/setup.py#L67), but the requirement of nnmkkwii for numpy is not compatible with python=3.10 (https://github.com/espnet/espnet/actions/runs/4079849732/jobs/7031662783#step:6:200).

`nnmkkwii` is invoked only in the following script:
https://github.com/espnet/espnet/blob/08feae5bb93fa8f6dcba66760c8617a4b5e39d70/egs/jsut/tts1/local/prep_segments.py#L10

The function used here seems a simple text processing and can be implemented easily. I think nnmkkwii is too large dependency for this purpose because it includes even some cython codes.

I think rewriting it to a local function is better. How do you think?

kan-bayashi:
@r9y9 Is it possible to update nnmnkwii?

r9y9:
@kan-bayashi Let me take a look.

r9y9:
@kan-bayashi New release is up https://pypi.org/project/nnmnkwii/0.1.2/

kan-bayashi:
@r9y9 Thank you for blazing fast release! Super cool!

@kamo-naoyuki Could you try the new version?

kamo-naoyuki:
@r9y9 pyopenjtalk has same problem, could you have a look?

kan-bayashi:
It seems Ryuichi already fix the issue in v0.3.0 of pyopenjtalk.
I update the version in https://github.com/espnet/espnet/pull/4912

kan-bayashi:
pyopenjtalk seems to be fixed. But gensim has a problem in 3.10 :(

PR #4631, opened 2022-09-11
Updates on Muskit Migration 

A-Quarter-Mile:
Hi @ftshijt, here are some new updates:

- Add <breath mark> to prep_segment. Thus, sub xmls will align with segments generated from .lab file.
- Add xml_scp in stage 2, which takes place of midi_scp.


egs2/TEMPLATE/svs1/scripts/utils/copy_data_dir.sh
```diff
@@ -0,0 +1,142 @@
+#!/usr/bin/env bash
```
ftshijt, line 1: Another option is to still use Kaldi's one, but at the same time, setting utt_extra_files like https://github.com/espnet/espnet/blob/master/egs2/TEMPLATE/st1/st.sh#L609-L621

egs2/TEMPLATE/svs1/scripts/audio/format_scp.sh
```diff
@@ -29,6 +29,7 @@ EOF
 )
```
ftshijt, line 29: Given that we already have format_wav_scp.sh, how about we just initiate a format_musicxml_scp.sh directly. It could be more clear for those users adapting from other tasks

egs2/TEMPLATE/svs1/scripts/audio/format_scp.sh
```diff
@@ -57,6 +58,11 @@ if [ ! -f "${scp_dir}/wav.scp" ]; then
     echo "$0: Error: No such file: ${scp_dir}/wav.scp"
     exit 1
 fi
+if [ ! -d "xml_dump" ]; then
```
ftshijt, line 61: I would not recommend having a fixed directory here. Various reasons:
- It might be broken when you want to start multiple trials with the same notation (you cannot easily change the name)
- It increases the burden or understanding, as additional information from the recipe
- We already have the scp file, no need to fix the directory here

ftshijt (changes requested):

ftshijt:
You should also add shell TEMPLATE to here

egs2/TEMPLATE/svs1/scripts/audio/format_scp.sh
A-Quarter-Mile: For now, **format_wav_scp.py** and **format_xml_scp.sh** are both called in `scripts/audio/format_scp.sh` (just like what we have in Muskit). Would it be better if we split format_scp.sh into format_wav_scp.sh and format_xml_scp.sh under scripts/audio, or just maintain the status quo?

egs2/TEMPLATE/svs1/scripts/utils/copy_data_dir.sh
A-Quarter-Mile: Label and xml.scp are added into utt_extra_files in svs.sh.

egs2/TEMPLATE/svs1/scripts/audio/format_scp.sh
ftshijt: I would suggest separating them and using the format_wav_scp.{sh, py} from the ASR template. The major reason here is to take benefit from the timely update to those scripts if there are issues.

egs2/TEMPLATE/svs1/svs.sh
```diff
@@ -0,0 +1,1174 @@
...
+local_data_opts="" # Options to be passed to local/data.sh.
+
+# Feature extraction related
+feats_type=raw       # Feature type (fbank or stft or raw).
+audio_format=wav    # Audio format: wav, flac, wav.ark, flac.ark  (only in feats_type=raw).
+min_wav_duration=0.1 # Minimum duration in second.
+max_wav_duration=20  # Maximum duration in second.
+use_sid=false        # Whether to use speaker id as the inputs (Need utt2spk in data directory).
+use_lid=false        # Whether to use language id as the inputs (Need utt2lang in data directory).
+use_xvector=false    # Whether to use x-vector (Require Kaldi).
```
ftshijt, line 54: There is no need to require Kaldi now, you can refer to TTS template for a fix here

egs2/TEMPLATE/svs1/svs.sh
```diff
@@ -0,0 +1,1174 @@
...
+                scripts/feats/feat_to_shape.sh --nj "${_nj}" --cmd "${train_cmd}" \
+                    "${data_feats}${_suf}/${dset}/feats.scp" "${data_feats}${_suf}/${dset}/feats_shape"
+
+                # 4. Write feats_dim
+                head -n 1 "${data_feats}${_suf}/${dset}/feats_shape" | awk '{ print $2 }' \
+                    | cut -d, -f2 > ${data_feats}${_suf}/${dset}/feats_dim
+
+                # 5. Write feats_type
+                echo "${feats_type}" > "${data_feats}${_suf}/${dset}/feats_type"
+            done
```
ftshijt, line 412: I think we can safely remove this part. There is no need for us to maintain the Kaldi frontend anymore. We can just keep using our own feature extractor `raw`.

egs2/TEMPLATE/svs1/svs.sh
```diff
@@ -0,0 +1,1174 @@
...
+                    cut -f 2 -d " " "${data_feats}${_suf}/${dset}/utt2lang" | sort | uniq | \
+                        awk '{print $1 " " NR}' >> "${data_feats}${_suf}/${dset}/lang2lid"
+                fi
+                pyscripts/utils/utt2spk_to_utt2sid.py \
+                    "${data_feats}/org/${train_set}/lang2lid" \
+                    "${data_feats}${_suf}/${dset}/utt2lang" \
+                    > "${data_feats}${_suf}/${dset}/utt2lid"
+            done
+        fi
+    fi
```
ftshijt, line 466: I'm thinking of introducing a single new stage for those conditional token generation. You can keep it for now, but probably fixed it later after the first version.

egs2/TEMPLATE/svs1/svs.sh
```diff
@@ -0,0 +1,1174 @@
...
+                _max_length=$(python3 -c "print(int(${max_wav_duration} / ${_frame_shift} * 1000))")
+
+                cp "${data_feats}/org/${dset}/feats_dim" "${data_feats}/${dset}/feats_dim"
+                <"${data_feats}/org/${dset}/feats_shape" awk -F, ' { print $1 } ' \
+                    | awk -v min_length="${_min_length}" -v max_length="${_max_length}" \
+                        '{ if ($2 > min_length && $2 < max_length) print $0; }' \
+                        >"${data_feats}/${dset}/feats_shape"
+                <"${data_feats}/org/${dset}/feats.scp" \
+                    utils/filter_scp.pl "${data_feats}/${dset}/feats_shape"  \
+                    >"${data_feats}/${dset}/feats.scp"
```
ftshijt, line 517: If you remove the previous part of fbank_pitch, or fbank extraction, we can just remove this part as well. To make the TEMPLATE as simple as possible

egs2/TEMPLATE/svs1/svs.sh
```diff
@@ -0,0 +1,1174 @@
...
+            fi
+
+            # Remove empty text
+            <"${data_feats}/org/${dset}/text" \
+                awk ' { if( NF != 1 ) print $0; } ' >"${data_feats}/${dset}/text"
+
+            # fix_data_dir.sh leaves only utts which exist in all files
+            utils/fix_data_dir.sh "${data_feats}/${dset}"
```
ftshijt, line 525: Please add utt_extra_file to all `utils/fix_data_dir.sh` and `utils/copy_data_dir.sh` and `utils/validate_data_dir.sh`

egs2/TEMPLATE/svs1/scripts/audio/format_scp.sh
A-Quarter-Mile: format_scp.sh has been split into format_wav_scp.sh and format_xml_scp.sh.

egs2/TEMPLATE/svs1/scripts/audio/format_scp.sh
A-Quarter-Mile: Fix xml dir has been deleted from format_xml_scp.sh.

egs2/TEMPLATE/svs1/svs.sh
A-Quarter-Mile: TTS template keeps these opts. I haven't remove them in this commit.

egs2/TEMPLATE/svs1/svs.sh
A-Quarter-Mile: Removed.

egs2/TEMPLATE/svs1/svs.sh
A-Quarter-Mile: Keep it for this version temporarily.

egs2/TEMPLATE/svs1/svs.sh
A-Quarter-Mile: Removed.

egs2/TEMPLATE/svs1/svs.sh
A-Quarter-Mile: utt_extra_file has been added to utils/fix_data_dir.sh. However, we don't need to add it to utils/copy_data_dir.sh and utils/validate_data_dir.sh after check the result.

egs2/TEMPLATE/svs1/svs.sh
ftshijt: (You can add a TODO here for future reference)

egs2/TEMPLATE/svs1/svs.sh
ftshijt: I will update the TTS template shortly. 
```suggestion
use_xvector=false    # Whether to use x-vector 
```

egs2/TEMPLATE/svs1/svs.sh
ftshijt: After you add the TODO, I will merge the current PR

A-Quarter-Mile: Add TODO to it.

PR #4636, opened 2022-09-12
Add 2-pass SLU code for FSC Challenge

siddhu001:
Add 2-pass SLU code for FSC Challenge dataset. Also the code for running 2-pass SLU with ground truth transcript on SLURP dataset.


siddhu001:
@sw005320 @ftshijt This PR is ready to review. I have Add 2 pass SLU code for FSC challenge as well as uploaded all models of 2 pass SLU to hugging face.

sw005320:
LGTM in my side!
@ftshijt, can you review it?

egs2/fsc_challenge/slu1/local/score.sh
```diff
@@ -0,0 +1,53 @@
+#!/usr/bin/env bash
```
ftshijt, line 1: Given that we have a SLU-specified template, do you think it would be good to put the score.sh into `scripts/utils` and make it standard?

When I was using the SLU recipe, I found several times that, if just calling run.sh, usually there is no option to use some specific parameters (e.g., we will fix the 5best, but what if instead 5best, we want 10best). We need to open the score.sh -> understand the logic --> find corresponding directory --> call the score.sh again. This is sometimes confusing for newcomers.

It does not have to be settled in this PR. If this PR is in a hurry, we can firstly finish this one and leave that in the future.

egs2/slurp/slu1/run.sh
```diff
@@ -23,9 +27,10 @@ slu_config=conf/tuning/train_asr_bert_conformer_deliberation.yaml
     --max_wav_duration 30 \
     --feats_normalize utterance_mvn\
     --inference_nj 12 \
-    --pretrained_model ../../slurp_new/asr1/exp/asr_train_asr_conformer_raw_en_word/valid.acc.ave_10best.pth:encoder:encoder\
-    --inference_slu_model valid.acc.ave_10best.pth\
+    --nj 12\
+    --inference_slu_model 1epoch.pth\
```
ftshijt, line 31: typo?

egs2/slurp/slu1/run.sh
```diff
@@ -9,12 +9,16 @@ set -o pipefail
...
+local_data_opts="--gt false"
+# Make gt true to run using ground truth text as transcript
 
 slu_config=conf/tuning/train_asr_bert_conformer_deliberation.yaml
 
 ./slu.sh \
     --lang en \
     --ngpu 1 \
+    --stage 1 \
+    --stop_stage 1\
```
ftshijt, line 21: typo?

egs2/slurp/slu1/run.sh
siddhu001: Fixed! Thanks for catching this!

egs2/fsc_challenge/slu1/local/score.sh
siddhu001: I think this is a great point Jiatong! I agree that the current setup is confusing. I believe in slu.sh, this has been handled since it looks at the inference folder from the script itself (https://github.com/espnet/espnet/blob/a7bd6522b32ec6472c13f6a2289dcdff4a846c12/egs2/TEMPLATE/slu1/slu.sh#L1497). I will also ensure in a future PR that for all previous SLU recipes that use asr.sh also get inference folder from script itself so that score.sh does not need to be run seperately.

I have noted this and will open another PR after Interspeech for this. I will hopefully like to merge this PR before I leave for Interspeech.

ftshijt: Sounds great! Thanks for letting me know the status.

ftshijt: I can merge it after the CI test is fixed~

ftshijt:
Many thanks. Merged

PR #4647, opened 2022-09-16
Update the title of plots to be y-label vs x-label

pyf98:
This PR changes the title of a figure (accuracy, loss etc) to "y-axis label vs x-axis label", e.g., `acc vs epoch`, `loss vs epoch`. This should be a more common way of describing a figure.


kamo-naoyuki:
Thank you for pointing out it.

PR #4656, opened 2022-09-20
CI fix and some other minor recipe fixes

ftshijt:


ftshijt:
The remaining CI issues are from S3prl's dependency of Speechbrain. After my discussion with Leo, it seems that they are removing this dependency and will be fixed that by today, so I will pending the fix later.

pyf98:
Hi, do we want to remove this block in the installation doc?

https://github.com/espnet/espnet/blob/master/doc/installation.md?plain=1#L230

ftshijt:
Oh, yes, thanks for reminding

PR #4669, opened 2022-09-27
Add tests on MacOS and Windows (only installation)

kamo-naoyuki:


kamo-naoyuki:
This PR is hard because too many scripts don't work with macos or windows. I just limited this PR for the installation of espnet. Some scripts for installation also still don't work.

I'll add these OSs for CI test, however,  if you have a trouble with them when making new PRs, please add the exception to avoid the error, e.g.,

```sh
if [[ $(uname -s) =~ Linux ]]; then
   # Nothing to do
   exit 0
fi
```

Issue #4437, opened 2022-06-09
Development plan for ESPnet2 singing voice synthesis


A-Quarter-Mile:
We are now migrating [Muskit](https://github.com/SJTMusicTeam/Muskits), an open-source music processing toolkit, into ESPnet2.
Muskit focuses on benchmarking the end-to-end singing voice synthesis and expects to extend more tasks in the future. The main structure and base codes are adapted from ESPnet.

We also expect to make some new attempts in combination with the existing tasks (eg. TTS) under ESPnet2. We welcome your suggestions and contributions!

##  Code Merging

- [x] Merge modules from Muskit (mainly under the following two folders)
    - [x] tools/
    - [x] muskit/
 - [ ] Add authorship notes 

##  Networks

- [x] RNN-based non-autoregressive model
- [x] Xiaoice
- [ ] Sequence-to-sequence Transformer (with GLU-based encoder)
- [ ] MLP singer
- [x] Tacotron-singing 
- [ ] DiffSinger 
- [x] VISinger

## Recipes

- [ ] CSD
- [x] Itako
- [x] Kiritan
- [ ] KiSing
- [ ] Multilingual_four
- [ ] NIT_song070
- [ ] No7singing
- [x] Ofuton_p_utagoe_db
- [x] Oniku_kurumi_utagoe_db
- [x] Opencpop
- [x] PJS
- [ ] JSUT
- [ ] Ameboshi_ciphyer_utagoe_db

## Documentation

- [x] Installation 
- [x] Running instructions
- [x] Recipe explanation 
- [ ] pretrained_models 

## New Functions

- [x] Add musicXML in front-end
- [x] Add CI test
- [x] Upload to Huggingface

ftshijt:
Many thanks! Looks very exciting~

imiskolee:
oh, amazing!  did any processing now, which branch for it or not started yet?

ftshijt:
Thanks for your interest in our plan @imiskolee We do not have any open codebases published yet, most codebases are still under private development. Please stay tuned~

PR #4670, opened 2022-09-27
Singing Voice Synthesis Task for ESPnet

ftshijt:
This is the singing voice synthesis task for ESPnet, as noted in https://github.com/espnet/espnet/issues/4437

For now, we have supported a sample recipe and the corresponding setups (including RNN and RNN-DP models). 

Some major updates from original Muskits:
- We replace the raw music score from MIDI to MusicXML for easier usage in various cases
- We fixed some legacy issues with Muskits

It would be great if you could have some suggestions and comments.

We intend to distribute the work after this initial one. Therefore, we would probably add test functions in the following PR.


espnet2/train/preprocessor.py
```diff
@@ -944,3 +946,236 @@ def _speech_process(
...
+            midis = []
+            xml_timeseq = []
+            phn_cnt = 0
+            vowel = ["a", "i", "u", "e", "o"]
+            for i in range(len(syllables)):
+                # NOTE: Some phonemes are tagged differently
+                if syllables[i] == "へ":
+                    phn = ["h", "e"]
+                else:
+                    phn = pyopenjtalk.g2p(syllables[i]).split(" ")
```
ftshijt, line 1051: can we instead use `self.tokenizer` to do this part? The current procedure breaks the CI at https://github.com/espnet/espnet/actions/runs/3141052108/jobs/5103071721 . @A-Quarter-Mile

espnet2/bin/svs_inference.py
```diff
@@ -0,0 +1,532 @@
...
+from espnet2.torch_utils.device_funcs import to_device
+from espnet2.torch_utils.set_all_random_seed import set_all_random_seed
+
+from espnet2.svs.naive_rnn.naive_rnn import NaiveRNN
+from espnet2.svs.naive_rnn.naive_rnn_dp import NaiveRNNDP
+
+# from espnet2.svs.glu_transformer.glu_transformer import GLU_Transformer
+# from espnet2.svs.xiaoice.XiaoiceSing import XiaoiceSing
+# from espnet2.svs.xiaoice.XiaoiceSing import XiaoiceSing_noDP
+
```
sw005320, line 31: Maybe, we can remove these lines?

sw005320 (approved):
It looks good to me.
Since it is separated from the others at the task level, I don't think we would have some critical issues at the design level.

My general suggestion is to add a test (which is already mentioned that it will be done in the following PR) and document (e.g., main REAME.md and other places).

espnet2/bin/svs_inference.py
ftshijt: Sure. These are models to be added in future PRs. Will remove it for now

sw005320: If there are some intentions, you can keep these lines.

ftshijt:
The CI seems to be failed due to some `flack8` issues

ftshijt:
Please let us know if the PR looks good~ It would be great to merge for further distribution of works.

ftshijt:
As this is quite a dependency for following works, I would want to merge it for now~ There are plenty of TODOs to make it better. Please feel free to provide any comments if you have any~

PR #4689, opened 2022-10-04
VISinger

jerryuhoo:
Implement VISinger model


egs2/ofuton_p_utagoe_db/svs1/run.sh
```diff
@@ -14,6 +14,7 @@ n_shift=300
 win_length=1200
 
 score_feats_extract=frame_score_feats   # frame_score_feats | syllable_score_feats
```
ftshijt, line 16: For this specific case, we should use syllable_score_feats instead of frame_score_feats

espnet2/gan_svs/espnet_model.py
```diff
@@ -0,0 +1,1100 @@
...
+        Returns:
+            Dict[str, Any]:
+                - loss (Tensor): Loss scalar tensor.
+                - stats (Dict[str, float]): Statistics to be monitored.
+                - weight (Tensor): Weight tensor to summarize losses.
+                - optim_idx (int): Optimizer index (0 for G and 1 for D).
+
+        """
+        with autocast(False):
+            # Extract features
```
ftshijt, line 149: I think this part is very similar to svs.py, can we somehow pack the feature extract to a function and reuse the code?

espnet2/gan_svs/hifigan/__init__.py
```diff
@@ -0,0 +1,17 @@
+from espnet2.gan_tts.hifigan.hifigan import (
```
ftshijt, line 1: I think you can directly use the one from TTS by directly import them, so no need to re-define them here

espnet2/gan_svs/hifigan/hifigan.py
```diff
@@ -0,0 +1,745 @@
+# Copyright 2021 Tomoki Hayashi
+#  Apache 2.0  (http://www.apache.org/licenses/LICENSE-2.0)
+
+"""HiFi-GAN Modules.
```
ftshijt, line 4: Ditto

espnet2/gan_svs/hifigan/loss.py
```diff
@@ -0,0 +1,293 @@
+# Copyright 2021 Tomoki Hayashi
+#  Apache 2.0  (http://www.apache.org/licenses/LICENSE-2.0)
+
+"""HiFiGAN-related loss modules.
```
ftshijt, line 4: Ditto

espnet2/gan_svs/hifigan/residual_block.py
```diff
@@ -0,0 +1,97 @@
...
+This code is modified from https://github.com/kan-bayashi/ParallelWaveGAN.
+
+"""
+
+from typing import Any, Dict, List
+
+import torch
+
+
+class ResidualBlock(torch.nn.Module):
```
ftshijt, line 15: Ditto

espnet2/gan_svs/jets/alignments.py
```diff
@@ -0,0 +1,165 @@
...
+#  Apache 2.0  (http://www.apache.org/licenses/LICENSE-2.0)
+
+import numpy as np
+import torch
+import torch.nn as nn
+import torch.nn.functional as F
+from numba import jit
+
+
+class AlignmentModule(nn.Module):
```
ftshijt, line 11: Our expectation is that you do not copy everything in gan_tts but import them if do not need modification. We may have discussion later if the module needs some levels of changes.

ftshijt (changes requested):

espnet2/tasks/svs.py
```diff
@@ -106,6 +112,7 @@
         # mlp=MLPSinger,
         # xiaoice=XiaoiceSing,
         # xiaoice_noDP=XiaoiceSing_noDP,
+        vits=VITS,
```
ftshijt, line 115: no need for this?

jerryuhoo: Need this for inference.

ftshijt: I see, thanks!

jerryuhoo:
stage 8 (evaluation) causes an error because the names of ground truth wav files differ from generated files.

espnet2/tasks/svs.py
```diff
@@ -265,7 +272,8 @@ def build_preprocess_fn(
                 g2p_type=args.g2p,
                 fs=args.fs,
                 time_shift=args.feats_extract_conf["hop_length"]
-                / args.feats_extract_conf["fs"],
+                # / args.feats_extract_conf["fs"], # bug
```
ftshijt, line 275: Any clue for this one? We should have the argument in feats_extract

ftshijt (commented):
Many thanks for the update! The implementation is cool. I only have some minor comments. Please fix them, then we can get into the next stage. Good job!

egs2/TEMPLATE/svs1/svs.sh
```diff
@@ -710,19 +711,19 @@ if ! "${skip_train}"; then
                 
                 _opts+="--train_shape_file ${svs_stats_dir}/train/text_shape.${token_type} "
                 _opts+="--train_shape_file ${svs_stats_dir}/train/singing_shape "
-                _opts+="--train_shape_file ${svs_stats_dir}/train/durations_shape "
-                _opts+="--train_shape_file ${svs_stats_dir}/train/score_shape "
+                # _opts+="--train_shape_file ${svs_stats_dir}/train/durations_shape "
+                # _opts+="--train_shape_file ${svs_stats_dir}/train/score_shape "
```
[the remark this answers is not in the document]
jerryuhoo: Then it will cause error
```FileNotFoundError: [Errno 2] No such file or directory: 'exp/22k/svs_stats_raw_phn_None_zh/train/durations_shape'``` and also no score_shape

File: espnet2/bin/svs_inference.py [hunk 11]
```diff
@@ -231,7 +251,11 @@ def inference(
     key_file: Optional[str],
     train_config: Optional[str],
     model_file: Optional[str],
+    # maxlenratio: float, TODO: add maxlenratio
```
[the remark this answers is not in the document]
jerryuhoo: Is it necessary to add maxlenratio? Otherwise I can delete it.

espnet2/bin/svs_inference.py
```diff
@@ -260,6 +284,9 @@ def inference(
         train_config=train_config,
         model_file=model_file,
         use_teacher_forcing=use_teacher_forcing,
+        speed_control_alpha=speed_control_alpha,
```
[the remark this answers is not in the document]
jerryuhoo: removed

espnet2/bin/svs_inference.py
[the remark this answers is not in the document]
ftshijt: This would be helpful to auto-regressive modeling, but I don't think we need this at least for VITS. So probably we can just delete it now

espnet2/gan_svs/vits/duration_predictor.py
```diff
@@ -0,0 +1,244 @@
+# Copyright 2021 Tomoki Hayashi
+# Copyright 2022 Yifeng Yu
```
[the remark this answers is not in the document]
ftshijt: Sounds cool to me. We can just keep the VISinger related functions in this PR.

egs2/TEMPLATE/svs1/svs.sh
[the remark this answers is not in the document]
ftshijt: I see. This may indicate a bug in the stats collection. @A-Quarter-Mile Do you have similar issues here?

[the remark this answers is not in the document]
ftshijt: Let's just remove them for now (we are not using it anyway, if needed, we can easily add them back)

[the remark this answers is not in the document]
jerryuhoo: Should we add if condition since it could be used for other svs tasks?

espnet2/tasks/svs.py
ftshijt: Given that, we would have fs anyway, I do not think we need to call the divide here. @jerryuhoo could you reflect the calculation at https://github.com/espnet/espnet/blob/master/espnet2/train/preprocessor.py#L994-L995 and remove the line here ?

egs2/TEMPLATE/svs1/svs.sh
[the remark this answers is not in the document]
ftshijt: I don't think so, let's simply remove it

ftshijt:
After you fixes the two issues pointed out, I will merge this PR and we can work on the one that merges into the master branch.

PR #4701, opened 2022-10-09
Add espnet2 TTS recipe on M-AILABS

Takaaki-Saeki:
This PR adds a TTS recipe on [M-AILABS](https://www.caito.de/2019/01/03/the-m-ailabs-speech-dataset/).
TTS recipes on M-AILABS was supported in espnet1 but not supported in espnet2.

This recipe only uses a en_US single-speaker data.
I have uploaded a pretrained Tacotron2 model to [HuggingFace](https://huggingface.co/saefro991/m-ailabs_en-us_judy_phn_tacotron2).

I also prepared a script for multilingual data preparation: `local/data_multilingual.sh` to train a TTS model on the whole dataset.


ftshijt:
Looks great! Some minor suggestion: given that the `local/data.sh` is automatically called in the TEMPLATE, how about combine data.sh and data_multilingual.sh but with an additional augment? In that way, we can change the config simply by change the `--local_data_opts`

In addition, could you also update the entry in `egs2/TEMPLATE/README.md` ?

Takaaki-Saeki:
Thanks for the suggestion!
I have merged the en_us and multilingual recipes. 
Also updated the readme.

Takaaki-Saeki:
Hmm... it seems that some tests regarding `espnet2/slu/test_transcript_espnet_model.py` are not successful.

sw005320:
> Hmm... it seems that some tests regarding `espnet2/slu/test_transcript_espnet_model.py` are not successful.

@siddhu001, could you check this?
some interfaces are changed?

siddhu001:
@sw005320 It seems to be problem with hugging face.
```
E       TypeError: register_buffer() got an unexpected keyword argument 'persistent'
tools/venv/lib/python3.7/site-packages/transformers/models/bert/modeling_bert.py:198: TypeError
```
I think it is because of some huggingface version update. I am looking into this. @Takaaki-Saeki if you have any suggestions, please let me know!

ftshijt:
Many thanks! Merged.

PR #4714, opened 2022-10-18
Update conv_input.py

aky15:
The subsampling_factor should be 6 instead of 4 as the two pooling strides along the time axis are 3 and 2.


b-flo (approved):
Hi,

Good catch. I'm okay with these changes as it will also fix the required minimum length for last chunk during streaming process (which is based on the subsampling factor).

@sw005320 We can merge this now. I'll also open a PR to do further changes in `ConvInput.`

b-flo:
I'm closing this PR as it's superseded by #4720. Thanks a lot for pointing out the issue!

Issue #4716, opened 2022-10-18
Error in training enahncement model when using a different model config file


bunny-sung:
Hi,

I am running experiment derivied from the egs2/librimix/enh_diar1 and egs2/wsj0_2mix/enh1 directory. There is no error when I was using the default `conf/tuning/train_enh_conv_tasnet.yaml` and `train_enh_dprnn_tasnet.yaml`.

However when I switched to DPT network: `conf/tuning/train_enh_dptnet.yaml` in the `run.sh` for wsj0_2mix, I got this error at the training stage:

```
<DIR>/tools/anaconda/envs/espnet_enh_2022/bin/python3 <DIR>/espnet2/bin/enh_train.py --collect_stats true --train_data_path_and_name_and_type dump/raw/train/wav.scp,speech_mix,sound --train_data_path_and_name_and_type dump/raw/train/spk1.scp,speech_ref1,sound --train_data_path_and_name_and_type dump/raw/train/spk2.scp,speech_ref2,sound --valid_data_path_and_name_and_type dump/raw/dev/wav.scp,speech_mix,sound --valid_data_path_and_name_and_type dump/raw/dev/spk1.scp,speech_ref1,sound --valid_data_path_and_name_and_type dump/raw/dev/spk2.scp,speech_ref2,sound --train_shape_file exp_dptnet/enh_stats_8k/logdir/train.9.scp --valid_shape_file exp_dptnet/enh_stats_8k/logdir/valid.9.scp --output_dir exp_dptnet/enh_stats_8k/logdir/stats.9 --config conf/tuning/train_enh_dptnet.yaml
Traceback (most recent call last):
  File "<DIR>/tools/anaconda/envs/espnet_enh_2022/lib/python3.8/runpy.py", line 194, in _run_module_as_main
    return _run_code(code, main_globals, None,
  File "<DIR>/tools/anaconda/envs/espnet_enh_2022/lib/python3.8/runpy.py", line 87, in _run_code
    exec(code, run_globals)
  File "<DIR>/espnet2/bin/enh_train.py", line 23, in <module>
    main()
  File "<DIR>/espnet2/bin/enh_train.py", line 19, in main
    EnhancementTask.main(cmd=cmd)
  File "<DIR>/espnet2/tasks/abs_task.py", line 1019, in main
    cls.main_worker(args)
  File "<DIR>/espnet2/tasks/abs_task.py", line 1121, in main_worker
    model = cls.build_model(args=args)
[...]

PR #4719, opened 2022-10-19
Update outdated enh config files

Emrys365:
Reminded by https://github.com/espnet/espnet/issues/4716, I checked through all existing enh config files in `egs2/*/enh1/conf/tuning/*.yaml` and found three outdated config files that cannot be used with the latest code.
This PR fixes these config files.


sw005320:
Thanks, @Emrys365!

PR #4720, opened 2022-10-19
Update ConvInput for Transducer

b-flo:
This is a follow up for #4714. Basically:

- Support subsampling factor of 4 and 6 for VGG
- Create boolean mask according to output size only.

About the second point: basically, we don't need to ensure the "correct" items are removed according to the convolution/max pooling operations. Padding values will be removed anyway and subsequent mask for streaming is created from the output of ConvInput.


b-flo:
@sw005320 There seems to be an issue with HF and some tests. Can I skip them and manually merge this PR? There is no need for reviews.

sw005320:
OK!

@siddhu001 ,
I know that you're busy, but could I ask you to take a look again?
If you can fix it easily, can you do that?

PR #4736, opened 2022-10-28
add src_sos & src_eos for mt task to address the index out of range w…

simpleoier:
add src_sos & src_eos for mt task to address the index out of range when src and tgt vocab are different


ftshijt:
I think similar problem could happen for the ST as well. Could you also reflect your changes that?

simpleoier:
> I think similar problem could happen for the ST as well. Could you also reflect your changes that?

@ftshijt Lol, this is from the same setting as in [st espnet model](https://github.com/espnet/espnet/blob/0ec9c9239bc780f57e78511d1dad140a3ac13929/espnet2/st/espnet_model.py#L77)

sw005320:
I see. Then, I'll merge this.

sw005320:
Thanks, @simpleoier!

PR #4746, opened 2022-11-01
add fleurs conformer+sc-ctc results

wanchichen:
Add Conformer + self-conditioned ctc results for FLEURS
Trained model: https://huggingface.co/espnet/wanchichen_fleurs_asr_conformer_scctc
Note: the trained checkpoint uses the older version of s3prl w/ the fairseq dependency 

The models that condition on LID use the newer version of s3prl. Will add in a separate PR with the code changes necessary to run them.


sw005320:
You can replace it as a default config.

sw005320:
Thanks a lot!
Let's use this pre-trained models as our base!!!

PR #4751, opened 2022-11-03
chunk_mask error

aky15:
in [espnet2/asr_transducer/encoder/modules/attention.py](https://github.com/espnet/espnet/blob/master/espnet2/asr_transducer/encoder/modules/attention.py),
masked_fill fills elements of tensor with 0 where mask is True, so the value of the mask should be True if chunk_mask is true or (instead of and) source mask is true.


b-flo (commented):
Hi,

I'm really sorry for the late answer, I was away from the project for a few weeks.

I'm okay with the fix, thanks a lot!
Some mistakes slipped in the current version, I'm preparing a patch as some of them are quite critical.

sw005320:
Thanks, @aky15 for your fix!
Thanks, @b-flo for your check!

PR #4769, opened 2022-11-16
Shell scripts for UASR processing

ftshijt:
Validated until stage 10


ftshijt:
I've test the functions until stage 12. For further processing, we need corresponding pytorch modeuls, which will be handled by Dongji

ftshijt:
Since it is a separate branch, I will first merge it

PR #4718, opened 2022-10-19
Add g2pk_explicit_space tokenizer

jonghwanhyeon:
This pull request adds `g2pk_explicit_space` `PhonemeTokenizer` to ESPNet.

Current `g2pk` tokenizer use a whitespace (`" "`) as word separators. But from what I've observed, ESPNet seems to ignore whitespace tokens when processing data.

Let's see the following example taken from `egs2/kss/tts1`:
```python
from espnet2.bin.tts_inference import Text2Speech
from espnet2.text.phoneme_tokenizer import PhonemeTokenizer

model = Text2Speech.from_pretrained("imdanboy/kss_tts_train_jets_raw_phn_null_g2pk_train.total_count.ave")

text = "안녕하세요. 세계입니다."

tokenizer = PhonemeTokenizer(model.train_args.g2p)
tokens_from_tokenizer = tokenizer.text2tokens(text)
print(tokens_from_tokenizer)
# ['ᄋ', 'ᅡ', 'ᆫ', 'ᄂ', 'ᅧ', 'ᆼ', 'ᄒ', 'ᅡ', 'ᄉ', 'ᅦ', 'ᄋ', 'ᅭ', '.', ' ', 'ᄉ', 'ᅦ', 'ᄀ', 'ᅨ', 'ᄋ', 'ᅵ', 'ᆷ', 'ᄂ', 'ᅵ', 'ᄃ', 'ᅡ', '.']

token_ids = model.preprocess_fn('<dummy>', data={"text": "안녕하세요. 세계입니다."})["text"]
tokens_from_preprocessor = [model.train_args.token_list[token_id] for token_id in token_ids]
print(tokens_from_preprocessor)
# ['ᄋ', 'ᅡ', 'ᆫ', 'ᄂ', 'ᅧ', 'ᆼ', 'ᄒ', 'ᅡ', 'ᄉ', 'ᅦ', 'ᄋ', 'ᅭ', '.', '<unk>', 'ᄉ', 'ᅦ', 'ᄀ', 'ᅨ', 'ᄋ', 'ᅵ', 'ᆷ', 'ᄂ', 'ᅵ', 'ᄃ', 'ᅡ', '.']

print(' ' in model.train_args.token_list) # False
```

As you can see, `g2pk` from `PhonemeTokenizer` separates words with whitespaces. However, `preprocess_fn` from the trained model treats whitespaces as `<unk>` token because there is no whitespace token in the model's `token_list`.

[...]

PR #4778, opened 2022-11-22
fix parsing token_list

imdanboy:
As reported in #4718,  whitespace```(' ')``` (word separator token in g2p_en, g2pk) is not parsed correctly. 

```token_list = [line.rstrip() for line in f]```
Currently when parsing each tokens from the line, ```str.rstrip``` removes not only ```'\n'``` but also whitespace token ```' '```

Therefore I suggest following method for parsing token_list
```token_list = [line[0] + line[1:].rstrip() for line in f]```


sw005320:
@kan-bayashi, could you check this PR?

kan-bayashi (approved):
Sorry for late reviewing. LGTM!

sw005320:
Thanks, both!
After the CI check, I'll merge it.

PR #4739, opened 2022-10-31
Adding OpenAI Whisper encoder to ASR frontend

slSeanWU:
Hi all,

Here are some results I got with the config associated with this PR
- `Librispeech`

    | WER | dev-clean  | test-clean | dev-other | test-other |
    |:-----: |:-----: | :-----: |:-----: |:-----: |
    | Whisper medium (250M params) | 5.1 | 5.2 | 10.5 | 10.7 |
    | WavLM large (320M params), single-layer feat | 3.7 | 3.6 | 6.6 | 6.7 |
- `CHiME-4`

  | WER | dt05-1ch-real | dt05-1ch-simu | et05-1ch-real | et05-1ch-simu |
  |:-----: |:-----: | :-----: |:-----: |:-----: |
  | Whisper medium (250M params) | 7.3 | 10.6 | 10.8 | 14.0 |
  | WavLM large (320M params), single-layer feat | 7.7 | 9.4 | 7.9 | 11.9 |

PR #4793, opened 2022-11-28
Add Full Whisper Model for Finetuning

slSeanWU:
Here are some results with the current implementation. Note that we should set `ctc_weight=0.0` during decoding
image


espnet2/asr/espnet_model.py
```diff
@@ -60,16 +60,28 @@ def __init__(
         sym_space: str = "<space>",
         sym_blank: str = "<blank>",
         extract_feats_in_collect_stats: bool = True,
+        sos_index: int = -1,
```
simpleoier, line 63: Can you also use `sym_sos` defined in the latest [espnet](https://github.com/espnet/espnet/blob/ca2193d54031eb469b5a02cf60c6ff0b185d7ca4/espnet2/asr/espnet_model.py#L64) to index the sos and eos symbols?

espnet2/asr/espnet_model.py
```diff
@@ -366,11 +383,13 @@ def _extract_feats(
         # for data-parallel
         speech = speech[:, : speech_lengths.max()]
 
-        if self.frontend is not None:
+        if self.frontend is not None and not self.is_encoder_whisper:
```
simpleoier, line 386: I think it is enough to just check whether the frontend is None. But you may add an assertion to check if whisper encoder and frontend exist at the same time in the model definition.

espnet2/asr/espnet_model.py
```diff
@@ -60,16 +60,28 @@ def __init__(
...
+        sos_index: int = -1,
+        eos_index: int = -1,
     ):
         assert check_argument_types()
         assert 0.0 <= ctc_weight <= 1.0, ctc_weight
         assert 0.0 <= interctc_weight < 1.0, interctc_weight
 
         super().__init__()
-        # note that eos is the same as sos (equivalent ID)
+        # NOTE (Shih-Lun): eos is the same as sos in ESPNet (equivalent ID)
```
simpleoier, line 71: ditto.

egs2/librispeech_100/asr1/conf/tuning/train_asr_whisper_full.yaml
```diff
@@ -0,0 +1,75 @@
+normalize: null
+
+encoder: whisper
+encoder_conf:
+    whisper_model: small
+    dropout_rate: 0.0
+    use_specaug: true
+    specaug_conf:
```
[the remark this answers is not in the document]
slSeanWU: Yes. This is mainly because `OpenAIWhisperEncoder()` uses its own waveform --> log-mel spectrum function. Thus, the specaug should be wrapped inside.

espnet/nets/e2e_asr_common.py
```diff
@@ -206,6 +215,11 @@ def convert_to_char(self, ys_hat, ys_pad):
             seq_hat_text = "".join(seq_hat).replace(self.space, " ")
             seq_hat_text = seq_hat_text.replace(self.blank, "")
             seq_true_text = "".join(seq_true).replace(self.space, " ")
+
+            if self.use_whisper_vocab:
```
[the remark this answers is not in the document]
slSeanWU: Sure, I think it's doable

espnet2/asr/espnet_model.py
slSeanWU: Ok! this makes sense

egs2/TEMPLATE/asr1/asr.sh
```diff
@@ -1329,6 +1343,7 @@ if ! "${skip_eval}"; then
                                   --token_type word \
                                   --non_linguistic_symbols "${nlsyms_txt}" \
                                   --remove_non_linguistic_symbols true \
+                                  --cleaner "${cleaner}" \
```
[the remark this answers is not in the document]
simpleoier: Yeah, I know. But this might affect other usecases.

egs2/TEMPLATE/asr1/asr.sh
```diff
@@ -318,6 +318,12 @@ elif [ "${token_type}" = char ]; then
 elif [ "${token_type}" = word ]; then
     token_list="${wordtoken_list}"
     bpemodel=none
+elif [ "${token_type}" = whisper_en ]; then # should make token_list an output filepath here
```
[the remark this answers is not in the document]
simpleoier: I knew it. I was suggesting to make this part a unified interface, similar to the `asr_task` used [here](https://github.com/espnet/espnet/blob/ca2193d54031eb469b5a02cf60c6ff0b185d7ca4/egs2/TEMPLATE/asr1/asr.sh#L998).

espnet2/asr/frontend/whisper.py
```diff
@@ -0,0 +1,158 @@
...
+class WhisperFrontend(AbsFrontend):
+    """Speech Representation Using Encoder Outputs from OpenAI's Whisper Model:
+
+    URL: https://github.com/openai/whisper
+    """
+
+    def __init__(
+        self,
+        fs: Union[int, str] = 16000,
+        n_fft: int = 400,
```
simpleoier, line 21: Is it possible to get the log-mel-filterbank paramters from the pretrained whisper model. If so, it is not necessary to use this arguments. We can use the as the default values.

sw005320:
@slSeanWU, I want to make sure.
Does this PR include #4739?
If so, I'll close #4739

slSeanWU:
> @slSeanWU, I want to make sure. Does this PR include #4739? If so, I'll close #4739

Yes, it does. Thanks!

sw005320:
Any update?

RuABraun:
Hi @slSeanWU I'm trying this PR while reiniting the token embedding, have issues with NaN losses, have you maybe dealt with this issue before?

slSeanWU:
> Hi @slSeanWU I'm trying this PR while reiniting the token embedding, have issues with NaN losses, have you maybe dealt with this issue before?

Yeah, that tends to happen in the first epoch. Doesn't ESPnet just ignore the NaN gradient steps by default?

RuABraun:
Yeah but the loss diverges.

I have sort_batch="descending" (as in your config), is that important?

Also I'm confused why is a frontend necessary since whisper model creates the logmel internally?

slSeanWU:
> Yeah but the loss diverges.
> 
> I have sort_batch="descending" (as in your config), is that important?
> 
> Also I'm confused why is a frontend necessary since whisper model creates the logmel internally?

Oh, I see. If you're using the entire Whisper encoder + decoder, the code will actually skip frontend operation.
We'll add assertions to make sure Whisper frontend and encoder is not used at the same time.

Not sure why your loss diverges, but I feel the batching thing won't matter too much.
Have you tried smaller learning rates or increasing grad accumulation steps?

sw005320:
@slSeanWU, any update?
Please fix the conflicts and reflect the comments.

slSeanWU:
@sw005320 @simpleoier I've addressed all comments above (except the one about unified interface for external tokenizers, which I think needs more discussion), and fixed the conflicts. Please take a look, thanks!

RuABraun:
For me the issue turned out to be mixed precision.

sw005320:
@slSeanWU, can you restart this PR?
This can be used for many projects.

slSeanWU:
@sw005320 I've added tests that cover all new python files in this PR.

For the config/results part, should I just put configs in `chime4` & `librispeech_100h` recipe folders, and update their READMEs?

egs2/chime4/asr1/README.md
```diff
@@ -45,3 +45,41 @@
...
+
+
+<!-- Generated by scripts/utils/show_asr_result.sh -->
+# RESULTS
+## Environments
+- date: `Tue Jan 10 04:15:30 CST 2023`
+- python version: `3.9.13 (main, Aug 25 2022, 23:26:10)  [GCC 11.2.0]`
+- espnet version: `espnet 202211`
+- pytorch version: `pytorch 1.12.1`
+- Git hash: `d89be931dcc8f61437ac49cbe39a773f2054c50c`
```
sw005320, line 57: Can you add a model link?

egs2/librispeech_100/asr1/README.md
```diff
@@ -100,3 +100,32 @@ Model: https://huggingface.co/pyf98/librispeech_100h_transformer
...
 |beam20_ctc0.3/test_clean|2620|66983|90.0|6.2|3.9|0.8|10.9|63.3|
 |beam20_ctc0.3/test_other|2939|66650|77.9|15.2|6.9|2.5|24.6|84.8|
+
+
+# RESULTS
+## Environments
+- date: `Mon Jan  9 23:06:34 CST 2023`
+- python version: `3.9.13 (main, Aug 25 2022, 23:26:10)  [GCC 11.2.0]`
+- espnet version: `espnet 202211`
+- pytorch version: `pytorch 1.12.1`
```
sw005320, line 110: ditto

espnet/nets/e2e_asr_common.py
```diff
@@ -120,7 +120,10 @@ def __init__(
         self.char_list = char_list
         self.space = sym_space
         self.blank = sym_blank
-        self.idx_blank = self.char_list.index(self.blank)
+        if self.blank in self.char_list:
+            self.idx_blank = self.char_list.index(self.blank)
+        else:
```
sw005320, line 125: Can you add a comment that this is to deal with the whisper model case?

espnet2/asr/espnet_model.py
```diff
@@ -71,7 +71,10 @@ def __init__(
         assert 0.0 <= interctc_weight < 1.0, interctc_weight
 
         super().__init__()
-        self.blank_id = token_list.index(sym_blank)
+        if sym_blank in token_list:
```
sw005320, line 74: ditto

Can you add a comment that this deals with the whisper model case?

sw005320 (commented):
LGTM
I left some minor comments.
Did you reflect @simpleoier's comments?

Once you reflect those, I'll merge them soon.

slSeanWU:
Thanks a lot, and yes, I addressed all of @simpleoier's comments except for the one on unified tokenizer interface, which I think should be dealt with in a separate PR.

RuABraun:
I used the installer for whisper and export vocab has a size of 50363 but token embedding has a size of Embedding(51864, 768) ?

slSeanWU:
> I used the installer for whisper and export vocab has a size of 50363 but token embedding has a size of Embedding(51864, 768) ?

Hi Rudolf,

The remaining tokens (idx 50363~51863) are [timestamp markers](https://github.com/openai/whisper/blob/main/whisper/tokenizer.py#L143) that don't appear in the vocab object.

sw005320:
Can you fix the CI error?

RuABraun:
> > I used the installer for whisper and export vocab has a size of 50363 but token embedding has a size of Embedding(51864, 768) ?
> 
> Hi Rudolf,
> 
> The remaining tokens (idx 50363~51863) are [timestamp markers](https://github.com/openai/whisper/blob/main/whisper/tokenizer.py#L143) that don't appear in the vocab object.

Makes sense thank you for explaining. However in the whisper decoder the vocab size is used as a check for whether to reinitialize the token embredding or not, seems to me it will always reinitialize?

slSeanWU:
> Can you fix the CI error?

It's failing due to that the current CI script doesn't install Whisper, so the tests I wrote run into errors. @simpleoier suggested creating new CI testcases, but I don't know if there's anything I can do.

slSeanWU:
> > > I used the installer for whisper and export vocab has a size of 50363 but token embedding has a size of Embedding(51864, 768) ?
> > 
> > 
> > Hi Rudolf,
> > The remaining tokens (idx 50363~51863) are [timestamp markers](https://github.com/openai/whisper/blob/main/whisper/tokenizer.py#L143) that don't appear in the vocab object.
> 
> Makes sense thank you for explaining. However in the whisper decoder the vocab size is used as a check for whether to reinitialize the token embredding or not, seems to me it will always reinitialize due to the mismatch (since the vocab size is determined by the len of token list which is created by the export)?

Hi, the current token list export [script](https://github.com/espnet/espnet/blob/89d97c794ad0139241f6374b0667ed1adee7411e/espnet2/bin/whisper_export_vocabulary.py#L57) (run in `asr.sh` stage 5) will ensure the list has either 51864 (for `en`) or 51865 (for `multilingual`) tokens. So embeddings shouldn't get reinitialized.

sw005320:
I'd recommend you check the other PRs.
You could refer to the longformer case https://github.com/espnet/espnet/search?q=longformer

RuABraun:
> > > > I used the installer for whisper and export vocab has a size of 50363 but token embedding has a size of Embedding(51864, 768) ?
> > > 
> > > 
> > > Hi Rudolf,
> > > The remaining tokens (idx 50363~51863) are [timestamp markers](https://github.com/openai/whisper/blob/main/whisper/tokenizer.py#L143) that don't appear in the vocab object.
> > 
> > 
> > Makes sense thank you for explaining. However in the whisper decoder the vocab size is used as a check for whether to reinitialize the token embredding or not, seems to me it will always reinitialize due to the mismatch (since the vocab size is determined by the len of token list which is created by the export)?
> 
> Hi, the current token list export [script](https://github.com/espnet/espnet/blob/89d97c794ad0139241f6374b0667ed1adee7411e/espnet2/bin/whisper_export_vocabulary.py#L57) (run in `asr.sh` stage 5) will ensure the list has either 51864 (for `en`) or 51865 (for `multilingual`) tokens. So embeddings shouldn't get reinitialized.

Thanks! Turns out I wasn't on the latest commit

sw005320:
Instead of ignoring the import during the test, please include it... (otherwise, this test becomes useless)

- For example, please check`longformer` and `s3prl` First, you put the whisper installation in the extra modules (see https://github.com/espnet/espnet/blob/master/tools/Makefile#L32).
- Then, add it to the module list in https://github.com/espnet/espnet/blob/master/tools/check_install.py#L15 to include the whisper installation in the installation check
- Also, add the whisper installation to https://github.com/espnet/espnet/blob/master/ci/install.sh#L23
- Finally, please add an import error handling in `path.sh`, e.g., https://github.com/espnet/espnet/blob/99f04d5428d45260d22eba903c2c4f6c27369899/egs2/tsukuyomi/tts1/local/path.sh#L2

This (importing an external library) is not a typical case and we do not have documentations. Instead, please check the other libraries (again longformer, s3prl, pyopenjtalk, etc.) and follow their way.

slSeanWU:
> Instead of ignoring the import during the test, please include it... (otherwise, this test becomes useless)
> 
> * For example, please check`longformer` and `s3prl` First, you put the whisper installation in the extra modules (see https://github.com/espnet/espnet/blob/master/tools/Makefile#L32).
> * Then, add it to the module list in https://github.com/espnet/espnet/blob/master/tools/check_install.py#L15 to include the whisper installation in the installation check
> * Also, add the whisper installation to https://github.com/espnet/espnet/blob/master/ci/install.sh#L23
> * Finally, please add an import error handling in `path.sh`, e.g., https://github.com/espnet/espnet/blob/99f04d5428d45260d22eba903c2c4f6c27369899/egs2/tsukuyomi/tts1/local/path.sh#L2
> 
> This (importing an external library) is not a typical case and we do not have documentations. Instead, please check the other libraries (again longformer, s3prl, pyopenjtalk, etc.) and follow their way.

Thanks for the detailed guide. I've included the tests and the CI is passing. Sorry that I didn't get it at first when you mentioned the `longformer` case.

sw005320:
Thanks a lot, @slSeanWU!
This is an outstanding contribution!

As a follow-up PR, can you add a brief description of this whisper model to https://github.com/espnet/espnet/blob/master/README.md?
I want to make your contribution more visible.

slSeanWU:
> Thanks a lot, @slSeanWU! This is an outstanding contribution!
> 
> As a follow-up PR, can you add a brief description of this whisper model to https://github.com/espnet/espnet/blob/master/README.md? I want to make your contribution more visible.

Sure thanks!

geoffvdr:
Hi @slSeanWU, would it be possible to share a `run.sh` example script for how to do zero-shot evaluation? Like the one you used for Whisper small full zero-shot on Librispeech100 for example. Thanks

lalimili6:
@slSeanWU, Could you mind sharing `run.sh` for an example code?
best regard

lalimili6:
Can add tensorboard? 
best regards

slSeanWU:
> Hi @slSeanWU, would it be possible to share a `run.sh` example script for how to do zero-shot evaluation? Like the one you used for Whisper small full zero-shot on Librispeech100 for example. Thanks

Hi @geoffvdr. Apologies for the late reply. If you haven't sorted out zero-shot evaluation, I just modify the config yaml with:
```
max_epoch: 1
optim_conf:
    lr: 0.0
```
to run a dry epoch and then it'll evaluate after that.

slSeanWU:
> Can add tensorboard? best regards

Hi @lalimili6, could you provide a bit more context? Thanks a lot!

lalimili6:
Hi, I want you to show me the training curves.

PR #4794, opened 2022-11-28
Add MEDIA and PortMEDIA results for XLS-R with mBART-50

akreal:


sw005320 (commented):
LGTM!

My only comment is that we can make this script as default, e.g.,
`run-hugging-face.sh` —> `run.sh`

You can change all of your PRs.
If there is an existing `run.sh` and you want to keep both, you may move/rename yours  as `local/run_hf.sh` And  adding a usage of `local/run_hf.sh` in `README.md`

PR #4799, opened 2022-12-02
Fix an audio format issue in some enh recipes

YoshikiMas:
I added `--audio_format wav` option to `run.sh` for WHAM!, WHAMR! and SMS-WSJ, based on discussion with @Emrys365. Since their data generation scripts use `32-bit Floating Point PCM`, the conversion to `16-bit Signed Integer PCM` in the stage 3 does not work well (in detail, [this part](https://github.com/espnet/espnet/blob/cb06bb1a9e5e5355a02d8c1871a8ecfafd54754d/egs2/TEMPLATE/asr1/pyscripts/audio/format_wav_scp.py#L186)). In addition, I updated outdated config files in the three recipes.

Note that I only ran the recipe for WHAMR!.


Emrys365 (approved):
LGTM. Thanks a lot @YoshikiMas!

sw005320:
Thanks, @YoshikiMas!

PR #4557, opened 2022-08-04
Adding MFA scripts

iamanigeeit:
From https://github.com/espnet/espnet/issues/4521 : this commit adds scripts necessary to convert MFA to ESPnet formats.

- `mfa.sh` automates the entire conversion process
- `tts.sh` has been modified to pass arguments to `data.sh` for MFA
- `data.sh` has been edited to produce required text and durations files

To create .lab files for other datasets, define the relevant function in `mfa_format.py`.

PR #4801, opened 2022-12-02
Adding MFA scripts for LJSpeech

iamanigeeit:
(cherry picked from commit f10f911ef0f841bc3c0ffd8bf6258efebdf4411a)

Original PR at: https://github.com/espnet/espnet/pull/4557


Fhrozen:
Just to confirm, did you try the code?. Bc, when I tried, it did not work well.
I mean, some punctuation makes the mfa align generates many OOV (sp) samples with the `us_mfa_english`, and if you tried to use `arpa` then the q is not recognized. At the end, it is better to train a mfa model (and also g2p) using the espnet g2p as reference. 
I did something similar but hardcoded for ljspeech:

https://github.com/Fhrozen/espnet/blob/tst-tts/egs2/ljspeech/tts1/local/mfa_aligments.sh
https://github.com/Fhrozen/espnet/blob/tst-tts/egs2/ljspeech/tts1/local/get_phones_alignments.py

Using these files, in case I obtained a <spn> symbol from the alignment, the program just split into similar frames using espnet g2p as reference for the phonemes of the word.

Fhrozen:
I uploaded some samples of the fastspeech2 trained with mfa (espnet2 g2p for training mfa g2p) + hifi gan 
https://1drv.ms/u/s!AliZ3I0uDW8HhKB14mq0Sumx_vSbpg?e=lFJ4BC

kan-bayashi (commented):
I cleaned codes
- clean bash script
- apply black and isort for python scripts
- move run_mfa.sh in local
- add check of textgrids

PR #4804, opened 2022-12-04
[Draft] Generating Artifacts

Fhrozen:
@sw005320 @kamo-naoyuki 

Let me know if I misses smth while reading the code, but in ESPnet 1,  you were able to generate ctc-logits plots.
I also observed that some other frameworks such as coqi-tts and the others code have a sequence to generate artifacts, such as images or audio samples (in case of TTS).
Espnet2 lacks this function (or I am unsure if it is hidden to activate).

In this PR, I added a general sequence to get Mels comparison from a tts training, such as the following img.

LJ049-0008

I was training to generate audio samples during training, but it will require a little more effort(unless using gan_tts, but not helpful when training the tts and not the vocoder).

Let me know about it, or which direction should take to generate the mentioned artifacts.

### DIFF
--- a/espnet2/tasks/abs_task.py
+++ b/espnet2/tasks/abs_task.py
@@ -328,6 +328,13 @@ class AbsTask(ABC):
             "This option makes sense only when attention-based model. "
             "We can also disable the attention plot by setting it 0",
         )
+        group.add_argument(
+            "--num_valid_artifacts",
+            type=int,
+            default=5,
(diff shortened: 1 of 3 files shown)


sw005320:
Thanks, @Fhrozen, for the discussion.
I think it is good to include @kan-bayashi in this PR.
@kan-bayashi, do you have any comments? Can you review this PR?

Fhrozen:
@sw005320 @kan-bayashi  any comment about this?

PR #4805, opened 2022-12-05
Fixing How2-2000h Data preparation and Seq Length Assert for Longformer Encoder

roshansh-cmu:
We changed the data download mechanism for How2-2000h recently. This impacts the asr1 and sum1 recipes for how2_2000h. 
Therefore, we updated the README and data preparation scripts.

- [x] Data Prep and README for asr1
- [x] Data Prep and README for sum1
- [x] Sequence length assert should not be checked in espnet_model.py for longformer since we may do some padding in the encoder to optimize GPU usage. This is a package default and not easy to modify at this time, so the assert runs always except when the encoder's selfattn is "lf_selfattn". 
- [x] Add HOW2_2kH to db.sh  

Please consider and merge at your convenience.


sw005320:
Thanks a lot, @roshansh-cmu!

PR #4817, opened 2022-12-09
Add unified S3PRL adapter interface

karthik19967829:
This PR adds
1.  a unified adapter interface that should work with all the S3PRL front-ends that are build on top wav2vec2 like Transformerencoder . 
2. Dependency related external library like Fairseq is removed to make installation and maintenance easier 
3. Usage is made easy for end-user by requiring only a config change as indicated in example config to insert adapters


sw005320:
Can you share a model link and results in README.md?

egs2/commonvoice/asr1/conf/adapter_example.yaml
```diff
@@ -0,0 +1,56 @@
+#inserted to 0,1,2
```
simpleoier, line 1: What does this comment mean?

egs2/commonvoice/asr1/conf/adapter_example.yaml
```diff
@@ -0,0 +1,56 @@
...
+  frontend_conf:
+    upstream: hubert_large_ll60k # Note: If the upstream is changed, please change the input_size in the preencoder.
+    add_adapters: True
+    adapter_config:
+      adapter_down_dim: 192
+      adapt_layers: '[23, 12, 11]' # Add the layers index (0-indexed) to be adapted
+  download_dir: ./hub
+  multilayer_feature: True
+
+freeze_param: ["encoder","decoder","ctc"]
```
simpleoier, line 45: If encoder, decoder, and ctc are frozen, do they need some initialization?

espnet2/asr/frontend/adapter_utils/adapters/adapter_transformer.py
```diff
@@ -0,0 +1,216 @@
...
+            layer_result = x
+
+            x = self.dropout3(x)
+            x = residual + x
+            x = self.final_layer_norm(x)
+
+        return x, (attn, layer_result)
+
+
+class AdapterTransformerSentenceEncoderLayer(TransformerSentenceEncoderLayer):
```
simpleoier, line 120: Does this Adapter transformer layer only support upstream models like HuBERT / Wav2Vec? What about other upstream models that are not from fairseq? How about Torchaudios?

espnet2/asr/frontend/adapter_utils/add_adapters.py
```diff
@@ -0,0 +1,81 @@
...
+    if adapt_layers == []:
+        print(">> adapt_layers is an empy list. No adapters will be inserted.")
+        return s3prl_upstream_model
+    orig_param_num = count_params(s3prl_upstream_model)
+
+    # freeze all layers
+    for param in s3prl_upstream_model.parameters():
+        param.requires_grad = False
+
+    for layer_idx, layer in enumerate(s3prl_upstream_model.model.encoder.layers):
```
simpleoier, line 18: `model.encoder.layers` is very specific. What could happen to an upstream with a different model level?

espnet2/asr/frontend/adapter_utils/add_adapters.py
```diff
@@ -0,0 +1,81 @@
...
+)
+
+
+def add_adapters(s3prl_upstream_model, adapter_down_dim, adapt_layers=None):
+    if adapt_layers == []:
+        print(">> adapt_layers is an empy list. No adapters will be inserted.")
+        return s3prl_upstream_model
+    orig_param_num = count_params(s3prl_upstream_model)
+
+    # freeze all layers
```
simpleoier, line 14: This part and the following freeze non-adapter layers is duplicated, isn't it?

espnet2/asr/frontend/adapter_utils/add_adapters.py
```diff
@@ -0,0 +1,81 @@
...
+        adapter_added_layer.load_state_dict(orig_state_dict, strict=False)
+
+        # overwrite original layer with adapter-added layer
+        s3prl_upstream_model.model.encoder.layers[layer_idx] = adapter_added_layer
+
+    # print resulting model stats
+    new_param_num = count_params(s3prl_upstream_model)
+    new_trainable_param_num = count_params(s3prl_upstream_model, only_trainable=True)
+
+    print(f"  * original model weights: {orig_param_num:,}")
```
simpleoier, line 63: It is better to be consistent with other files to use `logging`. I would also suggest to be more clear in description: 1. it calculates the parameters of s3prl frontend, not the espnet model; 2. clarify that it is before adding adapter and after adding adapter, instead of `original model` and `new model`.

espnet2/asr/frontend/s3prl.py
```diff
@@ -55,6 +56,13 @@ def __init__(
             "HubertModel",
         ]:
             upstream.model.encoder.layerdrop = 0.0
+            if getattr(self.args, "add_adapters", False):
```
simpleoier, line 59: where is `self.args` from?

simpleoier (commented):
Thanks for preparing this new feature!
I left some comments! Besides, I would suggest to make a new frontend for adapter s3prl frontend. It can inherit from the s3prl frontend, but adding some adapter ad-hoc arguments / checkings / asserts, etc.

espnet2/asr/frontend/s3prl.py
simpleoier: I don't think this add_adapters is unified for all s3prl upstream. It is only for HuBERT and Wav2Vec.

PR #4821, opened 2022-12-10
Add Native onnx export support to encoders and CTC

karthik19967829:
This PR updates make_pad_mask and related functions to make ESPnet transformer/conformer encoder and CTC ONNX export friendly . 
This particular change would reduce maintenance efforts of espnet_onnx package to be directly export encoders and CTC using ESPnet core package and avoid re-implimentation

### DIFF
--- a/espnet/nets/pytorch_backend/nets_utils.py
+++ b/espnet/nets/pytorch_backend/nets_utils.py
@@ -60,9 +60,55 @@ def pad_list(xs, pad_value):
 
     return pad
 
-
-def make_pad_mask(lengths, xs=None, length_dim=-1, maxlen=None):
+def make_pad_mask(lengths, maxlen=None):
     """Make mask tensor containing indices of padded part.
+
+    Args:
+        lengths (LongTensor or List): Batch of lengths (B,).
+
+    Returns:
+        BoolTensor: Mask tensor containing indices of padded part.
+
+    Examples:
+        >>> lengths = [5, 3, 2]
+        >>> make_pad_mask(lengths)
+        masks = [[0, 0, 0, 0 ,0],
+                 [0, 0, 0, 1, 1],
+                 [0, 0, 1, 1, 1]]
+    """
+    lengths = lengths.long()
+    bs = lengths.size(0)
+    if maxlen is None:
+            maxlen = lengths.max()
+
+    seq_range = torch.arange(0, maxlen, dtype=torch.int64, device=lengths.device)
+    seq_range_expand = seq_range.unsqueeze(0).expand(bs, maxlen)
+    seq_length_expand = seq_range_expand.new(lengths).unsqueeze(-1)
+    mask = seq_range_expand >= seq_length_expand
+
+    return mask
+
+def make_non_pad_mask(lengths):
+    """Make mask tensor containing indices of non-padded part.
+
+    Args:
+        lengths (LongTensor or List): Batch of lengths (B,).
+
+    Returns:
+        BoolTensor: Mask tensor containing indices of padded part.
+
+    Examples:
+        >>> lengths = [5, 3, 2]
+        >>> make_non_pad_mask(lengths)
+        masks = [[1, 1, 1, 1 ,1],
+                 [1, 1, 1, 0, 0],
+                 [1, 1, 0, 0, 0]]
+    """
+
+    return ~make_pad_mask(lengths)
+
+def make_pad_mask_with_reference(lengths, xs=None, length_dim=-1, maxlen=None):
+    """Make mask tensor with the shape of a reference tensor.
 
     Args:
         lengths (LongTensor or List): Batch of lengths (B,).
@@ -77,7 +123,7 @@ def make_pad_mask(lengths, xs=None, length_dim=-1, maxlen=None):
                 dtype=torch.bool in PyTorch 1.2+ (including 1.2)
 
     Examples:
-        With only lengths.
+        With only lengths. Equivalent to regular make_pad_mask(lengths)
 
         >>> lengths = [5, 3, 2]
         >>> make_pad_mask(lengths)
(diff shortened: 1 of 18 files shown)


espnet/nets/pytorch_backend/nets_utils.py
```diff
@@ -60,10 +60,56 @@ def pad_list(xs, pad_value):
...
+        BoolTensor: Mask tensor containing indices of padded part.
+
+    Examples:
+        >>> lengths = [5, 3, 2]
+        >>> make_pad_mask(lengths)
+        masks = [[0, 0, 0, 0 ,0],
+                 [0, 0, 0, 1, 1],
+                 [0, 0, 1, 1, 1]]
+    """
+    lengths = lengths.long()
```
Masao-Someki, line 79: When `type(lengths)` is `List`, this line will cause an error.
If we support both `torch.Tensor` and `List`, then we need to add if sentence like

```python
if isinstance(lengths, List):
    lengths = torch.Tensor(lengths)
```

espnet/nets/pytorch_backend/transformer/attention.py
```diff
@@ -75,14 +75,10 @@ def forward_attention(self, value, scores, mask):
...
-            mask = mask.unsqueeze(1).eq(0)  # (batch, 1, *, time2)
-            min_value = torch.finfo(scores.dtype).min
-            scores = scores.masked_fill(mask, min_value)
-            self.attn = torch.softmax(scores, dim=-1).masked_fill(
-                mask, 0.0
-            )  # (batch, head, time1, time2)
-        else:
-            self.attn = torch.softmax(scores, dim=-1)  # (batch, head, time1, time2)
+            mask = mask.unsqueeze(1)
+            mask = (~mask)*-10000
```
Masao-Someki, line 79: Please insert a space around `*` to match the other parts

espnet2/bin/asr_export_onnx.py
```diff
@@ -0,0 +1,184 @@
...
+    if not test_input:
+        test_input = dummy_input
+
+    test_feats, test_feats_lens = test_input
+    enc_out, enc_out_lens, _ = encoder(test_feats, test_feats_lens)
+    enc_out = enc_out.detach().numpy()
+
+    options = rt.SessionOptions()
+    options.graph_optimization_level = rt.GraphOptimizationLevel.ORT_ENABLE_ALL
+    onnx_model = rt.InferenceSession(out_filename, options)
```
Masao-Someki, line 57: We need to specify providers here with `onnxruntime>=1.9`.
We can add `providers=['CPUExecutionProvider']` to infer with CPU, or `'CUDAExecutionProvider` to use GPU.
Sample: https://onnxruntime.ai/docs/execution-providers/CUDA-ExecutionProvider.html#samples

espnet2/bin/asr_export_onnx.py
```diff
@@ -0,0 +1,184 @@
...
+    # test ONNX CTC and compare outputs with PyTorch CTC
+    if not test_input:
+        test_input = dummy_input
+
+    ctc_out = ctc.log_softmax(dummy_enc_out)
+    ctc_out = ctc_out.detach().numpy()
+
+    options = rt.SessionOptions()
+    options.graph_optimization_level = rt.GraphOptimizationLevel.ORT_ENABLE_ALL
+    onnx_ctc = rt.InferenceSession(out_filename, options)
```
Masao-Someki, line 104: As above, we need to specify providers.

espnet/nets/pytorch_backend/nets_utils.py
kenzheng99: Yeah, I also thought about this, but wasn't sure exactly in which cases we are passing in tensors vs lists. @Masao-Someki do you know if the ONNX export is able to properly build the computational graph in the case that it is a Python list? I was assuming that it might not work, so was thinking it might be better to change everything into tensors.

espnet2/bin/asr_export_onnx.py
kenzheng99: Can you explain more what specifying the provider does? If it's only for optimization purposes, I think it might not be necessary here, this call is mostly a sanity check to make sure that the export was successful. (in fact, I think we might   not even need the `graph_optimization_level`)

espnet/nets/pytorch_backend/nets_utils.py
Masao-Someki: @kenzheng99
I think torch can export the model properly. I'm not sure if any function call uses this function with `List` as input, but I think limiting the input type is better for readability.

espnet2/bin/asr_export_onnx.py
Masao-Someki: @karthik19967829
Maybe your onnxruntime version is old. We need to specify providers when we use `onnxruntime>=1.9.0`, so would you test this script with the latest onnxruntime?
When I try to create InferenceSession with onnxruntime==1.11, I got this error.
```python
import onnxruntime as ort
m = ort.InferenceSession('./ctc.onnx')
```

> ValueError: This ORT build has ['CUDAExecutionProvider', 'CPUExecutionProvider'] enabled. Since ORT 1.9, you are required to explicitly set the providers parameter when instantiating InferenceSession. For example, onnxruntime.InferenceSession(..., providers=['CUDAExecutionProvider', 'CPUExecutionProvider'], ...)

Masao-Someki: @kenzheng99 
Sorry, I mentioned the wrong person.

espnet/nets/pytorch_backend/transformer/attention.py
sw005320: This will change the results. So, I'm hesitating about this change.
Is it necessary?
Can you also confirm that this does not affect the result?

Also, `-10000` should be a variable and can be changed.

sw005320 (commented):
`make_pad_mask` is a very important function, and it is not safe to replace it with `make_pad_mask_with_reference`

Where is the original `make_pad_mask` used? If there is no case, we can rename the current `make_pad_mask_with_reference` as `make_pad_mask`

kenzheng99:
> `make_pad_mask` is a very important function, and it is not safe to replace it with `make_pad_mask_with_reference`
> 
> Where is the original `make_pad_mask` used? If there is no case, we can rename the current `make_pad_mask_with_reference` as `make_pad_mask`

Sorry for the confusion, `make_pad_mask_with_reference` is actually the unchanged old function, and `make_pad_mask` is the new simplified one that we implemented. I swapped the names around since I thought it made more sense like this, but open to other naming suggestions.

The `make_pad_mask` (new function) in this branch is used anytime the `xs` input is not required, like in most of the transformer-based model code. I think this should be safe, as the outputs are identical to the old function when `xs` is `None`.

espnet/nets/pytorch_backend/transformer/attention.py
kenzheng99: I think it might not be strictly necessary, @Masao-Someki have you tested ONNX export without this change?

I do like this implementation better though, its cleaner in both the code and the ONNX computational graph in my opinion. It might affect activation values a bit but it didn't change any of the final output transcripts in our testing, but maybe more testing is needed to confirm this. Masao's codebase also uses this change.

sw005320: I like this way since it would not change the behavior depending on the program (which is important for ONNX)
However, we should at least check 
- how it would affect to the performance.
- It is better to have a unit test.

sw005320:
I want to minimize the change comes from this PR.
So, my suggestion is to keep `make_pad_mask` as it is and call new `make_pad_mask` as `make_pad_mask_without_reference` or `make_pad_mask_onnx` or whatever.
Thus, we can minimize the change.

espnet/nets/pytorch_backend/transformer/attention.py
Masao-Someki: @kenzheng99 @sw005320 
When I implemented this `-10000.0` mask, I referenced the implementation of huggingface transformers. Currently, the implementation seems to be fixed to the minimum value [link](https://github.com/huggingface/transformers/blob/4f1c9d162ee53004be4d63b76627bd0c2f5a31a9/src/transformers/models/codegen/modeling_codegen.py#L519-L525).
The output probability of onnx and torch might differ slightly (MSE < 1e-10), but I didn't observe the accuracy drop. Unit test of this parity check is implemented in the espnet_onnx repository [link](https://github.com/espnet/espnet_onnx/blob/18eb341d44ccf83c3ab35bc040b4102a73602922/tests/unit_tests/test_inference_asr.py#L58-L65).
We only need to apply the `-10000` mask for conversion, so we can use the old function while training and switch to the onnx version when conversion. I'm unsure if we can do this, but switching the mask function in `make_pad_mask` inside may be a better implementation:

```python
def make_pad_mask(...):
  if onnx_converion:
    return make_pad_mask_onnx(...)
  else:
    # old implementation
```

kenzheng99:
moved to https://github.com/espnet/espnet/pull/4928

Masao-Someki:
We moved this issue to #4928, and then it was moved to #5326, which is already merged and closed. I think we can close this.

PR #4822, opened 2022-12-13
fix the issue of  start and end times not appear in pairs in calculat…

haihua:
…e_rtf.py

### DIFF
--- a/utils/calculate_rtf.py
+++ b/utils/calculate_rtf.py
@@ -75,24 +75,23 @@ def main():
         audio_durations = []
         start_times = []
         end_times = []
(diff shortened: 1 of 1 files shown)


sw005320:
@espnetUser, can you check this?
It looks correct to me.

haihua:
Sorry, I just modified the calculate_rtf.py and tested it.
Our observation: if some utterances have not been normally decoded, as a
result, there is not 'best hypo', and failed to pair.
Then it failed for the following assert statements.
Our point is to ignore such utterances.

Why did this modification cause so many errors ?

Regards,

Thanks

PR #4824, opened 2022-12-15
Multilingual superb

ftshijt:
Recipe for multilingual SUPERB


ftshijt:
Still need to add dedicated scoring functions to make it smooth to evaluation related to the benchmark. But the preparation and training should be fine.

ftshijt:
The PR shall be ready to merge. The customized scoring will be finished for the next step.

sw005320:
OK, thanks @ftshijt!
I’m assuming that the data download part would also be added later.

PR #4830, opened 2022-12-17
Update utils.py

aky15:
The usage of left_chunk_size is different from num_left_chunks in k2. 
By definitions in [tutorial](https://github.com/espnet/espnet/blob/master/doc/espnet2_tutorial.md) and [encoder](https://github.com/espnet/espnet/blob/master/espnet2/asr_transducer/encoder/building.py), it's the number of frames in the left context, so left_chunk_size = num_left_chunks * chunk_size.

### DIFF
--- a/espnet2/asr_transducer/utils.py
+++ b/espnet2/asr_transducer/utils.py
@@ -96,7 +96,7 @@ def make_chunk_mask(
         if left_chunk_size <= 0:
             start = 0
         else:
-            start = max(i  // chunk_size  * chunk_size - left_chunk_size, 0)
+            start = max(i // chunk_size  * chunk_size - left_chunk_size, 0)
 
         end = min((i // chunk_size + 1) * chunk_size, size)
         mask[i, start:end] = True
--- a/espnet2/asr_transducer/utils.py
+++ b/espnet2/asr_transducer/utils.py
@@ -96,7 +96,7 @@ def make_chunk_mask(
         if left_chunk_size <= 0:
             start = 0
         else:
-            start = max((i - left_chunk_size) // chunk_size  * chunk_size, 0)
+            start = max(i  // chunk_size  * chunk_size - left_chunk_size, 0)
 
         end = min((i // chunk_size + 1) * chunk_size, size)
         mask[i, start:end] = True
--- a/espnet2/asr_transducer/utils.py
+++ b/espnet2/asr_transducer/utils.py
@@ -96,7 +96,7 @@ def make_chunk_mask(
         if left_chunk_size <= 0:
             start = 0
         else:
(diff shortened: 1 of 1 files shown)


b-flo (commented):
Hi,

Thanks for the PR!

> The usage of left_chunk_size is different from num_left_chunks in k2.

Could you elaborate, please?
I may be missing something here but our `make_chunk_mask` is equivalent to `subsequent_chunk_mask` in k2 (minus the bitwise negation operator applied outside the method in their version). `left_chunk_size` and `num_left_chunks` are used in the same manner in both versions AFAIK.

aky15:
I mean the usage of left_chunk_size is somewhat inconsistent with the definition in [the tutorial](https://github.com/espnet/espnet/blob/master/doc/espnet2_tutorial.md) and [the encoder](https://github.com/espnet/espnet/blob/master/espnet2/asr_transducer/encoder/building.py), which is 'the number of frames in left context'.  If left_chunk_size and num_left_chunks are used in the same manner, maybe it's better to rename it for the sake of clarity?

aky15:
The left_context and right_context in inference are used in the same manner as num_left_chunks and num_right_chunks, but denoted as the num of frames in left/right context in the tutorial.

b-flo:
> I mean the usage of left_chunk_size is somewhat inconsistent with the definition in [the tutorial](https://github.com/espnet/espnet/blob/master/doc/espnet2_tutorial.md) and [the encoder](https://github.com/espnet/espnet/blob/master/espnet2/asr_transducer/encoder/building.py), which is 'the number of frames in left context'. If left_chunk_size and num_left_chunks are used in the same manner, maybe it's better to rename it for the sake of clarity?

OK, I get it now. Yes, it may be better to do some renaming and also improve the doc-strings and tutorials.
We should ensure back compatibility though (by adding new argument and setting the dest of the old argument to the new one).

> The left_context and right_context in inference are used in the same manner as num_left_chunks and num_right_chunks, but denoted as the num of frames in left/right context in the tutorial.

Please note that the inference part is wrong on multiple aspects. A new version will be added where `right_context` was dropped, `chunk_size` was replaced by `decoding_window` (input audio length in ms) and `left_context` behave as expected (L previous frames as left context for current computations).

I should have opened a PR last week but I'm waiting for some feedback to finalize the version. There is a slight performance degradation with small windows (<320ms) and some calculations are technically not correct (but intended to correct the observed issues). Sorry for that delay.

aky15:
OK, Thanks for your contribution!

b-flo:
FYI: I updated the docs and tutorial in the new version, you can find it [here](https://github.com/b-flo/espnet/tree/refactoring). Feel free to tell me if it's not clear enough.

PR #4832, opened 2022-12-18
Fixed pad mode for librosa.stft

Masao-Someki:
This PR fixes the `librosa.stft` call in the default frontend.

The default parameter of the `pad_mode` for `torch.stft` and `librosa.stft` is different, so we need to use the same value to get the same output.

### DIFF
--- a/espnet2/layers/stft.py
+++ b/espnet2/layers/stft.py
@@ -117,6 +117,7 @@ class Stft(torch.nn.Module, InversibleInterface):
                 hop_length=self.hop_length,
                 center=self.center,
                 window=window,
+                pad_mode='reflect',
             )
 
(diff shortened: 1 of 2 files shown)


sw005320:
Thanks a lot!
Can you add a unit test to check that both stft produces the same result?

Masao-Someki:
@sw005320 
I added a parity check for `torch.stft` and `librosa.stft`. The `atol` value in `allclose` function is the same as the test script of the PyTorch test ([link](https://github.com/pytorch/pytorch/blob/6faa6c68e8b76fb68f3a2b2783685102d0e87c00/test/test_spectral_ops.py#L919)).

sw005320:
Cool!

PR #4836, opened 2022-12-22
Fix data prep in GigaSpeech

pyf98:
Hi, the original data prep script seems not working. Probably the interface of the `GigaSpeech` repo has changed. The new usage can be found here: https://github.com/SpeechColab/GigaSpeech/blob/main/toolkits/kaldi/gigaspeech_data_prep.sh#L67


sw005320:
Thanks!

Issue #4839, opened 2022-12-28
"nlsyms_txt doesn't exist" in Stage 13: scoring


pyf98:
**Describe the bug**

When executing scoring for WSJ in ESPnet2 ASR, I am warned that "data/nlsyms.txt" doesn't exist: [code](https://github.com/espnet/espnet/blob/master/espnet2/text/word_tokenizer.py#L34). As a result, the special symbols such as `<NOISE>` are not removed for calculating error rates. However, the following argument is passed to `asr.sh`:

```bash
./asr.sh \
    ...
    --nlsyms_txt data/nlsyms.txt \
```

The issue comes from [this line](https://github.com/espnet/espnet/blob/master/egs2/TEMPLATE/asr1/asr.sh#L1413):
```bash
_opts+="--non_linguistic_symbols \"${nlsyms_txt}\" "
```
After changing it to `_opts+="--non_linguistic_symbols ${nlsyms_txt} "`, the issue seems to be fixed.

sw005320:
Thanks a lot for the report!
Can you tell me how much the score is changed with this?

pyf98:
For a Conformer model:
Using `nlsyms_txt=none`:
```
### WER

|dataset|Snt|Wrd|Corr|Sub|Del|Ins|Err|S.Err|
|---|---|---|---|---|---|---|---|---|
|decode_lm_lm_train_lm_transformer_en_char_valid.loss.ave_asr_model_valid.acc.ave/test_dev93|503|8334|93.0|5.2|1.8|0.8|7.8|61.2|
|decode_lm_lm_train_lm_transformer_en_char_valid.loss.ave_asr_model_valid.acc.ave/test_eval92|333|5700|95.6|3.1|1.3|0.5|4.9|45.3|

### CER

|dataset|Snt|Wrd|Corr|Sub|Del|Ins|Err|S.Err|
|---|---|---|---|---|---|---|---|---|
|decode_lm_lm_train_lm_transformer_en_char_valid.loss.ave_asr_model_valid.acc.ave/test_dev93|503|49334|96.6|1.1|2.2|0.6|4.0|61.2|
|decode_lm_lm_train_lm_transformer_en_char_valid.loss.ave_asr_model_valid.acc.ave/test_eval92|333|33740|97.5|0.6|1.8|0.4|2.9|45.3|
```

Using `nlsyms_txt=data/nlsyms.txt`:
```
### WER

[...]

PR #4841, opened 2022-12-28
Fix argument parsing for non_linguistic_symbols in asr.sh

pyf98:
This PR fixes https://github.com/espnet/espnet/issues/4839


sw005320:
Many thanks, @pyf98!

PR #4843, opened 2022-12-29
Add a warning statement when the hypo length equals to the max out length.

pengchengguo:
Add a warning statement when the hypo length equals the max out length.
The user should consider increasing the `maxlenratio` to avoid the decoding being stopped incorrectly by the max out length limitation.


sw005320:
Thanks!
@brianyan918, can you also add the same warning for ST/MT?
We could avoid the maximum output length issue in the ST/MT side.

PR #4856, opened 2023-01-07
Add E-Branchformer results on Librispeech

kkim-asapp:
This PR is about updating README for E-Branchformer results on Librispeech.
I also add a description about an uploaded E-Branchformer model on Huggingface repo.


sw005320:
Thanks a lot!

PR #4861, opened 2023-01-10
Support SOT training on LibriMix data.

pengchengguo:
Xuankai and I would like to include SOT training (https://arxiv.org/pdf/2003.12687.pdf) in ESPnet.

We modify the CSV files of the original LibriMix repo by adding random overlap offsets from 1-1.5s. This PR provides a unified simulation metric and a comparable benchmark for non-full-overlapped multi-speaker ASR.

Note that it is not ready for merging yet. We will update some results and pre-trained models later.

TODO:
- [x] Results of Transformer PIT w/ or w/o WavLM
- [x] Results of Transformer SOT w/ or w/o WavLM
- [x] Results of Conformer SOT w/ or w/o WavLM


egs2/librimix/sot_asr1/conf/tuning/train_pit_asr_transformer_wavlm.yaml
```diff
@@ -0,0 +1,90 @@
+# network architecture
```
simpleoier, line 1: PIT model may use different data preparation with SOT. It may be a bit surprising to put this config, especially this recipe folder is under `sot_asr1`. I don't know. Maybe we can explicitly make some notes about this config, e.g. what data preparation, what purpose, what differences, and how to use are about it.

egs2/librimix/sot_asr1/conf/tuning/train_pit_asr_transformer_wavlm.yaml
```diff
@@ -0,0 +1,90 @@
...
+freeze_param: [
+    "frontend.upstream"
+]
+
+# network architecture
+frontend: s3prl
+frontend_conf:
+    frontend_conf:
+        upstream: wavlm_local
+        path_or_url: "/home/work_nfs6/pcguo/asr/librimix/hub/wavlm_large.pt"
```
simpleoier, line 13: This path is for your local experiments.

egs2/librimix/sot_asr1/conf/tuning/train_sot_asr_conformer_wavlm.yaml
```diff
@@ -0,0 +1,88 @@
...
+freeze_param: [
+    "frontend.upstream"
+]
+
+# network architecture
+frontend: s3prl
+frontend_conf:
+    frontend_conf:
+        upstream: wavlm_local
+        path_or_url: "/home/work_nfs6/pcguo/asr/librimix/hub/wavlm_large.pt"
```
simpleoier, line 13: The path of the config is for your own.

simpleoier (commented):
Thanks @pengchengguo !

egs2/librimix/sot_asr1/conf/tuning/train_pit_asr_transformer_wavlm.yaml
pengchengguo: The motivation for including PIT  is to compare PIT and SOT on the same dataset (overlapped speech with a certain time delay). It is ok for me to remove it and only keep the SOT file here~

egs2/librimix/sot_asr1/conf/tuning/train_pit_asr_transformer_wavlm.yaml
pengchengguo: Fixed.

egs2/librimix/sot_asr1/conf/tuning/train_sot_asr_conformer_wavlm.yaml
pengchengguo: Fixed.

egs2/librimix/sot_asr1/local/metadta_Libri2Mix_offset/Storage_info.txt
```diff
@@ -0,0 +1,154 @@
+842M	Libri2Mix/wav16k/max/dev/s2
```
sw005320, line 1: where is it used?

egs2/librimix/sot_asr1/local/metadta_Libri2Mix_offset/libri2mix_dev-clean.csv
```diff
@@ -0,0 +1,3001 @@
+mixture_ID,source_1_path,source_1_gain,source_2_path,source_2_gain,noise_path,noise_gain,offset
```
sw005320, line 1: can you explain why this is needed?
to avoid the full overlap in the original librimix?

sw005320 (commented):
- does it have to be `sot_asr1`?
since it just changes the config, I think we can move this to `asr1` and change `local/data.sh` to `local/data_sot.sh`, and it is called from the original `local/data.sh`
If this makes conflicts with the existing data directory, your current idea is better.
- it is better to add some descriptions to README.md, for example, why we need CSV files, what is the SOT text format, and how `<sc>` symbols are treated in the tokenizer through the `--add_nonsplit_symbol` option

egs2/librimix/sot_asr1/local/metadta_Libri2Mix_offset/libri2mix_dev-clean.csv
pengchengguo: Yes, the updated csv file includes an additional column of "offset", which is used as the time delay when simulating overlapped data. Here we randomly generate offsets from 1s-1.5s.
The motivation for fixing the offset is to allow people to do a fair comparison on the same simulated data among different models.

egs2/librimix/sot_asr1/local/metadta_Libri2Mix_offset/Storage_info.txt
pengchengguo: I will remove it.

egs2/TEMPLATE/asr1/asr.sh
```diff
@@ -766,6 +772,9 @@ if ! "${skip_data_prep}"; then
 
             _opts="--non_linguistic_symbols ${nlsyms_txt}"
 
+            if ${sot_asr} && [ "${token_type}" = char ]; then
```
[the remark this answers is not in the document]
pengchengguo: - The main difference between `sot_asr1` and `asr1`  is the data simulation process and the training data. If merging these two directories, we may need to introduce additional files (eg. `data/{train_sot,dev_sot,test_sot}`, `run_sot.sh`). I think it is ok to add some files but the results in `README.md` can not be compared with each other (due to the different training data, full-overlapped or partial-overlapped).
- I will add more descriptions about SOT
- "only char case?", in our experiments, SOT only converges well on char units, thus we only considered this case, I will check the compatibility with other units like BPE.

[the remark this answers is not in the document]
sw005320: OK, thanks.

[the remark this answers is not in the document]
simpleoier: > only char case?
When we prepare the text, `<sc>` is inserted between speakers, with space on the left and right.
1) For the word case, `<sc>` is automatically regarded as a single word unit. So no special process for it.
2) For the BPE case, we use `_opts_spm+=" --user_defined_symbols=<sc>"` to take care of it. It will not be mixed in other BPE subwords.
3) For the char case, we need to add this special arguments to let the text processing know this is a single unit. It wouldn't be split into `<`, `s`, `c`, `>`.

sw005320:
@pengchengguo, can we try to finish this PR?

sw005320:
There is a conflict due to the recent refactoring of `asr.sh`
Can you fix this?

pengchengguo:
> There is a conflict due to the recent refactoring of `asr.sh` Can you fix this?

Sure, I am checking it now.

simpleoier (approved):

egs2/librimix/sot_asr1/conf/tuning/decode_pit.yaml
sw005320, line 1: where was it used?

pengchengguo: Removed

sw005320 (approved):

sw005320:
LGTM, thanks a lot!

pengchengguo:
Sorry, it seems the change (remove useless `decode_pit.yaml` file) was not submitted successfully, I may need to open another PR to remove this file or we can leave it to a later PR, like PR for updating results.

sw005320:
OK!
I think this is my mistake.
Sorry...

PR #4882, opened 2023-01-21
Add E-Branchformer for GigaSpeech

pyf98:
Hi, this PR adds a config and pre-trained model for GigaSpeech. The encoder config is based on the LibriSpeech recipe.


sw005320:
Thanks a lot!

PR #4888, opened 2023-01-25
Update collect stats stage so that less memory cost in Utt_mvn

simpleoier:
If utt_mvn is used and `extract_feats_in_collect_stats` is False in config, the scripts would skip building model and no dummy features are generated.
This can reduce the memory usage in Stage 10 when large SSL models are used, e.g. HuBERT.

### DIFF
--- a/egs2/TEMPLATE/asr1/asr.sh
+++ b/egs2/TEMPLATE/asr1/asr.sh
@@ -1106,6 +1106,10 @@ if ! "${skip_train}"; then
         for i in $(seq "${_nj}"); do
             _opts+="--input_dir ${_logdir}/stats.${i} "
         done
+        if [ "${feats_normalize}" != global_mvn ]; then
+            # Skip summerizaing stats if not using global MVN
+            _opts+="--skip_sum_stats"
+        fi
         # shellcheck disable=SC2086
         ${python} -m espnet2.bin.aggregate_stats_dirs ${_opts} --output_dir "${asr_stats_dir}"
 
--- a/egs2/TEMPLATE/diar1/diar.sh
+++ b/egs2/TEMPLATE/diar1/diar.sh
@@ -367,6 +367,10 @@ if ! "${skip_train}"; then
         for i in $(seq "${_nj}"); do
             _opts+="--input_dir ${_logdir}/stats.${i} "
         done
+        if [ "${feats_normalize}" != global_mvn ]; then
+            # Skip summerizaing stats if not using global MVN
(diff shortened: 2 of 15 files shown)


pengchengguo (approved):
Sorry didn't finish it in time. It looks good to me.

sw005320:
I just want to make sure. 
Don't we need to change anything on the config side in the recipe?

simpleoier:
@sw005320 I don't think we need to change it on the config side in the recipe. I tried to make it compatible with previous configs / pretrained models in the following two ways:
1. Find the hyper-parameter `extract_feats_in_collect_stats` from task model configs, [here](https://github.com/espnet/espnet/pull/4888/files#diff-475643ecd83de2b2b9d751832830d3db2cf07c2945732964ce0c1788b967464eR1147)
2. Keep the place holder of `extract_feats_in_collect_stats` in the espnet_model.

sw005320:
OK, thanks for the confirmation!

kamo-naoyuki:
@simpleoier I noticed this change now. Thank you, I agree this is useful option, but I think `model_conf.extract_feats_in_collect_stats` is bad place to specify this.

Why not adding just `--extract_feats_in_collect_stats` option simply?

simpleoier:
@kamo-naoyuki I use `model_conf.extract_feats_in_collect_stats` because we had this attribute before. Then we don't have to change the pre-trained models. And we don't introduce a new argument.
Please let me know your ideas about it.

kamo-naoyuki:
> because we had this attribute before.

I can't understand why you had this attribute in the pre-trained model before this PR, but it's okay.

Could you implement `--extract_feats_in_collect_stats` and keep also model_conf.extract_feats_in_collect_stats for backward compatibility, and als show deprecated warning.

Currently, user can't know the option of `model_conf.extract_feats_in_collect_stats` now by the help command. In addition, it's very strange to control the behaviours of `collect_stats` by `model_conf`

simpleoier:
> I can't understand why you had this attribute in the pre-trained model before this PR, but it's okay.

It was used to avoid doing model forward in collect_stats when using self-supervised learning frontends. 

> Could you implement --extract_feats_in_collect_stats and keep also model_conf.extract_feats_in_collect_stats for backward compatibility, and als show deprecated warning if using model_conf.extract_feats_in_collect_stats

Sounds good. I'll do it soon.

kamo-naoyuki:
> Sounds good. I'll do it soon.

Thank you!

PR #4892, opened 2023-01-27
Add Conv2dSubsampling1 module and test it in AphasiaBank ASR recipe

tjysdsg:
`Conv2dSubsampling1` module is similar to `Conv2dSubsampling` except the subsampling ratio is 1:1.

AphasiaBank English ASR experiments using E-Branchformer+WavLM showed a CER decrease from 17.7 to 17.0 by switching from `Conv2dSubsampling2` to `Conv2dSubsampling1`.

The user can specify `input_layer: conv2d1` in the encoder configuration to use this module.

### DIFF
--- a/egs2/aphasiabank/asr1/README.md
+++ b/egs2/aphasiabank/asr1/README.md
@@ -11,8 +11,9 @@
 
 ## asr_train_asr_ebranchformer_small_wavlm_large1
 
-- [train_asr_ebranchformer_small_wavlm_large.yaml](conf/tuning/train_asr_ebranchformer_small_wavlm_large.yaml)
+- [train_asr_ebranchformer_small_wavlm_large1.yaml](conf/tuning/train_asr_ebranchformer_small_wavlm_large1.yaml)
 - Control group data is included
+- Downsampling rate = 2 = 2 (WavLM) * 1 (`Conv2dSubsampling1`)
 - [Hugging Face](https://huggingface.co/espnet/jiyang_tang_aphsiabank_english_asr_ebranchformer_small_wavlm_large1)
 
 ### WER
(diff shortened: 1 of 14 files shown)


simpleoier (approved):
Better results!
Looks good to me in general!
Just one concern about the name of the new module, `Conv2dSubsampling1` doesn't do subsampling actually.

tjysdsg:
> Better results! Looks good to me in general! Just one concern about the name of the new module, `Conv2dSubsampling1` doesn't do subsampling actually.

Oh right, what's your suggestion? Should we just call it `Conv2d1` or `Conv2dInputLayer`?

egs2/aphasiabank/asr1/README.md
```diff
@@ -9,6 +9,24 @@
 - Git hash: `39c1ec0509904f16ac36d25efc971e2a94ff781f`
     - Commit date: `Wed Dec 21 12:50:18 2022 -0500`
 
+## asr_train_asr_ebranchformer_small_wavlm_large1
+
+- [train_asr_ebranchformer_small_wavlm_large.yaml](conf/tuning/train_asr_ebranchformer_small_wavlm_large.yaml)
```
pyf98, line 14: Can you add a short note here to explain the difference? E.g., this config uses xxx as the input layer which does not perform downsampling.

pyf98: Is this config correct? It looks same as the previous one.

tjysdsg: Thanks! Fixed in 4734d75

sw005320:
> > Better results! Looks good to me in general! Just one concern about the name of the new module, `Conv2dSubsampling1` doesn't do subsampling actually.
> 
> Oh right, what's your suggestion? Should we just call it `Conv2d1` or `Conv2dInputLayer`?

I think `Conv2dSubsampling1` is OK, but it is better to add a comment that this does not do any subsampling in appropriate places.

tjysdsg:
@sw005320 I made the comment clearer about the functionality in c1dc65f

sw005320:
Thanks a lot!

PR #4895, opened 2023-01-30
single language track setups

DanBerrebbi:
I changed some single language langs : 
I removed pol because similar to rus and nob because similar to swe,
I added French because there was no roman language(french, spanish, italian, portuguese ...) and added Swahili because there was no African language. 
For dataset selections, it is summarized on the last page of https://docs.google.com/document/d/1sb8SyDjcMf7FDiZHH8wVcZ0EADtXdNBF3LpA9Cu0I1k/edit

 
Capture d’écran 2023-01-30 à 11 51 17

For test sets format, I think that it is good to keep only one test set per language with all the datasets of this lang. This way it is an easy decoding process and then WE can split it the decoded file to have scores per dataset and so compute metrics for domain shifts ... . So we have flexibility for scoring and the user has a simple process.

Points to be discussed : 
* lang choices
* VoxPopuli not working 
* Should we use speed pert ? In my opinion no

### DIFF
--- a/egs2/msuperb/asr1/local/single_lang_data_prep.py
+++ b/egs2/msuperb/asr1/local/single_lang_data_prep.py
@@ -21,7 +21,22 @@ DATA = [
     "voxpopuli",
 ]
 
-SINGLE_LANG = ["eng", "deu", "rus", "pol", "swe", "jpn", "cmn", "sat", "nob", "xty"]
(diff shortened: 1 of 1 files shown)


ftshijt (commented):
Looks good to me. Please follow our discussion and fix the CI issues. Let's accelerate the process~ Thanks @DanBerrebbi

Issue #4866, opened 2023-01-13
On how2 and how2-2000h recipes


chlorane:
I tried to use how2 recipe from egs2/how2/asr1. However, when I execute run.sh, an error happened
local/data.sh: error executing wget http://islpc21.is.cs.cmu.edu/ramons/iwslt2019.tar.gz
It seems the link here is unavailable.

Also, I tried to run the new how2-2000h recipe from egs2/how2_2000h/asr1. However, error still happens.
2023-01-13T10:15:50 (data.sh:50:main) stage 1: Data preparation and verification
utils/fix_data_dir.sh: file data/tr_2000h_utt/feats.scp is not in sorted order or not unique, sorting it
utils/fix_data_dir.sh: filtered data/tr_2000h_utt/segments from 950026 to 0 lines based on filter /tmp/kaldi.T4ZL/recordings.
utils/fix_data_dir.sh: filtered data/tr_2000h_utt/wav.scp from 950026 to 0 lines based on filter /tmp/kaldi.T4ZL/recordings.
fix_data_dir.sh: no utterances remained: not proceeding further.

How could I solve them?

sw005320:
@roshansh-cmu, could you respond to it?

sw005320:
@roshansh-cmu, this is a reminder

roshansh-cmu:
Thank you Shinji- I have investigated this and am preparing a PR to resolve all issues.

roshansh-cmu:
@chlorane - please pull the fixes made and retry the data prep. It works end-to-end without any issues at this time. 

Also, remember that you must request "(audio_2000) fbank+pitch features in Kaldi scp/ark format" as mentioned in our [How2 github](https://github.com/srvk/how2-dataset). The IWSLT version is for 300h and not for summarization.

sw005320:
Thanks, @roshansh-cmu!

smilepokemmo:
@chlorane Hello, thank you for your work! May I ask if you still have a backup of the How2_2000h dataset? The official sources no longer provide this dataset and suggest obtaining it from other downloaders. Unfortunately, the dataset I downloaded earlier is missing some files - specifically, I may be missing cv05utt, cv05_sum, utt2spk, and spk2utt. Additionally, could you provide the complete dataset structure? This will help me further confirm which parts are missing.

PR #4899, opened 2023-01-31
Fix dataprep issues based on updated data release via Google form

roshansh-cmu:
This fix addresses issues raised in #4866 .


sw005320:
Thanks, @roshansh-cmu!

PR #4902, opened 2023-02-01
How2 fix README, incorrect url

roshansh-cmu:
Fixes bug in URL which may not work for all


sw005320:
Thanks for taking care of the update of `black` in addition to your original update...
I'll merge it after the CI check.

sw005320:
https://github.com/espnet/espnet/actions/runs/4064966275/jobs/6999036605#step:7:328
Sorry, @roshansh-cmu, it is not related to your PR, but can you fix it?

b-flo:
Hi @roshansh-cmu ,

>  temporarily patched this by adding the file to flake8 black list, but this may occur in other places as well.

The easiest way would be to omit the rule in  `test_flake8.sh` by adding `--ignore E231` to the flake8 call. Given the rule, I think we can safely discard it.

sw005320:
@roshansh-cmu, I think https://github.com/espnet/espnet/pull/4907 seems to fix the issue.
So, you can make a new PR and add these changes to the latest master.

PR #4905, opened 2023-02-01
standardized inference and number of iterations for mSuperb single lang track

DanBerrebbi:


egs2/msuperb/asr1/conf/tuning/train_asr_fbank_single.yaml
```diff
@@ -30,14 +24,14 @@ num_workers: 4
 batch_type: sorted
 batch_size: 8
 accum_grad: 4
-max_epoch: 50
+max_epoch: 200
```
ftshijt, line 27: I think we can just delete this line?

egs2/msuperb/asr1/conf/tuning/train_asr_s3prl_single.yaml
```diff
@@ -43,14 +36,14 @@ num_workers: 4
 batch_type: sorted
 batch_size: 8
 accum_grad: 4
-max_epoch: 50
+max_epoch: 200
```
ftshijt, line 39: ditto

egs2/msuperb/asr1/run_single_lang.sh
```diff
@@ -39,6 +39,7 @@ for duration in 10min 1h; do
             --feats_type raw \
             --asr_config "${asr_config}" \
             --inference_config "${inference_config}" \
+            --inference_asr_model "valid.loss.ave_5best.pth" \
```
ftshijt, line 42: I think the default one is the ave ones right? In that case, no need to change it

DanBerrebbi: oh yes I forgot about this softlink

ftshijt:
I think you also need some updates to some other scripts (due to the version change)

DanBerrebbi:
> 

Yes I saw it
I'll update after I make sure the current files work well without any bug which I am doing now

ftshijt (approved):
Looks good to me! thanks

PR #4912, opened 2023-02-04
update pyopenjtalk version to 0.3.0

kan-bayashi:


ftshijt:
May I know if we would significantly differ in the g2p method for the version updates? Since we are at the time releasing a benchmark with a prerequisite on it, it would be great if we could stick to a stable version.

kan-bayashi:
I think only the difference is interface. Let me check.
I will modify the code to accept version v0.1.6 ~ v0.3.0.

PR #4913, opened 2023-02-05
MSUPERB setting update

ftshijt:
Use phone as output for msuperb single-lang track in jpn and cmn.


DanBerrebbi:
Looks good to me;

Issue #4908, opened 2023-02-02
Redundant decoder in ESPnetASRModel when ctc_weight is 1.0


pyf98:
**Describe the bug**
When `ctc_weight=1.0`, the ESPnetASRModel should not have a decoder in its parameters, regardless of the decoder config. However, now it will add a default RNN decoder.

Note that the typical way to use CTC only is to set `ctc_weight: 1.0` and remove `decoder` and `decoder_conf` in the training config.

I believe this bug is introduced in a recent commit, because it worked well previously.

https://github.com/espnet/espnet/blob/master/espnet2/asr/espnet_model.py#L146

It should be something like this:
```
            if ctc_weight == 1.0:
                self.decoder = None
            else:
                self.decoder = decoder
```

sw005320:
Thanks!
@ftshijt, can you fix it?
Also, is there a way to add a test to ensure this does not happen?

ftshijt:
I changed the previous behavior mainly because the `self.decoder = None` is not user-aware from the config (e.g., user can still set decoder with anything, but eventually, there is no decoder is initialized). 

For that purpose, I add a `None` decoder and ask the user to set `decoder=None` when they do the initialization. So probably, we may just throw an error (e.g., you should set decoder to none etc.).

What's your opinion on it?

pyf98:
This will affect our previous CTC configs and models.

ftshijt:
I see. So I will add a warning msg and recover the previous options. thanks for bringing up the issue.

pyf98:
Actually, I think the decoder should be `None` by default. The user should specify a decoder when they really want to use one.

ftshijt:
Yeah, I thought that would be better, but the only concern I had at the time is that, it may lead to some issues to other parts (e.g., test functions etc.) Anyway, since we brought up the issue, let's just do it and fixes issues with the change.

PR #4917, opened 2023-02-06
set default none decoder for ASR

ftshijt:
Tries to fix the issues in https://github.com/espnet/espnet/issues/4908

### DIFF
--- a/ci/test_integration_espnet2.sh
+++ b/ci/test_integration_espnet2.sh
@@ -27,7 +27,7 @@ for t in ${feats_types}; do
     for t2 in ${token_types}; do
         echo "==== feats_type=${t}, token_types=${t2} ==="
         ./run.sh --ngpu 0 --stage 6 --stop-stage 13 --skip-upload false --feats-type "${t}" --token-type "${t2}" \
-            --asr-args "--max_epoch=1" --lm-args "--max_epoch=1" --python "${python}"
+            --asr-args "--max_epoch=1 --decoder rnn" --lm-args "--max_epoch=1" --python "${python}"
     done
 done
 echo "==== feats_type=raw, token_types=bpe, model_conf.extract_feats_in_collect_stats=False, normalize=utt_mvn ==="
--- a/ci/test_integration_espnet2.sh
+++ b/ci/test_integration_espnet2.sh
@@ -164,7 +164,7 @@ fi
 if python -c 'import torch as t; from packaging.version import parse as L; assert L(t.__version__) >= L("1.2.0")' &> /dev/null;  then
     cd ./egs2/mini_an4/enh_asr1
     echo "==== [ESPnet2] ENH_ASR ==="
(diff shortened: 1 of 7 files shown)


sw005320:
I added @pyf98 for this PR.

sw005320:
@pyf98, would it solve all issues?
I think making `rnn` as the default decoder sounds weird, and I think it is correct.

sw005320:
Actually, there are a lot of CI errors, @ftshijt

ftshijt:
> Actually, there are a lot of CI errors, @ftshijt

Yeah, sorry that I did not find time to fix the errors. Will try to fix those later

pyf98:
It can solve the issue when we run new experiments using prior CTC configs. However, for previously uploaded models, the issues still exists? For example, the decoder was set to `rnn` by default in this model: https://huggingface.co/pyf98/librispeech_100_ctc_e_branchformer/blob/main/exp/asr_train_asr_ctc_e_branchformer_e12_raw_en_bpe5000_sp/config.yaml#L5190

If we download and load it, the decoder will be set to RNN?

sw005320:
> It can solve the issue when we run new experiments using prior CTC configs. However, for previously uploaded models, the issue still exists? For example, the decoder was set to `rnn` by default in this model: https://huggingface.co/pyf98/librispeech_100_ctc_e_branchformer/blob/main/exp/asr_train_asr_ctc_e_branchformer_e12_raw_en_bpe5000_sp/config.yaml#L5190
> 
> If we download and load it, the decoder will be set to RNN?

Oh, this is complicated.
But in this case, if we use CTC weight for 1.0 during inference, would the result be the same?

pyf98:
I think it does not affect the final result. It just causes some redundant parameters in the model, which is not good. It has an issue for `load_state_dict` when `strict=True`.

pyf98:
I think we can change the default decoder to `none`.

But to keep compatibility, we still need to set the decoder as None in the model code when ctc weight is 1.0.

sw005320:
I see. Good idea.
@ftshijt, can you test it?

ftshijt:
Sure, I agree with Yifan on the potential behaviors. I think the only point here is the initialization with strict loading checks. Do you guys know some of the available pre-trained models with only CTC?

sw005320:
How about this?
https://github.com/espnet/espnet/blob/master/egs2/librispeech_100/asr1/README.md#e-branchformer-with-ctc

sw005320:
@ftshijt, I just want to make sure whether it is working with existing pre-trained CTC models without changing the configurations.
Did you confirm it? (I just want to double-check)

pyf98:
I'm also curious. I don't see the related modifications in the code. How is it solved?

ftshijt:
> I'm also curious. I don't see the related modifications in the code. How is it solved? Am I missing anything?

Sorry, my bad. I did not push that to remote. The current one should prevent issues when loading previous checkpoints that are trained with ctc only.

ftshijt:
(I checked with the `ebranchformer checkpoint` on `mini_an4`)

ftshijt:
This PR is ready to merge

sw005320:
LGTM.
@pyf98, can you also check it?

pyf98:
looks good!

sw005320:
Thanks, @ftshijt!

PR #4920, opened 2023-02-06
Update CHiME-7 Task 1 README.md

popcornell:
I wanted to update the readme with data generation instruction but mergify bot merged my PR

### DIFF
--- a/egs2/chime7_task1/asr1/run.sh
+++ b/egs2/chime7_task1/asr1/run.sh
@@ -117,6 +117,8 @@ if [ ${stage} -le 2 ] && [ ${stop_stage} -ge 2 ]; then
 
     if [ ${dset_name} == dipco ]; then
       channels=2,5,9,12,16,19,23,26,30,33 # in dipco only using opposite mics on each array, works better
+    elif [ ${dset_name} == chime6 ] && [ ${dset_part} == dev ]; then
+      channels=0,3,4,7,8,11,12,15,16,19
     fi
 
     if [ ${dset_part} == train ]; then
@@ -141,7 +143,7 @@ if [ ${stage} -le 3 ] && [ ${stop_stage} -ge 3 ]; then
   asr_train_set=kaldi/train_all_mdm_ihm_rvb_gss
   asr_cv_set=kaldi/chime6/dev/gss # use chime only for validation
   # Decoding on dev set because test is blind for now
-  asr_tt_set="kaldi/chime6_dev/dev/gss/"
+  asr_tt_set="kaldi/chime6/dev/gss/ kaldi/dipco/dev/gss/ kaldi/mixer6/dev/gss/"
   # these are args to ASR data prep, done in local/data.sh
   data_opts="--stage $asr_dprep_stage --chime6-root ${chime6_root} --train-set ${asr_train_set}"
   data_opts+=" --manifests-root $manifests_root --gss_dsets $gss_dsets --gss-dump-root $gss_dump_root"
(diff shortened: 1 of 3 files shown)


sw005320:
Thanks!

egs2/chime7_task1/asr1/run.sh
```diff
@@ -141,7 +143,7 @@ if [ ${stage} -le 3 ] && [ ${stop_stage} -ge 3 ]; then
   asr_train_set=kaldi/train_all_mdm_ihm_rvb_gss
   asr_cv_set=kaldi/chime6/dev/gss # use chime only for validation
   # Decoding on dev set because test is blind for now
-  asr_tt_set="kaldi/chime6_dev/dev/gss/"
+  asr_tt_set="kaldi/chime6/dev/gss/ kaldi/dipco/dev/gss/ kaldi/mixer6/dev/gss/"
```
kamo-naoyuki, line 146: By stage 4 of asr.sh, some recordings in tr and cv sets are removed due to "max_wav_duration" option, and this changes the contents  of `kaldi/chime6/dev/gss`.  Thus, we can't include same data set as cv set in test sets.

popcornell: Unless I made some error this is avoided in the dev set here: https://github.com/espnet/espnet/blob/9c7bde4f1241147fe2ad2c731f6681baf0e0f9a7/egs2/chime7_task1/asr1/run.sh#L138

Note that I still discard some super short segments, but these are practically irrelevant for the final result (<0.2 seconds).  
I still discard them because there are then issues with the WavLM frontend. 
Anyway I will solve it with a proper evaluation script. But again they won't change the WER practically on the dev set. 

Does ESPNet discard them ? This is something I ve not accounted for as I thought the original kaldi data would not be modified. If that is the case I could create two copies. 
But IMO this should be addressed at toolkit level at some point as it is quite an unexpected behaviour.

kamo-naoyuki: >But again they won't change the WER practically on the dev set.

Indee, some audio files are removed from the dev set in this recipe. Please check.

> But IMO this should be addressed at toolkit level at some point as it is quite an unexpected behaviour.

The `valid_set` is originally used in the trainig for validation. If long audio files are not removed, it causes memory error.  I think this is inevitable.

 Normally, we prepare a valid set different set from test sets.

kamo-naoyuki: >  I thought the original kaldi data would not be modified.

Do you mean that the original kalid data is `dump/~` and is it not modified?
 
 Yes, It's modified. Maybe, you thought logn/short files are ignored at the python level, but actually we modify it at the file level.

kamo-naoyuki: I rethinked it with the current `asr.sh` about this problem. To avoid it, we probably should generate another data sets at stage4 (removing short/long stage), e.g. `${train_set}` -> `${train_set}_removed`

Actually, we already generated ${data_feats}/org to keep the original data, but this is not used in the evaluation steps.

@sw005320 Do you have some idea?

popcornell: From a purely theoretical perspective, IMO the best practice maybe would be to do it after you load the manifest, in the python code.

popcornell: https://github.com/espnet/espnet/pull/4942 right now I ve solved by copying the data. 
WER increases to 37.1 BTW. I edited the README.md with the new, worse results. 

I think we can merge that PR, in the meantime I will also re-run the experiments and then I will adjust the README.md and add the ablations that were there before.

Issue #4919, opened 2023-02-06
librosa.util.exceptions.ParameterError: Window size mismatch: 512 != 400 when using streaming transformer model inference on CPU


bmilde:
I've successfully trained a streaming transformer for German with 13000 hours of data and end-to-end punctuation. See https://huggingface.co/speechcatcher/speechcatcher_german_espnet_streaming_transformer_13k_train_size_m_raw_de_bpe1024 . Espnet2 with asr.sh is really nice, thanks for that!

Following https://github.com/espnet/notebook/blob/master/espnet2_streaming_asr_demo.ipynb, inference also works on the training machine (Linux with Cuda). When I was trying to do inference with my model on a Mac mini (M1, arm) I was getting this error though:
 
```

  File "/Users/me/projects/speechcatcher/speechcatcher.py", line 66, in <module>
    recognize("test1.wav")
  File "/Users/me/projects/speechcatcher/speechcatcher.py", line 52, in recognize
    results = speech2text(speech=speech[i*sim_chunk_length:(i+1)*sim_chunk_length], is_final=False)
  File "/Users/me/projects/speechcatcher/speechcatcher_env/lib/python3.10/site-packages/torch/autograd/grad_mode.py", line 27, in decorate_context
    return func(*args, **kwargs)
  File "/Users/me/projects/speechcatcher/speechcatcher_env/lib/python3.10/site-packages/espnet2/bin/asr_inference_streaming.py", line 310, in __call__
    feats, feats_lengths, self.frontend_states = self.apply_frontend(
  File "/Users/me/projects/speechcatcher/speechcatcher_env/lib/python3.10/site-packages/espnet2/bin/asr_inference_streaming.py", line 253, in apply_frontend
    feats, feats_lengths = self.asr_model._extract_feats(**batch)
  File "/Users/me/projects/speechcatcher/speechcatcher_env/lib/python3.10/site-packages/espnet2/asr/espnet_model.py", line 407, in _extract_feats
    feats, feats_lengths = self.frontend(speech, speech_lengths)
  File "/Users/me/projects/speechcatcher/speechcatcher_env/lib/python3.10/site-packages/torch/nn/modules/module.py", line 1130, in _call_impl
    return forward_call(*input, **kwargs)
  File "/Users/me/projects/speechcatcher/speechcatcher_env/lib/python3.10/site-packages/espnet2/asr/frontend/default.py", line 87, in forward
    input_stft, feats_lens = self._compute_stft(input, input_lengths)
[...]

PR #4922, opened 2023-02-07
Use native CPU version of STFT on newer pytorch versions, fix librosa window size < ftt

bmilde:
See discussion in https://github.com/espnet/espnet/issues/4919


sw005320:
@pyf98, can you check it?
I think it looks good to me.

pyf98:
Looks good to me!

sw005320:
Thanks, @bmilde!

PR #4925, opened 2023-02-08
Fixing some issues with chime7-task1 baseline

popcornell:
many thanks to Kamo Naoyuki @kamo-naoyuki . 
I think there was a mistake on the close talk microphones for dipco dev (they were removed for eval). 
I will also address your other useful comments but I am unsure about the sox one.


kamo-naoyuki:
@popcornell About dipco,  this fix works for me. Thank you!

kamo-naoyuki:
I got the following error at stage4

```
Stage 4: Enhance segments using GSS
local/run_gss.sh: line 54: affix[@]: unbound variable
```

I guess it can be fixed by

```
"${affix[@]}" -> ${affix[@]:-}
```

However, I'm not sure why using an array variable  i.e. `affix=()`.

popcornell:
I am using that variable because I want that when channels is undefined it is not passed so GSS uses all available ones

kamo-naoyuki:
> I am using that variable because I want that when channels is undefined it is not passed so GSS uses all available ones.

The following is enough for your case. It is not necessary to use array variable.

```
  affix=
  if ! [ $channels == all  ]; then
    affix+="--channels=$channels"
  fi
```

> It runs fine on my side so this is strange. What does ${affix[@]:-} mean ? I am not a great bash expert

This script oviously can't work with "set -u" if `channels !=  all ` i.e. $channels is empty.

Please check.

- ${foo:-} means return a empty string if foo is undfined
- If you give `"${foo}"` to gss command, it is interpreted as argument even if foo is empty and it causes an arguments error for gss command.  `${foo}` can avoid this problem.

popcornell:
I am starting to add this pretrained model https://huggingface.co/popcornell/chime7_task1_asr1_baseline
WER (%) on (GSS on dev) of 
1. CHiME-6  35.5
2.  DiPCo 39.3 
3. Mixer 6 20.6 

Note that now i discarded sentences < 0.2 seconds (not many on dev) so it is not true WER for the challenge.

egs2/chime7_task1/asr1/run.sh
```diff
@@ -54,17 +54,25 @@ use_word_lm=false
...
 . ./utils/parse_options.sh
 
+
 # ESPNet does not scale parameters with num of GPUs by default, doing it
 # here for you
 asr_batch_size=$(calc_int 128*$ngpu) # reduce 128 bsz if you get OOMs errors
 asr_max_lr=$(calc_float $ngpu/10000.0)
 asr_warmup=$(calc_int 40000.0/$ngpu)
 
+if [ $decode_only == 1 ]; then
```
popcornell, line 72: if decode only i ran data prep and GSS only on development sets

YoshikiMas:
I faced minor issues in `install_dependencies.sh`.
I'm not sure but I think [this line](https://github.com/espnet/espnet/blob/5bf6b5cbffbbee52a94049b38678b72e91c82319/egs2/chime7_task1/asr1/local/install_dependencies.sh#L22) could be `${MAIN_ROOT}/tools/installers/install_s3prl.sh`.
Also, [this line](https://github.com/espnet/espnet/blob/5bf6b5cbffbbee52a94049b38678b72e91c82319/egs2/chime7_task1/asr1/local/install_dependencies.sh#L26) should be `${MAIN_ROOT}/tools/installers/install_gss.sh`.

popcornell:
Indeed thanks Yoshiki !

egs2/chime7_task1/asr1/local/install_dependencies.sh
```diff
@@ -2,6 +2,12 @@
 set -euo pipefail
 [ -f ./path.sh ] && . ./path.sh
 
+
+if ! command conda  &>/dev/null; then
+  echo "Conda comment not found, please follow the instructions on
```
sw005320, line 7:
```suggestion
  echo "Conda command not found, please follow the instructions on
```

sw005320:
LGTM
Once the CI check is passed, I'll merge this.

popcornell:
Note that I will need to change the README.md too probably based on what we decide tomorrow about the channel selection stuff.

popcornell:
Rules are not clear not if https://github.com/espnet/espnet/blob/ecf3bd3deae09f8040371387468f6567c2fd0a1e/egs2/chime7_task1/asr1/run.sh#L129 should be allowed or not. IMO it should be allowed but then we have to distinguish it from automatic or manual domain classification. 
The model remains unique but you just select a bunch of microphones with some a priori knowledge for example of the topology of the array (e.g. linear in chime6 and i use only outer mics)

sw005320:
OK, please make additional PRs.

PR #4926, opened 2023-02-09
fix extracted feature dummy generation

roshansh-cmu:


sw005320:
Thanks, @roshansh-cmu!!
Is it possible to create a simple and small integration test (only for this stage) to check the feature input case?

sw005320:
e.g, we can add this case in https://github.com/espnet/espnet/blob/master/ci/test_integration_espnet2.sh#L18-L25

roshansh-cmu:
We should be able to add a test where we first do fbank_pitch and copy over that feats.scp and try to use "extracted" features. 

What this PR fixes though is cases where we have a segments file ( recording ID != utterance ID ; which means AN4 might not be the best test ) or where wav.scp already exists.

sw005320:
I think we can just prepare small files instead of using an4 files.
If the data size is small (just one or a few feature files with segment, it should be), we can also include it in our package.

roshansh-cmu:
Will think on it and maybe make a separate PR for the integration test ? I think we need a dataset other than an4, one where rec-id != utt-id. It seems challenging to think of public datasets that follow that. Do you have any suggestions ?

sw005320:
We even do not need to prepare an actual dataset.
We just create one feature (randomly sampling or one example from some recording) and a segment file.
If we need, we can also prepare other dummy files.
This is just to be used for this test.
We do not need to go to the other stages.

PR #4927, opened 2023-02-09
IAM OCR recipe updates

kenzheng99:
small update to the IAM OCR recipe, with pretrained model link and retrained results.

(forgot to do this last semester oops)


ftshijt:
Thanks for the update! Looks great to me

PR #4928, opened 2023-02-09
refactor make_pad_mask to be ONNX compatible

kenzheng99:
Refactor `make_pad_mask` function into `make_pad_mask_simple` when the `xs` input is not needed. This simplified implementation supports JIT tracing which is required for ONNX export.

This PR replaces https://github.com/espnet/espnet/pull/4821. I have removed the attention change, as I found it to not be strictly necessary to support the ONNX export. I have also removed our testing code to unblock CI tests, may consider adding some of it back as proper espnet tests later.

### DIFF
--- a/espnet/nets/pytorch_backend/nets_utils.py
+++ b/espnet/nets/pytorch_backend/nets_utils.py
@@ -60,6 +60,26 @@ def pad_list(xs, pad_value):
 
     return pad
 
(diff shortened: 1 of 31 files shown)


Masao-Someki:
I will continue working on this content on #5326, so this PR can be closed.

sw005320:
#5326

PR #4929, opened 2023-02-09
Update the chunk iterator for the TSE task

Emrys365:
This PR updates the chunk iterator to make it more general for different tasks.

I added a new argument `excluded_key_prefixes` for [ChunkIterFactory](https://github.com/espnet/espnet/blob/master/espnet2/iterators/chunk_iter_factory.py) to allow certain keys to be ignored when checking the length consistency for each sample:
```diff
            for key in sequence_keys:
-               if len(batch[key]) != len(
-                   batch[sequence_keys[0]]
-               ) and not key.startswith("enroll_ref"):
+               if key.startswith(self.excluded_key_prefixes):
+                   # ignore length inconsistency for `excluded_key_prefixes`
+                   continue
+               if len(batch[key]) != len(batch[sequence_keys[0]]):
                    raise RuntimeError(
                        f"All sequences must has same length: "
                        f"{len(batch[key])} != {len(batch[sequence_keys[0]])}"
```

This can be useful for audio-to-audio tasks like target speaker extraction (TSE), where additional features are required as input, which do not necessarily have the same length as the input/target signal.

> In our previous discussion in the SE meeting, we decided to make a new chunk iterator for such purpose. But after some  coding, I find the current minimal changes in `espnet2/iterators/chunk_iter_factory.py` are not so bad.
>
> If you still feel it necessary to make a new script, I will do it.


espnet2/iterators/chunk_iter_factory.py
```diff
@@ -86,6 +88,11 @@ def __init__(
         self.batch_size = batch_size
         self.seed = seed
         self.shuffle = shuffle
+        self.excluded_key_pattern = (
+            "(" + "[0-9]*)|(".join(excluded_key_prefixes) + "[0-9]*)"
```
sw005320, line 92: Are you only assuming the cases of numbers are appended?
If so, it should be documented in `add_argument` or other places.
Adding a comment to the config is also informative.

PR #4931, opened 2023-02-09
Update Muskits

A-Quarter-Mile:
Hi @ftshijt, here are new updates:

- change frame & syllable level feature calculation.
- Update naive_rnn & xiaoice.

Some details in inference and collect_states needs to be discussed.
Auto-regressive model and PHONEix will be updated after that.


ftshijt:
Please complete the CI tests this time

PR #4932, opened 2023-02-10
Update inference and training setting for mSuperb multilingual model

guapaQAQ:


ftshijt:
Many thanks! looks great to me

PR #4941, opened 2023-02-15
Fix device mismatch error in GPU decoding with PyTorch 1.13

pyf98:
Starting from PyTorch 1.13, we will encounter an error if a cpu tensor is indexed by a cuda tensor: https://github.com/pytorch/pytorch/releases/tag/v1.13.0

image

This affects our standard GPU decoding. (Not sure if it affects other decoding algorithms.)


sw005320:
Thanks, @pyf98!

PR #4944, opened 2023-02-17
Issue if train_set or valid_set are included in test sets

kamo-naoyuki:
Issue:

If using a test_set as the train_set or valid_set in asr.sh, the test set is modified by `stage 4 Remove long/short utt`

Modify:

~~- Current behaviour: `${data_feats}/org/${dset}` at stage 3 -> `${data_feats}/${dset}` at stage 4~~
~~- In this PR:  `${data_feats}/${dset}` at stage 3 -> `${data_feats}/${dset}_flt` at stage 4~~

I only modified asr.sh in this PR, but all templates has same problem (due to my bad original template script...)

@sw005320


kamo-naoyuki:
I changed my mind.

I implemented `--filtered_train_key_text` and `--filtered_valid_key_text` for `espnet2/bin/*_train.py`. 

With giving a text file containing IDs to be filtered, 
The samples specified by this option are excluded from the training. 

e.g. 

- wav.scp:

```
IDa a.wav
IDb b.wav
IDc b.wav
```

- filtered_key.txt

```
IDb
```

In this case, `IDb` is excluded, and `IDa` and `IDc` remain for the training.

 I also changed `asr.sh` to create `filtered_key.txt` at stage 4 instead of  creating a new dataset.

kamo-naoyuki:
I  changed my mind again.

Filtering short/long utterances by the option of the python tool is better way as a viewpoint for a smart recipe, but it could make some overhead  for start at the startup.

Creating another dataset is a dirty way, but actually efficient for training speed.

I'll think about it.

PR #4945, opened 2023-02-19
Fix '__floordiv__ is deprecated' warnings 

fujimotos:
When we do this with `torch >= 0.12.1`:

    speech2text = Speech2Text(...)

PyTorch emits a couple of deprecation warnings:

    UserWarning: __floordiv__ is deprecated, and its behavior will
    change in a future version of pytorch. It currently rounds toward 0
    (like the 'trunc' function NOT 'floor'). This results in incorrect
    rounding for negative values. To keep the current behavior, use
    torch.div(a, b, rounding_mode='trunc')

Fix the problematic code accordingly, so that users won't see confusing warnings.

### DIFF
--- a/espnet/nets/batch_beam_search.py
+++ b/espnet/nets/batch_beam_search.py
@@ -4,9 +4,12 @@ import logging
 from typing import Any, Dict, List, NamedTuple, Tuple
 
 import torch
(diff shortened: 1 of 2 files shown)


sw005320:
Thanks, @fujimotos!

PR #4947, opened 2023-02-20
Add resource limit check before submitting jobs

tjysdsg:
This PR adds an automatic resource limit check based on the number of hours the job is set to run for.

- If not specified in slurm.conf, the default limitation of CPU and GPU jobs is 100 hours.
- The user can specified max number of hours a job can run for in slurm.conf or command line using `gpu_limit` and `cpu_limit` options.
- slurm.pl will cancel job submission if the job time exceeds the limit
    - The string format of `time` option is parsed according to https://slurm.schedmd.com/sbatch.html
    - When calculating job time, the script will multiply `num_nodes` and `num_threads` with the `time` option.
    - For example, if `time=2-10:00:00`, `num_threads=8`, and `num_nodes=2`, the total run time is 928 hours

There are several things that need to be addressed for this implementation to be merged:

- Where to put the new slurm.pl. Original source is from kaldi
- How to implement this feature for SGE and other cluster software
- How to test this in CI

Thanks @siddalmia for the suggestion!

### DIFF
--- a/utils/slurm.pl
+++ b/utils/slurm.pl
@@ -1,0 +1,708 @@
+#!/usr/bin/env perl
+use strict;
+use warnings;
+
+# Copyright 2012  Johns Hopkins University (Author: Daniel Povey).
+#           2014  Vimal Manohar (Johns Hopkins University)
+#           2015  Johns Hopkins University (Yenda Trmal <jtrmal@gmail.com>>)
+# Apache 2.0.
+
+use File::Basename;
+use Cwd;
+use Getopt::Long;
+
+# slurm.pl was created from the queue.pl
+# queue.pl has the same functionality as run.pl, except that
+# it runs the job in question on the queue (Sun GridEngine).
+# This version of queue.pl uses the task array functionality
+# of the grid engine.  Note: it's different from the queue.pl
+# in the s4 and earlier scripts.
+
+# The script now supports configuring the queue system using a config file
+# (default in conf/queue.conf; but can be passed specified with --config option)
+# and a set of command line options.
+# The current script handles:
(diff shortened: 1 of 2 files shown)


utils/slurm.pl
```diff
@@ -0,0 +1,708 @@
...
+
+my $max_time_limit = $cpu_limit;
+if ($use_gpu > 0) {
+  $max_time_limit = $gpu_limit;
+}
+if ($hours > $max_time_limit) {
+    print STDERR "$0: Job time ($hours hours) exceeds the limit ($max_time_limit hours), ";
+    print STDERR "cancelling job submission\n";
+    exit(1);
+}
```
tjysdsg, line 375: These lines are modified

utils/slurm.pl
```diff
@@ -0,0 +1,717 @@
...
```
tjysdsg, line 384: These lines were modified

utils/slurm.pl
```diff
@@ -0,0 +1,732 @@
...
```
tjysdsg, line 399: Changes

pyf98:
I didn't read it carefully, but just note that CPU/GPU hours are not the only factor that affects the charged services units. Some clusters allow requesting more memory for a single cpu core and the units to be charged will increase proportionally.

kamo-naoyuki:
It's better to request this PR to kaldi repo. We are not able to maintain slurm.pl.

tjysdsg:
> It's better to request this PR to kaldi repo. We are not able to maintain slurm.pl.

Thanks for pointing, I plan to implement the feature using launch.py instead

PR #4953, opened 2023-02-21
Fix in gen_task1_data.sh from CHiME7

boeddeker:
I got the error:

> local/gen_task1_data.sh: line 36: [-d: command not found 

and PyCharm complained:

> Couldn't parse this test expression. Fix to allow more checks.
> See SC1073.
> You need a space after the [ and before the ].
> See SC1035.

I think someone forgot to add whitespaces, otherwise it is not valid bash code.


sw005320:
@popcornell, is it already considered in your PR, or should I merge it?

sw005320:
OK, thanks!

PR #4958, opened 2023-02-22
Add E-Branchformer results in some recipes

pyf98:
This PR adds E-Branchformer results in the following recipes:

- Aidatatang: AED
- Aishell: CTC
- JSUT
- LibriSpeech 100: Transducer


sw005320:
Thanks a lot!

PR #4964, opened 2023-02-24
Add 'six' as a required Python module

fujimotos:
When we run `pip install espnet` in a clean environment:

```console
$ python3 -m venv venv
$ . ./venv/bin/activate
$ pip install espnet
```

it results in a missing module exception:

```python3
>>> from espnet2.bin.asr_inference import Speech2Text
  File "<stdin>", line 1, in <module>
  ...
  File "./espnet/nets/pytorch_backend/rnn/attentions.py", line 5, in <module>
    import six
ModuleNotFoundError: No module named 'six'
```

This is because ESPnet depends on `six`, but forgot to add it to the
requirement list. Fix it.


sw005320:
Thanks a lot!

Issue #4940, opened 2023-02-15
[Question] How to properly train-infer transformer on padded batches of sequences?


Penchekrak:
I wonder how to tackle the following problem, which arise when I train on batches of padded sequences:
- [relative positional encoding](https://github.com/espnet/espnet/blob/master/espnet/nets/pytorch_backend/transformer/embedding.py#L260) is dependent on sequence length and for padded batch it is only valid for longest sequence in batch
- `rel_shift` [operation](https://github.com/espnet/espnet/blob/master/espnet/nets/pytorch_backend/transformer/attention.py#L237) also seem to work correctly only for longest sequence in intermediate attention matrix.
Suppose now, that training on padded batches, without any modifications to the above code, went fine: when I perform inference of such model on individual sequences, those positional encodings differ from what model have seen during training. Can this discrepancy be somehow mitigated?

sw005320:
@pengchengguo, can you answer this question?
I'm also curious.

Penchekrak:
Let me elaborate slightly on this issue.
Consider padded batch with following exemplar mask:
![image](https://user-images.githubusercontent.com/1496107/219317699-2e3da85e-d044-434f-9ac8-04b53120ea31.png)
masked relative positional encoding for batched computation looks like (only zeroth channel is displayed here and below)
![image](https://user-images.githubusercontent.com/1496107/219317935-5f44772a-bb7d-4d8f-abf0-cb490d9c1249.png)
whereas positional encoding for individual element-wise computation after appropriate padding and stacking together looks like 
![image](https://user-images.githubusercontent.com/1496107/219318383-da07effb-1562-4f58-bf65-3a03e30b7db1.png)
For `matrix_bd` inside attention layer it is more or less the same story: 
`rel_shift` output for batched matrices looks like (again, only zeroth channel, non-vanishing entries of tensors are displayed)
![image](https://user-images.githubusercontent.com/1496107/219318779-d351f74f-2063-4081-b443-55af93d2f383.png)
(which looks closer to what is described in original paper)
but for stacked matrices for element-wise computation it is
![image](https://user-images.githubusercontent.com/1496107/219319015-e246d50e-5790-4173-80cb-f2989db14c60.png)


I believe this can affect model performance considerably, so I'm eager to find, how to make individual and batched computations more similar

pengchengguo:
Hi @Penchekrak,

I am trying to follow your comments and sorry I do not understand how to get the third figure.
[...]

PR #4966, opened 2023-02-25
Padded sequences fix issue #4940

Penchekrak:


sw005320:
Thanks a lot, @Penchekrak!
Did you test this in our standard conformed setup?
I'm curious about the speed and performance especially for the long audio (>20 seconds)

Penchekrak:
> Thanks a lot, @Penchekrak! Did you test this in our standard conformed setup? I'm curious about the speed and performance especially for the long audio (>20 seconds)

@sw005320, thank you for taking a look at my contribution. No, I did not test those changes with regard to performance (both in terms of computational demand and training quality) thoroughly yet. My intuition, however, that this will decrease both speed and quality, as it involves heavier computations and introduces much less noise into training (noise being very helpful to prevent overfitting). I assume that this code should be only used in cases where computational speed is much less important than consistency between inference and training regimes. There is still a lot of work to be done (ie testing, benchmarking, etc) and I'm open to your recommendations in this regard as I want to push this pull request to mergeable state :)

PR #4972, opened 2023-02-28
Fix a 'ref_channel'-related issue in espnet2/bin/enh_inference.py

Emrys365:
Fix the issue reported in https://github.com/espnet/espnet/discussions/4970.


sw005320:
Thanks for the quick fix!

Issue #4973, opened 2023-02-28
[BUG]there are incompatible upgrade in parratio==6.0.0


yt605155624:
please check https://github.com/timmahrt/praatIO/issues/54

timmahrt:
@yt605155624 Hello, I've created a PR that should fix the issue.

I'm not able to test the PR so I'm not familiar with espnet. Can you test it?

Alternatively you should be able to resolve the issue with:
```
pip uninstall praatio
pip install praatio==5.1.1
```

yt605155624:
I'm not the maintainer neither， PaddleSpeech's users report this bug to me，and I have noticed this to espnet and  MFA，I thought the maintainer will review  your PR soon @kan-bayashi

PR #4978, opened 2023-03-03
fix: resolve upgrade issues with praatio 6.0; lock praatio version

timmahrt:
This is a minor fix. It resolves https://github.com/espnet/espnet/issues/4973

Before this fix:
- users cannot use espnet's code that uses praatio (1 script)

After this fix:
- the affected script can be used
- the version of praatio is locked to avoid similar issues in the future

TODO:
- I am the maintainer of praatio. I'm not familiar with espnet, so its helpful if someone can validate the fix. 🙇


setup.py
```diff
@@ -72,7 +72,7 @@
         "fastdtw",
         "nara_wpe>=0.0.5",
         "sacrebleu>=1.5.1",
-        "praatio>=5.1.1",  # for librispeech phoneme alignment
+        "praatio >=6,<7",  # for librispeech phoneme alignment
```
kan-bayashi, line 75:
```suggestion
        "praatio>=6,<7",  # for librispeech phoneme alignment
```

simpleoier (approved):
Thanks! The updates works perfectly in my tests. LGTM!
(sorry for my late review.)

PR #4984, opened 2023-03-06
test

kamo-naoyuki:


kamo-naoyuki:
Would you happen to have any idea about this error?

https://github.com/espnet/espnet/actions/runs/4340074973/jobs/7578272513#step:18:2537

This error may happen only for enh tse task.

@ShigekiKarita

ShigekiKarita:
hmm no idea. Perhaps, multiple processes can write .coverage at the same time?

kamo-naoyuki:
Thanks, Shigeki. Could you tell me how we generate .coverage in CI? I don't understand the setting basically.

kamo-naoyuki:
Is the following command generating the .converage, right?

https://github.com/espnet/espnet/blob/b5b2b11e2e234be8c10549346dbd33e8c278fa4a/ci/test_integration_espnet2.sh#L8

kamo-naoyuki:
I tested running `utils/run.pl JOB=1:10 JOB.log coverage run --append empty.py`. 

I couldn't reproduce `Couldn't use data file '.../.coverage': no such table: meta`, but it is sometimes successful, and sometimes is failed with `sqlite3.OperationalError: no such table: coverage_schema` or `Couldn't use data file '.../.coverage': no such table: coverage_schema`
Potentially, parallel running for `coverage run --append` causes the problems.

I'm not sure these errors come from the same reason, but I found that once .coverage was generated successfully, I didn't get the error after it.

I'll try to modify the script to generate `.coverage` at the first in a single process.

PR #4986, opened 2023-03-06
Change to use Ubuntu-latest instead of Ubuntu-18.04 in CI

kamo-naoyuki:


kamo-naoyuki:
```
This is a scheduled Ubuntu-18.04 brownout. The Ubuntu-18.04 environment is deprecated and will be removed on April 1st, 2023
```

I changed  to use Ubuntu-latest instead of Ubuntu-18.04

PR #4997, opened 2023-03-11
Modify format_wav_scp.py to support PCM of uint8, int32, float32, float64, etc.

kamo-naoyuki:
### Issue of the current script

- `np.int16` is specified for the soundfile.read: `soundfile.read(-. dtype=np.int16)`. This indicates some problems if the PCM data has uint8 or float type.
- Using kaldiio for segmentation, but kaldiio requires that the input file is wav file

### The modification of PR

- I chanced to to `soundfile.read()` withtou specifying the dtype. Note that `soundfile.read()` always returns float64 array in spite of the dtype of input pcm, and `soundfile.write` write a float64 array in int16 pcm.
- I implemented a segmentation module independent from kaldiio. This realizes supporting flac format for inputting.

e.g. The following `wav.scp` with `segments` also works.

```
ID1 cat a.flac |
```


kamo-naoyuki:
I'll merge this PR if there are no comments.

PR #5003, opened 2023-03-13
fix s3prl upstream attribute bug

jwrh:
fix s3prl upstream attribute bug: variable `upstream` of the class `S3PRLUpstream` does not have an `model` attribute and thus the `if` statement that is responsible for setting `encoder.layerdrop = 0` never gets evaluated to `True`. Quick Fix includes using the `upstream` attribute of `upstream` to correctly get attribute values.


sw005320:
Thanks for your quick action!

PR #5004, opened 2023-03-14
Fix Whisper tokenizer CI error

slSeanWU:
Install `openai-whisper==20230308` to use huggingface tokenizer


sw005320:
Thanks for the quick fix!

PR #5009, opened 2023-03-15
Fix typeguard version

silvanocerza:
On 14th of March version `3.0.0` of `typeguard` has been released, it contains breaking changes so there are import failures importing certain `espnet` modules that also import `typeguard`.

This PR fixes the issue by pinning the version to latest non breaking one.

I decided to go this way since it's easist fix, updating all the imports would take much more time and I don't have enough knowledge about the codebase to do it correctly.


kamo-naoyuki:
Thanks!

PR #5020, opened 2023-03-16
Modify asr.sh

kamo-naoyuki:
This PR includes my self-answer to #4944.

- I changed `asr.sh` to raise an error if `--train_set` equal to `--valid_set`,  or `--train_set` is also included in `--test_sets`.
- If `--valid_set` is included in `--test_sets`, `--eval_valid_set` option is enabled. 
-  if `--eval_valid_set` is enabled, `dump/org/${valid_set}` is evaluated at the decoding stage instead of `dump/${valid_set}`

Someone might still think we should filter the long/short utterances in the training python process, but finally, I concluded it's a bad idea. Please let me go in this direction.

I also changed the behavior of the `--skip_train` option:

- Before this PR: The data preparation for `${train_set}` and `${valid_set}` still works. This is inconvinient if using a pre-trained model and evaluation is only required.
- After this PR:  The data preparation for `${train_set}` and `${valid_set}` can be also skipped.

Finally, I added a new feat-type: `--feats_type raw_copy`.  This is almost the same as `--feats_type raw_copy`, but it can skip `format_wav_scp.py`.  

This type is useful if the file format is already correct.　Sometimes, we create a new evaluation set from an existing dataset, e.g.  applying a speech enhancement method. In this case.  the data set may be already correct, so we might want to skip 
`format_wav_scp.py`.
 
Please note that if an user specifies `--feats_type raw_copy`, the user is responsible to guaranteed that the data format follows correctly our requirements.

### DIFF
--- a/ci/test_integration_espnet2.sh
+++ b/ci/test_integration_espnet2.sh
@@ -27,6 +27,9 @@ token_types="bpe char"
 for t in ${feats_types}; do
     ./run.sh --stage 2 --stop-stage 4 --feats-type "${t}" --python "${python}"
 done
+cp -r dump/raw data/
+./run.sh --stage 2 --stop-stage 4 --feats-type "raw_copy" \
+    --train_set raw/train_nodev --valid_set raw/train_dev --test_sets raw/test --python "${python}"
 for t in ${token_types}; do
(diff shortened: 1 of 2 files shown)


popcornell:
This is great !

PR #5049, opened 2023-03-22
Add streaming speech enhancemnt inference.

LiChenda:
Hi,
I added `espnet2/bin/enh_inference_streaming.py` to support the streaming inference of some causal enh-models.

Now, the separators can implement a `forward_stream` method (a new abstract method in Enh Separator), and then it can be called in `espnet2/bin/enh_inference_streaming.py`.
Currently, I implement the `forward_stream` function for three separators:

1.  TCN (Consistency is OK, performance needs to be optimized)
2.  LSTM (Consistency is OK)
3.  SkiM (Consistency is OK, currently only  support `no segment padding` mode, `segment padding` will be updated soon)

Some separated waves (upper is from old sequence inference, lower is from the streaming inference):

image

Future works:

- [x] Prepare a streaming demo in Colab notebook
- [ ] Performance optimization and support more separators.

------

Updated: 

I prepared a streaming demo for speech separation. The notebook demo requires the new features of this PR, so I'll update it to the [demo repo](https://github.com/espnet/notebook) when this PR gets merged.

Now, anyone with the following link can post comments on it, it would be great if you could leave your comments and help improve it.:
https://colab.research.google.com/drive/17vd1V78eJpp3PHBnbFE5aVY5uMxQFL6o?usp=sharing


LiChenda:
Hi, I also prepared a streaming demo for speech separation. The notebook demo requires the new features of this PR, so I'll update it to the [demo repo](https://github.com/espnet/notebook) when this PR gets merged.

Now, anyone with the following link can post comments on it, it would be great if you could leave your comments and help improve it.:
https://colab.research.google.com/drive/17vd1V78eJpp3PHBnbFE5aVY5uMxQFL6o?usp=sharing

sw005320:
Cool!
Would it be better to include the installation part?

image

LiChenda:
> Cool! Would it be better to include the installation part?
> 
> image

Done! Now you @sw005320  can run it in the Colab notebook. But the RTF is about 2 with the Colab's runtime... In my local machine, it's smaller than 1 😂. Seems we need more optimization.

espnet2/bin/enh_inference_streaming.py
```diff
@@ -0,0 +1,545 @@
...
+
+        # b. Enhancement/Separation Forward
+        frame_feature, f_lens = self.enh_model.encoder(speech_mix, lengths)
+
+        (
+            frame_out,
+            self.streaming_states,
+            _,
+        ) = self.enh_model.separator.forward_streaming(
+            frame_feature, self.streaming_states
```
Emrys365, line 210: Better to also add `f_lens` as an argument as some models may require such information.

espnet2/bin/enh_inference_streaming.py
```diff
@@ -0,0 +1,545 @@
...
+        lengths = speech_mix.new_full(
+            [batch_size], dtype=torch.long, fill_value=speech_mix.size(1)
+        )
+
+        # a. To device
+        speech_mix = to_device(speech_mix, device=self.device)
+        lengths = to_device(lengths, device=self.device)
+
+        # b. Enhancement/Separation Forward
+        frame_feature, f_lens = self.enh_model.encoder(speech_mix, lengths)
```
Emrys365, line 203: Do we need to consider possible buffering in the encoder and decoder as well if we feed the frames one-by-one? I mean if the encoder/decoder is also a multi-layer module, it might be wrong to simply use the default forward function.

espnet2/bin/enh_inference_streaming.py
```diff
@@ -0,0 +1,545 @@
...
+            _,
+        ) = self.enh_model.separator.forward_streaming(
+            frame_feature, self.streaming_states
+        )
+
+        frame_out = torch.cat(frame_out, 0)
+
+        lengths = torch.cat([lengths, lengths])
+
+        waves = self.enh_model.decoder(frame_out, lengths)[0].unbind(dim=0)
```
Emrys365, line 217: ditto

espnet2/bin/enh_inference_streaming.py
Emrys365: Certain cases may make this problem critical. For example, if an STFT encoder with `center=True` is used, it will apply zero-paddings to both size and the output will be multiple frames.

espnet2/bin/enh_inference_streaming.py
```diff
@@ -0,0 +1,545 @@
...
+        assert isinstance(batch, dict), type(batch)
+        assert all(isinstance(s, str) for s in keys), keys
+        _bs = len(next(iter(batch.values())))
+        assert len(keys) == _bs, f"{len(keys)} != {_bs}"
+        batch = {k: v for k, v in batch.items() if not k.endswith("_lengths")}
+
+        speech = batch["speech_mix"]
+
+        # split continuous speech into small chunks to simulate streaming
+        speech_sim_chunks, rest_pad = split_audio(speech, frame_size, hop_size)
```
Emrys365, line 356: We need to be careful when manually generating the frames because it should be consistent with the procedure defined in the encoder. For example, depending on the padding mode of the encoder, we may need to apply some paddings (can be non-zero) to the left/right side of the original input signal.

Emrys365: If we do that manually, then we should ignore the builtin padding operations in the actually encoder module. Things become too complicated...

espnet2/bin/enh_inference_streaming.py
```diff
@@ -0,0 +1,545 @@
...
+
+        (
+            frame_out,
+            self.streaming_states,
+            _,
+        ) = self.enh_model.separator.forward_streaming(
+            frame_feature, self.streaming_states
+        )
+
+        frame_out = torch.cat(frame_out, 0)
```
Emrys365, line 213: Could you annotate the shape of the output?

espnet2/bin/enh_inference_streaming.py
```diff
@@ -0,0 +1,545 @@
...
+
+        lengths = torch.cat([lengths, lengths])
```
Emrys365, line 215: I cannot understand why **two** `lengths` are concatenated here.

espnet2/bin/enh_inference_streaming.py
```diff
@@ -0,0 +1,545 @@
...
+            f"model must inherit {AbsESPnetModel.__name__}, but got {type(model)}"
+        )
+    model.to(device)
+    if model_file is not None:
+        if device == "cuda":
+            # NOTE(kamo): "cuda" for torch.load always indicates cuda:0
+            #   in PyTorch<=1.4
+            device = f"cuda:{torch.cuda.current_device()}"
+        model.load_state_dict(torch.load(model_file, map_location=device))
+    return model
```
Emrys365, line 78: You could simply import these functions from `espnet2.bin.enh_inference` to reduce the number of lines.

espnet2/bin/enh_inference_streaming.py
```diff
@@ -0,0 +1,545 @@
...
+
+class SeparateSpeechStreaming:
+    """SeparateSpeech class
+
+    Examples:
+        >>> import soundfile
+        >>> separate_speech = SeparateSpeech("enh_config.yml", "enh.pth")
+        >>> audio, rate = soundfile.read("speech.wav")
+        >>> separate_speech(audio)
+        [separated_audio1, separated_audio2, ...]
```
Emrys365, line 89: The examples should be updated.

espnet2/bin/enh_inference_streaming.py
```diff
@@ -0,0 +1,545 @@
...
+    # audio: B, T
+    batch_size, audio_len = audio.shape
+
+    rest = frame_size - (hop_size + audio_len % frame_size) % frame_size
+
+    if rest > 0:
+        pad = torch.zeros((batch_size, rest), dtype=audio.dtype).to(audio.device)
+        audio = torch.cat([audio, pad], 1)
+
+    pad_aux = torch.zeros((batch_size, hop_size), dtype=audio.dtype).to(audio.device)
```
Emrys365, line 419:
```suggestion
    if rest > 0:
        pad = torch.zeros((batch_size, rest), dtype=audio.dtype, device=audio.device)
        audio = torch.cat([audio, pad], 1)

    pad_aux = torch.zeros(
        (batch_size, hop_size), dtype=audio.dtype, device=audio.device
    )
```

espnet2/bin/enh_inference_streaming.py
```diff
@@ -0,0 +1,545 @@
...
+
+    """
+
+    def __init__(
+        self,
+        train_config: Union[Path, str] = None,
+        model_file: Union[Path, str] = None,
+        inference_config: Union[Path, str] = None,
+        ref_channel: Optional[int] = None,
+        normalize_output_wav: bool = False,
```
Emrys365, line 99: This argument seems unused.

espnet2/bin/enh_inference_streaming.py
Emrys365: In cases where a multi-layer encoder/decoder is used, it might be difficult to define an exact `frame_size` or `hop_size`.

espnet2/enh/layers/skim.py
```diff
@@ -295,19 +328,86 @@ def _padfeature(self, input):
...
+                    dtype=input_frame.dtype,
+                ),
+                torch.zeros(
+                    1,
+                    B,
+                    self.hidden_size,
+                    device=input_frame.device,
+                    dtype=input_frame.dtype,
+                ),
+            )
```
Emrys365, line 352:
```suggestion
        def empty_seg_states():
            shp = (1, B, self.hidden_size)
            return (
                torch.zeros(*shp, device=input_frame.device, dtype=input_frame.dtype),
                torch.zeros(*shp, device=input_frame.device, dtype=input_frame.dtype),
            )
```

espnet2/enh/separator/skim_separator.py
```diff
@@ -129,6 +129,82 @@ def forward(
...
+    )
+    separator.eval()
+
+    print(
+        "Number of parameters: "
+        f"{humanfriendly.format_size(sum(p.numel() for p in separator.parameters()))}"
+    )
+
+    input_feature = torch.randn((1, SEQ_LEN, 128))
+    ilens = torch.LongTensor([SEQ_LEN, SEQ_LEN, SEQ_LEN])
```
Emrys365, line 187:
```suggestion
    ilens = torch.LongTensor([SEQ_LEN])
```

Emrys365 (commented):
Could you move some of the tests to `test/espnet2/enh`?

espnet2/bin/enh_inference_streaming.py
LiChenda: Add `forward_streaming` for the encoder and decoder now. 

> In cases where a multi-layer encoder/decoder is used, it might be difficult to define an exact frame_size or hop_size.
It can be determined by the ABSEncoder and ABSEncoder abstract function:
```
  @property
  def frame_size(self) -> int:
      raise NotImplementedError

```

espnet2/bin/enh_inference_streaming.py
LiChenda: Done.

espnet2/bin/enh_inference_streaming.py
LiChenda: Still, keep it manually outside the model.  Builtin padding is ignored by new `forward_streaming` in the encoder and decoder.

espnet2/bin/enh_inference_streaming.py
LiChenda: Done.

espnet2/bin/enh_inference_streaming.py
LiChenda: Removed.

espnet2/bin/enh_inference_streaming.py
LiChenda: Done.

espnet2/bin/enh_inference_streaming.py
LiChenda: Done.

espnet2/bin/enh_inference_streaming.py
LiChenda: Add  to unit tests.

espnet2/enh/encoder/stft_encoder.py
```diff
@@ -54,3 +66,41 @@ def forward(self, input: torch.Tensor, ilens: torch.Tensor):
...
+            B, 1, F
+        """
+
+        assert (
+            input.dim() == 2
+        ), "forward_streaming only support for single-channel input currently."
+
+        windowed = self._apply_window_func(input)
+
+        feature = torch.fft.fft(windowed)
```
Emrys365, line 99: We should condition the operation on the `one_sided` argument:
```python
feature = torch.fft.rfft(windowed) if self.onesided else torch.fft.fft(windowed)
```

espnet2/enh/encoder/stft_encoder.py
```diff
@@ -54,3 +66,41 @@ def forward(self, input: torch.Tensor, ilens: torch.Tensor):
...
+        if self.stft.onesided:
+            feature = feature[..., 0 : -(self.n_fft) // 2 + 1]
```
Emrys365, line 101: We can condition the operation on the `one_sided` argument:
```python
feature = torch.fft.rfft(windowed) if self.stft.onesided else torch.fft.fft(windowed)
```

test/espnet2/enh/decoder/test_stft_decoder.py
```diff
@@ -58,3 +60,24 @@ def test_STFTDecoder_invalid_type(
...
         y, ilens = decoder(real, x_lens)
+
+
+def test_conv_enc_dec_streaming():
+    input_audio = torch.randn((1, 16000))
+    ilens = torch.LongTensor(
+        [
+            16000,
+        ]
+    )
```
Emrys365, line 71:
```suggestion
    ilens = torch.LongTensor([16000])
```

espnet2/enh/encoder/stft_encoder.py
```diff
@@ -54,3 +66,41 @@ def forward(self, input: torch.Tensor, ilens: torch.Tensor):
...
+        feature = feature.unsqueeze(1)
+
+        feature = ComplexTensor(feature.real, feature.imag)
```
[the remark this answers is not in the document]
LiChenda: Just keep the output type consistent with forward.

espnet2/bin/enh_inference_streaming.py
```diff
@@ -0,0 +1,509 @@
...
+    audio = torch.cat([pad_aux, audio, pad_aux], 1)
+
+    new_len = audio.shape[1]
+
+    audio = [
+        audio[:, i * hop_size : i * hop_size + frame_size]
+        for i in range((new_len - frame_size) // hop_size + 1)
+    ]
+
+    return audio, rest
```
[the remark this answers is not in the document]
LiChenda: I tried your frame segment implementation, and although the number of frames is the same as the `torch.stft`, the spectrum output is not consistent. Would you like to help debug it if you have time?

sw005320:
It looks good to me. I'll merge it after the CI is passed.
In addition to the boundary issue in STFT, I want you to consider the following items:
- examples, models, and results for several recipes
- documentations
- making a simple notebook demo

espnet2/enh/separator/tcn_separator.py
```diff
@@ -114,6 +114,79 @@ def forward(
...
+            stream_outputs.append(frame_out)
+        end = time.time()
+        print(f"streaming processing, time cost: {end - start}")
+        for i in range(SEQ_LEN):
+            for s in range(num_spk):
+                torch.testing.assert_allclose(
+                    stream_outputs[i][s], seq_output[s][:, i : i + 1, :]
+                )
+
+    print("Streaming OKey")
```
Emrys365, line 192: We can remove this part as the unit test already covers it.

espnet2/enh/separator/skim_separator.py
```diff
@@ -129,6 +129,84 @@ def forward(
...
+            stream_outputs.append(frame_out)
+        end = time.time()
+        print(f"streaming processing, time cost: {end - start}")
+        for i in range(SEQ_LEN):
+            for s in range(num_spk):
+                torch.testing.assert_allclose(
+                    stream_outputs[i][s], seq_output[s][:, i : i + 1, :]
+                )
+
+    print("Streaming OKey")
```
Emrys365, line 212: ditto

espnet2/enh/separator/rnn_separator.py
```diff
@@ -122,3 +122,86 @@ def forward(
...
+            stream_outputs.append(frame_out)
+        end = time.time()
+        print(f"streaming processing, time cost: {end - start}")
+        for i in range(SEQ_LEN):
+            for s in range(num_spk):
+                torch.testing.assert_allclose(
+                    stream_outputs[i][s], seq_output[s][:, i : i + 1, :]
+                )
+
+    print("Streaming OKey")
```
Emrys365, line 207: ditto

Emrys365 (commented):
I have some additional comments.

espnet2/bin/enh_inference_streaming.py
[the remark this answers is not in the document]
Emrys365: OK. I will look at it later today.

espnet2/bin/enh_inference_streaming.py
```diff
@@ -0,0 +1,509 @@
...
+        >>> separate_speech = SeparateSpeechStreaming("enh_config.yml", "enh.pth")
+        >>> audio, rate = soundfile.read("speech.wav")
+        >>> frame_size, hop_size = separate_speech.frame_size, separate_speech.hop_size
+        >>> speech_sim_chunks, rest_pad = split_audio(wav, frame_size, hop_size)
+        >>> output_chunks = [[] for ii in range(separate_speech.num_spk)]
+        >>>
+        >>> for chunk in speech_sim_chunks:
+        >>>     output = separate_speech(chunk)
+        >>>     for channel in range(separate_speech.num_spk):
+        >>>         output_chunks[channel].append(output[channel])
```
nateanl, line 49:
```suggestion
        >>>     for spk in range(separate_speech.num_spk):
        >>>         output_chunks[spk].append(output[spk])
```

espnet2/bin/enh_inference_streaming.py
```diff
@@ -0,0 +1,509 @@
...
+    num_chunks = len(chunks)
+    batch_size = chunks[0].shape[0]
+    audio_len = int(hop_size * num_chunks + frame_size - hop_size)
+
+    output = torch.zeros((batch_size, audio_len), dtype=chunks[0].dtype).to(
+        chunks[0].device
+    )
+
+    for i, chunk in enumerate(chunks):
+        output[:, i * hop_size : i * hop_size + frame_size] += chunk
```
nateanl, line 358: how's the energy of the overlapped region between two consecutive chunks? I feel the energy is doubled.

espnet2/enh/decoder/stft_decoder.py
```diff
@@ -59,3 +73,60 @@ def forward(self, input: ComplexTensor, ilens: torch.Tensor):
...
 
         return wav, wav_lens
+
+    def _get_window_func(self):
+        window_func = getattr(torch, f"{self.window}_window")
+        window = window_func(self.win_length)
+        n_pad_left = (self.n_fft - window.shape[0]) // 2
+        n_pad_right = self.n_fft - window.shape[0] - n_pad_left
+        window = torch.cat(
+            [torch.zeros(n_pad_left), window, torch.zeros(n_pad_right)], 0
```
nateanl, line 83: why the window tensor is zero-padded?

espnet2/bin/enh_inference_streaming.py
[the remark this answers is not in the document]
Emrys365: > I tried your frame segment implementation, and although the number of frames is the same as the `torch.stft`, the spectrum output is not consistent. Would you like to help debug it if you have time?

What do you mean by `the spectrum output is not consistent`? I actually obtain the same framed time-domain signals after dropping the last frame from your code:
```python
import math

import librosa
import torch
import torch.nn.functional as F

from espnet2.bin.enh_inference_streaming import split_audio


def frame(x, frame_length, hop_length, dim=-1):
    if x.size(dim) < frame_length:
        raise ValueError(
            "Input is too short (n={:d})"
            " for frame_length={:d}".format(x.size(dim), frame_length)
        )

    if hop_length < 1:
        raise ValueError("Invalid hop_length: {:d}".format(hop_length))

    n_frames = 1 + (x.size(dim) - frame_length) // hop_length
    strides = list(x.stride())

    if dim == -1 or dim == x.dim() - 1:
        shape = list(x.shape[:-1]) + [frame_length, n_frames]
        strides = strides + [hop_length]

    elif dim == 0:
        shape = [n_frames, frame_length] + list(x.shape)[1:]
        strides = [hop_length * strides[0]] + strides

    else:
        raise ValueError("Frame dim={} must be either 0 or -1".format(dim))

    return x.as_strided(shape, strides, storage_offset=0)


if __name__ == "__main__":
    x = torch.rand(2, 16000)
    frame_length = 512
    hop_length = 128
    center = True

    if center:
        signal_dim = x.dim()
        extended_shape = [1] * (3 - signal_dim) + list(x.size())
        pad = int(frame_length // 2)
        x2 = torch.nn.functional.pad(x.view(extended_shape), [pad, pad])
        x2 = x2.view(x2.shape[-signal_dim:])

    # (2, 512, 126)
    splited_th = frame(x2, frame_length, hop_length, dim=-1)
    splited_np = librosa.util.frame(x2.numpy(), frame_length, hop_length, axis=-1)
    # (2, 512, 127)
    splited, rest = split_audio(x, frame_size=frame_length, hop_size=hop_length) 
    splited = torch.stack(splited, dim=-1)

    torch.testing.assert_allclose(splited_th, splited_np)
    torch.testing.assert_allclose(splited_th, splited[..., :-1])

```

espnet2/enh/decoder/stft_decoder.py
```diff
@@ -59,3 +73,60 @@ def forward(self, input: ComplexTensor, ilens: torch.Tensor):
...
+        )
+        return window
```
Emrys365, line 85: This function seems unused in `forward_streaming`, which I think should be used.

espnet2/bin/enh_inference_streaming.py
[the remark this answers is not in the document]
Emrys365: And we can recover the waveform from the framed signal via overlap-add:
```python
length = 16000
start = n_fft // 2 if center else 0
end = -n_fft // 2 if length is None else start + length
channel, n_fft, n_frame = splited.shape

y = torch.nn.functional.fold(
    splited,
    (1, (n_frame - 1) * hop_length + frame_length),
    (1, frame_length),
    stride=(1, hop_length),
).squeeze(2).squeeze(1)[:, start:end]
# in practice, you need to replace `torch.ones_like(splited)` with the actual window function
envelop = torch.nn.functional.fold(
    torch.ones_like(splited),
    (1, (n_frame - 1) * hop_length + frame_length),
    (1, frame_length),
    stride=(1, hop_length),
).squeeze(2).squeeze(1)[:, start:end]
y = y / envelop

torch.testing.assert_allclose(x, y)
```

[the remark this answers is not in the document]
Emrys365: In the above, I assume `n_fft == frame_length`. If not, please stick to the value of `n_fft` instead of `frame_length`.

espnet2/enh/decoder/stft_decoder.py
Emrys365: I think it is for the case where the window size is smaller than `n_fft`. Same as in https://github.com/librosa/librosa/blob/main/librosa/core/spectrum.py#L509

espnet2/bin/enh_inference_streaming.py
Emrys365: I agree. The current code only works for some specific windows that always sums to 1 in the overlapped regions.

Please refer to the implementation in torchaudio: https://github.com/pytorch/audio/blob/release/0.5/torchaudio/functional.py#L185-L217

test/espnet2/enh/decoder/test_stft_decoder.py
```diff
@@ -58,3 +60,24 @@ def test_STFTDecoder_invalid_type(
         )
         x_lens = torch.tensor([300 * hop_length, 295 * hop_length], dtype=torch.long)
         y, ilens = decoder(real, x_lens)
+
+
+def test_conv_enc_dec_streaming():
```
Emrys365, line 65: Now I think we need to test the consistency for STFTEncoder and STFTDecoder both separately and jointly.
That is, make sure:
  * the streaming mode of STFTEncoder/STFTDecoder is equivalent to the corresponding normal mode
  * the cascaded STFTEncoder and STFTDecoder can reconstruct the input waveform

espnet2/enh/separator/tcn_separator.py
LiChenda: Removed.

espnet2/enh/encoder/conv_encoder.py
```diff
@@ -45,3 +53,7 @@ def forward(self, input: torch.Tensor, ilens: torch.Tensor):
         flens = (ilens - self.kernel_size) // self.stride + 1
 
         return feature, flens
+
+    def forward_streaming(self, input: torch.Tensor):
```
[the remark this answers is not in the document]
LiChenda: Unit test added. The default ConvEnc/ConvDec always uses `padding=0`, and it is not configurable at the module level. So we don't need to consider it as all the existing models use `padding=0`.

espnet2/bin/enh_inference_streaming.py
LiChenda: Done! And now, all the consistency checks are passed.

espnet2/enh/decoder/stft_decoder.py
LiChenda: solved.

test/espnet2/enh/decoder/test_stft_decoder.py
LiChenda: Done.

LiChenda:
Thanks for @Emrys365 @nateanl 's comments. I refactored the code slightly, the frame split and merge functions are now moved into `AbsEncoder` and `AbsDecoder`, and implemented by their subclass. That is because the Conv- and STFT-based Enc/Dec act differently.  And that should be more flexible to implement streaming functions for other new Enc/Decs. 

BTW, all the consistency checks have been passed now.

espnet2/bin/enh_inference_streaming.py
```diff
@@ -0,0 +1,458 @@
...
+        )
+
+        # split continuous speech into small chunks to simulate streaming
+        speech_sim_chunks = separate_speech.frame(speech)
+        output_chunks = [[] for ii in range(separate_speech.num_spk)]
+
+        # the main loop for streaming processing
+        for chunk in speech_sim_chunks:
+            # process a single chunk
+            output = separate_speech(chunk)
```
Emrys365, line 321:
```suggestion
            output = separate_speech(chunk, fs=fs)
```

espnet2/bin/enh_inference_streaming.py
```diff
@@ -0,0 +1,458 @@
...
+        separate_speech.reset()
+
+        # merge chunks
+        waves = [separate_speech.merge(chunks, lengths) for chunks in output_chunks]
+
+        waves = [
+            (w / abs(w).max(dim=1, keepdim=True)[0] * 0.9).cpu().numpy() for w in waves
+        ]
+
+        # waves = [w.cpu().numpy() for w in waves]
```
Emrys365, line 336: remove this line

espnet2/enh/decoder/stft_decoder.py
```diff
@@ -72,3 +81,116 @@ def forward(self, input: ComplexTensor, ilens: torch.Tensor):
...
+        )
+
+        output_wav = output_wav.squeeze(1)
+
+        n_pad_left = (self.n_fft - self.win_length) // 2
+        output_wav = output_wav[..., n_pad_left : n_pad_left + self.win_length]
+
+        return output_wav * self._get_window_func()
+
+        # return output_wav
```
Emrys365, line 116: ditto

espnet2/enh/decoder/stft_decoder.py
```diff
@@ -72,3 +81,116 @@ def forward(self, input: ComplexTensor, ilens: torch.Tensor):
...
         return wav, wav_lens
+
+    def _get_window_func(self):
+        window_func = getattr(torch, f"{self.window}_window")
+        window = window_func(self.win_length)
+        n_pad_left = (self.n_fft - window.shape[0]) // 2
+        n_pad_right = self.n_fft - window.shape[0] - n_pad_left
+        # window = torch.cat(
+        #     [torch.zeros(n_pad_left), window, torch.zeros(n_pad_right)], 0
+        # )
```
Emrys365, line 92: ditto

Emrys365 (approved):
Thanks! I only have some minor comments.

espnet2/enh/decoder/stft_decoder.py
LiChenda: Done

espnet2/bin/enh_inference_streaming.py
LiChenda: Done

espnet2/enh/decoder/stft_decoder.py
LiChenda: Done

sw005320:
Is it OK to merge it?

Emrys365:
Yes, I think it is in a good shape now.

PR #5054, opened 2023-03-23
First diarization system based on Pyannote. 

popcornell:
I added a very simple diarization system based on pyannote, it mainly serves the purpose of showcasing how to integrate the diarization with the rest. 
TS-VAD incoming in the meantime. 

Made another folder into the `chime7_dasr` recipe called diar_asr1 as it does diarization + asr (actually also enhancement via GSS). 
Lots of symbolic links to the `asr1` folder however.


popcornell:
I am trying to improve results a bit by trying different hyper-parameters for diarization (e.g. merging segments that are X apart from same speaker) and trying to weight more/less the CTC in decoding. 

Right now unfortunately the DER and JER are quite good, but WER is terrible: 

Scenario | DER | WER
-- | -- | --
CHiME6 | 44% | 70%
DiPCo | 35% | 63%
Mixer 6 | 19% | 35%

</div></b>

I've tried to re-use as much as possible code from the other recipe to make it easier to maintain. 
The drawback is that training of the ASR model should be done in the `asr1` recipe. 

The diarization system is a multi-channel extension of https://huggingface.co/pyannote/speaker-diarization and uses the pre-trained components from it. 
The segmentation model is used to perform channel selection every 5 seconds chunk (0.5 seconds hop size): I run it on each channel and the channel with the most activations is chosen as the best. 
I do not use the top-k channels because I found it slows a lot the clustering step and I want the baseline to be fast (ASR is slow already enough). Using top-k should improve however. I added these comments in the README.

egs2/chime7_task1/diar_asr1/baseline_logs/inference.log
simpleoier, line 1: You may want to delete the log file?

popcornell: I kept it on purpose  so people can see the full results and compare with their

egs2/chime7_task1/diar_asr1/README.md
```diff
@@ -0,0 +1,188 @@
...
++----+---------------+---------------+---------------+----------------------+----------------------+---------------+-----------+---------+-------------+--------------------+--------------------------+-----------------+-----------------+----------------------+---------+-----------------+-------------+--------------+----------+
+```
+We can see that the DER and JER values are quite competitive on CHiME-6 with the state-of-the-art (e.g. JHU CHiME-6 submission) but
+the WER figures are quite poor. <br>
+**Note that here we report DER and JER which are computed against manual annotation for CHiME-6.** <br>
+The DER and JER obtained with respect to the [forced-alignment annotation](https://github.com/chimechallenge/CHiME7_DASR_falign) will be higher. In general this model can be optimized to reach ~46% DER with respect to such annotation on CHiME-6 (thus roughly comparable to Kaldi TS-VAD implementation) but
+we found this configuration to lead to worse WER overall.
+
+This may due to the fact that we use an E2E ASR system which may be more sensitive to segmentation errors compared to
```
sw005320, line 95: We could also state the benefit of the simple training recipes thanks to e2e.
As a baseline, we should claim that we consider the balance of simplicity and high performance.

egs2/chime7_task1/diar_asr1/README.md
```diff
@@ -0,0 +1,188 @@
...
+--decode-only 1 --gss-max-batch-dur 30-360-DEPENDING_ON_GPU_MEM \
+--pyan-use-pretrained popcornell/pyannote-segmentation-chime6-mixer6
+```
+---
+**If you want only to generate data you can run only stage 0.**
+```bash
+./run.sh --chime6-root YOUR_PATH_TO_CHiME6 --dipco-root PATH_WHERE_DOWNLOAD_DIPCO \
+--mixer6-root YOUR_PATH_TO_MIXER6 --stage 0 --stop-stage 0
+```
+Please refer to `egs2/chime7_task1/README.md` for additional infos such as
```
sw005320, line 157: From my environment, this part is highlighted. I don't know why.

image

egs2/chime7_task1/diar_asr1/local/database.yml
```diff
@@ -0,0 +1,49 @@
+Databases:
```
sw005320, line 1: Where and how is this file used?
It would be better to add one line comment in the header.

egs2/chime7_task1/asr1/local/da_wer_scoring.py
```diff
@@ -683,6 +683,9 @@ def get_sess2segs(segments):
         print("####################################################################")
         print("### Macro-Averaged Metrics across all Scenarios (Ranking Metric) ###")
         print("####################################################################")
-        macro_avg = scenario_wise_df.mean(0).to_frame().T
+        import pdb
+
+        pdb.set_trace()
```
sw005320, line 688: Do we need these lines?

sw005320 (commented):
I left some comments.
They look good to me, overall.

sw005320 (approved):
Thanks!

egs2/chime7_task1/diar_asr1/README.md
popcornell: I added some lines about it.

egs2/chime7_task1/diar_asr1/README.md
popcornell: I fixed this

egs2/chime7_task1/asr1/run.sh
```diff
@@ -175,6 +178,10 @@ if [ ${stage} -le 3 ] && [ ${stop_stage} -ge 3 ]; then
     pretrained_affix+="--download_model ${use_pretrained}"
   fi
 
+  if [ $diar_score ]; then
```
desh2608, line 181: Maybe this should be changed to `if [ $diar_score -eq 1 ]; then`, otherwise this condition will always evaluate to True.

PR #5055, opened 2023-03-24
Fix the condition for maxlen warning in beam search

pyf98:
This PR fixes the condition of warning that best hypo length is equal to max length in beam search.


sw005320:
Thanks a lot!
I think it must be correct, and I'll merge it, but I just want @pengchengguo to take a look at it.

pengchengguo:
This modification takes into account the case where `maxlen != x.shape[0]`, thank @pyf98 to fix this problem.

PR #5068, opened 2023-03-30
Tedlium3 recipe

Some-random:


sw005320:
LGTM!
Thanks!

Can you try E-Branchformer?
You can find a training config in the other recipes.

Issue #4910, opened 2023-02-04
s3prl inference issue (wav2vec2) 


SatwikDutta:
**Describe the bug**
ASR model has been trained and packed along with other required materials (stage 14) to be used for inference in another cluster/system. During inference "RuntimeError: Error(s) in loading state_dict for ESPnetASRModel" error is encountered, even though the model reflected by asr_model_file argument is present. 

**Basic environments:**
- OS information: Linux 5.15.0-48-generic #54~20.04.1-Ubuntu SMP Thu Sep 1 16:17:26 UTC 2022 x86_64
- python version: `3.8.5 (default, Sep  4 2020, 07:30:14)  [GCC 7.3.0]`
- espnet version: `espnet 0.10.7a1`
- pytorch version: `pytorch 1.10.1+cu113`
- Git hash: `18ca0993d53b8ec4a6ec44de87d2bd663734b06d`
  - Commit date: `Tue Feb 8 21:40:17 2022 +0900`
 
**Environments from `torch.utils.collect_env`:**
e.g.,
```
Collecting environment information...
PyTorch version: 1.10.1+cu113
Is debug build: False
CUDA used to build PyTorch: 11.3
ROCM used to build PyTorch: N/A

OS: Ubuntu 20.04.5 LTS (x86_64)
GCC version: (Ubuntu 9.4.0-1ubuntu1~20.04.1) 9.4.0
Clang version: Could not collect
CMake version: version 3.16.3
Libc version: glibc-2.31

Python version: 3.8.5 (default, Sep  4 2020, 07:30:14)  [GCC 7.3.0] (64-bit runtime)
Python platform: Linux-5.15.0-48-generic-x86_64-with-glibc2.10
Is CUDA available: True
CUDA runtime version: 11.3.109
GPU models and configuration: 
GPU 0: Tesla K80
GPU 1: Tesla K80
GPU 2: Tesla K80
GPU 3: Tesla K80
GPU 4: Tesla K80
GPU 5: Tesla K80

Nvidia driver version: 470.141.03
cuDNN version: Could not collect
HIP runtime version: N/A
MIOpen runtime version: N/A

Versions of relevant libraries:
[pip3] numpy==1.21.1
[pip3] numpydoc==1.1.0
[pip3] pytorch-ranger==0.1.1
[pip3] pytorch-wpe==0.0.1
[pip3] torch==1.10.1+cu113
[pip3] torch-complex==0.2.1
[pip3] torch-optimizer==0.1.0
[...]

PR #5074, opened 2023-04-01
Bug Fix: pretrained s3prl-frontend based models loaded with parameters key mismatch error

simpleoier:
This is related to #4910 .
Thanks @boeddeker for the solution!

### DIFF
--- a/espnet2/tasks/abs_task.py
+++ b/espnet2/tasks/abs_task.py
@@ -1875,6 +1875,30 @@ class AbsTask(ABC):
                 # NOTE(kamo): "cuda" for torch.load always indicates cuda:0
                 #   in PyTorch<=1.4
                 device = f"cuda:{torch.cuda.current_device()}"
-            model.load_state_dict(torch.load(model_file, map_location=device))
+            try:
+                model.load_state_dict(torch.load(model_file, map_location=device))
+            except RuntimeError:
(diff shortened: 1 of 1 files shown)


sw005320:
Can you fix some CI errors?
Then, I'll merge it.

sw005320:
Thanks, @simpleoier!
@boeddeker, could you test it?

espnet2/tasks/abs_task.py
```diff
@@ -1875,6 +1875,30 @@ def build_model_from_file(
                 # NOTE(kamo): "cuda" for torch.load always indicates cuda:0
                 #   in PyTorch<=1.4
                 device = f"cuda:{torch.cuda.current_device()}"
-            model.load_state_dict(torch.load(model_file, map_location=device))
+            try:
+                model.load_state_dict(torch.load(model_file, map_location=device))
```
boeddeker, line 1879: Could you change this to 

```python
state_dict = torch.load(model_file, map_location=device)
try:
    model.load_state_dict(state_dict)
```
and remove the later call of `torch.load`?

boeddeker:
Sorry for the late reply. Yes, the code works to load an old model. Thanks.
The downside is just, that there is now a file system overhead, because the file is loaded multiple times.

Could that be fixed? (See my comment in the code)

PR #5084, opened 2023-04-06
Add InterCTC to E-Branchformer encoder, and the ability to save InterCTC inference output to files

tjysdsg:
### Add InterCTC with optional self-conditioning to E-Branchformer encoder

This was originally written by @wanchichen.
The implementation is nearly the same as InterCTC in `TransformerEncoder` and `ConformerEncoder`.

### Add the ability to write predictions of InterCTC layer(s) to file during inference

The output is saved at `path/to/asr_decode_dir/test_set/logdir/output.*/1best_recog/encoder_interctc_layer<layer_idx>.txt`.

`<layer_idx>` corresponds to the indices set in model config file.

For example, if the config contains this line: `interctc_layer_idx: [6,12]`, then there will two files corresponding the output of the 6th and the 12th encoder layer: `encoder_interctc_layer6.txt` and `encoder_interctc_layer12.txt`.

### Misc

The code in this PR was used in many of my Aphasia speech detection experiments and produced good results. I also added more tests to make sure ASR without InterCTC still works.

@sw005320 I mentioned there was a bug I encountered in my experiments in our last meeting, but looks like it has already been patched.

### DIFF
--- a/espnet2/asr/encoder/e_branchformer_encoder.py
+++ b/espnet2/asr/encoder/e_branchformer_encoder.py
@@ -10,11 +10,12 @@ Reference:
 """
 
 import logging
-from typing import Optional, Tuple
+from typing import List, Optional, Tuple
 
 import torch
 from typeguard import check_argument_types
 
+from espnet2.asr.ctc import CTC
 from espnet2.asr.encoder.abs_encoder import AbsEncoder
 from espnet2.asr.layers.cgmlp import ConvolutionalGatingMLP
 from espnet2.asr.layers.fastformer import FastSelfAttention
@@ -207,6 +208,8 @@ class EBranchformerEncoder(AbsEncoder):
         linear_units: int = 2048,
         positionwise_layer_type: str = "linear",
         merge_conv_kernel: int = 3,
+        interctc_layer_idx=None,
+        interctc_use_conditioning: bool = False,
     ):
         assert check_argument_types()
         super().__init__()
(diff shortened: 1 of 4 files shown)


sw005320:
Thanks!

espnet2/asr/encoder/e_branchformer_encoder.py
```diff
@@ -386,6 +397,8 @@ def forward(
         xs_pad: torch.Tensor,
         ilens: torch.Tensor,
         prev_states: torch.Tensor = None,
+        ctc: CTC = None,
+        layer: int = None,
```
pyf98, line 401: I think the name `layer` is a bit hard to understand. We can call it `max_layers`?

espnet2/bin/asr_inference.py
```diff
@@ -420,16 +431,41 @@ def __call__(
...
+        self, intermediate_outs: List[Tuple[int, torch.Tensor]]
+    ) -> Dict[int, List[str]]:
+        assert check_argument_types()
+
+        exclude_ids = [self.asr_model.blank_id, self.asr_model.sos, self.asr_model.eos]
+        res = {}
+        token_list = self.beam_search.token_list
+
+        for layer_idx, encoder_out in intermediate_outs:
+            y = self.asr_model.ctc.argmax(encoder_out)[0]  # batch_size = 1
```
pyf98, line 461: The intermediate CTC only supports greedy decoding? I think it's fine.

tjysdsg: Yes. I think greedy decoding can probably cover most use cases

sw005320:
Thanks a lot!

PR #5087, opened 2023-04-06
Slue PR configs

siddhu001:
Upload configs and provide access to best pre-trained models for SLT paper (https://arxiv.org/abs/2211.05869).


siddhu001:
Hi @ftshijt,

This PR is ready for review. The primary modifications include the addition of configuration files for the best-trained model in the SLT paper and the ESPnet-SLUE toolkit, which was presented at the ASRU workshop. I've also implemented some minor fixes to the scoring function and the 2-pass SLU implementation. Could you please review it?

siddhu001:
Hi @sw005320,

This PR is related to slue-voxceleb and slue-voxpopuli recipes for ESPnet-SLUE toolkit. Please let me know if you have any comments before we can merge this PR!

sw005320:
OK, I'll start to review.
@ftshijt, can you also start it?

egs2/slue-voxpopuli/slu1/run.sh
```diff
@@ -0,0 +1,35 @@
...
+# -e 'error', -u 'undefined variable', -o ... 'error in pipeline', -x 'print commands',
+set -e
+set -u
+set -o pipefail
+
+train_set="train"
+valid_set="devel"
+test_sets="test devel"
+
+slu_config=conf/tuning/train_asr_no_pretrain.yaml
```
ftshijt, line 12: Consider make it standardize by setting conf/train.yaml

egs2/slue-voxpopuli/slu1/conf/tuning/train_asr_vq_apc.yaml
```diff
@@ -0,0 +1,88 @@
+specaug: specaug
```
ftshijt, line 1: It seems to me that the change of many configs are quite similar, do we want to keep all of them? (I'm ok with it if that's for public benchmark, but ideally, it would be better if we could summarize them into fewer configs, which is also easier for users)

ftshijt (commented):
Thanks for the update! Looks very good to me. Two minor comments

egs2/slue-voxpopuli/slu1/run.sh
siddhu001: Done!

egs2/slue-voxpopuli/slu1/conf/tuning/train_asr_vq_apc.yaml
siddhu001: Yes, I considered it, but similar configs (like the ones using pre-trained ASR) have different sets of hyperparameters (like warmup steps, learning rate) selected based on validation performance. Hence, grouping similar configs would reduce reproducibility, so I think it is better to keep all of the configs.

siddhu001:
@sw005320 I think this PR is ready to be merged!

egs2/slue-voxpopuli/slu1/README.md
sw005320, line 1: The table is not correctly displayed.
Can you confirm it?
image

siddhu001: Thanks for catching this! I have fixed this now!

PR #5100, opened 2023-04-10
[SVS] Extend SingingGenerate

A-Quarter-Mile:
Hi @ftshijt, here are updates on Muskits:

- Add stage 10 & 11 for uploading model
- Supporting SingingGenerate like Text2Speech


ftshijt:
Many thanks!

PR #5104, opened 2023-04-13
Update AphasiaBank Recipe

tjysdsg:
This PR includes new experiments of AphasiaBank English subset for my conference paper.

Besides run scripts and model configurations, I've added scripts for evaluation (Aphasia detection accuracy, ASR CER/WER of patients vs. control group, and so on). Some of the scripts are not needed for reproducing the experiments but I still included them for completeness and future reference, for example the ASR encoder profiling script provided by @pyf98.

Note that `config.py` is very long and ugly. It's okay because I hardcoded everything to make sure it's easy to debug and reproduce experiments.

There are a few things I need to complete before merging but feel free the review the rest of the files:

- [x] Present all results in `README.md`
- [x] Update the [HuggingFace model](https://huggingface.co/espnet/jiyang_tang_aphsiabank_english_asr_ebranchformer_small_wavlm_large1)
- [x] Upload the best-performing models to HuggingFace
- [x] Train some models for a few epochs and run evaluation to make sure everything is still working after I cleaned the code


egs2/aphasiabank/asr1/conf/train_asr.yaml
ftshijt, line 1: Could you still keep the symlink for train_asr.yaml?

egs2/aphasiabank/asr1/conf/tuning/train_asr_ebranchformer_small_wavlm_large1_interctc9.yaml
```diff
@@ -0,0 +1,115 @@
+# Uses 2 V100 (32GB)
```
ftshijt, line 1: Since the configs and results with inter ctc are highly similar, maybe we can consider simplify the PR by merging some of them?

egs2/aphasiabank/asr1/local/calc_macs_flops.sh
```diff
@@ -0,0 +1,24 @@
+#!/bin/bash
```
ftshijt, line 1: I think this could be a general functionality for the modeling. It would be great if you can extend this into a general function in `pyscripts` and `scripts`. (Not a must)

egs2/aphasiabank/asr1/local/clean_hyp_annotations.py
```diff
@@ -0,0 +1,38 @@
+"""
+Remove [APH] and [NONAPH] tags from the hypothesis file.
```
ftshijt, line 2: Given that you already registered APH and NONAPH into nlsyms.txt, it is necessary to have this script.

egs2/aphasiabank/asr1/local/config.py
```diff
@@ -0,0 +1,1407 @@
+"""
```
ftshijt, line 1: One of the possibility for the exact configs is to use another git to hold the information. Then we can make the code in espnet cleaner.

FYI: https://github.com/espnet/espnet/blob/master/egs2/puebla_nahuatl/asr1/local/data.sh#L48

egs2/aphasiabank/asr1/local/spk_info.txt
```diff
@@ -0,0 +1,744 @@
+ACWT01a	63.9	moderate	broca
```
ftshijt, line 1: Similar to the last comment, we may consider to hold the info in a separate repo.

egs2/aphasiabank/asr1/local/score_per_severity.sh
```diff
@@ -0,0 +1,76 @@
+#!/usr/bin/env bash
```
ftshijt, line 1: In `asr.sh`, we have an option to create `local/score.sh` to adopt corpus-specific scoring. Could you help to integrate your local score scripts into that?

FYI, you may check the open_li52 or ml_superb recipe to check how to use `local/score.sh`

egs2/aphasiabank/asr1/run.sh
```diff
@@ -1,28 +1,34 @@
...
-# -e 'error', -u 'undefined variable', -o ... 'error in pipeline', -x 'print commands',
+
+# Aphasia English E-Branchformer experiment
+
 set -e
 set -u
 set -o pipefail
 
+asr_tag="ebranchformer_wavlm"
+
```
ftshijt, line 10:
```suggestion
```
Please standardlize the recipe for the final version by removing some of the set and use default instead.

egs2/aphasiabank/asr1/run.sh
```diff
@@ -1,28 +1,34 @@
...
 train_set="train"
 valid_set="val"
 test_sets="test"
 include_control=true
+tag_insertion=none
 
-asr_config=conf/tuning/train_asr_ebranchformer_small_wavlm_large.yaml
+asr_config=conf/tuning/train_asr_ebranchformer_small_wavlm_large1.yaml
```
ftshijt, line 17:
```suggestion
asr_config=conf/train_asr.yaml
```

egs2/aphasiabank/asr1/run.sh
```diff
@@ -1,28 +1,34 @@
...
-inference_config=conf/decode.yaml
-
 ./asr.sh \
+  --asr_tag "${asr_tag}" \
   --lang en \
-  --inference_nj 100 \
-  --ngpu 1 \
+  --inference_nj 2 \
+  --gpu_inference true \
+  --ngpu 2 \
```
ftshijt, line 30:
```suggestion
```

ftshijt (changes requested):
Thanks for the great effort. I left some comments for better shaping the recipe.

egs2/aphasiabank/asr1/local/calc_macs_flops.sh
tjysdsg: Yifan actually has a separate repo for this (https://github.com/pyf98/espnet_utils), but I can extend this in a new PR.

egs2/aphasiabank/asr1/run.sh
tjysdsg: I specified the full path to the config file because the value of `feats_normalize` depends on whether WavLM is being used or not (https://github.com/espnet/espnet/pull/5104/files#diff-1aba163fe023c77a4da555552adb37dc70fb8de832562d4891075fe10c08cae1R20-R24).

This is also the reason that I removed `train_asr.yaml` since there is not way to tell if WavLM is used inside the config.

Do you have suggestions for a better way to do this?
One solution I have in mind is writing a very simple python script under `local/` which loads the yaml file and detect if SSLR is used and print true/false.
Then I can call it in `run.sh` and assign a suitable value to `feats_normalize` based on the output.

egs2/aphasiabank/asr1/local/clean_hyp_annotations.py
tjysdsg: My bad, I forgot to mention that `--nlsyms_txt` needs to be unset during scoring to keep the tags.
By doing this, we can calculate the WER of overall, patients', and control group's subset.

egs2/aphasiabank/asr1/conf/tuning/train_asr_ebranchformer_small_wavlm_large1_interctc9.yaml
tjysdsg: I'll remove the interctc9 config as it's mostly the same as interctc6.
But for other configs I think maybe it's best to keep them for easier reproducibility.

egs2/aphasiabank/asr1/local/clean_hyp_annotations.py
tjysdsg: Added instructions in README.md

egs2/aphasiabank/asr1/local/score_per_severity.sh
tjysdsg: After some consideration, I think it might be better to let the user perform scoring manually. As noted in the README file, `--nlsyms_txt "local/nlsyms.txt"` is used during training but `--nlsyms_txt none` should be used for scoring. Without modifying asr.sh, it's better for user to call it separately to avoid errors

tjysdsg:
Hi @ftshijt, besides two unresolved conversations, the PR is now ready for another review, thanks!

egs2/aphasiabank/asr1/local/calc_macs_flops.sh
ftshijt: Sounds cool! Either add dependency to Yifan's repo or adding it into scripts/pyscripts would be a great help.

egs2/aphasiabank/asr1/local/score_per_severity.sh
ftshijt: By using `local/score.sh` you have the freedom to choose whether to include nlsyms, so I still think it would be better to incorporate in that way. https://github.com/espnet/espnet/blob/master/egs2/TEMPLATE/asr1/asr.sh#L1692 

But it's up to your choice. Simply adding more instructions to readme is also fine.

tjysdsg:
After internal discussion with @ftshijt, I'll open a new PR to put AphasiaBank- specific scoring code in `local/score.sh`. Current PR is ready for merging @sw005320

egs2/aphasiabank/asr1/run_interctc.sh
sw005320, line 1: Can you move this under the `local` directory?

tjysdsg: Done!

PR #5107, opened 2023-04-13
Fix sentencepice version to v0.1.97

kamo-naoyuki:
I don't know the detail of the update. Please unfix if anyone knows.


kamo-naoyuki:
Note that the new sentencepiece supports python3.11. We need to use the new sentencepiece in the future.

PR #5114, opened 2023-04-14
Update ReazonSpeech training recipe for v1.1.0

fujimotos:
This is the first batch of ReazonSpeech v1.1.0 patches.

It adds a couple of tweaks to the training recipe, so that we can
train a model more efficiently & stably.

* Enable `macaron_style` and change `rel_pos_type` to "latest", which
  we needed in order to get training to converge stably.

* Enable Automatic Mixed Precision (AMP) to speed up training.

* Increase `--max_wav_duration` from 14s to 30s.

* Tune the batch size for a machine with eight cards of Nvidia A100.


sw005320:
Thanks a lot!

PR #5120, opened 2023-04-17
Support Whisper-style training as a new task S2T

pyf98:
Hi, this PR adds a new task `s2t1` (speech-to-text) in ESPnet2 which follows OpenAI Whisper's training style. It has two major features:
- It uses special tokens as task specifiers (e.g., transcribe, translate) or prediction targets (e.g., language ID) so that a single encoder-decoder model can perform multiple tasks for multiple languages.
- It supports conditional generation where the condition is the previous sentence in a long talk.

The training data has the following format:
```
<sop> prev<sos><category><task><starttime1> utt1<endtime1><starttime2> utt2<endtime2><eos>
```
where `<sop>` is a special token denoting the start of prev/prompt sentence. The timestamps are also treated as special tokens because the audio has a fixed length (30s) and resolution (20ms). An example looks like:

```
<sop> I'm going to talk today about energy and climate.<sos><en><transcribe><0.00> And that might seem a bit surprising, because my full-time work at the foundation is mostly about vaccines and seeds, about the things that we need to invent and deliver to help the poorest two billion live better lives.<14.12><15.36> But energy and climate are extremely important to these people; in fact, more important than to anyone else on the planet.<24.26><eos>
```

During data preparation, three text files are generated:
- `text` contains the normal target sentence, i.e., the text between `<sos>` and `<eos>`.
- `text.prev` contains the previous sentence, i.e., the text between `<sop>` and `<sos>`. This might be unavailable at the beginning of a talk. In such cases, a special token `<na>` will be used.
- `text.ctc` contains the ASR transcript without any special token, which is used for the CTC loss. For ASR utterances, this can be derived from `text`, but for ST utterances, this is in a different language. If the ASR transcription is not available, `<na>` will be used.

For decoding, the model can perform utterance-level ASR or ST, which follows the same procedure as the standard tasks. It can also perform long-form ASR or ST based on the predicted timestamps.


sw005320:
Very cool, @pyf98!
@pengchengguo and @ftshijt, can you review this PR?

egs2/TEMPLATE/s2t1/s2t.sh
```diff
@@ -0,0 +1,1853 @@
...
+            --input "${data_feats}/lm_train.txt" --output "${token_list}" ${_opts} \
+            --field 2- \
+            --cleaner "${cleaner}" \
+            --g2p "${g2p}" \
+            --write_vocabulary true \
+            --add_symbol "${blank}:0" \
+            --add_symbol "${oov}:1" \
+            --add_symbol "${sop}:-1" \
+            --add_symbol "${eos}:-2" \
+            --add_symbol "${sos}:-3"
```
ftshijt, line 949: Is there a specific reason that you use -1, -2, and -3. I remember the default for padding is set as -1, not sure that would be affected.

egs2/TEMPLATE/s2t1/s2t.sh
```diff
@@ -0,0 +1,1853 @@
...
+                fi
+            done
+        done
+
+    done
+fi
+
+
+if [ ${stage} -le 13 ] && [ ${stop_stage} -ge 13 ] && ! [[ " ${skip_stages} " =~ [[:space:]]13[[:space:]] ]]; then
+    log "Stage 13: Scoring"
```
ftshijt, line 1616: I guess you are still working on the scoring. Will check it later after it is settled~

egs2/must_c_v2/s2t1/conf/tuning/train_s2t_ebf_lr1e-3_warmup5k.yaml
```diff
@@ -0,0 +1,101 @@
...
+#     apply_freq_mask: true
+#     freq_mask_width_range:
+#     - 0
+#     - 27
+#     num_freq_mask: 2
+#     apply_time_mask: true
+#     time_mask_width_ratio_range:
+#     - 0.
+#     - 0.05
+#     num_time_mask: 5
```
ftshijt, line 35:
```suggestion
```

egs2/must_c_v2/s2t1/conf/tuning/train_s2t_ebf_lr1e-3_warmup5k.yaml
```diff
@@ -0,0 +1,101 @@
...
+    betas:
+    - 0.9
+    - 0.98
+    eps: 1.0e-06
+    weight_decay: 0.0
+scheduler: warmuplr
+scheduler_conf:
+    warmup_steps: 5000
+
+batch_type: unsorted
```
ftshijt, line 88: Is there a specific reason that we make it `unsorted`?

espnet2/bin/s2t_inference.py
```diff
@@ -0,0 +1,1040 @@
...
+                lm=scorers["lm"] if "lm" in scorers else None,
+                lm_weight=lm_weight,
+                multi_blank_durations=multi_blank_durations,
+                multi_blank_indices=multi_blank_indices,
+                token_list=token_list,
+                **transducer_conf,
+            )
+            beam_search = None
+            hugging_face_model = None
+            hugging_face_linear_in = None
```
ftshijt, line 190: Please remove the unnecessary functions for this PR. non only for the case that we probably won't test those options in a single PR, but it makes the PR a bit massive. Could you please remove those?

espnet2/bin/s2t_inference.py
```diff
@@ -0,0 +1,1040 @@
...
+            y = self.s2t_model.ctc.argmax(encoder_out)[0]  # batch_size = 1
+            y = [x[0] for x in groupby(y) if x[0] not in exclude_ids]
+            y = [token_list[x] for x in y]
+
+            res[layer_idx] = y
+
+        return res
+
+    def _decode_single_sample(self, enc: torch.Tensor):
+        if self.beam_search_transducer:
```
ftshijt, line 452: The inference seems not be finished, will check later.

ftshijt, line 452: The inference seems not be finished, will check later.

espnet2/s2t/espnet_model.py
```diff
@@ -0,0 +1,574 @@
...
+if V(torch.__version__) >= V("1.6.0"):
+    from torch.cuda.amp import autocast
+else:
+    # Nothing to do if torch<1.6.0
+    @contextmanager
+    def autocast(enabled=True):
+        yield
+
+
+class ESPnetS2TModel(AbsESPnetModel):
```
ftshijt, line 36: Maybe we can inherit ASR model? That would be especially helpful to get the latest updates from there.

ftshijt (commented):
Thanks for the great effort! Since there is still some ongoing implementation, please let me know if the followings are ready. Some comments as follows:

espnet2/bin/s2t_inference.py
pyf98: Inference (utterance-level or long-form) is finished.

espnet2/s2t/espnet_model.py
pyf98: This seems great. Let me check it. Thanks

espnet2/tasks/s2t.py
```diff
@@ -0,0 +1,596 @@
...
+    "preprocessor",
+    classes=dict(
+        s2t=S2TPreprocessor,
+    ),
+    type_check=AbsPreprocessor,
+    default="s2t",
+)
+
+
+class S2TTask(AbsTask):
```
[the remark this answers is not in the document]
pyf98: Thanks. Will check it.

egs2/must_c_v2/s2t1/conf/tuning/train_s2t_ebf_lr1e-3_warmup5k.yaml
pyf98: Just to make it as simple as possible. All the utterances have the same length 30s. So there is no need to sort.

egs2/TEMPLATE/s2t1/s2t.sh
pyf98: I just want to make them at the end of the vocab. In the original ASR, we only use one symbol for SOS and EOS, which is inserted at the end (-1).

Actually -1 means the last element. The token int is always positive, so it should not affect padding. Please correct me.

egs2/TEMPLATE/s2t1/s2t.sh
```diff
@@ -0,0 +1,1853 @@
...
+            # In supervised learning based speech enhancement / separation, it is equivalent to number of speakers.
+num_inf=    # Number of inferences output by the model
+            # Note that if it is not specified, it will be the same as num_ref. Otherwise, it will be overwritten.
+            # In MixIT, number of outputs is larger than that of references.
+
+# Upload model related
+hf_repo=
+
+# Decoding related
+use_k2=false      # Whether to use k2 based decoder
```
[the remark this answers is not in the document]
pyf98: Sounds good.

egs2/TEMPLATE/s2t1/s2t.sh
ftshijt: My apology, it was my mistake.

egs2/TEMPLATE/s2t1/s2t.sh
```diff
@@ -0,0 +1,1803 @@
...
+else
+    log "${help_message}"
+    log "Error: not supported: --feats_type ${feats_type}"
+    exit 2
+fi
+
+# Extra files for prev/prompt and ASR CTC
+utt_extra_files="text.prev text.ctc"
+
+num_inf=${num_inf:=${num_ref}}
```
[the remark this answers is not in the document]
pyf98: Thanks. I will remove these multi-speaker related variables.

espnet2/bin/s2t_inference.py
pyf98: Thanks. I simplified s2t inference a bit.

sw005320:
@pyf98, how about working toward merging this PR?
The version 1 is finished and we are moving to version 2.
This is a good timing.

We can make follow up PRs (version 2 or @simpleoier’s fix for long recording), later.

espnet2/train/preprocessor.py
```diff
@@ -1340,3 +1340,179 @@ def __call__(
...
+                        # because no space should be present between special tokens.
+                        # This works for bpe, but probably not for the other token types.
+                        text_ints = text_ints[1:]
+
+                        # Remove timestamps
+                        if self.train and np.random.uniform() > self.time_apply_prob:
+                            # Timestamps are continuous ints
+                            text_ints = text_ints[
+                                np.logical_or(
+                                    text_ints < self.first_time,
```
simpleoier, line 1488: This may not work for `char` token types. Two main problems I noticed: 1. the order of tokens is following the frequency, not strictly continuous. 2. if a token does not occur in the text, it will be skipped in the token list. To address the 1st problem, maybe it is safer to use
```
self.all_time_ints = np.array(
      [self.token_id_converter.token2id[x] for x in all_time_tokens]
)
time_indices = np.in1d(text_ints, self.all_time_ints)
text_ints = text_ints[~time_indices]
```

pyf98: Thanks for the comment! Currently, I think only the `bpe` token type can work from training to inference. I can add a type check in the `s2t.sh` which terminates the program if any other token is specified.

Also, speed perturbation is not supported because the timestamps will change.

sw005320:
I have several comments/discussions.

1. We should add the unit test and integration test. We could do it later
2. Should we keep `mixed_v1` or remove it later? Same for mixed_v2 if we finish mixed_v3. I feel we may remove it and keep the latest ones.
3. I agree with @ftshijt that inheriting the ASR task/model would be better.

egs2/TEMPLATE/s2t1/s2t.sh
```diff
@@ -0,0 +1,1786 @@
...
+fi
+
+
+if [ ${stage} -le 5 ] && [ ${stop_stage} -ge 5 ] && ! [[ " ${skip_stages} " =~ [[:space:]]5[[:space:]] ]]; then
+    if [ "${token_type}" = bpe ]; then
+        log "Stage 5: Generate token_list from ${bpe_train_text} using BPE"
+
+        mkdir -p "${bpedir}"
+        # shellcheck disable=SC2002
+        cat ${bpe_train_text} | cut -f 2- -d" "  > "${bpedir}"/train.txt
```
simpleoier, line 881: When you use multiple languages, the text data size of different languages are quite different. Would this affect the BPE model? For example, in the extreme case, the rare-resource languages would only have a very small portion in the vocabulary. What do you think of this problem?

pyf98: Yes, this is a problem. Actually in my case, sentencepiece will sample a fixed amount of lines from the entire data (due to memory issue), which can make the situation even worse.

Currently, I don't think our model will work well on low-resource languages due to the imbalanced training data.

We need a better solution as we scale up to more languages.

pyf98: For example, we can just re-use GPT2's vocabulary.

simpleoier: I see. Do you think training separate BPE models for different languages and then somehow combining them together is a solution?

pyf98: I'm not sure, since it will make the process very complicated.

I think this is a general machine learning problem. There might be some solutions in NLP.

egs2/mixed_v3/s2t1/run.sh
```diff
@@ -0,0 +1,41 @@
...
+
+nbpe=50000
+s2t_config=conf/train_s2t_transformer_conv2d_size1024_e24_d24_lr3e-4_warmup20k.yaml
+inference_config=conf/decode_s2t.yaml
+
+# inference only args
+# --cleaner whisper_en --hyp_cleaner whisper_en
+./s2t.sh \
+    --stage 11 \
+    --stop_stage 11 \
```
sw005320, line 21: fix?

pyf98:
Hi @sw005320 , can we merge this PR now (given the ASRU result)? Thanks!

espnet2/bin/s2t_inference.py
sw005320, line 1: It seems that this part does not have good test coverage.
Are there some specific reasons?

espnet2/bin/s2t_inference_language.py
sw005320, line 1: What is the difference between s2t_inference vs. this?

sw005320 (commented):
It almost ready but I just want to check some points.

espnet2/bin/s2t_inference.py
pyf98: One reason is that we cannot easily test the long-form decoding. It relies on the predicted timestamps. If we use random input or random model, we cannot generate structured output which makes the decoding hang forever.

espnet2/bin/s2t_inference_language.py
pyf98: s2t_inference requires a language token and performs the full decoding.
Here, it only predicts the language token, which means we set the output length to be exactly 1.

egs2/mixed_v3/s2t1/local/stats.py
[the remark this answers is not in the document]
pyf98: It shows the number of hours for every language pairs in the specified data directory. It is not used for training or decoding. Just a separate utility tool

espnet2/bin/s2t_inference_language.py
sw005320: Would it be possible to add an option to make the output length 1 to realize this with s2t_inference?
If there are any other changes, we can leave it as it is.

sw005320:
Thanks for your great efforts.
I just merged this PR.
However, this lacks the documentation.
So, please add the documents in the following place.

1. by following `https://github.com/espnet/espnet/blob/master/egs2/TEMPLATE/*/README.md`, please add the document to describe the detail of this new template
2. please add a brief explanation to the main document https://github.com/espnet/espnet/blob/master/README.md

PR #5124, opened 2023-04-17
[SVS] Update XiaoiceSing

A-Quarter-Mile:
Hi @ftshijt, here are updates on Muskits:

- Add slur in svs
- Add pitch_loss and vuv_loss for XiaoiceSing following XiaoiceSing2.

### DIFF
--- a/espnet2/gan_svs/espnet_model.py
+++ b/espnet2/gan_svs/espnet_model.py
@@ -88,6 +88,8 @@ class ESPnetGANSVSModel(AbsGANESPnetModel):
         duration_ruled_phn_lengths: Optional[torch.Tensor] = None,
         duration_syb: Optional[torch.Tensor] = None,
         duration_syb_lengths: Optional[torch.Tensor] = None,
+        slur: Optional[torch.Tensor] = None,
+        slur_lengths: Optional[torch.Tensor] = None,
         pitch: Optional[torch.Tensor] = None,
         pitch_lengths: Optional[torch.Tensor] = None,
         energy: Optional[torch.Tensor] = None,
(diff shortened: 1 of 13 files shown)


espnet2/gan_svs/espnet_model.py
```diff
@@ -333,6 +340,8 @@ def collect_feats(
         duration_ruled_phn_lengths: Optional[torch.Tensor] = None,
         duration_syb: Optional[torch.Tensor] = None,
         duration_syb_lengths: Optional[torch.Tensor] = None,
+        slur: Optional[torch.Tensor] = None,
```
ftshijt, line 343: I guess you also need to add slur preparation in the TEMPLATE/svs1/svs.sh and data.sh ?

espnet2/svs/xiaoice/XiaoiceSing.py
```diff
@@ -455,7 +484,9 @@ def forward(
         midi = midi[:, : midi_lengths.max()]  # for data-parallel
         label = label[:, : label_lengths.max()]  # for data-parallel
         tempo = tempo[:, : tempo_lengths.max()]  # for data-parallel
-
+        if self.loss_function == "XiaoiceSing2":
+            log_f0 = pitch[:, : pitch_lengths.max()]
+            vuv = log_f0 != 0
```
ftshijt, line 489: For log f0, why you are checking the 0 value? Isn't should be to check whether it is negative or positive?

ftshijt (commented):
LGTM! please check the vuv part to make sure it is correct

espnet2/svs/xiaoice/XiaoiceSing.py
A-Quarter-Mile: Fixed it by setting negative log_f0 to zero.

PR #5127, opened 2023-04-19
Add data augmentation to ReazonSpeech recipe

fujimotos:
This is the last PR to upgrade the ReazonSpeech recipe to v1.1.

* Related: #5114

In particular, this integrates the data augmentation strategies which we
implemented for ReazonSpeech v1.1.

 - Noise enhancement using MUSAN
   *  Overlay a random noise sample from MUSAN on top of raw audio.
 - Random audio cutoff
   * Trim leading segments from audio files.
 - Multi-Utterance
   * Combine several records to generate multi-utterance entries.

With these augmentation applied, we can cofirm that CER scores for
standard Japanese datasets drop more than 1%.


sw005320:
@pyf98 and @ftshijt, can you check this PR?

@fujimotos, I think I forgot to ask in the previous PR.
Can you add README.md with the results and model link?
You can find several examples in the other recipes (e.g., https://github.com/espnet/espnet/blob/master/egs2/csj/asr1/README.md)

egs2/reazonspeech/asr1/README.md
```diff
@@ -0,0 +1,22 @@
...
+
+## Environment
+
+- Python version: `3.8.16`
+- ESPnet version: `espnet 202301`
+- PyTorch version: `pytorch 1.12.1`
+
+## Pretrained model
+
+https://huggingface.co/reazon-research/reazonspeech-espnet-next
```
ftshijt, line 22: We has a auto-generated results file located in `exp/<your_model>/RESULTS.md` could you please use that to add the results at here?

fujimotos: > We has a auto-generated results file located

Sorry, next time we train ReazonSpeech model, we will make sure to
salvage RESULTS.md from training environment.

We periodically train a new model, so it won't be long before we can
follow it up and submit a PR that add the score table.

sw005320:
OK, but the table in README is not only designed for the latest result, but it is very valuable for people to check the record of the improvements in the models.
Thus, I strongly recommend it even for the current models. 
As @ftshijt says, the README generation is automated, and your model is already uploaded in the HF hub. 
So, I think you don't need so much effort.

pyf98:
Looks good to me.

egs2/reazonspeech/asr1/README.md
```diff
@@ -0,0 +1,36 @@
...
+- pytorch version: `pytorch 1.12.1+cu102`
+- Git hash: `9203b110547fba7609653e673097601043761ea5`
+  - Commit date: `Fri Apr 21 10:15:47 2023 +0900`
+
+## exp/asr_train_asr_conformer_raw_jp_char
+### WER
+
+|dataset|Snt|Wrd|Corr|Sub|Del|Ins|Err|S.Err|
+|---|---|---|---|---|---|---|---|---|
+|decode_asr_lm_lm_train_lm_jp_char_valid.loss.ave_asr_model_valid.acc.ave/test|64|64|53.1|46.9|0.0|0.0|46.9|46.9|
```
sw005320, line 26:
```suggestion
```

PR #5130, opened 2023-04-20
add e-branchformer result for tedlium3 and add checker for text output length

Some-random:


sw005320:
Thanks!

- Can you upload a model to huggingface hub and add a link to READEME.md?
- Can you move the e-branchformer results to the top of READEME.md?

sw005320:
@Some-random, any update?

espnet2/train/preprocessor.py
```diff
@@ -332,6 +332,12 @@ def _text_process(
             text = self.text_cleaner(text)
             tokens = self.tokenizer.text2tokens(text)
             text_ints = self.token_id_converter.tokens2ids(tokens)
+            if len(text_ints) > 100:
+                logging.warning(
+                    "The length of the text output exceeds 100, "
+                    "which may cause OOM on the GPU."
+                    "Please ensure that the data processing is correct and verify it."
+                )
```
b-flo, line 340: @sw005320 @Some-random Sorry, really late on that (I didn't sync w/ master in a while) but why was it added and how was the limit defined?

I was a bit confused when I saw that thousand of warnings were displayed in my training logs!

sw005320: Yes, it was added to a warning of the large memory consumption, but `100` is probably too small.
So, we can change it to `500` or more.
Sorry about it.

b-flo: Thanks and no worries, I'll open a PR!

PR #5138, opened 2023-04-24
Fix issues in split scps

pyf98:
Hi, this PR fixes two issues in the script `split_scps.py`. Please let me know if you spot any potential problems.

1. The original script opens and closes an output file everytime it writes a line. If the file has millions of lines, this process will take a huge amount of time. Now, I open all files before writing, which reduces the time from over 1 day to a few seconds in my case.
2. `prev_key` seems not updated and remains `None`, so I added line 55.

### DIFF
--- a/espnet2/bin/split_scps.py
+++ b/espnet2/bin/split_scps.py
@@ -33,12 +33,14 @@ def split_scps(
         (Path(output_dir) / name).mkdir(parents=True, exist_ok=True)
 
     scp_files = [open(s, "r", encoding="utf-8") for s in scps]
-    
+
     # Create output files in 'w' mode to overwrite existing files if any
     out_files = {
         name: {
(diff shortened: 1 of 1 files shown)


kamo-naoyuki:
Okay, thanks. I read this PR

Opening before writing or opening when writing actually both has a problem

- Opening before writing: Consuming many file descriptors, It's impossible to run this program if using many files by os limitations.
- Opening when writing: Taking many times

However, I don't have a good solution to this problem. To be honest, I think we should use a different format for this separation.

e.g. Multiple scps -> JSONL -> split -> Regenerate multiple scps

But, I don't have a large motivation for this because I don't use such a large dataset personally. If you have a time and sometimes use this function, could you consider to create better system for this?

For now, I'll merge this PR, thanks.

pyf98:
Thanks! I'll consider it.

I was training some LMs and the text had more than 40M lines which caused issues in loading data. So I tried to use the `num_splits_lm` option.

PR #5139, opened 2023-04-24
Add missing filter_scps scripts and note about kaldi for diarization example of mini_librispeech

toto6038:
In egs2/mini_librispeech/diar1 example, I found some problems:
- The Perl script in utils/data/get_utt2dur.sh is unable to process scp file of the dataset used in this example, e.g. data/train_clean_5/wav.scp. It results in the dependency of Kaldi despite it's a egs2 example.
- The Perl script used to filter scps is missing.

To address these problems, I make some changes as follows:
- Add a NOTE.md to warn user about the dependency of Kaldi. The warning wan't added to README.md considering that the file is generated by a script.
- Copy [filter_scps.pl](https://github.com/kaldi-asr/kaldi/blob/master/egs/wsj/s5/utils/filter_scps.pl) from Kaldi's repository.


kamo-naoyuki:
Thanks!

PR #5159, opened 2023-05-05
fix the gradient backward issue when joint training with s3prl frontend

simpleoier:
In s3prl, `feature_grad_mult` was set to 0. Thus the forward is in the context of `torch.no_grad()` [here](https://github.com/s3prl/s3prl/blob/57c3ba0b2215edf0aacea968d16493788a48a424/s3prl/upstream/wavlm/WavLM.py#L361C1-L367). It will stop the gradient back-propagation in joint training. To fix it, just set it to be 1 manually


sw005320:
@Emrys365, can you review this PR?
I hope this would solve our issues completely.

espnet2/asr/frontend/s3prl.py
```diff
@@ -50,14 +50,10 @@ def __init__(
...
+            if getattr(upstream.upstream.model, "feature_grad_mult", None):
+                upstream.upstream.model.feature_grad_mult = 1.0
         upstream.eval()
-        if getattr(
-            upstream.upstream, "model", None
-        ) is not None and upstream.upstream.model.__class__.__name__ in [
-            "Wav2Vec2Model",
-            "HubertModel",
-        ]:
-            upstream.upstream.model.encoder.layerdrop = 0.0
```
Emrys365, line 60: Why are these lines removed?

Emrys365 (approved):
LGTM. I can also verify the gradient issue is resolved on my side.

Emrys365 (approved):
LGTM. I can also verify the gradient issue is resolved on my side.

espnet2/asr/frontend/s3prl.py
pengchengguo: I think this is because S3PRL already sets `encoder_layerdrop=0` when initializing an upstream, so we no longer need to do this in ESPnet. Am I right?

simpleoier: Yes, it is correct.

sw005320:
Thanks a lot for fixing it!

PR #5162, opened 2023-05-10
Limit the Numpy version (<1.24) to fix CI error temporarily.

simpleoier:
This PR is intended to fix the CI errors. Due to the [deprecation](https://numpy.org/devdocs/release/1.24.0-notes.html#expired-deprecations) of numpy aliases, e.g. np.float, np.int, errors were raised in several packages (e.g. fairseq, librosa, etc).

Timeline:
1. Try to remove fairseq dependency, espeically in test.
2. Several other packages raise the same error. So the current solution is to constrain the numpy version (<1.24).

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -24,7 +24,8 @@ requirements = {
         "torch>=1.3.0",
         "torch_complex",
         "nltk>=3.4.5",
-        "numpy",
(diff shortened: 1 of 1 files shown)


sw005320:
Thanks!
It would be great if you also remove corresponding parts later since they are not used anymore.

PR #5173, opened 2023-05-17
[SVS] Add new recipe

A-Quarter-Mile:
Hi @ftshijt, here are new updates:

- Add ameboshi recipe
- Add itako recipe

Some changes on raw data have been merged into local/.


A-Quarter-Mile:
The modifications of raw data various differently and requires manually check. Modifications cannot be unified easily so that the codes may be a little bit dirty in prep_segments_*.py.

egs2/itako/svs1/conf/tuning/decode_rnn.yaml
```diff
@@ -0,0 +1,10 @@
+# This configuration is the decoding setting for FastSpeech or FastSpeech2.
+
+##########################################################
+#                    DECODING SETTING                    #
+##########################################################
+# speed_control_alpha: 1   # alpha to control the speed of generated speech
+                           # 1 < alpha makes slower and 1 > alpha makes faster
```
ftshijt, line 7:
```suggestion
```

egs2/itako/svs1/local/prep_segments.py
```diff
@@ -0,0 +1,189 @@
...
+        help="threshold for silence identification.",
+        default=30000,
+    )
+    parser.add_argument(
+        "--silence", action="append", help="silence_phone", default=["pau"]
+    )
+    return parser
+
+
+def make_segment(file_id, labels, threshold=30, sil=["pau", "br", "sil"]):
```
ftshijt, line 77: It is fine to have the customized design for each recipe and that is the reason that we keep the file in local. But could you possibly add some comments/guidelines/suggestions to others (maybe in the recipe level readme or in the general readme in the template)? Many thanks!

ftshijt: This is not an big issue, we can fix it with Yuxun's PR

PR #5140, opened 2023-04-25
Standalone Transducer v1.1

b-flo:
This PR is an update for the standalone Transducer version. It stitches several things I'm using as a baseline for new features (that won't be included here). Mainly:

- General: Fix various docstrings, documentation, variable names, etc + add `LayerDrop`.
- Streaming: Rework audio processing + some minor fixes for performance.
- Encoder: Add `E-Branchformer` (offline/streaming).
- Decoder: Add `MEGA` (w/ chunking) + fix decoder interface.

Some results are given for Libri-100 in another [PR](https://github.com/espnet/espnet/pull/4698#issuecomment-1425897072).

P.S: It's missing unit and integration tests for MEGA, I'll add it later this week.

PR #5175, opened 2023-05-19
Update asr_transducer_inference.py

aky15:
Correct streaming RNN-T inference with right context.


b-flo:
Hi @aky15 ,

Thanks again for using/fixing this version! 
Unfortunately, the audio processing part is changed in #5140, so your PR won't work (`right-context` was dropped anyway). Feel free to take a look at the other PR though, external reviews are appreciated!

PR #5180, opened 2023-05-23
Update SVS README.md

jerryuhoo:
This VISinger setting is based on OpenCPop, maybe we should make the same modifications to the settings of Ofuton as well.


egs2/TEMPLATE/svs1/README.md
```diff
@@ -281,35 +281,36 @@ $  --pretrained_model /exp/xiaoice-2-24-250k/500epoch.pth:svs:svs.fftsinger \
 
 
 ### VISinger training
```
ftshijt, line 283: Can you also add the paper references to the section as well?

egs2/TEMPLATE/svs1/README.md
```diff
@@ -281,35 +281,36 @@ $  --pretrained_model /exp/xiaoice-2-24-250k/500epoch.pth:svs:svs.fftsinger \
...
-The VITS config is hard coded for 22.05 khz or 44.1 khz and use different feature extraction method. (Note that you can use any feature extraction method but the default method is linear_spectrogram.) If you want to use it with 24 khz or 16 khz dataset, please be careful about these point.
+The VISinger config is hard coded for 44.1 khz and use different feature extraction method. (Note that you can use any feature extraction method but the default method is `fbank`.) If you want to use it with 24 khz or 16 khz dataset, please be careful about these point.
```
ftshijt, line 284: Can we relax the sampling rate requirement? I think 44.1khz is still a very high standard to be used in many applications.

egs2/TEMPLATE/svs1/README.md
ftshijt, line 283: May be we can change the title to `VISinger (1+2)` to note that we support both versions?

ftshijt (commented):
Thanks for the fix. Can you also update the readme in https://github.com/espnet/espnet#svs-singing-voice-synthesis ?

egs2/TEMPLATE/svs1/README.md
jerryuhoo: > 

How about adding 22.05 khz config files in both opencpop and ofuton? How should I name them?

ftshijt: I think we can just add a 22.05 suffix to those. Just in case someone would be interested in that

ftshijt: For pre-trained models, we can just do the 44.1 if the dataset supports (future action items)

PR #5182, opened 2023-05-24
Update generate_librimix_sd.sh

leepeiying:
When executing the Shell Script, a bad interpreter: No such file or directory error is reported.
So I change the first line, in this generate_librimix_sd.sh and it's work.


sw005320:
Which system are you using?
Please check https://www.baeldung.com/linux/bash-shebang-lines
We consistently use `#!/usr/bin/env bash`  for more portability.

leepeiying:
I using Ubuntu 20.04.6 LTS (GNU/Linux 5.4.0-146-generic x86_64), so I do this change is not correct?

sw005320:
Thanks a lot!

PR #5183, opened 2023-05-24
CHiME-7 DASR adding evaluation inference + adding support to use diarization baseline "pre-computed" JSONs

popcornell:
@Emrys365 does this fix it ?


Emrys365:
I think we can ignore this issue. Because the default value of `use_chime6_falign` in [egs2/chime7_task1/asr1/run.sh](https://github.com/espnet/espnet/blob/master/egs2/chime7_task1/asr1/run.sh) is 0. So the `CHiME7_DASR_falign` will not be used.

But probably some descriptions should be added to mention when we should set this argument to 1.

egs2/chime7_task1/diar_asr1/run.sh
```diff
@@ -65,6 +66,20 @@ if [ -z "$gss_asr_stage" ]; then
...
+  diar_inf_dset="eval"
+fi
+
+
+if [ -n "${download_baseline_diarization}" ]; then
+  # TODO
+  log "Using organizer-provided JSON manifests from the baseline diarization system."
+  #git clone
+  #mv
+  stage=4
```
simpleoier, line 79: Just want to double check if we can skip stage 3 if we have this. How can we download baseline diarization results?

egs2/chime7_task1/asr1/run.sh
```diff
@@ -67,7 +67,7 @@ nbpe=500
 asr_max_epochs=8
 # put popcornell/chime7_task1_asr1_baseline if you want to test with pretrained model
 use_pretrained=
-decode_only=0
+decode_only=
```
simpleoier, line 70: I'm concerned that when it is empty, line 206 `data_opts+=" --decode-only $decode_only"` will raise some error, e.g. `Error: No positional arguments are required`.

simpleoier (commented):
I left some comments in the code, which may cause some issues. But I'm not 100% sure myself.

egs2/chime7_task1/diar_asr1/run.sh
```diff
@@ -51,7 +52,7 @@ gss_max_batch_dur=90
 
 # ASR config
 use_pretrained=
-decode_only=1
+decode_only=
```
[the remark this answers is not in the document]
popcornell: I should add quotes, good catch

egs2/chime7_task1/diar_asr1/README.md
```diff
@@ -147,14 +147,17 @@ You can also play with diarization hyperparameters such as:
 
 as said merge-closer can have quite an impact on the final WER.
 
+**NOTE**
+We found the diarization baseline to be highly sensitive to the
```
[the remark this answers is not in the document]
popcornell: completed

egs2/chime7_task1/asr1/run.sh
popcornell: now it is an empty string.

popcornell:
FYI: there is still the numpy problem: 

> ERROR: Could not install packages due to an OSError: [Errno 2] No such file or directory: '/Users/runner/work/espnet/espnet/tools/venv/envs/espnet/lib/python3.10/site-packages/numpy-1.24.3.dist-info/METADATA'

popcornell:
@simpleoier Can we merge this ?

simpleoier (approved):
LGTM!
But there are many changes in non-relevant files. Probably due to black version. Also, @popcornell, there is one failure test: `check_kaldi_symlinks`, can you solve it?
@sw005320 Can we merge this?

popcornell:
I can rechange them back. 
What black version are you using btw ?

popcornell:
I guess then it is fine ?! Like the black version is not enforced explicitly. 
Maybe we should fix it however and upgrade it explicitly by one PR once in a while (with explicit reformatting of everything). 
I think these questions can be answered only by @sw005320

Emrys365:
The `black` formatting is now always enforced in https://github.com/espnet/espnet/blob/master/.pre-commit-config.yaml#L16-L20, which done every time you push a new commit.

sw005320:
Yes, this part is not an issue.
However, CI says that `egs2/TEMPLATE/asr1/utils/data/extend_segment_times.py` and `tools/kaldi/egs/wsj/s5/utils/data/extend_segment_times.py` are different.
https://github.com/espnet/espnet/actions/runs/5192602992/jobs/9361995304?pr=5183

If they are necessary changes, it would be better to move this file to the local (recipe-specific) directory.
If their changes are due to the format only,  it would be better to change it.

The issue is that they look very different due to the different formats, and I could not figure out the above changes.

popcornell:
Made a fresh PR because it is faster. Moved to https://github.com/espnet/espnet/pull/5228

PR #5193, opened 2023-05-31
Add new functions and fix some bugs in SE

Emrys365:
This PR adds two main functions to ESPnet-SE (speech enhancement):
  * DNSMOS and PESQ evaluation
    * We can now evaluate the [DNSMOS](https://github.com/microsoft/DNS-Challenge/blob/master/DNSMOS/README.md) scores with `espnet2/enh/layers/dnsmos.py`.
    * We can also evaluate the PESQ score (be careful about [the license](https://www.itu.int/rec/T-REC-P.862-200511-I!Amd2/en)).
  * utt2category support
    * Now the `enh.sh` pipeline can automatically add the category information (utt2category) of each sample to the mini-batch by setting `--use_utt2category true`, which will be later processed by the preprocessor (`espnet2/train/preprocessor.py`) to be converted into unique integers.
    * The chunk-based iterator also supports constructing mini-batches based on the category information, so that each mini-batch only contains samples of the same category.

In addition:
  * The bugs in the data preparation for recipes `egs2/vctk_noisy/enh1` and `egs2/vctk_noisyreverb/enh1` have been fixed.
  * The SE losses with a weight of 0 will now be calculated without gradient to reduce the memory footprint.
  * A new argument `channel_reordering` is added to EnhPreprocessor and TSEPreprocessor to support data augmentation for multi-channel signals.
  * A new argument `--skip_stats_npz` is added to `espnet2/tasks/abs_task.py` to allow skipping the preparation of `*_stats.npz` during the "Collecting Stats" stage . This is enabled by default for the SE task because these stats will not be used later and can save the computation.
  * A new learning rate scheduler has been added in `espnet2/schedulers/warmup_reducelronplateau.py`, which combines the WarmupLR and ReduceLROnPlateau schedulers.


egs2/TEMPLATE/enh1/enh.sh
```diff
@@ -71,6 +71,7 @@ is_tse_task=false   # Whether perform the target speaker extraction task or norm
 use_dereverb_ref=false
 use_noise_ref=false
 extra_wav_list= # Extra list of scp files for wav formatting
+use_utt2category=false
```
simpleoier, line 74: Maybe we can follow the same style used in enh_asr1, without explicitly having this flag variables? Because I think it is a bit redundant.

egs2/TEMPLATE/enh1/enh.sh
```diff
@@ -160,6 +161,7 @@ Options:
     --use_noise_ref    # Whether or not to use noise signal as an additional reference
                          for training a denoising model (default="${use_noise_ref}")
     --extra_wav_list   # Extra list of scp files for wav formatting (default="${extra_wav_list}")
+    --use_utt2category # Whether or not to load the utt2category file for training (default="${use_utt2category}")
```
simpleoier, line 164: ditto.

egs2/TEMPLATE/enh1/enh.sh
```diff
@@ -669,6 +672,21 @@ if ! "${skip_train}"; then
             done
         fi
 
+        # Add the category information at the end of the data path list
```
simpleoier, line 675: It is a bit redundant I think. Maybe we can do something in the sampler and preprocessor instead of here.

espnet2/bin/enh_scoring.py
```diff
@@ -64,6 +66,41 @@ def scoring(
         format="%(asctime)s (%(module)s:%(lineno)d) %(levelname)s: %(message)s",
     )
 
+    if dnsmos:
```
simpleoier, line 69: This is a good feature. Can you add pesq as well?

espnet2/bin/enh_scoring.py
```diff
@@ -64,6 +66,41 @@ def scoring(
...
+        if dnsmos_args["mode"] == "local":
+            from espnet2.enh.layers.dnsmos import DNSMOS_local
+
+            if not Path(dnsmos_args["primary_model"]).exists():
+                raise ValueError(
+                    f"The primary model '{dnsmos_args['primary_model']}' does not exist. "
+                    "You can download the model from https://github.com/microsoft/"
+                    "DNS-Challenge/tree/master/DNSMOS/DNSMOS"
+                )
+            if not Path(dnsmos_args["p808_model"]).exists():
```
simpleoier, line 79: A minor suggestion, this dnsmos is very specific to p808 model. It would be better if this can be more flexible for future.

espnet2/tasks/abs_task.py
```diff
@@ -1265,6 +1273,7 @@ def main_worker(cls, args: argparse.Namespace):
                 ngpu=args.ngpu,
                 log_interval=args.log_interval,
                 write_collected_feats=args.write_collected_feats,
+                skip_stats_npz=args.skip_stats_npz,
```
simpleoier, line 1276: Is this the same purpose as in line 1169 of this file? We set the model to None if we do not want it. Then in collect_stats function, we won't actually do collect_stats.
If so, we can combine this with the existing one and redesign the logic.

simpleoier (commented):
Thanks for the efforts! I left some comments. I mainly concern about the utt2category and collect_stats parts.

espnet2/bin/enh_scoring.py
Emrys365: Yes, I can add the code. Please review whether we will have the license issue.
> The [`pesq`](https://github.com/ludlows/PESQ) python package seems to violate the PESQ license.

egs2/TEMPLATE/enh1/enh.sh
Emrys365: Sure. I will remove it.

espnet2/bin/enh_scoring.py
Emrys365: I don't know if there are other options for the DNS model. I just followed the script provided by the official repository https://github.com/microsoft/DNS-Challenge/blob/master/DNSMOS/dnsmos_local.py#LL23C47-L23C47

egs2/TEMPLATE/enh1/enh.sh
Emrys365: I think that will require more changes in more scripts. The current design only needs modifications in `enh.sh`, `espnet2/tasks/enh.py`, and `espnet2/train/preprocessor.py`.

sw005320:
Can you add some tests for skipping the statistics collection and preprocessor?
It's OK if it is difficult.

Emrys365:
> Can you add some tests for skipping the statistics collection and some changes in preprocessor? It's OK if it is difficult.

Sure. For the `skipping the statistics collection' part, I think it is automatically covered by the integration tests. I will try to add some tests for the preprocessor.

sw005320:
See https://github.com/espnet/espnet/actions/runs/5182160044/jobs/9338608132?pr=5193#step:18:2471

This PR is good for my side.
So, if @Emrys365, @simpleoier, and CI are OK, please merge this PR.

simpleoier (approved):
LGTM!
A minor suggestion, it would be better to add acknowledgement or reference to DNSMOS or PESQ.

Issue #5195, opened 2023-05-31
Mask-CTC: Inferene Process on GPU


david-gimeno:
It happens when I try to do inference with a Mask-CTC model after setting ```device="cuda:0```:
```
Traceback (most recent call last):
  File "vsr_main.py", line 224, in <module>
    inference(args.output_dir, speech2text, test_loader, "test")
  File "vsr_main.py", line 77, in inference
    result = speech2text(torch.squeeze(xs_pad, 0))
  File "/home/david/anaconda3/envs/espnet/lib/python3.8/site-packages/torch/utils/_contextlib.py", line 115, in decorate_context
    return func(*args, **kwargs)
  File "/home/david/phd/E2E-SPEECH-RECOGNITION/src/bin/asr_inference_maskctc.py", line 127, in __call__
    hyp = self.s2t(enc[0])
  File "/home/david/anaconda3/envs/espnet/lib/python3.8/site-packages/torch/nn/modules/module.py", line 1501, in _call_impl
    return forward_call(*args, **kwargs)
  File "/home/david/espnet/espnet2/asr/maskctc_model.py", line 309, in forward
    mask_idx = torch.nonzero(probs_hat[y_idx] < p_thres).squeeze(-1)
RuntimeError: indices should be either on cpu or on the same device as the indexed tensor (cpu)
```
There are no errors if I run the inference process on the CPU but I am working with videos and it takes more time I expected. Thanks in advance,

David.

sw005320:
Thanks for the report.
Actually, we did not implement the GPU inference version of mask CTC since it was fast enough for many ASR applications.
Thanks for raising an issue.

@YosukeHiguchi
Is it possible to fix this easily?
Since the massive encoder has become a trend, GPU-supported mask CTC would be an important feature.

YosukeHiguchi:
@david-gimeno
I just tested GPU inference on a Mask-CTC and it worked fine on my end. Could you check whether the variable `enc[0]` is stored on the GPU before it is passed to the inference algorithm?

david-gimeno:
It works! Thank you so much :) Best regards from Spain,

David.

sw005320:
Very cool!

PR #5196, opened 2023-05-31
Bug fix for #5195

YosukeHiguchi:
This PR fixes a bug in Mask-CTC decoding using GPU (#5195).


sw005320:
Thanks for your quick action!

Issue #5198, opened 2023-06-01
The specaug applied in whisper encoder seems wrong


zhenhaoyong:
The shape of feats after `log_mel_spectrogram` is `[B, 80, T]`. However, the `specaug` requires the shape of input to be `[B, T, F]`.

        feats, feats_lens = self.log_mel_spectrogram(xs_pad, ilens)

        if self.specaug is not None and self.encoders.training:
            feats, feats_lens = self.specaug(feats, feats_lens)

        xs_pad, olens = self.whisper_encode(feats, feats_lens)

```
def time_warp(x: torch.Tensor, window: int = 80, mode: str = DEFAULT_TIME_WARP_MODE):
    """Time warping using torch.interpolate.

    Args:
        x: (Batch, Time, Freq)
        window: time warp parameter
        mode: Interpolate mode
    """
```

sw005320:
Thanks for the report.
@siddhu001, can you confirm it?

siddhu001:
Yes this is a known issue. I made a small PR https://github.com/espnet/espnet/pull/5206 to quickly fix this.

PR #5206, opened 2023-06-02
Quick fix for whisper specaug

siddhu001:
The specaug applied in whisper encoder has the issue mentioned here - https://github.com/espnet/espnet/issues/5198. This PR fixes it


sw005320:
Thanks for your quick action!

PR #5223, opened 2023-06-08
Fix bugs in mfa_format.py

G-Thor:
I'm a big fan of the MFA integration in ESPNet.

This PR fixes a couple of minor bugs in mfa_format.py that caused issues for me.

1. A floating point error caused the no. of frames computed from the mfa alignment file's 'end' value to (very rarely) differ from the statistics extraction step (stage 5 of TTS recipes) for FastSpeech2. Obtaining the value directly from the wav file eliminates that issue. It's a little bit slower but you're guaranteed to get the right number of frames every time. See below for example of this error.
2. I added the option to leave the text_cleaner argument blank, in case of pre-normalised transcripts in a non-English language.
3. I added an .rstrip("/") to ignore trailing slashes in the corpus_dir argument.

Example of fp-error:
```
>>> samples = 251392
>>> sr = 22050
>>> dur_float = samples / sr
>>> hop_length = 256
>>> true_frames = samples / hop_length
>>> float_frames = dur_float * sr / hop_length
>>> print(f"true no. of frames: {true_frames}. float error: {float_frames}")
true no. of frames: 982.0. float error: 981.9999999999999
>>> print(f"true no. of frames: {int(true_frames)+1}. frames when passed through float: {int(float_frames)+1}")
true no. of frames: 983. frames when passed through float: 982
```

### DIFF
--- a/egs2/TEMPLATE/asr1/pyscripts/utils/mfa_format.py
+++ b/egs2/TEMPLATE/asr1/pyscripts/utils/mfa_format.py
@@ -143,7 +143,7 @@ def get_parser():
     return parser
 
 
-def get_phoneme_durations(data: Dict, original_text: str, fs: int, hop_size: int):
+def get_phoneme_durations(data: Dict, original_text: str, fs: int, hop_size: int, n_samples: int):
     """Get phohene durations."""
     orig_text = original_text.replace(" ", "").rstrip()
     text_pos = 0
@@ -226,7 +226,7 @@ def get_phoneme_durations(data: Dict, original_text: str, fs: int, hop_size: int
     # STFT frames calculation: https://github.com/librosa/librosa/issues/1288
     # centered stft
 
-    total_durations = int(Decimal(str(maxTimestamp)) * fs / hop_size) + 1
+    total_durations = int(n_samples / hop_size) + 1
     timing_frames = [int(timing * fs / hop_size) + 1 for timing in timings]
     durations = [
         timing_frames[i + 1] - timing_frames[i] for i in range(len(timing_frames) - 1)
@@ -263,6 +263,8 @@ def validate(args):
 
 def make_durations(args):
     """Make durations file."""
+    import soundfile as sf
+
     wavs_dir = Path(args.corpus_dir)
     textgrid_dir = args.textgrid_dir
     train_text_path = args.train_text_path
@@ -276,9 +278,15 @@ def make_durations(args):
                 len(lab_paths) > 0
             ), f"The folder {wavs_dir} does not contain any transcription."
             for lab_path in lab_paths:
+                wav_path = lab_path.as_posix().replace(".lab", ".wav") # Assumes .wav files are in same dir as .lab files
+                if not os.path.exists(wav_path):
+                    logging.warning("There is no wav file for %s, skipping.", lab_path)
+                    continue
+                with sf.SoundFile(wav_path) as audio:
+                    no_samples = audio.frames # get no. of samples directly from .wav file
                 filename = (
                     lab_path.as_posix()
-                    .replace(args.corpus_dir + "/", "")
+                    .replace(args.corpus_dir.rstrip("/") + "/", "")
                     .replace(".lab", "")
                 )
                 with open(lab_path) as lab_file:
@@ -290,7 +298,7 @@ def make_durations(args):
                 with codecs.open(tg_path, "r", encoding="utf-8") as reader:
                     _data_dict = json.load(reader)
                 new_phones, durations = get_phoneme_durations(
-                    _data_dict, original_text, args.samplerate, args.hop_size
+                    _data_dict, original_text, args.samplerate, args.hop_size, no_samples
                 )
                 key = filename.split("/")[-1]
                 text_file.write(f'{key} {" ".join(new_phones)}\n')
@@ -326,6 +334,8 @@ def make_labs(args):
     import soundfile as sf
 
     from espnet2.text.cleaner import TextCleaner
+    if not args.text_cleaner:
+        args.text_cleaner = None
 
     corpus_dir = Path(args.corpus_dir)
     cleaner = TextCleaner(args.text_cleaner)


Fhrozen:
@G-Thor Thank you for your PR.
While 2 and 3 seems well addressed, I found a problem with 1.
I also tried to use reading the file, which may increase the processing time, but more importantly that the frequency sample of the wav file from the local file may be different from the one that will be used during training.

Example:
In this data prep, the output sampling rate is 48KHz (which is employed to generate the wav file at `data/local/mfa/corpus`)
https://github.com/espnet/espnet/blob/096e2bb770fdbb6ecc2367c169787474e3f63a2b/egs2/tsukuyomi/tts1/local/data_prep.sh#L35

But the training will be performed at 24KHz (which is used to calculate the durations by multiplying the `maxtime x Sampling rate`)
https://github.com/espnet/espnet/blob/096e2bb770fdbb6ecc2367c169787474e3f63a2b/egs2/tsukuyomi/tts1/run.sh#L8
This will generate an assertion error during stage 5 and so on, so I cannot fully accept this PR.

 I also observed the example of fp-error, and that is why I am using `Decimal.`  I did not observe any issue with the calculation recently, and I've been using multiple sets and configurations. So, if you find any particular configuration or data set, please report it to test it.

@sw005320 @G-Thor, As I mentioned, if you modify this PR to focus on 2 and 3, then I will approve. 1 needs a different direction.

G-Thor:
@Fhrozen Thanks so much for your review and your work on MFA integration.

I did not consider this downsampling issue. Good catch. This should definitely be fixed.

I considered the added processing time, but in my experience the difference is noticable, but not significant in the scale of model training.

Unfortunately the use of Decimal does not fix the fp-error, since it is originated during the duration measurement in MFA, which returns a float. So the error is already present in the str being converted to Decimal.
Indeed the values in the example I provided come from an audio sample that caused assertion errors due to a mismatch in no. of frames.

We'd either need to read the no. of samples from a downstreamed wav, which requires running this step after stage 2 of the TTS recipe and supplying the path to the `${data_feats}${_suf}/${dset}` used therein (the output dir of `format_wav_scp.sh`), or account for downsampling in the calculation.

What if we were to read in the original sample rate (`source_fs`) alongside the no. of samples and do the frame calculation as follows:
```
total_durations = int(int(n_samples * fs / source_fs) / hop_size) + 1
```
We need to cast the intermediate calculation to int because the total no. of samples is an integer.
I'm not sure this works yet, as I'm unsure how no. of samples is rounded during downsampling. I'll look into it and report back.

1 is the main issue for me here so I would definitely like to find a direction that works for it.

egs2/TEMPLATE/asr1/pyscripts/utils/mfa_format.py
```diff
@@ -263,6 +265,8 @@ def validate(args):
 
 def make_durations(args):
     """Make durations file."""
+    import soundfile as sf
```
Fhrozen, line 268: remove from here and move to the top

egs2/TEMPLATE/asr1/pyscripts/utils/mfa_format.py
```diff
@@ -327,6 +345,9 @@ def make_labs(args):
```
Fhrozen, line 345: Remove the import from here and move to the top

egs2/TEMPLATE/asr1/pyscripts/utils/mfa_format.py
```diff
@@ -276,9 +280,19 @@ def make_durations(args):
                 len(lab_paths) > 0
             ), f"The folder {wavs_dir} does not contain any transcription."
             for lab_path in lab_paths:
+                wav_path = lab_path.as_posix().replace(
+                    ".lab", ".wav"
+                )  # Assumes .wav files are in same dir as .lab files
+                if not os.path.exists(wav_path):
+                    logging.warning("There is no wav file for %s, skipping.", lab_path)
+                    continue
+                with sf.SoundFile(wav_path) as audio:
```
Fhrozen, line 289: Try with:
```python
data, orig_rate = sf.read(wav_path)
no_samples = int(data.shape[0] * (args.samplerate / orig_rate))  # Assuming 1-D
```
And remove importing Decimal at top

Fhrozen (changes requested):
I added some comments.
Please, try with the modification and let me know if this work on your end.
Try also running until stage 5 to confirm no issues with the Duration extraction.

If this fixes the issue, try to fix the CI comments and will be ready to merge.

egs2/TEMPLATE/asr1/pyscripts/utils/mfa_format.py
Fhrozen: Modify Lines 289~292 with the given comment

G-Thor: Thanks for your suggestion. I used sf.SoundFile() instead of sf.read(), since in my tests it is faster but gives the same result.
If you have a specific reason to use sf.read() instead, I'll gladly replace it.

sw005320: Interesting.
Do you know what the difference is?

G-Thor:
Thanks for your review.
I've implemented the requested changes. I used sf.SoundFile() rather than sf.read() since it is faster. Let me know if you strongly prefer sf.read()

During my testing on tsukuyomi I encountered an issue in mfa_format.py. see commit [552b4c0](https://github.com/espnet/espnet/pull/5223/commits/552b4c0143f5a2087925cc74d2f5f49d9d2ee736) and [pyopenjtalk's  changes](https://github.com/r9y9/pyopenjtalk/pull/40).

After implementing this fix I encountered no errors in running mfa.sh nor in running the recipe up to stage 5 using fastspeech2 config and mfa-derived durations.

egs2/TEMPLATE/asr1/pyscripts/utils/mfa_format.py
sw005320: Got it.
OK, then, for this use, we can just use SoundFIle()

Fhrozen:
LGTM

sw005320:
Thanks!

PR #5229, opened 2023-06-14
Update import order

ftshijt:
To fix a potential issue when speechbrain is not installed but want to use RawNet speaker embedding extraction


sw005320:
Thanks, @ftshijt!

Issue #5237, opened 2023-06-16
Mismatched keys in pretrained JETS model


iamanigeeit:
Hello @imdanboy,

When i try to load the pretrained model from HuggingFace, it seems that `tts.discriminator.msd.discriminators.0.layers` should have `weight_v` and `weight_g` but it only has `weight`.

```
os.chdir(LJSPEECH_DIR)
pretrained_dir = LJSPEECH_DIR / "exp/tts_train_jets_raw_phn_tacotron_g2p_en_no_space"
pretrained_model_file = pretrained_dir / "train.total_count.ave_5best_new.pth"
pretrained_tts = Text2Speech.from_pretrained(
    train_config=pretrained_dir / "config.yaml",
    model_file=pretrained_model_file,
    device=device
)
pretrained_model = pretrained_tts.model
os.chdir(PWD)
```

Error:
```
RuntimeError: Error(s) in loading state_dict for ESPnetTTSModel:
	Missing key(s) in state_dict: "tts.discriminator.msd.discriminators.0.layers.0.0.weight_g", "tts.discriminator.msd.discriminators.0.layers.0.0.weight_v", "tts.discriminator.msd.discriminators.0.layers.1.0.weight_g", "tts.discriminator.msd.discriminators.0.layers.1.0.weight_v", "tts.discriminator.msd.discriminators.0.layers.2.0.weight_g", "tts.discriminator.msd.discriminators.0.layers.2.0.weight_v", "tts.discriminator.msd.discriminators.0.layers.3.0.weight_g", "tts.discriminator.msd.discriminators.0.layers.3.0.weight_v", "tts.discriminator.msd.discriminators.0.layers.4.0.weight_g", "tts.discriminator.msd.discriminators.0.layers.4.0.weight_v", "tts.discriminator.msd.discriminators.0.layers.5.0.weight_g", "tts.discriminator.msd.discriminators.0.layers.5.0.weight_v", "tts.discriminator.msd.discriminators.0.layers.6.0.weight_g", "tts.discriminator.msd.discriminators.0.layers.6.0.weight_v", "tts.discriminator.msd.discriminators.0.layers.7.weight_g", "tts.discriminator.msd.discriminators.0.layers.7.weight_v". 
[...]

Issue #4595, opened 2022-08-23
A bug in HiFiGANScaleDiscriminator


JoeyHeisenberg:
seems that weight norm doesn't work in HiFiGANScaleDiscriminator, in which we only get conv1d

https://github.com/espnet/espnet/blob/f274ebed88e3b4820b23ea71fb7f9f6d56706be9/espnet2/gan_tts/hifigan/hifigan.py#L410-L573

kan-bayashi:
Thank you for your report, you are right.

kan-bayashi:
Related: https://github.com/kan-bayashi/ParallelWaveGAN/issues/309

PR #5240, opened 2023-06-19
fix norm compatibility in scale discriminator

kan-bayashi:
Before https://github.com/espnet/espnet/commit/1b45bdc8c98b3b6bf149df241869aa63051acbd1, the norm for scale discriminator is not applied correctly. This causes mismatch of parameters with configuration. As a result, we cannot load the pretrained models.

~~To solve this issue, when parameter mismatch happens in loading, we remove the norm at first, load the parameters, and then apply the norm in post-hook functions.~~

It seems that applying weight norm and spectral norm for pretrained parameters causes unexpected behavior.
Therefore, I changed to just show the message about the training error when fine-tuning and the instruction to use pretrained parameters.
Just using pretrained model for the inference is fine.

Fix the following issues:
- #4595 
- #5237 
- #5239

### DIFF
--- a/espnet2/gan_tts/hifigan/hifigan.py
+++ b/espnet2/gan_tts/hifigan/hifigan.py
@@ -528,12 +528,18 @@ class HiFiGANScaleDiscriminator(torch.nn.Module):
             raise ValueError("Either use use_weight_norm or use_spectral_norm.")
 
         # apply weight norm
+        self.use_weight_norm = use_weight_norm
         if use_weight_norm:
             self.apply_weight_norm()
 
         # apply spectral norm
+        self.use_spectral_norm = use_spectral_norm
         if use_spectral_norm:
             self.apply_spectral_norm()
+
+        # backward compatibility
+        self._register_load_state_dict_pre_hook(self._load_state_dict_pre_hook)
+        self.register_load_state_dict_post_hook(self._load_state_dict_post_hook)
 
     def forward(self, x: torch.Tensor) -> List[torch.Tensor]:
         """Calculate forward propagation.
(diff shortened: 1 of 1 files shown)


kan-bayashi:
`register_load_state_dict_post_hook` seems available from pytorch `1.12` :(
I will consider another method.

iamanigeeit:
> `register_load_state_dict_post_hook` seems available from pytorch `1.12` :( I will consider another method.

Thanks for the update! I was going to upgrade my driver, cuda and pytorch just to make it work.

iamanigeeit:
> `register_load_state_dict_post_hook` seems available from pytorch `1.12` :( I will consider another method.

I think you have to check the `state_dict` for the MultiScaleDiscriminator only, because `state_dict` contains norm for other parameters.

```
if self.use_weight_norm and not any(
            ["weight_g" in k for k in state_dict.keys()]
        )  # this will always be False 
```

espnet2/gan_tts/hifigan/hifigan.py
```diff
@@ -572,6 +577,86 @@ def _apply_spectral_norm(m: torch.nn.Module):
...
+                logging.debug(f"Spectral norm is removed from {m}.")
+                torch.nn.utils.remove_spectral_norm(m)
+            except ValueError:  # this module didn't have weight norm
+                return
+
+        self.apply(_remove_spectral_norm)
+
+    def _load_state_dict_pre_hook(
+        self,
+        state_dict,
```
kan-bayashi, line 606: @iamanigeeit I think the `state_dict` here includes only keys corresponding to the current modules:
```python
(Pdb++) state_dict.keys()
dict_keys(['msd.discriminators.0.layers.0.0.weight', 'msd.discriminators.0.layers.0.0.bias', 'msd.discriminators.0.layers.1.0.weight', 'msd.discriminators.0.layers.1.0.bias', 'msd.discriminators.0.layers.2.0.weight', 'msd.discriminators.0.layers.2.0.bias', 'msd.discriminators.0.layers.3.0.weight', 'msd.discriminators.0.layers.3.0.bias', 'msd.discriminators.0.layers.4.0.weight', 'msd.discriminators.0.layers.4.0.bias', 'msd.discriminators.0.layers.5.0.weight', 'msd.discriminators.0.layers.5.0.bias', 'msd.discriminators.0.layers.6.0.weight', 'msd.discriminators.0.layers.6.0.bias', 'msd.discriminators.0.layers.7.weight', 'msd.discriminators.0.layers.7.bias'])
```
So I think we do not need to filter. Could you check the current code?

iamanigeeit: I added in `hifigan.py`
```
        for x in state_dict.keys():
            print(x)
        if self.use_weight_norm and not any(
            ["weight_g" in k for k in state_dict.keys()]
        ):
```

Then:
```
import os
os.environ['CUDA_LAUNCH_BLOCKING'] = "1"
from pathlib import Path
from espnet2.bin.tts_inference import Text2Speech
PWD = %pwd
PWD = Path(PWD)
LJSPEECH_DIR = (PWD / '../egs2/ljspeech/tts1/').resolve()
device = 'cuda'
os.chdir(LJSPEECH_DIR)
pretrained_dir = LJSPEECH_DIR / "exp/tts_train_jets_raw_phn_tacotron_g2p_en_no_space"
pretrained_model_file = pretrained_dir / "train.total_count.ave_5best.pth"
pretrained_tts = Text2Speech.from_pretrained(
    train_config=pretrained_dir / "config.yaml",
    model_file=pretrained_model_file,
    device=device
)
pretrained_model = pretrained_tts.model
os.chdir(PWD)
```

feats_extract.logmel.melmat
normalize.mean
normalize.std
pitch_normalize.mean
pitch_normalize.std
energy_normalize.mean
energy_normalize.std
tts.generator.encoder.embed.0.weight
tts.generator.encoder.embed.1.alpha
tts.generator.encoder.encoders.0.self_attn.linear_q.weight
tts.generator.encoder.encoders.0.self_attn.linear_q.bias
...
tts.discriminator.mpd.discriminators.4.output_conv.bias
tts.discriminator.mpd.discriminators.4.output_conv.weight_g
tts.discriminator.mpd.discriminators.4.output_conv.weight_v
tts.mel_loss.wav_to_mel.logmel.melmat

kan-bayashi: Does it work for you ?
```python
        current_module_keys = [
            x for x in self.state_dict().keys() if x.startswith(prefix)
        ]
        if self.use_weight_norm and not any(
            ["weight_g" in k for k in current_module_keys]
        ):
```

iamanigeeit: Sorry, i got disconnected due to network upgrade.

Yes, this works -- i am on PyTorch 1.11. Thanks!

kan-bayashi:
@sw005320 Can we ignore the CI error on MacOS?

@ftshijt Maybe SVS pretrained models have the same issue, could you check this PR?

ftshijt:
> @sw005320 Can we ignore the CI error on MacOS?
> 
> @ftshijt Maybe SVS pretrained models have the same issue, could you check this PR?

Thanks for the fix, it looks good to me! Should be compatible with the current SVS models.

PR #5268, opened 2023-06-30
Add support for K2 pruned transducer loss

b-flo:
This PR add capabilities to train with k2 pruned transducer loss. It's working but a WIP.
It's was not a priority but I'm currently doing a major work/rework on the training pipeline (initialization, schedulers, criterions, normalization, loss and auxiliary losses, etc), I think it should be considered.

Also:

- Use `maes` instead of `default` in `ErrorCalculator` to avoid endless/long loop in some cases (e.g.: unrestricted decoding with constrained model training).
- Use `torch.tensor` instead of `torch.LongTensor` in all decoders' `batch_score`. The legacy constructor will raise device mismatch error with torch >1.10.

To add:

- [x] Tests
- [x] Docs

### DIFF
--- a/espnet2/asr_transducer/error_calculator.py
+++ b/espnet2/asr_transducer/error_calculator.py
@@ -39,8 +39,8 @@ class ErrorCalculator:
         self.beam_search = BeamSearchTransducer(
             decoder=decoder,
             joint_network=joint_network,
-            beam_size=1,
-            search_type="default",
+            beam_size=2,
+            search_type="maes",
             score_norm=False,
         )
 
--- a/espnet2/asr_transducer/error_calculator.py
+++ b/espnet2/asr_transducer/error_calculator.py
@@ -38,6 +38,18 @@ class ErrorCalculator:
         """Construct an ErrorCalculatorTransducer object."""
         super().__init__()
 
+        # (b-flo): Since the commit #8c9c851 we rely on the mAES algorithm for
+        # validation instead of the default algorithm.
+        #
+        # With the addition of k2 pruned transducer loss, the number of emitted symbols
+        # at each timestep can be restricted during training. Performing an unrestricted
+        # (/ unconstrained) decoding without regard to the training conditions can lead
+        # to huge performance degradation. It won't be an issue with mAES and the user
+        # can now control the number of emitted symbols during validation.
+        #
+        # Also, under certain conditions, using the default algorithm can lead to a long
+        # decoding procedure due to the loop break condition. Other algorithms,
+        # such as mAES, won't be impacted by that.
         self.beam_search = BeamSearchTransducer(
             decoder=decoder,
             joint_network=joint_network,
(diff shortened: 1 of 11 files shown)


ftshijt:
Maybe not related, but curious, are you using the latest k2 for this one? I remember we are using an old fixed version for that. Is it possible to change that to up-to-date version?

b-flo:
I think it's in a OK state for now. I need to do additional testing to ensure/fix some general parameters/variables depending on how k2 kernel is set but it can be done in later PRs.

Also, for reference, for a E-Branchformer/RNN trained with either loss kernel:

| Kernel | batch_bins (accum_grad) | max_cached_mem (GB) | dev_clean (%WER) | dev_other (%WER) | test_clean (%WER) | test_other (%WER) |
|-|-|-|-|-|-|-|
| warp-transducer  | 4000000 (8) | 37.4G | 5.7 | 16.8 | 6.0 | 17.1 |
| k2 pruned  (regular) | 4000000 (8) | 28.84G | 5.6 | 16.7 | 6.1 | 17.1 |
| k2 pruned  (regular) | 8000000 (4) | 37.50G | 5.7 | 16.8 | 6.1 | 17.1 |

I don't provide training time here because it's a bit difficult to properly monitor in my current environment.

espnet2/asr_transducer/espnet_transducer_model.py
```diff
@@ -362,39 +419,144 @@ def _calc_transducer_loss(
...
                 )
                 exit(1)
 
-        loss_transducer = self.criterion_transducer(
-            joint_out,
-            target,
-            t_len,
-            u_len,
-        )
+        with autocast(False):
```
b-flo, line 427: It's an oversee on my part, it should have been included in #5140 for amp usage. The (auto)castings will be changed when float16 support is merged to `warp-transducer`.

espnet2/asr_transducer/error_calculator.py
```diff
@@ -39,8 +41,9 @@ def __init__(
         self.beam_search = BeamSearchTransducer(
             decoder=decoder,
             joint_network=joint_network,
-            beam_size=1,
-            search_type="default",
+            beam_size=2,
+            search_type="maes",
```
sw005320, line 45: Is it OK to change it from `default` to `maes`?
What is the intention of this change?

b-flo: > Is it OK to change it from default to maes?

It is, it won't make a difference for the reporting.

> What is the intention of this change?

To control the number of emitted symbols at each timestep in both training and validation.
With k2, we can end-up with a big performance degradation if we enforce an unrestricted decoding without regards to the training condition/parameters (See. https://arxiv.org/pdf/2211.00484.pdf). With `maes`, it'll be "stable" under all conditions.

Also, in some cases you can end-up with a really long looping procedure with `default` (e.g.: during early stages with some architecture such as MEGA or if there is a training/decoding condition mismatch as I said). It won't happen with other decoding algorithms.

sw005320: Thanks for the clarification.
I also encountered the issue of the number of emitted symbols, and this change sounds good to me.
Can you leave this discussion somewhere in the source code?

b-flo: Done but I'm not sure it's in a right place nor in a right form..!

Issue #5266, opened 2023-06-29
max_start_idx in get_random_segments.py must be >= 0


iamanigeeit:
While training JETS i kept having failures on short audio files like LibriTTS utt_id 3607_29116_000041_000001. This is caused by `feats_lengths < segment_size` here: 
https://github.com/espnet/espnet/blob/abd3aa7674e34acb0c490031c92b64d874099351/espnet2/gan_tts/utils/get_random_segments.py#L28-L33

`start_idxs` will be negative. Then training will fail here due to a 0-sized Tensor slice.
https://github.com/espnet/espnet/blob/abd3aa7674e34acb0c490031c92b64d874099351/espnet2/gan_tts/utils/get_random_segments.py#L56-L57

Fix:
Add `max_start_idx[max_start_idx < 0] = 0`

sw005320:
Thanks for pointing out the issue; your fix sounds reasonable.
Could I ask you to make a PR?

sucveria:
How about adding the following fix?
```
diff --git a/espnet2/gan_tts/utils/get_random_segments.py b/espnet2/gan_tts/utils/get_random_segments.py
index e6180e1c1..672bf44ab 100644
--- a/espnet2/gan_tts/utils/get_random_segments.py
+++ b/espnet2/gan_tts/utils/get_random_segments.py
@@ -27,6 +27,7 @@ def get_random_segments(
     """
     b, c, t = x.size()
     max_start_idx = x_lengths - segment_size
+    max_start_idx = torch.where(max_start_idx < 0, 0, max_start_idx)
     start_idxs = (torch.rand([b]).to(x.device) * max_start_idx).to(
         dtype=torch.long,
     )
```

kan-bayashi:
Fixed in #5274

PR #5274, opened 2023-07-03
Fix random segments

iamanigeeit:
https://github.com/espnet/espnet/issues/5266


kan-bayashi (approved):
This is good idea. LGTM!

PR #5276, opened 2023-07-03
Some name update for ml-superb

ftshijt:
This is mainly for the case of potential issue in dict generation as 10min and 1h might have different token list during training.


ftshijt:
@sw005320 Hi Shinji, could you check if the PR is ready to merge? Since the modification only related to tags etc, it would be great if we can merge it for easier use for other users.

ftshijt:
> Sorry for my late actions...

No worries, many thanks!

Issue #5259, opened 2023-06-26
Increase default f0max for TTS, and Praat is a better pitch extractor than Dio?


iamanigeeit:
Hello,

400 Hz is fine for read speech, but not suitable for expressive / emotional TTS applications. For example, 
[0001_001491](https://github.com/espnet/espnet/files/11870982/0001_001491.txt) (please rename to .wav as Github doesn't allow .wav uploads) from the [Emotional Speech Dataset](https://hltsingapore.github.io/ESD/) has 500+ Hz segments according to Praat, and some other samples even have 800+ Hz.

Praat (you can play back the extracted pitch contour and verify the pink dots are accurate)
![image](https://github.com/espnet/espnet/assets/13131208/4eb12bd0-470a-41b3-a9f5-59735d9701c5)

Pitch (blue) extracted with Dio f0max at 400 Hz -- notice the cutoff at 400 Hz, and the missing syllables:
![image](https://github.com/espnet/espnet/assets/13131208/aea8742a-9919-43e5-ad9c-b5a2a3eee22a)

The Dio default is actually 800. Changing f0max to 800 helps, but there's some discontinuities:
![image](https://github.com/espnet/espnet/assets/13131208/d3bfceac-1cd8-476b-bdc6-8dd90d98b5b2)

I have no signal processing background, so if i'm doing something wrong, please let me know!

iamanigeeit:
Update:
RAPT algorithm is supposedly better than Praat for f0 extraction, but the Python implementation at [pysptk](https://github.com/r9y9/pysptk/tree/master) does not work.

Running
```
from pysptk import sptk
f0s = pysptk.sptk.swipe(audio, fs=16000, hopsize=256, max=800)
```
gives a similar plot to Praat with more random points:
![image](https://github.com/espnet/espnet/assets/13131208/36a300a8-1622-412a-9b85-81f487f0351b)

Running
```
import librosa
f0s = librosa.yin(audio, fmin=50, fmax=800, sr=16000, frame_length=1024, hop_length=256)
```
also gives a similar messy plot:
![image](https://github.com/espnet/espnet/assets/13131208/b11c73f5-f684-4396-abb4-5c7a039747f1)

Fortunately Praat has a nice Python wrapper at [Parselmouth](https://github.com/YannickJadoul/Parselmouth).
```
import parselmouth
sound = parselmouth.Sound(audio, 16000)
f0s = sound.to_pitch(pitch_floor=75.0, pitch_ceiling=800.0)
[...]

PR #5279, opened 2023-07-04
Add praat pitch

iamanigeeit:
For https://github.com/espnet/espnet/issues/5259

Unfortunately i lumped my own Dio changes into the same commit, but the f0min and f0max i use are the defaults for Dio. Also, `n_fft` is not used.

I am not sure if i need to include `parselmouth` in `setup.py` or not, so i have not added it.


ftshijt:
Many thanks for your contribution!

For the package, I would recommend to make it optional as other packages in https://github.com/espnet/espnet/tree/master/tools/installers

Specifically, please see the following suggestions:
- first, add a installer for praat in `tools/installers`
- then, add a corresponding entry in `tools/Makefile`, making it optional
- when user initialize a praat pitch extractor, it might be good to have a `assert` check to verify if the package is installed or not, and direct them to installer if the package is not installed
- for better integration, it would be great if you could also add a unit test for the function in `test/espnet2`

iamanigeeit:
Ok, i'll add them when i get PraatPitch to train with decent results.

PR #5285, opened 2023-07-06
A few updates for asr2 and hubert

simpleoier:
### Update
1. Update `hubert.sh` to support download pretrained ckpt.
2. Update data loading part in SSL feature readers, related to asr2 and hubert.
    * ESPnet data iterator is now supported for batch loader and audio reader.
    * Using batch loader can reduce the time it takes for feature dumping and pseudo-label generation.


egs2/TEMPLATE/asr1/pyscripts/feats/ssl_feature_utils.py
```diff
@@ -39,30 +41,61 @@ def format_feature_conf_str(feature_conf: str):
...
+        batch_bins=batch_bins,
+        shape_files=[utt2num_samples],
+    )
+    batches = list(sampler)
+    iterator = SequenceIterFactory(
+        dataset=dataset,
+        batches=batches,
+        collate_fn=CommonCollateFn(float_pad_value=0.0, int_pad_value=-1),
+        num_workers=2,
+    ).build_iter(0)
```
ftshijt, line 64: Is there a specific reason of using numel sampler and sequence iterator, I thought streaming iterator would be enough (it is also easier to trace as it does not go through the sampler)

simpleoier: Not really. I just think it is too complicated to define multiple choices of iterators / samplers. So I randomly chose one. Not sure which one is good enough. But in the end, we may find one that covers most of the use cases, which may be enough, right?

simpleoier: The reason for using numel sampler and sequence iterator instead of streaming iterator is because the utterances are sorted, thus less computations are wasted due to padding. Another reason for using numel sampler is that it is easier to avoid OOM error than others, e.g. sorted sampler.

ftshijt: That's fair, thanks for the clarification.

sw005320:
Do these new parts go through the test?

simpleoier:
Yeah, it will be tested in `test_integration_espnet2.sh`

sw005320:
OK, sounds good.
After we fix the CI issue (sorry about it), I'll merge it.

PR #5297, opened 2023-07-12
fix the s3prl frontend gradient backprop bug, ensuring feature_grad_mult=1.0

simpleoier:
### Fix the `feature_grad_mult` for s3prl frontend.
* When `feature_grad_mult` is set to `0.0` by s3prl, the if clause will not be executed.
* Sometime this if clause can still be executed when `feature_grad_mult=0.0`, due to float(0.0) precision issue.
@Emrys365 @YoshikiMas


YoshikiMas:
I can observe clear differences in `backward_time` and `grad_norm` in the 1-1989 batch, with this change. I will be able to check the 1st epoch performance within a few days.

YoshikiMas:
I get the accuracy of the first epoch that is comparable with that with the older version of S3PRL while causing artifacts in enhanced speech.

simpleoier:
@YoshikiMas Thank you!

sw005320:
@YoshikiMas and @simpleoier, can I merge this PR?

PR #5313, opened 2023-07-20
Add causal LM to Hugging Face Transformers Decoder

akreal:


b-flo (commented):
LGTM. Some notes:

- It may be better to split into two PR the new feature and the new recipe.
- Is there some things related to pythia that we could set by default? Symbols related options for example.

Also, unrelated but some files for main ASR task are becoming huge and hard to follow with all these options/features...

sw005320 (commented):
- Please modify the description of libritts in https://github.com/espnet/espnet/blob/master/egs2/README.md to include the ASR task
- year, as @b-flo mentioned, `asr_interface.py` is getting bigger and bigger. Since this PR is an extension of the HF decoder, which was already implemented here, we can keep it. However, if we have more extensions in the future, we can think of splitting `asr_interface.py` w/ or w/o the HF decoder.

akreal:
Thank you for the review, @b-flo and @sw005320 !

> It may be better to split into two PR the new feature and the new recipe.

I've removed the new recipe for now.

> Is there some things related to pythia that we could set by default? Symbols related options for example.

If you refer to the things like `sym_sos`, they need to be customized for HF models in the configuration, because ESPnet already has the default values which are different:
https://github.com/espnet/espnet/blob/42f31823df4b4fab0dc68bd34a3ac5c08fe6034b/espnet2/asr/espnet_model.py#L65-L68

akreal:
The new test fails in the CI, but works for me locally. I'll investigate this.

akreal:
I reproduced the CI problem under python 3.7. It turned out that the latest Transformers do not support python 3.7, so the tests were running with the previous version of Transformers and there were some problem with the models I used for testing. I changed the models and they seem to work in python 3.7 and the previous Transformers version.

sw005320:
Thanks a lot!

PR #5315, opened 2023-07-20
Update README.md

sw005320:
I added several recent references (but not for ST v2, which will be added after some PRs are merged).
Also, I fixed some typos.


sw005320:
I merged it since it would not be related to CI.

PR #5316, opened 2023-07-20
Update setup.py

sw005320:
Fix CIs due to Pillow
https://github.com/JaidedAI/EasyOCR/issues/1077


sw005320:
Confirmed that `test_integration_espnet1.sh` is successfully finished.

PR #5317, opened 2023-07-20
Update test_integration_espnet1.sh

sw005320:
Remove some espnet1 integration test


sw005320:
It is the same as #5328

PR #5318, opened 2023-07-21
[WIP] CoVoST2 ASR2 recipe and new ST2 recipe

simpleoier:
## ST2 recipe
A combination tasks of ASR + ST + MT, which could use speech discrete tokens and text transcriptions together.
(More details will be filled)

## ASR2 recipe
Add ASR2 recipe for CoVoST2 data.


egs2/TEMPLATE/st2/setup.sh
```diff
@@ -0,0 +1,58 @@
...
+for f in st2.sh path.sh ; do
+    target=../../TEMPLATE/st2/"${f}"
+    ln -sf "${target}" "${dir}"
+    targets+="${dir}/${target} "
+done
+
+
+# Symlinks to TEMPLATE
+for f in db.sh scripts pyscripts utils steps; do
+    target=../../TEMPLATE/asr1/"${f}"
```
ftshijt, line 53:
```suggestion
    target=../../TEMPLATE/st2/"${f}"
```

egs2/TEMPLATE/st2/st2.sh
```diff
@@ -0,0 +1,2023 @@
...
+st_exp=        # Specify the directory path for ST experiment.
+                # If this option is specified, st_tag is ignored.
+st_stats_dir=  # Specify the directory path for ST statistics.
+st_config=     # Config for st model training.
+st_args=       # Arguments for st model training, e.g., "--max_epoch 10".
+                # Note that it will overwrite args in st config.
+ignore_init_mismatch=false      # Ignore initial mismatch
+num_splits_st=1                # Number of splitting for lm corpus.
+src_case="ts"
+src_lang=wavlm_large_21_km2000  # source language abbrev. id (e.g., wavlm_large_21_km2000)
```
ftshijt, line 115: As for ST, we use source language transcript to conduct multi-task (without it, training from scratch would be very difficult). Do you have a reason of change the src_lang, instead of adding arguments for discrete processing?

egs2/covost2/asr2/README.md
```diff
@@ -0,0 +1,53 @@
+# ASR2 recipe for LibriSpeech960
```
ftshijt, line 1: We can safely remove this file

ftshijt (commented):
The main implementation looks good. Only one concern in the potential confusion of "src" in ST task.

In ST task, we usually would do multi-task, while predicting source language transcript, where we refer as "src". It is different from the input speech. Could you please factor out that part if possible instead of reusing? Cause it would potentially bring a lot of confusions between the st1 and st2 implementation.

egs2/TEMPLATE/st2/st2.sh
simpleoier: This is a good point. Now the names of speech tokens / src text language / src lang / tgt lang are a bit confused. Let me think of some new names.

simpleoier: BTW, now it is actually multi-task learning (asr + st, mt is optional). It is okay to train from scratch.

simpleoier: Renamed as `speech_token_case/lang` and `src_tgt_text_case/lang`.

egs2/TEMPLATE/asr1/pyscripts/feats/dump_km_label.py
sw005320, line 1: Thanks!
What does it look like?
Could you paste an example?

egs2/TEMPLATE/asr2/README.md
```diff
@@ -140,30 +142,57 @@ See also:
...
+2. cd egs2/<recipe name>/asr2
+3. ../../TEMPLATE/asr2/setup.sh . # create the default files
+4. cd local
+5. ln -s ../../asr1/local/* .     # add symlinks for data preparation files in asr1 except for path.sh
+6. cd ../
+7. cp ../../librispeech/asr2/run.sh .         # copy run.sh
+8. cp ../../librispeech/asr2/conf/tuning/train_discrete_asr_e_branchformer1.yaml conf/ # copy training conf
+9. cp ../../librispeech/asr2/conf/decode_ctc0.3.yaml conf/     # copy confs
+10. EDIT run.sh by checking ../asr1/run.sh
+  a. We may skip an LM
```
sw005320, line 157:
```suggestion
  a. We may skip an LM by adding an option `--use_lm false`
```

egs2/TEMPLATE/asr2/README.md
```diff
@@ -140,30 +142,57 @@ See also:
...
 
-Run `run.sh`, which conducts all of the stages explained above.
-```sh
-$ ./run.sh
+## Tips for asr2 experiments
+* **Kmeans is important**
+   * SSL model choice can affect the performance a lot, e.g. wavlm models may not work for non-English data,
+   * Layer selection is also important: different layers retain different information. For example, based the training criterion, the 24-th layer of HuBERT_large is trying to match the information from HuBERT_base layer 9. If you didn't have experience, please refer to the Fig. 4 of this [CCA paper](https://arxiv.org/pdf/2211.03929.pdf), which is usually helpful.
+   * Number of kmeans clusters also affect the variance in pronunciation, etc.
+   * Please check the kmeans labels in `dump/extracted/{kmeans_feat_type}/layer{layer}/{dset}/pseudo_label_km{ncluseters}.txt`. In my experience, a good km result for ASR should have an obvious pattern of repeatitions, e.g.
```
sw005320, line 168: Maybe, you can add a bad example as well?

egs2/TEMPLATE/st2/st2.sh
sw005320, line 1: @ftshijt, can you review it?

egs2/librispeech/asr2/run.sh
```diff
@@ -28,7 +28,7 @@ src_case="rm"
 tgt_case="ts"
 
 ./asr2.sh \
-    --kmeans_opts "--batch_bins 4800000" \
+    --kmeans_opts "--batch_bins 4800000 --nj 4" \
```
sw005320, line 31: You may need to increase `--num_threads` to deal with large memory consumptions in scikit-learn?
(Ideally, I want you to solve it by avoiding using such less-refined implementation.

espnet2/bin/mt_inference.py
```diff
@@ -58,17 +71,42 @@ def __init__(
...
         assert check_argument_types()
 
+        if quantize_mt_model or quantize_lm:
+            if quantize_dtype == "float16" and torch.__version__ < LooseVersion(
+                "1.5.0"
+            ):
+                raise ValueError(
+                    "float16 dtype for dynamic quantization is not supported with "
+                    "torch version < 1.5.0. Switch to qint8 dtype instead."
+                )
```
sw005320, line 89: since CI does not support 1.5.0, we can remove these lines

espnet2/tasks/mt.py
sw005320, line 1: CTC BPE token part looks complicated and tricky.
It requires some documentation (in the source code and asr2 or st2 documents).

espnet2/train/preprocessor.py
```diff
@@ -556,6 +556,7 @@ def __init__(
         speech_name: str = "speech",
         text_name: List[str] = ["text"],
         tokenizer_encode_conf: List[Dict] = [dict(), dict()],
+        not_available_symbol: str = None,
```
sw005320, line 559: Can you explain about it and embed the explanation in the source coed?

egs2/TEMPLATE/asr1/pyscripts/feats/dump_km_label.py
simpleoier: It is like the tqdm progress bar, with `percentage progress`, `current batch / total_batch`, `time_so_far / est_time_to_finish`, `time_per_batch`. 
```
0%|          | 1/1577 [00:20<8:49:13, 20.15s/it]
```

egs2/TEMPLATE/asr2/README.md
simpleoier: Sure.

egs2/librispeech/asr2/run.sh
simpleoier: OK. Working on this item.

espnet2/train/preprocessor.py
simpleoier: I'll put the following explanation down below.
```
        # not_available_symbol is a special symbol as placeholder in the text. e.g.
        #      "utt_id <na>" an item in the text input
        # Then such samples will not have the corresponding text signals.
        # The resulting tensor would be processed as torch.longtensor([-1])
```

espnet2/tasks/mt.py
simpleoier: The idea is simple. Maybe my way is a bit complicated. In st2, we use different the text targets for CTC and attention-decoder.
* CTC target: ASR transcriptions for ASR or ST, while <not_available> is used in MT.
* Att-Dec target: ASR transcriptions for ASR, while translation transcriptions for ST / MT.
For this purpose, we need different text input as data. In ESPnet preprocessor, the number of tokenizer should match the text input.
```
        assert (
            len(token_type) == len(token_list) == len(bpemodel) == len(text_name)
        ), "token_type, token_list, bpemodel, or processing text_name mismatched"
```

egs2/TEMPLATE/st2/st2.sh
```diff
@@ -0,0 +1,2066 @@
...
+st_config=     # Config for st model training.
+st_args=       # Arguments for st model training, e.g., "--max_epoch 10".
+                # Note that it will overwrite args in st config.
+ignore_init_mismatch=false      # Ignore initial mismatch
+num_splits_st=1                 # Number of splitting for lm corpus.
+speech_token_case="ts"          # speech discrete token  case. ts: true sequence, rm: remove duplicated tokens
+speech_token_lang="wavlm_large_21_km2000"  # speech discrete token type abbrev. id (e.g., wavlm_large_21_km2000)
+src_tgt_text_case="lc.rm"       # source / target transcript case. Note, all source / target text should use the same case for now.
+src_tgt_text_lang=en            # source / target language abbrev. id (e.g., en). Multiple langs are supported to support multiple tasks, with space between (e.g., "es/en"), from data's perspect of view, src_lang of text is the first.multiple tasks, with space between (e.g., "es/en"), from data's perspect of view, src_lang of text is the first.
+tgt_tasks="asr/st"              # task abbrev. id (e.g., st). Multiple tasks are supported to support multiple tasks, with space between (e.g., "asr/st")
```
ftshijt, line 118: space between or "/" between?

egs2/TEMPLATE/st2/st2.sh
```diff
@@ -0,0 +1,2066 @@
+#!/usr/bin/env bash
```
ftshijt, line 1: The script looks great!

One issue remains on my side is the support of multi-reference scenarios for ST evaluation (I double checked the support, but it seems not supported in preparation yet) Please refer  https://github.com/espnet/espnet/blob/master/egs2/TEMPLATE/st1/st.sh#L547-L553 for some details on how we process those

ftshijt: For more information on the name:
https://github.com/espnet/espnet/blob/master/egs2/TEMPLATE/st1/st.sh#L1585C75-L1585C75

egs2/TEMPLATE/st2/st2.sh
```diff
@@ -0,0 +1,2066 @@
...
+    _corpus="${_task%/*}"
+    _model_name="${_creator_name}/${_corpus}_$(basename ${packed_model} .zip)"
+
+    # copy files in ${dir_repo}
+    unzip -o ${packed_model} -d ${dir_repo}
+    # Generate description file
+    # shellcheck disable=SC2034
+    hf_task=automatic-speech-recognition
+    # shellcheck disable=SC2034
+    espnet_task=ASR
```
ftshijt, line 2048: consider changing it?

egs2/TEMPLATE/st2/st2.sh
ftshijt: Since it was for multi-task, is it a good idea to put in st task or a more general one?

I'm asking cause the major reason we split st from asr in the previous context is:
- the different architectures between each other (for st only tasks, we also have a few unique architecture designs for each component (e.g., separate asr/mt decoder, two-pass framework with multi-decoder, etc.)
- data preparation (design for tgt_text, src_text, and src_speech)
- specific evaluation (bleu calculation and multi-reference support)

however, I feel many of the above parts are not shared (e.g., the architecture is still discrete asr; we skip the joint-task framework, instead, we do either asr or st in a single run; no support to multi-reference scoring)

Given the above reasons, I lean to call it s2t2 instead of st2. Please let me know you thoughts!

PR #5321, opened 2023-07-21
Docker Update

Fhrozen:
# Whaaat?

Update the docker devel file.

# Why?

Update the docker image containers with new installation and fix issues of espnet2 missing.


sw005320:
Can you also update a docker-related document?
https://github.com/espnet/espnet/blob/master/doc/docker.md should be updated.
You can make a separate PR.

Fhrozen:
@sw005320  I will modify the documentation in a different PR.

PR #5322, opened 2023-07-21
fix a data preparation issue for librimix recipe.

LiChenda:
In `egs2/librimix/asr1/local/data.sh`,  L116, if `mix_f` matches multiple files, the `grep` command (in some version) will output the specific matched file name (which is not the expected output) and lead to a wrong `utt_id` in data preparation.

So I updated the data preparation scripts to avoid this issue.


sw005320:
Thanks, @LiChenda!

PR #5323, opened 2023-07-21
Multilingual Librispeech (MLS) refactor ASR1 recipe

juice500ml:
### TO-DO List
- [x] Bug: Missing `db.sh`
- [x] Bug: If there's no `download` folder, `local/data.sh` dies.
- [x] Bug: Default downloads only one dataset (spanish)
- [x] Bug: Removes files even if it's a normal tar.gz
- [x] Bug: German is missing
- [x] Bug: Wrong parameter passed to fix_data_dir.sh
- [x] Feature: Add md5sum check for downloaded files
- [x] Refactor: Implement simple hashtable library in shell + add corresponding tests
    - [x] Dict generation
    - [x] Dict key add (set)
    - [x] Dict key delete
    - [x] Dict key get
    - [x] Dict key traverse
    - [x] Dict value traverse
- [x] Refactor: `local/data_prep.py` uses os.path rather than pathlib.Path
- [x] Feature: Implement 1hr/9hr/full data splits
- [x] Feature: Implement merged datasets (mls_all_train = mls_de_train + mls_en_train + ...)


ftshijt:
For a quick note on downloading 1 data default, those it is a multilingual dataset, we usually default support a monolingual recipe (as in commonvoice and voxforge), cause there are such needs from people. So I guess we may keep the convention, especially considering that the data is very large for MLS on `en`.

juice500ml:
@ftshijt Great point! I'll modify the code to accommodate both cases: (1) all langs and (2) specified lang.

utils/simple_dict.sh
```diff
@@ -0,0 +1,82 @@
+#!/usr/bin/env bash
```
ftshijt, line 1: Using `utils` is for espnet1 style, where we are not using mostly (and recommend not to use, since we actually remove the checks to utils in espnet2 case). For this function, you might add it to `egs2/TEMPLATE/asr1/scripts` instead of here

ftshijt (commented):
LGTM! Just a minor comment.

utils/simple_dict.sh
juice500ml: Aha, didn't know that! Thanks for pointing that out!! My intention of simple_dict.sh was to design a shell library that can be used to other projects too (Python's basic data structure is list and dict, so I thought we needed dict on shell for many python programmers, whereas list is simple enough already, so didn't implement it). So, (1) will it be a good idea to maintain a shell library in the first place, and (2) if so, where should I put it for the uses in espnet2?

ftshijt: Sorry for the late response! It would be good to have utility functions, but in that case, I would still recommend putting the one in `egs2/TEMPLATE/asr1/scripts/utils` as we won't include utils as in the `path.sh` and almost all the tasks have symlink to asr1

ftshijt: Or if @sw005320 would have other ideas, please let us know

ftshijt:
LGTM! Let's merge it

utils/simple_dict.sh
sw005320: For (2), as Jiatong suggested, we usually put the recipe-related utility espne2 scripts in `egs2/TEMPLATE/asr1/scripts `

juice500ml: Yeah, I'll relocate that file in the https://github.com/espnet/espnet/pull/5441 PR! I don't want the merge conflict to be too severe (because that PR and this PR has diverged a bit), so I'll try to merge this PR ASAP, and then modify that in the follow-up PR.

ftshijt:
Many thanks!

PR #5328, opened 2023-07-21
remove some tests in espnet1 integration test

sw005320:


sw005320:
We can reduce the time from 14 to 9 minutes.

PR #5341, opened 2023-07-22
Update Scoring for Speech Summarization from NLG-Eval to Huggingface Evaluate

roshansh-cmu:
The NLG-Eval doesn't seem to be updated recently and is incompatible with the latest versions of python, numpy. 

This PR modifies all scoring to rely on the huggingface evaluate package instead. 

Note:

- The ROUGE-L scores are slightly higher than previously obtained using NLG-Eval. However, the NLP community uses evaluate nowadays to compute scores, so it makes sense to retain this. 
- I also updated local/score.sh because the ls command format does not work with certain shell types 
- I removed any checks on nlgeval and bert_score wherever they appear and added the evaluate package to setup.py for installation.


roshansh-cmu:
Requesting a review from @Jungjee if you have any time ? Thanks !

setup.py
```diff
@@ -87,6 +87,7 @@
         "fairscale",
         "transformers",
         "gtn==0.0.0",
+        "evaluate",
```
Jungjee, line 90: can you add the version that worked for you?

Jungjee (commented):
LGTM! Just added one minor comment.

setup.py
roshansh-cmu: I think the evaluate codebase is updated with transformers, so maybe getting the latest version like transformers is best ?

sw005320:
LGTM!

sw005320:
I also made a label SSUM by considering the growth of this area. This is the first PR/issue having it.

PR #5345, opened 2023-07-22
asr2 recipe for WenetSpeech

anthony-wss:
## What?

Implement the discrete-token ASR for WenetSpeech corpus.

## Why?

This corpus has only asr1 recipe, so I am extending it.

## See also

None. Still in progress.


egs2/wenetspeech/asr2/run.sh
```diff
@@ -0,0 +1,51 @@
...
+src_nbpe=6000   # I use src_nbpe=6000 for 2000-cluster kmeans.
+tgt_nbpe=5000   # if token_joint is True, then only tgt_nbpe is used
+
+# ts: true sequence
+# rm: deduplicated sequence which removes duplicated tokens
+src_case="rm"
+tgt_case="ts"
+
+./asr2.sh \
+    --kmeans_opts "--batch_bins 4800000" \
```
simpleoier, line 31: From the discussion with Shinji and others, we may explicitly specify the number of jobs used in kmeans by default. For example, if we use 4 jobs, then we can use:
```
--kmeans_opts "--batch_bins 4800000 --nj 4" \
```

simpleoier (approved):
Thanks! It looks good to me in general.

egs2/wenetspeech/asr2/run.sh
```diff
@@ -5,33 +5,42 @@ set -e
 set -u
 set -o pipefail
 
-
-kmeans_feature="wavlm_large/21"  # use model_type/layer_index
+export CUDA_VISIBLE_DEVICES=4,5,6,7
```
simpleoier, line 8: Please clean this script, especially those related to your own environment.

sw005320:
Can you add a result to README.md that is similar to the other recipe?

PR #5349, opened 2023-07-22
Update the usage

sw005320:
## What?

Update the usage page

## Why?

The [usage part](https://espnet.github.io/espnet/tutorial.html) is outdated. Also, it mixes the espent1 and espnet2 explanations.

- [x] prepare the document for espnet1 to make the usage separate
- [x] separate the espent1 and espnet2 explanations


kan-bayashi:
I forgot auto merge update. I will do it today.

PR #5320, opened 2023-07-21
PR Labels for CI control 

Fhrozen:
# What

Implement the use of PR labels in the CI tests.

# Why

Reduce the time of the CI by constraining the test to the respective modifications.

PR #5351, opened 2023-07-22
Fix documentation skip CI

Fhrozen:
## What?

Fix CI variable

## Why?

CI were skipping doc verification due to mislabel.

## See also

#5320


kan-bayashi (approved):
Sorry for my mistake! Thanks!

PR #5358, opened 2023-07-23
ESPnet-SPK: EER validation efficiency improvement

Jungjee:
## What?

Updates ESPnet-SPk EER validation mechanism

## Why?

Current version extracts all speaker embeddings from trials where the duplication is severe.
In case of Vox1-O evaluation protocol, there are 37k trials but only 5k unique utterances.
Previous version would extract 74k speaker embeddings whereas new version would extract 5k.


espnet2/train/spk_trainer.py
```diff
@@ -72,41 +72,65 @@ def validate_one_epoch(
...
+                utt_id, batch["speech"], batch["speech2"]
+            ):
+                _utt_id_1, _utt_id_2 = _utt_id.split("*")
+                if _utt_id_1 not in utt_id_list:
+                    utt_id_list.append(_utt_id_1)
+                    speech_list.append(_speech)
+                if _utt_id_2 not in utt_id_list:
+                    utt_id_list.append(_utt_id_2)
+                    speech_list.append(_speech2)
+
```
sw005320, line 99: How about adding logging to report some statistics?
- at least some counts (e.g., speaker embedding vectors, how many similarities)
- other statistics, e.g., the global mean/variance of the speaker embedding vectors, scores, duration, etc.

This is just a suggestion, but this kind of statistics makes us easy to detect some irregular behaviors in advance.

sw005320: Also, if we provide logging before/after speaker embedding/scoring computations give us more information about the computational cost.

sw005320 (commented):
LGTM.

Can you add a comment to the header of the source about why we need a speaker trainer?
(I think this is because the validation scheme differs from the normal task, right)?

Jungjee:
> I think this is because the validation scheme differs from the normal task, right?

Yes, correct! I will add comments and statistics.

PR #5361, opened 2023-07-23
Update TEMPLATE_HF_Readme.md (fix bash typo)

akreal:
## What?

Fix bash typo.

## Why?

It pops up when I try to upload a model to the HF hub with `./run.sh --skip_upload_hf false --stage 18 --hf_repo espnet/xyz`.


sw005320:
Thanks, @akreal!

PR #5362, opened 2023-07-23
Couple of tweaks for asr2.sh for the HF hub upload

akreal:
## What?

 - Set the `lang` variable
 - Enable LFS tracking for `*.mdl` files

## Why?

- I think setting the language variable is needed (HF hub complains when language is set to `noinfo`).
- I'm not sure if setting the LFS tracking for `*.mdl` files is needed. Does it work without it for everyone else? I have not investigated what is the expected behaviour here, but in my environment HF hub complained that `*.mdl` is not under LFS.


simpleoier (approved):
Thanks! This modification is very necessary.

sw005320:
Thanks a lot!

PR #5368, opened 2023-07-24
add --num-workers 0 option to truck the coverage

kan-bayashi:
## What?

- Add `--num_workers 0` for integration test

## Why?

- If we launch data loader in another thread, coverage does not truck the processing.

## See also

- https://shinjiwjhu.slack.com/archives/C05H3BMJMHT/p1690127512703049


sw005320:
Many thanks, @kan-bayashi!
I added @Jungjee to this thread.

sw005320:
Thanks a lot!
This is very important since we are starting to upgrade the data preprocessing part.

PR #5370, opened 2023-07-24
Adding general data augmentation methods for speech preprocessing

Emrys365:
## What?

This PR adds a series of data augmentation techniques for preprocessing speech data in various tasks:

The supported data augmentation techniques include:
```yaml
effects_dict = {
    "lowpass": lowpass_filtering,
    "highpass": highpass_filtering,
    "bandpass": bandpass_filtering,
    "bandreject": bandreject_filtering,
    "contrast": contrast,
    "equalization": equalization_filtering,
    "pitch_shift": pitch_shift,
    "speed_perturb": speed_perturb,
    "time_stretch": time_stretch,
    "preemphasis": preemphasis,
    "deemphasis": deemphasis,
    "clipping": clipping,
    "polarity_inverse": polarity_inverse,
    "reverse": reverse,
    "corrupt_phase": corrupt_phase,
}
```

The data augmentation methods can be easily configured via the yaml file:
```yaml
preprocessor: default
preprocessor_conf:
    fs: 16000
    data_aug_effects:   # no need to set the "sample_rate" argument for each effect here
        - [0.1, "contrast", {"enhancement_amount": 75.0}]
        - [0.1, "highpass", {"cutoff_freq": 5000, "Q": 0.707}]
        - [0.1, "equalization", {"center_freq": 1000, "gain": 0, "Q": 0.707}]
        - - 0.1
          - - [0.3, "speed_perturb", {"factor": 0.9}]
            - [0.3, "speed_perturb", {"factor": 1.1}]
    data_aug_num: [1, 4]
    data_aug_prob: 1.0
```

> The above configuration will apply 1~4 different data augmentations during training, including "contrast", "highpass", "equalization", and "speed_perturb", each having a weight of 0.1 to be sampled.
>
> Their arguments are specified respectively.
>
> For the last augmentation, the two "speed_perturb" methods are mutually exclusive methods and only one will be applied.

## Why?

Current preprocessors are not flexible to support applying various data augmentation methods at the same time.


ci/test_integration_espnet2.sh
```diff
@@ -170,9 +176,11 @@ if python -c 'import torch as t; from packaging.version import parse as L; asser
     for t in ${feats_types}; do
         echo "==== feats_type=${t} with preprocessor ==="
         ./run.sh --ngpu 0 --stage 2 --stop-stage 10 --skip-upload false --feats-type "${t}" --ref-num 1 --python "${python}" \
-            --extra_wav_list "rirs.scp noises.scp" --enh_config ./conf/train_with_preprocessor_debug.yaml --enh-args "--num_workers 0"
-        ./run.sh --ngpu 0 --stage 2 --stop-stage 10 --skip-upload false --feats-type "${t}" --ref-num 1 --python "${python}" \
-            --enh_config conf/train_with_dynamic_mixing_debug.yaml --ref-num 2 --enh-args "--num_workers 0"
+            --extra_wav_list "rirs.scp noises.scp" --enh_config ./conf/train_with_preprocessor_debug.yaml
+        ./run.sh --ngpu 0 --stage 5 --stop-stage 10 --skip-upload false --feats-type "${t}" --ref-num 1 --python "${python}" \
+            --enh_config conf/train_with_data_aug_debug.yaml
+        ./run.sh --ngpu 0 --stage 2 --stop-stage 10 --skip-upload false --feats-type "${t}" --ref-num 2 --python "${python}" \
```
sw005320, line 182: Please revert `"--num_workers 0"` otherwise, it does not go through the coverage test.

sw005320:
@Jungjee, can you review this PR?

ci/test_integration_espnet2.sh
Emrys365: done

espnet2/layers/augmentation.py
```diff
@@ -0,0 +1,535 @@
...
+    Note: this function is slow.
+
+    Args:
+        waveform (torch.Tensor): audio signal (..., time)
+        sample_rate (int): sampling rate in Hz
+        n_steps (int): the (fractional) steps to shift the pitch
+            -4 for shifting pitch down by 4/`bins_per_octave` octaves
+            4 for shifting pitch up by 4/`bins_per_octave` octaves
+        bins_per_octave (int): number of steps per octave
+        n_fft (int): length of FFT (in second)
```
Jungjee, line 248: float?

espnet2/layers/augmentation.py
```diff
@@ -0,0 +1,535 @@
...
+    """
+    orig_freq = sample_rate
+    source_sample_rate = int(factor * orig_freq)
+    target_sample_rate = int(orig_freq)
+
+    gcd = math.gcd(source_sample_rate, target_sample_rate)
+    source_sample_rate = source_sample_rate // gcd
+    target_sample_rate = target_sample_rate // gcd
+
+    ret = torchaudio.functional.resample(
```
Jungjee, line 303: Did you consider applying one without pitch shift? Would it cause severe more computation? 
Also, how's the training training speed with this augment (bottlenecks in loading)?

espnet2/layers/augmentation.py
```diff
@@ -0,0 +1,535 @@
...
+    if win_length is None:
+        win_length = n_fft
+    if window is not None:
+        window_func = getattr(torch, f"{window}_window")
+        window = window_func(win_length, dtype=waveform.dtype, device=waveform.device)
+    spec = torch.stft(
+        waveform, n_fft, hop_length, win_length, window=window, return_complex=True
+    )
+    freq = spec.size(-2)
+    phase_advance = torch.linspace(0, math.pi * hop_length, freq)[..., None]
```
Jungjee, line 350: is [..., None] equivalent to .unsqueeze(-1) here?

espnet2/layers/augmentation.py
```diff
@@ -0,0 +1,535 @@
...
+            "ULAW" (mu-law), and "ALAW" (a-law).
+            For more details see torchaudio.backend.sox_io_backend.save().
+        bits_per_sample (int or None, optional): change the bit depth
+            for the supported formats
+            For more details see torchaudio.backend.sox_io_backend.save().
+
+    Returns:
+        ret (torch.Tensor): compressed signal (..., time)
+    """
+    ret = torchaudio.functional.apply_codec(
```
Jungjee, line 396: how about adding some warning or exception to not be called for unwanted torch version? (or in a different place because if you put that here, it can be called too often)

espnet2/train/preprocessor.py
```diff
@@ -220,12 +225,27 @@ def __init__(
...
+            self.data_aug = None
+        self.data_aug_prob = data_aug_prob
+
+    def _convolve_rir(self, speech, power, rirs, tgt_fs=None, single_channel=False):
         rir_path = np.random.choice(rirs)
         rir = None
         if rir_path is not None:
-            rir, _ = soundfile.read(rir_path, dtype=np.float64, always_2d=True)
+            rir, fs = soundfile.read(rir_path, dtype=np.float64, always_2d=True)
+            if tgt_fs and fs != tgt_fs:
```
Jungjee, line 242: maybe better to warn of raise something because sample rate mismatch may not be intended

Jungjee (commented):
LGTM, mostly added suggestions/questions, not mandatory.

espnet2/layers/augmentation.py
Emrys365: Thanks for noticing that!

espnet2/layers/augmentation.py
Emrys365: `speed_perturb` and `time_stretch` are two different time scaling method. The former changes the pitch while the latter does not. I think it is dependent on the use case. So I just provide both for the user to choose.

I haven't strictly tested the speed difference yet. Will do some test later.

espnet2/layers/augmentation.py
Emrys365: Yes. These are just the same operations.

espnet2/layers/augmentation.py
Emrys365: For now, I think I can just raise NotImplementedError for this function.

espnet2/train/preprocessor.py
Emrys365: Done

Emrys365:
> LGTM, mostly added suggestions/questions, not mandatory.

Thanks a lot!

Emrys365:
Strangely, I can locally pass the test in `ci/test_import_all.py` with Python 3.10.12 and PyTorch 1.13.1. I'm not sure what is the cause of the segmental fault in the CI test.

sw005320:
Thanks, @Emrys365!

PR #5376, opened 2023-07-27
AAI task

bloodraven66:
## What?

This is a new task in espnet - acoustic to articulatory inversion (aai) or speech inversion

## Why?

I'm not sure if there is sufficient interest to integrate into espnet, but this branch could be of use to speech production researchers, so adding it here for visibility. I've observed performance gains using ASR encoders in espnet.

## See also

Stages-
Stage 1-4: Normal ASR data prep states
Stage 5: Collect stats
Stage 6: Train
Stage 7: Decode and score

Regarding data prep - 
articulatory features to be used in `text` file as 
`utt_id` `path_to_articulatory_features.pt`

Metrics-
Corpus level Pearson's correlation coefficient shown at end of stage 7


b-flo:
> I'm not sure if there is sufficient interest to integrate into espnet, but this branch could be of use to speech production researchers, so adding it here for visibility. I've observed performance gains using ASR encoders in espnet.

Thanks a lot for the contribution!

It may be but some discussion is needed for the integration, the PR in its current form is unreviewable/untrackable. At quick glance, I would change the following:

- Split the PR into two: feature addition and recipe
- Symlink every shared utility with asr or other tasks and recipe ones.
- Inherit/re-use utilities, tools, etc from dependencies (e.g.: a lot of scripts I've seen adapted from fairseq and k2 are -almost- identical minus formatting)

From there, it would be easier to work on the task design and how it could be related to other one(s).

sw005320:
@bloodraven66, this is super cool!
Let's try to make it in good shape.
Speech production is an important theme.

bloodraven66:
> It may be but some discussion is needed for the integration, the PR in its current form is unreviewable/untrackable. At quick glance, I would change the following:
> 
> * Split the PR into two: feature addition and recipe
> * Symlink every shared utility with asr or other tasks and recipe ones.
> * Inherit/re-use utilities, tools, etc from dependencies (e.g.: a lot of scripts I've seen adapted from fairseq and k2 are -almost- identical minus formatting)
> 
> From there, it would be easier to work on the task design and how it could be related to other one(s).

Alright, let me change the commits

b-flo:
> @b-flo , how does it look now? I've removed the recipe and symlinked existing copies.

Thanks a lot, it's in good shape for a review now!

PR #5235, opened 2023-06-16
CFSD, SECS metrics for TTS 

imdanboy:
Hi, I tried to implement following TTS objective metrics (#1665 ) by simply utilizing related espnet script and the pretrained models.

1) Conditional Frechet Speech Distance (CFSD)
2) Speaker Embedding Cosine Similarity (SECS)

(Example TTS paper where those metrics are used: [ADAPTER-BASED EXTENSION OF MULTI-SPEAKER TEXT-TO-SPEECH MODEL FOR NEW SPEAKERS](https://arxiv.org/abs/2211.00585))

Example results on LJSPEECH eval set:
||CFSD|SECS|MCD|logF0 RMSE|
|---|---|---|---|---|
|[JETS](https://huggingface.co/imdanboy/ljspeech_tts_train_jets_raw_phn_tacotron_g2p_en_no_space_train.total_count.ave)|6.6239 ± 2.4985|0.8465 ± 0.0534|6.6817 ± 0.5373|0.2905 ± 0.0682|
|[VITS](https://zenodo.org/record/5443814)|6.8165 ± 1.6455|0.8440 ± 0.0585|6.9935 ± 0.6215|0.2812 ± 0.0589|
|[joint_finetune](https://zenodo.org/record/5498896)|6.1623 ± 1.4250|0.8481 ± 0.0564|6.7344 ± 0.6166|0.2839 ± 0.0739|

How about this?

PR #5377, opened 2023-07-28
add test script on evaluate_secs, cfsd

imdanboy:
## What?

Add test script for `evaluate_secs, evaluate_cfsd`

## Why?

Also change evaluate script to accept argument of download_dir so that the pretrained model is cleaned up after testing

## See also

#5235


sw005320:
Sorry for my late response.
This PR looks very good to me!

PR #5378, opened 2023-07-28
minor fix: correct wrong comments

imdanboy:
## What?

Correct wrong comments

## Why?

According to the forward method of `LogMel` class, `spec` argument seems to inform wrong tensor shape.


ftshijt:
Thanks for your fix!

PR #5379, opened 2023-07-28
Add LibriTTS-R recipe

ShigekiKarita:
## What?

Add a new LibriTTS-R recipe by forking egs2/libritts/tts1

## Why?

In our experiments, it improves TTS quality from the existing LibriTTS.

## See also

For details, see https://arxiv.org/abs/2305.18802


kan-bayashi (approved):
LGTM

PR #5365, opened 2023-07-24
ESPnet-SPK: sampler

Jungjee:
## What?

A specialized data sampling strategy for speaker recognition is added. 
It can be used (recommended) for classification-trained tasks such as language identification and anti-spoofing, etc.

## Why?

Has a big impact on performance, but not implemented yet.

PR #5385, opened 2023-07-31
ESPnet-spk: add more model architectures

Jungjee:
## What?

Add widely used state-of-the-art models. ECAPA-TDNN and MFA-Conformer.

## Why?

Extension of supported models are necessary.

Should be merged after #5365 is merged


Jungjee:
Will be merged with #5408

PR #5387, opened 2023-08-02
Add quotation marks to install_datasets.sh

qmeeus:
## What?

Added quotation marks around `datasets>=2.0.0` in `tools/installers/install_datasets.sh`

## Why?

So that bash does not consider the redirection in line `python3 -m pip install datasets>=2.0.0`


ftshijt:
Thanks for your udpate!

PR #5414, opened 2023-08-17
Update the Enh framework to support training with variable numbers of speakers

Emrys365:
## What?

This PR adds support for variable numbers of speakers when training speech separation models.

Major updates are added in
  * [egs2/TEMPLATE/enh1/enh.sh](https://github.com/Emrys365/espnet/blob/espnet2_enh/egs2/TEMPLATE/enh1/enh.sh)
    * A new argument `--variable_num_refs` is added to specify whether we are training models with variable numbers of speakers in each mixture sample.
    * A new multi-column format for reference files including `spk1.scp`, `dereverb1.scp`, and `enroll_spk1.scp` is introduced. Each row may contain variable columns, where each column starting from the 2nd column represents a different reference speaker.
  * [espnet2/fileio/multi_sound_scp.py](https://github.com/Emrys365/espnet/blob/espnet2_enh/espnet2/fileio/multi_sound_scp.py)
    * A new sound loader to created to support loading different numbers of audios in different samples. The corresponding data type `variable_columns_sound` is added to [espnet2/train/dataset.py](https://github.com/Emrys365/espnet/blob/espnet2_enh/espnet2/train/dataset.py) and [espnet2/train/iterable_dataset.py](https://github.com/Emrys365/espnet/blob/espnet2_enh/espnet2/train/iterable_dataset.py).
    * This is used to load `spk1.scp` and `dereverb1.scp` (if provided) when `variable_num_refs` is true.
    * Note that `enroll_spk1.scp` is loaded as the `text` type, so there is no problem if the enrollment audios of the same sample have different lengths.
  * [espnet2/train/preprocessor.py](https://github.com/Emrys365/espnet/blob/espnet2_enh/espnet2/train/preprocessor.py)
    * A new argument `speech_segment: Optional[int] = None` is added to `EnhPreprocessor` and `TSEPreprocessor` to allow only loading a fixed-length segment from each audio. This is useful if we want to control the balance between data of different lengths.
    * A new argument `avoid_allzero_segment: bool = True` is added to `EnhPreprocessor` and `TSEPreprocessor`. This argument is only useful when `speech_segment` is specified. If this argument is True, the segmentation process will ensure that all reference audios in the selected segment are not all-zero.
    * A new argument `flexible_numspk: bool = False` is added to `EnhPreprocessor` and `TSEPreprocessor` to handle the variable-speaker-number inputs in the `speech_ref1`, `dereverb_ref1`, and `enroll_ref1`. 
  * [espnet2/enh/espnet_model_tse.py](https://github.com/Emrys365/espnet/blob/espnet2_enh/espnet2/enh/espnet_model_tse.py)
    * A new argument `flexible_numspk` is added to handle training with variable numbers of speakers.

In addition, some updates are added for future extension (no test is available now):
  * [espnet2/tasks/abs_task.py](https://github.com/Emrys365/espnet/blob/espnet2_enh/espnet2/tasks/abs_task.py)
    * A new argument `--allow_multi_rates` is added to allow loading audios with different sampling rates. This is useful if we want to build a universal model that can handle different sampling rates.
  * [espnet2/enh/espnet_model.py](https://github.com/Emrys365/espnet/blob/espnet2_enh/espnet2/enh/espnet_model.py)
    * A new argument `flexible_numspk` is added to handle training with variable numbers of speakers.
    * However, since the currently implemented speech separation models do not support generate variable numbers of outputs, this function is not tested.
  * [espnet2/enh/extractor/abs_extractor.py](https://github.com/Emrys365/espnet/blob/espnet2_enh/espnet2/enh/extractor/abs_extractor.py)
      * A new input argument `additional` is added to the `AbsExtractor` forward method. This makes it consistent with the `AbsSeparator`, which is flexible for future extension.

## Why?

Speech separation with an unknown number of speakers is a popular research topic but has not been supported in ESPnet due to its variable data size.

## Note

Currently the inference and evaluation stages are not updated to support variable numbers of speakers. This part needs further discussion to make a good design.


espnet2/train/preprocessor.py
```diff
@@ -1056,6 +1080,31 @@ def _apply_to_all_signals(self, data_dict, func):
...
+    def _random_crop_range(self, data_dict, num_spk, uid=None, max_trials=10):
+        # Randomly crop the signals to the length `self.speech_segment`
+        assert self.speech_segment > 0, self.speech_segment
+        speech_refs = [
+            data_dict[self.speech_ref_name_prefix + str(spk + 1)]
+            for spk in range(num_spk)
+        ]
+        length = speech_refs[0].shape[0]
+        tgt_length = self.speech_segment
+        assert length >= self.speech_segment, (length, tgt_length)
```
Emrys365, line 1092: One thing worth discussing is whether we should allow the input signal to be shorter than the specified `speech_segment`.

sw005320:
@simpleoier, can you review this PR?

sw005320:
@LiChenda, can you also review this PR?

LiChenda:
> @LiChenda, can you also review this PR?

Sure, I'll review it.

espnet2/fileio/multi_sound_scp.py
```diff
@@ -0,0 +1,104 @@
...
+        key4 /some/path/d1.wav
+        ...
+
+        >>> reader = SoundScpReader('wav.scp', stack_axis=0)
+        >>> rate, stacked_arrays = reader['key1']
+        >>> assert stacked_arrays.shape[0] == 3
+
+        Note:
+            All audios in each sample must have the same sampling rates.
+            Audios of different lengths in each sample will be right-padded with np.nan
```
LiChenda, line 31: Why padded with `np.nan` here?

espnet2/tasks/enh.py
```diff
@@ -328,6 +328,28 @@ def add_task_arguments(cls, parser: argparse.ArgumentParser):
             help="The set of all possible categories in the dataset. Used to add the "
             "category information to each sample",
         )
+        group.add_argument(
+            "--speech_segment",
+            type=int_or_none,
+            default=None,
+            help="Truncate the audios to the specified length if not None",
```
LiChenda, line 335: Truncate the audio to the specified length (If I understand correctly, the length is in sampling points?)

espnet2/tasks/enh.py
```diff
@@ -328,6 +328,28 @@ def add_task_arguments(cls, parser: argparse.ArgumentParser):
...
+        )
+        group.add_argument(
+            "--avoid_allzero_segment",
```
LiChenda, line 338: Is it necessary to provide this option? I feel the default `True` is fine.

espnet2/train/dataset.py
```diff
@@ -64,10 +66,11 @@ def __getitem__(self, key: str) -> np.ndarray:
                     f"Unexpected type: {type(retval[0])}, {type(retval[1])}"
                 )
 
-            if self.rate is not None and self.rate != rate:
-                raise RuntimeError(
-                    f"Sampling rates are mismatched: {self.rate} != {rate}"
-                )
+            if not self.allow_multi_rates:
+                if self.rate is not None and self.rate != rate:
```
LiChenda, line 70: How about reducing two `if`  into one？
`if not self.allow_multi_rates and self.rate is not None and self.rate != rate:`

LiChenda (commented):
I left some comments on this PR. Do you have a plan to add some related recipes which have variable number of speakers?

sw005320 (approved):

espnet2/tasks/enh.py
Emrys365: Sometimes we may want to train the model to generate silence, and sometimes we may want to avoid the silence as the training target depending on the loss function we use. So I just make this argument to enable both possibilities.

espnet2/train/dataset.py
Emrys365: OK

espnet2/tasks/enh.py
Emrys365: Yes it's number of samples. Have updated the help-string.

Emrys365:
> I left some comments on this PR. Do you have a plan to add some related recipes which have variable number of speakers?

Yes, the recipe will be added in the followup PRs soon.

Emrys365:
Can we merge this PR?

LiChenda (approved):

sw005320:
Thanks a lot for your great efforts!

Issue #5415, opened 2023-08-22
Again: ValueError: assignment destination is read-only


mdecerbo:
I am also occasionally getting the "ValueError: assignment destination is read-only" crash mentioned in #2644 . 
For me it has most often occured when using compressed features (e.g., created with `copy-feats --compress=true`) but sometimes happens even with the features are not compressed. Most common place is within `time_warp()` on lines 35-38 here:

https://github.com/espnet/espnet/blob/66ccb857ae804a5970c5d0d57bb5dda4de645234/espnet/transform/spec_augment.py#L35-L39

When the problem occurs, the underlying numpy array is for some reason not marked as writeable. So checking whether it's writeable fixes the issue for me.

mdecerbo:
The PR was merged, thanks!

PR #5416, opened 2023-08-22
In spec_augment.py, check whether an array is writeable before modifying it inplace

mdecerbo:
This fixes occasional, unpredictable "ValueError: assignment destination is read-only" crashes for me. See #5415 .


sw005320:
@pyf98, do you think it is OK?

espnet/transform/spec_augment.py
pyf98, line 1: It looks good to me (although I have never encountered this issue before).

pyf98 (approved):

sw005320:
Thanks, @mdecerbo and @pyf98!

mdecerbo:
Thanks to you @pyf98 and @sw005320 for merging the patch, now I don't have to keep remembering to make that change each time I update my espnet checkout! 🙂

PR #5436, opened 2023-09-18
Add BibleTTS recipe

wyh2000:
## What?
BibleTTS training recipe and corresponding pre-trained models

## Why?
The goal is to build a TTS character based system using high-quality BibleTTS dataset. BibleTTS is a large high-quality open Text-to-Speech dataset with up to 80 hours of single speaker. It releases aligned speech and text for six languages spoken in Sub-Saharan Africa. There are two options: 
1) Train VITS models from scratch for each language. 

2) Finetune from a pretrained TTS model to accelarate training stage.

## See also
BibleTTS : https://masakhane-io.github.io/bibleTTS/


egs2/bibletts/tts1/conf/tuning/finetune_joint_conformer_fastspeech2_hifigan.yaml
```diff
@@ -0,0 +1,271 @@
+# This EXPERIMENTAL configuration is for ESPnet2 to finetune
```
ftshijt, line 1: We would like to minimize the config size as much as possible. For new recipes, please try to just upload the one with regard to your pre-trained models.

egs2/bibletts/tts1/run.sh
```diff
@@ -0,0 +1,45 @@
...
+
+train_set=tr_no_dev
+valid_set=dev
+test_sets="dev eval1"
+
+train_config=conf/train.yaml
+inference_config=conf/decode.yaml
+
+# g2p=g2p_en # Include word separator
+g2p=g2p_en_no_space # Include no word separator
```
ftshijt, line 28: Why do you use an English g2p for modeling? I thought this recipe was for multilingual purposes.

Your results seem to be with char? please align with your model with the setting~

egs2/bibletts/tts1/run.sh
```diff
@@ -0,0 +1,45 @@
...
+
+./tts.sh \
+    --lang en \
```
ftshijt, line 31: you may change the lang tag with recipes on different langs

egs2/bibletts/tts1/run.sh
wyh2000: I set g2p=none and token_type=char in the running command in "README.md" file. I will also modify them in the "run.sh" file

sw005320 (approved):

PR #5454, opened 2023-09-28
Do not use root logger in Beamsearch

### LINKED ISSUE #5281: [Feature request] want to choose log level for `best hypo` (transcribed texts)
Hello,

In ESPNet, transcribed texts sometimes appear in logging with root & INFO level, e.g., https://github.com/espnet/espnet/blob/3651c2e67126c4544820cf148407be7f2679866c/espnet2/bin/asr_inference.py#L479

I think this behavior is not favorable for commercial use. Many web services often log at the Info level, so the content of transcribed customer audio data, which should be kept confidential, is stored. Note that root level logging is difficult to control in a service environment.

A simple solution to this problem would be to stop using the root logger, i.e., to introduce 
`logger = logging.getLogger(__name__)`.

vsd-vector:
## What?

Do not use root logger in beam search classes and use get_logger(__name__) instead.

## Why?

This PR allows to configure logging level of beam search modules separately from other modules. For example, currently root logger gets spammed with transcription results, which might be undesirable. 

## See also

Potentially fixes #5281


sw005320:
Thanks a lot!

Issue #5453, opened 2023-09-28
ESPnet2 Contextual_block_conformer activation function for linear layers


espnetUser:
Hi @sw005320 and @eml914,

I noticed that the activation function for the linear PositionwiseFeedforward layers in the contextual block conformer encoder is set to default activation (ReLU) and not to Swish as is done in the normal (non-streaming) Conformer encoder (and the original paper):

https://github.com/espnet/espnet/blob/master/espnet2/asr/encoder/contextual_block_conformer_encoder.py#L146C1-L152C14

https://github.com/espnet/espnet/blob/master/espnet2/asr/encoder/conformer_encoder.py#L204C1-L211C14

The aishell streaming conformer config file does list Swish as the activation function for the encoder part:

https://github.com/espnet/espnet/blob/master/egs2/aishell/asr1/conf/train_asr_streaming_conformer.yaml#L14C27-L14C27

Could you please let me know if that was an oversight or there was a reason to go for ReLU instead of Swish (e.g., did not make a difference and was faster etc.). 

Thank you!

eml914:
Hi, thank you for pointing us out!
Indeed, the activation function is not set in the streaming version. I believe it is simple implementation error.
@D-Keqi can you check the codes please?

D-Keqi:
@espnetUser Thanks for picking it out, yes, this is a little error. We could correct it or would you be happy to make a PR for it?

espnetUser:
@D-Keqi: I will make a PR for this issue.

espnetUser:
@D-Keqi: Made the PR with small change to fix the activation function. Can you please review? Note, I did not have time to evaluate the effect of this change on streaming Conformer model training. Not expecting a lot to change but with these things one never knows until one actually runs it ;)

PR #5467, opened 2023-10-09
Fix to use proper activation function in contextual block conformer e…

espnetUser:
…ncoder

## What?

https://github.com/espnet/espnet/issues/5453: Added missing activation parameter to linear PositionwiseFeedforward layers arguments for contextual block Conformer encoder

## Why?

For Conformer encoder linear PositionwiseFeedforward layers typically use Swish rather than ReLU (default) activation function 

## See also

Original Conformer paper: https://arxiv.org/abs/2005.08100

### DIFF
--- a/espnet2/asr/encoder/contextual_block_conformer_encoder.py
+++ b/espnet2/asr/encoder/contextual_block_conformer_encoder.py
@@ -149,6 +149,7 @@ class ContextualBlockConformerEncoder(AbsEncoder):
                 output_size,
                 linear_units,
                 dropout_rate,
+                activation,
             )
         elif positionwise_layer_type == "conv1d":
             positionwise_layer = MultiLayeredConv1d


sw005320:
@espnetUser and @D-Keqi,
I'm not sure I can merge this until we could confirm the performance.
Is it ongoing?

espnetUser:
@D-Keqi: I don't have results yet and am currently tied up with other work. I can post results later when I got time to run another training w/wo this fix for the activation function.

D-Keqi:
> @D-Keqi: I don't have results yet and am currently tied up with other work. I can post results later when I got time to run another training w/wo this fix for the activation function.

OK, I expect I might be able to give it a go next week later due to the GPU service maintenance.

D-Keqi:
I gave it a go on Aishell-1 and found that ReLU actually works better than Swish for the linear PositionwiseFeedforward layer in my case. So we might want to leave the current version unchanged @sw005320. If @espnetUser has other results you'd like to share that's very welcome too.

PR #5481, opened 2023-10-17
Update perform_kmeans.sh

A-Quarter-Mile:
Hi, @ftshijt ~
Update the calculation of `portion_nutt`. 

If the portion is set to 1.0, to prevent the results from exceeding the total amounts. 
The condition is suggested to changed into follows:
{if (int($1 * $2)+1 < $1) {print(int($1 * $2)+1)} else {print($1)}}


ftshijt:
I feel it is reasonable, but maybe @simpleoier can double check it?

simpleoier:
@juice500ml is actually working on the same thing.

juice500ml:
Yeah, current master branch version fixed this issue!
https://github.com/espnet/espnet/blob/master/egs2/TEMPLATE/asr1/scripts/feats/perform_kmeans.sh#L112-L113

ftshijt:
Okay, then I will close the PR.

PR #5485, opened 2023-10-19
TTS evaluation script and monitoring functionality using MOS prediction model

Takaaki-Saeki:
## What?

This PR:
1. Adds an evaluation script for TTS using a MOS prediction model. It uses [SpeechMOS](https://github.com/tarepan/SpeechMOS) toolkit developed by @tarepan. It currently uses a pretrained [UTMOS](https://arxiv.org/abs/2204.02152) (Best performing method in [VoiceMOS Challenge 2022](https://voicemos-challenge-2022.github.io/)) strong learner model. Thanks to the toolkit, the predictor can be loaded directly from `torch.hub` without the need to install a specific module.
2. Adds a functionality to monitor the predicted MOS values during training of fully end-to-end TTS models.  Currently, it is supported for VITS and JETS.

## Why?

1. In addition to MCD and F0 RMSE, an evaluation script based on MOS prediction should be added. Evaluation by MOS prediction can be done even without ground truth speech.
2. For TTS (especially GAN-based TTS), it is often difficult to determine the best number of training steps. Monitoring the predicted MOS during training should help to improve the TTS performance.

## See also
- The figure shows the predicted MOS plot during the training of VITS on LJSpeech ,though it is still in early iterations. The predicted MOS value is increasing.
generator_predicted_mos.


egs2/TEMPLATE/asr1/pyscripts/utils/evaluate_pseudomos.py
```diff
@@ -0,0 +1,196 @@
...
+
+    if torch.cuda.is_available() and ("cuda" in args.device):
+        device = torch.device(args.device)
+    else:
+        device = torch.device("cpu")
+
+    # Load predictor for UTMOS22.
+    predictor = torch.hub.load(
+        "tarepan/SpeechMOS:v1.1.0", "utmos22_strong", trust_repo=True
+    ).to(device)
```
ftshijt, line 164: It is a very cool add-on. Do you also consider support other audo-MOS evaluation tools in the future?

egs2/TEMPLATE/asr1/pyscripts/utils/evaluate_pseudomos.py
```diff
@@ -0,0 +1,196 @@
+#!/usr/bin/env python3
```
ftshijt, line 1: You may consider update the README at `egs2/TEMPLATE/tts1`

egs2/TEMPLATE/asr1/pyscripts/utils/evaluate_pseudomos.py
Takaaki-Saeki: Currently I am not going to add it, but adding other MOS prediction tools sounds good.
So I have added a flag `mos_pred_tool` to switch the MOS prediction toolkits.

egs2/TEMPLATE/asr1/pyscripts/utils/evaluate_pseudomos.py
Takaaki-Saeki: Good point, thanks! Added the readme.

sw005320:
@Takaaki-Saeki, very cool!

- Can you add a test? You can check https://github.com/espnet/espnet/tree/master/test_utils
- I was thinking of suggesting to add this evaluation metric to the results of some recipes, but then I found that the TTS recipe usually do not provide the results (only pre-trained models). It does not even include the evaluation in template. We should do some updates…

sw005320:
After you add a test, I can merge this PR.

Takaaki-Saeki:
> * I was thinking of suggesting to add this evaluation metric to the results of some recipes, but then I found that the TTS recipe usually do not provide the results (only pre-trained models). It does not even include the evaluation in template. We should do some updates…

Thanks! This is an interesting direction and the objective results should be included in the TTS recipes.
For an objective evaluation of TTS recipes, it would be good to have a clearly defined test and dev sets, as in ASR (although this criterion needs to be discussed). For example, IIUC, LJSpeech and VCTK do not have predefined test sets, while LibriTTS has one as in Librispeech.
Along to this, TTS benchmarking (such as superb) with espnet2 would also be worthwhile.
Let me consider it as a future PR.

sw005320:
Cool!
We can discuss and design it!

ftshijt:
Thanks for the contribution!

PR #5494, opened 2023-10-21
Fix bugs of Whisper SOT.

pengchengguo:
Fix bugs in the Whisper SOT implementation.

1. In `whisper_export_vocabulary.py`, the type of `--sot_asr` should be `str2bool`;
2. In `class CommonPreprocessor_multi`, add `whisper_language` and `whisper_task` to support more datasets.


sw005320:
Thanks!

PR #5510, opened 2023-10-29
Voice conversion task in Espnet2

wyh2000:
## What?
We add voice conversion task in espnet2, and support both parallel data and non-parallel data.

## Why?
To support more voice conversion models and datasets

## See also
https://github.com/espnet/espnet/tree/master/egs/vcc20


egs2/TEMPLATE/vc1/README.md
```diff
@@ -0,0 +1,1048 @@
+# ESPnet2 VC Recipe TEMPLATE
```
ftshijt, line 1: reminder to change the tts content later

egs2/TEMPLATE/vc1/setup.sh
```diff
@@ -0,0 +1,64 @@
...
+# Copy
+for f in cmd.sh conf local; do
+    target="${dir}"/../../TEMPLATE/tts1/"${f}"
+    cp -r "${target}" "${dir}"
+    targets+="${dir}/${target} "
+done
+
+
+# Symlinks to TEMPLATE/tts1
+for f in tts.sh path.sh sid; do
```
ftshijt, line 44:
```suggestion
for f in vc.sh path.sh sid; do
```

egs2/TEMPLATE/vc1/sid/README.md
```diff
@@ -0,0 +1 @@
+The scripts in this directory were copied from Kaldi-ASR, https://github.com/kaldi-asr/kaldi, and their licenses follow the original license, https://github.com/kaldi-asr/kaldi/blob/master/COPYING, of Kaldi.
```
ftshijt, line 1: you may use symlink for these. I'm also not sure if we would want to keep this as default, as Kaldi's version is quite old (where I think it might be better to keep it as an option instead of the default one). Also due to the reason that our ESPnet version would be on quite soon.

egs2/TEMPLATE/vc1/sid/compute_vad_decision.sh
```diff
@@ -0,0 +1,86 @@
+#!/bin/bash
```
ftshijt, line 1: Ditto, please use symlink here

ftshijt (commented):
For data preparation, what is the typical data format that are different than the base-kaldi scps?

PR #5512, opened 2023-10-30
Add Libriheavy small and medium ASR2 recipes

akreal:
## What?

ASR2 recipe for Libriheavy medium subset.

## Why?

I'm running experiments with a causal LM and ASR2 on this dataset and would like to have a comparison point without a causal LM.

## See also

https://github.com/k2-fsa/libriheavy

I'll add results and a model link in about a week.


akreal:
I also added the recipe for the small subset. It should be ready for review now.

The official baseline seems to do case sensitive scoring for the non-normalized text, so I added `-s` to `score_opts` in these recipes.

sclite skips all lines starting with `**` and prints a warning for all lines starting with `*` (it's hardcoded as a comment character). I've added the removal of `*` in beginnings of texts in the scoring stage of `asr2.sh`. I've manually verified that it works as intended on the dev set, it should have no effect on other recipes.

sw005320 (approved):
Thanks, @akreal!

PR #5514, opened 2023-10-30
Add causal LM for ASR2

akreal:
This PR also includes few things which are not directly related to the causal LM, but are needed for the recipe:

- add an option to disable bias in the CTC output layer;
- add CTC-only training for ASR2 (I use this and the previous one to load a pretrained LM head into the CTC output layer and prealign encoder output with the pretraind LM embeddings);
- add `--pretrained_model` option to the ASR2 recipe to load the prealigned encoder parameters;
- allow dimensionality projection in the length adaptor to happen after the subsampling instead of before the subsampling (this needs less parameters and seems to work better).

To do:

- [x] add tests
- [x] add recipe

### DIFF
--- a/egs2/libriheavy_small/asr2/local/run_hf.sh
+++ b/egs2/libriheavy_small/asr2/local/run_hf.sh
@@ -31,10 +31,17 @@ tgt_case="ts"
 
 embed_tokens_path=data/mistral7b-instruct02_embed_tokens.pth
 
+# The training is performed in two stages:
+# Stage 1: pretrain the encoder for prediction of LM token embeddings,
+# this is done with CTC loss and frozen LM token embedding matrix;
+# Stage 2: fine-tune the encoder from Stage 1 as a part of AED model where the decoder is a frozen causal LM.
+# Theoretically, training only with the Stage 2 is possible, but in practice it fails to converge.
+
 local/export_hf_embed_tokens.sh \
     mistralai/Mistral-7B-Instruct-v0.2 \
     ${embed_tokens_path}
 
+# Stage 1: perform all recipe steps up to the training and train with CTC loss.
 ./asr2.sh \
     --kmeans_feature "${kmeans_feature}" \
     --kmeans_opts "--batch_bins 4800000 --nj 4" \
@@ -63,6 +70,8 @@ local/export_hf_embed_tokens.sh \
     --lm_train_text "data/${train_set}/text.${tgt_case}.${tgt_lang}" "$@" \
     --stop_stage 13
 
+# Stage 2: skip all recipe steps up to the training and train with AED (CE) loss
+# using the pretrained parameters from the Stage 1.
 ./asr2.sh \
     --kmeans_feature "${kmeans_feature}" \
     --kmeans_opts "--batch_bins 4800000 --nj 4" \


akreal:
@sw005320 this should be ready for a review (the CI failures are not related to my changes).

The recipe is based on LibriHeavy Small and Mistral-7B-Instruct-v0.2. The results actually show a regression in WER (15.4%&rarr;20.8% for test_other), but a different configuration might demonstrate LM benefits:
 - use the Mixtral-8x7B-v0.1 LM;
 - use the LibriHeavy Medium data;
 - fine-tune or jointly train on speech translation data (e.g. CoVoST2) or Dynamic-SUPERB.

Unfortunately, I don't have access to enough compute to try these configurations.

sw005320:
Thanks a lot, @akreal!
@pengchengguo and @simpleoier, can you review this PR?

egs2/libriheavy_small/asr2/local/run_hf.sh
```diff
@@ -0,0 +1,96 @@
...
+src_case="rm"
+tgt_case="ts"
+
+embed_tokens_path=data/mistral7b-instruct02_embed_tokens.pth
+
+local/export_hf_embed_tokens.sh \
+    mistralai/Mistral-7B-Instruct-v0.2 \
+    ${embed_tokens_path}
+
+./asr2.sh \
```
simpleoier, line 38: Can you add some comments why running asr2.sh twice?

espnet2/mt/espnet_model.py
```diff
@@ -63,6 +64,7 @@ def __init__(
         self.src_eos = src_vocab_size - 1 if src_vocab_size else None
         self.vocab_size = vocab_size
         self.src_vocab_size = src_vocab_size
+        self.src_ignore_id = src_ignore_id
```
simpleoier, line 67: I'm afraid that this makes the previous models incompatible.

simpleoier (commented):
Thanks for adding this new support! I left some comments. Please let me know if there is any misunderstanding.

egs2/libriheavy_small/asr2/local/run_hf.sh
akreal: Done! Please let me know if I should clarify it more.

espnet2/mt/espnet_model.py
akreal: I guess it would, if `src_ignore_id` would register as a torch model parameter and it would be missing from `state_dict` in the previous models. However, it appears to be only a class attribute and `torch.load` does not pay attention to it.

I tested the baseline model from this recipe (`espnet/akreal_lh_small_asr2_e_branchformer_wavlm_large_21_km2k_bpe_rm6k_bpe_ts3k_sp`) on a master code with these two extra lines in `espnet2/mt/espnet_model.py`, and it worked:
```
src_ignore_id: int = -1,
...
self.src_ignore_id = src_ignore_id
```

Please let me know if you meant some other type of incompatibility.

akreal:
> Thanks for adding this new support! I left some comments. Please let me know if there is any misunderstanding.

Thank you for your review! I addressed your comments. Please let me know if there are further questions or suggestions.

PR #5516, opened 2023-10-30
replace old mustc links with new instructions

brianyan918:
The mustc download links were changed. I'm updating the scripts accordingly.


sw005320 (approved):

eilson-he:
Hi, seems like the new links are broken as well (https://mt.fbk.eu/must-c-release-v1-0/). Do you know if there is any other workaround?

PR #5517, opened 2023-10-30
Add E-Branchformer model checkpoint in OWSM v2

pyf98:
This PR adds the model link to E-Branchformer in OWSM v2.
Despite being trained for fewer updates, E-Branchformer shows clear improvement over Transformer at such a large scale.
We will train and release E-Branchformer based models at even larger scales.

image


sw005320:
Thanks!

PR #5105, opened 2023-04-13
[WIP] add Bayes Risk CTC to Espnet2

jctian98:
This PR adds Bayes Risk CTC criterion to Espnet2 ASR recipe.

Paper link: https://openreview.net/forum?id=Bd7GueaTxUz&referrer=%5BAuthor%20Console%5D(%2Fgroup%3Fid%3DICLR.cc%2F2023%2FConference%2FAuthors%23your-submissions)

Work in progress:
1) validation experiments for offline and online cases;
2) add test code
3) make it compatible with more more ASR encoders besides conformer.

PR #5519, opened 2023-11-01
Add Bayes Risk CTC (reworked)

jctian98:
Add Bayes Risk CTC (BRCTC) criterion and a corresponding recipe
Previous [PR](https://github.com/espnet/espnet/pull/5105) is deprecated 

BRCTC can control the alignment prediction of CTC. In the recipe, it helps to down-sample the encoder output to achieve faster decoding. Details are in: https://openreview.net/forum?id=Bd7GueaTxUz

Dependency: Implementation relies on k2: https://github.com/k2-fsa/k2

Experiments on Aishell, work with Conformer, Hybrid CTC/Attention system, dev / test CER%
Baseline: 4.2 / 4.7
BRCTC: 4.4 / 4.7
Average number of frames for encoder output (on test set):
Baseline: 156
BRCTC: 45

BRCTC is found more compatible with `legacy rel_pos_type`. If this is set to `latest`, slight degradation might be observed.
On Aishell, dev/tests CER%:
Baseline: 4.2 / 4.5
BRCTC: 4.4 / 4.8


jctian98:
I think this PR is ready for review. 
(Only Codecov fails)

sw005320:
Cool!

espnet2/asr/bayes_risk_ctc.py
pyf98, line 1: This file does not have a CI test?

pyf98:
Looks good to me.

espnet2/asr/bayes_risk_ctc.py
sw005320: > This file does not have a CI test?

I agree.
@jctian98, please add some tests.
Then, I'll review it again.

jctian98:
Will add ci

jctian98:
CI test case is added and the main body of `bayes_risk_ctc.py` is covered.
Can you restart the CI test online?
(1) one test fails due to time out problem
(2) codecov doesn't pass (do I still need to do something to pass this?

espnet/nets/pytorch_backend/nets_utils.py
sw005320, line 1: Can you increase the test coverage of `trim_by_ctc_posterior()` if possible?

espnet2/asr/bayes_risk_ctc.py
sw005320: Not sure, but it looks like this module does not seem to be tested (or might be an issue in codecov).
Can you double check it?

jctian98:
Not sure why the codecov fails again - even the init function of the object is not covered.
Another failure could be occasional

maybe ready for review?

sw005320:
At least, based on https://github.com/espnet/espnet/blob/eb3efd0cda81f8b75e9aaf054b12668c117255e8/espnet2/asr/ctc.py#L191-L219, the test does not go through the loss function computation.
Therefore, your modifications in `test/espnet2/asr/test_ctc.py` (or even `gtnctc`) do not provide a meaningful test. So, please remove them and add the test to cover the loss computation test.


(I'm not sure about the following code)
https://github.com/espnet/espnet/blob/eb3efd0cda81f8b75e9aaf054b12668c117255e8/test/espnet2/asr/test_ctc.py#L17-L20

sw005320:
> I think the codecov doesn't test `bayes_risk_ctc.py` mainly because of this:
> 
> ```
> if self.ctc_type == "brctc" and not torch.cuda.is_available():
>         self.ctc_type = "builtin"
> ```
> 
> So the test case may need GPU. (seems CUDA is needed if the k2 is built with GPU)

I see.
Hmm, `brctc` also only supports GPU...

jctian98:
When GPU is not available I get this error when importing GPU version of `k2`
`ImportError: libcuda.so.1: cannot open shared object file: No such file or directory`
Maybe I can fix this locally by adding some `PATH`, I'll take a look.

But I'm not sure if the online CI test environment provides the CUDA .so files (even though there is no GPU available).
If it has, I think I can write a test case to confirm BRCTC has the identical output like builtin CTC (if no risk is applied)
If it doesn't, I think it cannot be tested without GPU (is there any case that CI tests with GPU?)

b-flo:
> But I'm not sure if the online CI test environment provides the CUDA .so files (even though there is no GPU available).
> If it has, I think I can write a test case to confirm BRCTC has the identical output like builtin CTC (if no risk is applied)
> If it doesn't, I think it cannot be tested without GPU (is there any case that CI tests with GPU?)

Until recently github-hosted runners didn't support GPU but there are [future plans](https://resources.github.com/devops/accelerate-your-cicd-with-arm-and-gpu-runners-in-github-actions/) for that. Until then, we would need to rely on a self-hosted runners.

AFAIK, the current practice in ESPnet is to skip such tests when no GPU is available (or rely on `importorskip` for GPU-dependant lib, such as k2).

test/espnet2/asr/test_ctc.py
```diff
@@ -14,25 +14,25 @@ def ctc_args():
     return h, h_lens, y, y_lens
 
 
-@pytest.mark.parametrize("ctc_type", ["builtin", "gtnctc"])
+@pytest.mark.parametrize("ctc_type", ["builtin", "gtnctc", "brctc"])
```
sw005320, line 17:
```suggestion
@pytest.mark.parametrize("ctc_type", ["builtin"])
```

test/espnet2/asr/test_ctc.py
```diff
@@ -14,25 +14,25 @@ def ctc_args():
...
 def test_ctc_forward_backward(ctc_type, ctc_args):
     ctc = CTC(encoder_output_size=10, odim=5, ctc_type=ctc_type)
     ctc(*ctc_args).sum().backward()
 
 
-@pytest.mark.parametrize("ctc_type", ["builtin", "gtnctc"])
+@pytest.mark.parametrize("ctc_type", ["builtin", "gtnctc", "brctc"])
```
sw005320, line 23:
```suggestion
@pytest.mark.parametrize("ctc_type", ["builtin"])
```

test/espnet2/asr/test_ctc.py
```diff
@@ -14,25 +14,25 @@ def ctc_args():
...
 def test_ctc_softmax(ctc_type, ctc_args):
     ctc = CTC(encoder_output_size=10, odim=5, ctc_type=ctc_type)
     ctc.softmax(ctc_args[0])
 
 
-@pytest.mark.parametrize("ctc_type", ["builtin", "gtnctc"])
+@pytest.mark.parametrize("ctc_type", ["builtin", "gtnctc", "brctc"])
```
sw005320, line 29:
```suggestion
@pytest.mark.parametrize("ctc_type", ["builtin"])
```

test/espnet2/asr/test_ctc.py
```diff
@@ -14,25 +14,25 @@ def ctc_args():
...
 def test_ctc_log_softmax(ctc_type, ctc_args):
     ctc = CTC(encoder_output_size=10, odim=5, ctc_type=ctc_type)
     ctc.log_softmax(ctc_args[0])
 
 
-@pytest.mark.parametrize("ctc_type", ["builtin", "gtnctc"])
+@pytest.mark.parametrize("ctc_type", ["builtin", "gtnctc", "brctc"])
```
sw005320, line 35:
```suggestion
@pytest.mark.parametrize("ctc_type", ["builtin"])
```

sw005320 (commented):
Suggestions based on https://github.com/espnet/espnet/pull/5519#issuecomment-1805646401

jctian98:
I have additional findings:
To run the test case with K2, GPU is NOT needed. However, for GPU version of K2, CUDA dependency (e.g., `libcuda.so.1`) is needed for import / module initialization processing.
So we can complete the test as long as the CUDA is installed and detected, even without GPU hardware available.

I'll write a new test for BRCTC. It will pass as long as CUDA files are available; it will be skipped if not. GPU will not be needed in that test.
However, if Github runners don't have the CUDA files, the test is still not effective.

I'll remove the current test cases for `gtnctc` and `brctc`

jctian98:
I've updated a new test case for BRCTC. But seems the `bayes_risk_ctc.py` is still not tested due to the reason above.

sw005320:
Thanks a lot, @jctian98!

PR #5521, opened 2023-11-01
Update attention.py

aky15:
fix rwkv attention calculation bugs. @b-flo 

Apply minus and exp operation to the time_decay factor, as https://github.com/huggingface/transformers/blob/ae093eef016533a3670561fa9e26addb42d446d1/src/transformers/models/rwkv/modeling_rwkv.py#L219C5-L219C40


b-flo:
Hi @aky15 ,

It should be intended but it may have be an error on my part as I tested a lot of thing with RWKV. Do you have some comparison between versions?

I don't have my notes at hand so I can't further comment but I recall this part was brought up when discussing paper formula and numerically stable implementation(s) of WKV computation.

aky15:
Hi, @b-flo 
I found that without minus and exp operations, the output of the RWKV SelfAttention at training does not agree with the output at inference. To be more specific, wkv calculated using https://github.com/espnet/espnet/blob/e7377efb1a4faac66b75db6aaaaadbadae0bc82e/espnet2/asr_transducer/decoder/modules/rwkv/attention.py#L352 is not equal to wkv calculated using https://github.com/espnet/espnet/blob/e7377efb1a4faac66b75db6aaaaadbadae0bc82e/espnet2/asr_transducer/decoder/modules/rwkv/attention.py#L364. When I add the minus and exp operations, the outputs become identical.

b-flo:
Hi @aky15 ,

Could you merge master branch changes please? After CI re-runs, I'll merge the PR, thanks!

b-flo:
Merging, thanks @aky15 (and @sw005320 for merging master). If you conduct some experiments with RWKV, feel free to report them! I'm interested!

aky15:
I conducted some experiments with RWKV as streamable RNN-T encoders @b-flo . https://arxiv.org/pdf/2309.14758.pdf FYI. I'll try RWKV as the transducer decoder later.

b-flo:
Thanks for the share, I'll throughly read this!
For the decoder part, imo there are no real benefit to use more complex architecture (same as MEGA, Transformer, etc) without rethinking its role and/or incorporating additional mechanism to tackle specific limitations.

aky15:
According to the results of k2 stateless trasducer, a trivial decoder such as a simple CNN works as well.
BTW, do you have any plan to support CNN as transducer decoder in Espnet? @b-flo

b-flo:
> According to the results of k2 stateless trasducer, a trivial decoder such as a simple CNN works as well.

I concur. For vanilla Transducer, the stateless decoder (w/ or w/o convolution augmentation) works well.

>  BTW, do you have any plan to support CNN as transducer decoder in Espnet? @b-flo

Hum, it should be an option to current [StatelessDecoder](https://github.com/espnet/espnet/blob/master/espnet2/asr_transducer/decoder/stateless_decoder.py) but it doesn't seem so. Dunno if I removed it intentionally or forgot some commits in related PR.

Anyway, I can add support. Bear with me for a while though, I have some other priorities right now.

PR #5523, opened 2023-11-01
Improve error robustness of unit tests

Emrys365:
## What?

This PR improves two unit tests to avoid occasional errors caused by numerical precision or randomness.

  * [test/espnet2/asr/frontend/test_frontend.py](https://github.com/espnet/espnet/blob/master/test/espnet2/asr/frontend/test_frontend.py): use a predefined 2-ch signal instead of a randomly sampled signal for beamforming to avoid singular covariance matrix.
  * [test/espnet2/enh/layers/test_complex_utils.py](https://github.com/espnet/espnet/blob/master/test/espnet2/enh/layers/test_complex_utils.py): fix the random seed to avoid occasional error when the numerical difference is slightly larger than 1e-06.

## Why?

Sometimes we may observe errors as in https://github.com/espnet/espnet/actions/runs/6719461822/job/18261148245 which are essentially numerical issues. After the modifications in this PR, these errors should be avoided in most cases.

### DIFF
--- a/test/espnet2/asr/frontend/test_frontend.py
+++ b/test/espnet2/asr/frontend/test_frontend.py
@@ -2,7 +2,48 @@ import pytest
 import torch
 
 from espnet2.asr.frontend.default import DefaultFrontend
-from espnet2.torch_utils.set_all_random_seed import set_all_random_seed
+
+
+random_speech = torch.tensor(
+    [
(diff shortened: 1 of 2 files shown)


sw005320:
Thanks, @Emrys365!

Issue #5528, opened 2023-11-03
pyopenjtalk issue


sw005320:
There is a windows installation issue in pyopenjtalk https://github.com/espnet/espnet/actions/runs/6736679956/job/18354953375
@r9y9, could you take a look at this?

Runtrons:
Having the same issue on Mac too.

r9y9:
Thanks for the report. Tthe new release v0.3.3 (https://github.com/r9y9/pyopenjtalk/releases/tag/v0.3.3) should fix this issue.

sw005320:
Thanks a lot, @r9y9!

r9y9:
Shall we close this?

sw005320:
Sorry. I'm closing the issue.

PR #5531, opened 2023-11-05
Update install_pyopenjtalk.sh

sw005320:
## What?

Update pyopenjtalk

## Why?

The previous version was failed for the windows installation.

## See also

https://github.com/espnet/espnet/issues/5528


sw005320:
@kan-bayashi, do we need to update install_tdmelodic_pyopenjtalk.sh? or can we leave it as it is?

PR #5532, opened 2023-11-06
Fix sentencepiece version for Python 3.11

fujimotos:
## What?

This patch fixes the version specifier for `sentencepiece` in setup.py.

## Why?

SentencePiece adds support for Python 3.11 in v0.1.98.

> https://github.com/google/sentencepiece/releases/tag/v0.1.98

Since ESPnet pins sentencepiece to v0.1.97, users with recent Python version can
see failure during installation. 

Note: We received such a bug report today from a user who uses Python 3.11.5,
so proceeded to submit this patch.

## See also

* @kamo-naoyuki pinned the sentence piece version to v0.1.97 in #5107.


kamo-naoyuki:
Sentencepiece v0.1.9.8 seems to break backward compatibility. 
I don't know the details of the update. Please ask someone in the know.

PR #5533, opened 2023-11-06
Avoid problem during averaging when parameter-tying exists

tzyll:
## Why?
When parameter-tying exists in a pytorch network, different keys share the same memory which will be processed multiple times if not separated using clone().


sw005320:
Will it break the compatibility, or can we still use the old model checkpoint?

tzyll:
Hi, the modification won't affect the old model checkpoint.

> Will it break the compatibility, or can we still use the old model checkpoint?

sw005320:
Thanks!
One more request.
Could you show me that the old one and this do not change the results (model parameters themselves and ASR results)?

PR #5535, opened 2023-11-07
Improve error robustness of unit tests

Emrys365:
## What?

This PR is a followup PR of https://github.com/espnet/espnet/pull/5523 to improve two unit tests to avoid occasional errors caused by numerical precision or randomness.

  * [test/espnet2/asr/frontend/test_frontend.py](https://github.com/espnet/espnet/blob/master/test/espnet2/asr/frontend/test_frontend.py): use a predefined 2-ch signal instead of a randomly sampled signal for beamforming to avoid singular covariance matrix.

## Why?

Sometimes we may observe errors as in https://github.com/espnet/espnet/actions/runs/6719461822/job/18261148245 which are essentially numerical issues. After the modifications in this PR, these errors should be avoided in most cases.

### DIFF
--- a/espnet/nets/pytorch_backend/frontends/dnn_wpe.py
+++ b/espnet/nets/pytorch_backend/frontends/dnn_wpe.py
@@ -1,11 +1,11 @@
 from typing import Tuple
 
 import torch
-from pytorch_wpe import wpe_one_iteration
 from torch_complex.tensor import ComplexTensor
 
 from espnet.nets.pytorch_backend.frontends.mask_estimator import MaskEstimator
 from espnet.nets.pytorch_backend.nets_utils import make_pad_mask
+from espnet2.enh.layers.wpe import wpe_one_iteration
 
 
 class DNN_WPE(torch.nn.Module):
--- a/espnet/nets/pytorch_backend/frontends/dnn_wpe.py
+++ b/espnet/nets/pytorch_backend/frontends/dnn_wpe.py
@@ -3,9 +3,9 @@ from typing import Tuple
 import torch
 from torch_complex.tensor import ComplexTensor
 
+from espnet2.enh.layers.wpe import wpe_one_iteration
 from espnet.nets.pytorch_backend.frontends.mask_estimator import MaskEstimator
 from espnet.nets.pytorch_backend.nets_utils import make_pad_mask
-from espnet2.enh.layers.wpe import wpe_one_iteration
 
 
 class DNN_WPE(torch.nn.Module):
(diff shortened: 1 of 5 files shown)


sw005320:
Thanks, but we still have an issue...

Emrys365:
This is weird. I cannot reproduce the error with the same PyTorch version...

sw005320:
Is this operation actually used?

Emrys365:
Are you talking about the `DefaultFrontend` with WPE and beamforming? I think we usually use the S2T function (SE+ASR) instead of ASR only, where the new `NeuralBeamformer` (espnet2/enh/separator/neural_beamformer.py) is used.

sw005320:
I mean the matrix inverse function, not wpe or beamformer.
I remember that there are several implementations for the complex matrix inverse operation.

Emrys365:
Because it is using the `pytorch_wpe` package for WPE (where the CI error occurs), the matrix inverse operation is conducted based on `torch_complex`. Both packages are not maintained in our codebase.

sw005320:
Got it.
Is there a way to avoid the issue of going through torch_complex by copying them inside espnet and using the other matrix inverse in the future?

I want to remove such dependencies in the future.

Emrys365:
I think so. It is what I did before, and probably we can just replace
```python
from espnet.nets.pytorch_backend.frontends.dnn_beamformer import DNN_Beamformer
from espnet.nets.pytorch_backend.frontends.dnn_wpe import DNN_WPE
```
with the new implementations in espnet2
```python
from espnet2.enh.layers.dnn_beamformer import DNN_Beamformer
from espnet2.enh.layers.dnn_wpe import DNN_WPE
```

sw005320:
Sounds good!
With this, can we remove these dependencies?
If so, we may consider it in the other PRs. This will simplify things.

Emrys365:
I think at least we can remove `pytorch_wpe`. It was only used in `espnet/nets/pytorch_backend/frontends/dnn_wpe.py`. I have verified locally that replacing espnet1 implementations with espnet2 ones can pass the tests.

sw005320:
Nice!
This is a good part of having a unit test!
We can safely replace it :)

Emrys365:
`torch_complex` is still now used in Enh since some complex-valued functions have not been implemented yet in torch.

sw005320:
OK, it passes the test!
Can you also remove `pytorch_wpe,` then?

Emrys365:
Sure.

sw005320:
Can you remove it at https://github.com/espnet/espnet/blob/master/setup.py#L42?

Emrys365:
Yes, I already removed it in the latest commit.

sw005320:
Thanks a lot!
This was a difficult bug, but it brought an outcome to remove the old dependency!
Very good!

PR #5539, opened 2023-11-08
Add instructions for finetuning owsm

pyf98:
## What?

This PR adds a step-by-step guide on how to fine-tune existing OWSM models on a specific dataset.


sw005320:
Thanks, @pyf98!
I see how it is complicated.
@juice500ml and @simpleoier, it is a good example of making it work with a simple espnet

PR #5540, opened 2023-11-09
fix kaldi dependency in asr2

ftshijt:
## Why?

Since we changed to our own utils and steps, we do not support default Kaldi installation anymore, so better to change asr2 symlink to utils/steps of our own.


simpleoier (approved):
LGTM! Thanks!

PR #5548, opened 2023-11-15
Add PyTorch & GPU support for DNSMOS calculation

Emrys365:
## What?

This PR adds support for PyTorch-based DNSMOS calculation in addition to the existing ONNX-based one. This is based on a third-party package [`onnx2torch`](https://github.com/ENOT-AutoDL/onnx2torch) which converts the ONNX-format model to a PyTorch module.

## Why?

Current DNSMOS calculation basically relies on the ONNX wrapper, where the GPU-based processing is not well supported. By converting the ONNX-format model to a PyTorch module using `onnx2torch`, we can now easily use CUDA devices to greatly speed up the calculation.

In addition, since the model is converted to a PyTorch module, gradient backpropagation is also supported, which can be used to provide supervision during training later on.


sw005320:
It requires additional packages for the test https://github.com/espnet/espnet/actions/runs/6872983326/job/18692312021?pr=5548#step:9:2429

sw005320:
How about only adding it in https://github.com/espnet/espnet/blob/master/ci/install.sh?

Emrys365:
OK. I'll try it.

sw005320:
Thanks a lot!

PR #5567, opened 2023-11-27
Fix except

takenori-y:
## What?
Use a correct one-line except. 

## Why?
```
from espnet2.tasks.tts import TTSTask
```
causes a segmentation fault in my environment (Python 3.9.0)

## See also
- https://docs.python.org/3/tutorial/errors.html#handling-exceptions
- https://stackoverflow.com/questions/6470428/catch-multiple-exceptions-in-one-line-except-block


sw005320:
Thanks a lot, @takenori-y!

Issue #5580, opened 2023-12-05
minlenratio not effective


jctian98:
**Describe the bug**
Please let me know if I'm wrong: the `minlenratio` argument beam search algo is not working in 
https://github.com/espnet/espnet/blob/master/espnet/nets/beam_search.py and the inherited batch_beam_search object

To fix it, the `minlen` or `minlenratio` should be considered in `post_process` when adding an ended hyp into `ended_hyps` list.

When working on an attention-only model, it occasionally predicts `<eos>` in the first place, which makes the best_hypo an empty string. `minlenratio` is a heuristic way to avoid this situation, but it doesn't work based on my observation. After fixing it locally, the issue is solved. 

I think the issue was originally found in https://github.com/espnet/espnet/issues/2953 but has not been solved yet. I can open a PR to fix it if we confirm this is an error.

Issue #5573, opened 2023-11-29
Increasing the beam size can sometimes lead to the elimination of inference in ASR on custom dataset.


hagakure333:
I'm training a speech recognition model on a custom dataset. When I increase the beam size, inference disappearance sometimes occurs. What could be potential reasons for this? Thank you.

beam size =50， wer=23.0
![消失](https://github.com/espnet/espnet/assets/67449891/fe7763d0-4ff0-43ce-a7ee-709f0b554274)

beam size =20，wer=22.1
![消失なし](https://github.com/espnet/espnet/assets/67449891/0d08a72f-628d-4834-9df8-5bc9313d0308)

inference config
`beam_size: 20
lm_weight: 0.0
minlenratio: 0.05
penalty: 0
ctc_weight: 0.0`

asr config
`batch_type: folded
batch_size: 64
accum_grad: 2
max_epoch: 70
patience: 5

init: xavier_uniform
best_model_criterion:
-   - valid
    - acc
    - max
keep_nbest_models: 10

encoder: transformer
encoder_conf:
    output_size: 256
    attention_heads: 4
    linear_units: 2048
    num_blocks: 12
    dropout_rate: 0.1
    positional_dropout_rate: 0.1
    attention_dropout_rate: 0.1
    input_layer: conv2d
    normalize_before: true

decoder: transformer
decoder_conf:
    attention_heads: 4
    linear_units: 2048
    num_blocks: 6
    dropout_rate: 0.1
    positional_dropout_rate: 0.1
    self_attention_dropout_rate: 0.1
    src_attention_dropout_rate: 0.1

model_conf:
    ctc_weight: 0.0
    lsm_weight: 0.1
    length_normalized_loss: false

optim: adam
optim_conf:
    lr: 0.0002
scheduler: warmuplr
scheduler_conf:
    warmup_steps: 20000
specaug: specaug
specaug_conf:
    apply_time_warp: true
    time_warp_window: 2
    time_warp_mode: bicubic
    apply_freq_mask: true
    freq_mask_width_range:
    - 0
    - 15
    num_freq_mask: 2
    apply_time_mask: true
    time_mask_width_range:
    - 0
    - 20
    num_time_mask: 2
`

espnet version ==0.10.0

sw005320:
Can you increase `minlenratio`?
This may avoid having too short sentences.

hagakure333:
[...]

PR #5581, opened 2023-12-06
Make minlenratio effective

jctian98:
## What?
This is to respond to issues https://github.com/espnet/espnet/issues/5573 and https://github.com/espnet/espnet/issues/5580 to make the decoding hyper-parameter `minlenratio` effective in beam search

Previously the `minlenratio` and the variable `minlen` were calculated but not applied, which leads to the issue as described below:
Some models intend to predict the special symbol `<eos>` at the first several decoding steps. As long as the invalid `<eos>` remains in the top-k list, these invalid hypotheses are added to the ended hypothesis list. Since the invalid hypotheses are short, the accumulated posterior is also very high compared to other valid hypotheses with much longer text hypotheses. The invalid hypotheses are then selected in the finalization stage, most of which are empty strings. As a consequence, the overall decoding results would be empty or very short. The issue is more likely to be observed when beam size is large as the `<eos>` token is more likely to stay in the top-k list.

To overcome this issue, simply make the `minlenratio` hyperparameter effective by adding an `if` condition in the `post_process` function. Then pass a small heuristic value to it (e.g., 0.1). This will force all ended hypotheses to have a length of at least 10% of the input lengths.  Although it's heuristic, it works well generally.

This PR also adds a length normalization feature for the beam search object. If activated, the best hypothesis is selected by the best token-level score rather than the accumulated score.


sw005320:
LGTM!
Once you fix the CI error, I'll merge it.

espnet/nets/batch_beam_search.py
sw005320, line 1: Why it does not include the normalization?

sw005320:
@siddhu001, can you check the case of negative `minlenratio`?
Would it be an expected behavior?

espnet/nets/batch_beam_search.py
jctian98: I think the `batch_beam_search` object inherits the sort function from the `beam_search` object so. So we don't need to handle this in the `batch_beam_search` object.

sw005320:
After @siddhu001's confirmation, I'll merge this PR.

jctian98:
> @siddhu001, can you check the case of negative `minlenratio`? Would it be an expected behavior?

It's good to test with the attention-only model, as the ctc can greatly calibrate the length estimation so minlenratio will not be tested properly. 
Please let me know if there is any issue. Thanks @siddhu001

siddhu001:
> @siddhu001, can you check the case of negative `minlenratio`? Would it be an expected behavior?

Hi @sw005320,

Yes I believe the case of negative minlenratio is the expected behaviour. We follow a similar approach for negative maxlenratio (i.e. maxlenratio=-1*maxlenratio when maxlenratio<0)

@jctian98 thanks a lot for your efforts on this PR. This is very useful!

PR #5372, opened 2023-07-25
Add espnetez

Masao-Someki:
## What?
This PR adds `espnetez` package to make it easier to use ESPnet!

## Why?
ESPnet runs primarily with a shell script, which can be difficult for beginners to use for running all the stages. The `espnetez` tool provides a Pythonic frontend for users, making it more user-friendly.

PR #5584, opened 2023-12-10
Support external dataset library for ESPnetEasy

Masao-Someki:
## What?
This PR is to support several external dataset libraries or formats and create instruction documents.
Currently, I'm planning to support the following datasets:
- [x] torchdata (torchaudio.dataset)
- [ ] Lhotse
- [x] huggingface dataset (optional)
- [ ] WebDataset (optional)

## Why?
While progressing with the espnetez project, we found that it is possible to support external dataset formats.
Espnetez has the potential to develop finetuning Python code more easily, but we cannot take full advantage if we cannot define our custom dataset.
Therefore, the support of typical dataset libraries in this PR contributes to improving the usability of espnetez.

## See also
https://github.com/espnet/espnet/pull/5372


Masao-Someki:
Currently, several `iter_factory` related functions are not supported and I'm working on this part.

Vaibhavs10:
Hey @Masao-Someki @sw005320 - Lovely that you are adding support for datasets. Do let me know if you hit any roadblocks and I'd be more than happy to help you! 🤗 
Very exciting!!

Masao-Someki:
@sw005320 @wanchichen  @juice500ml
Hi, I've incorporated the template directory into the egsez and adjusted the recipe directory to align with the structure used in other recipes, like egs2.

Could you please review the recipe folder, especially for the directory structure? Any comments you may have would be greatly appreciated!

sw005320:
@Masao-Someki or others,
Can you fix the CI errors?

sw005320:
One discussion would be this
egsez/custom_data/asr/finetune_owsm.ipynb

Is it a better place?

1. maybe move it to TEMPLATE (custom data processing is regarded as one of the templates)?
2. just change the directory name from `custom_data` (or keep it)

Masao-Someki:
@sw005320 
I've previously discussed with @juice500ml about moving these documents to the `notebook` repository.

For instance, please refer to: https://masao-someki.github.io/notebook/

This example showcases a webpage generated with VuePress, which can convert markdown into such pages. Within this documentation, we have a `ESPnetEasy` section for storing tutorials. We can easily extend the pages to include `torch-lightning`, as Kwanghee has developed the tutorial before.

So this is one another option.

sw005320:
Thanks.
How should we add a test with this form?
Would it be on the notebook page?

siddhu001:
Yes, I will review this PR soon!

siddhu001:
I have two high-level comments on this PR:

1. Regarding my first comment, I prefer having the notebooks in a dedicated folder within the main espnet repository, such as an "egsez" folder, rather than in a separate repository like https://github.com/espnet/notebook. This arrangement consolidates all the necessary codebase and example instances into a single repository, which would make management and access easier.

2. I noticed there are no Pytest cases for this folder. I suggest adding Pytests in espnet/test/espnetez. This addition would not only ensure the robustness of the current implementation but also facilitate future updates to the project.

3. Additionally, concerning the discussion in https://github.com/espnet/espnet/pull/5584#issuecomment-1992357870, I agree that it would be beneficial to have some notebooks in the TEMPLATE folder. These notebooks could serve as helper templates that can be easily modified with minor adjustments to work on specific datasets.

espnetez/preprocess/sentencepiece.py
```diff
@@ -9,13 +9,15 @@
...
+    remove_characters: str = "",
 ):
-    """Create train.txt file for sentencepiece training from the given dump file.
+    """
+    Create train.txt file for sentencepiece training from the given dump file.
 
     Args:
         dump_text_paths (Union[str, Path]): Dump text file path.
         output_path (Union[str, Path]): Output directory for train.txt file.
+        remove_chars (str): Characters to be removed from the text.
```
siddhu001, line 20: *remove_characters

Masao-Someki:
Regarding the commit [d36ff42](https://github.com/espnet/espnet/pull/5584/commits/d36ff4203d4e9f3409b05893527735c93b77856b), this is the fix for the following error: 
```
ABCMeta object got multiple values for keyword argument
```
When we try to initialize CommonPreprocessor with the default config, there are several conflict arguments.

For instance, in the [following lines](https://github.com/espnet/espnet/blob/d0047402e830a3c53e8b590064af4bf70415fb3b/espnet2/tasks/asr.py#L429-L467), `token_type` is defined in `args.token_type`. However, when `args` is set as the default configuration, `args.preprocessor_conf` also contains `token_type`. This results in a conflict at [line 463](https://github.com/espnet/espnet/blob/d0047402e830a3c53e8b590064af4bf70415fb3b/espnet2/tasks/asr.py#L463), triggering the error mentioned above.

ci/test_integrate_easy.py
```diff
@@ -0,0 +1,177 @@
...
+    if args.task == "tts" or args.task == "gan_tts":
+        normalize = training_config["normalize"]
+        training_config["normalize"] = None
+        pitch_normalize = training_config["pitch_normalize"]
+        training_config["pitch_normalize"] = None
+        energy_normalize = training_config["energy_normalize"]
+        training_config["energy_normalize"] = None
+
+    # if args.run_config is not None:
+    #     training_config.update(parser.parse_args(shlex.split(args.run_config)))
```
sw005320, line 146: Do you want to keep these lines?
If so, please add some comments.

ci/test_integrate_easy.py
sw005320, line 1: The name is not appropriate.
Can you change it?
I think `ci/test_integration_espnetez.py` is fine.

egs2/mini_an4/asr1/conf/train_asr_transducer_debug.yaml
```diff
@@ -0,0 +1,51 @@
...
+    positional_dropout_rate: 0.1
+    attention_dropout_rate: 0.0
+    input_layer: conv1d2
+    normalize_before: true
+
+decoder: transducer
+decoder_conf:
+    rnn_type: lstm
+    num_layers: 1  # Decoder Layers
+    hidden_size: 640  # Decoder dim
```
sw005320, line 32: we can change it very small.

egs2/mini_an4/asr1/conf/train_asr_transducer_debug.yaml
```diff
@@ -0,0 +1,51 @@
...
+    dropout: 0.1
+    dropout_embed: 0.2
+
+
+joint_net_conf:
```
sw005320, line 37: ditto
we can change it very small.

test/test_easy.py
sw005320, line 1: test_ez

Please use `ez` instead of `easy`.

sw005320 (commented):
- please try to consistently use `ez`. `easy' is a too general word, and it causes confusion.
- did you include a fine-tuning test? This is the most important use case. We may use the OWSM base model (or we may make a very small pre-trained model for this test).

ci/test_integrate_easy.py
Masao-Someki: These lines are not required; I will remove them.

Masao-Someki:
@sw005320 
Thank you for your review. I overlooked adding tests for fine-tuning. I plan to use the trained model (small one) in the CI test so that both training and fine-tuning can be done in one file.

.github/workflows/ci_on_ubuntu.yml
siddhu001, line 1: Maybe add test_finetune_espnetez in this workflow.

.github/workflows/ci_on_centos7.yml
```diff
@@ -58,3 +58,4 @@ jobs:
           PATH="/opt/rh/devtoolset-7/root/usr/bin:${PATH:-}"
           ./ci/test_python_espnet1.sh
           ./ci/test_python_espnet2.sh
+          ./ci/test_python_espnetez.sh
```
siddhu001, line 61: Maybe add test_finetune_espnetez in this workflow.

.github/workflows/ci_on_debian11.yml
```diff
@@ -46,3 +46,4 @@ jobs:
         run: |
           ./ci/test_python_espnet1.sh
           ./ci/test_python_espnet2.sh
+          ./ci/test_python_espnetez.sh
```
siddhu001, line 49: Ditto as above

.github/workflows/ci_on_ubuntu.yml
siddhu001: I believe you're running it inside ./ci/test_python_espnetez.sh. In that case, you should relocate "test_finetune_espnetez.py" to the following directory: https://github.com/espnet/espnet/tree/master/test

This change is useful because files within the 'ci' directory are typically executed directly within the workflow.

ci/test_finetune_espnetez.py
siddhu001, line 1: This looks good but you can optionally check some of pytests inside https://github.com/espnet/espnet/blob/master/test/ and maybe instead of args you can use `pytest.mark.parametrize`.

sw005320, line 1: @Masao-Someki, how about `test_finetune_espnetez.py` --> `test_integration_espnetez_ft.py`?
This is still an integration test.
Changing it to finetune does lose the naming consistency.

ci/test_integration_espnetez.sh
siddhu001, line 1: Same as before. You can consider adding it to https://github.com/espnet/espnet/blob/master/test/ and maybe instead of args you can use pytest.mark.parametrize.

ci/test_integration_espnetez.py
siddhu001, line 1: Same as before. You can consider adding it to https://github.com/espnet/espnet/blob/master/test/ and maybe instead of args you can use pytest.mark.parametrize.

ci/test_integration_espnetez.sh
```diff
@@ -0,0 +1,165 @@
...
+#### Make sure chainer-independent ####
+python3 -m pip uninstall -y chainer
+
+# Download mini_an4 as test data and prepare flac data
+cd ./egs2/mini_an4/asr1
+./run.sh --stage 1 --stop-stage 1
+./run.sh --stage 2 --stop-stage 4 --feats-type "raw"
+cp -r dump/raw data/
+./run.sh --stage 2 --stop-stage 4 --feats-type "raw_copy" \
+    --train_set raw/train_nodev --valid_set raw/train_dev --test_sets raw/test
```
siddhu001, line 26: Why do we need to create dump directory twice?

test/test_ez.py
```diff
@@ -0,0 +1,220 @@
...
+        temp_dir = Path(temp_dir)
+        dataset = generate_random_dataset(5)
+        data_inputs = {
+            "speech": ["wav.scp", "sound"],
+            "text": ["text", "text"],
+        }
+        ez.data.create_dump_file(temp_dir / "dump", dataset, data_inputs)
+        ez.preprocess.prepare_sentences([temp_dir / "dump" / "text"], temp_dir / "spm")
+
+        # It seems training an sentencepiece in CI and pytest may cause error.
```
siddhu001, line 213: Maybe remove this?

siddhu001:
Thanks @Masao-Someki for your great work! I have added some minor suggestions but overall, the PR looks good to me.

.github/workflows/ci_on_ubuntu.yml
Masao-Someki: @siddhu001 
Thank you, I understand. Then test_integration_espnetez.py also needs to be relocated?

ci/test_finetune_espnetez.py
Masao-Someki: I think this finetuning test should be an integration test, because it requires trained checkpoints from the test_integration_espnetez.py

ci/test_integration_espnetez.sh
Masao-Someki: I refer to the espnet2 test to write this part, but will check if using the only line 23 is enough.

test/test_ez.py
Masao-Someki: I left this part to test the preprocessing function for sentencepiece.

sw005320 (approved):
This looks good to me.
After the CI test is passed, I’ll merge this PR.

sw005320:
Can you fix the CI issue?

PR #5600, opened 2023-12-21
Tts2 support parallel_wavgan vocoder

jctian98:
Can merge this first: it can work as expected.

TODO: will clean up the code: current dependency is not clean: we intend to make `tts` and `tts2` independent.


ftshijt:
Thanks for the update! It looks great to me

PR #5034, opened 2023-03-19
adapter modules update

jwrh:
Hi, this is the first part of `FindAdaptNet` PR update -- the adapter module.

PR #5609, opened 2024-01-05
Integrate adapter for s3prl frontend

Stanwang1210:
## What?

The PR is to support add adapter to s3prl frontend SSL models.
I integrate it with original lora implementation.
Currently, it only supports houlsby adapter. However, we can easily integrate other types of adapter very soon.
Two config yaml give examples of the configuration of adapter

TODO: Havn't done with the adapter load/save function. Will handle it soon

## Why?
The original lora implementation are not compatible with other kinds of adapter.

## See also
#5034
@ftshijt


sw005320:
Cool!
How about adding the result and corresponding config as well?

Stanwang1210:
For the lora case, I didn't modified any code except for the entry to _create_lora_adapter_ function.
So the main modification is at the houlsby adapter case.
For those target layers idx assigned in the [adapter_conf](https://github.com/Stanwang1210/espnet/blob/461bb332fbb97eb9e592e844bbcad0d7aceda6f0/egs2/ml_superb/asr1/conf/tuning/train_asr_s3prl_houlsby.yaml#L39C8-L39C8) ,
`model.frontend.upstream.upstream.model.encoder.layers.idx` will become `HoulsbyTransformerSentenceEncoderLayer`.

Basically, it looks like 
layer

Did I make it clear?

sw005320:
Cool.
Can you add some usages to https://github.com/espnet/espnet/blob/master/egs2/ml_superb/asr1/README.md?

@ftshijt, by following the convention, it would be better to have the result and model link.

Stanwang1210:
Sorry for the late reply
Now I add model link and result in the [README](https://github.com/Stanwang1210/espnet/blob/master/egs2/ml_superb/asr1/README.md#adapter-usage-guidelines).
Please let me know if I miss anything

@ftshijt

espnet2/train/trainer.py
```diff
@@ -210,10 +211,14 @@ def run(
...
-        if use_lora and lora is None:
-            raise RuntimeError("Requiring loralib. Do 'pip install loralib'")
+        adapter = getattr(trainer_options, "adapter", None)
+        use_adapter = getattr(trainer_options, "use_adapter", False)
+        save_adapter_only = getattr(trainer_options, "save_adapter_only", False)
+        if use_adapter:
+            if adapter == "lora" and lora is None:
+                raise RuntimeError("Requiring loralib. Do 'pip install loralib'")
+
+            # TODO: houlsby adapter may need S3PRL?
```
ftshijt, line 221: Is it possible to make it general to other modules as well? If that's difficult, we may keep only to s3prl for now.

In that case, you may consider add a check to ensure that s3prl is installed when using the houlsby adapter.

espnet2/train/trainer.py
```diff
@@ -359,12 +364,15 @@ def run(
...
                 # 4. Save/Update the checkpoint
-                if use_lora and save_lora_only:
-                    # Only the LoRA realted params are saved, not the whole model
-                    model_state_dict = lora.lora_state_dict(model)
-                else:
-                    # Save all params of the model
-                    model_state_dict = model.state_dict()
+                # TODO: This kind of design may not facilitate the integration of other kind of the adapter
+                # Maybe we can only save those with requires_grad = True ?
+                # If we use lora.lora_state_dict, we will not save the downstream model in SSL settings
```
ftshijt, line 369: I do not have a clear question to this one. But I feel only save models with requires_grad = True is a bit risky.

Ping @sw005320 @wanchichen @simpleoier @pyf98 for some discussion here.

ftshijt (commented):
Thanks for the updates! The current implementation seems cool to me, but I would like to hold more discussions on some design points.

espnet2/train/trainer.py
simpleoier: Is this update compatible with pretrained models which have configs `use_lora` and `save_lora_only`?

espnet2/train/trainer.py
simpleoier: I think so. One case is that it can fail if the pretrained SSL models are updated.

ftshijt: For that purpose, I would suggest we simply follow the setting with the current design.

ftshijt:
Could you fix the conflicts and update the pre-trained model in the readme session? then we can move forward to merge it as it is a dependency for some other projects~

espnet2/train/trainer.py
Stanwang1210: Sorry for the late reply.
@simpleoier I think it's compatible but need to do some revision.
And I did it by rename `use_lora` to `use_adapter` (more general), and the same way with `save_lora_only`.
The reason is that in most case, we add lora only on the pre-trained model. However, in SSL settings, most of the time we need to reinitialize a downstream model for each task. That is to say, when applying lora to SSL models as adapters, we need to save not only the [lora parameters](https://github.com/espnet/espnet/blob/b7b37f5689fa4c065f2b1a0df3f9b7da12190093/espnet2/train/trainer.py#L365) but also the downstream model and other tunable parameters. Therefore, I choose to save all parameters `requires_grad = True`.

espnet2/train/trainer.py
Stanwang1210: The reason why I choose to save all parameters with `requires_grad = True' is that in adapter setting, we usually need to save the downstream model or other tunable parameters together (Ex, the weighted_sum). Therefore, we may not able to follow what loralib did, which only save lora_layers. 
I understand your concern about the current design. However, for the case SSL models are updated, I think it would be better to save the SSL weights (or I can not see why we need to update it without saving it). If we have to follow what loralib does by offering an options which saves the adapter parameter only, then I   would like to add another option for saving parameters with `requires_grad = True'. In this case, we can fulfill the requirement but also make sure adapter and lora can work in s3prl settings.
Do you think it's a good idea?

Stanwang1210: I made some revisions here.
Please check [d7f0b39](https://github.com/espnet/espnet/pull/5609/commits/d7f0b399dd47ccb7dd2714b0b584fad8b5a3c2be)

@ftshijt @simpleoier

espnet2/train/trainer.py
Stanwang1210: I made some revisions here.
Please check [d7f0b39](https://github.com/espnet/espnet/pull/5609/commits/d7f0b399dd47ccb7dd2714b0b584fad8b5a3c2be)

sw005320:
Please fix the above conflict.

sw005320:
Please fix the CI error.
https://github.com/espnet/espnet/actions/runs/7868548740/job/21465964342?pr=5609#step:6:337

Stanwang1210:
@sw005320 
I found one error happened when I 'import transformers'
I would like to know the preference of espnet here.
Should I use 'try' & 'exception' here (Iess preferable for me) 
or try to extract the code from 'transformers' modules and put it in somewhere in the espnet repo (like espnet/espnet/nets/pytorch_backend/transformer)

Thank you

sw005320:
> @sw005320 I found one error happened when I `import transformers` I would like to know the preference of espnet here. Should I use `try` & `exception` here (Iess preferable for me) or try to extract the code from 'transformers' modules and put it in somewhere in the espnet repo (like espnet/espnet/nets/pytorch_backend/transformer)
> 
> Thank you

Which error?
Also, can you check how the other test avoids the issue?

egs2/ml_superb/asr1/README.md
```diff
@@ -96,6 +96,66 @@ General steps to run tasks in LID trask are as follows:
 ```
 ./run_multi.sh --asr_config <your_training_config> --duration {10min, 1h} --lid true --only_lid false
 ```
+## Adapter usage guidelines
```
sw005320, line 99: Would it be better to note that the challenge does not allow the fine-tuning of SSL models?
@ftshijt, what do you think?

espnet2/layers/houlsby_adapter_layer.py
```diff
@@ -0,0 +1,123 @@
...
+
+    def forward(self, x):
+        return self.houlsby_adapter(x)
+
+
+if not is_s3prl_available:
+    HoulsbyTransformerSentenceEncoderLayer = None
+else:
+
+    class HoulsbyTransformerSentenceEncoderLayer(TransformerSentenceEncoderLayer):
```
sw005320, line 35: It seems that this class does not go through the test.
Can you check it?

Stanwang1210: May you check this now?

Stanwang1210:
It seems like the CI error happened due to some weird [reason](https://github.com/espnet/espnet/actions/runs/7987462946/job/21810006956?pr=5609#step:9:540). Is this caused by my PR?

sw005320:
It's not related to your PR.
This happens accidentally.
I reran the CI.

ftshijt:
> @ftshijt, do you know why codecov complains about this? I think @Stanwang1210 correctly prepared the test.
> 
> If this is due to some issues in codecov, we can ignore it and merge this PR.
> 
> image

Not exactly from the codecov, I think it is mainly due to an in-successfully running of other CI tests (i.e., the time out for vits decoding), which automatically stop some running CIs, resulting in no execuation of the test function from Stan.

sw005320:
Thanks, @Stanwang1210!

pengchengguo:
It seems like we missed `lora.mark_only_lora_as_trainable(model)` in `espnet2/layers/create_adapter_fn.py` and all params will be optimized even with lora finetuning.

sw005320:
Thanks @pengchengguo!
Can you explain it a bit more?

pengchengguo:
Certainly. Before the integration, the LoRA fine-tuning process was implemented as follows:

```python
## find target param -> add LoRA layer -> make only LoRA params trainable
for key in key_list:
    parent_module, target_name, target_module = get_submodules(model, key)
    if not isinstance(target_module, lora.LoRALayer):
        new_module = create_new_module(target_module, rank, alpha, dropout_rate)
        replace_module(parent_module, target_name, target_module, new_module)
    else:
        continue

lora.mark_only_lora_as_trainable(model)
```

After the PR, the last line `lora.mark_only_lora_as_trainable(model)` was missing, which causes all parameters to be updated, even when using LoRA.

sw005320:
Oh, that would be a big issue.
@Stanwang1210, can you double-check it?

ftshijt:
I remember we used to fix the issue through this PR. https://github.com/espnet/espnet/pull/5726 but it seems not available anymore.

Stanwang1210:
@pengchengguo 

Sorry for the late reply!
Based on the discussion in PR #5726, we get to the conclusion that using `freeze_param` option should be fine.
If you have any questions about `freeze_param` option, feel free to let us know!

pengchengguo:
I see. Thanks for your responding.

PR #5620, opened 2024-01-10
CTC prefix score, fix if blank == eos

albertz:
In my case, I have the EOS/SOS token at index 0 in my vocab. The CTC code here also has blank index hardcoded at index 0. In principle, it should not be a problem to reuse the EOS token as blank index for CTC, however, the code was behaving incorrect for this case and overwrote the EOS scores.


sw005320:
Thanks, @albertz!
@takaaki-hori, can you review this PR?

takaaki-hori (approved):
Thank you for the fix. LGTM.

sw005320:
Thanks a lot, @albertz!

PR #5623, opened 2024-01-17
Fix relative setting of train-dev-test

ftshijt:
For easier switch between all, original, acesinger settings in KiSing-v2 corpus


ftshijt:
The TTS ci tests' failures seem to be from the error in loading speechMOS module introduced recently. Any thoughts on that would be appreciate! @Takaaki-Saeki

Takaaki-Saeki:
Sorry for the late reply and thanks for working on it!

There seem to be test time-out in downloading the SpeechMOS pretrained checkpoint during the test.
If we could increase the time and CPU for the test, it might prevent the time-out.
Or we might need to simply remove the tests.
Do you have any thoughts? @ftshijt

PR #5627, opened 2024-01-20
add interspeech2024_dsu_challenge/asr2

simpleoier:
## What?

Add ASR2 recipe for IS2024 DSU challenge baseline


ftshijt (commented):
One quick comment: it would be beneficial if we can add an entry in egs2/README.md for the recipe

sw005320:
@simpleoier, I think it is better to be merged quickly.
Can you reflect @ftshijt's comments?

PR #5640, opened 2024-01-27
bug fixes for must_c v2 recipe

jasonmusespresso:
## What?
- remove check on existance of destination data directory.
- remove `-d` option in `tar` uncompression

## Why?
for the first item, the directory should already exist because we put raw data archive in it. The original implementation skips when the directory already exists, which will result in slient failure.

## Tests
- tested on slurm with the en-de


sw005320:
Thanks!

PR #5642, opened 2024-01-29
Fix CI

siddhu001:
Fix CI check https://github.com/espnet/espnet/actions/runs/7679127517/job/20929707611?pr=5640#step:9:2368


sw005320:
Thanks!

PR #5648, opened 2024-02-01
added en-zh, and en-ja to MUST_C_V2 dataset (ST) recipes

jasonmusespresso:
## What?
- added and tested two new target languages

## Why?
language expansion for must_c_v2

## Test
en-ja: testing on slurm ⌛️
en-zh: testing on slurm ⌛️

### DIFF
--- a/egs/must_c/st1/local/download_and_untar.sh
+++ b/egs/must_c/st1/local/download_and_untar.sh
@@ -50,7 +50,15 @@ fi
 if [ ${version} = "v1" ]; then
     tar_path=${data}/MUSTC_v1.0_en-${lang}.tar.gz
 elif [ ${version} = "v2" ]; then
+  if [ "${lang}" = "de" ]; then
     tar_path=${data}/MUSTC_v2.0_en-${lang}.tar.gz
+  elif [ "${lang}" = "ja" ]; then
+    tar_path=${data}/MUSTC_v2.0__KSA__en-ja__without-H5.tar.gz
+  elif [ "${lang}" = "zh" ]; then
+    tar_path=${data}/MUST_v2.0_KSA__en-zh_without-H5.tar.gz
+  else  # other languages also use the default path
+    tar_path=${data}/MUSTC_v2.0_en-${lang}.tar.gz
+  fi
 fi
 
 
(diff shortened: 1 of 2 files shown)


jasonmusespresso:
FYI @brianyan918

weikeduo1:
must_c_v2/asr1 missing data.sh

sw005320:
- @brianyan918, can you review it?

- @jasonmusespresso, please add the results and config files. Also, please upload the model and put the link to the results.

sw005320:
See https://github.com/espnet/espnet/blob/master/egs2/aishell/asr1/README.md as a reference.

sw005320:
Also, see https://github.com/espnet/espnet/blob/master/CONTRIBUTING.md#132-espnet2-recipes

brianyan918:
LGTM thanks @jasonmusespresso

egs2/must_c_v2/st1/conf/slurm.conf
sw005320, line 1: we should not change it with our local setup

sw005320:
It looks good to me, but I want to have input from @brianyan918 

@brianyan918,
- can you confirm the change for `sacrebleu_opt_extra`?
- compared with the other recipes, the README.md style is different, which is fine. However, I could not find the model link for all models. @jasonmusespresso seems to follow this style. Is it OK? I think it is better to provide the model link.

brianyan918:
> It looks good to me, but I want to have input from @brianyan918
> 
> @brianyan918,
> 
> * can you confirm the change for `sacrebleu_opt_extra`?
> * compared with the other recipes, the README.md style is different, which is fine. However, I could not find the model link for all models. @jasonmusespresso seems to follow this style. Is it OK? I think it is better to provide the model link.

The sacrebleu_opt_extra looks good, but shouldn't there be some value passed from run.sh? @jasonmusespresso

jasonmusespresso:
> Details

@brianyan918 yes, added it. in my run, I pass it when calling ./run.sh (e.g. ./run.sh --stage 13 --stop-stage 17 --sacrebleu_opt_extra "-tok ja-mecab -l en-ja --smooth-method exp"...), which should also work.

PR #5661, opened 2024-02-15
Add evaluation scripts with DiscreteSpeechMetrics.

Takaaki-Saeki:
## What?
This PR adds evaluation scripts using discrete-speech-metrics, which can be used to evaluate text-to-speech, noisy speech, and speech enhancement.
The evaluation metrics are computed between the candidate speech and the corresponding reference speech.

<div align="left">

</div>

## Why?

The metrics in discrete-speech-metrics are effective for speech synthesis evaluation because they better correlate with human subjective evaluation scores than MCD.
They also show higher correlations than UTMOS if reference speech is available.

<div align="left">

</div>

They can also be effective for noisy speech evaluation.
<div align="left">

</div>

## See also

Toolkit: https://github.com/Takaaki-Saeki/DiscreteSpeechMetrics

### DIFF
--- a/egs2/TEMPLATE/asr1/pyscripts/utils/evaluate_speechbertscore.py
+++ b/egs2/TEMPLATE/asr1/pyscripts/utils/evaluate_speechbertscore.py
@@ -23,8 +23,8 @@ import librosa
 import numpy as np
 import soundfile as sf
 import torch
-from scipy import spatial
 from discrete_speech_metrics import SpeechBERTScore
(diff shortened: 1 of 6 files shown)


ftshijt (commented):
The scripts looks great to me! 

It would be great to
- add corresponding examples in `egs2/TEMPLATE/tts1/README.md` 
- corresponding entries in the `tools/Makefile`
- try-exception to check if discrete_speech_metrics are installed or not

Issue #5654, opened 2024-02-08
finetune whisper, preprocessor_conf, preprocessor_class, an unexpected keyword argument 'tokenizer_language'


anonymoussky:
I use "run_whisper_finetune.sh" in the "espnet/egs2/aishell/asr1" folder. I can successfully run the script "sh run.sh", but I came across this error below when I try to finetune whisper using aishell data: unexpected keyword argument 'tokenizer_language'

preprocessor='default', preprocessor_conf={'tokenizer_language': 'zh'}, required=['output_dir', 'token_list'], version='202402', distributed=False)
Traceback (most recent call last):
  File "/opt/miniconda/lib/python3.9/runpy.py", line 197, in _run_module_as_main
    return _run_code(code, main_globals, None,
  File "/opt/miniconda/lib/python3.9/runpy.py", line 87, in _run_code
    exec(code, run_globals)
  File "/opt/miniconda/lib/python3.9/site-packages/espnet2/bin/asr_train.py", line 23, in <module>
    main()
  File "/opt/miniconda/lib/python3.9/site-packages/espnet2/bin/asr_train.py", line 19, in main
    ASRTask.main(cmd=cmd)
  File "/opt/miniconda/lib/python3.9/site-packages/espnet2/tasks/abs_task.py", line 1120, in main
    cls.main_worker(args)
  File "/opt/miniconda/lib/python3.9/site-packages/espnet2/tasks/abs_task.py", line 1324, in main_worker
    preprocess_fn=cls.build_preprocess_fn(args, train=False),
  File "/opt/miniconda/lib/python3.9/site-packages/espnet2/tasks/asr.py", line 429, in build_preprocess_fn
    retval = preprocessor_class(
TypeError: __init__() got an unexpected keyword argument 'tokenizer_language'

I use docker: "./run.sh --docker-gpu 0 --is-egs2 --docker-egs an4/asr1 --ngpu 1"
The solution "pip uninstall openai-whisper and then re-install with pip install openai-whisper==20230308" did not help me.

sw005320:
Thanks for the report!
@pengchengguo, can you fix them?
It has been used for several places https://github.com/search?q=repo%3Aespnet%2Fespnet%20tokenizer_language&type=code

pengchengguo:
Hi @anonymoussky,

Thank you for pointing out the mistake, it has been corrected in https://github.com/espnet/espnet/pull/5666

sw005320:
Thanks, @pengchengguo
[...]

PR #5666, opened 2024-02-19
Correct the argument errors in the whisper tokenizer language.

pengchengguo:
## What?

Correct the argument errors in the whisper tokenizer language.

## See also
The argument for the whisper tokenizer language does not match as mentioned in https://github.com/espnet/espnet/issues/5654.


sw005320:
Thanks, @pengchengguo!

Issue #5656, opened 2024-02-10
Slue voxpopuli 


maherr13:
Hello, 

Can you tell me how can i reproduce the results mentioned in the following directory [slue-voxpopuli](https://github.com/espnet/espnet/tree/master/egs2/slue-voxpopuli/slu1) ?? 

Thanks for your great work

sw005320:
You can just specify the config files written in the README.md file when you run the experiments i.e., change https://github.com/espnet/espnet/blob/master/egs2/slue-voxpopuli/slu1/run.sh#L12 to the specific config file and `./run.sh`

maherr13:
Thanks for the response,

I just did that for the first experiment, and it seems the dataset is not downloaded automatically. 
```
2024-02-11T18:49:40 (slu.sh:238:main) ./slu.sh --lang en --ngpu 1 --use_lm false --token_type bpe --gpu_inference true --nbpe 1000 --bpe_nlsyms FILL,SEP,PLACE,QUANT,ORG,WHEN,NORP,PERSON,LAW --feats_type raw --audio_format flac.ark --speed_perturb_factors 0.9 1.0 1.1 --max_wav_duration 30 --feats_normalize utterance_mvn --slu_config conf/tuning/train_asr_no_pretrain.yaml --inference_config conf/decode_asr.yaml --train_set train --valid_set devel --test_sets test devel --lm_train_text dump/raw/train_sp/text --lm_train_transcript dump/raw/train_sp/transcript --bpe_train_text data/train/text
2024-02-11T18:49:41 (slu.sh:447:main) Stage 1: Data preparation for data/train, data/devel, etc.
2024-02-11T18:49:41 (data.sh:17:main) local/data.sh 
stage 1: Download data to downloads
2024-02-11T18:49:41 (data.sh:45:main) stage 2: Data Preparation
/content/espnet/egs2/slue-voxpopuli/slu1/local/data_prep_original_slue_format.py:7: DeprecationWarning: 
Pyarrow will become a required dependency of pandas in the next major release of pandas (pandas 3.0),
(to allow more performant data types, such as the Arrow string type, and better interoperability with other libraries)
but was not found to be installed on your system.
If this would cause problems for you,
please provide us feedback at https://github.com/pandas-dev/pandas/issues/54466
        
  import pandas as pd
Traceback (most recent call last):
  File "/content/espnet/egs2/slue-voxpopuli/slu1/local/data_prep_original_slue_format.py", line 47, in <module>
[...]

PR #5671, opened 2024-02-22
Fix speed perturbation when not using transcript in slu.sh

siddhu001:
## What?

I updated the "utt_extra_files" to be set as none (i.e., "") when not utilizing the transcript.

## Why?

This caused some issues when performing speed perturbation without utilizing the transcript.

## See also

https://github.com/espnet/espnet/issues/5656


sw005320:
Thanks!

PR #5677, opened 2024-02-26
Update bitrate calculation scripts for the IS24 discrete speech challenge

ftshijt:
Two scripts for bitrate calculation.

The `calculate_bitrate.py` is for processing the submission package (i.e., all submission from participants shall be able to use this one to compute)

The `convert_token2json.py` is for converting espnet baseline results into a submission package.

Please consider merging it in advance so @simpleoier can push the ASR side scoring and I update the TTS and SVS scoring.


sw005320:
OK, thanks.
Please also add the test for this function when you make a PR for scoring.

sw005320:
I'm merged it assuming that the CI error is not related to this.

PR #5678, opened 2024-02-26
Fix over-lengthy audio in ml_superb data prep

ftshijt:
Another issue resulting from https://github.com/espnet/espnet/pull/5676#issuecomment-1963497902 , some extremely lengthy utterances are included in the benchmark though they are not considered in the `asr1` setup.

To keep `asr1` and `asr2` aligned, we need to remove the over-lengthy audio in ml_superb data from the local/data.sh to prevent it from happening.

### DIFF
--- a/egs2/ml_superb/asr1/local/data_prep.py
+++ b/egs2/ml_superb/asr1/local/data_prep.py
@@ -265,11 +265,12 @@ if __name__ == "__main__":
                 if reserve_flag and utt_id not in FEW_SHOT_SELECTED_DATA[lang]:
                     continue
 
-                # skip wavefile over 30s, the default setting in ml-superb benchmark
+                # skip wavefile over 20s
+                # the default setting in ml-superb benchmark (asr1)
                 wav_details, sample_rate = sf.read(os.path.join(
                             args.source, dataset, lang, "wav", "{}.wav".format(utt_id)
                         ))
-                if len(wav_details) / sample_rate > 30:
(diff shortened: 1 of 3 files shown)


egs2/ml_superb/asr1/local/data_prep.py
```diff
@@ -263,6 +265,17 @@ def process_text(text):
...
                     continue
+
+                # skip wavefile over 20s
+                # the default setting in ml-superb benchmark (asr1)
+                wav_details, sample_rate = sf.read(
+                    os.path.join(
+                        args.source, dataset, lang, "wav", "{}.wav".format(utt_id)
+                    )
+                )
+                if len(wav_details) / sample_rate > 20:
```
sw005320, line 276: Can you make 20 seconds as a variable and 20 seconds as the default value?
Also, how about adding a warning to cut such data with the file ID, which will make it easy for the user to understand this thresholding process??

simpleoier:
With this fix #5682, `asr2` may be fine now. Similar strategy as in `asr1` is applied.

ftshijt:
> With this fix #5682, `asr2` may be fine now. Similar strategy as in `asr1` is applied.

Yeah, I saw the fix (which is very cool to the problem and also great for more general usage). But, at the same time, I feel it would be good to update the data preparation directly so that even if other might use ml-superb for other occasions, then can be aware of the removal of lengthy audio explicitly (not from inexplicit filtering applied in our recipe stages)

PR #5679, opened 2024-02-26
TransformerDecoder forward_one_step with memory_mask

albertz:
## What?

I added the `memory_mask` arg to `BaseTransformerDecoder.forward_one_step`.

Alongside, I made the `cache` arg a keyword-only arg, such that the user can not rely on the specific order of positional args. I think that's cleaner.

## Why?

I needed the `memory_mask` arg for `TransformerDecoder.forward_one_step` because I wanted to perform batched beam search (i.e. on multiple sequences at once, i.e. the batch dim in `memory` is >1).

### DIFF
--- a/espnet/nets/pytorch_backend/transformer/decoder.py
+++ b/espnet/nets/pytorch_backend/transformer/decoder.py
@@ -246,7 +246,7 @@ class Decoder(BatchScorerInterface, torch.nn.Module):
             x = self.output_layer(x)
         return x, tgt_mask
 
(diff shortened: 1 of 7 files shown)


sw005320:
Thanks, @albertz!

pyf98:
Looks good to me! Since we do not implement batched inference right now, it should not affect the current behavior.

I also found this issue when preparing the assignment for our speech course. In that assignment, I did a similar thing to add the mask so that we can use batched inference.

PR #5682, opened 2024-02-27
Update egs2/TEMPLATE/asr2/asr2.sh

simpleoier:
## What?

Updating the asr2.sh, filtering utterances at stage 4, and moving kmeans step after it, to stage 5.
In the previous script, filtering was done after kmeans, in stage 6. It only removes the utterances with 0-length in the target text. Now I make the filtering step as stage 4, taking into consideration of both wav length and target text length. The previous stage 6 is no longer needed.

## Why?

asr2.sh suffers from the too short / long utterances in kmeans pseudo-labeling step, because no filtering was done before kmeans.


sw005320:
Thanks!

PR #5688, opened 2024-02-29
Fix typo in summarization scoring

YoshikiMas:
I obtained a reasonable score with `result['rouge2']*100` instead of `result['rouge2']**100`.


sw005320:
Thanks, @YoshikiMas!

@roshansh-cmu, @siddhu001, @wanchichen, I just want to let you know about this bug. I hope this will not be critical for some of your experiments.

roshansh-cmu:
Thanks for the fix @YoshikiMas - it was on my TODO. 
However, it doesn't cause issues with my results since I had fixed it locally.

siddhu001:
Thanks for the fix @YoshikiMas! 

Fortunately, I only report ROUGE-L, so this adjustment does not affect my previous results.

PR #5692, opened 2024-03-02
Quick fix to calculation of bitrate

ftshijt:
## What?

A quick fix to the calculation of bitrate


simpleoier (approved):

sw005320:
Merged it by assuming that this does not affect CI

Issue #5701, opened 2024-03-14
Upgrade typeguard version


duguyue100:
**Describe the issue**

I'm using `jedi-language-server` which uses `pygls` under the hood. `pygls` in general requires `typeguard > 3`. However `espnet` requires a specific version of `typeguard=2.13.3`. As you can see, after installing `espnet` my code completion stops working. If I simply upgrade `typeguard`, then the `espnet` will stop working.

I'm not an expert on `typeguard` but I guess you chose this specific version for a reason.
Is it possible to upgrade the `typeguard` version in the next release?

I think there are many users like me who have this setup.

**A workaround**: I can downgrade `jedi-language-server` to 0.40.0 and `pygls` to 1.0.0 manually to avoid this problem. But it will become annoying when I accidentally upgrade the package and have to do the whole downgrading again.

sw005320:
We can successfully update `typeguard`

Issue #5660, opened 2024-02-14
Support for new typeguard version


albertz:
Currently you pin to an old typeguard version. This is a problem for me, because I want to use it together with some other code which requires a newer typeguard version.

As I understand https://github.com/espnet/espnet/pull/5009, just no-one so far has invested some effort in updating the code? But in principle, you are ok with updating the code?

As I understand it, instead of using `check_argument_types()` and `check_return_type()`, use the `@typechecked` decorator on the function.

Note, I also asked whether we can maybe get back `check_argument_types()` and `check_return_type()` in newer versions of typeguard, but I doubt that they would add it back. https://github.com/agronholm/typeguard/issues/437

sw005320:
Thanks, @albertz.

@kamo-naoyuki, can you answer it?

kamo-naoyuki:
@albertz Yes. It's great if you'll update.

sw005320:
We can successfully update `typeguard`

PR #5702, opened 2024-03-14
Upgrade typeguard

sw005320:
## What?

- use the latest version of `typeguard`
- moved `check_argument_types()` to the beginning of the function and renamed as `@typechecked`
- deleted `check_return_type` by assuming that `@typechecked` can cover it.

## Why?

There are several requests to update `typeguard`

## See also

https://github.com/espnet/espnet/issues/5701 
https://github.com/espnet/espnet/issues/5660
https://github.com/espnet/espnet/pull/5009


espnet2/uasr/espnet_model.py
```diff
@@ -119,8 +119,9 @@ def number_updates(self):
         return self._number_updates
 
     @number_updates.setter
+    @typechecked
     def number_updates(self, iiter: int):
-        assert check_argument_types() and iiter >= 0
+        iiter >= 0
```
simpleoier, line 124: This might not work without `assert`

simpleoier (approved):
Thanks a lot! This is a lot of effort!
There is one typo in the PR.
I think this should be good if it passes the CI tests.

PR #5703, opened 2024-03-15
fix a small issue in OWSM decode_long

jctian98:
## Why?
The last time-stamp prediction in the last chunk is slightly smaller than the whole speech length, making to an extra chunk that usually has < 1000 samples. The decoding behavior of this very short chunk leads to unpredictable behavior.

Remove one redundant line.

### DIFF
--- a/espnet2/bin/s2t_inference.py
+++ b/espnet2/bin/s2t_inference.py
@@ -188,6 +188,7 @@ class Speech2Text:
         lang_sym: str = "<eng>",
         task_sym: str = "<asr>",
         predict_time: bool = False,
+        skip_last_chunk_threshold: float = 0.2,
     ):
         assert check_argument_types()
 
@@ -344,6 +345,7 @@ class Speech2Text:
         self.device = device
         self.dtype = dtype
         self.nbest = nbest
+        self.skip_last_chunk_threshold = skip_last_chunk_threshold
 
         self.lang_sym = lang_sym
         self.task_sym = task_sym
@@ -566,23 +568,28 @@ class Speech2Text:
         if isinstance(speech, np.ndarray):
             speech = torch.tensor(speech)
 
-        assert (
-            speech.dim() == 1
-        ), f"speech must have one dimension, got size {speech.size()} instead"
+        if speech.dim() > 1:
+            assert (
+                speech.dim() == 2 and speech.size(1) == 1
+            ), f"speech of size {speech.size()} is not supported"
+            speech = speech.squeeze(1)  # (nsamples, 1) --> (nsamples,)
 
         utterances = []
         offset = 0
         text_prev = init_text
         while offset < len(speech):
             logging.info(f"Current start time in seconds: {offset / fs:.2f}")
-            if offset + segment_len > len(speech) and len(segment) / fs < 0.2:
+            segment = speech[offset : offset + segment_len]
+            if (
+                offset + segment_len > len(speech)
+                and len(segment) / fs < self.skip_last_chunk_threshold
+            ):
                 logging.warning(
-                    f"Skip the last clip as it's too short: {len(segment)/ fs:.2f}s"
+                    f"Skip the last clip as it's too short: {len(segment) / fs:.2f}s"
                 )
                 offset += segment_len
                 continue
 
-            segment = speech[offset : offset + segment_len]
             # segment will be padded in __call__
             result = self.__call__(
                 speech=segment,
@@ -718,6 +725,7 @@ def inference(
     lang_sym: str,
     task_sym: str,
     predict_time: bool,
+    skip_last_chunk_threshold: float,
 ):
     assert check_argument_types()
     if batch_size > 1:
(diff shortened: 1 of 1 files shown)


jctian98:
The Codecov fails since we don't have a test case for `decode_long`

Maybe it's hard to have a valid test case for `decode_long`, as it should properly predict the time stamps. However, for test cases, it cannot be easy. What do you think? @pyf98

sw005320:
OK, for the test code coverage.
If @pyf98 is OK for this change, I can merge it.

Where does 0.2 come from?
It would be safe to make it as a variable.

espnet2/bin/s2t_inference.py
```diff
@@ -575,6 +575,12 @@ def decode_long(
         text_prev = init_text
         while offset < len(speech):
             logging.info(f"Current start time in seconds: {offset / fs:.2f}")
+            if offset + segment_len > len(speech) and len(segment) / fs < 0.2:
```
pyf98, line 578: Should this be placed after `segment = speech[offset : offset + segment_len]`? Otherwise, `segment` is the previous one.
Also, we can make 0.2 as an argument in this function.
I think we do not need `offset + segment_len > len(speech)`. If `segment` is shorter than the segment length, this condition must be true.

espnet2/bin/s2t_inference.py
```diff
@@ -575,6 +575,12 @@ def decode_long(
...
+                logging.warning(
+                    f"Skip the last clip as it's too short: {len(segment)/ fs:.2f}s"
```
pyf98, line 580: Keep the style consistent. Now there is a space after `/` but no space before it.

pyf98: We can change "clip" to "segment" or "chunk"

pyf98:
The issue addressed here is:
When processing the last segment in a long audio, the end timestamp can be slightly shorter or longer than the actual speech length due to discretization (timestamps are multiples of 20ms).
If the prediction is longer, the decoding will terminate and there is no problem.
If the prediction is shorter, we need to decode one more segment which can be very short. We can skip it.

espnet2/bin/s2t_inference.py
jctian98: This is a mistake during copy-paste. Thanks!

jctian98:
It should be okay for review again.

espnet2/bin/s2t_inference.py
```diff
@@ -566,17 +568,28 @@ def decode_long(
...
 
         utterances = []
         offset = 0
         text_prev = init_text
         while offset < len(speech):
             logging.info(f"Current start time in seconds: {offset / fs:.2f}")
-
             segment = speech[offset : offset + segment_len]
+            if (
+                offset + segment_len > len(speech)
```
pyf98, line 584: Do we need this condition? This is equivalent to `len(segment) < segment_len`, where `segment_len` is much larger than our threshold in the next line. So, I think this condition will be satisfied automatically if the next condition is true.

espnet2/bin/s2t_inference.py
```diff
@@ -188,6 +188,7 @@ def __init__(
         lang_sym: str = "<eng>",
         task_sym: str = "<asr>",
         predict_time: bool = False,
+        skip_last_chunk_threshold: float = 0.2,
```
pyf98, line 191: I was thinking adding this argument in `decode_long` only, since it is specific to long-form decoding but never used in short-form `__call__`. If we remove it from `__init__`, we can also remove the argument in `inference` and `argparser`.

pyf98:
Thanks @jctian98!
Sorry for asking more, but I think we can add the argument only in `decode_long`. This is more consistent with the design of other arguments.

espnet2/bin/s2t_inference.py
jctian98: That's true. Didn't notice that if `len(segment) < segment_len`, that's the last chunk.

espnet2/bin/s2t_inference.py
jctian98: Revised.

jctian98:
fixed. Please review.

espnet2/bin/s2t_inference.py
```diff
@@ -531,6 +531,7 @@ def decode_long(
         end_time_threshold: str = "<29.00>",
         lang_sym: Optional[str] = None,
         task_sym: Optional[str] = None,
+        skip_last_chunk_threshold=0.2,
```
pyf98, line 534: One final suggestion: let's add a type hint here: `float = 0.2`

pyf98: I remember it exists in previous versions

pyf98:
LGTM

sw005320:
OK!
Does it come from this PR?
https://github.com/espnet/espnet/actions/runs/8364098601/job/22898621380?pr=5703#step:8:811

jctian98:
It fails on test-shell but this PR doesn't change shell scripts. 
Can we re-run the CI test?

jctian98:
CI fails again due to a time-out error.

sw005320:
Thanks, @jctian98!

PR #5709, opened 2024-03-19
Adding transformer_opt

soumimaiti:
## What?

Adding huggingface transformer opt model to voxtlm_v1

## Why?
Huggingface transformer opt model requires additional lm model file to matcha names

## See also
Transformer with opt models can be used with _transformer_opt_

### DIFF
--- a/egs2/voxtlm_v1/lm1/conf/decode_lm_asr.yaml
+++ b/egs2/voxtlm_v1/lm1/conf/decode_lm_asr.yaml
@@ -1,0 +1,5 @@
+log_level: INFO
+beam_size: 5
+penalty: 0.0
+maxlen: 150
+minlen: 1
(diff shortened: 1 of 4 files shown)


sw005320:
@wyh2000, can you review this PR?

wyh2000:
I checked these two updated files, and they look good! I noticed that there is no config file corresponding to the OPT pretrained model in the `egs2/voxtlm_v1/lm1/conf` folder. It might be better to include configuration files for these models.

sw005320:
> I checked these two updated files, and they look good! I noticed that there is no config file corresponding to the OPT pretrained model in the `egs2/voxtlm_v1/lm1/conf` folder. It might be better to include configuration files for these models.

Thanks!
Yes, it's better to add them.
@soumimaiti, can you do it?

PR #5579, opened 2023-12-05
Speaker embedding extractor (with ESPnet pre-trained speaker model)

ftshijt:
## What?

- Add ESPnet speaker embedding extractor (inference script)
- Add ESPnet speaker embedding extractor for TTS purpose
- Separate the spk embedding and id converting stage in TTS
    - For flexibility concerns (e.g., after formatting the waveform to use different speaker embedding)
- change xvector to spk_embed as suggested by @Jungjee 

## TODO
- [ ] replace the current spk.sh template with the new inference
- [x] upload pre-trained vctk model with the ESPnet speaker pre-trained model
- [x] test function for spk_inference.py

PR #5714, opened 2024-03-21
Fix stage references in generated run.sh in TTS recipes

G-Thor:
After an extra stage was added to tts.sh in #5579 , following stage numbers were updated.
A few were missed in the update and this PR covers those that remained.


sw005320:
Thanks, @G-Thor!

PR #5719, opened 2024-03-26
Modified easy to ez

Masao-Someki:
## What?

- Replaced `easy` to `ez`
- Replaced `Easy` to `EZ`

## Why?

`easy` or `Easy` might be confusing.


sw005320:
Thanks, @Masao-Someki!

@siddhu001 and @ftshijt
This will change the interface, and our assignment may not work after we merge it.
Can you take care of this change?

Issue #5698, opened 2024-03-10
Probably a bug in saving checkpoints and loading for inference


simpleoier:
**Describe the bug**
When I use LoRA fine-tuning, I noticed that the inference performance was pretty bad. I dug into it and found that the parameters of the original layers were quite different. This is due to the dropout layer, following the process below:
```
# During training
1. train the model for one epoch             --   model is in the train mode
2. validate for one epoch and save the ckpt  --   model is in the eval mode, parameters related to dropout would be scaled up
3. repeat train and validate

# During inference
1. load the model                      -- model is in the train mode, but the saved parameters are from the eval mode
2. turn the model to the eval mode     -- parameters would be scaled up again
```

This might affect the general models in ESPnet.

**Basic environments:**
- OS information: Linux 4.18.0-372.32.1.el8_6.x86_64 #1 SMP Fri Oct 28 15:56:52 EDT 2022 x86_64
- python version: `3.9.18 (main, Sep 11 2023, 13:41:44)  [GCC 11.2.0]`
- espnet version: `espnet 202402`
- pytorch version: `pytorch 2.1.0`
- Git hash: `ec9760b22654dc04eeecd37e2659ebda0325a786`
  - Commit date: `Sat Mar 2 01:25:58 2024 -0500`

**Environments from `torch.utils.collect_env`:**
```
Collecting environment information...
PyTorch version: 2.1.0
Is debug build: False
CUDA used to build PyTorch: 11.8
ROCM used to build PyTorch: N/A

OS: Springdale Open Enterprise Linux 8.6 (Modena) (x86_64)
GCC version: (GCC) 8.5.0 20210514 (Red Hat 8.5.0-10)
Clang version: Could not collect
CMake version: version 3.28.3
Libc version: glibc-2.28

Python version: 3.9.18 (main, Sep 11 2023, 13:41:44)  [GCC 11.2.0] (64-bit runtime)
Python platform: Linux-4.18.0-372.32.1.el8_6.x86_64-x86_64-with-glibc2.28
Is CUDA available: False
CUDA runtime version: No CUDA
CUDA_MODULE_LOADING set to: N/A
GPU models and configuration: No CUDA
Nvidia driver version: No CUDA
cuDNN version: No CUDA
HIP runtime version: N/A
MIOpen runtime version: N/A
Is XNNPACK available: True

CPU:
Architecture:        x86_64
CPU op-mode(s):      32-bit, 64-bit
Byte Order:          Little Endian
[...]

PR #5722, opened 2024-03-28
Fix LoRA issues when saving all parameters.

simpleoier:
## What?
This PR fixes discrepancy of parameters for the backbone model when applying LoRA and save all parameters.
Setting the model to eval mode at the fixing position can avoid the issue. Becase the weights would not be merged again when loading pre-trained checkpoints (in the eval mode) for inference. In addition, the weights would subtract the LoRA parameters in resuming training case, when calling `train()`. 
It should not affect the normal training, because `train()` (subtraction) is applied after `eval()` (addition). 
For example, in the LoRA Linear layers of [loralib](https://github.com/microsoft/LoRA/blob/main/loralib/layers.py#L62C1-L75C35), weights of the backbone model will be merged with LoRA parameters when switching between train and eval modes.
```
def train(self, mode: bool = True):
    nn.Embedding.train(self, mode)
    if mode:
        if self.merge_weights and self.merged:
            # Make sure that the weights are not merged
            if self.r > 0:
                self.weight.data -= (self.lora_B @ self.lora_A).transpose(0, 1) * self.scaling
            self.merged = False
    else:
        if self.merge_weights and not self.merged:
            # Merge the weights and mark it
            if self.r > 0:
                self.weight.data += (self.lora_B @ self.lora_A).transpose(0, 1) * self.scaling
            self.merged = True
```
Thanks to @pengchengguo for the discussion.

## Why?
In the case of saving all parameters in LoRA fine-tuning, the pre-trained parameters were saved in eval mode. In such cases, the original model's weights are merged with the LoRA weights, as shown in the code block above.
However, when resuming training or doing inference, the initialized model is in the train mode, different from the saved model.
1. When calling eval() again in inference, the parameters would add the LoRA parameters again , leading to poor performance.
2. When calling train() in resuming case, the parameters would not subtract the LoRA parameters from the merged weights, because it's already in the train mode. 

## See also
The issue was reported in #5698

### DIFF
--- a/espnet2/layers/create_adapter_fn.py
+++ b/espnet2/layers/create_adapter_fn.py
@@ -145,6 +145,10 @@ def create_lora_adapter(
             f"Target modules {target_modules} not found in the base model."
         )
 
+    # Set the model (originally in train mode) to eval mode
+    # This step can avoid merging LoRA weights again when loading pre-trained checkpoints
+    model.eval()
+
 
 def create_new_houlsby_module(target_module: torch.nn.Module, bottleneck: int):
     """Create a new houlsby adapter module for the given target module\n.
--- a/espnet2/layers/create_adapter_fn.py
+++ b/espnet2/layers/create_adapter_fn.py
@@ -146,7 +146,8 @@ def create_lora_adapter(
         )
 
     # Set the model (originally in train mode) to eval mode
-    # This step can avoid merging LoRA weights again when loading pre-trained checkpoints
+    # This step can avoid merging LoRA weights again
+    # when loading pre-trained checkpoints
     model.eval()
 
 


pengchengguo:
This PR will only impact LoRA training while saving all parameters.
We have tested it across different scenarios to ensure its accuracy:
1. Training from scratch;
2. Resuming a training process;
3. Inference.

espnet2/layers/create_adapter_fn.py
```diff
@@ -145,6 +145,10 @@ def create_lora_adapter(
             f"Target modules {target_modules} not found in the base model."
         )
 
+    # Set the model (originally in train mode) to eval mode
+    # This step can avoid merging LoRA weights again when loading pre-trained checkpoints
```
sw005320, line 149:
```suggestion
    # This step can avoid merging LoRA weights again
    # when loading pre-trained checkpoints
```

sw005320:
Thanks, @simpleoier and @pengchengguo!

Issue #5721, opened 2024-03-27
An error when using LoRA for s3prl frontend.


simpleoier:
**Describe the bug**
When applying LoRA fine-tuning on the s3prl frontend (e.g. hubert_base), the output has no gradients. More specifically, I simply used the last layer, instead of using the multi-layer weighted sum.

**Basic environments:**
 - OS information: Linux 4.18.0-372.32.1.el8_6.x86_64 #1 SMP Fri Oct 28 15:56:52 EDT 2022 x86_64
- python version: `3.9.18 (main, Sep 11 2023, 13:41:44)  [GCC 11.2.0]`
- espnet version: `espnet 202402`
- pytorch version: `pytorch 2.1.0`
- Git hash: `ec9760b22654dc04eeecd37e2659ebda0325a786`
  - Commit date: `Sat Mar 2 01:25:58 2024 -0500`

**Environments from `torch.utils.collect_env`:**
e.g.,
```
Collecting environment information...
PyTorch version: 2.1.0
Is debug build: False
CUDA used to build PyTorch: 11.8
ROCM used to build PyTorch: N/A

OS: Springdale Open Enterprise Linux 8.6 (Modena) (x86_64)
GCC version: (GCC) 8.5.0 20210514 (Red Hat 8.5.0-10)
Clang version: Could not collect
CMake version: version 3.28.3
Libc version: glibc-2.28

Python version: 3.9.18 (main, Sep 11 2023, 13:41:44)  [GCC 11.2.0] (64-bit runtime)
Python platform: Linux-4.18.0-372.32.1.el8_6.x86_64-x86_64-with-glibc2.28
Is CUDA available: False
CUDA runtime version: No CUDA
CUDA_MODULE_LOADING set to: N/A
GPU models and configuration: No CUDA
Nvidia driver version: No CUDA
cuDNN version: No CUDA
HIP runtime version: N/A
MIOpen runtime version: N/A
Is XNNPACK available: True

CPU:
Architecture:        x86_64
CPU op-mode(s):      32-bit, 64-bit
Byte Order:          Little Endian
CPU(s):              48
On-line CPU(s) list: 0-47
Thread(s) per core:  2
Core(s) per socket:  12
Socket(s):           2
NUMA node(s):        2
Vendor ID:           GenuineIntel
CPU family:          6
Model:               63
Model name:          Intel(R) Xeon(R) CPU E5-2670 v3 @ 2.30GHz
Stepping:            2
CPU MHz:             3100.000
[...]

PR #5726, opened 2024-04-01
Add LoRA only options

Stanwang1210:
## What?
I add an option `lora_only` for enabling `lora.mark_only_lora_as_trainable`, which freezes all parameters except for LoRA layers. The default value of `lora_only` is set to True.

## Why?
As suggested by @ftshijt , it's better for us to provide this option.

### DIFF
--- a/espnet2/tts/prodiff/denoiser.py
+++ b/espnet2/tts/prodiff/denoiser.py
@@ -343,6 +343,11 @@ class SpectogramDenoiser(nn.Module):
         post_var = beta * (1.0 - alph_prod_prv) / (1.0 - alph_prod)
         post_var = torch.log(torch.maximum(post_var, torch.full((1,), 1e-20)))
 
+        # allows non CPU denoising
+        coef1 = coef1.to(device)
+        coef2 = coef2.to(device)
+        post_var = post_var.to(device)
+
         # denoising steps
         for _step in reversed(range(0, self.timesteps)):
             # p-sample
--- a/espnet2/tts/prodiff/prodiff.py
+++ b/espnet2/tts/prodiff/prodiff.py
@@ -591,6 +591,7 @@ class ProDiff(AbsTTS):
         lids: Optional[torch.Tensor] = None,
         is_inference: bool = False,
         alpha: float = 1.0,
+        use_teacher_forcing: bool = False,
     ) -> Sequence[torch.Tensor]:
         """Calculate forward propagation without loss.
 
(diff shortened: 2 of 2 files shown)


ftshijt:
@simpleoier Hi, could you please have a check to see if your concern on the default behavior is fixed by this PR?

simpleoier:
Sorry, I may not fully understand the purpose of this PR.
This PR only adds a warning for s3prl frontend. How about replacing it with some assert checks? In that case, the training stops if LoRA is applied on s3prl frontend.
I'm afraid not all of us would read the logging information carefully.

Also, my original concern was that LoRA could not be used for HuBERT / WavLM / Wav2vec2 models in the current codebase.

Stanwang1210:
This PR is only to support the option of enabling `lora.mark_only_lora_as_trainable`. It is related to the dependency of espnetez. 
For the issue issue #5721, we will create another PR to fix it.

simpleoier:
> This PR is only to support the option of enabling lora.mark_only_lora_as_trainable

OK, now I understand the point. Thanks for the explanation.
If I understand it correctly, here if `lora_only` is set to False (before this PR), other parameters would be optimized together, right? This sounds a little bit weird to me, as LoRA is meant to be a PEFT method. 
My solution to add other trainable parameters is that if some parameters need to be optimized during training, it is better to add another process for `other_trainable_parameters`. It is more flexible. What do you think?

> @simpleoier Hi, could you please have a check to see if your concern on the default behavior is fixed by this PR?

@ftshijt I was trying to recall that. Maybe my previous concern was that the previous configs (maybe pretrained models) have set the arguments as `use_lora: true`. Now it is set to a new variable, named as 
```yaml
use_adapter: true
adapter: lora
adapter_conf: ...
```
It would be better to update those configs if there are any.

Stanwang1210:
@simpleoier 
First, your understanding toward `lora_only` is correct.  
However, I think that it's better to use `freeze_param` option to control which groups of parameters should be frozen or not. Simply freezing all other parameters when employing PEFT methods may cause the issue of unexpectedly freezing downstream models. 
Therefore, I prefer the pipeline that `create_adapter_fn` only adds adapter modules without modifying the `requires_grad` of parameters. 

For the `other_trainable_parameters`, it somehow overlaps with the `freeze_param` option because they are basically the opposite of each other. In other words, those parameters who are not specified in `freeze_param` should be specified in  `other_trainable_parameters`. Therefore, I think this design is a little bit redundant.

simpleoier:
Hi @Stanwang1210 , using `freeze_params` is fine.

Stanwang1210:
So, do we have any further problems about this PR?

PR #5732, opened 2024-04-04
ESPnet-codec Training and Setup

ftshijt:
## What?

New ESPnet codec project

The initial PR (to the local codec branch) supports:
- The library set up of the codec task
- The recipe setup for mini_an4(debugging purpose and future CI support) and libriTTS recipe
- A base training framework with soundstream model

## References

https://github.com/alibaba-damo-academy/FunCodec/tree/master
https://github.com/facebookresearch/audiocraft
https://github.com/facebookresearch/encodec
https://github.com/yangdongchao/AcademiCodec

## TODO in the following PRs:

- Decoding
- Evaluation
- Deployment
- Add additional models


ftshijt:
@jctian98 @Jungjee @wyh2000 Hi guys, may I request your review for this PR? Since it has a few framework-level design choices, it would be difficult to change in the later stage.

espnet2/gan_codec/shared/decoder/seanet.py
```diff
@@ -0,0 +1,252 @@
...
+def unpad1d(x: torch.Tensor, paddings: Tuple[int, int]):
+    """Remove padding from x, handling properly zero padding. Only for 1d!"""
+    padding_left, padding_right = paddings
+    assert padding_left >= 0 and padding_right >= 0, (padding_left, padding_right)
+    assert (padding_left + padding_right) <= x.shape[-1]
+    end = x.shape[-1] - padding_right
+    return x[..., padding_left:end]
+
+
+class NormConvTranspose1d(nn.Module):
```
Jungjee, line 32: maybe these layers could be gathered to a separate folder? Such as "espnet2/codec/layers"?

espnet2/gan_codec/shared/discriminator/stft_discriminator.py
```diff
@@ -0,0 +1,168 @@
...
+from typing import Any, Dict, List, Optional
+
+import numpy as np
+import torch
+import torch.nn as nn
+import torch.nn.functional as F
+from einops import rearrange
+
+
+class ModReLU(nn.Module):
```
Jungjee, line 21: ditto. Better for others to find if this layer would be in `layers` folder to me.

Jungjee (commented):
Thanks @ftshijt  for your cool work!
LGTM in general.

Two comments from me (maybe close to questions):
1. Having a `gan_codec` tasks sounds like there would be other `*_codec` tasks upcoming. How many tasks do you expect? Is is GAN vs. GAN-irrelevant?
2. Could we have a folder where we could see all sorts of modules used for the codec tasks? (gan_codec and others) I initially suggested `espnet2/codec/layer` but after reviewing everything maybe we could put all fundamental modules (that could be re-usable) to `espnet2/codec/shared`? The idea comes from my assumption that some modules could be used across different *_codec tasks. But maybe this is not the optimal selection.

egs2/libritts/codec1/conf/train_soundstream.yaml
```diff
@@ -0,0 +1,170 @@
...
+    weight_decay: 0.0
+scheduler2: exponentiallr
+scheduler2_conf:
+    gamma: 0.999875
+generator_first: true # whether to start updating generator first
+
+
+##########################################################
+#                OTHER TRAINING SETTING                  #
+##########################################################
```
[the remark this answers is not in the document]
ftshijt: It is applied to TTS-related config, and I found it very clear so I tried to follow it.

(I always intend to make a consistent style of doc strings and configs in ESPnet but have not found the time yet... )

espnet2/bin/gan_codec_train.py
```diff
@@ -0,0 +1,22 @@
...
+from espnet2.tasks.gan_codec import GANCodecTask
+
+
+def get_parser():
+    parser = GANCodecTask.get_parser()
+    return parser
+
+
+def main(cmd=None):
+    """GAN-based SVS training
```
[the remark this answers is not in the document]
ftshijt: This is a typo, I will fix it.

ftshijt:
> Thanks @ftshijt for your cool work! LGTM in general.
> 
> Two comments from me (maybe close to questions):
> 
> 1. Having a `gan_codec` tasks sounds like there would be other `*_codec` tasks upcoming. How many tasks do you expect? Is is GAN vs. GAN-irrelevant?
> 2. Could we have a folder where we could see all sorts of modules used for the codec tasks? (gan_codec and others) I initially suggested `espnet2/codec/layer` but after reviewing everything maybe we could put all fundamental modules (that could be re-usable) to `espnet2/codec/shared`? The idea comes from my assumption that some modules could be used across different *_codec tasks. But maybe this is not the optimal selection.

For 1, I'm following the setting to tts and svs. Using the name has the intention to state the task is trained with gan_trainer instead of the existing normal trainer used in many other tasks. I think we will only support the gan_codec task for codec training for now.

For 2, yeah, I tried to put some shared modules to `shared` folder.

espnet2/gan_codec/shared/decoder/seanet.py
ftshijt: Thanks for the suggestion! I generally agree with your point on setting a better organization of the modules. But I'm not pretty sure if the above layers are good to be separated. Mostly because they are not beyond the original torch function but just wrappers for the seanet module itself.

ftshijt: Please let me know if you think layers would be still better in the case~

espnet2/gan_codec/shared/discriminator/stft_discriminator.py
ftshijt: Similar to the above ones, I'm expecting the existing modules to be attached to the major components (the stft discriminator) tightly. In that case, I'm leaning toward saving it here instead of having another set for the layers.

If to make them separate, do you have some suggestions on how the layers folder would be organized? Would be great if you could share a bit more thoughts on this side so that I can understand more on how it could be in better shape. Again, appreciate your advice!

jctian98:
Some comments after discussion with @ftshijt Please let me know if I'm wrong.
(1) Since DAC / soundstream / encoded share the very similar structure like SEANet, quantizer, and overall encoder-decoder-discriminator design, it might be better to have a unified file (current soundstream.py) for them three rather than have the `espnet2/gan_codec/{soundstream, encoded, DAC}` folders. Instead, we can use `if/else` to specify the distinctions among these models.
(2) we can also make the loss classified and more configurable. Specifically, the losses can be categorized (my current thinking) adversarial loss / reconstruction loss / loss that specify the similarity between generator decoder and discriminator.

TiVA2024:
> Some comments after discussion with @ftshijt Please let me know if I'm wrong. (1) Since DAC / soundstream / encoded share the very similar structure like SEANet, quantizer, and overall encoder-decoder-discriminator design, it might be better to have a unified file (current soundstream.py) for them three rather than have the `espnet2/gan_codec/{soundstream, encoded, DAC}` folders. Instead, we can use `if/else` to specify the distinctions among these models. (2) we can also make the loss classified and more configurable. Specifically, the losses can be categorized (my current thinking) adversarial loss / reconstruction loss / loss that specify the similarity between generator decoder and discriminator.

My concern is that DAC actually modified some loss and quantizer details from soundstream. It might be more clear for DAC to have a separate `dac.py` file.

wyh2000:
> Some comments after discussion with @ftshijt Please let me know if I'm wrong. (1) Since DAC / soundstream / encoded share the very similar structure like SEANet, quantizer, and overall encoder-decoder-discriminator design, it might be better to have a unified file (current soundstream.py) for them three rather than have the `espnet2/gan_codec/{soundstream, encoded, DAC}` folders. Instead, we can use `if/else` to specify the distinctions among these models. (2) we can also make the loss classified and more configurable. Specifically, the losses can be categorized (my current thinking) adversarial loss / reconstruction loss / loss that specify the similarity between generator decoder and discriminator.

My concern is that for DAC, it actually modifies some loss and quantizer details from Soundstream. It might be clearer for DAC to have a separate `dac.py` file.

espnet2/gan_codec/shared/quantizer/modules/core_vq.py
```diff
@@ -0,0 +1,380 @@
...
+
+        if self.training:
+            quantize = x + (quantize - x).detach()
+
+        loss = torch.tensor([0.0], device=device, requires_grad=self.training)
+
+        if self.training:
+            if self.commitment_weight > 0:
+                commit_loss = F.mse_loss(quantize.detach(), x)
+                loss = loss + commit_loss * self.commitment_weight
```
wyh2000, line 321: It seems that there are two conflicted hyperparameters: `commitment_weight` and `lambda_commit`.

ftshijt: Thanks!

ftshijt:
Consider the following PR (decoding/evaluation), I will merge the current one first. Let's keep discussing better code organization.

PR #5724, opened 2024-03-29
Upgrade typeguard [Subst.]

Fhrozen:
## What?

Subst. from #5702

- use the latest version of typeguard
- moved check_argument_types() to the beginning of the function and renamed as @typechecked
- deleted check_return_type by assuming that @typechecked can cover it.

Note: Some variables required an explicit type change. Some were renamed, and others changed the type depending on the extension of the refactoring.

## Why?

There are several requests to update typeguard

## See also

https://github.com/espnet/espnet/issues/5701
https://github.com/espnet/espnet/issues/5660
https://github.com/espnet/espnet/pull/5009

PR #5736, opened 2024-04-06
CI Fix on scipy api change

ftshijt:
## What?

API fix for recent CI failure.


ftshijt:
My bad, already fixed in https://github.com/espnet/espnet/pull/5724

PR #5745, opened 2024-04-17
Bugfix: This line bombs and is not necessary

EvanFellman:
## What?

I commented out a debugging line that gave issues.

## Why?

The line was incorrect.


ftshijt:
This is likely due to an old version of typeguard. Can you try to upgrade Typeguard to the latest version?

sw005320:
I closed this PR as it is related to our recent change in typeguard, which is resolved.

PR #5744, opened 2024-04-17
Add recipe for IWSLT 2024 shared task Indic track

cromz22:
## What?

Add recipe for IWSLT 2024 shared task Indic track

## Why?

There was no recipe

## See also

- IWSLT 2024 Indic track: https://iwslt.org/2024/indic

PR #5748, opened 2024-04-19
Bug fixes for egs2/TEMPLATE/st1

cromz22:
## What?

1. Moved `TEMPLATE/st1/local/images` to `TEMPLATE/st1/TEMPLATE_local/images` and fixed paths in `README.md` (commit f63989b)
2. Changed `TEMPLATE/st1/st.sh` for better handling of packing and uploading (commit 3eef4f1)
3. Added `TEMPLATE/st1/local/TEMPLATE_HF_README_st.md` (commit 8379c54 and 3eef4f1)

## Why?

I've encountered following issues when adding a new ST recipe (#5744):
1. When running `TEMPLATE/st1/setup.sh` for initial setup, the image files only used for `TEMPLATE/st1/README.md` are also copied, which introduces redundant files (cf. commit 40793fe in pr #5744).
2. I see that uploading models to Zenodo is deprecated and uploading them to huggingface is encouraged (cf. [CONTRIBUTING.md](https://github.com/espnet/espnet/blob/master/CONTRIBUTING.md)). Currently, stage 14 is packing, stage 15 is uploading to Zenodo, stage 16 is uploading to huggingface. The problem is that stage 14 and stage 15 are closely tied with the same variable `skip_upload`, making it impossible to run stage 16 after 14, skipping 15. I've separated them by introducing variables `skip_packing` and `skip_upload_zenodo`, so that it becomes possible to run stage 16 after 14 by specifying `--skip_packing false --skip_upload_hf false`. I also improved some error handlings.
3. In stage 16, `TEMPLATE_HF_Readme.md` in the ASR template is used. However, when I run the stage it fails because variable `lang` (ASR language; to add tags of what languages the model can handle in the huggingface model page) used in the template is undefined. I replaced it with `src_lang` and `tgt_lang`, and put the file in `local/TEMPLATE_HF_README_st.md`. (p.s. commit message of 8379c54 is a bit broken, the message inside parentheses was supposed to be "change of ASR template lang to src_lang and tgt_lang")


ftshijt:
Many thanks for your update! The current change looks great to me. 

To make it consistent, it would be super great if you can also adopt the same strategy for the uploading procedure to other tasks (I think most the tasks are following the current strategy in their uploading stages).

ftshijt:
Also, please fix the CI issues when introducing this modification~

cromz22:
Thanks, will do.

> To make it consistent, it would be super great if you can also adopt the same strategy for the uploading procedure to other tasks (I think most the tasks are following the current strategy in their uploading stages).

Can I include this in this pr or should I open a new one?

ftshijt:
> Thanks, will do.
> 
> > To make it consistent, it would be super great if you can also adopt the same strategy for the uploading procedure to other tasks (I think most the tasks are following the current strategy in their uploading stages).
> 
> Can I include this in this pr or should I open a new one?

Since we also care about some ST-specific changes, it might be good to open a following up PR for this activity~ Again, thanks for fixing the issues.

ftshijt:
Many thanks for the update! Since the other tests are canceled ci tests, I can merge the current PR~

PR #5749, opened 2024-04-21
Apply normalization on validation set for CHiME-8 recipe

popcornell:
There was an error for which txt normalization was not applied on validation set during ASR training on C8 recipe. 
This fixes that.


sw005320:
Thanks, @popcornell!

popcornell:
All credits to Kohei Matsuura who pointed out this issue on CHiME slack.

PR #5750, opened 2024-04-22
Fix failing `data.sh` in OWSM

cifkao:
## What?

The`egs2/owsm_v1/s2t1/local/data.sh` script is exiting for every dataset: `./local/prepare_${dataset}.sh || echo "preparing ${dataset} failed"` gets executed first and always evaluates to true, so `exit 1` will always get executed. I grouped the `echo` and the `exit` to fix this.

@pyf98


sw005320:
Thanks!

Issue #5550, opened 2023-11-15
Query regarding whisper.


Naramkeshav59:
Hello,
Is there any way to use whisper model as frontend ?  As I cannot find whisper in s3prl hub.

Naramkeshav59:
what I have tried as config:
```batch_type: numel
batch_bins: 8000000
accum_grad: 4
max_epoch: 10
patience: none
# The initialization method for model parameters
init: xavier_uniform
best_model_criterion:
-   - valid
    - acc
    - max
keep_nbest_models: 3

encoder: conformer
encoder_conf:
    output_size: 512
    attention_heads: 8
    linear_units: 2048
    num_blocks: 12
    dropout_rate: 0.1
    positional_dropout_rate: 0.1
    attention_dropout_rate: 0.1
    input_layer: conv2d2
    normalize_before: true
    macaron_style: true
    pos_enc_layer_type: "rel_pos"
    selfattention_layer_type: "rel_selfattn"
    activation_type: "swish"
    use_cnn_module:  true
    cnn_module_kernel: 31
    interctc_layer_idx: [3, 6, 9,]
    interctc_use_conditioning: true

decoder: transformer
decoder_conf:
    attention_heads: 8
    linear_units: 2048
    num_blocks: 6
    dropout_rate: 0.1
    positional_dropout_rate: 0.1
    self_attention_dropout_rate: 0.1
    src_attention_dropout_rate: 0.1

model_conf:
    ctc_weight: 0.3
    lsm_weight: 0.1
    interctc_weight: 0.5
    length_normalized_loss: false
    extract_feats_in_collect_stats: false

optim: adam
optim_conf:
    lr: 0.0002
scheduler: warmuplr
scheduler_conf:
    warmup_steps: 25000

specaug: specaug
specaug_conf:
    apply_time_warp: true
    time_warp_window: 5
    time_warp_mode: bicubic
    apply_freq_mask: true
    freq_mask_width_range:
    - 0
    - 30
    num_freq_mask: 2
    apply_time_mask: true
    time_mask_width_range:
    - 0
    - 40
    num_time_mask: 2

freeze_param: [
"frontend.upstream"
]

frontend: whisper
frontend_conf:
    whisper_model: "large-v2"  # Note: If the upstream is changed, please change the input_size in the preencoder.
        #path_or_url: /DATA/Sougata/BTech/espnet/egs2/xlsr128minilibrispeech/asr1/large.pt
    download_dir: ./hub
    

[...]

PR #5760, opened 2024-04-24
Fix Whisper frontend

siddhu001:
## What?

Small Fix to Whisper frontend to address bugs likely caused by recent code updates.

## Why?

See https://github.com/espnet/espnet/issues/5550


sw005320:
Thanks!
Mac CI seems to be broken, but it is not related to this PR.
I'll merge this.

PR #5766, opened 2024-04-28
Add torch=2.4.0, 2.3.1, 2.2.2 to CI

kamo-naoyuki:
## What?

I added torch=2.3.0, 2.2.2 to CI


kamo-naoyuki:
@Emrys365 
Hi, do you know the following error? This error happens with pytorch=2.3.0 and DNSMOS. 

https://github.com/espnet/espnet/actions/runs/8868262918/job/24841978720?pr=5766#step:9:471

kamo-naoyuki:
@Emrys365

Could you help me to solve this problem?

Emrys365:
Sorry I missed the message! I will look into it.

ftshijt:
Hi @kamo-naoyuki , may I ask if there are any updates to the new torch version PR? 

I'm working on adding a few more Python versions into CI with the VERSA installation, but it seems that pytorch 3.12 only works for torch 2.2 and above, so I guess it would be great if we can depend on your great effort in this PR

kamo-naoyuki:
Sorry for late reply. I'll restart this update.

@Emrys365 Thank you for the suggestion. I'll try it

@ftshijt Maybe, we need to remove the restrictions for sentencepiece version, https://github.com/espnet/espnet/blob/1c1a12d2c7f613ac59cc996aad8777a8456e9e95/setup.py#L34, for python>=3.11.

https://github.com/espnet/espnet/pull/5532

I think we should use the latest version sentenpiece, but I haven't checked what was changed in sentencepiece. I hope some guys in the know for sentencepiece help us.

PR #5769, opened 2024-04-30
Added partially auto-regressive decoding

Masao-Someki:
## What?

Support partially auto-regressive decoding for ASR and S2T task.
Since this decoding process is based on BERT-CTC, we can only perform speech recognition.

## Why?

To speed up OWSM inference

## See also

https://arxiv.org/abs/2309.14922


Masao-Someki:
With the partially auto-regressive decoding we can speed up OWSM inference twice as fast as Whisper on Google colab. We can use PAR inference with the following code:

```python
from espnet2.bin.s2t_inference import Speech2Text
m = Speech2Text.from_pretrained(
    'espnet/owsm_v3.1_ebf',
    partial_ar=True,
    max_mask_parallel=10
)
```

- Summary
image

- test script
https://colab.research.google.com/drive/1d35QZJ4kHxK2qthYVXfDZc2zxYf5x_5q?usp=sharing

Masao-Someki:
The paper encountered a memory issue, but a minor change in attention ([commit](ee641aa63714ab0135846da2c8fae77d7e87eb5f)) resolved this problem. Additionally, it contributed to speeding up the inference process.

sw005320:
Thanks a lot, @Masao-Someki!

espnet2/bin/s2t_inference.py
```diff
@@ -189,6 +190,10 @@ def __init__(
         lang_sym: str = "<eng>",
         task_sym: str = "<asr>",
         predict_time: bool = False,
+        partial_ar: bool = False,
+        threshold_probability: float = 0.99,
+        max_seq_len: int = 5,
+        max_mask_parallel: int = 5,
```
pyf98, line 196: Can you move these new arguments before this line? In the current design, the arguments below that line can be overwritten in `__call__`, but these new arguments can only be provided in `__init__`.
```
# default values that can be overwritten in __call__
```

pyf98:
Thanks! This speed-up looks very cool.
Since it only supports ASR, can you add some assertions in `s2t_inference` to ensure the task is ASR?

BTW, I didn't read the paper. Why does it only support ASR? Does it rely on CTC? If not, ASR and ST would be similar?

espnet/nets/pytorch_backend/transformer/attention.py
```diff
@@ -53,7 +53,16 @@ def forward_qkv(self, query, key, value):
         n_batch = query.size(0)
         q = self.linear_q(query).view(n_batch, -1, self.h, self.d_k)
         k = self.linear_k(key).view(n_batch, -1, self.h, self.d_k)
-        v = self.linear_v(value).view(n_batch, -1, self.h, self.d_k)
+
+        if expand_value:
+            v_shape = value.shape
+            v = (
+                self.linear_v(value[:1, :, :])
```
pyf98, line 60: What does `expand_value` do? It seems to select only the first sample?
It would be better to add some comments to explain this argument and its purpose.

espnet2/asr/partially_AR_model.py
```diff
@@ -0,0 +1,247 @@
...
+from espnet.nets.beam_search_partially_AR import PartiallyARBeamSearch
+from espnet.nets.scorer_interface import ScorerInterface
+
+if V(torch.__version__) >= V("1.6.0"):
+    from torch.cuda.amp import autocast
+else:
+    # Nothing to do if torch<1.6.0
+    @contextmanager
+    def autocast(enabled=True):
+        yield
```
pyf98, line 27: Not sure if we still need these version checks. PyTorch 1.6.0 is quite old now.

espnet2/asr/partially_AR_model.py
```diff
@@ -0,0 +1,247 @@
...
+    """Mask-CTC-based partially autoregressive inference"""
+
+    def __init__(
+        self,
+        ctc: CTC,
+        decoder: AbsDecoder,
+        threshold_probability: float,
+        sos: int = 4999,
+        eos: int = 4999,
+        mask_token: int = 5000,
```
pyf98, line 40: Do we need these default values? They might depend on the specific model? Can we use None instead?

espnet2/bin/s2t_inference.py
```diff
@@ -269,40 +274,59 @@ def __init__(
...
-            sos=s2t_model.sos,
-            eos=s2t_model.eos,
-            vocab_size=len(token_list),
-            token_list=token_list,
-            pre_beam_score_key=None if ctc_weight == 1.0 else "full",
-            normalize_length=normalize_length,
-        )
+        if partial_ar:
+            beam_search_class = PartiallyARInference
+            beam_search = beam_search_class(
```
pyf98, line 279: We can remove the first line and directly use `PartiallyARInference` in the second line.

espnet2/bin/s2t_inference.py
```diff
@@ -1029,6 +1066,32 @@ def get_parser():
         "If not given, refers from the training args",
     )
 
+    group = parser.add_argument_group("Semi-AR related")
+    group.add_argument(
+        "--partial_ar",
+        action="store_true",
```
pyf98, line 1072: `store_true` is a bit tricky. Why not use `type=str2bool`? Please check the other boolean arguments.

espnet2/bin/asr_inference.py
```diff
@@ -1098,6 +1129,33 @@ def get_parser():
         default=False,
         help="If true, best hypothesis is selected by length-normalized scores",
     )
+
+    group = parser.add_argument_group("Semi-AR related")
+    group.add_argument(
+        "--partial_ar",
+        action="store_true",
```
pyf98, line 1136: Same for this one. Check comment in s2t inference.

pyf98:
Another question: Does this support an extra text prompt (i.e., the text before `<sos>` in OWSM)?

espnet2/bin/s2t_inference.py
```diff
@@ -478,11 +504,14 @@ def _decode_single_sample(self, enc: torch.Tensor):
...
             if isinstance(hyp.yseq, list):
-                token_int = hyp.yseq[:last_pos]
+                token_int = hyp.yseq[start_pos:last_pos]
             else:
-                token_int = hyp.yseq[:last_pos].tolist()
-            token_int = token_int[token_int.index(self.s2t_model.sos) + 1 :]
+                token_int = hyp.yseq[start_pos:last_pos].tolist()
+
+            if not self.partial_ar:
+                token_int = token_int[token_int.index(self.s2t_model.sos) + 1 :]
```
pyf98, line 514: There might be an extra text prompt before `sos`. Is this supported by the partial AR inference?

Masao-Someki:
Thank you @pyf98!

> BTW, I didn't read the paper. Why does it only support ASR? Does it rely on CTC? If not, ASR and ST would be similar?
Also, is there any degradation in performance?

It only supports ASR because PAR uses CTC output as its basis. In this decoding process, we first get the CTC output with the probability for each token, and then apply masking. After that, we use the decoder to predict the masked section in AR manner.
The accuracy relies on several hyperparameters, but basically there is almost no degradation. 
image


> Another question: Does this support an extra text prompt (i.e., the text before <sos> in OWSM)?

It should depend on the case. Considering that the decoder can use the extra prompt but CTC cannot, if we get CTC output with high probability (which means that there is no masks to be predicted with decoder) then PAR do not use extra prompt. If there are any tokens with lower probability, then PAR use decoder with extra prompt to predict the masked part.

espnet/nets/pytorch_backend/transformer/attention.py
Masao-Someki: Thank you, I will add comments. Actually this is very important to reduce memory usage, because the batch size of the PAR might become too large. We compute all beams for all masks in a batched manner, so for example, if we have 10 beam size and 10 masks, we have `batch_size=100`.
So we compute this memory-heavy calculation only one time. In my local PC, applying this trick makes inference faster (3.2sec -> 2.1sec)

espnet2/bin/s2t_inference.py
Masao-Someki: Am I right that those prompts are not generated by the decoder? If so, there will be no extra prompts generated by the `PartiallyARInference` class.
If extra prompts would be generated by decoder then we need some fix.

pyf98: They are not generated by the decoder. They are provided as the `hyp_primer` in the current beam search, just like `sos`.
Then I guess this prompt is not supported in the PartiallyAR interface

pyf98:
> Thank you @pyf98!
> 
> > BTW, I didn't read the paper. Why does it only support ASR? Does it rely on CTC? If not, ASR and ST would be similar?
> > Also, is there any degradation in performance?
> 
> It only supports ASR because PAR uses CTC output as its basis. In this decoding process, we first get the CTC output with the probability for each token, and then apply masking. After that, we use the decoder to predict the masked section in AR manner. The accuracy relies on several hyperparameters, but basically there is almost no degradation. image
> 
> > Another question: Does this support an extra text prompt (i.e., the text before  in OWSM)?
> 
> It should depend on the case. Considering that the decoder can use the extra prompt but CTC cannot, if we get CTC output with high probability (which means that there is no masks to be predicted with decoder) then PAR do not use extra prompt. If there are any tokens with lower probability, then PAR use decoder with extra prompt to predict the masked part.

Thanks for the explanation. I now understand it at a high level.

espnet2/bin/s2t_inference.py
Masao-Someki: Thank you, then we do not need to fix this part, as the output from the beam search class does not contain the extra prompts.
This `token_int` is the same format as Mask-CTC model, so it has mask token at the beginning and end.

Masao-Someki:
@pyf98 I noticed that there were quite a lot of noisy codes I used in my experiments, so I removed everything unrequired. I also added comments and re-formated codes.

espnet/nets/pytorch_backend/transformer/attention.py
pyf98: Thanks for the explanation. Have you added some comments about `expand_value`? Sorry if I missed it somewhere.

espnet/nets/beam_search_partially_AR.py
sw005320, line 1: Can you add a unit test?

espnet/nets/pytorch_backend/transformer/attention.py
sw005320: @Masao-Someki, can you add some comments about `expand_value` in this file?

espnet2/asr/decoder/transformer_decoder.py
sw005320, line 1: ditto.
please add a unit test to check this part.

espnet/nets/pytorch_backend/transformer/attention.py
Masao-Someki: I realized that we need to expand the `key` as well, so I renamed it to `expand_kv`. I added the following comment to clarify this change:
https://github.com/Masao-Someki/espnet/blob/7eeb3c0608b63dd46cb4eaad3990c2d6ba1a948b/espnet/nets/pytorch_backend/transformer/attention.py#L121-L126

Masao-Someki:
I've completed all the reviews and added unit tests!

espnet2/asr/partially_AR_model.py
sw005320, line 1: Do you know why many lines do not go through the test?
Since we created an instance in https://github.com/espnet/espnet/pull/5769/files#diff-17a5ec3fb4fd9d507bd7f2f97f11219d6f47310afb5a80ee31ece48696e193d7R36, I thought that most lines in this file are covered, but actually not.

Masao-Someki: I checked the log file of [unit_test_espnet1_and_espnet2_on_debian11](https://github.com/espnet/espnet/actions/runs/9225728298/job/25383857073?pr=5769#logs), it says we could cover most of the lines:

```
...
 espnet2/asr/partially_AR_model.py    88    3    97%
...
 espnet/nets/beam_search_partially_AR.py    172    15    91%
```
It seems that Codecov hasn't updated its results, likely due to the macos issue in CI.

sw005320:
Thanks, @Masao-Someki!

PR #5771, opened 2024-05-02
Updated results for CHiME-8 DASR baseline with new notsofar1 dev set

popcornell:
Just updated the results on the readme + made gss2lhotse compatible with new lhotse version


egs2/chime8_task1/README.md
```diff
@@ -52,21 +52,24 @@ significantly complicates speaker counting.
...
+| dev | notsofar1    |     0.462402 |    70035 |   151459 |        13116 |       37324 |           19595 |              120 |                5 |              612 |
 +-----+--------------+--------------+----------+----------+--------------+-------------+-----------------+------------------+------------------+------------------+
 ###############################################################################
 ### Macro-Averaged tcpWER for across all Scenario (Ranking Metric) ############
 ###############################################################################
 +-----+--------------+
 |     |   error_rate |
 |-----+--------------|
-| dev |      0.57401 |
+| dev |     0.656558 |
```
sw005320, line 72: I think this number has too many decimal points?

popcornell: must be an issue with tabulate. 
I can fix it

popcornell: I think it is better to leave as it is sorry. 
Because I ve to fix it via `https://github.com/chimechallenge/chime-utils` and then if people do not re-install it they will get slighly different values and be puzzled why they do not correspond. 
It is the output of the scoring script.

popcornell: I know it looks a bit odd

sw005320: ok, that makes sense.

sw005320:
Thanks for the quick update!

PR #5772, opened 2024-05-02
Lhotse support for espnet-ez by class inheritance

popcornell:
@Masao-Someki this is what I have in mind based on our discussion. 

Is it fine ? I will map internally the dataset to the structure of a HF dataset. 
What about multi-channel audio support however ? Will we lose it in this way ? 

Also lhotse supports various augmentations that we will lose in this way. 
It is the simplest way but maybe not as flexible IMO. Do you have other suggestions ?

### DIFF
--- a/espnetez/dataset.py
+++ b/espnetez/dataset.py
@@ -23,3 +23,36 @@ class ESPnetEZDataset(AbsDataset):
 
     def __len__(self) -> int:
         return len(self.dataset)
+
+
+class ESPnetEZLhotseDataset(ESPnetEZDataset):
+    """
+    A simple wrapper class for supporting lhotse manifests in ESPNet-EZ.
+
+    Args:
+        supervisions: lhotse.SupervisionSet, path to the lhotse supervision manifest.
+        recordings: lhotse.RecordingSet, path to the corresponding lhotse recordings manifest.
+    """
+    def __init__(self, supervisions, recordings):
+
+        # lazy importing for lhotse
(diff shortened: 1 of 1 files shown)


Masao-Someki:
Thank you @popcornell, I think it is fine for just using the lhotze as a dataset. And as for the augmentations, you are right, we cannot take its advantage in this way.

I think we might able to use the following code, which I've added to use the custom dataloader, to fully utilize the lhotze library:
https://github.com/espnet/espnet/blob/543f488281e45db3931d24a5366edc5f50167208/espnetez/task.py#L105-L115

We can utilize this feature by passing the dataloader in the definition of Trainer class:
```python
trainer = ez.Trainer(
        task=args.task,
        train_config=finetune_config,
        train_dataloader=<custom dataloader>,
        train_dataloader=<custom dataloader>,
        data_info=data_info,
        output_dir=exp_dir,
        stats_dir=stats_dir,
        ngpu=0,
    )
```

A problem is that the given dataloader will also be used in the `collect_stats` stage. This is not preferable, but now I don't have a good solution. Do you have any idea for this point..?

Masao-Someki:
@popcornell 
> I think current ESPNet for example at least feeds the data that has been speed augmented no ?

You are right, I overlooked this point!! Then I think just feeding the dataloader as above is enough!

> I see that the Trainer takes a data_info argument however.
> Why is that for ?

My apologies for the confusion, this is not necessary!

PR #5785, opened 2024-05-22
Update ci_on_macos.yml (fix CI with latest mac at arm architecture)

ftshijt:
## What?

CI fix for MacOS

### DIFF
--- a/.github/workflows/ci_on_macos.yml
+++ b/.github/workflows/ci_on_macos.yml
@@ -26,7 +26,6 @@ jobs:
       - uses: actions/setup-python@v4
         with:
           python-version: ${{ matrix.python-version }}
-          architecture: 'x64'
       - name: Get PR labels
         id: pr-labels
         uses: joerick/pr-labels-action@v1.0.9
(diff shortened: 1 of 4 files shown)


ftshijt:
FYI: https://github.com/actions/setup-python/issues/808

ftshijt:
Umm, it is not as easy as I thought since some c compiling is also having issues with M1 arm...

ftshijt:
Hi guys @sw005320 @kamo-naoyuki , I'm trying to fix the CI issue with MacOS but am not sure about the next move. Would be great if you have any comments:

The current issue is mostly from the change of MacOS-latest version being switched to M1 chip with arm architecture. After I change the ci test to arm architecture, two external tools start to have issues:
- We can use 2.4.12 version of SCTK to fix the issue (thanks to https://github.com/usnistgov/SCTK/commit/6eebc2cc7799e76c53850abf17600114c901903b ).
- The Sph2pipe however, does not have an upgrade option and the exact issue from our CI has not been fixed since this Feb https://github.com/robd003/sph2pipe/issues/3 )

For the following fixes, we may think of removing sph2pipe from the default installation for MacOS test specifically (we still need it for other tests) or use our own fixed fork of the sph2pipe instead of their old release. May I ask which one you would prefer?

sw005320:
Hi @ftshijt,
Thanks for taking care of this!

My preference is

> For the following fixes, we may think of removing sph2pipe from the default installation for MacOS test specifically (we still need it for other tests)

We may need this legacy format in the Linux environment, but we don't need it for MacOS.

PR #5788, opened 2024-05-25
Import Espnet in-house codec to downstream task

jctian98:
This PR imports our codec model to the downstream tasks. Specifically, it provides the tokenization script for codec inference.

TODO:
- [x] inference with some unwanted behavior. Bug fix
- [ ] Support Batch inference - low priority, will not be supported in this PR.

@ftshijt I think it's ready for merging into the codec branch.


ftshijt:
For more information, on the unintended issue (especially for other people using the tool), we detected that the issue is because of a pytorch version. The final model should be tested in either torch<2.1 or torch>=2.1 for consistency because of an API change in the torch side.

ftshijt:
Thanks for the update! LGTM

PR #5741, opened 2024-04-10
[SVS] VISinger Plus

jerryuhoo:
## What?

Use pre-trained SSL models as additional input information for VISinger 2.

## Why?

SVS enhancement.

## See also

@ftshijt

PR #5789, opened 2024-05-27
Revert change to abs_task to keep the consistency behavior

ftshijt:
The change was initially conducted in a previous PR at https://github.com/espnet/espnet/pull/5741 but it breaks the compatibility of the loading policy. To keep it unchanged, we revert back to the original config.


ftshijt:
SInce it is a reversion, directly merge to master

PR #5791, opened 2024-05-27
version 1 speechtokenizer

massabaali7:
## What?

Adding the speech tokenizer. 

## Why?

For the codec project


espnet2/gan_codec/encodec/seanet_st.py
```diff
@@ -0,0 +1,438 @@
+# Adapted from https://github.com/facebookresearch/encodec
```
ftshijt, line 1: For adding speechtokenizer, could you please limit the change to either shared folder or speechtokenizer?

ftshijt: Also, may I know what is the difference between this seanet and the existing one? If there is no significant changes, you can directly use the current seanet in the shared folder.

espnet2/gan_codec/shared/quantizer/residual_vq_st.py
```diff
@@ -0,0 +1,129 @@
+# Copyright 2024 Jiatong Shi
```
ftshijt, line 1: Ditto. What is the difference between this one and the existing residual_vq?

espnet2/gan_codec/speechtokenizer/speechtokenizer.py
```diff
@@ -0,0 +1,801 @@
...
+            codecs (torch.Tensor): neural codecs in shape ().
+        Returns:
+            torch.Tensor: resynthesized audio.
+        """
+        quantized = self.quantizer.decode(codes)
+        resyn_audio = self.decoder(quantized)
+        return resyn_audio
+
+
+class SpeechTokenizerDiscriminator(nn.Module):
```
ftshijt, line 712: You seem to resue the soundstream discriminator but it does not match the speech tokenizer paper. Could you please double check?

espnet2/gan_codec/speechtokenizer/speechtokenizer.py
```diff
@@ -0,0 +1,801 @@
...
+        max_idx = len(self.target_bandwidths) - 1
+
+        # randomly pick up one bandwidth
+        bw = self.target_bandwidths[random.randint(0, max_idx)]
+        
+        # Forward quantizer
+        quantized, _, _, commit_loss, quantized_list = self.quantizer(encoder_out, self.frame_rate, bw, layers=layers)
+        
+        feature = rearrange(quantized_list[0], 'b d t -> b t d')
+        feature = self.transform(feature)
```
ftshijt, line 633: where is the distillation loss?

ftshijt (commented):
Thanks for your contribution. For my understanding, the major problem for speechtokenizer is to introduce SSL (semantic) features to the model distillation training. May I ask where did you manage to input the semantic features?

PR #5795, opened 2024-05-30
Discrete TTS with Monotonic attention search via VAE

ftshijt:
The implementation of discrete TTS with monotonic attention search via VAE (with the new model, we can skip the need of duration).


ftshijt:
@wyh2000 @jctian98 Please help review the PR~ I will add a unit test for this model soon.

PR #5797, opened 2024-06-04
fix spk mfa-conformer forwarding

series2:
## What?

Fix spk mfa-conformer forwarding

## Why?

The current implementation appears to propagate only one layer of the model.

### DIFF
--- a/espnet2/spk/encoder/conformer_encoder.py
+++ b/espnet2/spk/encoder/conformer_encoder.py
@@ -312,8 +312,8 @@ class MfaConformerEncoder(AbsEncoder):
 
         intermediate_outs = []
         for layer_idx, encoder_layer in enumerate(self.encoders):
-            x, _ = encoder_layer(xs_pad, masks)
-            intermediate_outs.append(x[0])
+            xs_pad, _ = encoder_layer(xs_pad, masks)
+            intermediate_outs.append(xs_pad[0])
 
         x = torch.cat(intermediate_outs, dim=-1)
         x = self.ln(x).transpose(1, 2)  # (#batch, L, output_size)
--- a/espnet2/spk/encoder/conformer_encoder.py
+++ b/espnet2/spk/encoder/conformer_encoder.py
@@ -304,7 +304,7 @@ class MfaConformerEncoder(AbsEncoder):
             or isinstance(self.embed, Conv2dSubsampling6)
             or isinstance(self.embed, Conv2dSubsampling8)
         ):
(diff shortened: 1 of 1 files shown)


sw005320:
Thanks for the report!
@Jungjee, can you confirm it?

Jungjee:
Thank you, @series2 for detecting this error.
@sw005320, yes, this fix is correct. I will also check whether this issue also exists in other models.

Jungjee:
Branchformer and E-branchformer is okay. 

@series2 , to be consistent with other code, do you mind reverting this change and instead change line 307 to 
`x, _ = self.embed(x, masks)`?
I believe the effect should be the same.

series2:
Thanks for the quick reply.
Either one is fine with me.

I changed the code to xs_pad using the following url as a reference.
https://github.com/espnet/espnet/blob/master/espnet2/asr/encoder/conformer_encoder.py

series2:
@Jungjee
I have modified the code. Could you please confirm it?

And a question.
This modification could increase the score even more, but I would like to know if you have any idea why the [voxceleb README](https://github.com/espnet/espnet/blob/master/egs2/voxceleb/spk1/README.md) results don't look so bad even with one layer of Conformer.

Jungjee:
@series2, LGTM!

> I would like to know if you have any idea why the [voxceleb README](https://github.com/espnet/espnet/blob/master/egs2/voxceleb/spk1/README.md) results don't look so bad even with one layer of Conformer

I was also surprised with the results considering this bug. I am running another MFA-Conformer experiment to check the performance difference.

series2:
Thank you for your thoughts.
If you have an experiment that reflects the change, Could you publish it in the official espnet repository, e.g., README under egs2/?

sw005320:
Thanks, @series2!

@Jungjee, please make a PR once you finish the experiments and update a model!

Jungjee:
@sw005320, @series2   since the bug is already fixed, ESPnet code-wise, I couldn't find something to update.
I updated the huggingface model. The new EER is 0.78% (previous: 0.86%).

sw005320:
@Jungjee, do you mean that you also updated https://github.com/espnet/espnet/blob/master/egs2/voxceleb/spk1/README.md?

Jungjee:
@sw005320 oh, right, I didn't update that one. I will open a PR to update the `readme.md`. Please let me know if there is something else I'm missing.

sw005320:
OK, if you have other models that have not been updated, please do so.
The file was not updated for a long time.

PR #5803, opened 2024-06-06
Update specaug.py

aky15:
## What?

Support configuration and control of the methods of applying time and freq mask, i.e. masking with 0 or masking with the mean value. Masking with 0 by default.

## Why?

Helps to switch the methods of applying time and freq mask.

### DIFF
--- a/espnet2/asr/specaug/specaug.py
+++ b/espnet2/asr/specaug/specaug.py
@@ -33,6 +33,7 @@ class SpecAug(AbsSpecAug):
         time_mask_width_range: Optional[Union[int, Sequence[int]]] = None,
         time_mask_width_ratio_range: Optional[Union[float, Sequence[float]]] = None,
         num_time_mask: int = 2,
+        replace_with_zero: bool = True,
     ):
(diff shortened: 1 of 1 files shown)


sw005320:
Thanks.
I see that you have prepared the flag.
Will you prepare the option argument to replace the value?

aky15:
Yes, I add the default value True to make 'replace_with_zero' an option argument. @sw005320 
Masking with 0 or by mean value has already been implemented in [mask_along_axis function](https://github.com/espnet/espnet/blob/1bbdf9626efae145988911290d69ddfeadf6c949/espnet2/layers/mask_along_axis.py#L78). However, the flag is missed in the SpecAug so I added it.

sw005320:
Oh, I see.
Do you mean that the default behavior would be changed with this?

@pyf98, can you check this PR?

pyf98:
Thanks for adding the option. @aky15 I guess it won't change the default behavior? If this is true, then it looks good.

aky15:
Yes, it won't change the default behavior. @sw005320 @pyf98

sw005320:
OK, thanks!

PR #5804, opened 2024-06-06
[WIP] Codec ASR2 feature and LibriSpeech recipe

jctian98:
As a downstream task of the codec project, this PR allows asr2 template to support codec tokens as the input. It also contains a LibriSpeech-960hrs recipe. 

Experiments are running. Will update the results here soon.
@ftshijt


ftshijt:
Thanks for the update!

PR #5826, opened 2024-06-29
update MFA-Conformer performance after fixing the bug in #5797

Jungjee:
## What?

There was a bug in the MFA-Conformer model's architecture which was fixed in #5797.
This PR updates the performance of the re-trained MFA-Conformer. 
The model is available in Huggingface.


sw005320:
Thanks!

series2:
@Jungjee 
Thanks for the update the model.
I am currently trying to reproduce it.
However, I am having trouble because some of the published [config](https://huggingface.co/espnet/voxcelebs12_mfaconformer_mel/blob/main/data/user_data/jeeweonj/espnet_jee/egs2/voxceleb/spk1/exp/spk_train_mfa_conformer_adamw_fixEncoderBug2_raw_sp/config.yaml) are different from the config output when I experimented with `conf/tuning/train_mfa_conformer_adamw.yaml`.
Also, when I run the experiment with the latest state of the master branch, the eer.png file shows a tendency to overlearn at around 4 epochs, which was not seen before the fix.

Can you please release `config:conf/tuning/train_mfa_conformer_adamw_fixEncoderBug2.yaml`?
It would be great if you could release a reproducible environment, even if it is in the `Jungjee/espnet` repository.

Your kind response is always helpful.

series2:
Thank you!

PR #5828, opened 2024-07-03
CI Fix - Updated dependencies

Fhrozen:
## What?

Update dependencies

## Why?

Current dependencies are generating CI issues.

## See also

@sw005320 I tried to fix using a new version of fairseq. The CI crashing issue is cause due to the Omegaconf version mismatch. Faiseq uses  <2.1, while s3prl needs > 2.1.
Currently, seems that fairseq has not plan to update its dependencies any time soon. So I have to possible fixes:
- Force s3prl to use a older version (however, not sure if this will be a good long time solution and new models may not be supported).
- Move only the code from fairseq employed into ESPnet or a ESPnet related repo (a new one). Take in consideration that fairseq last version was released in 2022, and just small fixes were applied.

Please let me know which possible solution could be the best (other options are welcome).


sw005320:
OK.
The second option sounds better, but how much work do we need?

Fhrozen:
I observed that Huber, AVHuber, and Wav2Vec are only employed. Maybe additionally adding WavLM, which seems to require few files to copy. I will generate a folder `espnet2/fairseq` which will contain the files and remove the fairseq installation/requirement. In any case, I will try to test an application for converting faiseq native models into jit script, so it will be possible to import any model.
Please, let me know if there is any question.

adding @simpleoier for comments.

Fhrozen:
@sw005320, Regarding Centos7 CI: 

> CentOS 7 reached its end of life on June 30, 2024. As of July 1st, 2024, CentOS will no longer receive official updates, patches or support. 

https://www.suse.com/c/centos-7-support-for-post-eol-security-and-stability/#:~:text=CentOS%207%20reached%20its%20end,comprehensive%20guide%20to%20CentOS%20EOL.

The github action@checkout cannot find also the docker container base/7/x86_64 due to the EOF.
Should I remove the CI by the moment? maybe we can build our own container, but if EOF, not sure if it will be required.

Fhrozen:
@sw005320 the ci tests are fixed.
- A new repo with the only necessary files for fairseq are located at espnet/fairseq. The CI shows no problems, but in particular cases where ppl start using functions that are not cover by the CI tests will reveal possible missing functions/classes on fairseq.
- Centos7 tests were removed due to EOL support. Please, let me know if this was ok.

sw005320:
Great. @Fhrozen!
This fixes our long-term issues due to fairseq!

PR #5830, opened 2024-07-05
A fix to newer torch version (compatible to old version with typecheck)

ftshijt:
## What?

With the old setting, there will be a typeguard error from error_queue. This is an essential fix for torch>=2.2

## Why?

Follow the existing spawn implementation

## See also

https://github.com/pytorch/pytorch/blob/main/torch/multiprocessing/spawn.py


ftshijt:
The CI issue is not related to the updates, will wait https://github.com/espnet/espnet/pull/5828 to merge first

sw005320:
Thanks, @ftshijt!

PR #5834, opened 2024-07-09
Fixed kising/acesinger google drive download

HANJionghao:
## What?

Resolved issues with the KiSing/ACE-Opencpop (acesinger) Google Drive download.

## Why?

Previous download commands were no longer functional.


sw005320:
Thanks!

PR #5537, opened 2023-11-07
SSL w/o torchaudio dependency

wanchichen:
## What?

This PR allows for HuBERT pre-training w/o using torchaudio, allowing for more customization and use of different ESPnet components. It also introduces some tricks to better support large-scale training. 

Features:
- Flash Attention (by @pyf98)
- Activation checkpointing in E-Branchformer
- Convolutional Feature Extractor as frontend
- Convolutional positional embeddings
- Efficient batch sampler for large-scale
- Efficient multi-shard iterator for large-scale
- WavLM noise augmentation
- More efficient distributed training
- Recommended tunable DDP args for multi-node training
- Cross-Entropy loss for HuBERT SSL
- Intermediate losses for HuBERT SSL
- HuBERT SSL using filterbank input features
- More detailed GPU memory reporting

Supports only HuBERT for Transformer and E-Branchformer so far. 

To do:

- [ ] Fbank training configs
- [ ] Conformer, branchformer encoder
- [ ] check in `hubert.sh` which implementation to use

PR #5839, opened 2024-07-11
Add flash-attn

wanchichen:
## What?

Support flash attention in Transformer-based encoders and decoders.

NOTE: I have no idea how we can have CI tests since flash-attn requires specific hardware

Huggingface simply skips the flash-attention tests if the required GPUs are not available: https://github.com/huggingface/transformers/pull/25598

Breaking apart features from

https://github.com/espnet/espnet/pull/5537

Original implementation by @pyf98 .


sw005320:
@Fhrozen fixed it (Super thanks, @Fhrozen!).

sw005320:

sw005320:
@pyf98, this is a reminder.

pyf98:
Looks good. @wanchichen Is there any part different from my original code?

sw005320:
- is there any error handling when we do not install the flash attention?
- can you add an example (e.g., config?)?

sw005320:
@wanchichen, any follow up?

wanchichen:
I will add a config for librispeech 100

tools/installers/install_flash_attn.sh
sw005320, line 1: Can we use flash attention installation default while it would be skipped if the environment does not support it?

wanchichen: Sure! Should we also specify flash attention as the default implementation then?

sw005320: Yeah, that's my preference if we can easily handle the case that the user's environment cannot support it.

sw005320:
Can you finish this PR?
Also, please add this new function to the top README.md

wanchichen:
I have added flash attention to the default espnet install. It will be skipped if the system requirements are not met (such as in the CI).

In the API, flash attention will be the default if the system supports it, unless the users manually disables it. 

I am not sure what is the cause of the failing CI tests though

sw005320:
Thanks, @wanchichen!

PR #5842, opened 2024-07-17
Implementation of EVA: Robust Audiovisual Speech Recognition Models with Mixture-of-Experts

wyh2000:
## What?

This is an implementation of EVA: Robust Audiovisual Speech Recognition Models with Mixture-of-Experts. It supports audiovisual ASR for unconstrained videos.
EVA implementation is based on OWSM v3.1, with additional visual information as the encoder input.

## See also

[AVFormer](https://openaccess.thecvf.com/content/CVPR2023/papers/Seo_AVFormer_Injecting_Vision_Into_Frozen_Speech_Models_for_Zero-Shot_AV-ASR_CVPR_2023_paper.pdf)
[AVATAR](https://gabeur.github.io/avatar-visspeech)


sw005320:
Thanks!
I have some high-level comments.
1. @pyf98, is it OK to have this kind of extension in the s2t template? Another option is to create a new template with the visual information. 
2. I'm not sure that we can replace `s2t_train` with `vis_s2t_train` in https://github.com/espnet/espnet/pull/5842/files#diff-bc3fb50d902f9d1e761ffb2c1244a4adab1825956d23235b22722e7f353d2398R1239
3. can you add a test?
4. `vis_s2t`does not follow the naming convention. `s2t_vis` would be better.

pyf98:
Thanks for the great work! @wyh2000 

It may be better to have a separate template for visual speech-to-text (maybe `vs2t`?) Vision-speech multi-modal models can have good potential in the future. A separate template is more flexible to extend and maintain. For example, you may need additional preprocessing stages or even training stages, which are not directly supported in `s2t` template. 

Also, the current `s2t` task has been used in many of our applications. We'd better not change the interface too much.

Since the current work is mainly based on `s2t`, we can reuse many scripts and model definitions. The extra work of making a template should not be too much.

PR #5847, opened 2024-07-24
[WIP] New Homepage for ESPnet

Masao-Someki:
The new homepage proposal with Sphinx and VuePress! 
You can see the demo-page from the following link: https://masao-someki.github.io/espnet_draft_home_page/

| Landing page | API Documents | Notebooks | Search box |
| --- | --- | --- | --- |
|  image  |  image |   image | image |

## What?

The `./ci/doc.sh` script will automatically generate the HTML files for the documentation. Performance has been optimized, as evidenced by the Lighthouse results.

| from PC | from mobile |
| --- | --- |
| image | image |

## Why?

The current documentation is difficult to integrate with the Notebook repository because Sphinx generates docstrings in HTML format, while the notebooks are in Jupyter Notebook format. Additionally, changing the directory structure is challenging, leading to inefficient maintenance.

To address this, we convert all documents into Markdown format and generate the HTML files using VuePress. With VuePress, we can define the page structure based on the directory structure of each Markdown file, eliminating the need to create a TOC or specific reStructuredText (RST) files. This approach also allows us to easily switch to other static site generators, such as [MkDocs](https://www.mkdocs.org/) and [Hugo](https://gohugo.io/).

## ToDO
- [x] Fix bug in script
- [x] Add link or button to go to source of the package


juice500ml:
Currently struggling while building this. It seems that the current version of sphinx is severely outdated, and we should consider upgrading it. For example, recommonmark is now deprecated. https://github.com/readthedocs/recommonmark/issues/221

sw005320:
> Currently struggling while building this. It seems that the current version of sphinx is severely outdated, and we should consider upgrading it. For example, recommonmark is now deprecated. [readthedocs/recommonmark#221](https://github.com/readthedocs/recommonmark/issues/221)

It sounds good to me to update Sphinx.
But Sphinx usually has complicated dependencies, so we fixed the version.
So, I hope it will not bring further complications.

Masao-Someki:
I apologize for the delayed response, could you please specify where the problem occurs? Is it related to Sphinx?

> It seems that the current version of sphinx is severely outdated, and we should consider upgrading it. For example, recommonmark is now deprecated. https://github.com/readthedocs/recommonmark/issues/221

Thank you, I will look into it.

juice500ml:
Yes, I'm a bit worried about the implications. I'd usually avoid updating too much... If you use Python 3.10, I think that problem persists. Let me try downgrading Python and run it again. Actually, nearly all the steps for espnet/ci/doc.sh raises issues for me. Possibly because I only installed `espnet[doc]` via pypi on a barebone Python?

Masao-Someki:
Would you reinstall ESPnet with all the necessary dependencies, such as `espnet[doc]` or `espnet[recipe]`, and run the command again? If you encounter any issues with the updated ESPnet, please let me know…

sw005320:
Can you fix the CI errors?

Masao-Someki:
Thank you, @juice500ml! The CI for document generation is now working successfully!

sw005320:
We still have several CI errors.
Please fix them

sw005320:
Did you try it in your local environment?
It would be easier to fix the error than to ask CI often.

Masao-Someki:
I'm really sorry about this, but when I tested it in my local environment, I encountered another error. It seems like it might take a while to resolve, so I'm running the jobs now...

sw005320:
I see. Make sense.
But since this is a CPU job, we can build a virtual Linux environment in your terminal PC and create exactly the same setup by following https://github.com/espnet/espnet/tree/master/.github/workflows

Masao-Someki:
@juice500ml

It looks like the `gtn` package wasn't installed during the `install espnet` process in the CI log from #5766. The `gtn` seems to be included in the current CI because the `all` dependency group is added to the `doc` group:

https://github.com/Masao-Someki/espnet/blob/386b291df6e4db4272bddddb19a4ca4877186950/setup.py#L123

```
requirements["doc"].extend(requirements["all"])
```

I think we can resolve the CI issues by removing this. Could you please confirm if it's okay to revert this line?

juice500ml:
I'm pretty sure that CI will work now. I'm planning to make another PR (within today) that makes updates on the doc design, etc. based on the feedbacks from other people. I don't want to make the PR to be bigger than this :)

doc/usage2rst.sh
```diff
@@ -8,6 +8,8 @@ if [ $1 == "--help" ]; then
 fi
 
 real=$(realpath $1)
+# githash=`git rev-parse HEAD`
+githash=8574aa8e0d5968916df1df42c28621a48b2a5281
```
sw005320, line 12: what's this?
is it OK to fix the hash?

sw005320 (commented):
- Thanks. I agree that the PR should not be bigger than this. I'll just wait for the next PR to check the content, but anyway, can @Masao-Someki and @juice500ml to cross-review this PR again to make sure?
- I'm just concerned that `setup.py` specifies the version or restricts the version. If they are necessary, it's OK

doc/usage2rst.sh
Masao-Someki: @sw005320 
Sorry, this is a debug code! I will check all diffs and fix debug lines!!

Masao-Someki:
Regarding the modifications to `setup.py`:

- We restricted `setuptools` to `<74.0.0` because the latest version (`74.0.0`) removed a functionality related to `distutils`, which would cause an installation error.
- We upgraded Sphinx to the latest version, adding a version restriction to prevent potential CI errors in case of a major Sphinx update.
- Due to the Sphinx upgrade, `recommonmark` was deprecated. The recommended replacement for Markdown support is `myst-parser`.
- We've added a Jupyter notebook as a demo. To convert `.ipynb` files to Markdown, we now use `jupyterlab`.
- We are using `sphinx-markdown-builder` to build Markdown files from RST files.

juice500ml:
I reviewed the PR & refactored some minor stuff. I think it's okay now, so let's wait for the checks to finish!

sw005320:
OK
After CI is fixed, I'll merge this PR.

PR #5851, opened 2024-07-29
fix bugs in SVS1

South-Twilight:
## What?
1. fix recipes: itako, kiritan, namine_ritsu_utagoe_db, natsume
    - fix problems in data_preparation stage.
2. update README.md to make it easier to understand
    - update run command
3. fix some config setting in opencpop
    - duration dim and some params in run.sh
## Why?

fix bugs as shown above


sw005320:
Do these PRs change the results?
(I guess so, since you changed some configurations).
In this case, can you also update the results and models?

South-Twilight:
> Do these PRs change the results? (I guess so, since you changed some configurations). In this case, can you also update the results and models?

These PRs will not change the model results. Previously, when running these models, we modified the model parameters to those in the current PR to adapt the specific recipe. This time, we have changed these settings to be the default settings.

sw005320:
Got it.
In this case, after the review by @ftshijt, I'll merge this PR.

espnet2/fileio/score_scp.py
```diff
@@ -58,7 +58,16 @@ def __init__(
...
-        tempo = int(m[0][2].number)
+
+        # NOTE(Yuxun): tempo with '(playback only)' returns None in item[2].number
+        tempo = None
+        for item in m:
+            tempo_number = item[2].number
+            if tempo_number is not None:
+                tempo = tempo_number
+                break
+        tempo = int(tempo)
```
ftshijt, line 69: Is it okay to use only one tempo for all music score object?

egs2/namine_ritsu_utagoe_db/svs1/conf/decode.yaml
```diff
@@ -1,8 +0,0 @@
-# This configuration is the decoding setting for FastSpeech or FastSpeech2.
```
ftshijt, line 1: I think we default would need the conf/deode.yaml for decoding stage?

egs2/itako/svs1/local/prep_segments_from_xml.py
```diff
@@ -99,20 +99,21 @@ def make_segment(file_id, tempo, notes, threshold, sil=["P", "B"]):
                 segment = SegInfo()
         # remove rest note
         if "itako04" in file_id:
-            if i in [191, 193, 194, 195, 197, 199, 200, 201, 202, 203, 207]:
```
ftshijt, line 102: Given their future potential data updates, how about we explicitly note the version of the dataset (if no specific version number if provided, let's simply use date) in README for user's reference?

ftshijt (commented):
Thanks for the update! I've left my comments as follows:

egs2/namine_ritsu_utagoe_db/svs1/svs.sh
sw005320, line 1: why it's here instead of pointing out `svs.sh` in TEMPLATE?

ftshijt: Oh, yeah, somehow my comment for this one is not shown up correctly. It should not be duplicated.

ftshijt:
Thanks for the update! LGTM now

sw005320:
Thanks, @South-Twilight!

PR #5856, opened 2024-08-02
Add DeepSpeed trainer for large-scale training

jctian98:
## What?

This PR adds another trainer object that warp the [DeepSpeed](https://github.com/microsoft/DeepSpeed) so that it will automatically handle many trainer-related things, especially some advanced features:
  * Tensor partition, with zero-{1,2,3} strategy
  * Activation checkpointing
  * Parameter offloading
  * Distributed checkpoint saving / loading
  * full BF16 training
  * on-cpu optimizer
  * etc.

(1) This DeepSpeed trainer is based on the data parallelism and will allow us to train model as large as ~13B. 
(2) It can be smoothly switched from previous ESPnet trainer. 
(3) Unlike model parallelism (will be needed when >13B), this trainer doesn't pose any requirement to the model architecture.

To use it, simply change add these lines in training config:
`use_deepspeed: true`
`deepspeed_config: <path-to-config>.json`
Most trainer-related options with be moved to `deepspeed_config`, so that the training config will only need to define things like model arch and data loader.

Discussion:
what is the good place to add a README.md file? or is it needed?

Will request @wanchichen  to take a look.

Thanks


sw005320:
@wanchichen, can you review this PR?

espnet2/train/deepspeed_trainer.py
```diff
@@ -0,0 +1,273 @@
...
+                log_interval = 100
+
+        for iiter, (utt_id, batch) in enumerate(
+            reporter.measure_iter_time(iterator, "iter_time"), 1
+        ):
+            assert isinstance(batch, dict), type(batch)
+
+            with reporter.measure_time("step_time"):
+                # (0) ensure all ranks have not finished.
+                dist.all_reduce(iterator_stop, ReduceOp.SUM)
```
wanchichen, line 185: This all-reduce is not necessary (same with the one in `valid_one_epoch`). We can remove lines 185-187

wanchichen:
Great work! The implementation is very clean, so I do not have many comments. Only two questions:

- Does full bf16 training work with models that use conv. layers?
- Is zero-3 tested? We might also be able to remove the logging all reduce (https://github.com/espnet/espnet/blob/b49ba36eb8b63dc7383a68c23b1757e7e4aa59cf/espnet2/train/deepspeed_trainer.py#L196), since I don't think all devices will have unique metrics in the zero3 case, which makes the operation very wasteful.

jctian98:
@wanchichen  Many thanks for the review.
We can definitely remove the all-reduce for logging, I'm just curious about why `I don't think all devices will have unique metrics in the zero3 case`. I think it should be identical to DDP, could you explain a bit? after that we can simply remove these lines or make them optional. 

I tested with Zero3 but not all of its functions. After this PR is merged, we can later add some examples with zero2 and zero3.

jctian98:
@wanchichen  Thanks for the review!
* With BF16 and the zero2 config I posted, it can train a conformer model without errors / warnings.
* I have tried zero3 and it can work - but slower than zero2 maybe be due to heavier cpu-gpu communication. We can further explore zero3 config, but maybe not need to change the code.
* Could you explain a bit why `I don't think all devices will have unique metrics in the zero3 case`? is there anything that is different from standard DDP (in terms of computing results)?

espnet2/bin/spk_embed_extract.py
sw005320, line 1: do we need this?
you removed this in the other part of the latest commit.

espnet2/tasks/abs_task.py
sw005320, line 1: ditto
do we need this?
you removed this in the other part in the latest commit.

sw005320:
- Do you have some example usages (e.g., config)?
- It's better to mention the DeepSpeed support in the top README.md

sw005320:
Sounds good.
Is it everything, @jctian98?
If it is ready, I'll merge this PR.

jctian98:
> Sounds good. Is it everything, @jctian98? If it is ready, I'll merge this PR.

We may still discuss with @wanchichen about if we should remove the `all_reduce`s in the training loop to avoid communication overhaed. But we can merge for now, and revise it later if necessary.

egs2/an4/asr1/conf/deepspeed_zero2.json
```diff
@@ -0,0 +1,47 @@
...
+      "allgather_bucket_size": 5e8
+    },
+    "zero_optimization": {
+      "stage": 2,
+      "contiguous_gradients": true,
+      "overlap_comm": true,
+      "reduce_scatter": true,
+      "reduce_bucket_size": 5e8,
+      "allgather_bucket_size": 5e8
+    },
```
xingchensong, line 23: redundant configuration?

egs2/an4/asr1/conf/deepspeed_zero2.json
```diff
@@ -0,0 +1,47 @@
+{
+    "train_batch_size": 32,
```
xingchensong, line 2: `train_batch_size` (32 in this case) might be conflict with `train.yaml::batch_size` (64 in this case)

espnet2/train/deepspeed_trainer.py
xingchensong: I think that forced synchronization is necessary, as it functions like a join operation, allowing all ranks to start forwarding simultaneously and avoiding timeouts. Torch_DDP does not require this synchronization operation, partly because DDP has its own join function, and partly because, for distributed training like DeepSpeed, the communication between machines is more complex and timeouts are more likely to occur. Therefore, sacrificing some waiting time (in fact, I believe the proportion of waiting time will not be very high) in exchange for training stability is necessary.

jctian98:
@xingchensong Many thanks for the review!!! I have updated the example config. 

Current example is simple a toy. Later when we run some experiments at large scale we can share our deepspeed config together with the recipes.

@sw005320  if CI is ok, I think this PR is ready to merge.

espnet2/train/deepspeed_trainer.py
jctian98: Thanks for the both comments :)
Sice the ZeRO will always have extensive communication, maybe this force sync is ok to use. I'll post more observation if I find these lines are the performance bottleneck.

PR #5857, opened 2024-08-02
Support Codec feature for TTS2 task

wyh2000:
## What?

To support discrete TTS with codec features.

## Why?

We add parallel decoder layers and final projection layers to support codec predict based on FastSpeech 2 architecture.


ftshijt:
LGTM! Could you fix the CI test?

For the design, may I ask why you choose to predict the different stream with separate decoders instead of the the same one with different prediction head?

wyh2000:
> LGTM! Could you fix the CI test?
> 
> For the design, may I ask why you choose to predict the different stream with separate decoders instead of the the same one with different prediction head?

Thank @ftshijt  for checking the code. I fixed the CI test. 
From the experiment results, using parallel decoders looks better than parallel prediction heads. So I chose parallel decoders here.

ftshijt:
Many thanks for the update! You may consider adding CI test for the new modules later.

PR #5858, opened 2024-08-05
Fix bug in asr2.sh

juice500ml:
## What?
- Typo-style bug fixes in asr2.sh template recipe

## Why?
- The current version does not work for librispeech-100 recipe.


juice500ml:
Renamed the branch, hence the PR got automatically closed.

PR #5859, opened 2024-08-05
Fix bug in asr2.sh

juice500ml:
## What?
- Typo-style bug fixes in asr2.sh template recipe

## Why?
- The current version does not work for librispeech-100 recipe.

## See also
- After this fix, we reproduced similar librispeech-100 results. (Left: ours, Right: original README) Actually, the WER/CER is slightly better, so we might want to update the numbers + commit hashes.
image

### DIFF
--- a/egs2/TEMPLATE/asr2/asr2.sh
+++ b/egs2/TEMPLATE/asr2/asr2.sh
@@ -819,7 +819,7 @@ if [ ${stage} -le 5 ] && [ ${stop_stage} -ge 5 ] && ! [[ " ${skip_stages} " =~ [
 
         _suf=
         if [ -n "${layer}" ]; then
-            _suf="layer${layer}/"
+            _suf="layer${layer}"
         fi
 
         if [ "${src_case}" = ts ]; then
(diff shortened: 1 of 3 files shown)


juice500ml:
@TomohikoNakamura It's a bit odd to see that TER's #wrd is different. Any idea why?

sw005320:
thanks, @juice500ml.
Did you confirm it is working?
If so, I'll merge this PR.

juice500ml:
Yesterday I realized that the branch was somewhat stale, so I'm currently running again, just to make sure! I'll comment here again after check is done.

juice500ml:
Ran another run with the updated branch, and additionally found that `path.sh` was missing. Added that in 366cee4, and it looks good now!

sw005320:
> Ran another run with the updated branch, and additionally found that `path.sh` was missing. Added that in [366cee4](https://github.com/espnet/espnet/commit/366cee49359d47dce859bb4b24e3c8c206dd96f2), and it looks good now!

Thanks!
After the CI test, I'll merge this PR.

sw005320:
Thanks!

PR #5862, opened 2024-08-05
Codec CI and VERSA installaion

ftshijt:
## What?

Codec CI related and VERSA installation setup


ftshijt:
I'm not pretty confident on the better hyper-params for funcodec. @wyh2000 could you please also check to see if we want to add more for other codec domains? Thanks ahead

sw005320:
Please fix the CI error (package dependency)

ftshijt:
@sw005320 I think the current issue is that python 3.7 does not support `transformers>=4.36.2` but only up to 4.30.2. However, the requirement is from other versa dependencies which cannot be adjusted on our side. 

In that case, do you think if we can remove python 3.7 (which was released in 2018)? For compensation, I can add python 3.11 and 3.12 into the ci.

sw005320:
> @sw005320 I think the current issue is that python 3.7 does not support `transformers>=4.36.2` but only up to 4.30.2. However, the requirement is from other versa dependencies which cannot be adjusted on our side.
> 
> In that case, do you think if we can remove python 3.7 (which was released in 2018)? For compensation, I can add python 3.11 and 3.12 into the ci.

Good ideas.
Let's do that.

ftshijt:
Additionally, I disable the default gtn installation as it was not updated since Mar. 2022 and cannot support python 3.11 and above

ftshijt:
For an update, the current CI issue was due to the issue in the original fairseq but then introduced to s3prl. I will try to start a PR in s3prl for a fix https://github.com/s3prl/s3prl/issues/522

Additional reference: https://github.com/facebookresearch/fairseq/issues/5012

ftshijt:
Hi @sw005320 , the CI has been fixed. Please feel free to review it again and merge if it looks good~ It would be a prerequisite to some of my following updates to specific modules.

espnet2/train/gan_trainer.py
sw005320, line 1: Is it related to this PR?
In any case, if it is correct, it is fine.

ftshijt: Yeah, this is a fix to the encodec model, where the discriminator could have a probability to skip update (which could cause distributed training errors).

sw005320:
OK, thanks!

PR #5883, opened 2024-08-29
Bug fix for source link

Masao-Someki:
## What?

The source link on the homepage does not work.
This is a bugfix for it

## Why?

#5847 

## See also

- Source link
image

- bug
image

- Fixed version (tested on local)
image


sw005320:
Thanks!

PR #5890, opened 2024-09-03
Add missing functionalities for espnetez

Masao-Someki:
## What?

I added two functionalities I used for my experiments.

- Function to insert new special tokens to the pretrained tokenizer, converter, and embedding layer. You can refer to the ESPnetEZ demo for SLU task for detailed usage.
- Add initialization functionality for custom preprocess function. It should take speech and text. Detail usage should be the same as

## Why?

- Several ESPnet demo requires this functionalities, and these are important features for fine-tuning purpose.

## See also

- ESPnetEZ demo for SLU task
- ESPnetEZ demo for ST task for huggingface integration.


Masao-Someki:
Note that notebooks are in PR now.
https://github.com/espnet/notebook/pull/48

Masao-Someki:
For the preprocessing function, it was a little complicated and I think this should be fixed. I will modify ez part to ignore default preprocessor even in the `collect_stats` phase. 

- Basically, when user uses custom dataset and lambda function in `data_info`, then the default preprocessor in ESPnet will be ignored during the training phase.
- However, for the `collect_stats` phase, even if users uses their custom dataset the default preprocessor was not ignored. This is why I added `build_preprocess_fn`. But this is not a good situation. I will fix this.

Masao-Someki:
I have checked this modification works on Google colab.

sw005320:
Thanks, @Masao-Someki.
Sorry that I overlooked this PR.
It looks good to me, and after the CI check, I'll merge it.

The current issue is that we do not have a connection with a notebook, and it is difficult to validate this change.
Please consider it for our future development.

sw005320:
Thanks, @Masao-Someki!

PR #5894, opened 2024-09-04
Categorical Balnced Chunk iterator

ftshijt:
This is the implementation of a categorical balanced chunk iterator, currently designed for codec training in multi-domains. 

Differences to existing iterators:
- chunk_iterator: chunk_iterator has category supports, but it only supports using one category in a batch, no randomness in the batch
- category_iterator: category_iterator does not support chunks

The current category_chunk_iterator is an extended chunk version of category_iterator. The current usage case is in codec training.


sw005320:
Please fix the CI error

sw005320:
Is it possible to add some tests for this?
This is a very useful extension and it would be a base for many other implementations.

ftshijt:
> Is it possible to add some tests for this? This is a very useful extension and it would be a base for many other implementations.

Sure thing! I'm almost done with the performance checking, will add that soon

ftshijt:
@sw005320 I think this PR is ready to merge. Please let me know if you have other suggestions. I've checked some intermediate results and will push to huggingface once the training converged.

PR #5904, opened 2024-09-21
[WIP] SpeechLM-based SE for CHiME-4

YoshikiMas:
This PR aims to support the speech enhancement (SE) task in SpeechLM style. All the scripts in `chime4/speechlm1/local` are from `chime4/enh1/local`.

ToDo:

- [x] Exploring an appropriate decoding configuration
- [x] Enhancing the evaluation stage


jctian98:
@YoshikiMas Many thanks for making this PR!!! 
As we discussed, maybe we can proceed with the following items:
(1)  `Exploring an appropriate decoding configuration` - FYI, to generate the audio/speech, we usually need to do sampling. Greedy search may not work properly. For this specific task, the input and output should have the same length so that we can pose a strong constraint on the `maxlenratio` and `minlenratio` parameters during inference.
(2)  `Enhancing the evaluation stage` - Yes, please make it.
(3) Training from scratch will cause over-fitting. It would be great if we could try to fine-tune the pre-trained TTS model. The training curves (loss and layer-wise accuracy) are helpful in diagnosing the training progress.
(4) It would also be great if you could share your prepared data and trained models following the stages 11-13.

Again, thanks so much for this great contribution!! 💯

YoshikiMas:
Thanks for your suggestions! I'll share the updates as soon as possible.

YoshikiMas:
Through discussion with @jctian98, we found that the current decoding implementation faces an error with a strict `maxlenratio` and `minlenratio` pair like `(1.0, 1.0)`. So, I currently use the top-K sampling only with `minlenratio`.

In that case, popular SE metrics like PESQ are not applicable since the length of the enhanced speech is different from the original one. So, the current `eval_se.sh` follows `eval_tts.sh`.

YoshikiMas:
Since `ValleLM` is deprecated in the latest version, I switched the model to `ARDelayLM`.

jctian98:
@YoshikiMas   Thanks for the update!!!
SE is a kind of special case the input/output should have the same length. I'll look into the inference code and make sure a fixed-length inference is well-supported!
DNS1 only has 100 hours and may be too small for an LM. We may discuss more offline. I'll start to deal with DNS2 from my side with your code and template.

YoshikiMas:
I have downloaded the raw data (clean speech and noise) on our server. And, what you have to do is just changing the data mixing config from 100 to 3000. Let me share some details offline

jctian98:
Great, I'll take care of this scaling up and let you know!

PR #5909, opened 2024-09-24
Bugfix for the paper publish workflow

juice500ml:
## What?
Paper publish build is failing

## Why?
Github actions' `actions/upload-artifact` was deprecated.
I bumped up from v1 to v4, because it doesn't seem to have difference in the interface.


sw005320:
Thanks for your quick fixes, @juice500ml!

PR #5910, opened 2024-09-25
add vc ssl2codec template to ESPnet_speechlm

wyh2000:
## What?

add voice conversion task and vocoder task to ESPnet speechlm.


jctian98:
@ftshijt Since this is simple, I think we can merge it directly.

PR #5913, opened 2024-09-27
Add Hugging Face Front End

taiqihe:
## What?

Added front end wrapper for Hugging Face audio models.

## Why?

Hugging Face is one of the most popular and active repository of pretrained models. This enables loading audio encoders supported by HF and integrating them directly into ESPNet. There are models supported by HF such as [`w2v-bert-2`](https://huggingface.co/facebook/w2v-bert-2.0) that are not supported by S3PRL.


sw005320:
Thanks, @taiqihe.
I think this is good in general!

Do you have some examples of making [w2v-bert-2](https://huggingface.co/facebook/w2v-bert-2.0) work?
I don't want to make this PR very large, so we may not have to think of incorporating it in this PR, but I just want to know the use case.

espnet2/asr/frontend/huggingface.py
```diff
@@ -0,0 +1,81 @@
...
+            self.fs = fs
+        else:
+            self.resample = False
+
+    def output_size(self) -> int:
+        return self.encoder.config.hidden_size
+
+    def forward(
+        self, inputs: torch.Tensor, input_lengths: torch.Tensor
+    ) -> Tuple[torch.Tensor, torch.Tensor]:
```
simpleoier, line 55: For better understanding, can you add some docstrings about input?

espnet2/asr/frontend/huggingface.py
```diff
@@ -0,0 +1,81 @@
...
+        with torch.no_grad():
+            # Reobtain jagged inputs to feed into the HF processor
+            device = inputs.device
+            inputs = [arr[:l].cpu().numpy() for arr, l in zip(inputs, input_lengths)]
+            if self.resample:
+                inputs = [
+                    librosa.resample(
+                        arr, orig_sr=self.fs, target_sr=self.processor.sampling_rate
+                    ).ravel()
```
simpleoier, line 64: Can I ask what would happen if the input is multi-channel? Will the `ravel()` pack everything into 1-D array?

simpleoier: I guess the intended input would be single-channel. Maybe it's better to add a shape checking somewhere before?

espnet2/asr/frontend/huggingface.py
```diff
@@ -0,0 +1,81 @@
...
+    def forward(
+        self, inputs: torch.Tensor, input_lengths: torch.Tensor
+    ) -> Tuple[torch.Tensor, torch.Tensor]:
+        with torch.no_grad():
+            # Reobtain jagged inputs to feed into the HF processor
+            device = inputs.device
+            inputs = [arr[:l].cpu().numpy() for arr, l in zip(inputs, input_lengths)]
+            if self.resample:
+                inputs = [
+                    librosa.resample(
```
simpleoier, line 62: Can we use pytorch resample?
I wonder if there is a specific reason for using librosa resample, and convert input tensor to cpu numpy array.
Does the `processor`(`AutoFeatureExtractor`) support tensor and gpu computing?

espnet2/asr/frontend/huggingface.py
simpleoier, line 1: If a test file can be added, it would be better.

simpleoier (commented):
Thanks for the new feature!

espnet2/asr/frontend/huggingface.py
taiqihe: As far as I know, HF processors don't support tensor inputs.

sw005320:
@taiqihe, can you respond to @simpleoier's reviews?
They are reasonable.

I see the value of your PR, and I want to merge it with some modifications.

taiqihe:
@simpleoier I removed the librosa resampling part and added a test for a few frontends.

I also added checks against Whisper and Encodec which need special attention to integrate. I'm not planning to implement those for now.

I'm also wondering if this should be renamed to TransformersFrontend instead, because this is really about loading models implemented in the HF transformers code base, and not loading arbitrary models hosted on the Hugging Face Hub.

espnet2/asr/frontend/huggingface.py
simpleoier: Thanks for adding more description! It would be even better to follow this [example](https://github.com/espnet/espnet/blob/master/espnet2/asr/frontend/windowing.py#L57-L66). When a new user of ESPnet or this frontend want to use or extend it, it's more informative and helpful.

taiqihe: I reformatted the docstring

simpleoier (approved):
LGTM!

sw005320:
It looks good to me, but do you have some examples?

taiqihe:
> It looks good to me, but do you have some examples?

What would be a good place to put examples? I can provide some sample configs.

sw005320:
If you are working on some existing recipe in espnet, you can add a config to the recipe config folder and the results, and model link to README.md, e.g., 
https://github.com/espnet/espnet/tree/master/egs2/librispeech_100/asr1

BTW, you still encounter some issues in the test.
@juice500ml, can you help @taiqihe?

juice500ml:
@taiqihe I checked the two failed tests. It seems that `debian11` test is giving timeouts because the hf models are too big. How about removing `wav2vec2-xls-r-300m` and `w2v-bert-2.0` (large models) from the tests? I think the remaining base models would be enough to test it. Also, for `ubuntu` test, I'm currently checking locally via [act](https://nektosact.com/usage/index.html). I'll let you know if we need additional fix on the code.

juice500ml:
FYI, failed `unit_test_espnet1_and_espnet2_on_debian11` comes from `evaluate_pseudomos.py`'s `torch.hub.load("tarepan/SpeechMOS:v1.2.0", "utmos22_strong")` calling too often, resulting in `HTTP Error 403: rate limit exceeded`, which would naturally be fixed when rerunning the tests.

taiqihe:
The two models that are left are still timing out. It can finish within 4s on my computer but I can't seem to add an exception for these tests.

juice500ml:
Does the bottleneck comes from model downloading OR inference? Maybe you can feed shorter audios (i.e., [change 1s to 0.1s](https://github.com/espnet/espnet/pull/5913/files#diff-62480259617d041515de22448fa4f6207eb9825055b9389653f1ed58d65ad1ceR13-R14)) if the latter's the problem.

juice500ml:
Also, varying audio lengths within the batch can be useful to test AutoFeatureExtractor. https://github.com/espnet/espnet/blob/master/test/espnet2/asr/frontend/test_whisper.py#L65-L75

juice500ml:
@taiqihe, I think there's a hard limit on the test suite's timeout, mainly because the existing tests are already quite big, so that we cannot increase more. Also, the existing tests use Whisper-Tiny (39M) with maximum 0.2s input. Hence, even for base models (~100M), 0.2s might be too big.  For similar reasons, we can't also do large models (~300M). How about removing the large models, and changing the input size to 640 frames (40ms)? I think that'll be around the minimum size. I checked with WavLM-base, and it seems to be it.

juice500ml:
Checked the commit, such a neat solution! I think it's likely that the tests will pass now. I checked the failing tests, now it's failing elsewhere, likely due to the occasional randomness inside the CIs. Xiaoice SVS timeouts for some reason?

taiqihe:
@juice500ml Thank you for your suggestions! It does look like the failing test is not related to this PR.

juice500ml:
I believe the PR is ready to merge, @sw005320 Thanks for your awesome contribution, @taiqihe :)

sw005320:
OK, done!

I think this PR has a lot of potential, but to utilize the potential, we need some documentation and examples.
@taiqihe, it would be great if you make some follow-up PRs to add them.

PR #5918, opened 2024-10-04
Mean normalization issue with commonpreprocessor when using data augmentation

popcornell:
many datasets such as Multi-lingual librispeech, probably other librivox material and also librispeech have the issue of not having the signal mean normalized. 
This is what it leads to when people do data augmentation with common preprocessor such as creating multi-speaker speech.
The offset is cumulated on overlapped speech parts and the signal clips.

image

I think this bug potentially affected lots of ESPnet recipes and experiments making the performance worse. 
E.g. it may happen also that RIR files are not mean normalized or also noise, especially in large scale datasets where quality may be poor. The effect will be compounded in these cases quickly making garbage .wav files like these. 
I am attempting a quick fix here. But we need more efficient and elegant code fix. 
@jctian98 @sw005320

### DIFF
--- a/espnet2/train/preprocessor.py
+++ b/espnet2/train/preprocessor.py
@@ -275,6 +275,7 @@ class CommonPreprocessor(AbsPreprocessor):
         self.audio_pad_value = audio_pad_value
 
     def _convolve_rir(self, speech, power, rirs, tgt_fs=None, single_channel=False):
+        speech = speech - np.mean(speech, 1, keepdims=True)
         rir_path = np.random.choice(rirs)
         rir = None
         if rir_path is not None:
@@ -286,6 +287,7 @@ class CommonPreprocessor(AbsPreprocessor):
                 rir = rir[:, chs]
             # rir: (Nmic, Time)
             rir = rir.T
+            rir = rir - np.mean(rir, -1, keepdims=True) # subtract mean in case RIR has problems 
             if tgt_fs and fs != tgt_fs:
                 logging.warning(
                     f"Resampling RIR to match the sampling rate ({fs} -> {tgt_fs} Hz)"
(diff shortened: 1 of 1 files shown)


sw005320:
Thanks, @popcornell.'
Indeed, it would be an issue.
There are not so many recipes using dynamic data augmentation, and I think this will just remove some bugs.
It would not affect many others.

@Emrys365, what do you think?

popcornell:
> There are not so many recipes using dynamic data augmentation, and I think this will just remove some bugs.
> It would not affect many others.

True but even if augmentation is not done, the signal still contains offset. 
May be good or bad when one uses pre-trained models that take the waveform as input e.g. neural codecs, HuBERT etc. 
This depends whether or not these models were trained to deal with signals with such offset or not.

sw005320:
I see.
Then, in the general inference case, we also need to implement online normalization processing?

popcornell:
> I see.
> Then, in the general inference case, we also need to implement online normalization processing?

Yes and no ? 
I mean realistically I think this is an artifact from defective very bad quality recording units I guess. 
You can assume that your application will have a decent one and do not care. You do normalization offline cos it is part of the data cleaning process. 

Or if you don't assume that and you want that your model is robust to such situation (which maybe you can prevent by adaptive filtering easily BTW in actual deployment, so it is separate from the model). 
For causal/streaming systems yes. E.g. Kaldi CMVN is on sliding windows. There are two options now (time has changed): 

1. Make the mean normalization on sliding windows of waveform (cos many models now take the waveform), this is akin to actually have an adaptive IIR filter to track the mean pretty much. 
2. Have an option that we do not perform normalization, but we delegate normalization to the model or feature extraction part. 
    - e.g. torchaudio sliding window cmvn or some cumulative layer norm in the model for example.  
    
Dealing such offset in the model is probably better but maybe we should implement both.

PR #5925, opened 2024-10-10
fix: collcate_fn -> collate_fn

kalvinchang:
## What?
* minor spelling fix in documentation: collcate_fn -> collate_fn

## Why?
* While looking for documentation on ESPnet's dataloading, I noticed that "collate_fn" was misspelled as "collate_fn", confusing me for a second.


sw005320:
Thanks, @kalvinchang!

PR #5931, opened 2024-10-19
Training code patches

### LINKED ISSUE #5921: Memory bug in trainer.py script
Hi,
While monitoring the memory allocation using NVIDA tools during ASR training, I encounter a bug in the espnet2/train/trainer.py. 
While doing a forward pass to calculate the loss in line 614 (retval = model(**batch)), the reference to the previous loss would cause the previous batch processing gradient to reside in memory (line 620). Since this is a significant memory allocation at least for E2E ASR, releasing this memory can free up to ~25% GPU memory.  With this bug it is difficult to tune the batch size without considering the memory used in the previous batch processing which is indead random. After this fix the memory is reset for a new batch processing, making it possible to increase the batch size and get the most out of the GPU memory for training speedup. In our case, this led to a 2x speedup.

This is a general bug which can lead to memory release during training and as a result training speedup by increasing the batch size.
 - OS information: Ubuntu 22.04.5 LTS
 - python version: Python 3.8.13
 - espnet version: latest
 - pytorch version  py3.8_cuda11.8_cudnn8.7.0_0
 -

wanchichen:
## What?

Fixes #5921.

I also observed that a failed flash attention import will silence any `logging` with a level of `INFO` or below. 
Changing it to a print statement to avoid the issue for now, as I still cannot find the root cause.

### DIFF
--- a/espnet/nets/pytorch_backend/transformer/attention.py
+++ b/espnet/nets/pytorch_backend/transformer/attention.py
@@ -18,7 +18,7 @@ try:
     from flash_attn import flash_attn_func, flash_attn_varlen_func
     from flash_attn.bert_padding import pad_input, unpad_input
 except Exception as e:
-    logging.info(f"Failed to import Flash Attention, using ESPnet default: {e}")
+    print(f"Failed to import Flash Attention, using ESPnet default: {e}")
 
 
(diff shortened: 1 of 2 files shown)


sw005320:
Thanks!

Issue #5936, opened 2024-10-29
Eval2000 text preprocess bug 


Swagger-z:
**Describe the bug**
Hi, I noticed the perl command bug haven't been fixed yet
         

> | perl -pe 's| \(\%.*\)||g' | perl -pe 's| \<.*\>||g' \

 line 64  in  egs2/swbd/asr1/local/data.sh

mentioned in https://github.com/espnet/espnet/issues/726#issue-443472412

This bug caused eval2000's transcription to omit some words, hope it will be fixed soon

sw005320:
Thanks for reminding us.
Yes, we'll work on this issue.
@siddhu001

siddhu001:
Thanks a lot @Swagger-z for reminding. I have investigated this and created a PR: https://github.com/espnet/espnet/pull/5941 to resolve this issue.

PR #5941, opened 2024-10-31
Add SWBD text processing fix

siddhu001:
## What?

This is in reference to issue https://github.com/espnet/espnet/issues/726 and https://github.com/espnet/espnet/issues/5936. Based on these discussions, I have made the necessary adjustment to the Perl command used in the SWBD text processing.

## Why?

The current Perl command executed on transcript files:

`perl -pe 's| \(\%.*\)||g'`

had a problem where, if a transcription line contained two groups of parentheses with text inside, the command treated the two groups as a single block, removing more content than intended.

For example, when running the script, this line:

`"This is huh (%HESITATION) a not so (%HESITATION) inspired example"`
was processed into:

`"This is huh inspired example"`
To address this, I updated the command as suggested, adding a question mark ? after the asterisk * in the regular expression:

`perl -pe 's| \(\%.*?\)||g'`

## Impact of the Change
After implementing this fix, I ran the data processing and observed two types of changes:

### Correction of Errors:

```
Original Text: BUT (%HESITATION) USED TO GO OUT YOU KNOW (WITH) SEVEN EIGHT TEN THAT RANGE
Before the Fix: "but seven eight ten that range"
After the Fix: "but used to go out you know with seven eight ten that range"
```
This demonstrates that there was a bug in the original SWBD data processing, and the fix effectively resolves this issue by preserving the intended content.

### Handling Disfluencies:

```
Original Text: (%HESITATION) I (JU-) I JUST SURVIVED LOS ANGELES BUT I WAS JUST DOWN THERE LAST WEEK SO
Before the Fix: "i just survived los angeles but i was just down there last week so"
After the Fix: "i ju- i just survived los angeles but i was just down there last week so"
```
The previous version of the SWBD data processing tended to remove repetitions and disfluencies, which could be useful in some cases. However, with the current fix, we retain the disfluencies as they appear in the original audio, which I believe is preferable since it aligns more closely with the original transcripts.

### DIFF
--- a/egs2/swbd/asr1/local/data.sh
+++ b/egs2/swbd/asr1/local/data.sh
@@ -13,7 +13,7 @@ SECONDS=0
 
 
 stage=1
-stop_stage=1
+stop_stage=3
 
 log "$0 $*"
 . utils/parse_options.sh
--- a/egs2/swbd/asr1/local/data.sh
+++ b/egs2/swbd/asr1/local/data.sh
@@ -13,7 +13,7 @@ SECONDS=0
 
 
 stage=1
-stop_stage=3
+stop_stage=1
 
 log "$0 $*"
 . utils/parse_options.sh
@@ -61,7 +61,7 @@ if [ ${stage} -le 1 ] && [ ${stop_stage} -ge 1 ]; then
     paste -d "" \
          <(cut -f 1 -d" " data/${x}/text.org) \
          <(awk '{$1=""; print tolower($0)}' data/${x}/text.org \
-         | perl -pe 's| \(\%.*\)||g' | perl -pe 's| \<.*\>||g' \
+         | perl -pe 's| \(\%.*?\)||g' | perl -pe 's| \<.*\>||g' \
          | sed -e "s/(//g" -e "s/)//g") \
          | sed -e 's/\s\+/ /g' > data/${x}/text.org2 # for ci check
     # remove the file with empty text, otherwise bug in stage calc perplexity


siddhu001:
@sw005320 can I merge this PR?

sw005320:
Would it change the results?
If so, I think it's better to include them.
Can you do that?

siddhu001:
Sure thing! I will check if this leads to change in results!

siddhu001:
Hi @sw005320,

I ran the previously trained model (https://huggingface.co/pyf98/swbd_e_branchformer) with the updated evaluation after the data preparation fix and observed improvements across all test sets and metrics. I also addressed a minor bug in local/score.sh related to incorrect paths. Given these results, I believe it would be beneficial to proceed with merging this PR!
Thanks!

sw005320:
Cool!
Maybe, we reached the SOTA number.

BTW, this will affect training.
So, can you also run training (with @pyf98?)?
We might get further improvements.

sw005320:
I'll merge this PR after CI, but if you have the bandwidth, please also work on training.

siddhu001:
Thanks a lot Shinji!

Currently, in local/data.sh, the Perl command is applied only to the eval2000 transcription, so I initially thought this fix wouldn’t impact model training. I’ll take a closer look and discuss with @pyf98 to check if there’s a similar issue with generating the training and validation transcripts. If I find any related issues, I’ll open another PR to update the training accordingly.

Thanks again for your feedback!

sw005320:
> Currently, in local/data.sh, the Perl command is applied only to the eval2000 transcription, so I initially thought this fix wouldn’t impact model training.

Oh, I see.
Then, you may not have to do it.

siddhu001:
@sw005320 , all the CI have passed and I will merge this PR soon!

sw005320:
Thanks!

PR #5946, opened 2024-11-07
removed ''continue'' statement from the for loop in run_mono.sh

Trikaldarshi:
## What?

removed ```continue``` statement from the ```for``` loop in ```run_mono.sh```

## Why?

Since both 10min and 1h loops are skipped, no training or ASR processing is done, and the script would prematurely reach the final echo and python commands without running any of the actual ASR tasks.


ftshijt:
Thanks for bringing up the issue and fixing it! I think you can directly remove the line (after you make the change I will merge the PR).

sw005320:
Thanks, @Trikaldarshi!

PR #5947, opened 2024-11-08
add visinger2 plus readme

jerryuhoo:
## What?

Add instruction for training VISinger 2 Plus @ftshijt


ftshijt (commented):
Thanks! Can you also add an entry at ##Table of Contents in the markdown?

PR #5951, opened 2024-11-12
Fix versa interface

ftshijt:
We have synced the scorer interface for versa, so we do not need a ESPnet-specified interface anymore.


sw005320:
How about including the use of versa (with minimum installation) as a test?

PR #5966, opened 2024-11-26
Phoneme Recognition with IPAPack (Zhu et al 2024) 

kalvinchang:
## What?
* new recipe for the IPAPack dataset (see README)
* trains an encoder-decoder baseline from scratch
* trains an encoder-decoder baseline with XEUS features from scratch
    * provides script to extract XEUS features in an offline fashion to avoid wasted computation of frozen S3Ms

## Why?

* introduces a phoneme recognition dataset to ESPnet, which are quite few


kalvinchang:
Original code: https://github.com/kalvinchang/espnet/tree/ipapack_encoder-decoder
(commit history)

I am not merging `ipapack_encoder-decoder` because that branch contains XEUS

egs2/ipapack/asr1/local/generate_utt2dur.py
sw005320, line 1: It's better not to use the absolute paths.

akreal:
This looks interesting!

Regarding why XEUS features do not perform well, it seems that output of one layer only is used (layer "-2"), do I get it right?  Could this be the reason, because normally it's weighted sum of all layers' outputs?

kalvinchang:
> Great!
> 
> * Can you tell me why XEUS features do not perform well? Can you paste the learning curves?

asr_train_asr_transformer_raw_word (no XEUS features)

asr_train_asr_xeus_transformer_extracted_word (XEUS features)

It seems like XEUS provides the model with a head start, but eventually, the head start decreases.

Another reason could be that DoReCo is never used in the dev set (Table 1 of https://aclanthology.org/2024.naacl-long.43.pdf).

kalvinchang:
> This looks interesting!
> 
> Regarding why XEUS features do not perform well, it seems that output of one layer only is used (layer "-2"), do I get it right? Could this be the reason, because normally it's weighted sum of all layers' outputs?

Thanks for your interest :)

Indeed, we only use the penultimate output of one layer to save computation and memory. Dumping all layers in an offline fashion would take 24 TB of data (1 layer is already 1 TB). Shinji suggested going back to online feature extraction, which allows us to take a weighted sum.

sw005320:
@ftshijt, can you also check this PR?

egs2/ipapack/asr1/README.md
```diff
@@ -0,0 +1,195 @@
...
+### CER
+
+|dataset|Snt|Wrd|Corr|Sub|Del|Ins|Err|S.Err|
+|---|---|---|---|---|---|---|---|---|
+|decode_asr_asr_model_valid.acc.best/test_doreco|5093|396745|70.3|17.5|12.2|3.3|33.0|100.0|
+
+### TER
+
+|dataset|Snt|Wrd|Corr|Sub|Del|Ins|Err|S.Err|
+|---|---|---|---|---|---|---|---|---|
```
ftshijt, line 68: We can simply remove TER if not used (same for WER if it is not properly used here)

sw005320:
My concern about this PR is the normalization issue in the test set evaluation, which makes our evaluation difficult. So, can you revisit this part and handle it appropriately?

kalvinchang:
> My concern about this PR is the normalization issue in the test set evaluation, which makes our evaluation difficult. So, can you revisit this part and handle it appropriately?

Sounds good! I will prepare IPAPack+ (20k hours) from Jian, which already does a lot of the normalization we need to do.

egs2/ipapack/asr1/conf/tuning/train_asr_transformer.yaml
```diff
@@ -0,0 +1,52 @@
...
+encoder_conf:
+    output_size: 512
+    attention_heads: 8
+    linear_units: 2048
+    num_blocks: 12
+    dropout_rate: 0.1
+    positional_dropout_rate: 0.1
+    attention_dropout_rate: 0.0
+    input_layer: conv2d
+    # 40 ms resolution (1/4 length)
```
kalvinchang, line 28: this could actually be hurting performance if using XEUS features

since the frame rate of XEUS is 50 fps

I will use conv2d1 (no downsampling) in the future

ftshijt: Usually for ASR, we use conv2d2 for SSL feature. (50% downsample)

Using half is a good balance to efficiency and modeling performance. But maybe phone would be a different story?

PR #5969, opened 2024-12-03
[Bugfix] fix keyword argument error in stage 7 of diar.sh

eric102004:
## What?

Fix the order of arguments so that keyword arguments (`--collar` `--fs` `--frame_shift`) can be properly passed into `score_der.sh`.

## Why?

I found that the above three arguments could not be passed into `score_der.sh` so it sometimes used the incorrect sampling rate to calculate the DER even when we set the correct one in run.sh.


sw005320:
Thanks, @eric102004!

PR #5974, opened 2024-12-07
fix: gradient truncation bug in pit_solver.py

YuzhuWang-code:
## What?

When training a speech separation model with multiple losses, only the losses with independent_perm=True have gradients, while the gradients of all other losses with independent_perm=False are truncated. 

## Why?

The root cause is the use of torch.tensor() to create a new tensor in pit_solver.py (when independent_perm=False). Here, torch.tensor() function detached from the computational graph when creating a new tensor, not preserving the gradient. 

## Solution

A directly available solution is replacing torch.tensor() with torch.stack()​.

## Verification

When checking the combined loss (L= alpha1 * loss1 + alpha2 *loss2 ), L has a gradient, where loss1 has a gradient and loss2 has no gradient. We compute the gradient norm during backpropagation to verify the contribution of each loss component in the combined loss, and we found that the contribution always comes from the loss with independent_perm=True only. Here is an example of gradient tracking.

## GRADIENT TRACKING VERIFICATION in pit_solver

1. Original Implementation Verification:
--------------------------------------------------

Original Implementation Results:

 Permutation Loss (batch 0, spk 0):
    Value: 1.525882
    Gradient: True

 Permutation Loss (batch 0, spk 1):
    Value: -0.320282
    Gradient: True

 After first torch.tensor operation:
    Value: 0.602800
    ↓ ↓ ↓ Gradient flow stopped at this point ↓ ↓ ↓
    Gradient: False

 After second torch.tensor operation:
    Value: 0.602800
    Gradient: False

2. Stack Implementation Verification:
--------------------------------------------------

Stack Implementation Results:

 Permutation Loss (batch 0, spk 0):
    Value: 1.525882
    Gradient: True

 Permutation Loss (batch 0, spk 1):
    Value: -0.320282
    Gradient: True

 After first torch.stack operation:
    Value: 0.602800
    Gradient: True

 After second torch.stack operation:
    Value: 0.602800
    Gradient: True

--------------------------------------------------

Results Comparison:
torch.tensor Loss: 0.602800 (Gradient: False)
torch.stack Loss: 0.602800 (Gradient: True)

### DIFF
--- a/espnet2/enh/loss/wrappers/pit_solver.py
+++ b/espnet2/enh/loss/wrappers/pit_solver.py
@@ -98,9 +98,9 @@ class PITSolver(AbsLossWrapper):
                     perm0 = perm_.unsqueeze(1)
                 stats[k] = new_v.gather(1, perm0.to(device=new_v.device)).unbind(1)
         else:
-            loss = torch.tensor(
+            loss = torch.stack(
                 [
-                    torch.tensor(
+                    torch.stack(
                         [
                             pre_hook(
                                 self.criterion,


YuzhuWang-code:
@sw005320

sw005320:
Thanks for the report.
This sounds very critical.
@Emrys365, can you check this PR?
Do you know if you come up with where this would affect mainly? Would ASR and enhancement joint training be affected by this? (@simpleoier and @YoshikiMas )

simpleoier:
@sw005320 I had a quick check of the code. ASR has its own pit loss wrapper, which is not affected. But this affects enh and enh-asr joint training models, when using multiple losses and setting `independent_perm=False`. I found one example that might be affected [here](https://github.com/espnet/espnet/blob/master/egs2/wsj0_2mix_spatialized/enh_asr1/conf/tuning/train_enh_asr_beamformer_fbank_transformer.yaml#L67).

simpleoier (approved):
Thanks for fixing it!

YoshikiMas:
Thank you for bringing this up. I expect that this would not directly affect SSE-ASR unless you use multiple SSE losses during the joint training.

In detail, the current implementation of SSE-ASR computes the ASR losses, CTC and AED, after sorting the enhanced speech based on the SSE or CTC loss. Then, SSE and ASR losses are [just added](https://github.com/espnet/espnet/blob/efd0d59150b1f0077b8fbd0326fa9040365e7277/espnet2/enh/espnet_enh_s2t_model.py#L302) (not  `torch.mean` following `torch.tensor`). Also, `enh_s2t_model` has [its own implementation for PIT with the CTC loss](https://github.com/espnet/espnet/blob/efd0d59150b1f0077b8fbd0326fa9040365e7277/espnet2/enh/espnet_enh_s2t_model.py#L421), and it looks good to me.

simpleoier:
@YoshikiMas Yes, you're right. I overlooked that there is single SSE loss. So far, the SSE-ASR joint training was not affected.

sw005320:
Thanks, @YuzhuWang-code!
I just merged the code.

@simpleoier, @YoshikiMas, @Emrys365, please let me know if there are other examples for which we need to update the recipes.

YoshikiMas:
- I found some outdated config files for `wsj0_2mix` have multiple loss functions and `independent_perm: False`. They might be affected by this issue.
- [The example](https://github.com/espnet/espnet/blob/master/egs2/wsj0_2mix_spatialized/enh_asr1/conf/tuning/train_enh_asr_beamformer_fbank_transformer.yaml#L67) brought up by @simpleoier looks unnatural to me as it has only one SSE loss but with `independent_perm: False`. I might misunderstand something and would like to hear @Emrys365's intent.
- In [my example](https://github.com/espnet/espnet/blob/master/egs2/wsj0_2mix_spatialized/enh1/conf/tuning/train_enh_tfgridnet_waspaa2023.yaml), the second criterion (CI-SDR) is with `independent_perm: False`, but it is for monitoring CI-SDR by setting the weight to zero. So, the gradient for the second criterion should originally be zero.

Emrys365:
Sorry I totally missed the message... I think the above bug both
1. affected speech separation training in [`espnet2/enh/espnet_model.py`](https://github.com/espnet/espnet/blob/master/espnet2/enh/espnet_model.py) when multiple losses are used with some loss wrappers using `independent_perm: False`
    > This will cause the losses with `independent_perm: False` to be ignored during back-propagation.
2. and affected joint SSE-ASR training in [`espnet2/enh/espnet_enh_s2t_model.py`](https://github.com/espnet/espnet/blob/master/espnet2/enh/espnet_enh_s2t_model.py) when `calc_enh_loss: true` and `independent_perm: False` are used.
    > This will lead to no effective SSE loss (without gradient) in joint training, i.e., only the ASR loss is used for end-to-end optimization.

Emrys365:
> [The example](https://github.com/espnet/espnet/blob/master/egs2/wsj0_2mix_spatialized/enh_asr1/conf/tuning/train_enh_asr_beamformer_fbank_transformer.yaml#L67) brought up by @simpleoier looks unnatural to me as it has only one SSE loss but with independent_perm: False. I might misunderstand something and would like to hear @Emrys365's intent.

Originally, I thought `independent_perm: False` will have the same effect as `independent_perm: True` if both of them worked successfully (i.e., with effective gradients), because the optimal permutation will always be calculated from scratch if no previous losses are available, even when `independent_perm: False` is used. It was not a good design though.

YoshikiMas:
> Originally, I thought independent_perm: False will have the same effect as independent_perm: True if both of them worked successfully (i.e., with effective gradients), because the optimal permutation will always be calculated from scratch if no previous losses are available, even when independent_perm: False is used. It was not a good design though.

Thank you for the clarification. Since this PR is already merged, was the issue solved?

Emrys365:
Yes, I believe this is now addressed. I also checked the use of `torch.tensor` in the codebase, and it seems to be all-clear.

PR #5976, opened 2024-12-09
[WIP] Update docstring

Masao-Someki:
## What?

Add or update docstrings using ChatGPT-4o-mini.

## Why?

To provide more detailed and comprehensive docstrings while reducing the manual workload.


sw005320:
Thanks!
@ftshijt has an idea of semi-automatically providing docstring with LLMs :)

@pyf98, can you take a look?

Masao-Someki:
The docstrings are automatically generated and inserted to the source code. The template I'm using to generate this code is the following:

```
Please generate a docstring for the {name}, which is part of the {package} package.
Include only docstrings, and do not include anything else.
If there is already a docstring, update it accordingly.

**Entire Python code**

\```python
{source_code}
\```

**Docstring**
```

{name} can be a class name or function name, and {package} can be a class name or file name. So for example, the template looks like:

```
Please generate a docstring for the encode, which is part of the ESPnetS2TModel package.
...
```

or 
```
Please generate a docstring for the ESPnetS2TModel, which is part of the espnet2/s2t/espnet_model.py package.
...

**Comment by pyf98** [comment]
Thanks. It looks interesting.
Do you provide the entire repo to LLM, or just a single script?
Does it automatically modify the original scripts? Or we still need to manually copy and paste to the correct place?

When working locally, we can use VSCode Copilot to generate something similar. It can also consider multiple files as context.

**Comment by sw005320** [comment]
E501 issues...
https://github.com/espnet/espnet/actions/runs/12245236204/job/34158714184?pr=5976

**Comment by Masao-Someki** [comment]
@pyf98
Thank you! I’ve uploaded the files here: https://github.com/Masao-Someki/docstring_generator
This tool automatically updates docstrings. Since I couldn’t find an existing repo like this, I created it when I need to write docstrings for ESPnetEZ.

I looked into Copilot briefly, but it seems it doesn’t support API calls from Python? If it does, we could consider exploring that!

@sw005320
I think we can address this using a formatting package or by adding a note about length directly to the template. I’ll try some ideas.

**Comment by pyf98** [comment]
@Masao-Someki Thanks for sharing.
Regarding Copilot, yes, it is mainly for local use in a workspace. So, APIs might be difficult. It basically calls one of the LLMs based on the provided context files: GPT 4o, o1-preview, o1-mini, Claude 3.5. 
The benefit is that Copilot offers free access to certain users: student, teacher, or maintainer of popular open-source repos.
Copilot also runs on GitHub webpage.

**Comment by sw005320** [comment]
Should we just rely on them, or some hallucinations may happen?

**Comment by Masao-Someki** [comment]
@sw005320 
I noticed that hallucinations happened sometimes in `args` field. I didn't check the detailed descriptions now, so I plan to look at these documents one by one after generating all documents, but overall I feel these documents are good.

**hallucination example**

`espnet2/gan_codec/encodec/encodec.py EncodecDiscriminator:forward()`

I think this mistake comes from the docstrings for `__init__`.

```python
    def forward(self, x: torch.Tensor) -> List[List[torch.Tensor]]:
        """
            Encodec Discriminator with only Multi-Scale STFT discriminator module.

        This class implements the Encodec Discriminator, which utilizes a
        Multi-Scale Short-Time Fourier Transform (STFT) for evaluating
        the quality of generated audio signals. The discriminator aims
        to distinguish between real and generated audio, contributing to
        the adversarial training process.

        Attributes:
            msstft (MultiScaleSTFTDiscriminator): The multi-scale STFT
                discriminator instance used for feature extraction.

        Args:
            msstft_discriminator_params (Dict[str, Any]): Parameters for the
                Multi-Scale STFT Discriminator, including:
                - in_channels (int): Number of input channels.
                - out_channels (int): Number of output channels.
                - filters (int): Number of filters in convolutions.
                - norm (str): Normalization choice for convolutional layers.
                - n_ffts (Sequence[int]): Sizes of FFT for each scale.
                - hop_lengths (Sequence[int]): Length of hop between STFT
                    windows for each scale.
                - win_lengths (Sequence[int]): Window sizes for each scale.
                - activation (str): Activation function choice for
                    convolutional layers.
                - activation_params (Dict[str, Any]): Parameters for the
                    activation function.

        Examples:
            >>> discriminator = EncodecDiscriminator()
            >>> input_signal = torch.randn(8, 1, 16000)  # Batch of 8, 1 channel, 16000 samples
            >>> outputs = discriminator(input_signal)
            >>> print(len(outputs))  # Number of scales
            >>> print(len(outputs[0]))  # Number of layers in the first scale
        """

        msstft_out = self.msstft(x)

        return msstft_out
```

I think the above mistake comes from the class definition.

PR #5980, opened 2024-12-12
ESPnet-SPK: add SASV support

Alexgichamba:
## What?

Couple of updates for ESPnet-SPK to support:

- training of integrated SASV models
  - generic multi-task learning (MTL) objectives
  - a MTL-aware category sampler
  - new metric: a-DCF (https://arxiv.org/abs/2403.01355)
-  embedding averaging for enrollment
- new ASVspoof5 recipe
- using precomputed feats
- a modified SKA-TDNN with cross attention for incorporating such feats

## Why?

To support the R&D of spoofing robust speaker verification systems


sw005320:
Thanks, @Alexgichamba!

@Jungjee, do you have the bandwidth to review this?
If not, I want to ask @ftshijt (after your deadline is over...)

sw005320:
There are several CI errors.
https://github.com/espnet/espnet/actions/runs/12302925617/job/34336883562?pr=5980
Please fix them.

Jungjee:
@sw005320  I'll take a look!

Alexgichamba:
> There are several CI errors. https://github.com/espnet/espnet/actions/runs/12302925617/job/34336883562?pr=5980 Please fix them.

Noted, will address them before proceeding.

egs2/TEMPLATE/spk1/spk.sh
```diff
@@ -72,6 +72,27 @@ inference_config=conf/decode.yaml   # Inference configuration
 inference_model=valid.eer.best.pth  # Inference model weight file
 score_norm=false      # Apply score normalization in inference.
 qmf_func=false        # Apply quality measurement based calibration in inference.
+cos_sim=false       # Use cosine similarity for scoring
```
Jungjee, line 75: I think we don't need this as we have `score_norm` and `qmf_func` to control the inference behavior. Can you let me know what this variable does?

egs2/TEMPLATE/spk1/spk.sh
```diff
@@ -262,8 +283,17 @@ if [ ${stage} -le 3 ] && [ ${stop_stage} -ge 3 ]; then
...
-            # category2utt will be used bydata sampler
+            # category2utt will be used by the data sampler
             cp data/"${train_set}/spk2utt" "${data_feats}/${train_set}/category2utt"
+
+            # copy files for spoofing task if multi-task sasv is enabled
+            if "${multi_task}" && "${sasv_task}"; then
+                cp data/"${train_set}/utt2spf" "${data_feats}/${train_set}/utt2spf"
+                cp data/"${train_set}/spf2utt" "${data_feats}/${train_set}/spf2utt"
+                # category2utt_2 will also be used by the data sampler
+                cp data/"${train_set}/spf2utt" "${data_feats}/${train_set}/category2utt_2"
```
Jungjee, line 294: what is `category2utt_2` for? Can you update the name to reflect what it contains?

egs2/TEMPLATE/spk1/spk.sh
```diff
@@ -72,6 +72,27 @@ inference_config=conf/decode.yaml   # Inference configuration
...
+# Use pseudomos scores
+pseudomos=false   # Use pseudomos scores for training
+train_utt2pmos=   # absolue path to utt2pmos file, required if pseudomos is true
+valid_utt2pmos=   # absolue path to utt2pmos file, required if pseudomos is true
+
+# use precomputed feats
+precomputed_feats=false        # Use precomputed features for training
+
+# Label related
+no_labels=false    # No labels for inference
```
Jungjee, line 95: I'm not sure about how several variables would be used. 
Several of them also seem to overlap with existing variables. 
Can you let me know what each variable does?

Jungjee: after all these changes, I'm starting to think that maybe adding an SASV task would be better. How does this sound to you?

egs2/asvspoof5/spk1/local/asvspoof5_train_data_prep.py
```diff
@@ -0,0 +1,48 @@
+# asvspoof5_train_data_prep.py
```
Jungjee, line 1: FYI, we're releasing the official ASVspoof5 dataset in a few days via zenodo. 
Can you please confirm if the formats match with the official public version?

I believe the version you're using is the one from the challenge.

espnet2/bin/spk_embed_extract.py
```diff
@@ -128,6 +128,7 @@ def extract_embed(args):
             output_dir=args.output_dir,
             custom_bs=bs,
             average=args.average_embd,
+            embed_avg=args.embedding_average,
```
Jungjee, line 131: ditto, how is `embed_avg` different from `average`?

Jungjee (commented):
Thanks, Alex for implementing SASV in ESPnet with your updated functions. 
I would need to review again more thoroughly, but even at this point, I have two main points to discuss:
1. do you think it's better to have sasv as a separate task?
2. how are many variables that you added different from existing ones that have similar names?

Alexgichamba:
> Thanks, Alex for implementing SASV in ESPnet with your updated functions. I would need to review again more thoroughly, but even at this point, I have two main points to discuss:
> 
> 1. do you think it's better to have sasv as a separate task?
> 2. how are many variables that you added different from existing ones that have similar names?

Hi Jeeweon,
Thanks for taking a look.

> 1. I didn't see enough variation to warrant a different task, but perhaps you see it differently?
> 2. A number, but this is definitely something I can clean up without much of a struggle.

egs2/TEMPLATE/spk1/spk.sh
Alexgichamba: It's providing an alternative to what was existing for computing scores (explicit cosine similarity vs average L2 distance)
https://github.com/espnet/espnet/blob/master/espnet2/train/spk_trainer.py#L132

egs2/TEMPLATE/spk1/spk.sh
Alexgichamba: Since category2utt was generic in naming, I assumed the other one could also be generic.
Since it's only seen by the sampler, which can be generically used.
But I can update the name. Should I update the name to spf2utt?

egs2/TEMPLATE/spk1/spk.sh
Alexgichamba: Ok let me try:
multi_task - require two sets of labels used along with sasv_task for multi task learning
sasv_task - defines that the secondary task is spoofing
spf_args - collects a bunch of args similar to spk_args
embed_avg - embedding averaging for enrollment utterances
embed_avg_args - picks up the extra args for embedding averaging (trial3.scp and trial4.scp)
pseudomos - if true then approximating pseudomos scores will be an secondary objective
train_utt2pmos and valid_utt2pmos - has pmos labels for a given utt
precomputed_feats - allows precomputed features to be passed as .npy arrays
no_labels - this is because by default ESPnet SPK requires trial_label but for ASVspoof5 test no such labels were available. I could have missed something because I struggled to find any other workaround for test sets where you only want to compute scores, since labels aren't present.

I see. Could you kindly let me know why a new task would be better?
I'm okay with it, if a new task will result in the best outcomes... then I will definitely switch to that.

egs2/asvspoof5/spk1/local/asvspoof5_train_data_prep.py
Alexgichamba: That's nice :)
please share the link when it's ready so I can adapt

espnet2/bin/spk_embed_extract.py
Alexgichamba: Yeah, this is something to clean up
Original functionality was different from what I wanted, and I didn't want to break existing recipes

PR #5982, opened 2024-12-13
remove gtn in setup.py

sw005320:
## What?

Remove `gtn` in setup.py

## Why?

`gtn` is not used in default.
Also, this seems to cause an issue in installation.


sw005320:
@Masao-Someki, can you check this?

Masao-Someki:
@sw005320 
I investigated the code where `gtn` is used. As a result, I found that it is used only when `gtnctc` is explicitly set for CTC, instead of the default `builtin`. Since there is no other part to use this, I agree with removing `gtn` from the dependencies.

As one suggestion, how about adding a try-except to functions like `from_pretrained`?

E.g.
https://github.com/espnet/espnet/blob/9e0876d058548931c8ae3f70dd81027f192e9066/espnet2/bin/asr_inference.py#L693-L701

Fhrozen:
@sw005320 Are you planning to merge this soon? Other PRs may need this to fix some issues.

sw005320:
will do, but I’m careful and at least want to check the CI.

sw005320:
I’ll go to sleep, but if this PR passes the CI, you can merge it.

Fhrozen:
FYI, the unit test will continue showing error due to a confict between protobuf and sentencepiece. I did not check which package is employing `protobuf==3.20.3`, but the current version of sentencepiece for espnet requires  `protobuf==3.14.0`. I added a PR for updating the sentencepiece version.

sw005320:
Do you mean this?
https://github.com/espnet/espnet/pull/5988
This may lose the compatibility.

Fhrozen:
I ran the CI test using pytorch=2.4, python=3.10 and could not replicate the error (Segmentation fault on Feature Collection for s2st_train. I am testing with pytorch 1.13.1 to check a possible version issue.

sw005320:
Thanks, @Fhrozen.
This is a breaking change, and we have to monitor its status for a while.

PR #5994, opened 2024-12-25
Add Scaled Dot Product Attention (SDPA) from PyTorch

pyf98:
## What?

This PR adds the SDPA implementation from PyTorch. It supports several implementations of the attention mechanism (including flash attention), some of which are more efficient than the default implementation.

See https://pytorch.org/docs/2.5/generated/torch.nn.functional.scaled_dot_product_attention.html for more information.

Note: SDPA does not seem to support flash attention with variable-length inputs. So, we still need to use the previous flash attention implementation.


sw005320:
Thanks!
Can you fix some CI errors?

sw005320:
Can you check some treatments for this library?
Our CI complains about itl
https://github.com/espnet/espnet/actions/runs/12507998828/job/34895445316?pr=5994

sw005320:
Thx!

PR #5999, opened 2024-12-28
Genshin Dataset recipe of vits model

Whale-Dolphin:
## What?

Add recipe for running TTS of vit model on genshin corpus.
Recipe for genshin dataset training and model in hf.

## Why?

To facilitate the experiments on genshin dataset, a popular game dubbling corpus, via ESPnet.

## See also

https://huggingface.co/WhaleDolphin/Genshin-vits-espnet2
https://github.com/AI-Hobbyist


egs2/TEMPLATE/tts1/tts.sh
```diff
@@ -1111,12 +1125,82 @@ else
...
+        if ${gpu_inference}; then
+            _cmd="${cuda_cmd}"
+            _ngpu=1
+        else
+            _cmd="${decode_cmd}"
+            _ngpu=0
+        fi
+
+        ${_cmd} --gpu "${_ngpu}" JOB=1:"${_nj}" "${_eval_dir}"/versa_eval.JOB.log \
+        python -W ignore -m versa.bin.scorer \
```
ftshijt, line 1182: Thanks for the update! I think we can also include  wer evaluation here with the text information?

egs2/TEMPLATE/tts1/tts.sh
```diff
@@ -1111,12 +1125,82 @@ else
...
+            --pred ${_eval_dir}/pred.JOB \
+            --score_config ${_score_config} \
+            --cache_folder ${_eval_dir}/cache \
+            --use_gpu ${gpu_inference} \
+            --output_file ${_eval_dir}/result.JOB.txt \
+            --io soundfile
+
+        python pyscripts/utils/aggregate_tts_eval.py \
```
ftshijt, line 1190: I did not found this script. Maybe forget to update?

egs2/genshin/tts1/conf/tuning/train_full_band_multi_spk_vits.yaml
```diff
@@ -0,0 +1,187 @@
+# This configuration is for ESPnet2 to train 44.1 kHz
```
ftshijt, line 1: We may consider simplify the config for the initial PR by just using the best config?

egs2/TEMPLATE/tts1/tts.sh
```diff
@@ -116,6 +116,16 @@ lang=noinfo      # The language type of corpus.
 text_fold_length=150   # fold_length for text data.
 speech_fold_length=800 # fold_length for speech data.
 
+# VERSA eval related
+skip_scoring=true # Skip scoring stages.
+skip_wer=true # Skip WER evaluation.
```
ftshijt, line 121: VERSA can also do WER, we can consider using versa only, which makes the maintain effort easier?

egs2/genshin/tts1/conf/tuning/train_full_band_multi_spk_vits.yaml
Whale-Dolphin: ok, I see

Whale-Dolphin:
OK, I'll see it right now!

egs2/genshin/tts1/run.sh
```diff
@@ -0,0 +1,50 @@
...
+    --feats_type raw \
+    --fs "${fs}" \
+    --n_fft "${n_fft}" \
+    --n_shift "${n_shift}" \
+    --token_type phn \
+    --cleaner tacotron \
+    --max_wav_duration 60 \
+    --g2p "${g2p}" \
+    --tts_task gan_tts \
+    --use_sid true \
```
ftshijt, line 41: I saw you are using sid options, which require all splits including all speakers. But the data preparation stage actually does not force this requirement (the set is split randomly). Is this intentional (also does the data set have any fixed train/dev/test split in its release)? 

If no official split is provided, how about introducing our own split and fix it to make sure the models are comparable in their training data every time it runs?

egs2/genshin/tts1/conf/tuning/train.yaml
```diff
@@ -0,0 +1,78 @@
+# This configuration is for ESPnet2 to train Tacotron 2. Compared to the
```
ftshijt, line 1: This one seems to be not used and correct, we may remove it?

ftshijt (commented):
Thanks! The recipe looks great in general. I'm having a few minor questions related.

egs2/genshin/tts1/conf/tuning/train.yaml
Whale-Dolphin: OK, I'll remove it. I misunderstood that a recipe must have a train.yaml file to pass the test. Sorry for my mistake.

egs2/genshin/tts1/run.sh
Whale-Dolphin: I'll finish the config and PR again!

Whale-Dolphin: Oh, another reason for using random splitting rather than sequential splitting is that the dataset is categorized by different types of speech (such as level-up dialogues and altitude-related conversations with other characters). If we were to use conventional sequential splitting, we would lose the variety across these categories.

ftshijt:
Many thanks for your contribution! Will merge the PR.

PR #6000, opened 2024-12-30
Fix issues in OWSM CTC segmentation

pyf98:
## What?

This PR fixes a few issues in OWSM-CTC forced aligner. The new script should be more reliable and memory-efficient than the previous one. The script has been used in the segmentation of large-scale YODAS2.

The PR also fixes a typo in the installation documentation.


sw005320:
Thanks!

PR #6005, opened 2025-01-04
Fix minor bug in OWSM-CTC preprocessor

pyf98:
## What?

During OWSM-CTC training, the language token is randomly masked so that the model will learn to predict it. However, this should not happen for validation.


sw005320 (approved):
Thanks for catching it.
Please merge this PR once the CI test is passed.

PR #6007, opened 2025-01-06
Implement unified batch decode interface for OWSM-CTC

pyf98:
## What?

- Implemented a unified batch decode interface for OWSM-CTC greedy search. The `decode_batch` method can decode a batch of audios which can be either short-form or long-form. Each audio can be provided as a path, a numpy 1-D array or a torch 1-D tensor. This makes the usage more flexible.
- Enabled flash attention for inference. With mixed precision and flash attention, we can decode ~200 samples at the same time on a GPU with 96GB memory.
- Added a stand-alone utility script to average model checkpoints.

### DIFF
--- a/espnet/nets/pytorch_backend/transformer/attention.py
+++ b/espnet/nets/pytorch_backend/transformer/attention.py
@@ -190,7 +190,7 @@ class MultiHeadedAttention(nn.Module):
             return self.linear_out(out)  # (batch, time1, d_model)
 
         # Use Flash Attention implementation
-        if self.training and self.use_flash_attn:
+        if self.use_flash_attn:
             try:
                 # In the causal case, the last row will be the key mask
                 key_nonpad_mask = mask[:, -1, :]  # (#batch, time2)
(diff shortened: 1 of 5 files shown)


pyf98:
This PR is ready.

espnet/nets/pytorch_backend/transformer/attention.py
sw005320, line 1: why do you need this change?

espnet2/bin/s2t_inference_ctc.py
sw005320, line 1: Can you add a test for this batch decoding?

espnet/nets/pytorch_backend/transformer/attention.py
pyf98: The previous code uses flash attention only during training. But we can also use it for inference.

espnet2/bin/s2t_inference_ctc.py
pyf98: Added.

sw005320:
Thanks!

PR #6008, opened 2025-01-08
BEATs Tokenizer Inference

Shikhar-S:
## What?

Add inference code for using BEATs as audio tokenizer.
Adds Random projection quantizer as described in [BEST-RQ](https://arxiv.org/pdf/2202.01855) paper.

## Why?

1. BEATs pre-training
2. SpeechLM experiments


Shikhar-S:
Requesting @jctian98 for review.
cc: @sw005320

jctian98:
Hi @Shikhar-S  Many thanks for the PR. It would be very helpful. In the perspective of future SpeechLM development, I have some high-level questions. Could you share your opinion?

For audio/speech tokenization, we currently have `codec` and `ssl` tokenization methods. Specifically, we usually concat `codec` and `ssl` in frame level; in SpeechLM we call it `codec_ssl` representation of speech.

In the context of BEATs, do you have any idea about how we can include the BEATs tokens in the future? say:
(1) we concat `codec`, `ssl` and `BEATs` all frame-by-frame
(2) we replace `ssl` with `beats` so we only concat `beats` and `codec` for non-speech audio
(3) `beats` only for non-speech audio (not very ideal because we can hardly generate audio then)
(4) any other plan?

The answer to this question may lead to different designs for beats tokenization. Currently, we have:
(a) `codec_tokenization.sh` for codec
(b) `ssl_tokenization.sh` for SSL
(c) `codec_ssl_tokenization.sh` that call `codec_tokenization.sh` and `ssl_tokenization.sh` and then concat them together.

Of course, this is only from SpeechLM development perspective, if you have other considerations, please also let us know. 
Many thanks!

Shikhar-S:
Thanks for the comments!

> In the context of BEATs, do you have any idea about how we can include the BEATs tokens in the future? say: (1) we concat `codec`, `ssl` and `BEATs` all frame-by-frame (2) we replace `ssl` with `beats` so we only concat `beats` and `codec` for non-speech audio (3) `beats` only for non-speech audio (not very ideal because we can hardly generate audio then) (4) any other plan?

I like (1). One downside is that it can blow up the seq lengths. 

For (2) we have to know before hand what style of task it is (speech vs audio) and switch tokenizers. Also there is little freedom for transfer learning from beats token to speech tasks (for example use of background context encoded in beats).

While (3) has same problems as (2), it would be nice to experiment with beats tokens for generating audio at some point.
 
To conclude, I think (1) is currently the best direction.


> The answer to this question may lead to different designs for beats tokenization. We should consider how to integrate BEATs tokenization. Currently, we have: (a) `codec_tokenization.sh` for codec (b) `ssl_tokenization.sh` for SSL (c) `codec_ssl_tokenization.sh` that call `codec_tokenization.sh` and `ssl_tokenization.sh` and then concat them together.

Thanks for explaining your design.
The reason for current design choice in this PR is that I was not aware of `ssl_tokenization.sh` ( I see now that it is in the speechlm branch)
We can go for different `x_tokenization.sh` scripts for different styles of tokenizers (`x`). The advantage would be less merge conflicts and the modularity will help us to combine them together with a single `combine_tokenization.sh` script later. I will change the current design for beats if this seems okay to you?

jctian98:
Thanks for the reply!!!

> I like (1). One downside is that it can blow up the seq lengths.

Just want to make sure the beats token is like SSL: 50 frames / second; 1 code / frame.
It will increase the number of tokens from 9 to 10 but it might be fine.

> We can go for different x_tokenization.sh scripts for different styles of tokenizers (x). The advantage would be less merge conflicts and the modularity will help us to combine them together with a single combine_tokenization.sh script later. I will change the current design for beats if this seems okay to you?

Yeah I also think this is a good idea. We can have a script like `beats_tokenization.sh` or similar. Could you also revise `codec_ssl_tokenization.sh` to combine any combination of `codec`, `ssl` and `beats`? later on we can use it consistently in our tokenization stages.

One more question: would `beats` be a proper name for `modality`? it's a name for one model, not a kind of models. If we have a better name for audio foundation model tokens, can we consider change it? (or `beats` is still good at this moment for me.

Many thanks again for making this PR!!!

Shikhar-S:
> Just want to make sure the beats token is like SSL: 50 frames / second; 1 code / frame. It will increase the number of tokens from 9 to 10 but it might be fine.

Yes beats is ~50 tokens per second.


> Could you also revise `codec_ssl_tokenization.sh` to combine any combination of `codec`, `ssl` and `beats`? later on we can use it consistently in our tokenization stages.

`codec_ssl_tokenization.sh` is not in the master branch right now. Let us get it in and then I can modify the scripts to integrate audio tokenization? Meanwhile, since this is an independent tokenizer we can check this in. I want to use it for pre-training audio models.

> One more question: would `beats` be a proper name for `modality`? it's a name for one model, not a kind of models. If we have a better name for audio foundation model tokens, can we consider change it? (or `beats` is still good at this moment for me.

Thanks for the suggesiton. Changed to audio_tokenization.sh

jctian98:
Sounds good! Then we can first have the script for `audio_tokenization.sh` :)

Shikhar-S:
> Sounds good! Then we can first have the script for `audio_tokenization.sh` :)

updated, ptal.

sw005320:
LGTM!
Thanks, @Shikhar-S!

PR #6017, opened 2025-01-20
cmu_kids

wangpuup:
## What?

Add ASR results on CMU kids data.


sw005320:
Thanks, @wangpuup!
Can you fix this CI error?
https://github.com/espnet/espnet/actions/runs/12876705297/job/35900083385?pr=6017#step:8:406

wangpuup:
@sw005320 Error has been fixed. Thank you!

sw005320:
Thanks a lot.
The PR looks good, and I merged this.

One comment (you may consider them in the follow-up PR)
- I'm concerned that the training and test data have the same text since the performance is too good. Can you check it?

wangpuup:
The current train/test/dev splits are defined in conf/file_list. In each wav ID (e.g., "fabm2ag1"), the first four initials represent the speaker id ("fabm"), the next three letters represent the reading material "2ag", and the last digit indicates whether the file is manually aligned. The audio files in train, test and dev splits are from different speakers, but there is overlapping reading material between the splits. I will make a new split.

sw005320:
Where does this split come from?
Is it official?
If so, we may keep it, but the issue is that e2e models memorize the output text pattern.
So, we may also conduct an experiment with our own split that removes the text overall and see how it works.
You may refer to some scripts in espnet (e.g., https://github.com/espnet/espnet/blob/master/egs/voxforge/asr1/local/split_tr_dt_et.sh)

wangpuup:
The split is from @balaji1312's share, and I think it's official.
Sure, I will make a separate split for some experiments.

balaji1312:
I believe the original split is based on prior work and was designed to ensure no speaker overlap. However, as @wangpuup  pointed out, the CMU Kids corpus includes identical prompts, leading to overlap in reading material across the splits. I agree that performing additional experiments on a new split without prompt overlap between the splits would be a good approach. The only consideration is whether this might significantly reduce the size of the training data, which we should carefully check.

Issue #5973, opened 2024-12-06
inference with owsm_v3.1_ebf_small


bloodraven66:
I am decoding with [owsm 3.1 small model](https://huggingface.co/espnet/owsm_v3.1_ebf_small) and having the following error while building the model - 

TypeError: __init__() got an unexpected keyword argument 'attention_qk_norm'

I see that the config file of the pretrained model has attention_qk_norm in ebranchformer encoder config, but current version of ebranchformer no longer has this. Should this be changed to [qk_norm](https://github.com/espnet/espnet/blob/master/espnet2/asr/encoder/e_branchformer_encoder.py)?

 

**Basic environments:**
 - Python 3.8.20 (default, Oct 15 2024, 11:13:24) 
 - espnet version: 202412 (latest as of December 5th 2024)
 - pytorch version - 2.3.0+cu121

**Task information:**
 - Task: S2T
 - Recipe: librispeech
 - ESPnet2

**To Reproduce**
1. download owsm 3.1 small pretrained model - huggingface_hub.snapshot_download(repo_id="espnet/owsm_v3.1_ebf_small")
2. fix paths for feats_stats.npz and bpe.model in pretrained config
3. run inference

```
dataset="librispeech"
test_sets="$dataset/test-clean $dataset/test-other"
inference_config=conf/decode_ctc1_bs1_lang-en_task-asr.yaml

./s2t.sh \
    --stage 12 \
    --lang en \
    --ngpu 0 \
    --gpu_inference false \
    --inference_nj 8 \
    --feats_type raw_copy \
    --use_lm false \
    --skip_data_prep false \
    --inference_s2t_model "valid.total_count.ave_5best.pth" \
    --skip_train true \
    --s2t_exp "<path>/s2t_train_s2t_ebf_conv2d_size768_e9_d9_piecewise_lr5e-4_warmup60k_flashattn_raw_bpe50000" \
    --inference_config "${inference_config}" \
    --test_sets "${test_sets}" "$@"
```
```
beam_size: 1
penalty: 0.0
maxlenratio: 0.0
minlenratio: 0.0
ctc_weight: 1.0
lm_weight: 0.0
lang_sym: <eng>
task_sym: <asr>
predict_time: false
```

**Error logs**
    main()
  File "/mnt/matylda4/udupa/exps/ctc_spec_decoding/espnet/espnet2/bin/s2t_inference.py", line 1103, in main
    inference(**kwargs)
[...]

PR #6025, opened 2025-01-23
fix: OWSM v3.1 - remove flash attention args

kalvinchang:
## What?

* In https://github.com/espnet/espnet/issues/5973, a user reported that ESPnetS2TModel (for OWSM v3.1) could not be instantiated
    * I ran into this issue too
    * The solution is to remove the flash attention arguments regarding qk_norm from the model
* I added a small note that the current configuration is for the medium model in case anyone is confused

## Why?
* flash attention arguments are not fields of the underlying ESPnetS2TModel
* these arguments were removed in HuggingFace (https://huggingface.co/espnet/owsm_v3.1_ebf_small/commit/e2f4e38bfb4b8319858ba910eb3af3eff272d3b9#d2h-406866) but not ESPnet

### DIFF
--- a/egs2/owsm_v3.1/s2t1/conf/train_s2t_ebf_conv2d_size1024_e18_d18_piecewise_lr2e-4_warmup60k_flashattn.yaml
+++ b/egs2/owsm_v3.1/s2t1/conf/train_s2t_ebf_conv2d_size1024_e18_d18_piecewise_lr2e-4_warmup60k_flashattn.yaml
@@ -76,6 +76,7 @@ decoder_conf:
     positional_dropout_rate: 0.1
     self_attention_dropout_rate: 0.1
     src_attention_dropout_rate: 0.1
+    use_flash_attn: true
 
 model_conf:
     ctc_weight: 0.3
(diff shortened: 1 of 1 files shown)


egs2/owsm_v3.1/s2t1/conf/train_s2t_ebf_conv2d_size1024_e18_d18_piecewise_lr2e-4_warmup60k_flashattn.yaml
```diff
@@ -68,10 +76,6 @@ decoder_conf:
     positional_dropout_rate: 0.1
     self_attention_dropout_rate: 0.1
     src_attention_dropout_rate: 0.1
-    self_attention_qk_norm: false   # qk norm
-    src_attention_qk_norm: false    # qk norm
-    self_attention_use_flash_attn: true
-    src_attention_use_flash_attn: true
```
pyf98, line 74: You can add `use_flash_attn: true` to the decoder as well.

pyf98:
Thanks for fixing the config.

pyf98:
LGTM!

sw005320:
Thanks!

PR #6030, opened 2025-01-30
Update tools/README.md

popcornell:
I think that the most failsafe installation procedure should fix python version of conda env to 3.8.  Tried a bunch of times for me it breaks with 3.10 and 3.12 always run into different problems


sw005320:
Make sense.
Actually, our demo also uses `3.8`

popcornell:
The documentation in https://espnet.github.io/espnet/installation.html also fixes it to 3.8 
I think also here is good to keep it this way.

sw005320:
Thanks!

PR #6031, opened 2025-01-31
Fix minor bug in Makefile

juice500ml:
## What?
Add `touch numpy.done` after numpy installation finishes normally.

## Why?
After numpy installation, Makefile did not indicate that installation was finished.
Hence, rerunning Makefile kept numpy to be reinstalled even after installation.


sw005320:
Thanks!

PR #6032, opened 2025-02-02
Create recipe for myst_ogi_cmu_kids

anyuyay:
Create recipe for myst_ogi_cmu_kids.


sw005320:
Can we come up with a better recipe name?
This is a bit too long.

anyuyay:
> Can we come up with a better recipe name? This is a bit too long.

Could it be kids_speech or moc_kids? moc_kids stands for myst+ogi+cmu

sw005320:
cool
kids_speech is good for me for now.
- we may think of better names later
- please add Jibo's test set
- add a description that this recipe is a combination of them.

PR #6038, opened 2025-02-09
Create svs1 recipe for gtsinger

Arllan-lanliu:
## What?

create svs1 recipe for GTSinger dataset


egs2/gtsinger/svs1/conf/tuning/train_xiaoice.yaml
```diff
@@ -0,0 +1,72 @@
+
+
+##########################################################
+#                  SVS MODEL SETTING                     #
+##########################################################
```
ftshijt, line 5: Please reduce the configurations (only keep the one that you used for the model training)/

egs2/gtsinger/svs1/svs.sh
```diff
@@ -0,0 +1,1219 @@
+#!/usr/bin/env bash
```
ftshijt, line 1: Please use symlink instead of copying the template file to here.

PR #6045, opened 2025-02-22
fix some minor errors in SID recipe

shimhz:
## What?

Minor fixes for SID recipe 
- preprocessor setting in no data augmentation condition
- speech shape mismatch when extracting test features

### DIFF
--- a/egs2/TEMPLATE/spk1/spk.sh
+++ b/egs2/TEMPLATE/spk1/spk.sh
@@ -497,7 +497,7 @@ if [ ${stage} -le 6 ] && [ ${stop_stage} -ge 6 ]; then
             --data_path_and_name_and_type ${_inference_dir}/trial.scp,speech,sound \
             --data_path_and_name_and_type ${_inference_dir}/trial2.scp,speech2,sound \
             --data_path_and_name_and_type ${_inference_dir}/trial_label,spk_labels,text \
-            --shape_file ${spk_stats_dir}/valid/speech_shape \
+            --shape_file ${spk_stats_dir}/test/speech_shape \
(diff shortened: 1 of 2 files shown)


sw005320:
Thanks!
LGTM.

@Jungjee, can you take a look at them?

egs2/TEMPLATE/spk1/spk.sh
```diff
@@ -497,7 +497,7 @@ if [ ${stage} -le 6 ] && [ ${stop_stage} -ge 6 ]; then
             --data_path_and_name_and_type ${_inference_dir}/trial.scp,speech,sound \
             --data_path_and_name_and_type ${_inference_dir}/trial2.scp,speech2,sound \
             --data_path_and_name_and_type ${_inference_dir}/trial_label,spk_labels,text \
-            --shape_file ${spk_stats_dir}/valid/speech_shape \
+            --shape_file ${spk_stats_dir}/test/speech_shape \
```
Jungjee, line 500: Thanks, @shimhz, but can you revert this line? 
I get where this is coming from, but I think this would better be dealt in another PR with other changes (which I'll need to work on sometime soon..)

Jungjee (commented):
Thanks for pointing out the corner case in RIR augmentation configuration! 

Can you revert the one line change in `spk.sh`? Other than that, LGTM!

egs2/TEMPLATE/spk1/spk.sh
shimhz: I understand what you mean. I reverted that line.

sw005320:
Thanks, @shimhz!

PR #6053, opened 2025-03-07
New SSL Recipe

wanchichen:
## What?

The current SSL recipe/code is very HuBERT-oriented. This PR rewrites it to support any generic SSL task, along with inference with XEUS.

## Major Changes
- Rename previous `ssl1` recipe to `hubert1`. -> HuBERT has many complicated offline steps, so I think its fine to treat it separately. This is also more inline with the actual implementation, which lives in `espnet2/hubert`
- Create new `ssl1` recipe for generic self-supervised learning.
- Create new `ssl` task. The `espnet_model` for this task follows this framework:
    - It contains a `frontend`, `preencoder`, `encoder` as the main modules. 
    - It is trained with 1 or more SSL Losses, which is defined in `espnet2.loss`.
       - Each defined SSL Loss has 2 attributes that need to be defined:
           - `util_attributes`: functions that must be performed for the task to work. For example, masking in hubert and EMA in data2vec. Different tasks may use the same functions, so we need to pre-define this to prevent redundant computation.
           - `required_inputs`: the data needed to perform loss calculation. For example, reconstrunction losses may also require the frontend feats, while HuBERT only needs the encoder output.

## Minor Changes
- Explicitly set `weights_only=False` for `torch.load()`, which is needed in PyTorch 2.6
- Average n best ckpts in deep speed trainer


sw005320:
@simpleoier, I'm not sure you have some bandwidth, but can you review this PR if possible?

sw005320:
@ftshijt, can you review this PR?

espnet/nets/pytorch_backend/transformer/embedding.py
sw005320, line 1: add a unit test

espnet2/train/deepspeed_trainer.py
sw005320, line 1: @jctian98, can you review the PR related to deepspeed?

espnet/nets/pytorch_backend/transformer/embedding.py
sw005320: if test/espnet2/ssl/test_ssl_espnet_model.py can handle it, it is fine.

espnet2/train/collate_fn.py
sw005320, line 1: This seems to have a lot of changes.
Can you explain these changes?

espnet2/ssl/utils/mask.py
sw005320, line 1: can you add some simple tests?

sw005320:
The PR looks good to me.
I want to have some clarifications about your changes.
I hope the others also review this PR quickly.

espnet/nets/pytorch_backend/transformer/embedding.py
wanchichen: I have a simple test in https://github.com/espnet/espnet/blob/63d38096d088ff11ff6cd290813b1700560bc829/test/espnet2/ssl/test_ssl_espnet_model.py#L26

espnet2/train/collate_fn.py
wanchichen: The most complex part comes from WavLM / Derverb. (lines 92 to 203)

Lines 92 to 128 is just loading the noises 

https://github.com/espnet/espnet/blob/63d38096d088ff11ff6cd290813b1700560bc829/espnet2/train/collate_fn.py#L155 performs reverb augmentation.

https://github.com/espnet/espnet/blob/63d38096d088ff11ff6cd290813b1700560bc829/espnet2/train/collate_fn.py#L175 does WavLM style

And the changes in `__call__()` are from:
- Calling the above functions for augmentation
- Consider cases where we do not have input `text` (ie. any non-HuBERT style SSL training)

sw005320: I see.
Can you add some comments to them?

egs2/librispeech/ssl1/run.sh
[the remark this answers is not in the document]
wanchichen: There is some redundancy, but I thought it would be better to separate them for a better user experience.

HuBERT has additional stages for
- Feature extraction
- K-means
- K-means inference

Which can be skipped for other models. Also, the design of `hubert1` has a loop that runs the following for each Hubert iteration
1. K-means training and inference (above)
2. Collect stats
3. SSL model training

This is good for HuBERT reproducibility since it takes care of all steps from start to finish, but it makes the recipe very complex and hard to modify  (and maybe confusing) for users doing other SSL training techniques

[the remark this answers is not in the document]
sw005320: Make sense.
How about adding this description in README.md under the template in ssl/hubert?
Please also add them to the task or other appropriate places in multiple sources to avoid confusions for the users. 
This clarification helps people to understand these two different (but related) tasks.

egs2/TEMPLATE/hubert1/README.md
```diff
@@ -0,0 +1,3 @@
+# Self-supervised Learning
```
simpleoier, line 1: Do you think it's better to add hubert specific information in this file?

egs2/librispeech/ssl1/run.sh
```diff
@@ -7,44 +7,19 @@ set -o pipefail
...
 
-train_config_iter0=conf/tuning/train_ssl_torchaudiohubert_base_960h_pretrain_it0.yaml
-train_config_iter1=conf/tuning/train_ssl_torchaudiohubert_base_960h_pretrain_it1.yaml
-train_config_iter2=conf/tuning/train_ssl_torchaudiohubert_large_960h_pretrain_it2.yaml
+train_config=conf/tuning/train_hubert.yaml
 
-./hubert.sh \
-    --ngpu 8 \
+./ssl.sh \
+    --stage 7 \
```
simpleoier, line 16: Does this recipe starts from stage 7?

simpleoier (approved):
Nice! This PR LGTM! Some newly added modules are lack of unit tests. Hope you can add them in the following PRs.

egs2/TEMPLATE/ssl1/ssl.sh
```diff
@@ -0,0 +1,585 @@
...
+train_set=     # Name of training set
+valid_set=     # Name of valid set
+lang=noinfo    # The language type of corpus.
+speech_fold_length=800 # fold_length for speech data during SSL training.
+text_fold_length=400   # fold_length for text data during SSL training.
+
+help_message=$(cat << EOF
+Usage: $0 --train-set "<train_set_name>" --valid-set "<valid_set_name>" --test_sets "<test_set_names>"
+
+Options:
```
ftshijt, line 85: minor comment: please remember to align the help message to the above arguments.

espnet2/ssl/espnet_model.py
```diff
@@ -0,0 +1,258 @@
...
+            pad_masks = make_pad_mask(speech_lengths).to(speech.device)
+            # Prioritize block masking if both are available
+            if "block_mask" in self.util_attributes:
+                speech, mask_info = self.util_modules["block_mask"](speech, pad_masks)
+            elif "mask" in self.util_attributes:
+                speech, mask_info = self.util_modules["mask"](speech, pad_masks)
+            data["mask_info"] = mask_info
+
+        # Flow Matching
+        # TODO: NOT IMPLEMENTED YET
```
ftshijt, line 193: minor comment, leaving TODO with your name ;)

ftshijt (commented):
Many thanks for the PR!

I left some minor comments, mostly related to documentation side. The design looks great to me.

In addition, it would be super helpful if you can include some examples for using the pre-trained SSL from the recipe for downstream applications. But we may target that in a later PR.

egs2/librispeech/ssl1/run.sh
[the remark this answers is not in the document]
wanchichen: just added!

espnet2/train/deepspeed_trainer.py
sw005320: @jctian98, this is a reminder.
This PR is very important and I want your review as soon as possible.

espnet2/ssl/espnet_model.py
```diff
@@ -0,0 +1,235 @@
...
+from espnet2.asr.specaug.abs_specaug import AbsSpecAug
+from espnet2.layers.abs_normalize import AbsNormalize
+from espnet2.ssl.loss.abs_loss import AbsSSLLoss
+from espnet2.torch_utils.device_funcs import force_gatherable
+from espnet2.train.abs_espnet_model import AbsESPnetModel
+from espnet.nets.pytorch_backend.nets_utils import make_pad_mask
+
+
+class ESPnetSSLModel(AbsESPnetModel):
+    """A generic SSL model"""
```
sw005320, line 25: Please add more explanations here (or other appropriate spaces).
What kind of SSL models are supported?

espnet2/ssl/utils/mask.py
sw005320: Did you add a test?

egs2/librispeech/ssl1/conf/deepspeed.json
```diff
@@ -0,0 +1,47 @@
+{
+    "train_micro_batch_size_per_gpu": 1,
+      "gradient_accumulation_steps": 1,
+      "gradient_clipping": 5.0,
+      "bf16": {
```
jctian98, line 5: BF16 is only suitable for Ampere and above. Add a comment here?

jctian98:
I checked the PR mainly for DeepSpeed-related features. Specifically (1) deepspeed_trainer; (2) model average; (3) related config file. I think the implementation is correct. 

From the reproducibility perspective:
(1) Can we provide guidance on how to install deepspeed and related plugins somewhere? like README.md file. I think people usually get in trouble with plugin compilations. 
(2) We can possibly provide a config file without deepspeed. Just ensure it can run quickly (but remind people we don't exact use that config file)
(3) For the HF models we need some extra setup for activation checkpointing besides in the deepspeed.json. If you also do the checkpointing, could you point out where that setup is? (usually a setup in modeling file)

egs2/librispeech/ssl1/conf/deepspeed.json
wanchichen: I put it in the readme instead. I dont think json supports comments natively.

espnet2/ssl/espnet_model.py
wanchichen: Done! For now its only hubert. Data2vec and dinosr are finished implementing, i am just verifying by training now.

espnet2/ssl/utils/mask.py
wanchichen: The original test already covered the usage a bit, but i added another to double check if the masking was done correctly.

wanchichen:
Thanks everyone for the comments! 

Downstream usage (tokenization and fine-tuning) will be in a separate PR.

> Can we provide guidance on how to install deepspeed and related plugins somewhere? 
I added some details in the librispeech recipe readme. But it will be difficult to make complete documentation since the installation for those packages are very system dependent.

> We can possibly provide a config file without deepspeed.
I added `train_hubert_amp.yaml` as a non-deepspeed config.

>  If you also do the checkpointing, could you point out where that setup is?
Oh i dont do checkpointing here since the models are very small. We can follow up in another PR.

sw005320:
Thanks!!

PR #6054, opened 2025-03-07
Add optional ESPnet preprocessor support to ESPnetEZ

HANJionghao:
## What?

- Added optional support for the ESPnet preprocessor in ESPnetEZ through the new "use_ez_preprocessor" argument in the config. 
- This change does not affect the existing functionality, as it introduces an additional configuration option.

## Why?

SVS preprocessing is complex. This enhancement simplifies the demo by providing an easier way to use default ESPnet preprocessors.

## See also

SVS demo PR: https://github.com/espnet/notebook/pull/52
This demo PR has been updated to reflect the introduction of "use_ez_preprocessor" and the code changes from the original branch in the previous PR.


sw005320:
@Masao-Someki, can you finish your review?

espnetez/dataset.py
```diff
@@ -138,10 +150,13 @@ def names(self) -> Tuple[str, ...]:
...
     def __getitem__(self, uid: Union[str, int]) -> Tuple[str, Dict]:
         idx = int(uid)
-        return (
-            str(uid),
-            {k: v(self.dataset[idx]) for k, v in self.data_info.items()},
-        )
+        if self.preprocess is None:
+            data = {k: v(self.dataset[idx]) for k, v in self.data_info.items()}
+        else:
+            data = self.preprocess(
```
Masao-Someki, line 156: Just for clarification.
This `data` from the preprocess function is the same format as the data in line 154?
```
data = {k: v(self.dataset[idx]) for k, v in self.data_info.items()}
```

Masao-Someki:
Hi, thank you @HANJionghao for this PR! I added my review.

espnetez/dataset.py
HANJionghao: Yes. These are the same format.

Masao-Someki (approved):

Masao-Someki (approved):

sw005320:
Can you check the CI error?
https://github.com/espnet/espnet/actions/runs/14115976549/job/39546201953?pr=6054

Masao-Someki:
@HANJionghao 
I checked the CI error, and I think the `preprocess` function should be placed before the `data = {k: ...}` line. For the `text`, it's already being tokenized in the `v(...)` process, so I suspect that passing the tokenized result to the preprocessor is what's causing the issue.

HANJionghao:
In the PR code, the preprocessor is called if use_ez_preprocessor=True is explicitly set in the config. Since this is a new config argument, it should not affect previous scripts. Investigating this CI issue right now

Masao-Someki:
@HANJionghao 
Hi, is it possible to add unit test on this feature? Codecov mentioned that there is no tests on the newly introduced part.

HANJionghao:
@Masao-Someki Unit test added and functionality verified. Let me know if anything else is needed for this PR.

Masao-Someki:
Thank you @HANJionghao, this looks good to me. Some CI checks aren’t running, but it seems the same issue exists on the main branch as well.

Issue #6057, opened 2025-03-08
Broken Link in ESPnet Documentation: 404 Error for Fine-Tuning Guide


hayabusa255:
I attempted to perform fine-tuning following the guidance provided in your documentation. However, I encountered a 404 error when accessing the following URL:

https://espnet.github.io/espnet/jvs/tts1/

Could you please provide the correct link or update the documentation accordingly?

Masao-Someki:
Thank you for reporting the issue, and apologies for the inconvenience. Here is the correct link:  
[https://github.com/espnet/espnet/blob/master/egs2/jvs/tts1/README.md](https://github.com/espnet/espnet/blob/master/egs2/jvs/tts1/README.md)  

We'll update the document soon!

PR #6058, opened 2025-03-10
Bugfix #6057

Masao-Someki:
## What?

Fixed link of TTS recipe

## Why?

Bugfix of #6057

## See also

-


sw005320:
Thanks, @Masao-Someki!

PR #6064, opened 2025-03-17
A quick fix for type error when dealing with multi-decoder (ST)

ftshijt:
With the latest typeguard, it will spot the type issue in multi-decoder decoding (when both ASR and ST search are returned together).

Adding a simple fix to allow a tuple type for the multi-decoder return cases.

(Noted that this fix is necessary to make sure the demonstration notebook is working smoothly)


sw005320:
Thanks, @ftshijt!

PR #6070, opened 2025-03-22
[fix] locale match bug

TomohikoNakamura:
## What?
I fixed a minor bug in `egs2/TEMPLATE/asr1/utils/validate_data_dir.sh`.
Replace [Line 129](https://github.com/espnet/espnet/blob/7c38a90360edd657ea4cab99779eae3f831256e0/egs2/TEMPLATE/asr1/utils/validate_data_dir.sh#L129) with `if locale -a | grep -E "^C\.(UTF-8|utf8)$" >/dev/null; then`.
This matches both `C.UTF-8` and `C.utf8`.

## Why?
An alert arises that libritts/dev-other text includes unprintable characters. It can be confirmed in [egs2/voxtlm_v1/lm1](https://github.com/espnet/espnet/tree/master/egs2/voxtlm_v1/lm1/).

I identified that this bug is caused by [Line 129](https://github.com/espnet/espnet/blob/7c38a90360edd657ea4cab99779eae3f831256e0/egs2/TEMPLATE/asr1/utils/validate_data_dir.sh#L129) of `egs2/TEMPLATE/asr1/utils/validate_data_dir.sh`.
In recent Ubuntu, `locale -a` yields like below.
```bash
user@machine:~$ locale -a
C
C.utf8
POSIX
en_US.utf8
ja_JP.utf8
```
It does not match the current implementation.
This falls `LC_ALL="en_US.UTF-8" grep -c '[^[:print:][:space:]]' $data/text`. And, libritts/dev-other text includes unprintable characters with `en_US.UTF-8`, which raises the error.
I confirmed that the text consists of all printable characters with `LC_ALL=C.UTF-8`.


egs2/TEMPLATE/asr1/utils/validate_data_dir.sh
```diff
@@ -126,7 +126,7 @@ fi
 num_utts=`cat $tmpdir/utts | wc -l`
 if ! $no_text; then
   if ! $non_print; then
-    if locale -a | grep "C.UTF-8" >/dev/null; then
+      if locale -a | grep -E "^C\.(UTF-8|utf8)$" >/dev/null; then
```
sw005320: The indent seems to be wrong.
Can you fix it?

sw005320:
This will break CI due to https://github.com/espnet/espnet/blob/master/ci/check_kaldi_symlinks.sh#L23-L25
(We try to keep these Kaldi-oriented scripts the same as those in Kaldi.)

It seems that this is critical, and we may need to maintain it by ourselves.
In this case, we may exclude this one from the above CI check (by skipping them like `utils/README.md` in https://github.com/espnet/espnet/blob/master/ci/check_kaldi_symlinks.sh#L16C19-L18

Can you fix it?

Fhrozen:
This pull request makes a minor update to the locale detection logic in the `validate_data_dir.sh` script to improve compatibility with different systems.

- Improved locale detection by updating the `grep` pattern to match both `C.UTF-8` and `C.utf8` variants, ensuring broader compatibility when setting the locale (`L`) variable.

PR #6067, opened 2025-03-18
Additional results for the discrete ASR challenge

juice500ml:
## What?
- asr1 recipe results for interspeech2024_dsu_challenge dataset
  - It contains layerwise results
- Add results of beam size=5 for quicker inference in asr2 recipe

## Why?

- It contains the results of the recent paper submission.

PR #6071, opened 2025-03-25
Downgrade Transformers version

Fhrozen:
## What?

Limit max version of Transformers

## Why?

The current version of transformer `4.50.0`, seems to be developed for newest pytorch versions (`>=2.3.0`)

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -84,7 +84,7 @@ requirements = {
         "torchaudio",
         "torch_optimizer",
         "fairscale",
-        "transformers<4.50.0",
+        "transformers",
         "evaluate",
     ],
     "setup": [
--- a/setup.py
+++ b/setup.py
@@ -84,7 +84,7 @@ requirements = {
         "torchaudio",
         "torch_optimizer",
         "fairscale",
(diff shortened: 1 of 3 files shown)


Fhrozen:
Planned to fix #6067

sw005320:
Thanks @Fhrozen!

Vaibhavs10 (commented):
hey hey - https://github.com/huggingface/transformers/releases (4.50.1) should fix this, so you should not need it.

sw005320:
Thanks, @Vaibhavs10!
@Fhrozen, can you remove the version restriction?

Fhrozen:
Version 4.50.1 fixes the `flatten_with_keys_fn` import error. So, I suppose that pytorch versions higher than `2.0.0`, should not have that issue. But, for `1.13`, the restriction is needed, because it does not have a the `compiler` folder.
I will check again and test if no issue is presented when the restriction is removed.

sw005320:
Maybe, we can remove the CI support for 1.13 in the near future.

PR #6083, opened 2025-04-01
Added Harvest algorithm as an option for F0 extraction (minor feature)

f-matano44:
## What?

Added Harvest algorithm as an option for F0 extraction.
As Harvest is also part of PyWorld like DIO, it has been implemented by extending the existing DIO class.

## Why?

Harvest is slower than DIO, but offers higher accuracy in F0 extraction, making it a suitable choice as an F0 extraction option.

## See also

As a reference, the following paper is available:
https://doi.org/10.21437/Interspeech.2017-68

### DIFF
--- a/espnet2/tasks/gan_svs.py
+++ b/espnet2/tasks/gan_svs.py
@@ -34,6 +34,7 @@ from espnet2.train.gan_trainer import GANTrainer
 from espnet2.train.preprocessor import SVSPreprocessor
 from espnet2.tts.feats_extract.abs_feats_extract import AbsFeatsExtract
 from espnet2.tts.feats_extract.dio import Dio
+from espnet2.tts.feats_extract.harvest import Harvest
 from espnet2.tts.feats_extract.energy import Energy
 from espnet2.tts.feats_extract.linear_spectrogram import LinearSpectrogram
 from espnet2.tts.feats_extract.log_mel_fbank import LogMelFbank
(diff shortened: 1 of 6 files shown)


sw005320:
Thanks!
@ftshijt, can you check this PR?

ftshijt:
Thanks for the update! The current implementation looks great to me. May I ask if there is any example pre-trained model with the updated pitch extractor?

f-matano44:
@ftshijt Thank you for reviewing my PR.
Unfortunately, I don't have any pre-trained model available to share.

sw005320:
@f-matano44

I think this is a good addition, but if we don't have an example, people may not be able to use it.
If possible, can you run one of the SVS recipes with harvest as an example and share the recipe and the corresponding recipe?

If @f-matano44 says that it is difficult to conduct such an experiment, @ftshijt, can we ask some of our colleagues to execute this?

f-matano44:
@sw005320 @ftshijt 

Thank you very much for your feedback.

Unfortunately, I currently don’t have access to a GPU powerful enough to run the SVS recipe and share the results.
Would it be possible for you to run the experiment on my behalf?

Switching to harvest-based pitch extraction is straightforward: simply update the configuration file as follows:

```diff
- pitch_extract: dio
+ pitch_extract: harvest
```

Thank you in advance for your help.

HANJionghao:
@f-matano44 Thanks again! Since online Harvest pitch extraction is slowing down training quite a lot, would it be possible to include the offline extraction as part of this PR?

Issue #6079, opened 2025-03-31
[OWSM fine-tunening]AttributeError: module 'numpy' has no attribute 'dtypes'


GoGoAsahi:
Hi, I'm trying to fine-tune the OWSM model following a [Tutorial.](https://espnet.github.io/espnet/notebook/ESPnetEZ/ASR/ASR_finetune_owsm.html) I've run into trouble which is AttributeError: module 'numpy' has no attribute 'dtypes'. My research environment is Google Colab. I wrote this code:

`!pip install espnet`
import os
from glob import glob
import numpy as np
import librosa

import torch
from espnet2.bin.s2t_inference import Speech2Text
from espnet2.layers.create_adapter_fn import create_lora_adapter
import espnetez as ez

DUMP_DIR = f"./dump"
CSV_DIR = f"./transcription"
EXP_DIR = f"./exp/finetune"
STATS_DIR = f"./exp/stats_finetune"
FINETUNE_MODEL = "espnet/owsm_v3.1_ebf"
LORA_TARGET = [
    "w_1", "w_2", "merge_proj"
]
LANGUAGE = "jpn"`

This is the error code:
---------------------------------------------------------------------------
AttributeError                            Traceback (most recent call last)
<ipython-input-4-acfc482ee4b3> in <cell line: 0>()
      6 
      7 import torch
----> 8 from espnet2.bin.s2t_inference import Speech2Text
      9 from espnet2.layers.create_adapter_fn import create_lora_adapter
     10 import espnetez as ez

25 frames
/usr/local/lib/python3.11/dist-packages/espnet2/bin/s2t_inference.py in <module>
     16 from espnet2.asr.partially_AR_model import PartiallyARInference
     17 from espnet2.fileio.datadir_writer import DatadirWriter
---> 18 from espnet2.tasks.lm import LMTask
     19 from espnet2.tasks.s2t import S2TTask
     20 from espnet2.text.build_tokenizer import build_tokenizer

/usr/local/lib/python3.11/dist-packages/espnet2/tasks/lm.py in <module>
     13 from espnet2.lm.seq_rnn_lm import SequentialRNNLM
     14 from espnet2.lm.transformer_lm import TransformerLM
---> 15 from espnet2.tasks.abs_task import AbsTask
     16 from espnet2.text.phoneme_tokenizer import g2p_choices
     17 from espnet2.torch_utils.initialize import initialize

[...]

PR #6086, opened 2025-04-06
Update numpy version

Masao-Someki:
## What?

Update numpy version in setup.py from `numpy<1.24` to `numpy<2.0.0`

## Why?

#6079 

## See also

#6079


sw005320:
Should we keep this PR?
What is the situation with the numpy version handling?
@Fhrozen, @ftshijt

Fhrozen:
What is the situation with the numpy version handling?
There are so many package that uses versions lower than `1.26` so even if we move it to `<2.0`, later during the installation, when other package requires it, the numpy package will be downgraded. 
This does not really affect the installation, so can be even merged.

PR #6088, opened 2025-04-10
ACE-Opencpop Data Preparation Quick Fix

HANJionghao:
## What?

- Removed song 2026 from the ACE-Opencpop (i.e., acesinger) dataset processing script.
- Performed minor code cleanup and variable renaming for clarity.

## Why?

- The audio of song 2026 in the ACE-Opencpop dataset is misaligned with its score. We excluded this song from the processed dataset to maintain data quality and avoid introducing noise.

## See also

None


ftshijt:
Many thanks for the update!

PR #6098, opened 2025-04-22
[SVS1]: update inference and evaluation

South-Twilight:
## What?

1. add inference with music score (syllabel level)
2. add versa evaluation for SVS1
3. update recipes: rename ameboshi to amaboshi, update acesinger
## Why?

Fix bugs and add new function.

## See also

Part of updates in [PR 6019](https://github.com/espnet/espnet/pull/6019).


egs2/TEMPLATE/asr1/pyscripts/utils/evaluate_token_accuracy.py
```diff
@@ -0,0 +1,105 @@
...
+                name = names[i]
+                if args.mix_type == "frame":
+                    name_gen_token = gen_token[i :: args.discrete_token_layers]
+                    name_gt_token = gt_token[i :: args.discrete_token_layers]
+                elif args.mix_type == "sequence":
+                    name_gen_token = gen_token[i * token_len : (i + 1) * token_len]
+                    name_gt_token = gt_token[i * token_len : (i + 1) * token_len]
+                acc = sum(gen == gt for gen, gt in zip(name_gen_token, name_gt_token))
+                acc = acc / token_len
+                acc_dict[name][key] = acc
```
ftshijt, line 84: It is fine to do per-utterance accuracy, but usually, we summarize and conduct a micro-average for this instead of a macro average. If you can find any reference using macro average here, we might keep it otherwise, I would suggest that we can use micro-average by summing all correct/error frames and then do the division.

ftshijt:
Thanks for the updates. Since the token accuracy is not a major implementation, I will first merge the PR. However, I would still recommend a future update to support the micro average as well.

PR #6105, opened 2025-04-29
VISinger2 Duration Loss Quick Fix

HANJionghao:
## What?

- Fixed a bug in espnet2/gan_svs/vits/vits.py where score_dur_loss mistakenly used phoneme-level durations as targets, same as phoneme_dur_loss. Changed the target of score_dur_loss from gt_dur to score_dur.

## Why?

- Currently, both channels in pred_dur are trained to predict phoneme durations, which are duplicated. This fix corrects the second channel to predict score-level durations and aligns with the original VISinger2 paper.

## See also

None


ftshijt:
Thanks for the fix!

Issue #6112, opened 2025-05-15
Lagacy torch tensor constructor causes issue when using LM for scoring


advaitvd:
**Torch Version I'm using:** _2.6.0+cu126_

**Issue Description:**

https://github.com/espnet/espnet/blob/ac2a7a422f04ef2904560d8f36d7c1226f9c4bcf/espnet2/asr/transducer/beam_search_transducer.py#L338
Lagacy constructor expects the device type to be cpu thus causing issue when the AM decoder is on a gpu during inference. The above code should be updated to
```
torch.tensor(
  ...
  dtype=torch.long
)
```

**Error Message:**
```
Lagacy constructor for initializing Long Tensor needs to be updated.
Traceback (most recent call last):                                                                                                
  File "/opt/miniconda/lib/python3.9/site-packages/gradio/queueing.py", line 536, in process_events                               
    response = await route_utils.call_process_api(                                                                                
  File "/opt/miniconda/lib/python3.9/site-packages/gradio/route_utils.py", line 322, in call_process_api                          
    output = await app.get_blocks().process_api(                                                                                  
  File "/opt/miniconda/lib/python3.9/site-packages/gradio/blocks.py", line 1935, in process_api                                   
    result = await self.call_function(                                                                                            
  File "/opt/miniconda/lib/python3.9/site-packages/gradio/blocks.py", line 1520, in call_function                                 
    prediction = await anyio.to_thread.run_sync(  # type: ignore                                                                  
  File "/opt/miniconda/lib/python3.9/site-packages/anyio/to_thread.py", line 56, in run_sync                                      
    return await get_async_backend().run_sync_in_worker_thread(                                                                   
  File "/opt/miniconda/lib/python3.9/site-packages/anyio/_backends/_asyncio.py", line 2461, in run_sync_in_worker_thread          
    return await future                                                                                                             File "/opt/miniconda/lib/python3.9/site-packages/anyio/_backends/_asyncio.py", line 962, in run                                 
    result = context.run(func, *args)                            
  File "/opt/miniconda/lib/python3.9/site-packages/gradio/utils.py", line 826, in wrapper                                         
    response = f(*args, **kwargs)                                
  File "/app/workspace/ASR/gr_cln.py", line 52, in speech_to_text                                                                 
    n_best = speech2text(audio_data)                             
  File "/opt/miniconda/lib/python3.9/site-packages/torch/utils/_contextlib.py", line 115, in decorate_context                     
    return func(*args, **kwargs)                                 
[...]

PR #6114, opened 2025-05-16
Fix for issue #6112 Lagacy torch tensor constructor causes issue when…

advaitvd:
… using LM for scoring

## What?

Changed all torch tensor initializations in espnet2 source code from the legacy
`torch.LongTensor(...)`
to
`torch.tensor(..., dtype=torch.long)`

## Why?

As described in issue #6112 , the legacy torch tensor constructor `torch.LongTensor(...)`, expects device to be cpu when passed as a constructor argument and throws an error if the device passed is cuda (check #6112 for detailed error message.). Besides, `torch.LongTensor(...)` can cause silent truncation of floats which might lead to bugs which could be hard to find. Switching to `torch.tensor` is the correct and recommended way to initialize tensors as per pytorch documentation.

## See also

Pytorch documentation recommends `torch.tensor` to initialize tensors.
https://docs.pytorch.org/docs/stable/tensors.html#torch.Tensor

### DIFF
--- a/espnet2/svs/feats_extract/score_feats_extract.py
+++ b/espnet2/svs/feats_extract/score_feats_extract.py
@@ -366,7 +366,7 @@ def expand_to_frame(expand_len, len_size, label, midi, duration):
         ListsToTensor(seq_label), dtype=torch.long, device=label.device
     )
     seq_label_lengths = torch.tensor(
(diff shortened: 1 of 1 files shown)


sw005320:
Thanks a lot, @advaitvd!

PR #6123, opened 2025-05-29
Update BSRNN implementations to support more flexible band-split schemes

Emrys365:
## What?

This PR updates the BSRNN implementation to add support of more band-split schemes (e.g., overlapping mel subbands).
In addition, this PR also adds checks of the `norm_type` argument to prevent unintended use when `causal=True`.
> For example, `norm_type="GN"` should not be used in the `causal=True` mode.

## Why?

This should benefit future model designs and may bring some performance gains.


sw005320:
Thanks, @Emrys365!

PR #6129, opened 2025-06-04
Changing order on CI for Whisper installation

Fhrozen:
## What?

Changing order on CI installation.

## Why?

Versa is using the latest whisper version. Whisper new version is not supported by ESPnet and does not have some hyperparameters (such as N_MEL). Changing the installation is an easy hack to fix CI checks.

## See also

Whisper.
ESPnet-3 commits


sw005320:
Thanks!

PR #6131, opened 2025-06-06
Fixed a typo that was causing data leakage.

Abdigal1:
Fixed a typo that was causing data leakage.

## What?

Fixed a typo that was preventing correct filtering of test IDs in `validated.tsv`.

## Why?

The filtering step correctly removed `val_ids`, but due to the typo, it did **not** remove `test_ids`. This caused data leakage, as test data was still present in the training set.

## See also

Aside from the typo, the original intent was to remove `client_ids` associated with the test set. However, since Common Voice includes repeated sentences spoken by different speakers, this method doesn't fully eliminate data overlap.

Therefore, in addition to fixing the typo here, I recommend removing the exact matching sentences from `validated.tsv` that are also present in the test set.
https://discourse.mozilla.org/t/why-train-tsv-includes-a-few-files-just-3-of-validated-set/36471/6
https://discourse.mozilla.org/t/common-voice-v1-corpus-design-problems-overlapping-train-test-dev-sentences/24288


sw005320:
Thanks for pointing out the issue!
@pyf98, can you review this?

pyf98:
Thanks! The discussions make sense to me. But the script was not prepared by me, so I'm not very familiar with this particular issue.

sw005320:
Who prepared it?
You may find it in history.

sw005320:
According to [this](https://github.com/espnet/espnet/commits/master/egs2/owsm_v3/s2t1/local/prepare_commonvoice.py), it seems that @jctian98 prepared this.
Can you confirm it?

sw005320:
@jctian98, can you answer it?

jctian98:
Hi @Abdigal1, thank you for bringing this to our attention. Yes, there is a typo that could cause potential test-set leakage. So, we definitely change `dev_ids -> test_ids` in this PR.

As it's been a while since we made this script, I want to double-check with you some details (in case you know it)

- How much leakage would it cause? Is `test.tsv` a subset of `validated.tsv`, or are they mostly exclusive? 
- Which version of CommonVoice are you using? We were working on `cv-corpus-13.0-2023-03-09`, in which the `.tsv` file used `client_id` rather than `sentence` as the first field. if this is different in different CommonVoice versions, could you add some comments to note this change?


Many thanks!

Abdigal1:
Hey @jctian98 , let me know if there's anything else I need to do for the PR to be approved.

Regarding your questions:

- In the version I used (`cv-corpus-20.0-2024-12-06`), the `test.tsv` file is a subset of `validated.tsv`, and I believe this is true for all versions of the [dataset](https://github.com/common-voice/cv-dataset?tab=readme-ov-file#about-the-dataset).

- All the fields listed [here](https://github.com/common-voice/cv-dataset?tab=readme-ov-file#fields) are present in the dataset. I think you're asking why the filtering is done by  `sentence`, which its not the first field, rather than by  `client_id`. This is because the same sentence can be spoken by multiple people (`client_id`). So, if we don’t filter by sentence, we could end up with the same sentence appearing in both training and testing sets—just spoken by different individuals.

jctian98:
Hi @Abdigal1, many thanks for the reply! This is super helpful. 

The PR looks good to me. @sw005320 

We are actively updating our CI pipeline, so there are some failed CI tests. We may merge it when CI is working properly.

wanchichen:
If you only filter by `sentence`, you are allowing cases where the same speaker appears in both train and test right? 

Ideally, you filter by both `sentence` and `client_id`, but that might be too aggressive and discard too much data.

If you had to choose one, I think filtering by `client_id` is actually correct here, IMO speaker leakage is more of a concern than text leakage. The overall OWSM dataset is so large that its likely there are many cases of common text phrases appearing in both train and test.

Abdigal1:
Hey! Sorry for the delay, @wanchichen . Given the context, it makes sense to me. I'm forking it and updating the filter to use `client_id`.

sw005320:
What is the current status of this PR, @Abdigal1?

sw005320 (approved):
Thanks a lot
This is critical!

PR #6135, opened 2025-06-09
Espnet3/installation

Masao-Someki:
## What?

Installation/CI fix for ESPnet-3 development. Work done by @Fhrozen 

## Why?

1. espnet-3 requires additional dependencies.
2. Since espnet installation contains so many dependencies, and some of them are not required for some tasks. We wanted to clean the dependencies based on use cases.

## See also

#6133 

## Major changes

`.devcontainer/`

- Would like to ask Nelson to explain this part, but basically creating devcontainer for ESPnet-3, and refactoring the code base.

`.github/`

- Addition of utilities to simplify environment setup for CI.

`.pre-commit-config.yaml`

- Commented out to temporarily disable functionality for automatic format execution.
- Will uncomment before merge.

`ci/`

- espnet-2: Mainly format and task additions. Want to run CI in parallel by adding tasks as arguments.
- For espnet-1 related items, changed paths to accommodate folder structure changes in test_utils.

- Removed espnet-ez related CI. (ESPnet-3 CI will be added in the coming PR)

`doc`, `vuepress`

- Homepage changes and documentation addition for devcontainer.

`egs*/espnet*`

- I'm sorry, I will delete items that is still not be reverted for formatting.
- For some changes are not related to formats, so currently asking to Nelson.

`test/` 

- Support `torch>2.6`.

`test_utils/`

- Moved espnet1 resources to `espnet1_scripts`.

`tools/` 
- torch2.7 support
- Changes to makefile to use pyproject.toml instead of setup.py.

Issue #6133, opened 2025-06-09
Development plan for ESPnet-3


Masao-Someki:
We want to develop the ESPnet-3, more pythonic/scalable framework with pytorch lightning and espnet recipes. We also aim to utilize multiple datasources, such as Huggingface Dataset, Lhotse, and pytorch dataset.

Based on ESPnet-EZ, we plan to shift from EZ to 3. Welcome your requests for these particular directions and contributions! The followings are our development plans.

## New Features

- [ ] Pytorch-lightning trainer (based on yifan's implementation)
- [ ] Parallel processing with Dask
- [ ] Data organizer (easily scale to multiple dataset)
- [ ] Inference runner/score runner
- [ ] collect stats

## Recipes

- [ ] TEMPLATE

**asr1**
- [ ] aishell
- [ ] LibriSpeech
- [ ] LibriSpeech-100h

**asr2**

- [ ] 2024 InterSpeech challenge

**enh1**

**s2t1**
- [ ] owsm

**spk1**

**st1**

**tse1**

**tts1**

**tts2**

---
The following tasks are optional as it takes longer time to train.

**codec1**
- [ ] amuse

**speechlm1**

**ssl1**

## Documentation

- [ ] Get Started
- [ ] Installation
- [ ] callbacks
- [ ] collect_stats
- [ ] config
- [ ] contribution
- [ ] data preparation
- [ ] batching
- [ ] multi-gpu/multi-node training
- [ ] tensorboard/wandb

## Others

- [ ] Fix installation
- [ ] Cleanup CI
- [ ] Leaderboard

- [ ] CI for notebooks

Fhrozen:
@dosu-bot, Can you help me here? I will like to make a new branch from PR [#6139](https://github.com/espnet/espnet/pull/6139) that is independent from the master (maybe an orphan branch). Could you guide me through?

Masao-Someki:
As one of the process of ESPnet-3 project, we'd like to remove ESPnet-1 support.

## 1. Steps to remove ESPnet-1
1. [x] List up all packages used from ESPnet-2. 
1. [x] List up all packages used from packages from step 2.(packages in 2)
1. [x] Extract required functions from packages from step 3.
1. [x] Remove ESPnet-1 from CI
1. [x] Decide which directory to move from espnet-1 to 2.
    1. Loop the following
    1. Move one package
    1. Replace import sentence
    1. Run CI
1. [x] Run CI for espnet-2 on local env
[...]

PR #6139, opened 2025-06-11
Espnet3/install only

Masao-Someki:
## What?

Update installation of espnet. With this PR, we can install espnet in a structured way.
E.g., `pip install espnet[all]`, `pip install espnet[core,task-asr]`

## Why?

The dependencies of espnet becomes very large. We wanted to split it.

## See also

#6133
#6135

## Note

This PR might include formatting automatically done by precommit.ci, but that will be merged to the master branch in another PR. This PR should only be merged after that PR.


Masao-Someki:
@Fhrozen Could you review this PR, and if there is any missing files please let me know!!

Fhrozen:
@Masao-Someki Under review.

Masao-Someki:
Thank you @Fhrozen and I didn't know that we had the copilot review! I'll definitely need that in my future PRs!
I'll fix them

Masao-Someki:
@Fhrozen just commented me that the devcontainer-related script should be the another PR, as it is not specifically related on installation.

I will remove those, and after merging the #6140 the number of files should be 5 (document, pyproject.toml, Makefile, setup.py, and .gitignore)

doc/installation.md
sw005320, line 1: Can you integrate the installation documents with the previous one?

`#### 4. Install ESPnet via `pyproject.toml` (Recommended for ESPnet3)` suddenly appears
If you think it's recommended, you should move it first and make the current installation documents under legacy.

.github/workflows/ci_on_ubuntu.yml
```diff
@@ -316,6 +316,7 @@ jobs:
         python3 -m pip install -U numba
         ./tools/installers/install_torch.sh false ${TH_VERSION} CPU
         ./tools/installers/install_chainer.sh CPU
+        cp ./tools/setup.py ./ && mv pyproject.toml pyproject.toml.tmp # && mv espnet3 .espnet3
```
sw005320: This looks hacky.
Why did you move setup.py under `tools/`?

sw005320: Please add the comments

pyproject.toml
```diff
@@ -0,0 +1,139 @@
...
+core = [
+  "torch>=2.6.0",
+  "torchaudio>=2.6.0",
+  "numpy",
+  "soundfile>=0.10.2",
+  "kaldiio>=2.18.0",
+  "espnet_tts_frontend",
+  "sentencepiece==0.2.0",
+  "pyworld>=0.3.4",
+  "g2p_en",
```
sw005320: why TTS related packages (tts_frontend, pywordl, g2p_en) are included in the core?

pyproject.toml
```diff
@@ -0,0 +1,139 @@
...
+task-asr = [
+  "ctc-segmentation>=1.6.6",
+  "editdistance",
+  "h5py",
+  "lhotse",
+  "jamo",
+  "torch_complex",
+  "opt_einsum",
+  "protobuf",
+  "tensorboard",
```
sw005320: why `tensorboard` is only listed in asr?

pyproject.toml
```diff
@@ -0,0 +1,139 @@
...
+  "dask_jobqueue",
+]
+
+task-asr = [
+  "ctc-segmentation>=1.6.6",
+  "editdistance",
+  "h5py",
+  "lhotse",
+  "jamo",
+  "torch_complex",
```
sw005320: This is used for boh asr and enh

pyproject.toml
```diff
@@ -0,0 +1,139 @@
...
+  "opt_einsum",
+  "protobuf",
```
sw005320: Is it needed?

pyproject.toml
```diff
@@ -0,0 +1,139 @@
...
+  "dask[distributed]",
+  "dask_jobqueue",
+]
+
+task-asr = [
+  "ctc-segmentation>=1.6.6",
+  "editdistance",
+  "h5py",
+  "lhotse",
+  "jamo",
```
sw005320: what's this?

pyproject.toml
```diff
@@ -0,0 +1,139 @@
...
+  "omegaconf",
+  "datasets",
+  "dask[distributed]",
+  "dask_jobqueue",
+]
+
+task-asr = [
+  "ctc-segmentation>=1.6.6",
+  "editdistance",
+  "h5py",
```
sw005320: this is used for many other places. https://github.com/search?q=repo%3Aespnet%2Fespnet+h5py&type=code
should it be in core?

tools/Makefile
```diff
@@ -128,7 +128,9 @@ lightning.done: pytorch.done
 # NOTE(kamo): conda_packages is not necessary for installation of espnet, but add it the dependencies just in case.
 espnet.done: pytorch.done conda_packages.done lightning.done
 	@echo NUMPY_VERSION=$(shell . ./activate_python.sh && python3 -c "import numpy; print(numpy.__version__)")
```
sw005320, line 130: This (copy, move, etc.) looks hacky
Why do we need this procedure?

tools/setup.py
```diff
@@ -141,7 +141,7 @@
     long_description=open(os.path.join(dirname, "README.md"), encoding="utf-8").read(),
     long_description_content_type="text/markdown",
     license="Apache Software License",
-    packages=find_packages(include=["espnet*"]),
+    packages=find_packages(include=["espnet", "espnet2"]),
```
sw005320: will you add espnet3 later?

pyproject.toml
sw005320, line 1: Can you make sure that each task's requirement is correct?

sw005320:
Please fix all CI issues.

.github/workflows/ci_on_ubuntu.yml
Masao-Someki: I thought it was for readability, but maybe we can keep it in the current directory.
@Fhrozen Is there any specific reason we moved the setup.py into `tools/` ?

pyproject.toml
Masao-Someki: Thank you, I'll revisit and fix them.

.github/workflows/ci_on_ubuntu.yml
Fhrozen: I did this to keep the dependencies of ESPnet 1/2 and ESPnet 3 separate. Because espnet1/2 are installed by the make file, I make the  `setup.py` to move between folders. I looked for keeping both files `pyproject.toml` and `setup.py` in the same directory and redirecting the installation when the `espnet3` is added, but it was a little messy. Let me know if this option is not valid, and will try to look for a different solution.

tools/Makefile
Fhrozen: espnet3 is implemented to run with only python. THe installation is executed using the `pyproject.toml` file. When using the `setup.py`, will only execute as espnet1/2 installation, and use the dependencies for espnet1/2. I keep the installation on espnet1/2 with the setup.py file, and espnet3 with pyproject.toml. Let me know which will be a better option for the installation.

sw005320: I see.
I feel that this is confusing.
Is it possible to combine them?
(If that makes `pyproject.toml` complicated, we may keep on your way)

Masao-Someki:
Thank you.
> what does "core" mean?
> Does it only perform inference?
> This definition and description are not clear.

For my understanding, the `core` is the packages that is required across many tasks.
For inference, we need to install task-specific dependency. (E,g., espnet_tts_frontend for some TTS models)

> I still cannot understand why this PR affects the Python script under the example directories, but I checked all of them and they are OK

They are formattings. After merging #6140 it, they should disappear.

pyproject.toml
Masao-Someki: I found that if we want to move torch_complex to task-specific installation, we need to mofdify 63 additional files. This is mainly used in asr/diar/enh and many test scripts, I'm thinking of moving it to `core` package..

sw005320:
> Thank you.
> 
> > what does "core" mean?
> > Does it only perform inference?
> > This definition and description are not clear.
> 
> For my understanding, the `core` is the packages that is required across many tasks. For inference, we need to install task-specific dependency. (E,g., espnet_tts_frontend for some TTS models)

If so, please clearly define it and document it.

> > I still cannot understand why this PR affects the Python script under the example directories, but I checked all of them and they are OK
> 
> They are formattings. After merging #6140 it, they should disappear.

What would affect this formatting?

Masao-Someki:
Oh, I just realized that for this PR the formatting is applied more than #6140...? I will check why

Masao-Someki:
I found that if I run black in local PC it generates different output from precomit's output. It seems that the current master is formatted on local PC. I will fix that.

Masao-Someki:
Hi, here's a description and discussion items on how I changed each packages:

* **pyworld**: It's used in almost all the `evaluate_f0` and `evaluate_uvu` scripts under each task. Judging by the script names, it seems like it's mainly used in TTS/VC. Should we keep it as a core dependency, or is it okay to move to task-tts?

* **lightning**: It's mainly used in ESPnet-3, and although it's supported in ESPnet-2 as a LightningTrainer, it feels a bit off to include it as a core dependency of ESPnet-2.. I'd like to have comments on this package.

* **nltk**: It's only used for evaluation in MT, ST, and dialogue. I'm not sure which dependency group is reasonable.

* **h5py**: Other than being used in `ESPnetDataset`, it's only used in tests and in espnet1. I moved it to the test dependency group, and making the dataset import it only when the h5py dataset type is actually used.

* **lhotse**: This is mainly used in diar-ASR/ASR and a few data preparation scripts under `egs2/`. It's not used at all in the main ESPnet code, so I removed it from the dependencies.

* **jamo**: This is used for Korean G2P. Moved it under TTS.

* **protobuf**: I couldn't find any actual usage under `egs/`, `egs2/`, `espnet/`, or `espnet2/`, so I removed it.

* **espnet_model_zoo**: Should we include this package in somewhere?

* **task-asr-state-space**: There are several packages that is only required from state-space model definition. so I moved them to `task-asr-state-space`.

* **tensorboard**, **torch_complex**, **einsum**, : I moved it to core, as it is used in several tasks.



Let me know your thoughts on each of these.

Masao-Someki:
I've created a new venv, run `pip install -e .[core]`, and ASR inference with the following model runs without issue.

 https://huggingface.co/pyf98/librispeech_100_ctc_e_branchformer

sw005320:
> Hi, here's a description and discussion items on how I changed each packages:
> 
> * **pyworld**: It's used in almost all the `evaluate_f0` and `evaluate_uvu` scripts under each task. Judging by the script names, it seems like it's mainly used in TTS/VC. Should we keep it as a core dependency, or is it okay to move to task-tts?

task-tts

> * **lightning**: It's mainly used in ESPnet-3, and although it's supported in ESPnet-2 as a LightningTrainer, it feels a bit off to include it as a core dependency of ESPnet-2.. I'd like to have comments on this package.
> * **nltk**: It's only used for evaluation in MT, ST, and dialogue. I'm not sure which dependency group is reasonable.

task-mt, task-st, task-sds

> * **h5py**: Other than being used in `ESPnetDataset`, it's only used in tests and in espnet1. I moved it to the test dependency group, and making the dataset import it only when the h5py dataset type is actually used.
> * **lhotse**: This is mainly used in diar-ASR/ASR and a few data preparation scripts under `egs2/`. It's not used at all in the main ESPnet code, so I removed it from the dependencies.

Are you sure?
I thought that we use this for our future development. @popcornell 

> * **jamo**: This is used for Korean G2P. Moved it under TTS.
> * **protobuf**: I couldn't find any actual usage under `egs/`, `egs2/`, `espnet/`, or `espnet2/`, so I removed it.
> * **espnet_model_zoo**: Should we include this package in somewhere?

We can keep on it on core.

> * **task-asr-state-space**: There are several packages that is only required from state-space model definition. so I moved them to `task-asr-state-space`.

I don't think this is the right word
Too many hyphens.
Why not keep it in ASR?

> * **tensorboard**, **torch_complex**, **einsum**, : I moved it to core, as it is used in several tasks.

Can you test it by using the unit/integration test?
Ideally, we want to test it by using their installation and corresponding integration test (e.g., task-asr and asr integration test).
This will ensure that the task-dependent installation can work correctly. 
We should not deploy it without the test.

Masao-Someki:
I'm not sure why, but I'm getting different formatting results from Black compared to what @Fhrozen gets on his local machine, even if we are using the same version. This inconsistency is the main reason we're seeing so many formatting changes in the files.

I asked him to run the formatter on his end to resolve the differences, which should make the changes disappear, but I don't think this is a sustainable situation...

Masao-Someki:
Not sure why the setuptools version has changed, but we need to use specific version to avoid import error of chainer. 
I found that in the CI the setuptools version is 79.0.1, but we need setuptools<74.0.0

Reference:
image

doc/installation.md
```diff
@@ -173,8 +173,81 @@ We also have [prebuilt Kaldi binaries](https://github.com/espnet/espnet/blob/mas
...
+    $ pip install -e ".[core]"
+    ```
+
+To install task-specific dependencies, you can add the desired options. For example:
+
+* For ASR:
+
+    ```sh
+    $ cd <espnet-root>
+    $ pip install -e ".[core,task-asr]"
```
sw005320: 1. Is it possible to automatically include core when we specify task-*?
2. what is the default behavior without specifying anything? `all`?
3. I still cannot understand the role of `core`. If it is used as a standalone? If not, I cannot understand why we make it as a package.

sw005320: if 1 is possible, my question about 3 will be gone.

sw005320: If we move libraries from `core` to `dependencies,` would 1 be solved?

Masao-Someki: The default behavior without specifying will install the following dependencies, but I think it's good to include `core` packages here, as it is required for other groups. I'll make commit on that

```
dependencies = [
    "setuptools>=38.5.1,<74.0.0",
    "configargparse>=1.2.1",
    "packaging",
    "typeguard",
    "humanfriendly",
    "scipy>=1.4.1",
    "numpy",
    "PyYAML>=5.1.2",
    "soundfile>=0.10.2",
]
```

sw005320: Thanks!
With this, we don't have to include a (vague) concept of `core.`
We can also avoid specifying both `core` and `task-` in the installation.

sw005320:
Please also check the CI errors.

ci/test_integration_espnet2.sh
```diff
@@ -148,7 +157,14 @@ done
 rm -rf exp dump data
 cd "${cwd}"
 
+# Uninstall task-dependency
+python3 test_utils/uninstall_extra.py
```
sw005320, line 161: Can this uninstall `[task-asr]` libraries?
Why not use `pip uninstall`?

doc/installation.md
```diff
@@ -173,8 +173,80 @@ We also have [prebuilt Kaldi binaries](https://github.com/espnet/espnet/blob/mas
...
+
+    ESPnet adopts the modern Python packaging standard using `pyproject.toml`.
+    You can install the **core dependencies** with:
+
+    * Core installation:
+
+        ```sh
+        $ cd <espnet-root>
+        $ pip install -e ".[core]"
+        ```
```
sw005320: We should remove the descriptions about `core`

doc/installation.md
```diff
@@ -173,8 +173,80 @@ We also have [prebuilt Kaldi binaries](https://github.com/espnet/espnet/blob/mas
...
+        cd tools
+        bash -c ". ./activate_python.sh; python3 check_install.py"
+        ```
+
+    3. [Optional] How dependency groups are organized
+        The ESPnet `pyproject.toml` defines optional dependency groups:
+
+        | Group      | Purpose                       |
+        | ---------- | ----------------------------- |
+        | `core`     | Core libraries and runtime    |
```
sw005320: We should remove

ci/test_integration_espnet2.sh
Masao-Someki: I tried `pip uninstall espnet[task-asr]` but it will uninstall only the espnet, and not the packages in [task-asr]. I used several commands/tools to uninstall only [task-asr] related packages, and this is the best way so far. Other tools couldn't uninstall dependencies or agrgessively uninstall everything including pip.

image
image
image

sw005320: I see.
Thanks for the explanation and investigation.

Masao-Someki:
@Fhrozen Once the CI issues are resolved, could you run the formatter on this branch? There have been several updates to files in `espnet2/` to avoid import errors, but the formatting differences make it hard to identify the actual changes.

espnet2/asr/state_spaces/utils.py
```diff
@@ -4,8 +4,15 @@
...
-from omegaconf import DictConfig, ListConfig
+try:
+    import hydra
+    from omegaconf import DictConfig, ListConfig
+except ImportError:
+    import logging
+
+    logging.warning(
+        "If you are running state-space model," "run `pip install espnet['task-asr']`."
+    )
```
sw005320: Is it correct?
`hydra` and `omegaconf` are related to `espnet-3`.
Why do they appear here?
What we need to install `task-asr`

pyproject.toml
```diff
@@ -0,0 +1,162 @@
...
+
+[tool.setuptools.packages.find]
+include = ["espnet*"]
+
+[tool.setuptools.package-data]
+espnet = ["version.txt"]
+
+[project.optional-dependencies]
+
+espnet-3 = [
```
sw005320, line 70: I cannot understand the position of this.
How is it used?

Masao-Someki: It will be used in espnet-3, but maybe we should add this dependency in future PR.

.github/workflows/ci_on_ubuntu.yml
sw005320: Would it be solved in the future?
If so, we can keep it, but otherwise, we should only maintain `pyproject.toml` since it is very confusing.

Since most of our espnet2 integration tests are working, I think we can unify `setup.py` with `pyproject.toml`

espnet2/asr/state_spaces/utils.py
Masao-Someki: It looks like the files under `espnet2/asr/state_spaces/` use Hydra and OmegaConf.
Also, `opt_einsum` is only used within this directory, so I’ve included these in the `task-asr` group...

sw005320:
@Masao-Someki and @Fhrozen, this might help



To reflect flake8 in the `.pre-commit-config.yaml` file in the `@espnet/espnet` repository, you should add a flake8 hook configuration under the `repos:` section. This ensures that flake8 will run as part of your pre-commit checks.

Here’s how you can do it:

1. **Locate `.pre-commit-config.yaml`**  
   Open the `.pre-commit-config.yaml` file at the root of your repository (or in the `/files` directory if that’s where it is).

2. **Add or Update the flake8 Hook**  
   Add the following block under the `repos:` section if it doesn’t exist, or update it if it does:

   ```yaml
   - repo: https://github.com/pycqa/flake8
     rev: 6.1.0  # or the latest stable version
     hooks:
       - id: flake8
         additional_dependencies: []  # Add any flake8 plugins you use here
   ```

3. **Example of a Complete Section:**

   ```yaml
   repos:
     - repo: https://github.com/pycqa/flake8
       rev: 6.1.0
       hooks:
         - id: flake8
           additional_dependencies: [flake8-bugbear]
           # You can add more flake8 plugins to the array if needed
   ```

4. **Install/Update Pre-commit Hooks**  
   After editing the file, run:
   ```
   pre-commit install
   pre-commit autoupdate
   ```

**Summary:**  
- Add the flake8 repo block to your `.pre-commit-config.yaml`
- Specify the `rev` (version)
- Use the `additional_dependencies` list for any plugins

Would you like to see the current contents of your `.pre-commit-config.yaml` file to get a context-aware example?

ci/test_integration_espnetez.sh
```diff
@@ -17,6 +17,14 @@ gen_dummy_coverage(){
 #### Make sure chainer-independent ####
 python3 -m pip uninstall -y chainer
 
+# First uninstall all espnet-related dependencies including all extras.
+# I use toml and load the pyproject.toml
+python3 -m pip install toml
+python3 test_utils/uninstall_extra.py
```
Fhrozen, line 23: this line is uninstalling nltk, because it is an optional package. Check if the package should become a permanent package or remove the test.

Fhrozen: Or maybe disable the test related to the package.

Fhrozen: nvm, I just found the solution

Masao-Someki: @Fhrozen Since NLTK is now used in many tasks, I think we can simply move it to the main dependency?

Masao-Someki: Also it will fix the CI issue?

pyproject.toml
```diff
@@ -0,0 +1,157 @@
...
+  "sphinx-argparse",
+  "sphinx-markdown-builder"
+]
+
+all = [
+  "espnet[task-asr]",
+  "espnet[task-tts]",
+  "espnet[task-enh]",
+  "espnet[task-spk]",
+]
```
sw005320: `all` does not seem to include all.
This will also cause some issues in the future when someone extends the task.
They may forget to add it to `all` (like you do).
Is there a way to include all tasks by using `task*` or something like that?

tools/Makefile
sw005320: I believe this is not resolved.
Why do we still maintain `setup.py` and `pyproject.toml`?

In my understanding, the current `all` in `pyproject.toml` is almost the same as `setup.py`
(Thus, the integration tests are all passed.)
Can you check them and resolve the confusing situations?

pyproject.toml
Fhrozen: From Gemini:

That is an excellent and insightful question that explores the boundaries of what pyproject.toml is designed to do.

The short answer is **no**, it is not possible to use wildcards (`task-*`) or loops directly within the `pyproject.toml` file to dynamically group dependencies.

### The Reason: pyproject.toml is Declarative, Not Programmatic
The `pyproject.toml file is designed to be a declarative and static configuration file. This means you declare what your project's configuration is, but you cannot include logic—like loops, conditionals, or wildcards—to compute that configuration. Its strength lies in its simplicity and predictability; any tool can read the file and know the project's metadata without executing any code.

### The Standard Way: Explicitly Listing the Extras
The correct and standard way to achieve your goal is to explicitly list each extra you want to include in your my-packages group. While it may feel repetitive, it is unambiguous and the officially supported method.

```toml
# pyproject.toml

[build-system]
requires = ["setuptools>=61.0"]
build-backend = "setuptools.build_meta"

[project]
name = "my-component-package"
version = "1.0.0"

# 1. Define your individual component extras
[project.optional-dependencies]
task-1 = [
    "pandas",
]
task-2 = [
    "matplotlib",
    "seaborn"
]
task-3 = [
    "scikit-learn"
]
# ...and so on for task-4, task-5, etc.

# 2. Create the group that explicitly includes the others
all-tasks = [
    "my-component-package[task-1]",
    "my-component-package[task-2]",
    "my-component-package[task-3]",
]

# You might also want a group for development tools
dev = [
    "my-component-package[all-tasks]", # Include all the above
    "pytest",
    "ruff"
]
```

### The Advanced Way: Adding Dynamic Logic with `setup.py`
If you absolutely must have dynamic logic (for example, in a very complex project with dozens of auto-generated extras), you can fall back to using a `setup.py` script.

This involves telling `pyproject.toml` that the `optional-dependencies` are "dynamic," meaning they will be determined at build time by the build backend (`setuptools`) running the `setup.py` script.

**This adds complexity and is generally not recommended unless you have a strong reason.**

### Conclusion

| Method|	How It Works|	Pros	|Cons|
|:--:|:--:|:--:|:--:|
|Standard (`pyproject.toml` only)	|Explicitly list each extra in the group.	|Simple, clear, standard, easy for tools to analyze.	|Requires manual updating if you add a new `task-N`.|
|Advanced (Dynamic `setup.py`)	|Use Python code to generate the list of extras.	|Automated, powerful for complex scenarios.	|Adds complexity, dependencies are not statically visible, can be harder to debug.|

Masao-Someki: @Fhrozen Could you try copilot review? If it can detect this error, we can continue with that.

Masao-Someki:
Okay, it seems we need additional test that checks `all` includes all dependencies. I'll prepare that in `test_import_all.py`

tools/Makefile
```diff
@@ -128,7 +128,9 @@ lightning.done: pytorch.done
...
+	cp ./setup.py ../ && mv ../pyproject.toml ../pyproject.toml.tmp
```
sw005320: This must be changed.

Masao-Someki:
I added additional dependencies to setup.py. These are for optional packages and are only installed when 'all' is specified in the extras.

pyproject.toml
```diff
@@ -21,22 +21,37 @@ classifiers = [
 requires-python = ">=3.10"
 license = { "text" = "Apache Software License" }
 dependencies = [
```
sw005320, line 26: how about using `==` for stable installation?

Masao-Someki: Thank you, this is a good idea, especially for packages like formatting tools or other libraries that don’t get frequent updates. 
However, after I checked all packages, I found that there are some libraries, like `datasets`, that are very actively developed (new version in almost one month). For these, using the specific version might prevent us from benefiting from important bug fixes or new features in a timely manner.

So how about using `==` for more stable or rarely updated packages, and using a more flexible version specifier (like `>=`) for actively maintained ones?

https://pypi.org/project/datasets/#history

Masao-Someki:
@sw005320 @Fhrozen 
If this PR looks good, could you please merge it?
To my understanding, it’s ready, and any remaining issues can be handled in a follow-up PR to espnet3.
CI for ESPnet3 is currently blocked by this PR, as it's the foundation for all related workflows.

pyproject.toml
```diff
@@ -20,38 +20,40 @@ classifiers = [
 ]
 requires-python = ">=3.10"
 license = { "text" = "Apache Software License" }
+
+# We can use newer setuptools after removing chainer support.
```
sw005320, line 24: It's better to make a comment for some packages that strictly require a specific version, e.g., `"sentencepiece==0.2.0",` (currently == or <=).

Also, what is the process of upgrading them?
When we fix the version, we can test upgrading to it. If it is successful, we will replace the versions with the latest ones.

Masao-Someki: I think we can use Dependabot or GitHub Actions to automatically comment or create PR.

- Dependabot: Automatically checks `pyproject.toml` and creates PRs for outdated packages.

- GitHub Actions + Scripts: Post the list of outdated packages as a comment on PRs.

sw005320:
Can you merge the update in the main repo to espnet3 and update this PR?

Masao-Someki:
Please merge #6192 first to merge master branch into espnet-3 and this PR

Masao-Someki:
@sw005320 
I've resolved the conflicts and all CI checks have passed. Could you please review and merge this PR if everything looks good to you?

sw005320:
@Masao-Someki, can you fix copilot's review?

Masao-Someki:
Hi @sw005320 or @Fhrozen, could you rerun the CI, please? It seems the CI fails when it tries to download a package via pip, and it should be passed.

sw005320:
OK, I merged it.
What is the next?

PR #6155, opened 2025-06-19
LID-1: Training and task setup

Qingzheng-Wang:
## What did you change?

This PR introduces the core training and evaluation pipeline for the LID task in ESPnet, including:
- `espnet2/bin/lid_train.py`: training entry point
- `espnet2/tasks/lid.py`, `espnet2/tasks/abs_task.py`: task definition and integration
- `espnet2/train/preprocessor.py`: updated to support LID-specific preprocessing

---

## Why did you make this change?

This is the foundational step to enable language identification (LID) training within ESPnet. It defines the task and enables downstream LID model training.

---

## Is your PR small enough?

Yes

---

## Additional Context

- Part 1 of the LID system.
- Follow-up PRs will introduce model definition, inference, samplers, and recipes

PR #6156, opened 2025-06-19
LID-2: Model, loss and pooling modules

Qingzheng-Wang:
## What did you change?

This PR introduces the core model components and loss functions for language identification (LID), located under the `espnet2/lid` directory.

Specifically, it includes:

- **Loss functions** (`espnet2/lid/loss`):
  - Adapted from ESPnet-SPK losses.
  - Modified to support LID-specific training by adding language classification accuracy computation and returning predicted language IDs.

- **Pooling functions** (`espnet2/lid/pooling`):
  - Adapted from ESPnet-SPK pooling layers.
  - Modified to ensure that mean, statistical, and attention-based pooling operate only on valid (unpadded) feature lengths.

- **`espnet2/lid/espnet_model`**

---

These modules are adapted from ESPnet-SPK components and revised to support language identification, including changes in output structure and padding handling.

---

## Is your PR small enough?

Yes

---

## Additional Context

- Part 2 of the LID system.
- Depends on: #6155

### DIFF
--- a/espnet2/lid/espnet_model.py
+++ b/espnet2/lid/espnet_model.py
@@ -1,0 +1,158 @@
+from typing import Dict, Optional, Tuple, Union
+
+import torch
+from typeguard import typechecked
+
+from espnet2.asr.encoder.abs_encoder import AbsEncoder
+from espnet2.asr.frontend.abs_frontend import AbsFrontend
+from espnet2.asr.specaug.abs_specaug import AbsSpecAug
+from espnet2.layers.abs_normalize import AbsNormalize
+from espnet2.lid.loss.abs_loss import AbsLoss
+from espnet2.lid.pooling.abs_pooling import AbsPooling
+from espnet2.spk.projector.abs_projector import AbsProjector
+from espnet2.torch_utils.device_funcs import force_gatherable
+from espnet2.train.abs_espnet_model import AbsESPnetModel
+
+
+class ESPnetLIDModel(AbsESPnetModel):
+    r"""ESPnet LID model
+    Support for language identification and language embedding extraction.
+    """
+
+    @typechecked
+    def __init__(
+        self,
+        frontend: Optional[AbsFrontend],
+        specaug: Optional[AbsSpecAug],
+        normalize: Optional[AbsNormalize],
+        encoder: Optional[AbsEncoder],
+        pooling: Optional[AbsPooling],
+        projector: Optional[AbsProjector],
+        loss: Optional[AbsLoss],
+        extract_feats_in_collect_stats: Optional[bool] = None,
+    ):
+
+        super().__init__()
+
+        self.frontend = frontend
+        self.specaug = specaug
+        self.normalize = normalize
+        self.encoder = encoder
+        self.pooling = pooling
+        self.projector = projector
+        self.loss = loss
+
+    @typechecked
+    def forward(
+        self,
+        speech: torch.Tensor,
+        speech_lengths: torch.Tensor,
+        lid_labels: Optional[torch.Tensor] = None,
+        extract_embd: bool = False,
+        **kwargs,
+    ) -> Union[
+        Tuple[torch.Tensor, torch.Tensor],
+        Tuple[torch.Tensor, Dict[str, torch.Tensor], torch.Tensor],
+        torch.Tensor,
+    ]:
+        r"""Forward pass of the LID model.
+
+        Processes raw speech through frontend, encoder, pooling, and loss modules.
+
+        Args:
+            speech: Input waveform tensor (batch_size, num_samples)
+            speech_lengths: Lengths of each input in the batch (batch_size,)
+            lid_labels: Ground truth language labels (batch_size,)
+            extract_embd: If True, return language embeddings and predictions (inference mode)
+
+        Returns:
+            - If extract_embd=True (inference mode):
+                Tuple(lang_embd, pred_lids)
+            - If training:
+                Tuple(loss, stats_dict, batch_weight)
+        """
+
+        if lid_labels is not None:
+            assert speech.shape[0] == lid_labels.shape[0], (
+                speech.shape,
+                lid_labels.shape,
+            )
+
+        batch_size = speech.shape[0]
+        stats = dict()
+
+        # 1. Extract feats
+        # Must transfer speech_lengths to extract_feats to get correct feat_lengths
+        feats, feat_lengths = self.extract_feats(speech, speech_lengths)
+        frame_level_feats = self.encode_frame(feats)
+
+        # 2. Aggregation into utterance-level
+        utt_level_feat = self.pooling(frame_level_feats, feat_lengths=feat_lengths)
+
+        # 3. (Optionally) Go through further projection(s)
+        lang_embd = self.project_lang_embd(utt_level_feat)
+
+        # 4. Calculate loss
+        # NOTE: If lid_labels is None, loss and accuracy are None
+
+        loss, accuracy, pred_lids = self.loss(lang_embd, lid_labels)
+
+        if extract_embd:
+            return lang_embd, pred_lids
+
+        stats["loss"] = loss.detach()
+        if accuracy is not None: # if not provide labels, accuracy is None
+            stats["accuracy"] = accuracy.detach()
+
+        loss, stats, weight = force_gatherable((loss, stats, batch_size), loss.device)
+        return loss, stats, weight
+
+    def extract_feats(
+        self, speech: torch.Tensor, speech_lengths: torch.Tensor
+    ) -> Tuple[torch.Tensor, torch.Tensor]:
+        batch_size = speech.shape[0]
+        speech_lengths = (
+            speech_lengths
+            if speech_lengths is not None
+            else torch.ones(batch_size).int() * speech.shape[1]
+        )
+
+        # 1. Extract feats
+        if self.frontend is not None:
+            feats, feat_lengths = self.frontend(speech, speech_lengths)
+        else:
+            feats = speech
+            feat_lengths = None
+
+        # 2. Apply augmentations
+        if self.specaug is not None and self.training:
+            feats, _ = self.specaug(feats, feat_lengths)
+
+        # 3. Normalize
+        if self.normalize is not None:
+            feats, _ = self.normalize(feats, feat_lengths)
+
+        return feats, feat_lengths
+
+    def encode_frame(self, feats: torch.Tensor) -> torch.Tensor:
+        frame_level_feats = self.encoder(feats)
+
+        return frame_level_feats
+
+    def project_lang_embd(self, utt_level_feat: torch.Tensor) -> torch.Tensor:
+        if self.projector is not None:
+            lang_embd = self.projector(utt_level_feat)
+        else:
+            lang_embd = utt_level_feat
+
+        return lang_embd
+
+    def collect_feats(
+        self,
+        speech: torch.Tensor,
+        speech_lengths: torch.Tensor,
+        lid_labels: torch.Tensor = None,
+        **kwargs,
+    ) -> Dict[str, torch.Tensor]:
+        feats, feats_lengths = self.extract_feats(speech, speech_lengths)
+        return {"feats": feats}
--- a/espnet2/lid/loss/aamsoftmax.py
+++ b/espnet2/lid/loss/aamsoftmax.py
@@ -1,0 +1,115 @@
+# Code adapted from https://github.com/espnet/espnet/blob/master/espnet2/spk/loss/aamsoftmax.py
+# Original code refer to https://github.com/clovaai/voxceleb_trainer/blob/master/loss/aamsoftmax.py
+# and https://github.com/wujiyang/Face_Pytorch (Apache License)
+
+import math
+
+import torch
+import torch.nn as nn
+import torch.nn.functional as F
+
+from espnet2.lid.loss.abs_loss import AbsLoss
+from typing import Optional, Tuple
+
+
+class AAMSoftmax(AbsLoss):
+    r"""Additive angular margin softmax.
+
+    Reference:
+    ArcFace: Additive Angular Margin Loss for Deep Face Recognition
+    https://arxiv.org/pdf/1801.07698
+
+    Args:
+        nout: Dimension of input features (embedding size)
+        nclasses: Number of output classes
+        margin: Angular margin for positive samples
+        scale: Feature scaling factor
+        easy_margin: Whether to use easy margin variant
+    """
+
+    def __init__(
+        self, 
+        nout: int, 
+        nclasses: int, 
+        margin: float=0.3, 
+        scale: int=15, 
+        easy_margin: bool=False, 
+    ):
+        super().__init__(nout)
+
+        self.m = margin
+        self.s = scale
+        self.in_feats = nout
+
+        self.weight = torch.nn.Parameter(
+            torch.FloatTensor(nclasses, nout), requires_grad=True
+        )
+        nn.init.xavier_normal_(self.weight, gain=1)
+        
+        self.easy_margin = easy_margin
+        self.cos_m = math.cos(self.m)
+        self.sin_m = math.sin(self.m)
+        self.th = math.cos(math.pi - margin)           # Threshold for easy_margin
+        self.mm = math.sin(math.pi - margin) * margin  # For numerical stability
+
+        self.ce = nn.CrossEntropyLoss()
+
+    def forward(
+        self, 
+        input: torch.Tensor, 
+        label: Optional[torch.Tensor]=None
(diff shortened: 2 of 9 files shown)


sw005320:
This pull request introduces a new language identification (LID) model to the ESPnet framework, along with supporting modules for feature extraction, pooling, and loss computation. The changes primarily focus on adding the `ESPnetLIDModel` class, implementing various loss functions, and providing pooling mechanisms for aggregating frame-level features into utterance-level embeddings.

### Language Identification Model:
* [`espnet2/lid/espnet_model.py`](diffhunk://#diff-0b495cb7ee0dbdaac7585af77d32a48b962a0052aee2489314f290b3c7672b01R1-R158): Added the `ESPnetLIDModel` class, which supports language identification and embedding extraction. It integrates frontend, encoder, pooling, and loss modules for processing speech data.

### Loss Functions:
* [`espnet2/lid/loss/aamsoftmax.py`](diffhunk://#diff-16964c8a9ed1077ad359e43bcbc018daf547f6f1a372a57936f475004bdb9a45R1-R112): Implemented the AAMSoftmax loss, which uses additive angular margin for better feature discrimination.
* [`espnet2/lid/loss/aamsoftmax_subcenter_intertopk.py`](diffhunk://#diff-f70fa8c50f079bc42c8f184f8e4ac8920e7cab9c58a7fb92897480bd5844f8d6R1-R191): Added a loss function combining ArcFace, sub-center enhancement, and inter-topK penalty for improved intra-class variation handling and hard negative sample penalization.
* [`espnet2/lid/loss/softmax.py`](diffhunk://#diff-177284531320a2a6b9b84cf605f3dae2d87b15e863cebed466f2a555b3107992R1-R60): Introduced a standard Softmax loss for language identification tasks.
* [`espnet2/lid/loss/abs_loss.py`](diffhunk://#diff-090487d95db2be05374956f1b4f27b523433056f5a3b326784b7902dee86b8a5R1-R16): Created an abstract base class `AbsLoss` to standardize loss function interfaces.

### Pooling Mechanisms:
* [`espnet2/lid/pooling/chn_attn_stat_pooling.py`](diffhunk://#diff-56f76c54d8563c1bcc682a3ccebc86795c6ea7ec3e17abc8929bbf85793a3593R1-R97): Added channel-attentive statistical pooling, which aggregates frame-level features into utterance-level embeddings using attention mechanisms.
* [`espnet2/lid/pooling/mean_pooling.py`](diffhunk://#diff-0f2f120a6fbafc838009ca9e9f591c3fd0f4503a670ac53ce236e84e15c009ffR1-R44): Implemented mean pooling for averaging frame-level features into a single utterance-level feature.
* [`espnet2/lid/pooling/abs_pooling.py`](diffhunk://#diff-9ee105fda5620769748db825d72e12e30361abc752e18148c4e935a585fc85b0R1-R14): Created an abstract base class `AbsPooling` to define pooling interfaces.

sw005320:
Please fix the conflict.

espnet2/lid/espnet_model.py
sw005320, line 1: It's more like a note or comment.
How do we handle the code-switching case?
Can we extend utt2lang to have multiple languages?

espnet2/lid/loss/aamsoftmax.py
```diff
@@ -0,0 +1,115 @@
...
+from typing import Optional, Tuple
+
+import torch
+import torch.nn as nn
+import torch.nn.functional as F
+
+from espnet2.lid.loss.abs_loss import AbsLoss
+
+
+class AAMSoftmax(AbsLoss):
```
sw005320: This sounds duplicated with `espnet2/spk/loss/aamsoftmax.py`?
Why do we need to have a new function?
Please explain it and add a comment if needed.

espnet2/lid/loss/aamsoftmax_subcenter_intertopk.py
sw005320, line 1: ditto
why do we need to duplicate it?
please explain/clarify how it differs to avoid some confusions.

espnet2/lid/espnet_model.py
Qingzheng-Wang: Currently it only supports one utterance corresponding to one language label.

espnet2/lid/loss/softmax.py
sw005320, line 1: ditto

espnet2/lid/pooling/abs_pooling.py
sw005320, line 1: ditto
Please explain why we need additional pooling functions

espnet2/lid/pooling/chn_attn_stat_pooling.py
sw005320, line 1: ditto
Please explain why we need additional pooling functions

espnet2/lid/loss/aamsoftmax_subcenter_intertopk.py
Qingzheng-Wang: The main differences are:

1. This implementation returns predicted labels (via argmax) during inference, enabling LID predictions.
2. It also computes and returns accuracy during training and evaluation, making it easier to monitor model performance.

sw005320:
@Jungjee, can you review this PR?
I'm actually fine with having some duplicated functions, but I want to make sure.

espnet2/lid/pooling/chn_attn_stat_pooling.py
Qingzheng-Wang: Compared to the SPK version (`espnet2/spk/pooling/*`), these implementations explicitly handle feature lengths, ensuring that mean and standard deviation pooling are computed only over valid (unpadded) frames.

Qingzheng-Wang: I have added comments explaining this.

espnet2/lid/pooling/stat_pooling.py
```diff
@@ -0,0 +1,64 @@
...
+
+
+class StatsPooling(AbsPooling):
+    """Aggregates frame-level features to single utterance-level feature.
+
+    Note:
+        Compared to the SPK version (`espnet2/spk/pooling/stat_pooling.py`),
+        this implementation explicitly handles feature lengths,
+        ensuring that mean and standard deviation pooling are
+        computed only over valid (unpadded) frames.
```
sw005320: Do you mean that the origional SPK's pooling does not care about the padded region?

@Jungjee, if so, I think it's better to use this for SPK.
Can you check it?

Qingzheng-Wang: Yes, the original SPK pooling aggregated over all frames, including padding frames.

sw005320:
@Jungjee, can you confirm it?
If the padding issue is true, this would be an improvement point of your work.

sw005320:
@holvan, can you check if these fixed pooling functions are working in your SPK experiments?
If it has some improvements (or it does not degrade the performance), we should replace pooling functions in both SPK and LID with these modified functions.

sw005320:
> @Jungjee, can you confirm it? If the padding issue is true, this would be an improvement point of your work.

@Jungjee, this is a reminder.
Can you take a look at this?
This is important for your SID implementation, and we need to review it carefully and make the necessary corrections.
I don't want to have twisted situations (e.g.,  LID has the correct implementation while SID has some bugs in pooling).

espnet2/lid/loss/aamsoftmax.py
Jungjee: +1.
I don't see critical difference compared to `espnet2/spk/loss/aamsoftmax.py`. 
Is there a reason why you're introducing this file instead of importing `espnet2/spk/loss/aamsoftmax.py`?

espnet2/lid/loss/aamsoftmax_subcenter_intertopk.py
```diff
@@ -0,0 +1,205 @@
+# Code adapted from:
```
Jungjee: ditto. Please let us know what is different to the original function in spk side and whether you can enhance that code instead of creating a nearly-duplicate.

espnet2/lid/loss/abs_loss.py
```diff
@@ -0,0 +1,16 @@
+from abc import abstractmethod
```
Jungjee: ditto. I am not currently sold on adding `espnet2/lid/loss` in general.

espnet2/lid/pooling/abs_pooling.py
Jungjee: +1

sw005320:
Thanks @Jungjee!
@Qingzheng-Wang, can you reflect the comments?

espnet2/lid/loss/aamsoftmax.py
Qingzheng-Wang: The main differences are:

1. This implementation returns predicted labels (via argmax) during inference, enabling output with LID predictions during inference.
2. It also computes and returns accuracy during training and evaluation, making it easier to monitor model performance.

Although the core principle of AAMSoftmax remains the same as in spk, this version uses a different output interface specific for the LID task. This modification is necessary for my design of the LID inference pipeline (#6157 ), which outputs the LID predictions.

espnet2/lid/loss/aamsoftmax_subcenter_intertopk.py
Qingzheng-Wang: My explanation is:

I want to return the LID predictions during inference (the LID inference pipeline is #6157 ). The predictions are obtained by applying argmax to the cosine similarity scores:

```
cosine = F.linear(
    F.normalize(input), F.normalize(self.weight)
)  # Output: (batch_size, nclasses)

pred_lids = torch.argmax(cosine, dim=1)
```

So, I rewrite the forward functions of original losses.

espnet2/lid/loss/aamsoftmax.py
sw005320: OK, please embed such a comment and avoid future confusions by the users

Jungjee: @Qingzheng-Wang @sw005320 , to me if it's merely additional returns, I still believe we could avoid adding dupliated functions but instead add these chages to ESPnet-SPK module side and update ESPnet-SPK to accomodate this change.

espnet2/lid/loss/aamsoftmax_subcenter_intertopk.py
Jungjee: I'm not sure if this difference is significant enough so that we need a new function/class..

espnet2/lid/loss/aamsoftmax.py
Qingzheng-Wang: But please consider, LID is designed only for classification, but SPK is designed for speaker embedding extraction.

espnet2/lid/loss/aamsoftmax_subcenter_intertopk.py
Qingzheng-Wang: I sincerely apologize, but I have already explained my considerations in detail.
It is possible to directly modify the SPK task, but is that really reasonable? From my understanding, the LID task primarily focuses on classification, while SPK provides embeddings that can support many more tasks, such as verification, in addition to classification.

Why would adding a separate loss implementation be considered a simple duplication? My intention was simply to ensure that the LID task is implemented correctly and does not interfere with the original design of the SPK task.

espnet2/lid/loss/aamsoftmax.py
sw005320: It's more like a balance of 1) having each module separate (more easily modularized) or 2) avoiding duplicated code.
We can make a decision one by one.
For this particular function, I think it is OK for me to duplicate it with some comments about how they differ.

Jungjee: I'm fine with however we proceed as long as @sw005320 is fine with it.
But, @Qingzheng-Wang , please note that the same losses in SPK side is also used for classification only because they are used only in the training phase. The way I see it, having acc printed out wouldn't harm SPK side.

More specifically,  

> This implementation returns predicted labels (via argmax) during inference, enabling output with LID predictions during inference.

Has nothing to do with SPK because in ESPnet-SPK, loss functions are detached in the first place

> It also computes and returns accuracy during training and evaluation, making it easier to monitor model performance.

Again, returning accuracy during training would be a plus for SPK side. 

Hope this clarifies!

sw005320: @Qingzheng-Wang, I think @Jungjee's points make sense to me.
So, I think we should follow his suggestion.

Qingzheng-Wang: Sure, I will make the modifications to the losses and poolings directly under the SPK module. Thanks for the clarifications and suggestions.

Jungjee: Thanks, @sw005320 and @Qingzheng-Wang!

sw005320:
@Jungjee, could you review it again since this PR now modifies the SID-related sources based on your recommendation.

Jungjee:
Let me take a look tonight!

espnet2/spk/espnet_model.py
```diff
@@ -99,12 +101,12 @@ def forward(
...
         # 1. extract low-level feats (e.g., mel-spectrogram or MFCC)
         # Will do nothing for raw waveform-based models (e.g., RawNets)
-        feats, _ = self.extract_feats(speech, None)
+        feats, feat_lengths = self.extract_feats(speech, speech_lengths)
 
         frame_level_feats = self.encode_frame(feats)
 
         # 2. aggregation into utterance-level
-        utt_level_feat = self.pooling(frame_level_feats, task_tokens)
+        utt_level_feat = self.pooling(frame_level_feats, feat_lengths=feat_lengths)
```
Jungjee, line 109: is the removal of `task_tokens` okay for the SPK side?

espnet2/spk/loss/aamsoftmax.py
```diff
@@ -15,65 +14,100 @@
...
+        nout: int,
+        nclasses: int,
+        margin: float = 0.3,
+        scale: int = 15,
+        easy_margin: bool = False,
+        **kwargs,
     ):
         super().__init__(nout)
 
-        self.test_normalize = True
```
Jungjee, line 34: Doesn't this break the SPK side at test phase? 
Please let me know if you checked this.

Jungjee (commented):
Thank you! 
I believe you would have tested already with LID.
But may I ask if you've tested the SPK side as well so that this PR doesn't break the SPK side? I was expecting something on the `spk/espnet_model.py` or relevant files to have some updates.

espnet2/spk/espnet_model.py
Qingzheng-Wang: Yes. None of the three pooling functions require task_tokens, so its removal will not affect the SPK side.

espnet2/spk/loss/aamsoftmax.py
Qingzheng-Wang: The test cases for spk have passed, and I could not find any usage of `test_normalize`, so I removed it.

espnet2/lid/espnet_model.py
```diff
@@ -0,0 +1,160 @@
...
+        else:
+            lang_embd = utt_level_feat
+
+        return lang_embd
+
+    def collect_feats(
+        self,
+        speech: torch.Tensor,
+        speech_lengths: torch.Tensor,
+        lid_labels: torch.Tensor = None,
```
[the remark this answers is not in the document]
Qingzheng-Wang: Fixed.

espnet2/spk/pooling/chn_attn_stat_pooling.py
```diff
@@ -24,31 +27,75 @@ def __init__(self, input_size: int = 1536):
...
+    ) -> torch.Tensor:
+        r"""Forward pass of channel-attentive statistical pooling.
+
+        Args:
+            x: Input feature tensor of shape (batch_size, feature_dim, seq_len)
+            feat_lengths: Optional tensor of shape (batch_size,) containing
+                          the valid length of each sequence before padding
+
+        Returns:
+            x: Utterance-level embeddings of shape (batch_size, 2 x feature_dim)
```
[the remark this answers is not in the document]
Qingzheng-Wang: Fixed.

Qingzheng-Wang:
> Thank you! I believe you would have tested already with LID. But may I ask if you've tested the SPK side as well so that this PR doesn't break the SPK side? I was expecting something on the `spk/espnet_model.py` or relevant files to have some updates.

Thanks for your comments. 
Yes, I have tested it by training on VoxCeleb, and confirmed that it does not break the SPK side.

sw005320:
Can you add a test for these new functions?

Qingzheng-Wang:
> Can you add a test for these new functions?

Yes, I will add tests.

sw005320:
OK
I just merged this PR since it took time, but please make sure to add the test for the new functions at your follow up PRs

PR #6161, opened 2025-06-19
[Pre-ESP3] 01 - Remove contraint Pytest version

Fhrozen:
## What did you change?

This pull request includes several updates to dependencies, test functions, and test annotations. The most important changes involve relaxing the version constraint for `pytest`, correcting a return value assignment in a test function, and adding an execution timeout annotation to a test case.

### Dependency updates:
* [`setup.py`](diffhunk://#diff-60f61ab7a8d1910d86d9fda2261620314edcae5894d5aaa236b821c7256badd7L94-R94): Relaxed the upper version constraint for `pytest` in the test dependencies, allowing versions beyond `<8.4.0`. (`[setup.pyL94-R94](diffhunk://#diff-60f61ab7a8d1910d86d9fda2261620314edcae5894d5aaa236b821c7256badd7L94-R94)`)

### Test function corrections:
* [`test/espnet2/enh/layers/test_complex_utils.py`](diffhunk://#diff-1d337a3d8f93885e45780a919094a96ed1c40d19a621b6727485d780b77dd4c0L166-R166): Fixed the return value assignment in `test_solve` by assigning the result of `torch.linalg.solve` to `ret2` instead of returning it directly. (`[test/espnet2/enh/layers/test_complex_utils.pyL166-R166](diffhunk://#diff-1d337a3d8f93885e45780a919094a96ed1c40d19a621b6727485d780b77dd4c0L166-R166)`)

### Test annotations:
* [`test/espnet2/enh/layers/test_dnsmos.py`](diffhunk://#diff-3eaaef39f730cabaf11de2a242e3b8c8ffe1ef85f2dc698f64e67af1471f156bR11): Added a `@pytest.mark.execution_timeout(6)` annotation to the `test_audio_melspec_consistency` function to enforce a timeout limit of 6 seconds for execution. (`[test/espnet2/enh/layers/test_dnsmos.pyR11](diffhunk://#diff-3eaaef39f730cabaf11de2a242e3b8c8ffe1ef85f2dc698f64e67af1471f156bR11)`)

---

## Why did you make this change?

This is a Pre-espnet 3 change to reduce changes on incoming PRs.

## Additional Context

Refs:
#6139

PR #6150, opened 2025-06-13
Apply format documents for new espnet version

Fhrozen:
## What did you change?

This pull request refactors multiple scripts across various directories to improve readability and maintainability by replacing complex `with open(...)` statements with a cleaner multi-line `with` block syntax. The changes affect scripts related to data preparation, text cleaning, and utility functions in ESPnet.

### Refactoring for improved readability:

* **Data preparation scripts:**
  - Updated `egs/an4/asr1/local/data_prep.py` to use multi-line `with` blocks for handling multiple file streams.
  - Similar refactoring applied to `egs/mini_an4/asr1/local/data_prep.py` and `egs/mini_an4/asr_mix1/local/data_prep.py`. [[1]](diffhunk://#diff-e9fc527a98b8b9215fde1d67e5bdede0dd7ea3d0a31329c9c53a3a20bdf1b069L31-R38) [[2]](diffhunk://#diff-e4bebb07076db5a8da99af08859df63d223f73da5df20bbacf235043ed3a4b6cL31-R38)
  - Refactored `egs/ru_open_stt/asr1/local/ru_open_stt_prepare_data.py` to improve file handling.
  - Updated `egs2/accentdb/asr1/local/data_prep.py` for cleaner file stream management.

* **Text cleaning scripts:**
  - Refactored `egs/jnas/asr1/local/clean_text.py`, `egs/jsut/tts1/local/clean_text.py`, and `egs/jvs/tts1/local/clean_text.py` to use the multi-line `with` block syntax for better readability. [[1]](diffhunk://#diff-9247e092c146f84d905b12985dd3cd6c05e10b3ea7b868d5493af64894bcc6d3L36-R39) [[2]](diffhunk://#diff-eca82927edfede9da71c90ff6f075e9b2d0953b6c82a8476946148e71af0b474L36-R39)
  - Updated `egs2/aphasiabank/asr1/local/clean_hyp_annotations.py` and related scripts to adopt the same syntax. [[1]](diffhunk://#diff-9d9961dd1bd37d6272748ee3c45fc0c77b0b55491cc9773a0d63b340da388223L31-R34) [[2]](diffhunk://#diff-713478a44919a6127104874b93cac963078ad42b14e9351897947c121f327065L13-R16) [[3]](diffhunk://#diff-0ef580b98f6fa893fb0e12606b096115548cbb3f872415dbc7866b3ad5055774L14-R17)

* **Utility scripts:**
[...]

PR #6164, opened 2025-06-19
[espnet3] Format files

Masao-Someki:
## What did you change?

Simply adding white space on espnetez/__init__.py. This will be removed by formatter.

---

A commit is required to make PR and trigger precommit.ci

---

## Is your PR small enough?

No. But this is because it includes formatting on the entire espnet repository.

---

## Additional Context

#6139 
#6150 
#6161


Masao-Someki:
Once precommit.ci completes formatting and these changes are merged into espnet/espnet3, the formatting issue in #6139 will be resolved.

Masao-Someki:
I’ll take a closer look at how black behaves in pre-commit.ci.

sw005320:
OK for me.
Can I merge this?

Masao-Someki:
@sw005320 I found the exact commit that changes130~ files by formatter. First I separated the commit into this PR for #6139. Now it is okay to merge into espnet/espnet3.
I will also create a PR to include all files in precommit. After merging this, formatting issue in #6139 should resolved.

pyproject.toml
sw005320, line 1: Should it be included in this PR?
My understanding is that this PR is only for formatting.

tools/setup.py
sw005320, line 1: Ditto

Should it be included in this PR?
My understanding is that this PR is only for formatting.

pyproject.toml
Masao-Someki: @sw005320 Thank you. It seems that without this commit, the formatter wouldn't be applied to those files.
To be honest, I was hesitant to include it, but this approach is the simplest and cleanest way to separate the formatting into a PR.

sw005320: OK, then can you add TODO to your roadmap and the source code to fix the duplicated pyptorject and setup.py issues?

Masao-Someki: Thank you— just to clarify, I believe there's no actual duplication between `pyproject.toml` and `setup.py` in this context.

The `setup.py` file has already been moved under the `tools/` directory and is no longer part of the build process for ESPnet3. It’s still needed for ESPnet2 CI, which is why we’re keeping it in `tools/`.

The reason for including `pyproject.toml` in this PR is that it changes how the formatter behaves — some directories (like `tools/`) were not previously formatted, and now they are. That’s why the formatting scope appears broader here.

sw005320: Just to clarify, I believe I mentioned that there would be confusion if we maintain both files.
We should at least provide some documents about it (otherwise, it will just confuse people).

Why do we need espnet2 CI in espnet3?

Masao-Someki: > We should at least provide some documents about it

I agree. I'll leave a comment at the top of `setup.py` suggesting that documentation be added. I think it’s best not to include it in this PR.



> Why do we need espnet2 CI in espnet3?

Since ESPnet3 relies on ESPnet2 (e.g., model definitions), having ESPnet2 CI helps ensure compatibility.

PR #6165, opened 2025-06-20
precommit+black to include all files

Masao-Someki:
## What did you change?

- Include all files under egs, egs2, espnet, espnet2, espnetez as formatter target.
- To properly exclude TEMPLATE files I used `force-exclude`

---

By formatting all files under the five directory, we will not suffer from weird formatting issue anymore in the future PR.

---

## Is your PR small enough?

Yes. But since this change will let the formatter to check all files under `egs`, `egs2`, `espnet`, `espnet2`, `espnetez`, there might be additional formatted files in this PR.

---

## Additional Context

#6133 
#6164

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -17,7 +17,7 @@ repos:
     rev: 25.1.0
     hooks:
     -   id: black
-        args: ["egs", "egs2", "espnet", "espnet2", "espnetez",
+        args: ["egs2", "espnet2", "espnetez",
                "--force-exclude", "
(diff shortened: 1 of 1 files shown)


Masao-Someki:
It seems including all files cause timeout. I will remove espnet-1 related files to see if still we have timeout.

Masao-Someki:
I looked into using `pre-commit`, but I couldn't find a way to manually configure a timeout. I also tried moving `pre-commit` to a `pre-push` hook, but then realized it would require developers to have `pre-commit` installed locally as a dependency.

The likely cause of the timeout is the inclusion of `egs2` and `espnet2`, but these directories are essential and cannot be excluded. Since there’s currently no good workaround for this, I will go ahead and close this PR.

That said, we can continue using the current `pre-commit` configuration, which applies formatting only to the updated files.

PR #6171, opened 2025-06-25
[espnet3-2] Add Config Loading script 

Masao-Someki:
## What did you change?

Added a utility module to support dynamic and recursive configuration loading for ESPnet-3, based on OmegaConf and Hydra conventions:

* **New module: `espnet3.utils.config`**

  * `load_config_with_defaults(path)`: Implements recursive merging of YAML configurations using Hydra-style `defaults`, without relying on Hydra's runtime.
  * Supports:

    * `_self_` inclusion
    * key-value based includes (e.g., `optimizer: adam` → loads `optimizer/adam.yaml`)
    * nested and recursive config resolution
   
* **Custom OmegaConf resolver**

  * Registered resolver: `load_line`, which loads and strips lines from a file (e.g., for vocab lists).

* **Added comprehensive unit tests** in `test_config.py`:

---

## Why did you make this change?

Hydra’s `@hydra.main` decorator and runtime logic impose restrictions when configurations need to be programmatically loaded or overridden (e.g., in multi-node job submissions, or config-only pipelines).

This utility offers:

* Hydra-compatible YAML resolution without runtime dependency

---

## Is your PR small enough?

Yes.
Note: This PR is developed upon the following PRs:
 * #6139
 * #6167 
and should be merged into espnet/espnet3 after both PR has merged.
Once #6139 and #6167 are merged, this PR will contain 4 commits, 3 changed files, and 393 additions.

---

## Additional Context

* #6133
The `load_config_with_defaults` utility is expected to become the standard config entrypoint for all ESPnet-3 scripts going forward.

PR #6167, opened 2025-06-22
[espnet3-1] Add Data Organizer

Masao-Someki:
## What did you change?

1. Removed ESPnet-EZ, as we are transitioning to ESPnet-3.
2. Added a data organizer module along with accompanying unit tests.

---

## Why did you make this change?

The new data organizer is designed to support multi-dataset experiments. It allows users to easily add or remove datasets from their experiments, improving flexibility and modularity.

---

## Is your PR small enough?

Yes. Although the removal of ESPnet-EZ results in 14 files being deleted, only 4 new files are added.
The added code is well-documented, with comprehensive docstrings, which accounts for a relatively high number of lines.

---

## Additional Context

#6133

PR #6172, opened 2025-06-25
[espnet3-3] Add trainer and model

Masao-Someki:
## What did you change?

Added the main training script for ESPnet-3, including the following components:
- trainer.py
- model.py
- dataloader.py (split from model.py due to its large functionality)

---

To support training in ESPnet-3 using PyTorch Lightning. This implementation is based on Yifan’s ESPnet-2 Lightning trainer and has been updated to offer a more flexible interface.

---

## Is your PR small enough?

Yes, the core content of this PR is focused and concise. Although it includes a significant number of additions, the code is well-documented and thoroughly tested.
Note that this PR depends on the following related PRs:

#6139
#6167
#6171

Once these are merged, the total number of affected files will be 13.

---

## Additional Context

#6133


Masao-Someki:
## Test Case for ESPnet3LightningTrainer


**Hydra-based Component Initialization**
| Test Name                  | Description                                                                  |
|---------------------------|------------------------------------------------------------------------------|
| test_logger_variants      | logger config (None or dict) correctly instantiates a TensorBoardLogger      |
| test_accelerator_variants | accelerator config (None or dict) correctly sets CPUAccelerator              |
| test_strategy_variants    | strategy config (None or DDP) sets correct training strategy                 |
| test_profiler_variants    | profiler config is properly instantiated (PassThrough or SimpleProfiler)     |
| test_plugins_variants     | plugins config correctly sets HalfPrecision plugin or None                   |

**Config Overrides Based on ESPnet3 Convention**
| Test Name                              | Description                                                                  |
|----------------------------------------|------------------------------------------------------------------------------|
| test_reload_dataloaders_enforced       | Forces `reload_dataloaders_every_n_epochs=1` if `iter_factory` is enabled   |
| test_distributed_sampler_disabled_if_espnet_sampler | Forces `use_distributed_sampler=False` when ESPnet sampler is used     |

**Method Delegation (fit / validate)**
| Test Name                    | Description                                                                  |
|-----------------------------|------------------------------------------------------------------------------|
| test_fit_calls_trainer_fit  | Ensures `fit()` forwards args/kwargs to `trainer.fit()` with model injection |
| test_validate_calls_trainer_validate | Ensures `validate()` forwards args/kwargs to `trainer.validate()`        |

**Assertion & Required Fields**
| Test Name                    | Description                                                                  |
|-----------------------------|------------------------------------------------------------------------------|
| test_missing_model_raises   | Raises `TypeCheckError` if `model` is None                                   |
| test_missing_expdir_raises  | Raises `TypeCheckError` if `expdir` is None                                  |
| test_missing_config_raises  | Raises `TypeCheckError` if `config` is None                                  |

Masao-Someki:
## Test Case for LitESPnetModel


**Normal Cases**
| Test Name                          | Description                                                              |
|-----------------------------------|--------------------------------------------------------------------------|
| test_training_step_runs           | Verifies that training_step executes and returns a loss                  |
| test_validation_step_runs         | Verifies that validation_step runs and returns a loss                    |
| test_is_espnet_sampler_flag       | Checks that is_espnet_sampler is set correctly based on iter_factory     |
| test_state_dict_and_load          | Ensures state_dict and load_state_dict are correctly forwarded           |
| test_use_espnet_collator_flag_false | Checks dataset.use_espnet_collator=False for non-CommonCollateFn        |

**Error / Edge Cases**
| Test Name                          | Description                                                              |
|-----------------------------------|--------------------------------------------------------------------------|
| test_nan_loss_skip                | Ensures batch is skipped when loss is NaN/Inf                            |
| test_dataloader_mismatch_raises  | Asserts error if train/valid dataloader types mismatch                   |
| test_mixed_optim_scheduler_raises| Asserts error if both optim and optims/schedulers are defined            |
| test_missing_optimizer_and_scheduler_raises | Raises ValueError when neither optimizer nor scheduler is specified     |

Masao-Someki:
## Test Case for DataLoaderBuilder


**Standard PyTorch DataLoader mode**
| Test Name                          | Description                                                              |
|-----------------------------------|--------------------------------------------------------------------------|
| test_batch_sampler_only           | Uses only batch_sampler and omits batch_size and shuffle                |
| test_sampler_only                 | Uses only sampler to drive the dataloader                               |
| test_collate_fn_none              | Sets collate_fn=None to use PyTorch's default_collate                   |
| test_common_collate_fn            | Uses CommonCollateFn and checks audio/audio_lengths formatting          |
| test_custom_collate_fn            | Uses a custom collate function and checks the batch structure           |
| test_sampler_and_batch_sampler_conflict | Ensures AssertionError if both sampler and batch_sampler are defined  |

**IterFactory Mode & YAML-based Integration**
| Test Name                          | Description                                                              |
|-----------------------------------|--------------------------------------------------------------------------|
| test_iter_factory_from_default_yaml_with_organizer | Uses a real YAML config with iter_factory and verifies batch structure |
| test_iter_factory_with_collate_fn | Confirms config-defined collate_fn takes precedence over argument       |

**Multiple Iterator Mode (Sharded Dataset)**
| Test Name                              | Description                                                              |
|---------------------------------------|--------------------------------------------------------------------------|
| test_multiple_iterator_shard_initialization | Checks correct shard is selected based on epoch                         |
| test_multiple_iterator_epoch_shard_switching | Ensures different shard is selected as epoch changes                  |

espnet3/trainer/model.py
```diff
@@ -0,0 +1,331 @@
+import logging
+
+import lightning as L
```
sw005320: Do you need to make it as `L`?
It is not frequently used, and `lightning` is more explicit

espnet3/trainer/trainer.py
```diff
@@ -0,0 +1,199 @@
+import warnings
+from argparse import Namespace
+from typing import Any, Dict, Union
+
+import lightning as L
```
sw005320: ditto
Do you need to make it as `L`?
It is not frequently used, and `lightning` is more explicit

espnet3/trainer/model.py
```diff
@@ -0,0 +1,331 @@
...
+from espnet3.trainer.dataloader import DataLoaderBuilder
+
+logger = logging.getLogger("lightning")
+
+
+class LitESPnetModel(L.LightningModule):
+    """
+    ESPnet3-specific PyTorch LightningModule wrapper.
+
+    This class handles model training, validation, optimizer/scheduler setup,
```
sw005320, line 23: add a description about organizer

espnet3/trainer/model.py
```diff
@@ -0,0 +1,331 @@
...
+        valid_dataset: Validation dataset.
+        collate_fn (Callable): Collation function used in DataLoader.
+        is_espnet_sampler (bool): Whether the model uses ESPnet's custom sampler.
+    """
+
+    def __init__(self, model, config):
+        super().__init__()
+        self.config = config
+        self.model = model
+        organizer = instantiate(config.dataset)
```
sw005320: I think `organizer` is a bit confusing.
Can you rename it?

espnet3/trainer/model.py
```diff
@@ -0,0 +1,337 @@
+import logging
+
+import lightning
+import torch
+from hydra.utils import instantiate
+from omegaconf import OmegaConf
+
+from espnet2.train.collate_fn import CommonCollateFn
+
+# Temporarily disabled for the code review.
```
popcornell, line 10: what does this mean ? Is it because you want this PR to be independent ?

espnet3/trainer/model.py
```diff
@@ -0,0 +1,337 @@
...
+        ), "Train and valid should have the same type of dataloader."
+
+        self.is_espnet_sampler = is_train_espnet
+
+        self.collate_fn = CommonCollateFn(int_pad_value=-1)
+        # define collate_fn. Default to ESPnet's CommonCollateFn
+        if hasattr(self.config.dataloader, "collate_fn"):
+            self.collate_fn = instantiate(self.config.dataloader.collate_fn)
+
+    def _sync2skip(self, flag_skip):
```
popcornell, line 78: Another way to do this is https://github.com/Lightning-AI/pytorch-lightning/issues/4956#issuecomment-1722500885 which will check also for invalid gradients.

popcornell: However it will need some testing for ddp. (I guess it should work because it calls self.zero_grad())

espnet3/trainer/model.py
sw005320: Yes.
He split several PRs and needs this treatment.
Please ignore them.

espnet3/trainer/model.py
Masao-Someki: Thank you! This was copied from Yifan's Lightning model module in ESPnet2. Since it's already in use there, I didn't want to introduce too many changes. The implementation looks good, so I'll leave this item in the load map for now—we can revisit and implement it later.

sw005320: Thanks, @popcornell.
Yeah, ddp test is a bit complicated, and we may take this option, but we may revisit once espnet3 recipe becomes more stable.

espnet3/trainer/model.py
```diff
@@ -0,0 +1,337 @@
...
+        self.valid_dataset = data_organizer.valid
+        self.nan_countdown = 0
+
+        # If user is trying to use both Pytorch dataloader and ESPnet's dataloader
+        # Then raise an error here.
+        is_train_espnet = False
+        if (
+            hasattr(self.config.dataloader.train, "iter_factory")
+            and self.config.dataloader.train.iter_factory is not None
+        ):
```
Emrys365, line 57: Can be shortened as

```python
        if getattr(self.config.dataloader.train, "iter_factory", None) is not None:
```

espnet3/trainer/model.py
```diff
@@ -0,0 +1,337 @@
...
+            is_train_espnet = True
+
+        is_valid_espnet = False
+        if (
+            hasattr(self.config.dataloader.valid, "iter_factory")
+            and self.config.dataloader.valid.iter_factory is not None
+        ):
```
Emrys365, line 64: Can be shortened as

```python
        if getattr(self.config.dataloader.valid, "iter_factory", None) is not None:
```

espnet3/trainer/model.py
```diff
@@ -0,0 +1,337 @@
...
+
+        any_invalid = self._check_nan_inf_loss(loss, batch_idx)
+        if any_invalid:
+            # skip this batch altogether on all workers.
+            return None
+
+        new_stats = {}
+        for k, v in stats.items():
+            if v is not None:
+                new_stats[f"{mode}/{k}"] = v.item()
```
Emrys365, line 165: Here, is `v` always a Tensor? If it belongs to any other type (e.g., float), this line will fail.

espnet3/trainer/trainer.py
```diff
@@ -0,0 +1,199 @@
...
+        # Since espnet's sampler requires to set the following configs:
+        # Reload dataloaders every epoch to reuse ESPnet's dataloader
+        # reload_dataloaders_every_n_epochs=1
+        # ESPnet's dataloader already shards the dataset based on distributed setups
+        # use_distributed_sampler=False
+        if hasattr(self.model.config.dataloader.train, "iter_factory"):
+            if (
+                hasattr(self.config, "reload_dataloaders_every_n_epochs")
+                and self.config.reload_dataloaders_every_n_epochs != 1
+            ):
```
Emrys365, line 114: Can be shortened as

```python
        if getattr(self.config, "reload_dataloaders_every_n_epochs", 1) != 1:
```

espnet3/trainer/trainer.py
```diff
@@ -0,0 +1,199 @@
...
+                warnings.warn(
+                    "ESPnet's dataloader requires to set"
+                    "reload_dataloaders_every_n_epochs = 1. "
+                    "Override the config to 1."
+                )
+            if (
+                hasattr(self.model.config, "use_distributed_sampler")
+                and not self.config.use_distributed_sampler
+            ):
```
Emrys365, line 123: Why using both `self.model.config` and `self.config` here? Are they the same object?

sw005320:
Thanks, @Emrys365.

@Masao-Someki, can you reflect them by making a follow-up email?

PR #6175, opened 2025-06-27
[espnet3-5] Add parallel module and collect_stats

Masao-Someki:
## What did you change?

- Added a `espnet3.parallel` module based on `Dask` for supporting parallel processing on various environments.
- Added a dask-based `collect_stats` module.
- The `collect_stats` function supports multiple execution modes: local, parallel (Dask-based), and multiple iterator (sharded datasets).

------

## Why did you make this change?

- To enable robust and flexible collection of feature-level statistics required for normalization or data analysis.
- `espnet3/parallel` is also required for parallel decoding in evaluation stage.

------

## Is your PR small enough?

Yes.

This PR mainly adds 2 files (`espnet3/parallel/parallel.py` and `espnet3/collect_stats.py`) with around 700 lines

------

## Additional Context

* #6133
* #6139 
* #6167 
* #6171 
* #6172 
* #6173

Note: Since I couldn't find a way to run CI with parallel processing, this PR does not contain unit tests.

PR #6173, opened 2025-06-25
[espnet3-4] Add support for multiple optimizers and schedulers

Masao-Someki:
## What did you change?

- Added new HybridOptim and HybridLRS classes under espnet3/trainer to enable unified management of multiple optimizers and schedulers.
- Enhanced LitESPnetModel to allow flexible configuration of multiple optimizers/schedulers, including strict validation:
    - Continued support for single optimizer/scheduler usage.
    - Each optimizer can be assigned parameters via the params key, with checks for overlaps, missing assignments, and required fields.
- Added comprehensive unit tests:
    - test/espnet3/test_hybrid_optim.py and test_hybrid_scheduler.py for HybridOptim and HybridLRS
    - test_model_with_optim_scheduler.py for various configuration and error scenarios (including both valid and invalid patterns).

---

## Why did you make this change?

- To support advanced training strategies such as per-module or partial training using multiple optimizers/schedulers.
- To avoid issues in PyTorch Lightning where training_step is called multiple times when using multiple optimizers.
- To improve maintainability and reliability by catching configuration and assignment errors early.

---

## Is your PR small enough?

Yes, the core content of this PR is focused and concise. Although it includes a significant number of additions, the code is well-documented and thoroughly tested.
Note that this PR depends on the following related PRs:

* #6139
* #6167
* #6171
* #6172

---

## Additional Context

- Core implementation is based on community discussion and recipes ([reference](https://github.com/Lightning-AI/pytorch-lightning/issues/3346#issuecomment-1478556073)), but with additional validation and safety checks.
- The test suite covers both normal and edge cases to ensure future extensibility and regression safety.
- #6133

PR #6178, opened 2025-06-29
[espnet3-6] Add evaluation scripts

Masao-Someki:
## What did you change?

* Added a new inference module under `espnet3/inference/`, including:

  * `abs_metrics.py`: Abstract base class for evaluation metrics.
  * `inference_runner.py`: A comprehensive, configurable inference engine supporting streaming and parallelism.
  * `score_runner.py`: Evaluation runner for decoding outputs using custom metrics.

* Introduced `espnet3/task.py` with helpers for task-specific model instantiation (`get_espnet_model`) and configuration saving.

---

To modularize and generalize the inference and evaluation pipelines in ESPnet3:

* Enables easier implementation and reuse of evaluation logic.
* Supports both offline and streaming inference.
* Allows flexible configuration with Hydra and parallel execution with Dask.

---

## Is your PR small enough?

Yes.

* ✅ Fewer than 20 files changed.
* ✅ Fewer than 1000 lines changed.

It has many files under `test_utils/espnet3/`, which is mainly used for invalid test cases.

---

## Additional Context

* #6133
* #6139 
* #6167 
* #6171 
* #6172 
* #6173
* #6175

Note: Since I couldn't find a way to run CI with parallel processing, this PR does not contain unit tests.


Masao-Someki:
### 💡 Provider/Runner Architecture

Based on Shinji's review I refactord runner classes and introduced a Provider/Runner classes that cleanly separates environment setup from execution.
`Provider` instantiates required items such as datasets and models (one per worker), while `Runner` handles the actual computation logic (e.g., `forward`).

It runs **seamlessly across local, multiprocessing, and cluster (Dask) modes** . No code changes needed, just `set_parallel(config.parallel)` and using different config will change the parallel job.

Users only implement `build_*`, and `forward`, all parallel setup and worker initialization happen automatically.

Provider example: this class instantiates dataset and model
https://github.com/Masao-Someki/espnet/blob/31fcab2aae0924d33aef306a4c8f0e9f251a9c15/test/espnet3/test_inference_runner.py#L20-L31

Runner example: this runner implements offline/online stft computation.
https://github.com/Masao-Someki/espnet/blob/31fcab2aae0924d33aef306a4c8f0e9f251a9c15/test/espnet3/test_inference_runner.py#L34-L119

Masao-Someki:
I will keep working on scoring metrics side

Masao-Someki:
I added sample code demonstrating how to run the **OWSM-V4 medium (1B)** model on **1,000 examples** from the *LibriSpeech test-clean* set using an **A40 GPU**.

Sample code: https://gist.github.com/Masao-Someki/6b72bb7e577ff4e5dfd20b38775b9871

| Environment  | # of GPUs | Time (sec) |
| ------------ | --------- | ---------- |
| local        | 1         | 1220       |
| local        | 2         | 695        |
| slurm / sync | 1         | 1336       |
| slurm / sync | 2         | 669        |
| slurm / sync | 4         | 369        |

In the **local** environment, I used a single job with multiple GPUs and ran inference across those GPUs.
 In **slurm/sync** mode, the main job is a lightweight CPU process that submits jobs to the cluster via `sbatch`. Model initialization occurs within those submitted jobs. In this mode, if the parent job is terminated, all child jobs are also terminated.

In **async** mode, a JSON file is generated containing the provider and runner class information, along with the list of data passed to each function (see `spec-f6c6db3b-r0.json` in the gist). Here, the main job is also a lightweight CPU process that submits jobs via Slurm. Unlike the sync mode, child jobs continue running even if the parent job is terminated, since the parent does not wait for child jobs to finish.

I will remove the scorer runner for now, as this PR already includes many new features. The scorer and other related components will be added in a future PR.

Masao-Someki:
@sw005320 @Emrys365 
I think this PR is now ready for review again. I apologize that the runner package has become a bit more complex.
Users should implement their code as shown in the gist, and developers can focus on implementing providers such as `inference_provider.py`

The main complexity is in `BaseRunner._run_async`, where we replaced the original `dask_jobqueue` command and now use it only for submitting jobs.

sw005320:
@Emrys365, can you check it?

Masao-Someki:
I realized we need to let users use the world rank and size in the runner class. I will fix this part!

espnet3/parallel/parallel.py
Emrys365, line 1: In this script, the modifications are removing several checks before submitting tasks. Are they not necessary anymore?

espnet3/runner/base_runner.py
```diff
@@ -0,0 +1,392 @@
...
+def get_job_cls(cluster, spec_path=None):
+    parent_cls = cluster.job_cls
+    assert spec_path is not None
+
+    class ASyncRunnerJob(parent_cls):
+        def __init__(self, *args, worker_extra_args=None, **kwargs):
+            self._user_worker_extra_args = worker_extra_args or []
+            super().__init__(*args, worker_extra_args=worker_extra_args, **kwargs)
+            python = sys.executable
+            current_file_path = Path(os.path.abspath(__file__)).resolve()
```
Emrys365: `Path.resovle` will make the path absolute.

```suggestion
            current_file_path = Path(__file__).resolve()
```

espnet3/runner/base_runner.py
```diff
@@ -0,0 +1,392 @@
...
+            # DictConfig -> dict
+            cfg_dict = OmegaConf.to_container(self.provider.config, resolve=True)
+            main_file = Path(sys.argv[0]).resolve()
+            for p in map(Path, sys.path):
+                try:
+                    rel = main_file.relative_to(p.resolve())
+                    module_name = ".".join(rel.with_suffix("").parts)
+                    break
+                except ValueError:
+                    continue
```
Emrys365: Would it be a problem if multiple parent paths of `main_file` are included in `sys.path`?
For example, when `sys.path` includes
```
['/opt/homebrew/bin',
 '/Users/Emrys/code/speech'
 '/Users/Emrys/code/speech/espnet']
```
The first match `'/Users/Emrys/code/speech'` will be chosen after the for-loop.

espnet3/runner/base_runner.py
```diff
@@ -0,0 +1,392 @@
...
+
+            provider_cls = f"{module_name}" f".{self.provider.__class__.__name__}"
```
Emrys365:
```suggestion
            provider_cls = f"{module_name}.{self.provider.__class__.__name__}"
```

espnet3/runner/base_runner.py
```diff
@@ -0,0 +1,392 @@
...
+                    params=getattr(self.provider, "params", {}) or {},
+                    indices=list(chunk),
+                    world_size=len(chunks),
+                    world_rank=rank,
+                    result_path=str(result_path),
+                    extras={},
+                )
+                spec_path = self.async_specs_dir / f"spec-{job_id}.json"
+                with open(spec_path, "w", encoding="utf-8") as f:
+                    json.dump(convert_paths(asdict(spec)), f, ensure_ascii=False)
```
Emrys365, line 278: Would it be good to always convert the paths in the config `spec` to absolute? I'm not sure whether relative paths will be a problem when submitting asynchronous jobs.

espnet3/runner/base_runner.py
```diff
@@ -0,0 +1,392 @@
...
+            3) Write results to ``async_result_dir``.
+        Args:
+            indices (Sequence[int]): Indices to process.
+
+        Returns:
+            List[Any] | None: Flattened results if gathered; otherwise ``None``.
+
+        Raises:
+            RuntimeError: If Dask client creation fails.
+            ValueError: If async sharding parameters are invalid.
```
Emrys365, line 239: Will Timeout or other errors (e.g., caused when long-time jobs are automatically killed or killed due to OOM) also be handled?

espnet3/runner/inference_provider.py
```diff
@@ -0,0 +1,165 @@
...
+        Example:
+            >>> provider = InferenceProvider(cfg, params={"device": "cuda"})
+            >>> env = provider.build_env_local()
+            >>> env.keys()
+            dict_keys(["dataset", "model", "device"])
+        """
+        config = self.config
+
+        dataset = self.__class__.build_dataset(config)
+        model = self.__class__.build_model(config)
```
Emrys365: Would it be fine to simply replace `self.__class__` with `self`?

espnet3/runner/inference_provider.py
```diff
@@ -0,0 +1,165 @@
...
+            >>> # Minimal sketch; actual keys depend on your subclass
+            >>> from omegaconf import OmegaConf
+            >>> cfg = OmegaConf.create({
+            >>>     "dataset": {"path": "data/test", "split": "test"}
+            >>> })
+            >>> ds = MyInferenceProvider.build_dataset(cfg)
+
+        Notes:
+            - Keep dataset construction side-effect free, except for necessary
+              lazy indexing or memory mapping.
```
Emrys365: Can you list some examples that should be avoided?

espnet3/runner/base_runner.py
Masao-Someki: Thank you!! I will fix this

espnet3/parallel/parallel.py
Masao-Someki: Sorry this turned out to be a bit long, but I think the details are necessary to explain the situation clearly. I’ve been struggling with this part for a while and couldn’t find a good solution, so I decided to remove this check function and document the behavior instead..

For example, consider running inference on the LS test data using OWSM. Since the model and data are too large to send as a list of objects to each worker, we initialize the dataset and model once per worker during initialization, and only send the indices from the main process. The user-defined initialization process is executed in the `setup_fn`, which returns a `dict`. This dictionary is then unpacked and passed to the function that user wants to parallelize.

However, there’s one issue. When users trigger parallel processing (e.g., using the `map` function — see lines 381 and 314 in `espnet3/parallel/parallel.py`), they can pass additional arguments to the function. As a result, the function that runs in parallel receives arguments by unpacking two dictionaries — the one from `setup_fn` and the runtime `kwargs`. If both dictionaries contain the same key, the function receives multiple values for the same argument, and an error occurs.

The check that was removed here verified this situation *before* starting the parallel processing. If duplicate keys were detected, it raised an error and stopped execution. However, in order to perform this check, the main process needs to execute `setup_fn` once to obtain its return keys. Depending on how `setup_fn` is written, this can be time-consuming, memory-intensive, or even require GPU resources — which are otherwise unnecessary for the main process.

In async mode, this is not a major issue because the main process exits right after dispatching the parallel jobs. But in sync mode, the main process would waste resources while the workers perform the actual processing. After considering whether it’s worth consuming those resources just for this check, I decided to remove it.

Now, looking at this again, there are a few possible directions:

- keep the current behavior where a collision causes an error at runtime,
- allow one of the dictionaries to overwrite the other (for example, `runtime kwargs` overwrite `setup_fn`), or
- restore the deleted check to validate keys before execution.

I’d like to open a discussion and hear opinions on which approach would be best.

espnet3/runner/base_runner.py
Masao-Someki: Thank you, I realized this part contains a serious bug. (This only works if users have runner/providers in the main file)
I will fix this part

espnet3/runner/base_runner.py
Masao-Someki: Thank you so much, and you are right, we need to have those paths as absolute paths.

espnet3/runner/base_runner.py
Masao-Someki: Hmm, I don't think we need to put those errors here, as those errors are not raised by espnet-3, but caused by user defined function or environment issue..

espnet3/runner/inference_provider.py
Masao-Someki: Thank you so much, I will remove it. I don't know why I put this..

espnet3/runner/inference_provider.py
Masao-Someki: I think I need to rephrase that sentence, thank you!

What I meant is that we need to avoid dataset initialization which takes very long time. Dask provides an argument that specifies how many seconds it should wait before it terminate workers and we need to adjust this value based on the dataset initialization time. If dataset initialization takes very long and it exceeds this value, all the workers will be terminated. I remember it was `death-timeout`

espnet3/runner/base_runner.py
Masao-Someki: I fixed this by simply getting module and class path, and replace `__main__` with the actual file name.
Thank you!

Masao-Someki:
@sw005320 @Emrys365 
I fixed the code and bugs pointed out by Emry's review. If this looks okay to you, I will now move on to some of the utility that is required for training and move on to the recipe!

sw005320:
Please fix the CI error

Masao-Someki:
The CI error for espnet2 package comes from huggingface side. 
I think this should be fixed in other PR into master branch.

sw005320:
Please check https://github.com/espnet/espnet/pull/6261

Can you import the espnet master now in another PR?

Masao-Someki:
Thank you, I will merge master into espnet3 branch!

Masao-Someki:
@sw005320 
I merged the master branch and all CI has passed. Could you check this PR again and if this looks okay to you, please merge this PR!

Masao-Someki:
When using this runners I found a issue in running parallel processing in sync mode. (rank is not passed via environment variable)
I will re-check this package and fix them

sw005320:
Can you make the test time shorter?
50 second is too large

espnet3/parallel/parallel.py
Emrys365: @Masao-Someki Thanks for your detailed explanation! I tried summarizing the problem you're facing and asked ChatGPT for a solution:

https://chatgpt.com/s/t_68f4481c3eb08191831c709e28258d49

In brief, I think the first three options it provided are worth trying:
1. Ask the user to specify the return keys of the `setup_fn` function explicitly:
    ```python
    setup_fn.output_keys = ["features", "labels", "metadata"]

    def worker(setup_fn):
        expected_keys = getattr(setup_fn, "output_keys", None)
    ```
2. Use dataclasses or typing.Annotated to define return structure:
    ```python
    from dataclasses import dataclass

    @dataclass
    class SetupFnOutput:
        features: list
        labels: list
        metadata: dict

    def setup_fn(data) -> SetupFnOutput:
        ...
        return SetupFnOutput(features=..., labels=..., metadata=...)


    def worker(setup_fn):
        rtype = proc_fn.__annotations__.get("return")
    ```
3. Add a dry-run or “describe” mode
    ```python
    def setup_fn(data=None, dry_run=False):
        if dry_run:
            return {"features": None, "labels": None, "metadata": None}
        # otherwise, perform expensive computation
    ...
    ```

espnet3/runner/base_runner.py
Emrys365: I see. I'm just curious whether it would be easy to debug (based on the error message espnet3 provides) when these errors occur.

espnet3/parallel/parallel.py
Masao-Someki: Thank you!!
I had considered predefining the `dry-run` flag and key, but both require extra user steps and introduce package-specific coding. That’s why I avoided this.. It adds complexity when writing `setup-fn`.

The dataclass idea is good, but it is also introducing additional item that user have to remember when using parallel package.. Just implementing a function to return `dict` is easy to understand/remember, so I feel it's too much to require additional annotation only for one testing the input.

I’ve been leaning toward handling this automatically (overwrite or raise an error) to keep user-side coding minimal.. but do you think it’s acceptable to require that extra step?

espnet3/runner/base_runner.py
Masao-Someki: Oh, thank you, for async mode it is simpler and easier I believe, as all workers are independent job and all logs are in their log files...

Do you think if we need detailed loggings? I think we can put detailed logs to help debugging, but I think we can create them in another PR to put loggings in all packages

espnet3/parallel/parallel.py
sw005320: My preference is to minimize the user-side coding.
So, we can continue it based on this philosophy.
If we have sufficient test code and documentation, we may not have future confusions.

sw005320:
Thanks!

PR #6179, opened 2025-06-30
[espnet3] ESPnet1 Support Sunset and Migration to `espnet2.legacy`

Masao-Someki:
## What did you change?

* Removed all ESPnet1-related CI testing, scripts, and references.
* Deleted ESPnet1 test and integration scripts (`test_python_espnet1.sh`, `test_shell_espnet1.sh`, `test_integration_espnet1.sh`).
* Removed the `espnet` package directory and migrated necessary utilities to `espnet2.legacy`.
* Adjusted all imports to use `espnet2.legacy` instead of `espnet`.
* Cleaned up CI pipelines to reflect ESPnet2-only support.
* Updated configuration files, removing `init: chainer` and replacing it with `init: normal`.

---

* To fully deprecate and remove ESPnet1 support from the repository.
* To streamline CI/CD pipelines and reduce maintenance overhead.
* To ensure a unified and modern codebase centered on ESPnet2.
* To eliminate legacy Chainer-based components which are no longer supported.

---

## Is your PR small enough?

No, but except for the following items, basically all files were updated via simply replacing or copy/paste.

* Selection of functions/classes.
* Update beam-search related unit tests to use espnet-2 functions
* CI fix, chainer removal

We may separate removal of chainer support as another PR to update dependencies.

---

## Additional Context

* #6133


sw005320:
@Masao-Someki, should this be processed at some point?

Masao-Someki:
While it's not strictly necessary, removing ESPnet-1 would allow us to eliminate several dependency-related limitations, such as version conflicts with Chainer dependencies. Since this is essentially the only benefit, I think this is not something that needs to be merged urgently.

.github/workflows/ci_on_ubuntu.yml
```diff
@@ -150,16 +87,17 @@ jobs:
...
-      - name: test utils
-        run: ./ci/test_utils.sh
-      - uses: codecov/codecov-action@v2
-        with:
-          flags: test_utils
-      - name: coverage erase
-        continue-on-error: true
-        run: |
-          source tools/activate_python.sh
-          coverage erase
```
sw005320: Can you make sure?
I think some of them would be used for the entire espnet project.

sw005320: You can ask LLMs to check them.

egs/csj/asr1/local/run_spm.sh
sw005320, line 1: We can remove `egs/` in the next PR

egs2/aishell/asr1/conf/train_asr_rnn.yaml
```diff
@@ -31,7 +31,7 @@ batch_type: folded
 batch_size: 30
 
 # optimization related
-init: chainer
+init: normal
```
sw005320, line 34: This is the initialization method specifically used in Chainer, and I think we don't have to change it.
(Instead, maybe, it's better to add a comment about it.)
`normal` does not sound informative.

sw005320: If this is OK, please change the all related parts.

espnet2/__init__.py
```diff
@@ -1,3 +1,8 @@
-"""Initialize espnet2 package."""
+"""Initialize espnet package."""
```
sw005320: is it OK?
I think we don't have to change this comment

espnet2/legacy/nets/pytorch_backend/transformer/mask.py
```diff
@@ -20,16 +20,3 @@ def subsequent_mask(size, device="cpu", dtype=torch.bool):
...
-    """Create mask for decoder self-attention.
-
-    :param torch.Tensor ys_pad: batch of padded target sequences (B, Lmax)
-    :param int ignore_id: index of padding
-    :param torch.dtype dtype: result dtype
-    :rtype: torch.Tensor (B, Lmax, Lmax)
-    """
-    ys_mask = ys_in_pad != ignore_id
-    m = subsequent_mask(ys_mask.size(-1), device=ys_mask.device).unsqueeze(0)
-    return ys_mask.unsqueeze(-2) & m
```
sw005320, line 35: Is this not used?
Can you confirm it?

espnet2/legacy/utils/cli_utils.py
```diff
@@ -7,12 +7,6 @@
 import numpy
 
 
-def strtobool(x):
-    """Convert string to boolean."""
-    # distutils.util.strtobool returns integer, but it's confusing,
-    return bool(dist_strtobool(x))
-
-
```
sw005320, line 15: Is this not used?
Can you confirm it?

espnet2/tasks/asr.py
```diff
@@ -265,7 +265,6 @@ def add_task_arguments(cls, parser: argparse.ArgumentParser):
             default=None,
             help="The initialization method",
             choices=[
-                "chainer",
```
sw005320, line 268: make sure that this is valid

espnet2/torch_utils/initialize.py
```diff
@@ -22,106 +22,58 @@ def initialize(model: torch.nn.Module, init: str):
...
-                n = data.size(1)
-                stdv = 1.0 / math.sqrt(n)
-                data.normal_(0, stdv)
-            elif data.dim() in (3, 4):
-                # conv weight
-                n = data.size(1)
-                for k in data.size()[2:]:
-                    n *= k
-                stdv = 1.0 / math.sqrt(n)
-                data.normal_(0, stdv)
```
sw005320, line 50: I think we can keep this initialization.
This is not based on the chainer module, but the way to intialize it.

sw005320: But again, if it is not used in recent code, we can remove it to simplify our code.

espnet2/torch_utils/initialize.py
```diff
@@ -22,106 +22,58 @@ def initialize(model: torch.nn.Module, init: str):
...
-                    torch.nn.init.kaiming_normal_(p.data, nonlinearity="relu")
-                elif init == "normal":
-                    torch.nn.init.normal_(p.data, mean=0.0, std=0.02)
-                else:
-                    raise ValueError("Unknown initialization: " + init)
-        # bias init
-        for name, p in model.named_parameters():
-            if ".bias" in name and p.dim() == 1:
-                p.data.zero_()
-                logging.info(f"Initialize {name} to zeros")
```
sw005320, line 90: Are these lines related to the chainer initialization?

espnet2/version.txt
sw005320, line 1: Do you need to change this?
We may move this to the top layer.

pyproject.toml
sw005320, line 1: What is the intention of creating a version for espnet2 instead of the entire espnet project?
Do you want to maintain separate version management for espnet2 and espnet3?
Could you explain this to me?

test/espnet2/legacy/test_batch_beam_search.py
sw005320, line 1: please explain the change of this part

sw005320 (commented):
I checked all changes.

- Again, I want to get some discussions about the `chainer` initialization
- Please explain the beam search part. This seems to have some refactoring.

.github/workflows/ci_on_ubuntu.yml
Masao-Someki: Oh, this is commented out for debug! I’ll fix it

sw005320:
I'm concerned that this will create a significant discrepancy with the ESPnet main repository.
I think it's better to do it on the main repository and merge this change to espnet3 if possible.

espnet2/torch_utils/initialize.py
Masao-Someki: Thank you, I just searched all "chainer" related part and removed it, but now I feel we can simply keep this, it is also required for reproduction.
I will recover this and `init: chainer` part.

sw005320: Again, I'm actually OK to remove it as it simplifies the code.
Again, if they are not used in the recent recipes or some important recipes (like LibriSpeech), we can safely remove them now.

Masao-Someki: I reviewed all occurrences of `init: chainer`, and it only appears in fairly old configurations (about 4–5 years ago). The most recent reference is in the MLS config, which was created 2 years ago.
I think we can safely remove this option.

espnet2/version.txt
Masao-Someki: Thank you, in the previous ESPnet, we had the version.txt file under `espnet/`, so I initially placed it in `espnet2/`.
I’ll move it to the top-level directory instead.

espnet2/legacy/nets/pytorch_backend/transformer/mask.py
Masao-Someki: Yes, it is not used in any espnet2/espnet3 projects
image

espnet2/legacy/utils/cli_utils.py
Masao-Someki: This is also not used, basically we use `distutils.util.strtobool`
image

test/espnet2/legacy/test_batch_beam_search.py
Masao-Someki: At that time I thought beam search can be one of the main feature of espnet. (or updated espnet-2.) So I thought it might not be proper to place it in `legacy` package. But I feel we can do this in another PR if required, we can just keep it in `legacy` as other package now. Sorry for the confusion.

Masao-Someki:
I understand your concern, it is true that this PR would resulting in a lot of diffs with master.. But ESPnet-3 development was initially okay with breaking changes, including some developments that are not backward compatible. This is one of the reasons I thought we can consider removing ESPnet-1 within ESPnet-3 development.
If we plan to merge this into the current master branch, it will initially only result in version removal. Therefore, we should announce it and explain to users in advance why this refactoring is necessary.
It may also be better to wait until after the ICASSP/ICLR deadline if we want to merge this change into master, since people may still be relying on it for their projects..

Masao-Someki:
How about we first make sure everything goes well with the espnet3 branch, and then gradually merge it into master?

sw005320:
My understanding was that you requested to import the master changes to fix the CI issue.
I think it's better to fix the CI issue first.
I think it's better to frequently update the master change.

Some changes (e.g., LID-related) are important and you should incorporate to espnet3

sw005320:
If you really want to avoid it, I'd like you to discuss it in https://github.com/espnet/espnet/pull/6231, where you're a reviewer of this PR, and you can expect this situation....

pyproject.toml
```diff
@@ -57,14 +57,13 @@ requires = ["setuptools==73.0.1", "wheel", "setuptools-scm>=8.0,<9.0"]
 build-backend = "setuptools.build_meta"
 
 [tool.setuptools.dynamic]
-version = {attr = "espnet.__version__"}
+version = {attr = "espnet2.__version__"}
```
sw005320, line 60: Did you decide to use `espnet2`?
Is it to make a distinction with `espnet3`?

(either is fine, but I want to confirm)

.github/workflows/ci_on_debian11.yml.eol
```diff
@@ -42,11 +42,9 @@ jobs:
         run: ./ci/install.sh
       - name: test shell
         run: |
-          ./ci/test_shell_espnet1.sh
```
sw005320, line 45: is it OK to remove it?
some of them are used for espnet2 recipes I believe

.github/workflows/ci_on_debian11.yml.eol
```diff
@@ -42,11 +42,9 @@ jobs:
...
           ./ci/test_shell_espnet2.sh
           # Note: ESPnet-3 does not have shell, so we will not check.
       - name: test python
         run: |
-          ./ci/test_python_espnet1.sh
```
sw005320, line 50: Is it OK to exclude this?
According to https://github.com/espnet/espnet/blob/master/ci/test_python_espnet1.sh#L19, this test seems to include many other espnet2-related tests (maybe, I'm wrong, but I could read like this)

sw005320:
Can you update the numpy version?
Now, it supports Numpy 2.0

pyproject.toml
Masao-Someki: No, simply espnet directory was removed, so I just using the espnet2 directory to get version info

.github/workflows/ci_on_debian11.yml.eol
Masao-Someki: I think we can remove this line. The script currently checks:
- flake8 for the espnet directory
- format check with pycodestyle (with several excluded directories)
- tests in test/

For the pycodestyle part, it is true that it checks the espnet2 directory in the current code. However, the exclude settings are exactly the same as those in `test_python_espnet2.sh`, so it will also checked in `test_python_espnet2.sh`.

https://github.com/espnet/espnet/blob/eaf9a83f5f5a7427651b2eb4b6438a9a6427ebf0/ci/test_python_espnet2.sh#L8C1-L16

sw005320: So, all test functions in https://github.com/espnet/espnet/tree/master/test are not related to espnet2?
I think so for some of them, but I'm not sure about all of them.
But maybe we can trust the espnet2 test.
In this case, we can also remove test files in https://github.com/espnet/espnet/tree/master/test (in another PR? Either is fine)

.github/workflows/ci_on_debian11.yml.eol
Masao-Someki: Thank you. I found that the following scripts were tested with `test_shell_espnet1.sh` and used from the `egs2`/`espnet2` directory:

* `utils/trim_silence.sh`
* `utils/make_fbank.sh`
* `utils/remove_longshortdata.sh`
* `utils/download_from_google_drive.sh`
* `utils/reduce_data_dir.sh`
* `utils/feat_to_shape.sh`

Since all of them are in the `utils/` directory, I’ll include the entire `utils` directory in `test_shell_espnet2.sh`.

Masao-Someki:
@sw005320 
For `ci/test_utils.sh`, I reviewed all the bats files and removed the test file that checks files not used in ESPnet-2. Since all bats files for integration tests have been removed, `test_utils.sh` now only runs unit tests. However, these unit tests were not executed in the original file, so we could consider removing the `test_utils.sh` process entirely..

sw005320:
Cool!
This is a great advance for espnet3

PR #6184, opened 2025-07-11
[svs]: fix bugs in svs.py

South-Twilight:
## What did you change?

update svs.py

---

fix bugs in svs.py

---

## Is your PR small enough?

yes

---


ftshijt:
Thanks for your fix!

PR #6188, opened 2025-07-17
docs: clarify CBHG encoder vs post‑net roles in Tacotron 1

ZhuoyanTao:
## What did you change?
Update the comment section immediately below CBHG class definition

---

The comment in the cbhg implementation did not correctly explain the roles of cbhg from Tacotron 1 in encoder and decoder.

---

## Is your PR small enough?

yes

---


sw005320:
Thanks!
This is a good clarification!

PR #6197, opened 2025-08-05
Workaround for shellcheck v0.11.0

Masao-Someki:
## What did you change?

Applied a workaround to fix compatibility issues introduced by ShellCheck v0.11.0.
Specifically:
- Replaced usage of `shellcheck-stable` with an explicit download and usage of `shellcheck v0.10.0`.
- Updated both `ci/test_shell_espnet1.sh` and `ci/test_shell_espnet2.sh` scripts accordingly.

---

ShellCheck v0.11.0 introduced issues that break our current CI setup.
To maintain CI stability and avoid downstream problems, we pinned the version to v0.10.0 as a temporary workaround.

---

## Is your PR small enough?

Yes

---

## Additional Context

This workaround should be removed once the problem with v0.11.0 has fixed.

### DIFF
--- a/ci/test_shell_espnet1.sh
+++ b/ci/test_shell_espnet1.sh
@@ -4,7 +4,9 @@ if [ ! -e tools/kaldi ]; then
     git clone https://github.com/kaldi-asr/kaldi --depth 1 tools/kaldi
 fi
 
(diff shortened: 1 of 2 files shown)


sw005320:
Thanks!

sw005320:
Thanks, @Masao-Someki!

PR #6221, opened 2025-08-23
Update numpy version

Fhrozen:
## Notes

Fix #6213

## What did you change?

This pull request updates the project's Python and dependency requirements to ensure compatibility with newer versions and to drop support for older Python versions. The key changes are focused on updating the `numpy` dependency and the supported Python versions in `setup.py`, as well as a minor update to the `tools/Makefile`.

Dependency and compatibility updates:

* Updated the `numpy` requirement in `setup.py` to `numpy>=2.2.0`, removing the previous restriction to versions below 1.24. This ensures compatibility with the latest `numpy` releases.
* Changed the minimum required Python version in `setup.py` from 3.7 to 3.9, and updated the list of supported Python versions to include 3.11, 3.12, and 3.13 while removing 3.7 and 3.8.

Build process update:

* Removed `chainer.done` from the `python` target dependencies in `tools/Makefile`, likely reflecting the project's move away from Chainer.

---

## Additional Context

The part of the code that requires `numpy<2.2` (if any part is found during evaluations) will not be removed, but will raise a `compatibility error`.

PR #6223, opened 2025-08-23
Update Whisper Version

Fhrozen:
## Note

During the update of `numpy` version, I found that `whisper==20230308` is downgrading torch to 2.0.0.
I made this PR to keep separated updates to the numpy version. 

## What did you change?

This pull request updates the handling of the `openai-whisper` dependency in both the `whisper_encoder.py` code and the installation script. The changes ensure compatibility with newer versions of the `openai-whisper` package, particularly addressing the removal of the `N_MELS` constant in recent releases.

Dependency and compatibility updates:

* Updated the installation script (`install_whisper.sh`) to install `openai-whisper` version `20250625` instead of `20230308`, reflecting the move to a newer tested version.
* In `whisper_encoder.py`, added logic to handle the absence of `N_MELS` in versions of `openai-whisper` after `20230308`. If the installed version does not include `N_MELS`, the code now uses the provided `input_size` as a fallback and logs a warning.
* Removed the import of `N_MELS` from the `whisper.audio` module during initialization to prevent import errors on newer versions.

General improvements:

* Added imports for `logging` and `packaging.version.parse` to support version checking and warning messages.


Fhrozen:
Merge this before #6221

sw005320:
I think you can merge this since it is a version change only

PR #6226, opened 2025-08-28
Fix GPU flag handling in tts.sh script

ZhuoyanTao:
## What did you change?

---
Fix the bug that makes versa always use gpu even when gpu_inference is False

---
Previously, the code used --use_gpu False to tell Versa Scorer not to use a GPU. However, versa reads the existence of this option as True, leading to exception when the node doesn't have gpu.

## Is your PR small enough?

---

### DIFF
--- a/egs2/TEMPLATE/tts1/tts.sh
+++ b/egs2/TEMPLATE/tts1/tts.sh
@@ -1155,15 +1155,11 @@ if ! "${skip_scoring}"; then
         if ${gpu_inference}; then
             _cmd="${cuda_cmd}"
             _ngpu=1
+			use_gpu_flag="--use_gpu"
         else
             _cmd="${decode_cmd}"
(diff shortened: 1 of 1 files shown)


sw005320:
@ftshijt, please check it

ftshijt:
Many thanks for catching the bug. I'm good with the usegpu, but not pretty sure for the reason of removing the gt_files for the evaluation, as some metrics will be not available if that is removed. Could you please share your rationale to that?

ZhuoyanTao:
> Many thanks for catching the bug. I'm good with the usegpu, but not pretty sure for the reason of removing the gt_files for the evaluation, as some metrics will be not available if that is removed. Could you please share your rationale to that?

Sorry the deletion was by accident. I've added it back. Thanks for catching that!

sw005320:
Thanks!

PR #6228, opened 2025-08-29
EOL of debian 11 support in favor of debian 12

Fhrozen:
## Notes

During test of #6221, an import issue was found related to numpy>=2.0.0. 
The error: ` version 'GLIBCXX_3.4.29' not found` is because Debian 11 only supports up to gcc-10 (GLIBCXX_3.4.28). 
- Tried to install gcc-11 during package installations, but the package was not found.
- Installed gcc-11 with conda and reinstalled numpy, but numpy did not detect the change.

Before merging the numpy>=2.0.0 PR, it will recommend ending support for Debian 11 and moving to Debian 12.

## What did you change?

This pull request updates the CI workflow to use Debian 12 instead of Debian 11 and updates the documentation to reflect this change. The most important changes are:

**CI Workflow Updates:**

* Added a new GitHub Actions workflow file, `.github/workflows/ci_on_debian11.yml.eol`, which sets up CI jobs to run tests in a Debian 12 container environment with Python 3.10 and updated dependencies.

**Documentation Updates:**

* Updated the CI status badges in `README.md` to reference Debian 12 instead of Debian 11, ensuring the documentation matches the new CI environment.


sw005320:
Thanks!

PR #6248, opened 2025-09-23
Get forced alignments from CTC model

Shikhar-S:
## What did you change?
Add forced alignment outputs based on CTC posteriors for the S2T model.

---

To get forced alignments from any CTC models in ESPnet.

---

## Is your PR small enough?
Yes


sw005320:
Thanks!
This is useful.
How (and when) would it be enabled?
Any example?

sw005320:
Also, please fix the CI errors

Shikhar-S:
> Thanks! This is useful. How (and when) would it be enabled? Any example?

Thanks for review. Added an example here - `egs2/ipapack_plus/s2t1/force_align.py`
It can only be used during inference in current implementation.

Shikhar-S:
Hi, this is ready to be merged.
@sw005320

sw005320:
OK, thanks!

PR #6242, opened 2025-09-16
[espnet3-5] (2) Add parallel module and collect_stats

Masao-Someki:
## What did you change?

- Added a `espnet3.parallel` module based on `Dask` for supporting parallel processing on various environments.
- Added a dask-based `collect_stats` module.
- The `collect_stats` function supports multiple execution modes: local, parallel (Dask-based), and multiple iterator (sharded datasets).

------

## Why did you make this change?

- To enable robust and flexible collection of feature-level statistics required for normalization or data analysis.
- `espnet3/parallel` is also required for parallel decoding in evaluation stage.

------

## Is your PR small enough?

Yes.

This PR mainly adds 2 files (`espnet3/parallel/parallel.py` and `espnet3/collect_stats.py`) with around 700 lines

------

## Additional Context

* #6133
* #6139 
* #6167 
* #6171 
* #6172 
* #6173
* #6175

This PR contains the same changes with #6175, but since the commit history got messed up, I recreated it here.

PR #6180, opened 2025-06-30
[espnet3-7] Add Callbacks

Masao-Someki:
## What did you change?

* Added a new module: `espnet3/trainer/callbacks.py`, which includes:

  * `AverageCheckpointsCallback`: a custom Lightning callback for averaging top-K model checkpoints.
  * `get_default_callbacks()`: a utility to create a standard set of callbacks including checkpointing, progress bar, and LR monitoring.
* Integrated `get_default_callbacks` into the training loop in `espnet3/trainer/trainer.py`.
* Created a new test module `test/espnet3/test_callback.py`:

---

## Why did you make this change?

* To support checkpoint ensembling by averaging top-K models.
* To standardize callback configuration in the trainer, ensuring consistency and reusability.
* To ensure correctness and robustness of the new logic via a comprehensive test suite.

---

## Is your PR small enough?

Yes. 3 files changed, ~500 additions.

---

## Additional Context

* #6133
* #6139 
* #6167 
* #6171 
* #6172 
* #6173
* #6175
* #6178

PR #6249, opened 2025-09-26
[espnet3-7](2) Add Callbacks

Masao-Someki:
## What did you change?

* Added a new module: `espnet3/trainer/callbacks.py`, which includes:

  * `AverageCheckpointsCallback`: a custom Lightning callback for averaging top-K model checkpoints.
  * `get_default_callbacks()`: a utility to create a standard set of callbacks including checkpointing, progress bar, and LR monitoring.
* Integrated `get_default_callbacks` into the training loop in `espnet3/trainer/trainer.py`.
* Created a new test module `test/espnet3/test_callback.py`:

---

* To support checkpoint ensembling by averaging top-K models.
* To standardize callback configuration in the trainer, ensuring consistency and reusability.
* To ensure correctness and robustness of the new logic via a comprehensive test suite.

---

## Is your PR small enough?

Yes. 3 files changed, ~500 additions.

---

## Additional Context

* #6133
* #6139 
* #6167 
* #6171 
* #6172 
* #6173
* #6175
* #6242
* #6178
* #6180

### DIFF
--- a/espnet3/trainer/callbacks.py
+++ b/espnet3/trainer/callbacks.py
@@ -56,7 +56,7 @@ class AverageCheckpointsCallback(Callback):
         self.output_dir = output_dir
         self.best_ckpt_callbacks = best_ckpt_callbacks
 
-    def on_fit_end(self, trainer, pl_module):
+    def on_validation_end(self, trainer, pl_module):
         if trainer.is_global_zero:
             for ckpt_callback in self.best_ckpt_callbacks:
                 checkpoints = list(ckpt_callback.best_k_models.keys())
--- a/espnet3/trainer/callbacks.py
+++ b/espnet3/trainer/callbacks.py
@@ -1,3 +1,4 @@
+import logging
 from pathlib import Path
 from typing import Any, List, Tuple, Union
 
@@ -98,6 +99,10 @@ class AverageCheckpointsCallback(Callback):
                         # (If there are any cases that requires averaging
                         #  or the other reducing method, e.g. max/min, for integer type,
                         #  please report.)
+                        logging.info(
+                            "The following parameters were only accumulated, "
+                            f"not averaged: {k}"
+                        )
                         pass
                     else:
                         avg_state_dict[k] = avg_state_dict[k] / len(checkpoints)
--- a/espnet3/trainer/callbacks.py
+++ b/espnet3/trainer/callbacks.py
@@ -60,18 +60,36 @@ class AverageCheckpointsCallback(Callback):
                 checkpoints = list(ckpt_callback.best_k_models.keys())
 
                 avg_state_dict = None
+                reference_keys = None  
                 for ckpt_path in checkpoints:
                     state_dict = torch.load(
                         ckpt_path,
                         map_location="cpu",
-                        weights_only=False,
-                    )["state_dict"]
+                        weights_only=False
+                    )
 
+                    # for deepspeed checkpoints
+                    if "module" in state_dict:
+                        state_dict = state_dict["module"]
+                    # for PytorchLightning checkpoints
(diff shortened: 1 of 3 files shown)


Masao-Someki:
Details on test cases is listed here: https://github.com/espnet/espnet/pull/6180#issuecomment-3020930189

sw005320:
I think this is a great idea to implement an average checkpoints (or other monitoring values) via callback.

@Emrys365, can you check this PR?

espnet3/trainer/callbacks.py
```diff
@@ -0,0 +1,205 @@
...
+                            avg_state_dict[k] = avg_state_dict[k] + state_dict[k]
+
+                for k in avg_state_dict:
+                    if str(avg_state_dict[k].dtype).startswith("torch.int"):
+                        # For int type, not averaged, but only accumulated.
+                        # e.g. BatchNorm.num_batches_tracked
+                        # (If there are any cases that requires averaging
+                        #  or the other reducing method, e.g. max/min, for integer type,
+                        #  please report.)
+                        pass
```
Emrys365, line 106: Better to report the parameter names that are only accumulated here so that users will be aware of the potential issue.

espnet3/trainer/trainer.py
```diff
@@ -88,19 +86,18 @@ def __init__(
...
+            expdir,
+            self.config.log_every_n_steps,
+            OmegaConf.to_container(best_model_criterion),
+        )
+        if getattr(self.config, "callbacks", None):
+            assert isinstance(
+                self.config.callbacks, ListConfig
+            ), "callbacks should be a list"
+            for callback in self.config.callbacks:
+                callbacks.append(instantiate(callback))
```
Emrys365, line 99: Can it happen that the defined callbacks in `self.config` are duplicates of the default callbacks? In that case, would it better to  detect such cases to display warnings?

Emrys365 (approved):
LGTM in general.

espnet3/trainer/callbacks.py
```diff
@@ -0,0 +1,205 @@
...
+                    else:
+                        avg_state_dict[k] = avg_state_dict[k] / len(checkpoints)
```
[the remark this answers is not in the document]
Masao-Someki: No, this is a trick from espnet2, and confirmed by yifan as well. I will not changing this part.

espnet3/trainer/trainer.py
Masao-Someki: Thank you, the current implementation does not account for cases where callbacks defined in `get_default_callbacks` overlap with those specified in the config. For example, `LearningRateMonitor` may be registered twice, which can result in duplicated log outputs.. 

In the current behavior of Lightning this is not a critical issue, so it does not immediately cause an error.

espnet3/trainer/callbacks.py
```diff
@@ -0,0 +1,212 @@
...
+    TQDMProgressBar,
+)
+from typeguard import typechecked
+
+
+@typechecked
+class AverageCheckpointsCallback(Callback):
+    """
+    A custom PyTorch Lightning callback that performs weight averaging over the top-K
+    checkpoints (according to specified metrics) at the end of training.
```
sw005320, line 20: Does model averaging only happen at the end of training?
This would be an issue when people try to get the intermediate results with the checkpoint.

Masao-Someki: Thank you! Right now it's only hooked into `on_fit_end`, so it fires once at the very end. I will switch to `on_validation_end` to trigge after every validation round.

sw005320: OK!

PR #6253, opened 2025-09-29
Restrict Docker Github Actions to Original Repo

Fhrozen:
## What did you change?

This pull request introduces a conditional check to the Docker publishing workflow to ensure that the job only runs when the current branch is `master` and the repository is `espnet/espnet`. This helps prevent unintended Docker image publishing from forks or non-master branches.

Workflow restriction:

* Added an `if` condition to the `docker` job in `.github/workflows/publish_docker_image.yml` so it only runs when on the `master` branch of the `espnet/espnet` repository.


sw005320:
Thanks, @Fhrozen

PR #6268, opened 2025-10-20
Codec codebase bug fixes: `detach()` in RVQ residual and `target_bandwidth` in inference

whr-a:
* Fix 2 bugs:
  * When calculating the residual in RVQ, `detach()` should be applied. Refer to Cisco's fix: [core_vq.py](https://github.com/cisco-open/espnet/blob/master/espnet2/gan_codec/shared/quantizer/modules/core_vq.py), and Moshi's implementation: [core_vq.py](https://github.com/kyutai-labs/moshi/blob/main/moshi/moshi/quantization/core_vq.py)
  * Bug in passing `target_bandwidth` during inference

### DIFF
--- a/espnet2/gan_codec/dac/dac.py
+++ b/espnet2/gan_codec/dac/dac.py
@@ -440,7 +440,7 @@ class DAC(AbsGANCodec):
             Tensor: Generated codes (T_code, N_stream).
 
         """
-        target_bw = kwargs.get('target_bw', None)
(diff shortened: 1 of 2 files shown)


sw005320:
Oh, this sounds critical.
Thanks for catching up on it.
@ftshijt, can you review this PR?

ftshijt:
Thanks for your fixing! The fixes look great to me.

PR #6270, opened 2025-10-24
[espnet3-8] Bugfix for recipe

Masao-Someki:
## What did you change?

* Updated `espnet3/collect_stats.py` with Runner/Provider classes:

  * This fix includes bugfix in collect_stats.
  
* Bug fixed `base_runner.py`:
  * Wrapped `parallel_for()` with `tqdm` progress bar for visual feedback during distributed runs.
  
* Bug fixed `espnet3/utils/config.py`:

    * Returned `ListConfig` instead of plain list in `load_line()`.
    * Added `OmegaConf.resolve()` after merging configs.

* Bug fixed `espnet3/trainer/model.py`: imported `Path` for consistency with file I/O utilities.

---

* The previous `collect_stats` system used separate implementations (`collect_stats_local`, `collect_stats_parallel`, etc.), which made maintenance and testing difficult.
* In the #6178 we implemented Runner/Provider classes for seamless integration of parallel processing/local execution. 
* This is much easier to maintain, so I modified from the parallel processing into runner/provider.

* Added `tqdm` progress monitoring to improve user experience during long Dask-based runs.
* The `config.py` changes improve Hydra compatibility and ensure proper type resolution in nested configurations.

---

## Is your PR small enough?

✅ **Yes.**

* 5 files changed
* ~450 insertions, ~20 deletions
* All changes are localized and focused on refactoring the `collect_stats` workflow.

---

## Additional Context

* #6133

### DIFF
--- a/espnet3/collect_stats.py
+++ b/espnet3/collect_stats.py
@@ -404,7 +404,7 @@ def collect_stats_multiple_iterator(
     task: Optional[str],
     batch_size: int,
     parallel_config: Any,
-    write_collected_feats: bool,
+    write_collected_feats: bool = False,
 ):
     """Collect stats on sharded datasets using Dask + setup_fn."""
     set_parallel(parallel_config)
--- a/test/espnet3/test_collect_stats.py
+++ b/test/espnet3/test_collect_stats.py
@@ -45,12 +45,13 @@ class DummyDataset:
     Each item has a variable time length and fixed feature dim.
     """
 
-    def __init__(self, n=10, base_len=3, dim=4):
+    def __init__(self, n=10, base_len=3, dim=4, use_espnet_preprocessor=False):
         self.n = n
         self.base_len = base_len
         self.dim = dim
         # Precompute lengths to keep things deterministic
         self.lengths = [self.base_len + (i % 3) for i in range(self.n)]
+        self.use_espnet_preprocessor = use_espnet_preprocessor
 
     def __len__(self):
         return self.n
@@ -61,7 +62,10 @@ class DummyDataset:
         # Create a simple pattern to let us verify sums deterministically
         # Shape: [T, D]
         x = torch.full((T, self.dim), float(idx), dtype=torch.float32)
-        return uid, {"x": x, "length": T}
+        if self.use_espnet_preprocessor:
+            return uid, {"x": x, "length": T}
+        else:
+            return {"x": x, "length": T}
 
     # For multiple-iterator tests: split items into shards round-robin
     def shard(self, shard_idx: int, num_shards: int = 2):
@@ -86,9 +90,21 @@ class DummyDataset:
 class DummyOrganizer:
     """Hydra-instantiable organizer that exposes .train and .valid datasets."""
 
-    def __init__(self, n_train=8, n_valid=5, base_len=3, dim=4):
-        self.train = DummyDataset(n=n_train, base_len=base_len, dim=dim)
-        self.valid = DummyDataset(n=n_valid, base_len=base_len, dim=dim)
+    def __init__(
+        self, n_train=8, n_valid=5, base_len=3, dim=4, use_espnet_preprocessor=False
+    ):
+        self.train = DummyDataset(
+            n=n_train,
+            base_len=base_len,
+            dim=dim,
+            use_espnet_preprocessor=use_espnet_preprocessor,
+        )
+        self.valid = DummyDataset(
+            n=n_valid,
+            base_len=base_len,
+            dim=dim,
+            use_espnet_preprocessor=use_espnet_preprocessor,
+        )
 
 
 class DummyCollate:
@@ -152,7 +168,9 @@ def make_model_cfg(scale: float = 1.0):
     )
 
 
-def make_dataset_cfg(n_train=8, n_valid=5, base_len=3, dim=4):
+def make_dataset_cfg(
+    n_train=8, n_valid=5, base_len=3, dim=4, use_espnet_preprocessor=False
+):
     return OmegaConf.create(
         {
             "_target_": "test.espnet3.test_collect_stats.DummyOrganizer",
@@ -160,6 +178,7 @@ def make_dataset_cfg(n_train=8, n_valid=5, base_len=3, dim=4):
             "n_valid": n_valid,
             "base_len": base_len,
             "dim": dim,
+            "use_espnet_preprocessor": use_espnet_preprocessor,
         }
     )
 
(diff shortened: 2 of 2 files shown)


Masao-Someki:
For the collect-stats part, I realized the current fix is not enough. 
1. I need to remove collect-stats files in utils/
2. Update unit tests
3. Collect-stats provider needs refactoring, it's very complicated and we can make it simple

While the update on current commits is enough for running recipe, I will fix the above items to make this PR better

jctian98:
LGTM

sw005320:
We need a careful document and explanation about what a runner is and what a provider is.
- Please embed this information in the source code
- Please add TODO to explain the runner and provider in our document for parallel processing

Masao-Someki:
Thank you, I added a todo comment, and I will create a PR on document! I have prepared it but couldn't make a PR as we have multiple changes to interfaces, but for now I think it is a good timing to make PR!

espnet3/collect_stats.py
```diff
@@ -1,16 +1,435 @@
...
+    write_collected_feats: bool = False,
+    collect_stats_kwargs: Optional[Dict[str, Any]] = None,
+):
+    """Process a batch of dataset indices and compute feature statistics."""
+
+    structured_items: List[Tuple[str, Any]] = []
+    for i in idxs:
+        item = dataset[i]
+        if isinstance(item, tuple) and len(item) == 2:
+            uid, sample = item
```
Emrys365, line 50: Is `uid` always a string here?

Masao-Someki:
I have added the document for parallel processing and runner/provider classes.
It is still not on PR but you can check form here
- parallel processing: https://github.com/Masao-Someki/espnet/blob/espnet3/documents/doc/espnet3/parallel.md
- runner and provider: https://github.com/Masao-Someki/espnet/blob/espnet3/documents/doc/espnet3/provider_runner.md

espnet3/collect_stats.py
```diff
@@ -1,16 +1,435 @@
...
+            for b_idx, uid in enumerate(uids_in_order):
+                seq = feat_batch[b_idx]
+                if len_batch is not None:
+                    L = int(len_batch[b_idx])
+                    seq = seq[:L]
+                else:
+                    seq = seq[None]
+
+                if not isinstance(seq, np.ndarray):
+                    seq = np.asarray(seq)
```
Emrys365, line 155: Is it possible that the feature `seq` is a torch tensor on a CUDA device? If this is possible, some more handling might be needed.

espnet3/collect_stats.py
```diff
@@ -1,16 +1,435 @@
...
+        params: Optional[Dict[str, Any]] = None,
+    ):
+        cfg = OmegaConf.create({})
+        cfg.model_config = model_config
+        cfg.dataset_config = dataset_config
+        cfg.dataloader_config = dataloader_config
+        cfg.mode = mode
+        cfg.task = task
+        cfg.shard_idx = shard_idx
+        cfg.update(**params)
```
Emrys365, line 253: Might raise an error if `params` is None.

espnet3/collect_stats.py
```diff
@@ -1,16 +1,435 @@
...
+
+    for shard_idx in range(num_shards):
+        sum_dict, sq_dict, count_dict = _collect_stats_common(
+            model_config=model_config,
+            dataset_config=dataset_config,
+            dataloader_config=dataloader_config,
+            mode=mode,
+            output_dir=output_dir,
+            task=task,
+            write_collected_feats=False,
```
Emrys365: Why is write_collected_feats set to `False` here?

tools/setup_anaconda.sh
```diff
@@ -83,7 +83,7 @@ else
     conda install -y conda
 fi
 
-conda install -y pip setuptools
+conda install -y pip==25.2 setuptools
```
Emrys365, line 86: Why using this specific version?

tools/setup_miniforge.sh
```diff
@@ -84,7 +84,7 @@ else
     conda install -y conda
 fi
 
-conda install -y pip setuptools
+conda install -y pip==25.2 setuptools
```
Emrys365, line 87: ditto

tools/setup_python.sh
```diff
@@ -42,4 +42,4 @@ export PATH=${PYTHON_DIR}:\${PATH}
 EOF
 
 . ./activate_python.sh
-python3 -m pip install -U pip wheel
+python3 -m pip install -U pip==25.2 wheel
```
Emrys365, line 45: ditto

tools/setup_venv.sh
```diff
@@ -46,4 +46,4 @@ EOF
 fi
 
 . ./activate_python.sh
-python3 -m pip install -U pip wheel
+python3 -m pip install -U pip==25.2 wheel
```
Emrys365, line 49: ditto

espnet3/collect_stats.py
Masao-Someki: Thank you, `uid` is always string if espnet's common preprocessor is used.
I thought the condition on if sentence is not correct, so I will change this to check if dataset is built for espnet's collate function like the following change:

```
if hasattr(dataset, "use_espnet_preprocessor") \
                and dataset.use_espnet_preprocessor:
```

espnet3/collect_stats.py
Masao-Someki: No, it is a numpy array on CPU. You can check here, the `feats` is generated in these lines and will be passed to the above lines.

https://github.com/Masao-Someki/espnet/blob/29118cbc0584479acea838efbee989ad93ae5287/espnet3/collect_stats.py#L74-L77

espnet3/collect_stats.py
Masao-Someki: Thank you, I will update to `params: Optional[Dict[str, Any]] = dict(),`

espnet3/collect_stats.py
Masao-Someki: Oh, thank you maybe this is from debugging  stage...?
I also noticed that I didn't allowed this flag with multiple iterator (which should be supported). So I fixed this part

tools/setup_anaconda.sh
Masao-Someki: This comes from a CI issue previously we had. However it seems for Jinchuan's speechlm PR, our CI can install espnet with pip>25.3... I am not sure the exact reason right now, but I will removing this. If the same CI issue happens, I'll fix the version as a workaround

Masao-Someki:
@Emrys365 
I'm sorry, I just realized this and fixed the reviews!
For the pip version, I think we can focus on this later.. I had put this in the loadmap.

PR #6271, opened 2025-10-24
[espnet3-9] Add Librispeech-100h ASR recipe

Masao-Someki:
## What did you change?

Added the following recipes

- template
- librispeech_100/asr1

---

To accelerate discussion and as the starter recipe for the espnet-3

---

## Is your PR small enough?

No, because this includes both template and actual recipe.
But I believe separating the PR for template and actual recipe makes it difficult to understand how they are connecting together.

---

## Additional Context

* #6270 (This PR includes the modification on this PR)
* #6133

### DIFF
--- a/espnet3/components/dataloader.py
+++ b/espnet3/components/dataloader.py
@@ -223,8 +223,7 @@ class DataLoaderBuilder:
         if factory_config["iter_factory"] is not None:
             # update shape files
             iter_factory_config = update_shard(
-                factory_config["iter_factory"],
-                shard_idx
+                factory_config["iter_factory"], shard_idx
             )
             return self._build_iter_factory(iter_factory_config, dataset)
         else:


Masao-Someki:
@popcornell 
Could you please review this PR, particularly the template-related parts in decode/asr.py and run.py?

I was considering two possible approaches for handling this:
1. Create a `TEMPLATE/<task>` directory, or
2. Create `decode/<task>.py`, then add the task name as an argument or config option in `decode.yaml`, and switch the decoding script in `run.py` accordingly.

I’d love to hear your thoughts on which direction makes more sense..

egs3/librispeech_100/asr/conf/default.yaml
sw005320, line 1: Do we need to prepare a common default hyperparameter?
This would be a risk that hyperparameters are distributed across multiple locations.

egs3/librispeech_100/asr1/conf/evaluate.yaml
```diff
@@ -0,0 +1,36 @@
+# Decoding configuration for LibriSpeech 100h
+defaults:
+  - default.yaml
+  - ../dataset.yaml
```
sw005320: What's this?

egs3/librispeech_100/asr1/conf/evaluate.yaml
```diff
@@ -0,0 +1,36 @@
...
+  asr_train_config: ${exp_dir}/config.yaml
+  asr_model_file: ${exp_dir}/last.ckpt
+  beam_size: 1
+  ctc_weight: 1.0
+
+
+metrics:
+  - metric:
+      _target_: template.metrics.wer.WER
+      clean_types: whisper_basic
```
sw005320: We should not use whisper filtering for the LibriSpeech WER computation.
This will reach different results from the others.
People do not do any normalization for the normal Librispeech evaluation.

egs3/librispeech_100/asr1/src/dataset.py
```diff
@@ -0,0 +1,94 @@
...
+
+
+class LibriSpeechDataset(TorchDataset):
+    """Minimal dataset for LibriSpeech prepared with ``create_dataset.py``."""
+
+    def __init__(
+        self,
+        data_dir: str | Path = "data",
+        split: str | None = None,
+        sample_rate: int = 16_000,
```
sw005320: what is `16_000`?

sw005320:
I still cannot understand why you keep
`TEMPLATE/template` direcotry.
Is it based on some well-known coding styles?
For me, this sounds redundant (and also meaningless).
But if you have some intention, please let me know.

egs3/librispeech_100/asr/run.py
sw005320, line 1: This looks good, but can you try to further minimize this?

espnet3/systems/base/bin/run.py
sw005320, line 1: What's the difference between run.py here and that in the recipe?
Would you like to prepare this at the template level?

egs3/librispeech_100/asr1/conf/evaluate.yaml
Masao-Someki: This is used to load the default or dataset configuration defined in a separate file. The dataset configuration (dataset.yaml) has already been discussed and will be separated. We can discuss how to handle default.yaml

egs3/librispeech_100/asr1/conf/evaluate.yaml
Masao-Someki: Thank you, I will fix this.

egs3/librispeech_100/asr1/src/dataset.py
Masao-Someki: Thank you, I will fix this to 16000

popcornell:
> @popcornell
> Could you please review this PR, particularly the template-related parts in decode/asr.py and run.py?
> 
> I was considering two possible approaches for handling this:
> 
> Create a TEMPLATE/<task> directory, or
> Create decode/<task>.py, then add the task name as an argument or config option in decode.yaml, and switch the decoding script in run.py accordingly.
> I’d love to hear your thoughts on which direction makes more sense..

I am thinking if it would be better to have a template for each task or not. 
Like TEMPLATE/asr/ so when other users will make a recipe for another task they can do TEMPLATE/diarization/. 

Or we keep template mostly with abstract classes (really an empty template which just describe minimal setup for a recipe) or with just one example for just one task (as you are doing now already) while following recipes on non-asr tasks will have instead to import from another recipe (e.g. i am doing a separation recipe and I will import from wsj0-2mix the inference function).  

These are basically all the options.

sw005320:
> Like TEMPLATE/asr/ so when other users will make a recipe for another task they can do TEMPLATE/diarization/.

This is exactly what we discussed and reached the same conclusion. Good to have your support.

popcornell:
Yeah I think that will be the cleanest. To have templates for all basic tasks.

egs3/librispeech_100/asr/conf/default.yaml
Masao-Someki: After merging the config file, the `default.yaml` now a symlink for the default.yaml in TEMPLATE!
I think this is okay as we have only one actual file?

egs3/librispeech_100/asr/conf/dataset.yaml
```diff
@@ -0,0 +1,34 @@
+# Dataset and dataloader configuration for LibriSpeech 100h
+dataset:
+  _target_: espnet3.data.DataOrganizer
+  train:
+    - name: train-clean-100
+      dataset:
+        _target_: src.dataset.LibriSpeechDataset
```
Emrys365, line 7: Does the current design only allow to load datasets defined under the same recipe directory (i.e., `egs3/librispeech_100/asr1/`**src/dataset.xxx**)?

Emrys365: I think it is helpful to support loading other datasets defined by other recipes, especially for conducting evaluations across domains.

Emrys365: BTW, the current configuration looks a bit redundant to me, as the same line is repeated for each subset. Is it possible to allow omitting this argument when only one `xxxDataset` is defined in `src/dataset.py`?

egs3/librispeech_100/asr/conf/dataset.yaml
```diff
@@ -0,0 +1,34 @@
...
+        split: test-other
+
+  preprocessor:
+    _target_: espnet2.train.preprocessor.CommonPreprocessor
+    train: true
+    token_type: bpe
+    token_list: ${tokenizer.save_path}/tokens.txt
+    bpemodel: ${tokenizer.save_path}/bpe.model
+    text_cleaner:
```
Emrys365, line 34: Is it OK to use an empty value for `text_cleaner`?

egs3/librispeech_100/asr/conf/dataset.yaml
```diff
@@ -0,0 +1,34 @@
...
+        _target_: src.dataset.LibriSpeechDataset
+        split: test-clean
+    - name: test-other
+      dataset:
+        _target_: src.dataset.LibriSpeechDataset
+        split: test-other
+
+  preprocessor:
+    _target_: espnet2.train.preprocessor.CommonPreprocessor
+    train: true
```
Emrys365, line 30: Will this work for the inference stage?

egs3/librispeech_100/asr/conf/evaluate.yaml
```diff
@@ -0,0 +1,43 @@
+# Decoding configuration for LibriSpeech 100h
+defaults:
+  - default.yaml
+  - dataset.yaml
+
+exp_tag: template
```
Emrys365, line 6: Better to change the name to make it more meaningful.

egs3/librispeech_100/asr/src/custom_model.py
```diff
@@ -0,0 +1,41 @@
...
+import torch.nn as nn
+
+from espnet2.bin.s2t_inference import Speech2Text
+
+
+class OWSMFinetuneModel(nn.Module):
+
+    def __init__(self):
+        super().__init__()
+        # About 30 sec
```
Emrys365, line 11: This will depend on the network connection status. Maybe removing this line or changing it to indicate the model size in MB?

egs3/librispeech_100/asr/src/custom_model.py
```diff
@@ -0,0 +1,41 @@
...
+        pretrained_model = Speech2Text.from_pretrained(
+            "espnet/owsm_v4_base_102M", lang_sym=f"<eng>", beam_size=1, device="cuda"
```
Emrys365, line 13: Is it better to allow the user to specify the beam size and device by adding arguments to `__init__`?

egs3/librispeech_100/asr/src/custom_model.py
```diff
@@ -0,0 +1,41 @@
...
+        speech,
+        speech_lengths,
+        text,
+        text_lengths,
+        text_ctc=None,
+        text_ctc_lengths=None,
+        text_prev=None,
+        text_prev_lengths=None,
+    ):
+        # OWSM (s2t_model) の forward にそのまま流す
```
Emrys365, line 29: Maybe translate it to English?

egs3/librispeech_100/asr/src/dataset.py
```diff
@@ -0,0 +1,102 @@
...
+            prepare_s2t=prepare_s2t,
+        )
+
+        data_root = Path(cfg.data_dir)
+        if not data_root.exists():
+            raise FileNotFoundError(
+                f"Dataset directory not found: {data_root.resolve().as_posix()}"
+            )
+        if cfg.split is None:
+            raise ValueError("Please specify the Hugging Face dataset split name.")
```
Emrys365, line 54:
```suggestion
            raise ValueError(
                "Please specify the Hugging Face dataset split name. "
                f"Available: {list(dataset.keys())}"
            )
```

egs3/librispeech_100/asr/src/dataset.py
```diff
@@ -0,0 +1,102 @@
...
+
+    def __len__(self) -> int:
+        return len(self.dataset)
+
+    def _build_example(self, idx: int, item: Dict) -> Dict:
+        example: Dict[str, object] = {}
+        speech_path = item.get("speech", {}).get("path")
+
+        if self.return_speech:
+            array = sf.read(speech_path)[0]
```
Emrys365, line 82: Maybe you should check the consistency of sampling rates (comparing with `self.sample_rate`) here.

egs3/librispeech_100/asr/run.sh
```diff
@@ -0,0 +1,13 @@
+#!/bin/bash
+
+source path.sh
+
+python run.py \
+    --stage all \
+    --train_config conf/tuning/train_e_branchformer.yaml \
+    --eval_config conf/evaluate.yaml \
+    \
```
Emrys365, line 9:
```suggestion
```

egs3/TEMPLATE/asr/conf/default.yaml
```diff
@@ -0,0 +1,85 @@
+num_device: 1
+num_nodes: 1
+task:
```
Emrys365, line 3: How about removing the null parameters?

egs3/librispeech_100/asr/conf/tuning/train_e_branchformer.yaml
```diff
@@ -0,0 +1,115 @@
...
+    account: bbjs-delta-cpu
+    queue: cpu
+    cores: 4
+    processes: 1
+    memory: 16GB
+    death_timeout: 180
+    # walltime: 1:00:00
+    log_directory: logs/
+    job_extra_directives:
+      - "--time=1:00:00"
```
sw005320, line 24: I repeated many times (and I don't remember how many) that the cluster-related config file should be separated, and every time you said that you forgot to do it...

egs3/librispeech_100/asr/conf/tuning/train_e_branchformer.yaml
sw005320, line 1: Can you reduce the config files if they are the same as the default value?

egs3/librispeech_100/asr/conf/dataset.yaml
sw005320: I agree

egs3/librispeech_100/asr/conf/dataset_owsm.yaml
sw005320, line 1: Why is the OWSM dataset included here?

egs3/librispeech_100/asr/conf/default.yaml
sw005320: Why not directly use `../../../TEMPLATE/asr1/conf/default.yaml` in `train_custom_model.yaml‎`
We don't need this file.
Please do not use a symlink of a symlink

egs3/librispeech_100/asr/conf/evaluate.yaml
sw005320, line 1: Inference and scoring must be separated
They are independent, depending on our use case.

egs3/TEMPLATE/asr/conf/dataset.yaml
sw005320, line 1: If this is the only information, I recommend you delete this and put it for each recipe

egs3/librispeech_100/asr/run.sh
```diff
@@ -0,0 +1,13 @@
+#!/bin/bash
+
+source path.sh
+
+python run.py \
```
sw005320, line 5: minimize the argument as much as possible

- `run.py` accepts default arguments, which are `conf/train.yaml` and `conf/evaluate.yaml`. Then, we don't need lines 7 and 8
- make `all` as a default valuer for `--stage` and you don't need line 6
- lines 10--12: ditto

By doing this, we only need 
```
source path.sh && python run.py
```

egs3/librispeech_100/asr/conf/evaluate.yaml
sw005320: I agree.

egs3/librispeech_100/asr/conf/evaluate.yaml
```diff
@@ -0,0 +1,59 @@
+# Decoding configuration for LibriSpeech 100h
+defaults:
+  - default.yaml
+  - dataset.yaml
```
sw005320, line 4: Why do we need them?

egs3/librispeech_100/asr/conf/evaluate.yaml
```diff
@@ -0,0 +1,59 @@
...
+  vocab_size: 5000
+  character_coverage: 1.0
+  model_type: bpe
+  save_path: ${data_dir}/${tokenizer.model_type}_${tokenizer.vocab_size}
+
+
+metrics:
+  - metric:
+      _target_: espnet3.systems.asr.metrics.wer.WER
+      clean_types:
```
sw005320, line 43: What's this?
if we don't need to specify the argument, we don't need to add

egs3/librispeech_100/asr/conf/evaluate.yaml
```diff
@@ -0,0 +1,59 @@
...
+  asr_model_file: ${exp_dir}/last.ckpt
+  beam_size: 1
+  ctc_weight: 0.3
+
+
+tokenizer:
+  vocab_size: 5000
+  character_coverage: 1.0
+  model_type: bpe
+  save_path: ${data_dir}/${tokenizer.model_type}_${tokenizer.vocab_size}
```
sw005320, line 37: why we need a tokenizer?
this should be incorporated in the model

egs3/librispeech_100/asr/conf/tuning/train_custom_model.yaml
```diff
@@ -0,0 +1,32 @@
+# Main training for LibriSpeech 100h ASR recipe
+defaults:
+  - default.yaml
+  - dataset_owsm.yaml
+
+
+exp_tag: template_3
```
sw005320, line 7: We should use the config file name as the default behavior.

egs3/librispeech_100/asr/src/custom_model.py
sw005320, line 1: I cannot understand the role of this file.
Would it be for fine-tuning?
How about explicitly naming the model as `ft.py`?

egs3/librispeech_100/asr/path.sh
```diff
@@ -0,0 +1,6 @@
+#!/bin/bash
+
+
+export PYTHONPATH=../../../:../../TEMPLATE/asr1:$(pwd):${PYTHONPATH}
```
sw005320, line 4: this must be updated.

egs3/librispeech_100/asr/src/dataset.py
sw005320, line 1: What is the role of the dataset.py and create_dataset.py behaviors?
Please specify them and also fix the role

egs3/TEMPLATE/asr/conf/default.yaml
sw005320, line 1: Please review these parameters so that we can minimize the config files in each recipe

egs3/TEMPLATE/asr/conf/evaluate.yaml
sw005320, line 1: We don't need this

egs3/TEMPLATE/asr/conf/train.yaml
sw005320, line 1: This is very confusing.
Please remove `default.yaml` and move its contents here

egs3/librispeech_100/asr/conf/evaluate.yaml
sw005320: - dataset.yaml
- train.yaml
- infer.yaml
- metric.yaml (or measure.yaml if we want to unify the verb)

egs3/TEMPLATE/asr/readme.md
sw005320, line 1: Please revisit it.

egs3/TEMPLATE/asr/run.py
```diff
@@ -0,0 +1,151 @@
...
+
+# Default stage list (can be extended/overridden by callers)
+DEFAULT_STAGES: List[str] = [
+    "create_dataset",
+    "train_tokenizer",
+    "collect_stats",
+    "train",
+    "evaluate",
+    "decode",
+    "score",
```
sw005320, line 22: Please revisit

egs3/TEMPLATE/asr/run.py
```diff
@@ -0,0 +1,151 @@
+# espnet3/utils/run_template.py
```
sw005320, line 1: what's this?

espnet3/systems/asr/metrics/cer.py
sw005320, line 1: Add a test for this by having several situations (e.g., lower case, upper case, mixed cases, handling multiple white spaces in the unicodes

espnet3/systems/asr/metrics/wer.py
sw005320, line 1: Ditto add a test for this by having several situations.

espnet3/systems/asr/task.py
sw005320, line 1: Add a comment that this is a file to use espnet2 task

espnet3/systems/base/inference.py
sw005320, line 1: Did you really refer to the inference API in espnet2 https://github.com/espnet/espnet/blob/master/espnet2/bin/asr_inference.py?

- This looks like an espnet2 ASR specific. This is not what we want. Can you dicuss it with @jctian98?
- There is no definition of the API class that performs simple inference (without reference or without parallelization)
- How to obtain the other decoding information, e.g., nbest, decoding states?

espnet3/systems/base/score.py
sw005320, line 1: We agreed to avoid to use `score`

sw005320: This is very confusing.
You use `metric` partly and still keep `score
This kind of ambiguity will introduce confusion.

espnet3/systems/base/scp_utils.py
sw005320, line 1: utility files should be under utilities.
not under the base

espnet3/systems/base/system.py
```diff
@@ -0,0 +1,66 @@
...
+
+
+class BaseSystem:
+    """Base class for all ESPnet3 systems.
+
+    Each system should implement the following:
+      - create_dataset()
+      - train()
+      - decode()
+      - score()
```
sw005320, line 22: rename them or use the words consistently

espnet3/systems/base/train.py
sw005320, line 1: This is lightning-based trainer
This name will be conflicted with the others in the future

test/espnet3/test_callback.py
sw005320, line 1: I recommend you test the directory to follow the same structures as the espnet3 directory as much as possible.

sw005320 (commented):
- Please revisit the default config structures and simplify them.
  - Remove `default.yaml` and all default parameters should be moved to `train.yaml`, `inference.yaml`, and `metric.yaml`
- Please reconsider the default behaviors for the experimental and data directory creations. You can follow espnet2 styles (but a Pythonic way)
- Should we prepare some directory structures for `components` in advance? I'm expecting this would become very messy

Masao-Someki:
Updated documents here: https://masao-someki.github.io/espnet_draft_home_page/

I hope the documents will help reviewers to understand the espnet3 package more detail.

I will work on fixing this PR based on the review and discussion with @popcornell.

sw005320:
Do you mean this?
https://masao-someki.github.io/espnet_draft_home_page/espnet3/recipe_directory.html
There is an error `&lt;system&gt;`

egs3/librispeech_100/asr/conf/dataset.yaml
Masao-Someki: Thank you, @Emrys365, and sorry for the late reply.

1. **Does the current design only allow loading datasets defined under the same recipe directory?**
   → No. As long as the dataset is importable, it can be used.

2. **Is it possible to omit this argument when only one `xxxDataset` is defined in `src/dataset.py`?**
   → Yes. We can omit the class definition if there is only a single dataset class available. However, since we need to support different data splits for training, validation, and testing, we could structure the configuration as follows:

```yaml
# Dataset configuration for LibriSpeech 100h
dataset:
  train:
    - name: train-clean-100
      dataset:
        split: train-clean-100
  valid:
    - name: dev-clean
      dataset:
        split: dev-clean
    - name: dev-other
      dataset:
        split: dev-other
  test:
    - name: test-clean
      dataset:
        split: test-clean
    - name: test-other
      dataset:
        split: test-other
```

egs3/librispeech_100/asr/conf/dataset.yaml
Masao-Someki: I think we can override the value to false in the `evaluate,yaml`. I will update the config file.

```
dataset:
  preprocessor:
    train: false
```

egs3/librispeech_100/asr/conf/dataset.yaml
Masao-Someki: Yes!
`text_cleaner:` with no value is parsed as `null`, and `OmegaConf.load` returns `None`. In `CommonPreprocessor`, `TextCleaner(text_cleaner)` receives `None`, which becomes an empty cleaner list, so no normalization is applied.

egs3/librispeech_100/asr/conf/evaluate.yaml
Masao-Someki: Thank you, I will change it to `librispeech100_asr_branchformer`

PR #6275, opened 2025-10-26
CI Installation fix

Fhrozen:
## What did you change?

This pull request updates installation scripts for several dependencies to use the `--no-build-isolation` flag when installing Python packages in editable mode. This change helps ensure that package installations use the current environment's dependencies, which can improve compatibility and reproducibility, especially when custom or pre-existing packages are required.

**Dependency installation improvements:**

* Updated the `pip install` command in `tools/installers/install_muskits.sh` for `ParallelWaveGAN` to use the `--no-build-isolation` flag.
* Updated the `pip install` command in `tools/installers/install_parallel-wavegan.sh` for `ParallelWaveGAN` to use the `--no-build-isolation` flag.
* Updated the `pip install` command in `tools/installers/install_warp-transducer.sh` for the `warp-transducer` PyTorch binding to use the `--no-build-isolation` flag.

# References:

Possible related to https://github.com/pypa/pip/issues/13627

Could not find any related installation fix that generated this issue.

### DIFF
--- a/tools/installers/install_muskits.sh
+++ b/tools/installers/install_muskits.sh
@@ -15,7 +15,7 @@ if [ ! -e muskits.done ]; then
                 rm -rf ParallelWaveGAN
                 git clone https://github.com/kan-bayashi/ParallelWaveGAN.git
                 cd ParallelWaveGAN
-                pip install -e .
+                pip install --no-build-isolation -e .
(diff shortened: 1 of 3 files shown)


sw005320:
Thanks!

PR #6260, opened 2025-10-07
SpeechLM Data Infra: Data batchfy, sampling and iterator

jctian98:
The data iterator block of the SpeechLM repo.

This PR should not be reviewed or merged before https://github.com/espnet/espnet/pull/6257

(1) espnet2/speechlm/configuration: general-purpose and language model-specific task template design
(2) espnet2/speechlm/dataloader/iterator.py: the iterator factory logic to build the dataloader object, with good state trace and deterministic behavior.
(3) espnet2/speechlm/dataloader/batch.py the method to batchify the data.
(4) Plus some minor fix.

Also see: https://github.com/espnet/espnet/pull/6258

(1) SpeechLM preprocessor;
(2) SpeechLM model files
(3) A speechLM-specific trainer
(4) A PR to connect all components
(5) Inference

PR #6280, opened 2025-10-27
[SpeechLM] Minor fix on data loading

jctian98:
This PR makes several minor improvements over previous PR #6257 #6260 

(1) using multiprocessing to build each single dataset for efficiency (dataset.py)
(2) Discard the very long examples, instead of raising an error
(3) Remove the previous `configuration` folder and move its "task_conf.py" to the dataloader folder.

Other changes are minor.

Request review: @Masao-Someki @wanchichen @siddhu001

### DIFF
--- a/espnet2/speechlm/bin/prepare_audio_lhotse.py
+++ b/espnet2/speechlm/bin/prepare_audio_lhotse.py
@@ -1,17 +1,8 @@
 #!/usr/bin/env python3
-"""Prepare audio metadata from Kaldi wav.scp and segments files using Lhotse.
+# Copyright 2025 Jinchuan Tian (Carnegie Mellon University)
+#  Apache 2.0  (http://www.apache.org/licenses/LICENSE-2.0)
 
-This script processes Kaldi format wav.scp and optional segments files to extract
-audio metadata (sampling rate, duration, etc.) and save it using Lhotse manifests.
-Supports multiprocessing for efficient processing of large datasets.
-
-Multi-channel Support:
-    - Single-channel audio: Creates MonoCut objects for efficiency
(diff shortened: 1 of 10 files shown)


jctian98:
I have reflected on the Gemini's review. 

I ensure the num_worker is at least 1 to avoid the multi-processing issue. But I think we may not need to special implementation when num_worker == 1 for simplicity.

The PR is ready to review.

espnet2/speechlm/utils/data.py
```diff
@@ -0,0 +1,124 @@
...
+            dtype=dtype, device=device, non_blocking=True
+        )
+
+    # Create length tensor
+    length_tensor = torch.tensor(lengths, dtype=torch.long, device=device)
+
+    return padded, length_tensor
+
+
+def to_device(data, device=None, dtype=None, non_blocking=False, copy=False):
```
Masao-Someki, line 92: Is this the same function as the existing to_device?

https://espnet.github.io/espnet/guide/espnet2/torch_utils/to_device.html

espnet2/speechlm/dataloader/iterator.py
Masao-Someki, line 1: As mentioned in another PR, please consider inheriting from `AbsIterFactory`.

jctian98: The code intentionally avoids any dependency on the espnet2 code. So maybe we can avoid this. 
Thank you!

espnet2/speechlm/utils/data.py
jctian98: yes it's the same.

siddhu001:
The PR looks ok to me.

PR #6284, opened 2025-10-28
codec fix: DDP logic and dead code revival logic

whr-a:
## What did you change?
* Corrected the "dead code" revival logic within the `EuclideanCodebook`'s EMA update.
* Added DDP (Distributed Data Parallel) synchronization (`all_reduce`) for all codebook updates, including K-Means init, EMA statistics (`cluster_size`, `embed_sum`), and dead code sampling.

1.  **To fix a EMA bug:** The previous logic for reviving dead code was non-functional. When a dead code (`self.embed`) was replaced with a new vector, its corresponding EMA state (`self.embed_avg`, `self.cluster_size`) was **not** reset. This caused the new vector to be immediately overwritten by a stale value (calculated from the old, dead state) in the *same* forward pass, preventing the codebook from recovering from collapse.
2.  **To fix DDP training:** In the previous module, EMA updates (for `cluster_size` and `embed_sum`) were calculated using only the `local` batch on each worker, also with no step-by-step synchronization. Synchronization only occurred intermittently during initialization (`init_embed_`) and code expiration (`expire_codes_`). `broadcast_tensors` was used to copy the buffers **from Rank 0** to all other workers, discarding the updates computed by other ranks. This PR changes the logic to use `all_reduce` on these statistics on every training step, ensuring the EMA update is calculated using the full **global batch** and keeping codebooks consistent across all workers.

## Is your PR small enough?
yes

## Additional Context
The logic in https://github.com/cisco-open/espnet/blob/master/espnet2/gan_codec/shared/quantizer/modules/core_vq.py is fully referenced.
Also compared to https://github.com/XiaomiMiMo/MiMo-Audio/blob/main/src/mimo_audio_tokenizer/quantization.py
The code has been validated for correctness through training.

### DIFF
--- a/espnet2/gan_codec/shared/quantizer/modules/core_vq.py
+++ b/espnet2/gan_codec/shared/quantizer/modules/core_vq.py
@@ -41,8 +41,6 @@ import torch.nn.functional as F
 from einops import rearrange, reduce, repeat
 from torch import einsum, nn
 
-from espnet2.gan_codec.shared.quantizer.modules.distrib import broadcast_tensors
-
 
 def default(val: Any, d: Any) -> Any:
     return val if val is not None else d


espnet2/gan_codec/shared/quantizer/modules/core_vq.py
```diff
@@ -232,24 +359,20 @@ def forward(self, x):
...
             ema_inplace(self.embed_avg, embed_sum.t(), self.decay)
-            cluster_size = (
-                laplace_smoothing(self.cluster_size, self.codebook_size, self.epsilon)
-                * self.cluster_size.sum()
-            )
-            embed_normalized = self.embed_avg / cluster_size.unsqueeze(1)
-            self.embed.data.copy_(embed_normalized)
+            self.update_ema()
+
+            self.expire_codes_(x)
```
ftshijt, line 375: If expire codes are in later stage, shall we do another all_reduce to keep all gpus the same?

whr-a: I think it's OK,
1. `expire_codes_` determines dead codes by checking `self.cluster_size < self.threshold_ema_dead_code`, and `self.cluster_size` has just been synchronized across all GPUs, so the calculated mask is identical.
2. The replacement vectors are determined by `self.sample_fn`, which in the DDP environment is `sample_vectors_distributed`. This function is designed for distributed sampling and ensures every GPU gets the exact same new vectors. Though each GPU will handle the dead codes independently, the resulting codebook will still remain synchronized."

ftshijt: Thanks for sharing the update. It looks good to me

ftshijt:
Thanks for your great contribution!

PR #6287, opened 2025-11-04
Add arkive data loading

wanchichen:
## What did you change?

This PR adds support for loading data using [arkive](https://github.com/wanchichen/arkive), designed for large-scale training.
Arkive stores metadata in `.parquets` and audio dumps in `.bin` files, avoiding the need to store many metadata and individual audio files.

I use DuckDB for loading the metadata parquet, making it extremely fast. For librispeech 960 on my system, it takes around 0.30 seconds to load a full metadata parquet (with sample rate, channel, duration info) and 1.15 seconds to load an equivalent kaldi `wav.scp`.

The `ArkiveAudioReader` allows ESPnet to read audio data using `arkive`. It requires an SQL query as input like:
```
name = '/work/nvme/bbjs/chen26/arkive/librispeech_test_processed.parquet'
query = f"SELECT utt_id, path, start_byte_offset, file_size_bytes, start_time, end_time FROM '{name}'"
reader = ArkiveAudioReader(query)
```

For distributed processing, one can also pass the world size and num_workers to filter the data and save memory.
```
ArkiveAudioReader(query, valid_ids=None, worker_id=0, world_size=64)
```

`espnet2/speechlm/bin/prepare_audio_arkive.py` can be used to convert kaldi audio files to the arkive format.

---

For now, i dont add a script to auto install `arkive` and `duckdb` since its only used in `speechlm`, and i think the install script will change with espnet3.

Kaldi format requires us to store too many extra files and occupies a lot of disk space. Lhotse is nice, but since it works on individual audio files, the disk i/o impact is quite severe. Arkive is designed to be more scalable and more adaptable to new modalities.

### DIFF
--- a/egs2/kinyarwanda_tts/tts1/Readme.md
+++ b/egs2/kinyarwanda_tts/tts1/Readme.md
@@ -1,0 +1,48 @@
+# KINYARWANDA TTS RECIPE
+
+This is the first ESPnet2 recipe for Kinyarwanda TTS. It implements a Tacotron 2 architecture with character-based tokenization, utilizing transfer learning from an English LJSpeech pre-trained model to support low-resource Bantu language synthesis.
+
+See the following pages for the usage:
+- [How to run the recipe](../../TEMPLATE/tts1/README.md#how-to-run)
+- [How to train with speaker ID](../../TEMPLATE/tts1/README.md#multi-speaker-model-with-speaker-id-embedding-training)
+- [How to train VITS](../../TEMPLATE/tts1/README.md#vits-training)
+
+See the following pages before asking the question:
+- [ESPnet2 Tutorial](https://espnet.github.io/espnet/espnet2_tutorial.html)
+- [ESPnet2 TTS FAQ](../../TEMPLATE/tts1/README.md#faq)
+
+# INITIAL RESULTS
+
+**Model:** Tacotron 2 (Transfer Learning from LJSpeech)
+**Vocoder:** Parallel WaveGAN (LJSpeech pre-trained)
+
+## Environments
+
+* **date:** `Tue Feb 17 20:26:00 UTC 2026`
+* **python version:** `3.12.0`
+* **espnet version:** `espnet 202412`
+* **pytorch version:** `pytorch 2.5.1`
+
+### UTMOSv2 (Strong)
+
+**Average:** 2.2103 (Evaluated on 400 test samples)
+
+### CER (with whisper-large-v3-kin-track-b)
+
+Evaluation performed using the Kaggle-winning ASR model for Kinyarwanda Track B to ensure a high-rigidity benchmark.
+
+| #Snt | #Wrd | Corr | Sub | Del | Ins | Err | S.Err |
+| :--- | :--- | :--- | :--- | :--- | :--- | :--- | :--- |
+| 400 | 34270 | 86.7 | 6.1 | 7.1 | 5.8 | 19.0 | 98.2 |
+
+### WER (with whisper-large-v3-kin-track-b)
+
+| #Snt | #Wrd | Corr | Sub | Del | Ins | Err | S.Err |
+| :--- | :--- | :--- | :--- | :--- | :--- | :--- | :--- |
+| 400 | 4768 | 58.0 | 33.4 | 8.6 | 6.3 | 48.3 | 98.2 |
+
+## Pretrained Models
+
+### espnet/egs2/kinyarwanda_tts/tts1/exp/tts_kinyarwanda_transfer_v1
+
+* [https://huggingface.co/Professor/kinyarwanda-tacotron2-espnet](https://huggingface.co/Professor/kinyarwanda-tacotron2-espnet)
--- a/egs2/kinyarwanda_tts/tts1/cmd.sh
+++ b/egs2/kinyarwanda_tts/tts1/cmd.sh
@@ -1,0 +1,110 @@
+# ====== About run.pl, queue.pl, slurm.pl, and ssh.pl ======
+# Usage: <cmd>.pl [options] JOB=1:<nj> <log> <command...>
+# e.g.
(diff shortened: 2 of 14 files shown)


espnet2/speechlm/bin/prepare_audio_arkive.py
```diff
@@ -0,0 +1,219 @@
...
+    logging.info(
+        f"Successfully processed {len(ark.data)}/{len(audio_paths)} recordings"
+    )
+
+    df = ark.data
+    cols = df.columns.tolist()
+    cols.remove("original_file_path")
+    cols_to_add = [df[col] for col in cols]
+    data = dict(zip(df["original_file_path"], zip(*cols_to_add)))
+
```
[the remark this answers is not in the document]
wanchichen: this is wrong lol, much slower like this

espnet2/speechlm/dataloader/multimodal_loader/audio_loader.py
```diff
@@ -17,6 +31,101 @@
...
+        result = duckdb.query(query)
+
+        # filter query result before loading to df
+        # avoids loading the whole query result into memory
+        if worker_id is not None:
+            assert (
+                world_size is not None
+            ), f"filtering by worker_id requires world_size, got {world_size}"
+            result = duckdb.query(
+                f"""
```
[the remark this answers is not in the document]
wanchichen: same as above

Qingzheng-Wang:
Please add `ArkiveAudioReader` here: https://github.com/espnet/espnet/blob/d93495baf1d5419df541801f2083bfd52844724c/espnet2/speechlm/dataloader/dataset.py#L21

espnet2/speechlm/dataloader/multimodal_loader/audio_loader.py
```diff
@@ -17,6 +31,100 @@
...
+
+    def __init__(
+        self,
+        query: str,
+        valid_ids: list = None,
+        worker_id: int = None,
+        world_size: int = None,
+    ):
+
+        result = duckdb.query(query)
```
Qingzheng-Wang: I think it would be better to keep the interface consistenti with the `LhotseAudioReader`:

```suggestion
class ArkiveAudioReader:
    """Dict-like lazy audio reader using arkive parquets

    Args:
        manifest_dir: Directory containing the arkive manifest files
            e.g. /path/to/dataset/audio1
        valid_ids: List of valid IDs to keep (optional, keeps all if None)
        worker_id: partition ids by worker (optional, keeps all if None)
        world_size: used for worker partitioning
        query: (optional) SQL query for advanced usage
            columns: [utt_id, path, start_byte_offset, file_size_bytes, start_time, end_time]
    """

    def __init__(
        self,
        manifest_dir: str,
        valid_ids: list = None,
        worker_id: int = None,
        world_size: int = None,
        query: Optional[str] = None,
    ):

        if query is None:
            query = f"SELECT * FROM '{manifest_dir}/metadata.parquet'"

        result = duckdb.query(query)
```

espnet2/speechlm/dataloader/multimodal_loader/audio_loader.py
```diff
@@ -17,6 +31,100 @@
...
+            )
+        )
+
+        if valid_ids:
+            data = {k: data[k] for k in set(valid_ids) if k in data}
+
+        self.data = data
+
+    def __getitem__(self, key: str) -> Tuple[np.ndarray, int]:
+        path, start_byte, file_size, start_time, end_time = self.data[key]
```
Qingzheng-Wang: `start_time` and `end_time` might be \<NA\> in pandas notation if not provide segments, so here we should convert them to None so that the subsequent code can correctly recognize and handle them:
```suggestion
path, start_byte, file_size, start_time, end_time = self.data[key]
        
        # Convert pandas NA to Python None
        if pd.isna(start_time):
            start_time = None
        if pd.isna(end_time):
            end_time = None
            
```

espnet2/speechlm/dataloader/multimodal_loader/audio_loader.py
```diff
@@ -9,6 +9,20 @@
 
 import numpy as np
```
Qingzheng-Wang, line 10:
```suggestion
import numpy as np
import pandas as pd
```

espnet2/speechlm/dataloader/multimodal_loader/audio_loader.py
wanchichen: I'm not 100 % sure if we should do this, since the query should be responsible for aggregating everything across multiple parquets and might require some complex logic.

For example, it may do some complex filtering and concat like 

```
SELECT * FROM ['file1.parquet', 'path/to/file2.parquet', 'another_folder/file3.parquet'] WHERE utt_id like 'libri%' as table1 
UNION ALL
SELECT doc_id as utt_id, file_path as path, start as start_byte_offset, file_size_bytes, start_time, end_time FROM file2.parquet WHERE file_size_byte > 100000
```

The suggested solution will be fine to allow for both interfaces (for people who dont want to write SQL) but im worried about the possibility of silent errors (like test-set corruption if someone has both train/test in the same parquet).

wanchichen: what do you think @jctian98 ?

sw005320:
@wanchichen, there are several CI errors.
Please handle them.

sw005320:
Fix this https://github.com/espnet/espnet/actions/runs/19321213055/job/55262933331?pr=6287#step:8:13

jctian98:
Already had some discussion with @wanchichen, will try to merge after some features are updated.
Please also fix CI.

wanchichen:
I improved the data loading to use polars instead of pandas, which is significantly more efficient. In the future, we should eventually move to integer index based loading once everything is unified, instead of utt id. Maintaining an utterance id to index mapping currently invokes a 30% memory overhead (although still more efficient than the alternatives...)

PR #6298, opened 2025-11-13
[SVS]: update recipe name

South-Twilight:
## What did you change?

Update amaboshi to ameboshi.

Fix recipe name.

## Is your PR small enough?

yes, just change directory name.


sw005320:
Thanks for the fix, @South-Twilight!

PR #6327, opened 2025-12-24
[espnet3-11] Directory update to espnet-3

Masao-Someki:
## What did you change?

* Reorganized the structure under `espnet3` and moved files into `components/`, `systems/`, `parallel/`, and `utils/`.
* Added and removed `__init__.py` files accordingly, and moved tests to match the new paths.
* Made minor updates to several files and to `test_utils/espnet3/config/model_ctc.yaml` to reflect the updated import paths.
* Update unit tests to shorten the module path inside. It is stored in string, but it cause the format issue (longer than 88 chars)

---

* To organize the directory structure to make maintenance easier.

---

## Is your PR small enough?

* No (46 files changed, 229 lines changed)
* But is adding/removing __init__.py and moving files, and fixed import sentence accordingly.

---

## Additional Context

* #6133 
* #6271

### DIFF
--- a/doc/espnet3/parallel.md
+++ b/doc/espnet3/parallel.md
@@ -99,11 +99,11 @@ ESPnet3 supports two main options for model definition:
    - Simply copy the model config and specify the ESPnet3 task to retain compatibility
 
    ```python
-   from espnet3.trainer import LitESPnetModel
+   from espnet3.components.modeling.lightning_module import ESPnetLightningModule
    from espnet3.task import get_espnet_model
 
    model = get_espnet_model(task="asr", config=config.model)
-   model = LitESPnetModel(model)
+   model = ESPnetLightningModule(model)
    trainer = Trainer(model, ...)
    ```
 
@@ -112,10 +112,10 @@ ESPnet3 supports two main options for model definition:
    - Hydra will instantiate it at runtime
 
    ```python
-   from espnet3.trainer import LitESPnetModel
+   from espnet3.components.modeling.lightning_module import ESPnetLightningModule
 
    model = hydra.utils.instantiate(config.model)
-   model = LitESPnetModel(model)
+   model = ESPnetLightningModule(model)
    trainer = Trainer(model, ...)
    ```
 


espnet3/components/data/collect_stats.py
```diff
@@ -14,10 +14,10 @@
 from espnet2.fileio.datadir_writer import DatadirWriter
 from espnet2.fileio.npy_scp import NpyScpWriter
 from espnet2.train.collate_fn import CommonCollateFn
+from espnet3.parallel.base_runner import BaseRunner
+from espnet3.parallel.env_provider import EnvironmentProvider
 from espnet3.parallel.parallel import set_parallel
-from espnet3.runner.base_runner import BaseRunner
-from espnet3.runner.env_provider import EnvironmentProvider
-from espnet3.task import get_espnet_model
+from espnet3.utils.task import get_espnet_model
```
Emrys365: I cannot quite understand why `task.py` is moved into the `utils` directory, which does not sound like a major component of the framework. Could you explain the motivation a bit?

espnet3/utils/task_utils.py
Emrys365, line 1: ditto

test/espnet3/components/callbacks/test_callback.py
```diff
@@ -7,7 +7,10 @@
 from lightning.pytorch.callbacks import LearningRateMonitor, ModelCheckpoint
 from omegaconf import OmegaConf
 
-from espnet3.trainer.callbacks import AverageCheckpointsCallback, get_default_callbacks
+from espnet3.components.callbacks.callbacks import (
```
Emrys365: This looks a bit weird to import `callbacks.callbacks`. How about renaming `callbacks.py` to something like `default_callbacks.py` if the directory will be extended to contain more functional callbacks?

espnet3/components/modeling/lightning_module.py
Emrys365, line 1: `model.py` sounds too broad in my view. The function of this script is similar to [`espnet2/train/trainer.py`](https://github.com/espnet/espnet/blob/master/espnet2/train/trainer.py) in ESPnet2 I think. Can this be merged with [`espnet3/components/training/trainer.py`](https://github.com/Masao-Someki/espnet/blob/011ea0ca3d5bd93c0f049745350d9c2b938360c9/espnet3/components/training/trainer.py)?

espnet3/components/data/collect_stats.py
Masao-Someki: Thank you, I think I need to change the file name into `task_utils.py`!
This file is providing several utility functions to connect espnet3 configurations and espnet2 tasks.

test/espnet3/components/callbacks/test_callback.py
Masao-Someki: Thank you, this is a very good point. I will change it to `default_callbacks.py`. We can support more optional callbacks as we develop the espnet3!

espnet3/components/modeling/lightning_module.py
Masao-Someki: Thank you!

I think `model.py` looked similar to `trainer.py` because it included setup functions for optimizers/schedulers, dataloaders, and so on. Since I was following the common practice in PyTorch Lightning to include these components in the `LightningModule`, I think it makes sense to keep this structure.

However, I agree that `model.py` was not a good name, so I have renamed it to `lightning_module.py`. Does this make the responsibility clearer to you?

Masao-Someki:
@sw005320 could you please take a look at this PR? If it looks good to you, we can merge it into `master` and move on to the next PR!

sw005320:
LGTM!
Thanks!

PR #6328, opened 2025-12-24
[espnet3-12] Add ASR system and inference packages

Masao-Someki:
## What did you change?

- Added `espnet3/systems/asr` package
- Added support on inference/measurement part.

---

To support ASR recipe in the next PR.

---

## Is your PR small enough?

Sorry, but no, it has 31 files with 2k line edits.
However, the largest file is the `systems/asr/task.py`, which is a direct copy from espnet2.
Also the larger file `espnet3/systems/asr/tokenizer/sentencepiece.py` was a copy from espnet-ez package.

Most of other espnet-3 files are less than 100 lines, which should make the review easier.

---

## Additional Context

This PR should be merged after #6327.

* #6133
* #6271
* #6327

### DIFF
--- a/espnet2/ps2st/espnet_model.py
+++ b/espnet2/ps2st/espnet_model.py
@@ -16,11 +16,17 @@ try:
         AutoProcessor,
         Qwen2AudioForConditionalGeneration,
     )
-    from transformers.modeling_utils import no_init_weights
 
     is_transformers_available = True
 except ImportError:
     is_transformers_available = False
+
+if is_transformers_available:
+    try:
+        # transformers < v5
+        from transformers.modeling_utils import no_init_weights
+    except ImportError:
+        from transformers.initialization import no_init_weights
 
 
 class ESPnetQwen2AudioModel(AbsESPnetModel):
--- a/espnet2/ps2st/qwen2_scorer.py
+++ b/espnet2/ps2st/qwen2_scorer.py
@@ -49,8 +49,9 @@ class Qwen2HFScorer(ScorerInterface):
             logits = out.logits[:, -1]
             past_kv = out.past_key_values
 
+            # Support both transformers V4 and V5
             past_len = (
-                past_kv[0][0].size(-2)
+                past_kv.layers[0].get_seq_length()
                 if past_kv is not None
                 else self.prefill_input_ids.size(1)
             )
--- a/espnet2/text/whisper_token_id_converter.py
+++ b/espnet2/text/whisper_token_id_converter.py
@@ -64,12 +64,20 @@ class OpenAIWhisperTokenIDConverter:
         self.tokenizer = copy.deepcopy(self.tokenizer)
         timestamps = [f"<|{i * 30 / 1500:.2f}|>" for i in range(0, 1501)]
         sc = [speaker_change_symbol] if sot else []
-        special_tokens = (
-            self.tokenizer.tokenizer.additional_special_tokens + timestamps + sc
-        )
-        self.tokenizer.tokenizer.add_special_tokens(
-            dict(additional_special_tokens=special_tokens)
-        )
+        # workaround for transformers v5
+        if hasattr(self.tokenizer.tokenizer, "additional_special_tokens"):
+            # For transformer < V5
+            special_tokens = (
+                self.tokenizer.tokenizer.additional_special_tokens + timestamps + sc
+            )
+            self.tokenizer.tokenizer.add_special_tokens(
+                dict(additional_special_tokens=special_tokens)
+            )
+        else:
+            # For transformer >= V5
+            self.tokenizer.tokenizer.add_special_tokens(
+                dict(extra_special_tokens=timestamps + sc)
+            )
         self.model_type = model_type
 
     def get_num_vocabulary_size(self) -> int:
--- a/espnet2/text/whisper_tokenizer.py
+++ b/espnet2/text/whisper_tokenizer.py
@@ -86,12 +86,18 @@ class OpenAIWhisperTokenizer(AbsTokenizer):
         # Whisper uses discrete tokens (20ms) to encode timestamp
         timestamps = [f"<|{i * 0.02:.2f}|>" for i in range(0, 1501)]
         sc = [speaker_change_symbol] if sot else []
-        special_tokens = (
-            self.tokenizer.tokenizer.additional_special_tokens + timestamps + sc
-        )
-        self.tokenizer.tokenizer.add_special_tokens(
-            dict(additional_special_tokens=special_tokens)
-        )
+
+        # workaround for transformers v5
+        if hasattr(self.tokenizer.tokenizer, "additional_special_tokens"):
+            # For transformer < V5
+            self.tokenizer.tokenizer.add_special_tokens(
+                dict(additional_special_tokens=timestamps + sc)
+            )
+        else:
+            # For transformer >= V5
+            self.tokenizer.tokenizer.add_special_tokens(
+                dict(extra_special_tokens=timestamps + sc)
+            )
 
     def __repr__(self):
         return (


Masao-Someki:
Hi @popcornell @wanchichen, if you have time, it would be great if you could take a look at the espnet3 system implementation (`espnet3/systems/asr` and `espnet3/systems/base`) and share your feedback!

Masao-Someki:
@Emrys365 Could you please review this PR? It introduces the `espnet3/systems/` directory and adds base systems under `espnet3/systems/base` along with corresponding unit tests, which makes the diff large..

If this feels too heavy to review in one go, please let me know, I’m happy to split it into multiple PRs.

espnet3/components/metrics/abs_metric.py
sw005320, line 1: I expect this to include many functions (e.g., import `versa` to support various generation metrics).
Would it be sufficiently general to handle such extensions?

espnet3/systems/asr/metrics/__init__.py
```diff
@@ -1 +1 @@
-"""ASR metric components for ESPnet3."""
+"""ASR metric implementations for decoding and evaluation."""
```
sw005320: what is decoding?

espnet3/systems/asr/inference.py
```diff
@@ -0,0 +1,110 @@
...
+            Dict with index, hypothesis string, and reference string.
+
+        Raises:
+            AssertionError: If required fields are missing from dataset items.
+        """
+        data = dataset[idx]
+        assert "speech" in data, "ASR inference requires 'speech' in dataset item."
+        assert "text" in data, "ASR inference requires 'text' in dataset item."
+        speech = data["speech"]
+        hyp = model(speech)[0][0]
```
sw005320: Is it correct?
Is it via beam search?

espnet3/systems/asr/metrics/cer.py
sw005320, line 1: Does this support the case normalization and punctuation removals?

espnet3/systems/asr/metrics/wer.py
sw005320, line 1: Ditto
Does this support the case normalization and punctuation removals?

Or will you assume that this is correctly done in the data preparation?
If so, I'd recommend you add such comments

espnet3/systems/base/inference.py
sw005320, line 1: Does this support batch inference?

espnet3/systems/base/measure.py
sw005320, line 1: Why do you mix up "score" and "metric"?
I asked many times not to do so.
This is very confusing.

espnet3/systems/base/system.py
```diff
@@ -0,0 +1,131 @@
...
+
+
+class BaseSystem:
+    """Base class for all ESPnet3 systems.
+
+    Each system should implement the following:
+      - create_dataset()
+      - train()
+      - decode()
+      - score()
```
sw005320: ditto

Can you use search to check all parts that say "score" and consistently change it?

espnet3/systems/base/system.py
```diff
@@ -0,0 +1,131 @@
...
+logger = logging.getLogger(__name__)
+
+
+class BaseSystem:
+    """Base class for all ESPnet3 systems.
+
+    Each system should implement the following:
+      - create_dataset()
+      - train()
+      - decode()
```
sw005320: This must be an inference.
Please do not use inconsistent naming at all.

espnet3/systems/base/system.py
```diff
@@ -0,0 +1,131 @@
...
+            if isinstance(model_cfg, DictConfig):
+                model_target = model_cfg.get("_target_")
+        logger.info(
+            "Training start | exp_dir=%s model=%s",
+            getattr(self.train_config, "exp_dir", None),
+            model_target or "<unknown>",
+        )
+        return train(self.train_config)
+
+    def evaluate(self, *args, **kwargs):
```
sw005320: Do you need this?

We can just use `infer()` and `measure` in a cascade.
Why do you need to combine?

test/espnet3/systems/base/test_metric.py
sw005320, line 1: ditto
score? measure? metric?

sw005320 (changes requested):
Please make sure to use consistent notations.
- score/metric/measure
- decode/inference

sw005320:
Please fix the CI errors

espnet3/components/data/dataset.py
```diff
@@ -41,7 +40,7 @@ class CombinedDataset:
     Raises:
         IndexError: If a requested index is outside the range of the combined dataset.
         ValueError: If index is a non-integer string or cannot be cast to int.
-        RuntimeError: If `get_text()` or `shard()` is called but not supported.
```
Emrys365, line 44: Can you explain why `get_text()` related comments and code are removed?

espnet3/components/metrics/abs_metric.py
```diff
@@ -0,0 +1,28 @@
...
+    @abstractmethod
+    def __call__(
+        self, data: Dict[str, List[str]], test_name: str, decode_dir: Path
+    ) -> Dict[str, float]:
+        """Compute metrics for a decoded test set.
+
+        Args:
+            data (Dict[str, List[str]]): Aligned fields from decode outputs.
+            test_name (str): Name of the test dataset (e.g., "test-other").
+            decode_dir (Path): Root path where decode results are stored.
```
Emrys365: Better to elaborate on the use cases of these arguments. It is not quite clear to me what `data` should contain and whether `test_name` can be randomly specified.


In addition, it might be better to rename `decode_dir` to `output_dir`, because the `AbsMetrics` is general and might be used for other tasks without having a decoding procedure.

espnet3/components/modeling/lightning_module.py
```diff
@@ -405,15 +405,24 @@ def collect_stats(self):
         Raises:
             AssertionError: If `config.statsdir` is not provided.
         """
-        assert hasattr(self.config, "statsdir"), "config.statsdir must be defined"
+        assert hasattr(self.config, "stats_dir"), "config.statsdir must be defined"
```
Emrys365: Should `config.statsdir` be `config.stats_dir` instead?

espnet3/parallel/inference_provider.py
```diff
@@ -0,0 +1,85 @@
...
+            params: Optional extra entries merged into the environment.
+        """
+        super().__init__(config)
+        self.params = params or {}
+        # Build once for local execution to avoid redundant IO
+        self._local_env = {
+            "dataset": self.build_dataset(config),
+            "model": self.build_model(config),
+            **self.params,
+        }
```
Emrys365: How about using `self._loval_env = self.make_worker_setup_fn()()`?

espnet3/systems/asr/inference.py
Emrys365: Does this method support parallel computing (e.g., multiprocessing)?

espnet3/systems/asr/system.py
```diff
@@ -0,0 +1,172 @@
...
+        logger.info("Collected %d transcript lines for tokenizer training", len(texts))
+        with open(output_path / "train.txt", "w", encoding="utf-8") as f:
+            f.write("\n".join(texts))
+
+        logger.info(f"Training tokenizer: {self.train_config.tokenizer.model_type}")
+        logger.info(f"Tokenizer output: {self.train_config.tokenizer.save_path}")
+
+        # Example placeholder:
+        train_sentencepiece(
+            output_path / "train.txt",
```
Emrys365: Might need to avoid overwriting a possibly existing `train.txt` file?

espnet3/systems/base/inference.py
```diff
@@ -0,0 +1,122 @@
...
+        "Starting inference | decode_dir=%s test_sets=%s",
+        getattr(config, "decode_dir", None),
+        test_sets,
+    )
+
+    for test_name in test_sets:
+        logger.info("===> Processing test set: %s", test_name)
+        config.test_set = test_name
+        provider = InferenceProvider(config)
+        runner = InferenceRunner(provider=provider, async_mode=False)
```
Emrys365: Does this template always run inference sequentially (sample by sample)?

espnet3/systems/base/system.py
```diff
@@ -0,0 +1,131 @@
...
+    def infer(self, *args, **kwargs):
+        """Run inference on the configured datasets."""
+        self._reject_stage_args("infer", args, kwargs)
+        logger.info(
+            "Inference start | decode_dir=%s",
+            getattr(self.infer_config, "decode_dir", None),
+        )
+        return inference(self.infer_config)
+
+    def measure(self, *args, **kwargs):
```
Emrys365: How about making this method private (i.e., `_measure`) since it is called by `evaluate`?

espnet3/systems/base/train.py
```diff
@@ -0,0 +1,107 @@
...
+        set_parallel(cfg.parallel)
+
+    if cfg.get("seed") is not None:
+        L.seed_everything(int(cfg.seed), workers=True)
+
+    torch.set_float32_matmul_precision("high")
+
+    task = cfg.get("task")
+    if task:
+        save_espnet_config(task, cfg, cfg.exp_dir)
```
Emrys365: Can this work if `task` is None?

espnet3/utils/scp_utils.py
```diff
@@ -0,0 +1,96 @@
...
+    }
+
+
+def get_class_path(obj) -> str:
+    """Return the fully qualified class path for an object."""
+    return f"{obj.__module__}.{obj.__class__.__name__}"
+
+
+def load_scp_fields(
+    decode_dir: Path,
```
Emrys365: This argument name `decode_dir` is too specific to decoding-related tasks.

test/espnet3/components/data/test_collect_stats.py
```diff
@@ -334,38 +332,25 @@ def test_collect_stats_multiple_iterator(tmp_path: Path):
...
-            mode_dir / f"{k}_stats.npz",
-            count=count_dict[k],
-            sum=sum_dict[k],
-            sum_square=sq_dict[k],
-        )
-    with open(mode_dir / "stats_keys", "w") as f:
-        f.write("\n".join(sum_dict) + "\n")
+    assert (mode_dir / "stats_keys").exists(), "stats_keys not written"
+
+    files = os.listdir(mode_dir)  # カレントディレクトリ
```
Emrys365:
```suggestion
    files = os.listdir(mode_dir)  # current directory
```

espnet3/components/metrics/abs_metric.py
Masao-Someki: Thank you! Since I had that in mind to when creating inference part, but let me check with Jiatong this part again.

espnet3/systems/asr/metrics/cer.py
Masao-Someki: It includes a cleaner derived from ESPnet2, so we can use the same configuration for `clean_types`!
(edit: I checked `TextCleaner` class now, and I think we may use `tacotron` cleaner for such case)

https://github.com/Masao-Someki/espnet/blob/9938f5ed70653c5892f1ab1705a6bcda9e463d15/espnet3/systems/asr/metrics/cer.py#L37

espnet3/systems/base/inference.py
Masao-Someki: Thank you, this is not supported..! To support this, we need to modify the `BaseRunner` to support batch processing. I will address this and add a unit test to verify the functionality!

espnet3/systems/asr/metrics/cer.py
Masao-Someki: It includes a cleaner derived from ESPnet2, so we can use the same configuration for clean_types! 
I checked `TextCleaner` class now, and I think we may use `tacotron` cleaner for such case.

reference: https://github.com/espnet/espnet/blob/8dd980a698e9541c5bfc4c7b76728e6cb832f190/espnet2/text/cleaner.py#L27

espnet3/systems/asr/inference.py
Masao-Someki: >Is it correct? Is it via beam search?

By default, I thought that the ESPnet2 ASR inference class is used... However, I will revisit this part, as it should be made more flexible..!

> Does this method support parallel computing (e.g., multiprocessing)?

Yes!
If the process is implemented in the runner class, it can be executed in parallel.

espnet3/components/data/dataset.py
Masao-Someki: Thank you. There are several reasons based on my experiments.

It was originally added as a convenience for extracting training text (e.g., tokenizer or language model preparation), but in practice it introduced several issues:

1. Performance and I/O concerns
    Calling `get_text()` over the entire dataset can be very slow. Depending on how a dataset is implemented, this may trigger metadata loading or even audio loading, which is unnecessary and expensive for text-only extraction.

2. Unclear responsibility of the Dataset API
   Dataset should be expected to return training examples used in the training loop. Adding `get_text()` mixes in a different responsibility (corpus extraction for dataset creation), which makes the API harder to understand, especially for users unfamiliar with the original intent.

3. Better alternatives exist in an earlier stage
   Extracting text directly from the dataset directory or manifests during a separate “create-dataset” or preprocessing stage is significantly faster, simpler, and more explicit. We want users to rely on that approach rather than calling `get_text()` on training datasets.

Based on these points, keeping `get_text()` in the Dataset/CombinedDataset API caused confusion and encouraged inefficient usage, so it was removed to keep the dataset interface focused and clear.

espnet3/components/metrics/abs_metric.py
Masao-Someki: Thank you!
As you pointed out, the purpose of `data`/`test_name` was not very clear, so I updated the docstring of `AbsMetrics.__call__` to include concrete examples of `data`, an explanation of additional inputs, and a clarification that test_name corresponds to the test set name defined by the data organizer.

espnet3/parallel/inference_provider.py
Masao-Someki: Thank you, this is a good idea!

espnet3/systems/asr/system.py
Masao-Someki: Thank you, I added `tokenizer.train_file` support in training config. If it is set, the training text is written to that path; otherwise it defaults to `train.txt`. In both cases, if the target file already exists, training fails with an error to avoid overwriting.

espnet3/systems/base/train.py
Masao-Someki: Yes, this works fine when task is None!

Masao-Someki:
## About [292886c](https://github.com/espnet/espnet/pull/6328/commits/292886c51df4a8c059aedec6ebb0e4a527efd567)
Background:
                                                                                                      
  - Inference logic assumed a fixed espnet2-style output format (e.g., model_output[0][0]), making it 
    fragile to differences across tasks or models.                                                    
  - We want to flexibly switch dataset input keys and output structures via YAML configuration.       
                                                                                                      
  Technical Changes:
                                                                                                      
  - The inference process has been externalized using `input_key` and `output_fn` (specified as an import 
    path). What to extract from the dataset and how to format model outputs are delegated to recipe   
    `src/infer.py`.
  - The runner raises a KeyError if `input_key` does not exist in the dataset.                          
  - The output format now requires `uttid`; all other keys returned by `output_fn` are written to SCP without assuming hyp/ref semantics.                                  
  - `InferenceProvider` now includes standard build_dataset and build_model implementations, eliminating
    the need for task-specific Provider classes.

espnet3/components/metrics/abs_metric.py
Masao-Someki: I confirmed that we can directly use VERSA with the current system output. However, I think we should invoke it via the CLI, since the interface for importing and using Python functions may change significantly. We can support this by calling VERSA when we implement the measuring stage in audio-generation systems like tts, svs, enh, etc.

espnet3/systems/base/train.py
Emrys365: To be clear, I meant to ask whether the entire `train.py` script can work properly if `task` is not specified. If `task` is None, what is the behavior of the script? What is the case where `task` could be None?

Masao-Someki: Sorry for the misunderstanding, here's a clarification of this part.
First, `task` is only needed when we want to use espnet2 Tasks/models. In this case, we need to save `config.yaml` in specific format to use the inference APIs (from espnet2).
If task is not set, we treat it as an espnet3-native, or user-defined custom model. In this case, `train.yaml` (and `infer.yaml`) are expected to fully define the model and inference setup, so we don't need to save the espnet2-style config.
For packaging/publishing a model, the design is to include `train.yaml` and `infer.yaml` (and of course, other configs).

espnet3/systems/base/inference.py
Masao-Someki: Thank you, now I supported the batch inference!
I will mark this as resolved, and I will put further discussions on the above shinji's review comment.

espnet3/systems/asr/inference.py
Masao-Someki: Thank you @sw005320 and @Emrys365 for your comments! I have reflected them and implemented the following items:

- A more flexible inference API in [292886c](https://github.com/espnet/espnet/pull/6328/commits/292886c51df4a8c059aedec6ebb0e4a527efd567). See https://github.com/espnet/espnet/pull/6328#issuecomment-3767017846 for more details.
- Batch inference via `batch_forward` function in [733394f](https://github.com/espnet/espnet/pull/6328/commits/733394f414cfb6822b59e49e223a62bf55967792)

sw005320: @Masao-Someki, I want to clarify.

Some inference methods natively support batch inference, while others may be performed in multiple passes.
Does this implementation support both cases?

Masao-Someki:
Thank you, @sw005320 and @Emrys365, for your reviews!
I've addressed all of the comments. If you have time, could you please take another look at this PR?

You can find the current ESPnet3 documentation here: [ESPnet3 Demo Homepage](https://masao-someki.github.io/espnet_draft_home_page/espnet3/). 
For the scope of this PR, please refer to the following page, which describes the expected additions: ([ESPnet3 Systems](https://masao-someki.github.io/espnet_draft_home_page/espnet3/core/systems.html))

The functionalities added in response to the review comments are documented here:
-  [`output_fn`: formatting model outputs into SCP fields](https://masao-someki.github.io/espnet_draft_home_page/espnet3/stages/inference.html#output-fn-formatting-model-outputs-into-scp-fields)
- [Batched inference](https://masao-someki.github.io/espnet_draft_home_page/espnet3/stages/inference.html#batched-inference-batch-size-batch-forward)

Masao-Someki:
It would also be great if @popcornell could review this PR, as it’s a fairly large one!

espnet3/systems/asr/inference.py
Masao-Someki: @sw005320 Yes. Here are the clarifications:
- **Batch inference**: This is now supported as of commit [733394f](https://github.com/espnet/espnet/pull/6328/commits/733394f414cfb6822b59e49e223a62bf55967792).
- **Multiple inference paths**: If a system requires two inference paths based on the input, we can simply use an `if–else` statement in the `forward` or `batch_forward` function.

I also noticed that there is currently no mechanism to configure hyperparameters that control the inference-path.
I will add a commit for a new config `infer.yaml: provider.params`, so that users can manage these settings directly from the config file.

espnet3/systems/asr/tokenizer/sentencepiece.py
```diff
@@ -0,0 +1,224 @@
...
+        ["<blank>", "<unk>"]
+        + [line.split("\t")[0] for line in lines][3:]
+        + ["<sos/eos>"]
+    )
+    with open(os.path.join(save_path, "tokens.txt"), "w") as f:
+        f.write("\n".join(vocabs))
+
+
+def add_special_tokens(
+    tokenizer, converter, embedding, special_tokens, insert_after="<st_zho>"
```
popcornell: the insert_after is very specific. I am wondering if we should not use default at all here since it can lead to bugs ?

popcornell (commented):
Looks good to me. I have a PR here with some corrections on the docstrings. 
https://github.com/Masao-Someki/espnet/pull/32/changes

espnet3/systems/asr/tokenizer/sentencepiece.py
Masao-Someki: Thank you, I have fixed this part!
- If `insert_after` is not specified, then we will add the special token at the last position.

espnet3/parallel/base_runner.py
```diff
@@ -190,6 +199,23 @@ def forward(idx: int, *, dataset, model, **env) -> Any:
         """
         raise NotImplementedError
 
+    @classmethod
+    def batch_forward(cls, indices: Iterable[int], *, dataset, model, **env) -> Any:
```
sw005320: Why do we need both forward and batch_forward?
Can we unify them?

egs3/TEMPLATE/asr/src/inference.py
sw005320, line 1: infer --> inference

Please make sure to unify the notation

espnet3/systems/base/measure.py
sw005320, line 1: measure --> metric

espnet3/systems/base/system.py
```diff
@@ -0,0 +1,124 @@
...
+            if isinstance(model_cfg, DictConfig):
+                model_target = model_cfg.get("_target_")
+        logger.info(
+            "Training start | exp_dir=%s model=%s",
+            getattr(self.train_config, "exp_dir", None),
+            model_target or "<unknown>",
+        )
+        return train(self.train_config)
+
+    def infer(self, *args, **kwargs):
```
sw005320, line 101: why `infer`?
inference?

espnet3/systems/base/system.py
```diff
@@ -0,0 +1,124 @@
...
+        """Run inference on the configured datasets."""
+        self._reject_stage_args("infer", args, kwargs)
+        logger.info(
+            "Inference start | infer_dir=%s",
+            getattr(self.infer_config, "infer_dir", None),
+        )
+        return inference(self.infer_config)
+
+    def measure(self, *args, **kwargs):
```
sw005320: measure --> metric

test/espnet3/systems/asr/test_asr_inference.py
sw005320, line 1: please make sure to batchfy it.

test/espnet3/systems/base/test_metric.py
sw005320, line 1: test_metric

espnet3/systems/base/system.py
popcornell: I think it is because it is the verb and it is a method. Inference would be more appropriate for a class/file name

espnet3/systems/base/system.py
popcornell: as before.

egs3/TEMPLATE/asr/src/inference.py
popcornell: I agree on this

espnet3/systems/base/measure.py
popcornell: this too since it is a file and not a method it should be the name not the verb

espnet3/systems/base/system.py
sw005320: Do you mean we should consistently change all our naming with the verb and method?
(e.g., train and training)?

If it is practical to change everything, we can do so.
Can you propose where we use train and where we use training?
Can you also check the entire code?

Note that I'm talking about the naming consistency.

sw005320: If you really want to do so, can you propose consistently changing everything?
For me, instead, unifying them is much easier, but if you can do it perfectly, this is great!

espnet3/components/optimizers/multiple_optimizer.py
sw005320, line 1: optim --> optimizer according to your naming convention?

sw005320: or `optimizers` (plural according to the naming conventions)

espnet3/systems/base/inference.py
sw005320, line 1: base --> bases?

sw005320: Under espnet3 and components, you use the plural form, while under the base, you use the singular form.
 I'd recommend unifying them, or if you separate them, make this distinction clear, and also add documents about the naming rules.

Masao-Someki: Thank you, I will fix it and check other part on this point!!

Masao-Someki: Thank you for pointing out this part!

Under `systems/`, each subdirectory is intended to be a system label (e.g., asr, tts, base), not a pluralized category. For instance, renaming `base` to `bases` would imply multiple base systems, whereas the intention is to have a single base system.

I agree that this distinction is not currently clear, so I will add the section about plural/singular form on the documentation about the naming convention.

espnet3/systems/base/system.py
```diff
@@ -0,0 +1,124 @@
...
+            "Inference start | inference_dir=%s",
+            getattr(self.infer_config, "inference_dir", None),
+        )
+        return infer(self.infer_config)
+
+    def metric(self, *args, **kwargs):
+        """Compute evaluation metrics from hypothesis/reference outputs."""
+        self._reject_stage_args("metric", args, kwargs)
+        logger.info(
+            "metricment start | metric_config=%s",
```
popcornell: what is metricment ?

espnet3/systems/base/metric.py
```diff
@@ -0,0 +1,56 @@
...
+
+from espnet3.components.metrics.abs_metric import AbsMetric
+from espnet3.utils.scp_utils import get_cls_path, load_scp_fields
+
+
+def metric(metrics_config: DictConfig):
+    """Compute metrics for each test set and write a metrics JSON file.
+
+    Args:
+        config: Hydra/omegaconf configuration with inference and metric settings.
```
popcornell: this does not match the function

espnet3/systems/asr/tokenizers/sentencepiece.py
```diff
@@ -0,0 +1,230 @@
...
+
+    Raises:
+        FileNotFoundError: If the specified `dump_text_path` does not exist.
+        Exception: If the training of the SentencePiece model fails for any
+            reason.
+
+    Examples:
+        >>> train_sentencepiece(
+        ...     dump_text_path='path/to/train.txt',
+        ...     output_path='path/to/output',
```
popcornell: it should be save_path now

Masao-Someki:
Thank you, I've fixed the review comment!

espnet3/components/trainers/trainer.py
sw005320, line 1: `training` --> `trainers`? If this directry means the collection of trainers?

sw005320: Note that I suggested `training` previously as a wide category, including all training classes/functions, but this directly seems to only include `trainer`. In this case, `trainers` are more appropriate.

espnet3/systems/base/inference.py
sw005320: I see. Thanks for the clarification.

espnet3/components/trainers/trainer.py
sw005320: How is it related to `base/training`?
Depending on the relationship, we may keep `training`

Masao-Someki: I also felt that changing this to **trainers** might actually be a better name, so I'll go ahead and make that change.
The `training (trainer)` module in the `components` directory would contain trainers or frameworks for common training logic, while `systems/base/training.py` would contain the actual training process. With this we can clearly separate system-specific training logic.

Also if we add a new trainer in the future, we can place it in `components`, and any system that wants to use it can simply import and use it from `systems/*/training.py`.

sw005320: okay, sounds good

sw005320:
After we fix the CI issue, I can merge this PR.

@popcornell and @Emrys365, if you catch any issues with naming or anything, please add comments.

Masao-Someki:
I've fixed the CI in another PR. After we check all CI passes we can merge this and move to espnet3-13 PR!

sw005320:
Thanks, @Masao-Someki!

Emrys365:
@Masao-Someki  I found some inconsistencies of `uttid` and `utt_id` in the merged code. Maybe this should be unified too?

Masao-Someki:
Thank you @Emrys365, I overlooked that one! I will fix it in the PR 3-14.5 (#6350), which is a bugfix PR.

Emrys365:
Should `inference_dir` be simplified to `infer_dir`? I saw the use of `infer_dir` in https://github.com/Masao-Someki/espnet/blob/espnet3/recipe/ls_asr100_2/egs3/librispeech_100/asr/conf/inference.yaml.

Also, `decode_dir` in some config files should also be unified to `infer_dir`, I think.

PR #6331, opened 2025-12-26
[espnet3-14] Add integration test

Masao-Someki:
## What did you change?
                                                                                                                             
  - Added an espnet3 integration test in CI and introduced TEMPLATE/mini_an4 ASR example configs, scripts, and dataset       
    generation code.                                                                                                         
  - Added jiwer as a dependency for the scoring stage.                                                                       
  - Moved the mini_an4 dataset archive to egs2 to avoid using egs.                                                           

  ## Why did you make this change?
                                                                                                                             
  - To validate espnet3 end‑to‑end behavior via a reproducible CI integration test.                                          
  - To declare the dependency required for scoring.                                                                          
  - To align with the policy of using egs2 instead of egs.                                                                   

                                                                                                           
  ## Is your PR small enough?                                                                                                
                                                                                                                             
  no                                                                                                                         
  (more than 20 files changed, because it includes TEMPLATE directory.) 
                                                                                             
  ## Additional Context                                                                                                      
                                                                                                                             
  - egs2/mini_an4/asr1/downloads.tar.gz is a binary update.

This PR should be merged after #6329.

* #6133
* #6327
* #6328
* #6329

PR #6329, opened 2025-12-24
[espnet3-13] Add logging utils

Masao-Someki:
## What did you change?

Added logging utility and attached loggers.

---

## Why did you make this change?

To enhance the logging feature of espnet3.

---

## Is your PR small enough?

yes, 18 files and -600 line edits.

---

## Additional Context

This PR should be merged after #6328.

* #6133
* #6323
* #6328

PR #6338, opened 2026-01-16
[espnet3-15] Add publication stage

Masao-Someki:
## What did you change?

Added the following stage for the publication
- `pack_model`: create a directory for uploading model into HF hub.
- `upload_model`: Upload model to HF hub.

---

To easily share the experimental results.

---

## Is your PR small enough?

Yes, it is 12 file changes with 808 additions after the #6331 has merged.

---

## Additional Context

This PR should be merged after #6331.

* #6133
* #6327
* #6328
* #6329
* #6331


Masao-Someki:
I have uploaded the model from integration test and it looks like this
https://huggingface.co/ms180/mini_an4_integration_test

Masao-Someki:
@sw005320 This PR is ready for review.

If we set the `HF_TOKEN` as a secret and attach the `test_upload` label, the integration test will trigger the `upload_model` stage. This will create the `espnet3/CI_mini_an4_training_asr_transformer` repository. I’ve currently tested this using my own account [here](https://huggingface.co/ms180/CI_mini_an4_training_asr_transformer/tree/main)

An additional integration test has been added to verify that the packaged model works correctly with the `InferenceModel` class (`ci/test_integration_espnet3_publication_check.py`).

~I also noticed an issue with the README in the Hugging Face Space, and I’ll address that bug shortly.~ -> Done

espnet3/publication/inference_model.py
```diff
@@ -0,0 +1,494 @@
...
+            }
+            try:
+                batch_result = self.runner_cls.forward(
+                    index_list,
+                    dataset=dataset,
+                    model=self.model,
+                    input_key=self.input_key,
+                    output_fn=self.output_fn,
+                )
+            except (TypeError, NotImplementedError, RuntimeError):
```
popcornell: This is maybe too aggressive error catching for me. 

Maybe you can do: 

```python
except (TypeError, NotImplementedError) as e:
    logger.debug(
        "Runner does not support batched inference; falling back to per-sample.",
        exc_info=True,
    )
    batch_result = None
except RuntimeError as e:
    logger.warning(
        "Batched inference raised RuntimeError (%s); falling back to per-sample. "
        "This may indicate CUDA OOM or a shape mismatch.",
        e,
    )
    batch_result = None
```

even better if you can break it up in 3 types of error by catching CUDA OOM expliocitly.

espnet3/utils/publication_utils.py
```diff
@@ -0,0 +1,628 @@
...
+    lines = []
+    for line in template.splitlines():
+        keys = pattern.findall(line)
+        if keys and any(not safe.get(k, "") for k in keys):
+            continue
+        lines.append(line)
+    return Template("\n".join(lines)).safe_substitute(safe)
+
+
+def _write_meta(
```
popcornell, line 510: For future & backward compatibility we can add a "schema_version" key. 
And then in from_packed we can control for the schema_version:

```python
schema = int(meta.get("schema_version", 0))
if schema == 0:
    # treat as legacy / pre-versioned; allow but warn
elif schema == 1:
    # current code path
else:
    raise ValueError(
        f"Bundle was produced by a newer pack_model (schema_version={schema}) "
        f"than this installation supports. Upgrade espnet3."
    )
```

jctian98:
My general comment is to include a version/GitHub commit ID for the model you publish. The codebase would be updated, and sometimes some old models cannot be reproduced with the new code. By recording a commit ID, users can at least find an old but runnable code example.

Masao-Someki:
Thank you @jctian98 , I added git-related items and the sample repository has updated [here](https://huggingface.co/ms180/CI_mini_an4_training_asr_transformer)

It has git origin, branch, commit id now.

egs3/mini_an4/asr/conf/publication.yaml
```diff
@@ -0,0 +1,10 @@
+
+pack_model:
+  include:
+    - ${recipe_dir}/src
+    - ${data_dir}/**/bpe.model
+    - ${data_dir}/**/bpe.vocab
+    - ${data_dir}/**/tokens.txt
+
+upload_model:
+  hf_repo: ms180/CI_mini_an4_training_asr_transformer
```
Fhrozen: Is this a test repo? should be left blank or parsed as required.

egs3/TEMPLATE/asr/src/hf_model_repo_readme_template.md
```diff
@@ -0,0 +1,66 @@
+---
+tags:
+- espnet
+- ${task}
```
Fhrozen: You can also find the `corpus` tag.

egs3/TEMPLATE/asr/src/hf_model_repo_readme_template.md
```diff
@@ -0,0 +1,66 @@
...
+language: ${lang}
+license: ${license}
+---
+
+# ESPnet3 ${task} model
```
Fhrozen: I am not sure if you are using the whole name (as HF-task like automatic speech recognition) or the abbrevation (as ESPnet-task like ASR)

Fhrozen (approved):
LGTM, minor comments

Fhrozen (approved):

.github/workflows/ci_on_ubuntu.yml
```diff
@@ -311,6 +320,45 @@ jobs:
...
+      github.event.pull_request.draft == false
+    env:
+      ESPNET3_PUBLICATION_TEST_UPLOAD: ${{ needs.process_labels.outputs.test_upload }}
+      HF_TOKEN: ${{ secrets.HF_TOKEN }}
+    steps:
+      - uses: actions/checkout@master
+      - uses: actions/cache@v5
+        with:
+          path: ~/.cache/pip
+          key: ${{ runner.os }}-pip-publication-3.10-2.9.1-${{ hashFiles('**/pyproject.toml') }}-${{ hashFiles('**/Makefile') }}
```
sw005320: This information is hard-coded.
If it is necessary, it is fine.
We (developers) need to take care of it.

espnet3/utils/publication_utils.py
```diff
@@ -0,0 +1,632 @@
...
+    """
+    pack_cfg = publication_config.pack_model
+
+    recipe_root = Path(training_config.recipe_dir).resolve()
+
+    # Create (or recreate) the output directory
+    exp_dir = Path(training_config.exp_dir)
+    out_dir = Path(getattr(pack_cfg, "out_dir", exp_dir / "model_pack"))
+    if out_dir.exists():
+        shutil.rmtree(out_dir)
```
popcornell, line 577: is there some mechanism that will prevent some stupid user (like me) to put by mistake out_dir = my training run and lose everything ?

espnet3/utils/publication_utils.py
```diff
@@ -0,0 +1,632 @@
...
+    out_dir: Path,
+    recipe_root: Path,
+) -> tuple[Path, Path]:
+    """Validate a named artifact and return its (src, dst) path pair.
+
+    Called by ``pack_model`` before copying each entry in ``pack_cfg.files``
+    and ``pack_cfg.yaml_files``. Raises if the source is missing or lives
+    outside the recipe root.
+    """
+    src = Path(src_path)
```
popcornell, line 109: will the artifact always be under recipe root ? 
What if one uses some huggingface downloaded model in their recipe ? How that is handled ? It is not put into ~/.cache/huggingface ? I am thinking e.g. codecs or tokenizers.

espnet3/utils/publication_utils.py
```diff
@@ -0,0 +1,632 @@
...
+
+    _write_meta(out_dir, files=files, yaml_files=yaml_files)
+    logger.info("Packed model to %s", out_dir)
+    return out_dir
+
+
+def _run(cmd: list[str], cwd: Path | None = None) -> str:
+    """Run a subprocess command and return stdout.
+
+    Used by ``upload_model`` to invoke ``huggingface-cli`` for repo creation
```
popcornell: stderr and stout is captured here no ? 
Is the user informed of the process is ongoing ? cos it may take some time. 
Maybe we can call in upload_model before this run: logger.info("uploading %s -> %s", pack_dir, repo)

egs3/TEMPLATE/asr/conf/publication.yaml
```diff
@@ -0,0 +1,75 @@
...
+    # These patterns are applied while copying `${exp_dir}` and any included
+    # directories.
+    - inference
+    - inference_*
+    - last.ckpt
+    - "**/step*.ckpt"
+    - "**/*.log"
+    - "**/tensorboard/**"
+    - "**/wandb/**"
+  files:
```
popcornell: files overrides exclude right ? because exclude will skip last.ckpt

egs3/TEMPLATE/asr/src/hf_model_repo_readme_template.md
Masao-Someki: It is the task name from espnet, so it is a shortened version.

.github/workflows/ci_on_ubuntu.yml
Masao-Someki: Thank you, we don't need to stick to this version. I will fix this part

espnet3/utils/publication_utils.py
Masao-Someki: That's right, it can happen in the current code!
I will add warning or additional flag that accept override or removal

egs3/TEMPLATE/asr/src/hf_model_repo_readme_template.md
Masao-Someki: The ${task} placeholder is resolved by `_infer_task_name()` in espnet3/utils/publish.py, for example, `espnet3.systems.asr.ASRTask` becomes `asr`.
If task is not set in the config, it falls back to the recipe directory
 (e.g. "asr").

egs3/TEMPLATE/asr/src/hf_model_repo_readme_template.md
Masao-Someki: Good idea, thanks!

egs3/mini_an4/asr/conf/publication.yaml
Masao-Someki: Thanks, and this is a test repo.
I will move it to espnet/

espnet3/utils/publication_utils.py
Masao-Someki: > will the artifact always be under recipe root ?

That was my understanding, but you're right, in practice external files can be used..! Like a pretrained model on a shared cluster, or a tokenizer from a shared directory.

I'm thinking of adding a warning if `pack_model` detects an external path, it will warn the user that the directory structure may not be preserved as expected, and suggest using a symlink under the recipe directory instead. 
Since if we have symlinks during packing dirs, the actual files will be copied and the structure under the link will be preserved correctly.

egs3/TEMPLATE/asr/conf/publication.yaml
Masao-Someki: Yes, I added a comment on the config

espnet3/utils/publication_utils.py
Masao-Someki: Thanks, I added the logging and now stderr/out will be visible!

Masao-Someki:
@Fhrozen Thank you! All CI has passed!!

Masao-Someki:
@sw005320 I think we can merge this PR if this looks ok to you!
The next PR is #6342, which adds the demo stage!

egs3/TEMPLATE/asr/conf/publication.yaml
```diff
@@ -0,0 +1,91 @@
...
+# registered as a named artifact in `meta.yaml` (step 2). Without the
+# exclude, `last.ckpt` would be copied twice to the same destination.
+#
+# `inference_dir` inside this block is used only to find the `metrics.json`
+# written by the `measure` stage for the bundled README and result summary.
+pack_model:
+  out_dir: ${exp_dir}/model_pack
+  allow_overwrite: false
+  inference_dir: ${inference_dir}
+  readme: egs3/TEMPLATE/asr/src/hf_model_repo_readme_template.md
```
sw005320: Please use the relative path or whatever to avoid  `egs3/TEMPLATE/asr`

egs3/TEMPLATE/asr/conf/publication.yaml
sw005320, line 1: avoid having a task (system) name dependency (e.g., `asr`) as much as possible

egs3/TEMPLATE/asr/conf/publication.yaml
```diff
@@ -0,0 +1,91 @@
...
+    - inference_*
+    - last.ckpt        # excluded from bulk copy; registered via `files` below
+    - "**/step*.ckpt"
+    - "**/*.log"
+    - "**/tensorboard/**"
+    - "**/wandb/**"
+  files:
+    # Named artifacts appear in `meta.yaml` and can be resolved by
+    # `InferenceModel.from_packed(...)`. `exclude` does not apply here.
+    asr_model_file: ${exp_dir}/last.ckpt
```
sw005320: Does this has to be `asr_`?
Can we use `model_file`?

This avoid to add dependencies of the task/system

egs3/TEMPLATE/asr/src/hf_model_repo_readme_template.md
```diff
@@ -0,0 +1,68 @@
+---
+tags:
+- espnet
+- ${task}
+- ${corpus}
+language: ${lang}
+license: ${license}
+---
+
+# ESPnet3 ${task} model
```
sw005320: task? system?
I'm confised

egs3/TEMPLATE/asr/src/hf_model_repo_readme_template.md
```diff
@@ -0,0 +1,68 @@
...
+
+${description}
+
+## Model
```
sw005320, line 14: - Total number of parameters
- Total number of "learnable" parameters

egs3/TEMPLATE/asr/src/hf_model_repo_readme_template.md
sw005320, line 1: The path/filename is redundant.
This is under the template directory. Why you need to add the `template` as a filename again?

sw005320: what does `repo` mean?

espnet3/publication/inference_model.py
sw005320, line 1: This requires other people to check it.
@Masao-Someki, please contact the main developers (e.g., @popcornell, @jctian98, etc.) to confirm this interface would be good for the other tasks/systems.

espnet3/utils/publication_utils.py
sw005320, line 1: I leave the detailed reviews to others, but please make sure each function goes through testing.

egs3/TEMPLATE/asr/conf/publication.yaml
Masao-Someki: Thank you!
This was for espnet2 models, so I removed from the TEMPLATE!

egs3/TEMPLATE/asr/conf/publication.yaml
Masao-Someki: Thank you, this part should use the recipe_dir!

egs3/TEMPLATE/asr/src/hf_model_repo_readme_template.md
Masao-Someki: Thank you, this should be system, and I will remove task.. For espnet2 models, we will define task in the config and it will be included as system.

egs3/TEMPLATE/asr/src/hf_model_repo_readme_template.md
Masao-Someki: Thank you, I will try to include model information as much as possible!
Currently I will update this PR to include:

  - total number of parameters
  - number of learnable params
  - non-trainable params
  - model size
  - dtype composition
  - model class
  - model repr (print(model) with details, which the UI can be expanded with click)

egs3/TEMPLATE/asr/src/hf_model_repo_readme_template.md
Masao-Someki: The filename originally tried to indicate that this README is for a Hugging
  Face model repo, but that is redundant under the template directory. I will
  rename it to `hf_model_readme.md`

sw005320, line 1: Is it duplicated?

egs3/TEMPLATE/asr/src/hf_model_repo_readme_template.md
sw005320: @Masao-Someki, we can delegate this to the others.

egs3/TEMPLATE/asr/src/hf_model_repo_readme_template.md
sw005320: @Masao-Someki, if it is solved, we can merge this PR.
Please discuss https://github.com/espnet/espnet/pull/6338#discussion_r3263056125 with the other team members and make a follow up PR

egs3/TEMPLATE/asr/conf/publication.yaml
```diff
@@ -0,0 +1,83 @@
...
+    - last.ckpt
+    - "**/step*.ckpt"
+    - "**/*.log"
+    - "**/tensorboard/**"
+    - "**/wandb/**"
+
+##########################################################
+#                  UPLOAD MODEL SETTING                  #
+##########################################################
+# Upload requires `huggingface-cli` and a valid Hugging Face login.
```
popcornell: huggingface-cli is deprecated. Should we still use it or move to hf command ?

Masao-Someki:
I just found one bug and I will update it. When we have symlink to very different directory (out of the recipe directory) it causes issue when resolving path..
I already fixed it, so I will update the `publication_utils.py` and will fix corresponding unit test.

espnet3/utils/publication_utils.py
Masao-Someki: image

I just tested this publication stage and it works well when we already have the directory.

egs3/TEMPLATE/asr/src/hf_model_repo_readme_template.md
Masao-Someki: Oh I will remove it, thank you! (It should be removed...)
Also I will check the additional items with others.

egs3/TEMPLATE/asr/conf/publication.yaml
Masao-Someki: You're right, let me change the instruction comment!!

espnet3/utils/publication_utils.py
```diff
@@ -0,0 +1,751 @@
...
+    README. Checks ``inference_dir`` on each provided config in priority order
+    (publication → metrics → inference) and returns the first ``metrics.json``
+    found.
+
+    """
+    for cfg in (publication_config, metrics_config, inference_config):
+        inference_dir = getattr(cfg, "inference_dir", None) if cfg else None
+        if not inference_dir:
+            continue
+        matches = list(Path(inference_dir).rglob("metrics.json"))
```
popcornell: do we have to use rglob here ? don t we know the path ? isnt the path simply inference_dir / metrics.json ?

espnet3/utils/publication_utils.py
popcornell: I think now is better but we should still check that the directory is a valid pack)model directory. 
E.g. it contains meta.yaml. 

I used claude for this snippet: 
```
if out_dir.exists():
    if not getattr(pack_cfg, "allow_overwrite", False):
        raise RuntimeError(
            f"out_dir already exists: {out_dir}. "
            "Set pack_model.allow_overwrite: true to overwrite."
        )
    _assert_safe_to_overwrite(out_dir, exp_dir=exp_dir, recipe_root=recipe_root)
    shutil.rmtree(out_dir)
out_dir.mkdir(parents=True)
```

```
def _assert_safe_to_overwrite(
    out_dir: Path, exp_dir: Path, recipe_root: Path
) -> None:
    """Refuse to rmtree anything that isn't clearly a prior pack output."""
    resolved = out_dir.resolve()

    # 1) Never follow a symlink — rmtree would delete the link's target.
    if out_dir.is_symlink():
        raise RuntimeError(f"Refusing to overwrite a symlink: {out_dir}")

    # 2) Never delete critical roots or an ancestor of the experiment.
    protected = {
        Path("/").resolve(),
        Path.home().resolve(),
        Path.cwd().resolve(),
        exp_dir.resolve(),
        recipe_root.resolve(),
    }
    if resolved in protected or resolved in exp_dir.resolve().parents:
        raise RuntimeError(
            f"Refusing to delete {out_dir}: protected or parent directory. "
            "Set pack_model.out_dir to a dedicated pack directory."
        )

    # 3) The real safeguard: only overwrite a directory we produced.
    #    Every pack_model() output contains meta.yaml (see _write_meta).
    if not (resolved / "meta.yaml").is_file():
        raise RuntimeError(
            f"Refusing to overwrite {out_dir}: it doesn't look like a previous "
            "pack_model() output (no meta.yaml). Delete it manually if intended."
        )
```

egs3/TEMPLATE/asr/conf/publication.yaml
popcornell: what about using the python library ? HfApi ? instead of subprocess and then running commands ?

espnet3/utils/publication_utils.py
```diff
@@ -0,0 +1,765 @@
...
+    try:
+        api.repo_info(repo_id=repo, repo_type="model")
+        return True
+    except RepositoryNotFoundError:
+        return False
+    except Exception as exc:
+        raise RuntimeError(f"Failed to check repo existence for {repo}: {exc}") from exc
+
+
+def upload_model(system) -> None:
```
popcornell, line 683: is there a way to upload a private model ?

espnet3/utils/publication_utils.py
```diff
@@ -0,0 +1,765 @@
...
+        system: ESPnet3 system instance with ``publication_config``.
+
+    Raises:
+        RuntimeError: If required config values are missing or upload fails.
+
+    Examples:
+        >>> upload_model(system)
+    """
+    publication_cfg = system.publication_config
+    repo = publication_cfg.upload_model.hf_repo
```
popcornell: can we check 

if not repo or repo == "yourname/your-model-repo": raise RuntimeError("Set publication.upload_model.hf_repo to your repo.")

Because otherwise it throws a weird error if the user does not change it

espnet3/utils/publication_utils.py
```diff
@@ -0,0 +1,765 @@
...
+    """Check if a Hugging Face model repo exists.
+
+    Used by ``upload_model`` before uploading to decide whether to create
+    the repo first via ``hf repos create``.
+
+    """
+    try:
+        from huggingface_hub import HfApi
+        from huggingface_hub.utils import RepositoryNotFoundError
+    except Exception as exc:
```
popcornell: HfHubHTTPError ?

popcornell:
As far as I understand, right now, if we run recipe e2e it tries to publish the model but it fails if we do not pass `--publication_config` right ? 
I am wondering if it is just better to raise a warning like publication stage skipped, in order to publish model you need to pass --publication_config etc etc. You can resume publication by etc etc

espnet3/utils/publication_utils.py
Masao-Someki: Thanks, removed!!

Masao-Someki:
@popcornell 
> As far as I understand, right now, if we run recipe e2e it tries to publish the model but it fails if we do not pass --publication_config right ?

It is slightly different. If not enough config is provided for the requested stages, run.py will raise error and stop processing.
You can check the [TEMPLATE/asr/run.py](https://github.com/Masao-Someki/espnet/blob/2be2f5413827a7c3107253256206c0f5916393fb/egs3/TEMPLATE/asr/run.py#L174-L199)

I think one way is we can remove publication stages from the default stage list (which runs with `--stages all`)

espnet3/utils/publication_utils.py
Masao-Someki: Thank you!
I added error when we got issue with uploading. This can happen when token might be wrong, people might push to someone else's repo (by mistakenly not updating from default config) etc

espnet3/utils/publication_utils.py
Masao-Someki: I changed to the python library, which has the private repo access.
So now we can add it in config like:

```
upload_model:
  hf_repo: yourname/your-model-repo
  private: false
```

sw005320:
@Masao-Someki, can you check this error?
It seems reproducible.

sw005320:
https://github.com/espnet/espnet/actions/runs/27015525556/job/79972055726?pr=6338

Masao-Someki:
The CI has passed. If this looks good to @sw005320, let's merge and move on to espnet3-16!

I found a bug in the espnet3-16 implementation, so I'll fix it along with any review feedback.

sw005320:
Thanks, @Masao-Someki!

PR #6339, opened 2026-01-16
inference trick from: "Improving Cross-Attention based on Positional Alignment during Inference for Robust Long-form Speech Recognition"

Miamoto:
## What did you change?

	•	Added the implementation from "Improving Cross-Attention based on Positional Alignment during Inference for Robust Long-form Speech Recognition"
	•	Introduced new decodingoption to enable inference trick: "inference_lf_trick: True" can be added in decode_asr.yaml file.

---

---

## Is your PR small enough?

yes

---


sw005320:
- Can you fix the CI error?
- Can you come up with a better name for `inference_lf_trick`?
- For the CTC prefix score, [margin](https://github.com/espnet/espnet/blob/df9514742b18d7f599be6fdc0db83cba627abf44/espnet2/legacy/nets/ctc_prefix_score.py#L30) is prepared in `CTCPrefixScorer`, but it cannot be configurable via the current option argument. So, how about using this?

PR #6354, opened 2026-02-02
[SpeechLM] Trainer, processor and model

jctian98:
This PR merges the recent updates about the trainer, processor, and model files inthe  SpeechLM module.

(1) abs_job.py: add the "is_train" argument in job template abstract class
(2) speechlm_job.py: update the preprocessor and collate_fn to support CFG and sequence packing.
(3) parallel.py: update the file to add the whole model inference logic.
(4) deepspeed_trainer.py: update small training setup.

### DIFF
--- a/.devcontainer/ci_cpu/espnet.dockerfile
+++ b/.devcontainer/ci_cpu/espnet.dockerfile
@@ -58,8 +58,9 @@ RUN apt-get update && \
 
 # Latest version of git
 ENV TZ=Etc/UTC
-ENV TH_VERSION=2.0.1
-ENV USE_CONDA=false
+ENV TH_VERSION=2.7.1
+ENV ESPNET_PYTHON_VERSION=3.12
+ENV USE_CONDA=true
 ENV CHAINER_VERSION=6.0.0
 ENV PATH=/opt/miniconda/bin:${PATH}
 
@@ -72,13 +73,5 @@ RUN mkdir -p /workspaces && \
 
 WORKDIR /opt
 USER ${USERNAME}
-
-RUN wget --tries=3 -nv "https://repo.anaconda.com/miniconda/Miniconda3-latest-Linux-x86_64.sh" -O miniconda.sh && \
-    bash miniconda.sh -b -p /opt/miniconda && \
-    /opt/miniconda/bin/conda config --prepend channels https://software.repos.intel.com/python/conda/ && \
-    rm miniconda.sh && \
-    conda install -y python=3.10 && \
-    conda install -c conda-forge libstdcxx-ng && \
-    conda clean -a -y
 
 WORKDIR /workspaces
--- a/.github/actions/install-ffmpeg/action.yml
+++ b/.github/actions/install-ffmpeg/action.yml
@@ -1,0 +1,11 @@
+name: 'Install FFmpeg'
+description: 'Installs FFmpeg via Homebrew on macOS'
+runs:
+  using: "composite"
+  steps:
+    - name: Install FFmpeg via Homebrew
+      run: brew install ffmpeg
+      shell: bash
+    - name: Verify FFmpeg installation
+      run: ffmpeg -version
+      shell: bash
--- a/.github/actions/install-system-dependencies/action.yml
+++ b/.github/actions/install-system-dependencies/action.yml
@@ -1,0 +1,22 @@
+name: Install System Dependencies
+description: Install required system packages for ESPnet testing
+
+inputs:
+  packages:
+    description: 'Space-separated list of packages to install'
+    required: false
+    default: 'cmake libsndfile1-dev bc sox ffmpeg libjpeg-dev'
+  enable-cache:
+    description: 'Enable apt cache to speed up installation'
+    required: false
+    default: 'true'
+
+runs:
+  using: "composite"
+  steps:
+
+    - name: Install system packages
+      shell: bash
+      run: |
+        sudo apt-get update -qq
+        sudo apt-get install -qq -y ${{ inputs.packages }}
--- a/.github/actions/prepare-environment/action.yml
+++ b/.github/actions/prepare-environment/action.yml
@@ -1,0 +1,39 @@
+name: Action prepare environment
+description: This action prepares the environment for testing.
+
+inputs:
+  os-version:
+    required: true
+    type: string
+  python-version:
+    required: true
+    type: string
+  pytorch-version:
+    required: true
+    type: string
+  chainer-version:
+    required: true
+    type: string
+  use-conda:
+    required: true
+    type: string
+
+
+runs:
+  using: "composite"
+  steps:
+    - uses: actions/setup-python@v6
+      if: ${{ inputs.os-version == 'ubuntu' }}
+      with:
+        python-version: ${{ inputs.python-version }}
+        architecture: 'x64'
+    # NOTE: System dependencies removed - they're not cached and will be installed in test jobs
+    - name: install espnet
+      shell: bash
+      env:
+        ESPNET_PYTHON_VERSION: ${{ inputs.python-version }}
+        TH_VERSION: ${{ inputs.pytorch-version }}
+        CHAINER_VERSION: ${{ inputs.chainer-version }}
+        USE_CONDA: ${{ inputs.use-conda }}
+      run: |
+        ./ci/install.sh
--- a/.github/workflows/ci_on_macos.yml
+++ b/.github/workflows/ci_on_macos.yml
@@ -21,14 +21,12 @@ jobs:
         use-conda: [true, false]
     steps:
       - uses: actions/checkout@master
+      - name: Install FFmpeg
+        uses: ./.github/actions/install-ffmpeg
       - uses: actions/cache@v5
         with:
           path: ~/.cache/pip
           key: ${{ runner.os }}-pip-${{ matrix.python-version }}-${{ matrix.pytorch-version }}-${{ hashFiles('**/pyproject.toml') }}-${{ hashFiles('**/Makefile') }}
-      - name: Install FFmpeg via Homebrew
-        run: brew install ffmpeg
-      - name: Run ffmpeg command
-        run: ffmpeg -version
       - uses: actions/setup-python@v6
         with:
           python-version: ${{ matrix.python-version }}
--- a/.github/workflows/ci_on_ubuntu.yml
+++ b/.github/workflows/ci_on_ubuntu.yml
@@ -28,50 +28,68 @@ jobs:
             echo "is_docker=false" >> "$GITHUB_OUTPUT"
           fi
 
-  unit_test_espnet2_and_integration_test_espnet2:
-    runs-on: ${{ matrix.os }}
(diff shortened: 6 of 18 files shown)


espnet2/speechlm/model/speechlm/lm/parallel.py
```diff
@@ -333,7 +410,7 @@ def _loss(self, hidden_states, input_ids, loss_mask):
             # Apply loss masks and compute weighted average
             loss = loss * loss_mask
             count = (loss_mask != 0.0).float()
-            loss = loss.sum() / count.sum()
+            loss = loss.sum() / count[:, :, 0].sum()
```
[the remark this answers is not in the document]
jctian98: This ensures we use frame-wise loss computing: a frame is effective if and only if the first stream is not empty. 
This is safe and is ensured in the preprocessing stage.

espnet2/speechlm/model/speechlm/lm/parallel.py
```diff
@@ -86,8 +90,9 @@ def from_pretrained(
                 embed_dim = model.config.hidden_size
                 new_embed_tokens = nn.Embedding(vocab_size, embed_dim, padding_idx=0)
                 new_lm_head = nn.Linear(embed_dim, vocab_size, bias=False)
-                new_embed_tokens.weight[0] = 0.0
-                new_lm_head.weight[0] = 0.0
+
+                nn.init.normal_(new_embed_tokens.weight, mean=0.0, std=0.02)
+                nn.init.normal_(new_lm_head.weight, mean=0.0, std=0.02)
```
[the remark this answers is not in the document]
jctian98: Setting padding tokens to 0 would empirically make the training crash for Qwen3. So we can keep the current setup.

espnet2/speechlm/model/speechlm/lm/parallel.py
```diff
@@ -183,19 +208,15 @@ def forward(self, **kwargs):
                     - logits (inference mode without loss_mask)
             """
             input_ids = kwargs["seqs"]
-            conti_feats = kwargs["conti_feats"]
-            loss_mask = kwargs.get("loss_masks", None)
+            loss_mask = kwargs.get("loss_masks")
```
Subohao, line 212: What if get function fails? Is a default value needed here?

espnet2/speechlm/model/speechlm/speechlm_job.py
```diff
@@ -383,11 +445,34 @@ def _apply_chat_template(self, task, data_dict):
...
+                elif msg[1] == "audio":
+                    # User/system use input audio IO, assistant uses output audio IO
+                    if msg[0] == "user" or msg[0] == "system":
+                        this_io = self.audio_input
+                    else:
+                        this_io = self.audio_output
+                else:
+                    raise ValueError(f"Not supported modality in dialogue: {msg[1]}")
+
+                msg = (msg[0], this_io, msg[2])
```
Qingzheng-Wang, line 464: Just a question: how are modality names structured in dialogue data (or other task formats)? 

I mean, is the data structured like:
```
[["user", "audio", audio_path], ["user", "text", text_content], ["assistant", "text", text_content]]? 
```

Or would it be better to use indexed names like:
```
[["user", "audio1", audio_path], ["user", "text1", text_content], ["assistant", "text1", text_content]]
```
with numbers, so we can handle multiple inputs of the same modality separately? 

For example, with 2 audio files:
```
[["user", "audio1", audio_path], ["user", "audio2", audio_path], ["user", "text1", text_content], ["assistant", "text1", text_content]]
```

espnet2/speechlm/model/speechlm/lm/parallel.py
```diff
@@ -48,8 +49,11 @@ def from_pretrained(
             cls,
             pretrained_model_name_or_path,
             multimodal_io,
+            vocab,
             vocab_intervals,
             max_loss_interval: int = 13192,
```
sw005320, line 54: What is 13192?
Why is it embedded?
It is better to be changed as an option

espnet2/speechlm/model/speechlm/lm/parallel.py
sw005320: ditto
Please avoid embedding the hyperparameters
At least, they are initialized at a class instance.

espnet2/speechlm/model/speechlm/lm/parallel.py
```diff
@@ -233,32 +250,92 @@ def _embed(self, input_ids, conti_feats):
...
 
             # (2) Convert tokens to embeddings and sum across streams
-            input_embeds = self.model.embed_tokens(input_ids).sum(dim=2)
+            # NOTE(Jinchuan): Padding tokens in stream > 0 are zeroed out.
+            # Cannot do the same for stream 0 in Qwen3 for numerical stability.
+            input_embeds = self.model.embed_tokens(input_ids)
+            input_embeds[..., 1:, :] = torch.where(
+                (input_ids[..., 1:] == 0).unsqueeze(-1), 0.0, input_embeds[..., 1:, :]
+            )
+            input_embeds = input_embeds.sum(dim=2)
```
sw005320, line 285: Please make sure that this goes through the unit test.
It must be buggy

espnet2/speechlm/model/speechlm/lm/parallel.py
```diff
@@ -348,4 +425,282 @@ def _loss(self, hidden_states, input_ids, loss_mask):
...
+                    decoded_sequences[0][-1, 0] != self.eot_token_id
+                    and num_msg >= len(enforce_modalities) - 1
+                ):
+                    break  # decode next segment only when ending with <|eot|>
+
+                num_msg += 1
+
+            return messages, cache
+
+        def inference_segment(
```
sw005320, line 526: Please add comments about `inference` and `inference_segment` methods.
How are they different?

espnet2/speechlm/model/speechlm/lm/parallel.py
```diff
@@ -348,4 +425,282 @@ def _loss(self, hidden_states, input_ids, loss_mask):
 
             return loss, stats
 
+        # Below are all inference logics
+        @torch.no_grad()
```
sw005320: Is this inference used only for validation, or is it used for inference in general?
Can you correctly specify the training and inference modes?
I can only see `@torch.no_grad()`

Please make sure that the evaluation mode is correctly specified (e.g., dropout, padding, (model averaging?) etc. in addition to @torch.no_grad())

espnet2/speechlm/model/speechlm/lm/parallel.py
```diff
@@ -348,4 +425,282 @@ def _loss(self, hidden_states, input_ids, loss_mask):
 
             return loss, stats
 
+        # Below are all inference logics
```
sw005320, line 429: Please make sure to add comments for each method for inference

espnet2/speechlm/model/speechlm/speechlm_job.py
```diff
@@ -71,18 +72,18 @@ def _build_vocabulary(self, num_special_tokens=256):
...
+            "<|eos|>",  # 2
+            "<|eot|>",  # 3
+            "<|system|>",  # 4
+            "<|user|>",  # 5
+            "<|assistant|>",  # 6
+            "<|text|>",  # 7
+            "<|audio|>",  # 8
+            "<|image|>",  # 9
+            "<|video|>",  # 10
+            "<|toolcall|>",  # 11
```
sw005320: What does `# number` mean?
Please also make sure to add the comments (definitions) of each special token

espnet2/speechlm/model/speechlm/lm/parallel.py
jctian98: loss_masks are always included in preprocessor for now

espnet2/speechlm/model/speechlm/speechlm_job.py
jctian98: dialogue data doesn't contain "modality names", it looks like:
`[role, modality, content]`
where modality is `text` or `audio` or others. 

You may see `audio1` or `text2` from other tasks, where all `text*` means text modality and so on. But dialogue data is an exception for maximum flexibility.

espnet2/speechlm/model/speechlm/lm/parallel.py
jctian98: This is an empirical value, based on our previous tokenizer setup: 5000 `SSL` code + 8 * 1024 `codec` code.

This is just a default value that can be overridden.

For the stream > 0 in multi-stream loss computing, loss is only computed on a sub-interval of the whole vocabulary (as the whole vocabulary is too large and expensive). This `max_loss_interval` specifies the maximum consecutive sub-intervals we can merge during loss computing. 

In practice, this setup would only impact the efficiency, not the correctness. And the 13k value here should be able to cover most audio tokens in one merged interval.

espnet2/speechlm/model/speechlm/lm/parallel.py
jctian98: removed, not necessary

espnet2/speechlm/model/speechlm/lm/parallel.py
jctian98: add docstring for all inference functions

espnet2/speechlm/model/speechlm/lm/parallel.py
jctian98: The inference function is only used for inference, not validation.
The model is set to `eval()` mode when initialized in `inference.py`, so it should be valid (especially for dropout and paddings, if used)
We don't use model averaging so far; if needed, that should be done offline before launching the inference.

espnet2/speechlm/model/speechlm/speechlm_job.py
jctian98: number is the token id; add extra explanation for each special token

espnet2/speechlm/model/speechlm/lm/parallel.py
sw005320: In this case, why you add `@torch.no_grad()`?

espnet2/speechlm/model/speechlm/lm/parallel.py
sw005320: You'd at least need to provide such a comment.
Please make sure to take care of the other people who will use this code.

sw005320: @jctian98, please reflect on the comments

sw005320, line 54:
```suggestion
vocab_intervals,
max_loss_interval: int = 13192,
# 13192 is an empirical value, based on our previous tokenizer setup: 
# 5000 SSL code + 8 * 1024 codec code.
# This is just a default value that can be overridden.
# See https://github.com/espnet/espnet/pull/6354/changes#r2772187777
```

sw005320:
Once the CI is passed, I'll merge this PR.

jctian98:
Thanks! I'll wait this PR to be merged

jctian98:
maybe ok to merge? @sw005320

sw005320:
Thanks!

PR #6356, opened 2026-02-02
Fix Whisper tokenizer to use extra_special_tokens for Transformers v5 compatibility

Masao-Someki:
## What did you change?

Updated the Whisper tokenizer logic to use `extra_special_tokens` instead of `additional_special_tokens` when reading/adding special tokens.

---

Transformers v5 removed `tokenization_utils_fast.py` and unified the tokenizer APIs. espnet/whisper still depended on older GPT2Tokenizer attributes (`additional_special_tokens`), which caused errors with the new API. Switching to `extra_special_tokens` fixes compatibility with v5.

---

## Is your PR small enough?

yes. 2 files, 4 add/deletions

---

## Additional Context

https://github.com/huggingface/transformers/blob/main/MIGRATION_GUIDE_V5.md

### DIFF
--- a/espnet2/ps2st/espnet_model.py
+++ b/espnet2/ps2st/espnet_model.py
@@ -16,6 +16,7 @@ try:
         AutoProcessor,
         Qwen2AudioForConditionalGeneration,
     )
+
     is_transformers_available = True
 except ImportError:
     is_transformers_available = False


sw005320:
Thanks!

How about considering the suggestions from the Gemini code reviews?

sw005320:
Thanks.
I did not fully understand it, but `self.tokenizer.tokenizer.extra_special_tokens` seems to be omitted with this change.
Is it OK?
Did you confirm by throwing some examples? (We may add a test about it)

Masao-Someki:
Thank you!

Tokens in `extra_special_tokens` are ignored inside the `add_special_tokens` function, so we can remove it from the original code.
In this code, tokens that already exist in `extra_special_tokens` are filtered out, so this part can be shortened from the beginning.
https://github.com/huggingface/transformers/blob/56c4b5e37f92b4ee8716beca3b26dee5c5e63dc5/src/transformers/tokenization_utils_base.py#L1176-L1184

I also found that this fix only supports `transformers >= 5.0.0`, so I recovered `additional_special_tokens` to maintain support for `transformers < 5.0.0` as well.

Masao-Someki:
(It appears that GitHub Actions is currently experiencing an incident. I'll re-run it once the service has recovered.)

PR #6363, opened 2026-02-11
fix: raise NotImplementedError and avoid mutable default args

Mr-Neutr0n:
## Summary

- **Fix `return NotImplementedError` -> `raise NotImplementedError`** in 5 abstract base classes. Using `return` silently returns the exception class object as a value instead of raising it, so subclasses that forget to override the `name` property would not get an error at runtime.

  Affected files:
  - `espnet2/s2st/losses/abs_loss.py`
  - `espnet2/s2st/aux_attention/abs_aux_attention.py`
  - `espnet2/enh/loss/criterions/abs_loss.py`
  - `espnet2/asvspoof/loss/abs_loss.py`
  - `espnet2/uasr/loss/abs_loss.py`

- **Fix mutable default argument `others={}`** -> `others=None` with a `None` guard in 4 enhancement loss wrapper `forward()` methods. Mutable default arguments in Python are shared across all calls, which can lead to subtle bugs if the dict is ever mutated.

  Affected files:
  - `espnet2/enh/loss/wrappers/pit_solver.py`
  - `espnet2/enh/loss/wrappers/dpcl_solver.py`
  - `espnet2/enh/loss/wrappers/multilayer_pit_solver.py`
  - `espnet2/enh/loss/wrappers/fixed_order.py`

## Test plan

- [x] All changes are minimal and behavior-preserving for correct usage
- [ ] Existing unit tests pass (`pytest test/espnet2/enh/loss/`)
- [ ] No new dependencies introduced


sw005320:
@Mr-Neutr0n,
What do you think of Gemini's reviews?

Mr-Neutr0n:
Thanks for the nudge on this one, Gemini had a fair point.

Fixed the Liskov Substitution violation in `AbsLossWrapper` — base class now declares `others: Optional[Dict] = None`, matching what all four subclasses (`DPCLSolver`, `FixedOrderSolver`, `MultiLayerPITSolver`, `PITSolver`) actually default to. Pushed as `de0f6441`.

The other Liskov point Gemini raised — `MultiLayerPITSolver.forward(self, ref, infs, others=None)` taking `infs` (plural) vs the base class `inf: List` — is a real signature divergence, but it's a deeper refactor (multi-layer PIT legitimately needs a list-of-lists input) and felt out of scope for the LSP-cleanup part of this PR. Happy to do it as a follow-up if the maintainers agree, but didn't want to bundle a behavior-preserving refactor with the typing cleanup.

All 36 wrapper tests pass locally (`test/espnet2/enh/loss/wrappers/`). CI should turn green on the next push.

Mr-Neutr0n:
@sw005320 following up on this. I addressed the Gemini review back in June: `AbsLossWrapper` now declares `others: Optional[Dict] = None` in the base class so it matches what `DPCLSolver`, `FixedOrderSolver`, `MultiLayerPITSolver` and `PITSolver` all actually default to, which was the Liskov violation it flagged.

Is there anything else you'd like changed before this can go in?

sw005320 (approved):

sw005320:
Thanks, @Mr-Neutr0n!

Mr-Neutr0n:
my pleasure @sw005320

Mr-Neutr0n:
Thanks for merging @sw005320, and for the review earlier in the year that caught the Liskov issue. Appreciate you picking it back up after the gap.

PR #6392, opened 2026-03-17
[espnet3-14.3] Add integration test for ASR

Masao-Someki:
## What did you change?

Added ASR integration test

---

To test espnet3 pipeline

---

## Is your PR small enough?

No, 28 files change and 2,069 addition.

Except for the following edits, the reviews are done in #6331.

- Please also add comments to the config like the that in the template directory

---

## Additional Context

* #6331
* #6390
* #6391


egs3/mini_an4/asr/conf/inference.yaml
```diff
@@ -0,0 +1,34 @@
...
+#   python run.py --stages infer --inference_config conf/inference.yaml
+#
+# Mini AN4 ASR inference config for ESPnet3 integration tests.
+
+##########################################################
+#                    PATHS AND INPUTS                    #
+##########################################################
+exp_tag: ${load_yaml:training.yaml,exp_tag}
+inference_dir: ${exp_dir}/inference
+dataset_dir: ${data_dir}/mini_an4
```
sw005320: This would be an issue as we need to specify the dataset name `mini_an4`, and it would cause some errors.
How about creating the `dataset_dir` under `egs3/mini_anr` so that we don't need to specify it?

Is there another way to avoid it?

egs3/mini_an4/asr/conf/training_asr_rnn.yaml
```diff
@@ -0,0 +1,149 @@
...
+dataset_dir: ${data_dir}/mini_an4
+
+##########################################################
+#                 DATASET CREATION SETTING               #
+##########################################################
+# Reuse the downloaded mini_an4 archive to create local manifests.
+create_dataset:
+  func: src.creating_dataset.create_dataset
+  dataset_dir: ${dataset_dir}
+  archive_path: ${recipe_dir}/../../egs2/mini_an4/asr1/downloads.tar.gz
```
sw005320: Ditto
Instead of specifying the data in egs2, it is better to put it under the recipe directory.

Also, please avoid using egs2 files.
Make espnet3 recipes independent as much as possible.

egs3/mini_an4/asr/conf/training_asr_rnn.yaml
```diff
@@ -0,0 +1,149 @@
...
+# This is a debug config for CI.
+# Mini AN4 ASR training config for ESPnet3 integration tests.
+
+##########################################################
+#                  SYSTEM AND PATH SETTING               #
+##########################################################
+task: espnet3.systems.asr.task.ASRTask
+
+exp_tag: train_asr_rnn_debug
+dataset_dir: ${data_dir}/mini_an4
```
sw005320: ditto

egs3/mini_an4/asr/conf/training_asr_rnn.yaml
```diff
@@ -0,0 +1,149 @@
...
+##########################################################
+# This is a debug config for CI.
+# Mini AN4 ASR training config for ESPnet3 integration tests.
+
+##########################################################
+#                  SYSTEM AND PATH SETTING               #
+##########################################################
+task: espnet3.systems.asr.task.ASRTask
+
+exp_tag: train_asr_rnn_debug
```
sw005320: Why `_debug`?
Please remove unnecessary names and avoid confusions

egs3/mini_an4/asr/conf/training_asr_rnn.yaml
```diff
@@ -0,0 +1,149 @@
...
+
+##########################################################
+#                    DATASET DEFINITION                  #
+##########################################################
+# Use the tiny train/dev split prepared for CI-sized ASR runs.
+dataset:
+  train:
+    - name: train_nodev
+      dataset:
+        _target_: src.dataset.MiniAN4Dataset
```
sw005320: ditto
This would bring some issues as we need to specify the dataset name

egs3/mini_an4/asr/conf/training_asr_rnn.yaml
```diff
@@ -0,0 +1,149 @@
...
+        manifest_path: ${dataset_dir}/manifest/train_nodev.tsv
+  valid:
+    - name: train_dev
+      dataset:
+        _target_: src.dataset.MiniAN4Dataset
```
sw005320: ditto

egs3/mini_an4/asr/conf/inference_transducer.yaml
```diff
@@ -0,0 +1,32 @@
...
+##########################################################
+exp_dir: ${load_yaml:training.yaml,exp_dir}
+inference_dir: ${exp_dir}/inference
+dataset_dir: ${load_yaml:training.yaml,dataset_dir}
+
+dataset:
+  test:
+    - name: test
+      dataset:
+        _target_: src.dataset.MiniAN4Dataset
```
sw005320: This would bring some issues, as we need to specify the dataset name

egs3/mini_an4/asr/conf/inference.yaml
```diff
@@ -0,0 +1,34 @@
...
+
+dataset:
+  test:
+    - name: test
+      dataset:
+        _target_: src.dataset.MiniAN4Dataset
```
sw005320: ditto
This would bring some issues as we need to specify the dataset name

egs3/mini_an4/asr/conf/training_asr_rnn_data_aug.yaml
```diff
@@ -0,0 +1,146 @@
...
+# This is a debug config for CI
+# Mini AN4 ASR training config for ESPnet3 integration tests.
+
+##########################################################
+#                  SYSTEM AND PATH SETTING               #
+##########################################################
+task: espnet3.systems.asr.task.ASRTask
+
+exp_tag: train_asr_rnn_data_aug_debug
+dataset_dir: ${data_dir}/mini_an4
```
sw005320: ditto

egs3/mini_an4/asr/conf/training_asr_rnn_data_aug.yaml
```diff
@@ -0,0 +1,146 @@
...
+#
+# This is a debug config for CI
+# Mini AN4 ASR training config for ESPnet3 integration tests.
+
+##########################################################
+#                  SYSTEM AND PATH SETTING               #
+##########################################################
+task: espnet3.systems.asr.task.ASRTask
+
+exp_tag: train_asr_rnn_data_aug_debug
```
sw005320: ditto.
why "debug"?

egs3/mini_an4/asr/conf/training_asr_rnn_data_aug.yaml
```diff
@@ -0,0 +1,146 @@
...
+dataset_dir: ${data_dir}/mini_an4
+
+##########################################################
+#                 DATASET CREATION SETTING               #
+##########################################################
+create_dataset:
+  func: src.creating_dataset.create_dataset
+  dataset_dir: ${dataset_dir}
+  archive_path: ${recipe_dir}/../../egs2/mini_an4/asr1/downloads.tar.gz
```
sw005320: ditto

egs3/mini_an4/asr/conf/training_asr_rnn_data_aug.yaml
```diff
@@ -0,0 +1,146 @@
...
+
+##########################################################
+#                    DATASET DEFINITION                  #
+##########################################################
+dataset:
+  train:
+    - name: train_nodev
+      dataset:
+        _target_: src.dataset.MiniAN4Dataset
```
sw005320: ditto

egs3/mini_an4/asr/conf/training_asr_rnn_data_aug.yaml
```diff
@@ -0,0 +1,146 @@
...
+        manifest_path: ${dataset_dir}/manifest/train_nodev.tsv
+  valid:
+    - name: train_dev
+      dataset:
+        _target_: src.dataset.MiniAN4Dataset
```
sw005320: ditto

egs3/mini_an4/asr/conf/training_asr_streaming.yaml
```diff
@@ -0,0 +1,141 @@
...
+##########################################################
+# This is a debug config for CI.
+# Mini AN4 streaming ASR training config for ESPnet3 integration tests.
+
+##########################################################
+#                  SYSTEM AND PATH SETTING               #
+##########################################################
+task: espnet3.systems.asr.task.ASRTask
+
+exp_tag: train_asr_streaming_debug
```
sw005320: ditto

egs3/mini_an4/asr/conf/training_asr_streaming.yaml
```diff
@@ -0,0 +1,141 @@
...
+dataset_dir: ${data_dir}/mini_an4
```
sw005320: ditto

egs3/mini_an4/asr/conf/training_asr_streaming.yaml
```diff
@@ -0,0 +1,141 @@
...
+
+##########################################################
+#                 DATASET CREATION SETTING               #
+##########################################################
+create_dataset:
+  func: src.creating_dataset.create_dataset
+  dataset_dir: ${dataset_dir}
+  archive_path: ${recipe_dir}/../../egs2/mini_an4/asr1/downloads.tar.gz
```
sw005320: ditto

egs3/mini_an4/asr/conf/training_asr_streaming.yaml
```diff
@@ -0,0 +1,141 @@
...
+
+##########################################################
+#                    DATASET DEFINITION                  #
+##########################################################
+# Standard mini_an4 train/dev manifests plus BPE preprocessing.
+dataset:
+  train:
+    - name: train_nodev
+      dataset:
+        _target_: src.dataset.MiniAN4Dataset
```
sw005320: ditto

egs3/mini_an4/asr/conf/training_asr_streaming.yaml
```diff
@@ -0,0 +1,141 @@
...
+        manifest_path: ${dataset_dir}/manifest/train_nodev.tsv
+  valid:
+    - name: train_dev
+      dataset:
+        _target_: src.dataset.MiniAN4Dataset
```
sw005320: ditto

egs3/mini_an4/asr/conf/training_asr_transducer.yaml
```diff
@@ -0,0 +1,143 @@
...
+##########################################################
+# This is a debug config for CI.
+# Mini AN4 transducer ASR training config for ESPnet3 integration tests.
+
+##########################################################
+#                  SYSTEM AND PATH SETTING               #
+##########################################################
+task: espnet3.systems.asr.task.ASRTask
+
+exp_tag: train_asr_transducer_debug
```
sw005320: ditto

egs3/mini_an4/asr/conf/training_asr_transducer.yaml
```diff
@@ -0,0 +1,143 @@
...
+dataset_dir: ${data_dir}/mini_an4
```
sw005320: ditto

egs3/mini_an4/asr/conf/training_asr_transducer.yaml
```diff
@@ -0,0 +1,143 @@
...
+
+##########################################################
+#                 DATASET CREATION SETTING               #
+##########################################################
+create_dataset:
+  func: src.creating_dataset.create_dataset
+  dataset_dir: ${dataset_dir}
+  archive_path: ${recipe_dir}/../../egs2/mini_an4/asr1/downloads.tar.gz
```
sw005320: ditto

egs3/mini_an4/asr/conf/training_asr_transducer.yaml
```diff
@@ -0,0 +1,143 @@
...
+
+##########################################################
+#                    DATASET DEFINITION                  #
+##########################################################
+# Share the same tiny manifests across the ASR debug configs.
+dataset:
+  train:
+    - name: train_nodev
+      dataset:
+        _target_: src.dataset.MiniAN4Dataset
```
sw005320: ditto

egs3/mini_an4/asr/conf/training_asr_transducer.yaml
```diff
@@ -0,0 +1,143 @@
...
+        manifest_path: ${dataset_dir}/manifest/train_nodev.tsv
+  valid:
+    - name: train_dev
+      dataset:
+        _target_: src.dataset.MiniAN4Dataset
```
sw005320: ditto

egs3/mini_an4/asr/conf/training_asr_transformer.yaml
```diff
@@ -0,0 +1,141 @@
...
+##########################################################
+# This is a debug config for CI.
+# Mini AN4 transformer ASR training config for ESPnet3 integration tests.
+
+##########################################################
+#                  SYSTEM AND PATH SETTING               #
+##########################################################
+task: espnet3.systems.asr.task.ASRTask
+
+exp_tag: train_asr_transformer_debug
```
sw005320: ditto

egs3/mini_an4/asr/conf/training_asr_transformer.yaml
```diff
@@ -0,0 +1,141 @@
...
+dataset_dir: ${data_dir}/mini_an4
```
sw005320: ditto

egs3/mini_an4/asr/conf/training_asr_transformer.yaml
```diff
@@ -0,0 +1,141 @@
...
+
+##########################################################
+#                 DATASET CREATION SETTING               #
+##########################################################
+create_dataset:
+  func: src.creating_dataset.create_dataset
+  dataset_dir: ${dataset_dir}
+  archive_path: ${recipe_dir}/../../egs2/mini_an4/asr1/downloads.tar.gz
```
sw005320: ditto

egs3/mini_an4/asr/conf/training_asr_transformer.yaml
```diff
@@ -0,0 +1,141 @@
...
+
+##########################################################
+#                    DATASET DEFINITION                  #
+##########################################################
+# Minimal train/dev split for transformer smoke tests.
+dataset:
+  train:
+    - name: train_nodev
+      dataset:
+        _target_: src.dataset.MiniAN4Dataset
```
sw005320: ditto

egs3/mini_an4/asr/conf/training_asr_transformer.yaml
```diff
@@ -0,0 +1,141 @@
...
+        manifest_path: ${dataset_dir}/manifest/train_nodev.tsv
+  valid:
+    - name: train_dev
+      dataset:
+        _target_: src.dataset.MiniAN4Dataset
```
sw005320: ditto

egs3/mini_an4/asr/conf/training_lm_rnn.yaml
```diff
@@ -0,0 +1,125 @@
...
+# This is a debug config for CI.
+# Mini AN4 LM-RNN training config for ESPnet3 integration tests.
+
+##########################################################
+#                  SYSTEM AND PATH SETTING               #
+##########################################################
+task: espnet3.systems.asr.task.ASRTask
+
+exp_tag: train_lm_rnn_debug
+dataset_dir: ${data_dir}/mini_an4
```
sw005320: ditto

egs3/mini_an4/asr/conf/training_lm_rnn.yaml
```diff
@@ -0,0 +1,125 @@
...
+
+##########################################################
+#                 DATASET CREATION SETTING               #
+##########################################################
+create_dataset:
+  func: src.creating_dataset.create_dataset
+  dataset_dir: ${dataset_dir}
+  archive_path: ${recipe_dir}/../../egs2/mini_an4/asr1/downloads.tar.gz
```
sw005320: ditto

egs3/mini_an4/asr/conf/training_lm_rnn.yaml
```diff
@@ -0,0 +1,125 @@
...
+
+##########################################################
+#                    DATASET DEFINITION                  #
+##########################################################
+# Reuse the ASR manifests and tokenize the transcript text for LM training.
+dataset:
+  train:
+    - name: train_nodev
+      dataset:
+        _target_: src.dataset.MiniAN4Dataset
```
sw005320: ditto

egs3/mini_an4/asr/conf/training_lm_rnn.yaml
```diff
@@ -0,0 +1,125 @@
...
+        manifest_path: ${dataset_dir}/manifest/train_nodev.tsv
+  valid:
+    - name: train_dev
+      dataset:
+        _target_: src.dataset.MiniAN4Dataset
```
sw005320: ditto

egs3/mini_an4/asr/conf/training_transducer_asr_conformer_rnnt.yaml
```diff
@@ -0,0 +1,126 @@
...
+##########################################################
+#                  SYSTEM AND PATH SETTING               #
+##########################################################
+num_device: 1
+num_nodes: 1
+task: espnet3.systems.asr.transducer_task.ASRTransducerTask
+
+recipe_dir: .
+data_dir: ${recipe_dir}/data
+exp_tag: train_transducer_asr_conformer_rnnt_debug
```
sw005320: ditto

egs3/mini_an4/asr/conf/training_transducer_asr_conformer_rnnt.yaml
```diff
@@ -0,0 +1,126 @@
...
+exp_dir: ${recipe_dir}/exp/${exp_tag}
+stats_dir: ${recipe_dir}/exp/stats
+inference_dir: ${exp_dir}/inference
+dataset_dir: ${data_dir}/mini_an4
```
sw005320: ditto

egs3/mini_an4/asr/conf/training_transducer_asr_conformer_rnnt.yaml
```diff
@@ -0,0 +1,126 @@
...
+
+##########################################################
+#                 DATASET CREATION SETTING               #
+##########################################################
+create_dataset:
+  func: src.creating_dataset.create_dataset
+  dataset_dir: ${dataset_dir}
+  archive_path: ${recipe_dir}/../../egs2/mini_an4/asr1/downloads.tar.gz
```
sw005320: ditto

egs3/mini_an4/asr/conf/training_transducer_asr_conformer_rnnt.yaml
```diff
@@ -0,0 +1,126 @@
...
+
+##########################################################
+#                    DATASET DEFINITION                  #
+##########################################################
+# Use the same manifest layout as the other mini_an4 debug configs.
+dataset:
+  train:
+    - name: train_nodev
+      dataset:
+        _target_: src.dataset.MiniAN4Dataset
```
sw005320: ditto

egs3/mini_an4/asr/conf/training_transducer_asr_conformer_rnnt.yaml
```diff
@@ -0,0 +1,126 @@
...
+        manifest_path: ${dataset_dir}/manifest/train_nodev.tsv
+  valid:
+    - name: train_dev
+      dataset:
+        _target_: src.dataset.MiniAN4Dataset
```
sw005320: ditto

egs3/mini_an4/asr/src/creating_dataset.py
```diff
@@ -0,0 +1,154 @@
...
+    with path.open("w", encoding="utf-8") as f:
+        for e in entries:
+            f.write(f"{e.utt_id}\t{e.wav_path}\t{e.text}\n")
+
+
+def create_dataset(dataset_dir: Path, archive_path: Path | None = None) -> None:
+    dataset_dir = Path(dataset_dir)
+    logger = setup_logger(name="mini_an4.create_dataset")
+
+    archive = Path("../../../egs2/mini_an4/asr1/downloads.tar.gz")
```
sw005320: avoid to specify the espnet2 directory

egs3/mini_an4/asr/src/tokenizer.py
sw005320, line 1: This seems to be dataset independent.
It would be better to move this to the recipe common place (e.g., template)

sw005320 (commented):
There are many dataset-dependent names or paths in the config file.
Please avoid them as much as possible.
You should make them under the recipe directory and avoid specifying the dataset names.

I think the tokenizer part is well-designed.
Please follow this way

sw005320:
Can you even remove `mini_an4/asr` from the config?
Since it will be under the `mini_an4/asr` recipe directory, we can extract this path information from the current directory information.

egs3/mini_an4/asr/dataset/__init__.py
```diff
@@ -0,0 +1,6 @@
+"""Mini AN4 dataset module."""
+
+from egs3.mini_an4.asr.dataset.builder import MiniAN4Builder as DatasetBuilder
```
Masao-Someki: I think we can just do `from .dataset.builder import ...`
Let me check

sw005320:
Yeah, let's try to minimize the dataset-dependent descriptions (but we should have some of them in the dataset/ directory).
We can make it as a policy.

sw005320:
This is much better.
Another question. Why do we need to declare `test` twice?
```  
  test:
    - name: test
      kwargs:
        split: test
```

sw005320:
What does `kwargs ` mean?

sw005320:
You can also add some comments about the train/valid/test data specifications in the config file.

egs3/mini_an4/asr/conf/inference.yaml
```diff
@@ -0,0 +1,32 @@
+##########################################################
+#                   Inference SETTING                    #
+##########################################################
+# Quick usage:
+#   python run.py --stages infer --inference_config conf/inference.yaml
+#
+# Mini AN4 ASR inference config for ESPnet3 integration tests.
```
sw005320: You can remove this line

egs3/mini_an4/asr/dataset/dataset.py
```diff
@@ -0,0 +1,118 @@
...
+
+    return entries
+
+
+# ---------------------------------------------------------------------------
+# Public dataset class
+# ---------------------------------------------------------------------------
+
+
+class MiniAN4Dataset(TorchDataset):
```
sw005320: Can you reconsider avoiding the mixed case here?
This could be a source of ambiguity in the class name, since the dataset name is tricky.
Then, the code becomes buggy
How about making them all upper or lower to avoid it?

sw005320: Another ambiguity is `_` 
BTW, the recipe name uses `_` while the class name doesn't use it. So, it is not consistent.

Masao-Someki: How about capitalizing both the first letter and the “D” in “Dataset”?
So we would have datasets like `Minian4Dataset`, `LibrispeechDataset`, `VctkDataset`, `CommonvoiceDataset`, etc.

sw005320: This works for me.
Good idea

sw005320: how about `_`?
we could convert from `mini_an4` to `Minian4` but we would not have the inverse.

Masao-Someki:
> Another question. Why do we need to declare test twice?

I think it is for name and split.
I add the name and split here, so that people can have multiple test sets.

For example, we can run tests on librispeech, vctk, or japanese dataset like reazon speech dev/test, etc
```
  test:
    - name: librispeech_test_clean
      ref: librispeech/asr
      kwargs:
        split: test_clean
    - name: librispeech_test_other
      ref: librispeech/asr
      kwargs:
        split: test_other
    - name: vctk
      ref: vctk/asr
      kwargs:
        split: test
    - name: reazonspeech_dev
      ref: reazonspeech/asr
      kwargs:
        split: dev
    - name: reazonspeech_test
      ref: reazonspeech/asr
      kwargs:
        split: test
```

name is for unique test set ID, split is a kwargs for the Dataset class in the ref.

> What does kwargs mean?

This is for the keyword arguments of the Dataset class under the ref.
E.g., Dataset class for `mini_an4/asr` is defined like this:
```

class MiniAN4Dataset(TorchDataset):
    """Mini AN4 dataset that returns ``{\"speech\", \"text\"}`` samples."""

    def __init__(self, split: str, recipe_dir: str | Path | None = None) -> None:
        self.split = split
        recipe_root = (
            Path(recipe_dir).resolve()
            if recipe_dir is not None
            else Path(__file__).resolve().parents[1]
        )
```

So that we can put `split` and `recipe_dir` in the kwargs.
It will be passed to the dataset class, so it depends on the implementation.

egs3/mini_an4/asr/dataset/dataset.py
Masao-Someki: It is true
Then how about capitalize after `_`?
E.g., `mini_an4/asr` -> `MiniAn4Dataset`, `ofuton_p_utagoe_db/svs` -> `OfutonPUtagoeDbDataset`

This approach makes a lot of sense to me.

sw005320:
I see.
Can you use a better name?
`name`: I understood
`ref`: ???
`kwargs and split`: ???

Masao-Someki:
@sw005320 
How about `data_src` and `data_src_args`?
So for example:
```
  test:
    - name: librispeech_test_clean
      data_src: librispeech/asr
      data_src_args:
        split: test_clean
        data_path: /path/to/data
        ...  # Other args for the Dataset class 
```

sw005320:
> @sw005320 How about `data_src` and `data_src_args`? So for example:
> 
> ```
>   test:
>     - name: librispeech_test_clean
>       data_src: librispeech/asr
>       data_src_args:
>         split: test_clean
>         data_path: /path/to/data
>         ...  # Other args for the Dataset class 
> ```

This is clearer!
Thanks!
@popcornell and @Emrys365, any comment?

Masao-Someki:
Thank you @sw005320, I will update based on this

egs3/mini_an4/asr/conf/training_asr_rnn.yaml
```diff
@@ -0,0 +1,145 @@
...
+
+##########################################################
+#            OPTIMIZER & SCHEDULER SETTING               #
+##########################################################
+optimizer:
+  _target_: torch.optim.AdamW
+  lr: 0.001
+  weight_decay: 0.0
+
+override scheduler:
```
popcornell: this yaml key has a space

sw005320: I see. This is interesting. This means we may use a bit more natural language-based descriptive names

popcornell: IDK if it was intended TBH

sw005320: I see. We can wait for @Masao-Someki's response

sw005320:
@Masao-Someki, please fix the CI error

egs3/mini_an4/asr/conf/inference.yaml
```diff
@@ -0,0 +1,33 @@
...
+    - name: test
+      kwargs:
+        split: test
+
+##########################################################
+#                    MODEL AND OUTPUT                    #
+##########################################################
+model:
+  _target_: espnet2.bin.asr_inference.Speech2Text
+  asr_train_config: ${exp_dir}/config.yaml
```
popcornell, line 41: the ${exp_dir} does not appear anywhere in this file. How is it resolved ? By hydra automatically ?

ci/test_integration_espnet3.sh
```diff
@@ -0,0 +1,54 @@
...
+    local runner=$2
+    local inference_config=$3
+
+    ln -sfn "${training_config}" conf/training.yaml
+    ${python} "${runner}" \
+        --stages create_dataset train_tokenizer collect_stats train infer measure \
+        --training_config conf/training.yaml \
+        --inference_config "${inference_config}" \
+        --metrics_config conf/metrics.yaml
+    rm -rf exp data
```
popcornell, line 33: can you also fix this based on Gemini suggestion ? I think this is wasting time no ? data can be retained

egs3/mini_an4/asr/conf/training_asr_rnn.yaml
Masao-Someki: This one is coming from hydra's config override.
Basically in espnet3 (=hydra+omegaconf), the user-defined configs are overriding the default values, so that

default config
```
scheduler:
  _target_: espnet2.schedulers.warmup_lr.WarmupLR
  warmup_steps: 15000
```

user config:
- If there's no scheduler config -> using the warmup lr.
- If modifying only the warmup steps user can write
```
scheduler:
  warmup_stemps: 5000
```

This is good for many cases, such as trainer configs we don't have to write everything.
However in some cases we want to change the entire scheduler config.

E.g., if user write as
```
scheduler:
  _target_: torch.optim.lr_scheduler.ReduceLROnPlateau
  mode: min
  factor: 0.5
  patience: 1
```

Then the final config would be like this:
```
scheduler:
  _target_: torch.optim.lr_scheduler.ReduceLROnPlateau
  mode: min
  factor: 0.5
  patience: 1
  warmup_stemps: 5000  # <- invalid config
```

One way is to override the scheduler config (as in the current code), or another way is to remove scheduler from default value.

egs3/mini_an4/asr/conf/inference.yaml
Masao-Someki: The configurations are using `egs3/TEMPLATE/asr/conf/*.yaml` as the default value.
So this `${exp_dir}` comes from this line: [exp_dir](https://github.com/Masao-Someki/espnet/blob/45528cb02d93e0941757ff346a5a0565f04f4422/egs3/TEMPLATE/asr/conf/inference.yaml#L16)

This is espnet3's functionality: [load_and_merge_config](https://github.com/Masao-Someki/espnet/blob/45528cb02d93e0941757ff346a5a0565f04f4422/espnet3/utils/config_utils.py#L254)

ci/test_integration_espnet3.sh
Masao-Someki: Hmm, I think it's still fine to keep it. The create-dataset stage (which generates the data/ directory) only takes about 0.07 seconds on GitHub Actions, so I don't think the execution time is really a concern. I left it as is mainly because changing it would add more lines, and the current version feels easier to read?

Masao-Someki:
I found that fixing this CI is not a simple problem, but requires some design-level brain storming..
Let me handle this first

egs3/mini_an4/asr/conf/training_asr_rnn.yaml
popcornell: Yes this is ok. 
It was just odd for me that the key did not have any space

egs3/mini_an4/asr/conf/inference.yaml
popcornell: Okok

ci/test_integration_espnet3.sh
popcornell: Ok !

Masao-Someki:
The change in [c78db55](https://github.com/espnet/espnet/pull/6392/commits/c78db5584553eff3f7735fc8a9a918fb5e86970a) removes the `load_yaml` resolver and stops config files from reaching into sibling configs to fetch `exp_tag` / `exp_dir`.  

Instead, experiment identity is handled at runner setup time:                                                                        
- configs are loaded with `resolve=False`                                                                                            
- when `training_config` is present, `exp_tag` and `exp_dir` are propagated into `inference_config` and `metrics_config`                                                                            
 - when `training_config` is absent, standalone inference/metrics configs must define `exp_tag` or a concrete `exp_dir`, otherwise the runner raises an error                                                                                                               

  ## Motivation

  `load_yaml` looked convenient for reusing values across config files, but it created a weaker config-loading path. 
In practice, that caused a few problems:
  - it read raw YAML files, so some configs defined by `TEMPLATE/**/*.yaml` or with `defaults:`, are not visible from the `load_yaml` resolver.
  - behavior became hard to track once configs started depending on other configs during interpolation

In the CI issue, the transducer failure was a concrete example of this.. 
`inference_transducer.yaml` read `exp_dir` from `training.yaml`, but `load_yaml` evaluated the raw training config without merging the default config, so resolution broke.
Rather than making cross-config reads more complex, this change removes that pattern and keeps experiment identity handling in `run.py`

## What changed

  - removed `load_yaml` from `espnet3/utils/config_utils.py`
  - added runner-side experiment-context helpers in `espnet3/utils/run_utils.py`
  - updated `egs3/TEMPLATE/asr/run.py` to apply and validate experiment context before final resolve
  - removed `load_yaml` usage from mini_an4 inference / transducer inference / metrics configs
  - moved helper coverage into `test/espnet3/utils/test_run_utils.py`
  - updated config-related tests accordingly

Masao-Someki:
@sw005320 If this looks okay to you, could you merge this PR? We can move to the next (3.14.5) bug fix PR!

Masao-Someki:
I have fixed the config name. I will handle other config-related items in the following PRs!

.github/workflows/ci_on_ubuntu.yml
```diff
@@ -275,6 +275,43 @@ jobs:
...
+    needs: process_labels
+    if: |
+      github.event.pull_request.draft == false
+    strategy:
+      max-parallel: 20
+      matrix:
+        python-version: ["3.10", "3.12"]
+        pytorch-version: [2.9.1, 2.10.0, 2.11.0]
+        use-conda: [false]
+        chainer-version: [6.0.0]
```
sw005320: Do you need Chainer?

egs3/mini_an4/asr/conf/inference.yaml
```diff
@@ -0,0 +1,32 @@
...
+##########################################################
+inference_dir: ${exp_dir}/inference
+
+# Dataset specification:
+#   test:
+#     - uses the "test" split defined in the dataset.dataset.Minian4Dataset
+dataset:
+  test:
+    - name: test
+      data_src_args:
```
sw005320, line 29: Do you need `args`?
data_src or data_sources would be good enough?

Also, it is better to add similar explanation comments to the training data

#                    DATASET DEFINITION                  #
# Dataset splits:
#   train:
#     - split="train" from the dataset.dataset.Minian4Dataset
#   valid:
#     - split="valid" from the dataset.dataset.Minian4Dataset
# Notes:
#   - actual data sources are defined via ${recipe_dir}

egs3/mini_an4/asr/conf/inference.yaml
```diff
@@ -0,0 +1,32 @@
...
+#                    PATHS AND INPUTS                    #
+##########################################################
+inference_dir: ${exp_dir}/inference
+
+# Dataset specification:
+#   test:
+#     - uses the "test" split defined in the dataset.dataset.Minian4Dataset
+dataset:
+  test:
+    - name: test
```
sw005320, line 28: I believe this should also include `valid`

egs3/mini_an4/asr/conf/metrics.yaml
```diff
@@ -0,0 +1,22 @@
+##########################################################
+#                 METRIC SCORING SETTING                 #
+##########################################################
+# Quick usage:
+#   python run.py --stages measure --metrics_config conf/metrics.yaml
+#
+##########################################################
+#                    PATHS AND INPUTS                    #
+##########################################################
+inference_dir: ${exp_dir}/inference
```
sw005320, line 10: Where do we specify the data?
Is it included in this directory specification?
I think it should be commented.

egs3/mini_an4/asr/conf/training_asr_rnn.yaml
sw005320, line 1: We can skip RNN related integration test.
This is not used anymore.

egs3/mini_an4/asr/conf/training_asr_rnn_data_aug.yaml
sw005320, line 1: ditto
We can skip RNN related integration test.
This is not used anymore.

sw005320, line 1: In this case, the data_aug part can be moved to the other config

egs3/TEMPLATE/asr/conf/inference.yaml
```diff
@@ -26,16 +33,23 @@ inference_dir: ${exp_dir}/inference
...
+#     - name: librispeech_test_clean
+#       data_src: librispeech/asr
+#       data_src_args:
+#         split: test_clean
+#         data_path: ${dataset_dir}
+#     - name: librispeech_test_other
+#       data_src: librispeech/asr
+#       data_src_args:
+#         split: test_other
+#         data_path: ${dataset_dir}
```
sw005320: The dev sets should be included

espnet3/components/data/dataset_module.py
```diff
@@ -0,0 +1,250 @@
...
+
+This module standardizes how ESPnet3 resolves dataset sources used in configs
+such as:
+
+.. code-block:: yaml
+
+    dataset:
+      train:
+        - data_src: mini_an4/asr
+          data_src_args:
```
sw005320, line 11: These lines look redundant.
Do we really need `data_src_args:`?

egs3/mini_an4/asr/conf/inference.yaml
Masao-Someki: Thank you, and yes, I think it is good to have `data_src_args`.
This would be passed directly to the dataset class. For example, suppose we have a dataset class like:

```
class MyDataset(Dataset):
  def __init__(self, split="train",):
    self.data = self._build_data(split)
  ...
```

Then we can feed the `data_src_args.split` to the `__init__()`.
We could remove `data_src_args`, but then we would need additional code to manually remove (pop) `data_src` from the config, and then feed the rest configs to `__init__()`.

sw005320:
Please check my other comments

.github/workflows/ci_on_ubuntu.yml
Masao-Someki: Thanks for the comment!
`chainer-version` was not actually used in the espnet3's CI workflow, so I removed it in [2e58522](https://github.com/espnet/espnet/pull/6392/commits/2e58522430455d638ceb10113cc8ca430d5db6b5)

Masao-Someki:
Thank you! 

* Chainer is removed for espnet3's CI in [2e58522](https://github.com/espnet/espnet/pull/6392/commits/2e58522430455d638ceb10113cc8ca430d5db6b5)
* I think `data_src_args` should stay as-is..
* I added dataset explanation comments to the training config in [1bbaeb9](https://github.com/espnet/espnet/pull/6392/commits/1bbaeb9d7e5317a1577b062b08aee0f47280f2e3) and [c8a80bf](https://github.com/espnet/espnet/pull/6392/commits/c8a80bf1be9b8c62076322c8bd3e5321de2b5c0c)!

egs3/mini_an4/asr/conf/inference.yaml
sw005320: @Masao-Someki, please check this comment
We usually do not have only the test data.
If it is intentional to reduce the computation cost, that is valid (but I think we do not have so many additions for the computation time)

egs3/TEMPLATE/asr/conf/inference.yaml
sw005320: @Masao-Someki, please check this comment
We usually do not have only the test data.
If it is intentional to reduce the computation cost, that is valid (but I think we do not have so many additions for the computation time)

espnet3/components/data/dataset_module.py
Masao-Someki: This is used to pass kwargs to the dataset class `MiniAn4Dataset`  in `egs3/mini_an4/asr/dataset/dataset.py`.

For `split` in this case, there are two options:
  1. Pass it explicitly via `data_src_args`.
  2. Define a default in the dataset class, for example
     `def __init__(self, split: str = "train", ...)`.

egs3/mini_an4/asr/conf/metrics.yaml
Masao-Someki: The test set is determined by this function. It searches for a directory under `inference_dir`, and the data in that directory is used to compute the metrics.

I added a comment to mention this point in the metrics.yaml

https://github.com/Masao-Someki/espnet/blob/7a3114e329acd8ac60084c082fa34971600b3812/espnet3/systems/base/metric.py#L17-L39

espnet3/utils/run_utils.py
sw005320, line 1: Can you add a test for this?
It looks tricky.

sw005320: If new additions are already covered by the test, that is OK.
Please respond to it.

Masao-Someki: Thank you, I checked and seems negative cases are not enough.
I will create another PR to check and add all unit tests for `espnet3/`.

sw005320:
Thanks!
What is the next?

Masao-Someki:
Thank you @sw005320 !!
Could you check #6428 first? The PR just enhances the unit tests mentioned in the above conversation.
Since all changes are unit tests, I think it will be easier to proceed.

PR #6393, opened 2026-03-17
Add uv.lock for Tracking Security Package issues with Python

Fhrozen:
There is a security issue that cannot be tracked because no lock file is available (https://github.com/espnet/espnet/security/dependabot/21)

I include that file with some observations on the installation and docs.

## What did you change?

This pull request introduces a new, faster installation option for ESPnet using the `uv` and `pixi` tools, along with a supporting script, and adds flexibility for specifying Python and PyTorch versions. The documentation is updated to reflect this new installation method, and the script is implemented with argument parsing and improved environment setup.

**Documentation updates:**

* Added a new "Option E" to the installation instructions in `doc/installation.md`, describing how to use the new `setup_uv.sh` script with `uv` and `pixi` for fast setup, including available options and a step-by-step explanation.

**New installation script and improvements:**

* Added the `tools/setup_uv.sh` script, which automates environment setup with `uv` and `pixi`, supports `--python` and `--torch` options for version selection, and provides a help message. The script installs dependencies, creates a virtual environment, installs the specified versions of PyTorch and torchaudio, and sets up ESPnet in editable mode.
* The script now dynamically uses the specified Python and PyTorch versions when creating the virtual environment and installing packages, instead of hardcoding them. [[1]](diffhunk://#diff-02ccdf959718f2e45fbf3a3482f6dcb8c0f4faf2342bee3c60f0eedca413a612L26-R62) [[2]](diffhunk://#diff-02ccdf959718f2e45fbf3a3482f6dcb8c0f4faf2342bee3c60f0eedca413a612L35-R71)
* Improved the generation of the `activate_python.sh` script to use the correct absolute path for activating the virtual environment.


Fhrozen:
BTW, CI for `uv` and other setup installation process will be prepared on a follow-up PR.

doc/installation.md
sw005320, line 1: Can you revisit the entire document with @Masao-Someki?
We don't need Step 1.
We can also reorder the option for the most frequent use cases.

tools/setup_uv.sh
sw005320, line 1: @Masao-Someki, please review this PR

uv.lock
sw005320, line 1: what's this?
binary file?
do we need this?

Fhrozen: Is not a binary, is a lock file similar to package.lock for npm. I am adding this due the high vulnerability issue raised by the dependabot: `https://github.com/espnet/espnet/security/dependabot/21`

Masao-Someki: @Fhrozen
I don't think adding `uv.lock` to this large project is a good idea.

It's definitely useful for fully reproducing a Python environment with `uv`, and I agree it works great for smaller projects (like a repo that just running an OWSM fine-tuning, if we had one).

However, in this repo we have many users relying on different environments, like Anaconda, venv, and other package managers, and introducing `uv.lock` could end up being confusing for them very much I think...

tools/setup_uv.sh
Masao-Someki: Maybe we can revisit this file later.

One advantage of using pixi, in my opinion, is that we can simply run `pixi shell` (or `eval $(pixi shell-hook)`) to enter the pixi's shell, and it will automatically set everything up.. including dependencies like `python`, `ffmpeg`, etc.
So it might make sense to run `pixi init` and add a step in `activate_python.sh` to enter the pixi shell.

I can open a separate PR to handle this, including updates to the Makefile, installer scripts, and related setup..

Masao-Someki (commented):
Overall, I think the best way to resolve this issue is to identify a version of setuptools that works well with both espnet and dependabot..

uv.lock
Fhrozen: @Masao-Someki, thanks for the details
I was adding it because the Dependabot needs it to check the pip dependencies. It can be ignored for different environments and other package managers. But if it becomes difficult to manage, then I can remove it.

PR #6420, opened 2026-04-17
[Speechlm] Update multimodal IO and dataloader modules

jctian98:
## Summary

This PR contains several improvements to the SpeechLM multimodal IO and dataloader modules.

### Dataloader (`multimodal_loader/`)
- Make `arkive` import optional (`audio_loader.py`, `text_loader.py`) so that the module can be imported without arkive installed; access is deferred to call sites.

### Multimodal IO (`multimodal_io/`)
- **`audio.py` (DiscreteAudioIO)**: Enable `flash_attention_3` for the Xcodec semantic model when available, avoiding slow CuDNN planning under the default SDPA attention. 
- **`audio.py` (ContinuousAudioIO)**: Handle transposed audio input by auto-transposing when `num_channels > num_samples`, and compute `frame_length` using the actual sample rate of the input (rescaling to the model's `sample_rate`) instead of assuming a fixed rate. Drop the now-unused `n_samples` trimming.
- **`text.py`**: Fix return type annotation of `decode_batch` (`List[str]` → `str`) to match the actual return value.
- **`abs_io.py`**: Remove `ABC` inheritance — `Module` already provides the needed base, and the double inheritance caused metaclass friction.

### Bin (`prepare_length_stats.py`)
- Catch `ValueError` from `DataIteratorFactory.build_iter()` in the worker and return an empty stats dict, so a single bad shard no longer kills the whole stat collection.


espnet2/speechlm/model/speechlm/multimodal_io/audio.py
```diff
@@ -189,6 +189,14 @@ def _init_codec(
...
             # checkpoint for general audio is: hf-audio/xcodec-hubert-general
             self.codec_model = XcodecModel.from_pretrained(codec_hf_model_tag)
+            # NOTE(Jinchuan): default SDPA attention may cause slow CuDNN planning.
+            # Use flash_attention_3 if available.
+            from transformers.utils import is_flash_attn_3_available
+
+            if is_flash_attn_3_available():
+                self.codec_model.semantic_model.config._attn_implementation = (
+                    "flash_attention_3"
+                )
```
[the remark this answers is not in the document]
jctian98: seems x-codec doesn't support this option. Will skip it

sw005320:
This PR should use https://github.com/wavlab-speech/omniio (private now).
@wanchichen

jctian98:
Maybe people can review this PR @sw005320 @wanchichen @Masao-Someki @cyhuang-tw 
The data infra `omniio` may be integrated in the future.

jctian98:
CI is fixed. waiting for review and merge

espnet2/speechlm/bin/prepare_length_stats.py
```diff
@@ -89,16 +89,22 @@ def worker(
...
+            world_size=world_size,
+            shuffle=False,
+            sequential_load=True,
+            num_workers=0,
+            collate_fn=lambda x: x[0],
+        ).build_iter()
+    except ValueError as e:
+        logging.getLogger(__name__).warning(
+            f"Worker {rank}: Error building iterator, skipping shard: {e}"
+        )
```
Masao-Someki, line 106: I just wonder what would happen in this case. E.g., if we use 10 GPUs and if one GPU has this error, does it mean that we train with 9 GPUs? Or training will fail in some other place?

Masao-Someki (approved):
Thanks @jctian98, I think there's no issue from the espnet3's perspective for this PR.

espnet2/speechlm/bin/prepare_length_stats.py
jctian98: This is a simple work around for the situation that you are working with `nj=200` but there are only <200 examples in a dev set.

Masao-Someki:
@jctian98 I think you should add a few more unit tests to cover your changes and help pass the Codecov CI check.

sw005320:
> The data infra `omniio` may be integrated in the future.

I understand your point, since @wanchichen isn't responsible.
However, this means that if you keep using an unofficial repo by @wanchichen as a core module, it will bring further disaster. It will also have a safety-related issue. 
`omniio` will be located under our organization, and at least, the other people can help maintain it even if @wanchichen is not available

@wanchichen, can you prioritize the development and the review of the speechlm PRs?
You have not responded to or reviewed this PR for more than a week, despite our frequent requests...

wanchichen:
The code lgtm. For the stuff related to Omni IO / arrive, you can swap to the implementation here: https://github.com/wavlab-speech/omniio/blob/main/omniio/interface.py

The API is exactly the same so it should be a 1-1 drop in replacement.

sw005320:
@jctian98, since @wanchichen updates the repo, please use it and finish this PR.

jctian98:
Is it normal that I label this PR as "queued for merge" and it merges the PR automatically? @sw005320 @Masao-Someki 

But anyway I've fixed the issue of code merge after this PR.

PR #6421, opened 2026-04-17
[speechlm] update modeling files

jctian98:
## Summary
- Add `parallel_pp.py`: pipeline-parallel multimodal LLM that extends
  `ParallelLLM` with stage-wise construction and forward. Each PP rank
  loads only its local transformer layers plus the peripherals needed
  for first / middle / last stages; registered as the `"parallel_pp"`
  LM in `SpeechLMJobTemplate`.
- Add `loss.py`: fused linear cross-entropy via Liger Kernel, avoiding
  materialization of the `[B*T, V]` logits tensor (~19GB saved for
  32k tokens × 150k vocab). Two calls handle stream 0 (full vocab) and
  streams 1+ (multimodal vocab range) with `reduction="sum"` to work
  around Liger's `reduction="none"` backward bug.
- Rework `parallel.py` to share embed / loss / stats machinery with the
  PP subclass and shed code that has moved into the new files.
- Extend `speechlm_job.py` vocabulary builder to produce per-class CE
  weights alongside the existing vocab/intervals, and wire in the new
  `parallel_pp` backend.


espnet2/speechlm/model/speechlm/lm/loss.py
```diff
@@ -0,0 +1,144 @@
...
+            None,  # softcap
+            True,  # return_z_loss
+            torch.float32,  # accum_dtype
+            False,  # use_token_scaling
+            True,  # return_token_accuracy
+        )
+
+    # ---- Combine ----
+    count = (loss_mask[:, :, 0] != 0).float().sum()
+    ce_loss = (s0_loss + mm_loss).float()
```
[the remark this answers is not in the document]
jctian98: This is intentional; the count is per frame, not per token. each frame can have multiple tokens.

espnet2/speechlm/model/speechlm/lm/parallel.py
```diff
@@ -49,95 +51,91 @@ def from_pretrained(
...
+                vocab_meta["mm_end"],
+            )
+            model.register_buffer("vocab_weight", vocab_meta["vocab_weight"])
+
+            model.vocab = vocab_meta["vocab"]
+            model.vocab_intervals = vocab_meta["vocab_intervals"]
+
             # (4) Setup multimodal IO handlers and adaptors
             # Discrete IOs use vocabulary, continuous IOs need linear adaptors
             model.multimodal_io_dict = nn.ModuleDict(multimodal_io)
```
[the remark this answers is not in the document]
jctian98: it's moved to another places, don't need to be here anymore.

espnet2/speechlm/model/speechlm/lm/parallel_pp.py
```diff
@@ -0,0 +1,511 @@
...
+                    hidden_states = layer_output[0]
+                else:
+                    hidden_states = layer_output
+
+                # HF decoder layers discard router_logits from the MoE
+                # block's return value. Retrieve them from the stashed
+                # attribute on GroupedMoeBlock instead. Use .clone() to
+                # detach from the stash so activation checkpointing
+                # recomputation cannot overwrite our copy.
+                mlp = getattr(layer, "mlp", None)
```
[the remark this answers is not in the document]
jctian98: it's verified ok, no need to change.

jctian98:
The PR has tried to increase the coverage as much as possible.
I think it's ready for review @sw005320 @Masao-Someki @wanchichen

espnet2/speechlm/model/speechlm/lm/parallel.py
```diff
@@ -49,95 +51,91 @@ def from_pretrained(
...
-                    if not hasattr(model, "lm_head"):
-                        raise AttributeError("Model must have 'lm_head' attribute")
+                nn.init.normal_(new_embed_tokens.weight, mean=0.0, std=0.02)
+                nn.init.normal_(new_lm_head.weight, mean=0.0, std=0.02)
 
-                    text_start, text_end = vocab_intervals["text"][0]
+                # (2.2) override by old embedding and lm head
+                assert hasattr(model.model, "embed_tokens")
+                assert hasattr(model, "lm_head")
+                text_start, text_end = vocab_meta["text_start"], vocab_meta["text_end"]
```
Masao-Someki, line 102: qq, am I right that the `text_start` and `test_end` corresponds to `sos` and `eos` token?

espnet2/speechlm/model/speechlm/lm/parallel.py
```diff
@@ -211,29 +171,122 @@ def forward(self, **kwargs):
...
+            # Forward through base transformer model.
+            # attn_args carries pre-computed flash attention kwargs:
+            #   pack mode:   cu_seq_lens_q/k, max_length_q/k
+            #                (avoids per-layer .item() sync)
+            #   bucket mode: empty (flash attention uses is_causal=True)
             output = self.model(
                 inputs_embeds=inputs_embeds,
                 position_ids=position_ids,
+                output_router_logits=True,
+                use_cache=False,
```
Masao-Someki, line 187: I'm not sure if this should be hardcoded or configurable. Could we allow users to change it via config instead? If this is the typical use case, though, leaving it hardcoded may be fine.

espnet2/speechlm/model/speechlm/lm/parallel.py
```diff
@@ -211,29 +171,122 @@ def forward(self, **kwargs):
...
+                **attn_args,
             )
 
+            model_output = (
+                output.last_hidden_state,
+                getattr(output, "router_logits", None),
+            )
+            target = (input_ids, loss_mask)
+            scale = kwargs.get("loss_scale", None)
+            return self._loss(model_output, target, scale=scale)
```
Masao-Someki, line 197: Hmm, from espnet3's perspective, it might be difficult to use this model, because it only returns loss.
How does the logging works?

espnet2/speechlm/model/speechlm/lm/parallel.py
jctian98: No, it's the start and end index of text vocab within the whole multimodal vocab; it doesn't necessarily start and end with `sos/eos`.

espnet2/speechlm/model/speechlm/lm/parallel.py
jctian98: If this is an MoE model, it's always necessary to compute the load_balance loss with such router logits; otherwise, there is no action (Dense models). 

Users might not need to really care about this implementation, which brings extra complexity to them. There should be no case where other configs are valid (unless they intentionally work on MoE balance issue, not like in speech research)

Same for use_cache: never use cache during training forward function.

espnet2/speechlm/model/speechlm/lm/parallel.py
jctian98: I was returning both loss and stats together in the previous implementation, but that was not compatible with torch's `pipeline parallel`, which only accepts loss as the output. 
So the loss is returned from a module-level attributes rather than from a dict in `return`

espnet2/speechlm/model/speechlm/lm/parallel.py
Masao-Someki: Thanks!

espnet2/speechlm/model/speechlm/lm/parallel.py
Masao-Someki: Thank you!
I think a little comments on this point would be helpful!

espnet2/speechlm/model/speechlm/lm/parallel.py
Masao-Someki: Understand, so if people wants to run their model with pipeline parallel, they need to return only the loss from the model..
Thank you, I'll revise the espnet3 side in this case.

wanchichen:
The technical implementation lgtm. Using liger kernels is much more efficient, we should actually consider using it for more parts of the code elsewhere in espnet.

espnet2/speechlm/model/speechlm/lm/parallel.py
sw005320: I see. This is a big constraint, but we have to follow Torch's pipeline in parallel. So, we can discuss how we can change ESPnet3 due to this.

sw005320:
Thanks!
What's next?

espnet2/speechlm/model/speechlm/lm/parallel.py
jctian98: Tbh, I guess we may not use this pipeline parallel feature recently and beyond LLM training. It's basically prepared for big LLMs (e.g., Qwen3-235B)

If that's a very big constraint and may not be generalizable to most other ESPnet3 applications, maybe we can make it a unique case in the SpeechLM module for this moment.

jctian98:
@sw005320 next is https://github.com/espnet/espnet/pull/6422

espnet2/speechlm/model/speechlm/lm/parallel.py
sw005320: I'm sure the "current" big LLMs won't be so big in 2 or 3 years.
So, we should support it.

PR #6423, opened 2026-04-17
[SpeechLM] trainer and utilities

jctian98:
## Summary

Add a torchtitan-based trainer for SpeechLM as an alternative to the existing DeepSpeed path. `train.py` now reads `trainer.type` from the training config and dispatches to either `DeepSpeedTrainer` (default, unchanged) or the new `TitanTrainer` / `TitanPPTrainer`.

- **`titan_trainer.py`** — FSDP2-based trainer, interface-compatible with `DeepSpeedTrainer`. Handles DCP checkpointing, optimizer/scheduler, logging, and optional from-scratch re-init.
- **`titan_trainer_pp.py`** — Subclass adding pipeline-parallel `train()` / `valid()` via torchtitan's PP schedule; `self.model` is always an `nn.ModuleList` of chunks so iteration stays uniform.
- **`bin/train.py`** — Adds `set_seed()` for cross-rank reproducibility, loads config early, and dispatches on `trainer.type`. DeepSpeed import is now lazy so non-deepspeed runs don't require it.
- **`requirement.txt`** — Pins the extra deps (`torchtitan`, `liger-kernel`, `transformers==4.57.1`, `wandb`, `polars`, `duckdb`, …).
- **`__init__.py`** — Removed (empty, no longer needed; the package has a `README.md`).


espnet2/speechlm/trainer/titan_trainer.py
```diff
@@ -0,0 +1,622 @@
...
+                    break
+                batch = to_device(
+                    batch, self.device, dtype=self.dtype, non_blocking=True
+                )
+
+                local_count = (batch["loss_masks"][:, :, 0] != 0).float().sum()
+                global_count = local_count.clone()
+                dist.all_reduce(global_count, op=dist.ReduceOp.SUM, group=self.dp_pg)
+                batch["loss_scale"] = self.dp_size / (global_count * grad_accum)
+
```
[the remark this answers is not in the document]
jctian98: The loader in ensured to have the same length. no problem here.

espnet2/speechlm/trainer/titan_trainer_pp.py
```diff
@@ -0,0 +1,310 @@
...
+                kwarg_mbs.append(
+                    to_device(batch, self.device, dtype=self.dtype, non_blocking=True)
+                )
+
+            for mb in kwarg_mbs:
+                local_count = (mb["loss_masks"][:, :, 0] != 0).float().sum()
+                global_count = local_count.clone()
+                dist.all_reduce(global_count, op=dist.ReduceOp.SUM, group=self.dp_pg)
+                mb["loss_scale"] = self.dp_size / (global_count * len(kwarg_mbs))
+
```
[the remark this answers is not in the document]
jctian98: same above

espnet2/speechlm/trainer/titan_trainer_pp.py
```diff
@@ -0,0 +1,310 @@
...
+        for key in keys:
+            val = stats.get(key, 0.0)
+            t = (
+                torch.tensor(val, device=self.device)
+                if not isinstance(val, torch.Tensor)
+                else val.clone()
+            )
+            dist.broadcast(t, group_src=pp_src, group=self.pp_pg)
+            result[key] = float(t.cpu())
+        return result
```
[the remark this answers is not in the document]
jctian98: all GPUs in the same pp rank should have the same stats

espnet2/speechlm/trainer/titan_trainer.py
```diff
@@ -0,0 +1,622 @@
...
+        for key, handle in handles:
+            if handle is not None:
+                handle.wait()
+
+        count = stats.pop("count")
+        for key in stats:
+            if key in self.count_normalized_keys:
+                stats[key] = stats[key] / count
+            else:
```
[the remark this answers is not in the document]
jctian98: it always has the count, as in the training loop

espnet2/speechlm/trainer/titan_trainer.py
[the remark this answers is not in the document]
jctian98: it's never zero. count is never zero

espnet2/speechlm/trainer/titan_trainer_pp.py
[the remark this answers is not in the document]
jctian98: same above

espnet2/speechlm/trainer/titan_trainer_pp.py
```diff
@@ -0,0 +1,310 @@
...
+                logger.info(f"step {self.global_step}, stats: {short}")
+
+            if self.global_step > 1 and self.global_step % self.gc_freq == 0:
+                gc.collect(1)
+
+            self.global_step += 1
+
+    def valid(self) -> None:
+        pass  # TODO(Jinchuan): Implement validation loop
+
```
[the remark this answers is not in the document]
jctian98: this is leaved for future.

wanchichen:
lgtm!

espnet2/speechlm/trainer/titan_trainer.py
```diff
@@ -0,0 +1,622 @@
...
+import math
+import time
+from pathlib import Path
+from typing import Any, Dict, Optional
+
+import torch
+import torch.distributed as dist
+import torch.distributed.checkpoint as dcp
+import torch.nn as nn
+import wandb
```
sw005320, line 23: Did `wandb` become the default now?

espnet2/speechlm/requirement.txt
sw005320, line 1: This package seems to be duplicated with the main pyproject.toml
Also, this will bring some inconsistent package issues. 
So, it should be added to `[project.optional-dependencies]`

Please also minimize the libraries
@Masao-Someki, please check this part carefully

test/espnet2/speechlm/trainer/conftest.py
```diff
@@ -1,20 +1,28 @@
 """Dependency stubs for espnet2/speechlm/trainer tests.
 
 Injects lightweight stubs for heavy optional dependencies (deepspeed,
-wandb, humanfriendly) so that tests can run in CI without those packages.
+wandb, humanfriendly, torchtitan, parallel_utils) so that tests can run
+in CI without those packages. The real packages are used when available.
```
sw005320: These sound very tricky for me, but if this is a widely used practice, it would be fine.

sw005320 (commented):
They look good to me overall, but we need to avoid its own requirement under the module as we prepare such a scheme in `pyproject.toml`

@Masao-Someki, please review this PR

espnet2/speechlm/trainer/titan_trainer.py
```diff
@@ -0,0 +1,622 @@
...
+    """TorchTitan-based trainer with FSDP2 support for SpeechLM.
+
+    This trainer provides distributed training using PyTorch's native FSDP2
+    (Fully Sharded Data Parallel) instead of DeepSpeed ZeRO. It maintains
+    interface compatibility with DeepSpeedTrainer for easy switching.
+
+    IMPORTANT: wandb is MANDATORY and must be initialized before creating this trainer.
+    The trainer will raise an error if wandb.run is None.
+    Wandb should always be initialized in offline mode for local-only logging.
+    All training metrics, losses, and stats are logged to local wandb files.
```
Masao-Someki, line 91: I think it may be good to have reason here on why wandb is required

espnet2/speechlm/trainer/titan_trainer.py
```diff
@@ -0,0 +1,622 @@
...
+        def lr_lambda(step):
+            if step < warmup_steps:
+                return float(step) / float(max(1, warmup_steps))
+            elif step < decay_end_step:
+                progress = (step - warmup_steps) / max(1, decay_end_step - warmup_steps)
+                return min_lr_ratio + 0.5 * (1.0 - min_lr_ratio) * (
+                    1.0 + math.cos(math.pi * progress)
+                )
+            else:
+                return min_lr_ratio
```
Masao-Someki, line 332: Does this titan trainer should only support this le scheduler?
If that is the case I think it is good to have some document on that point.

espnet2/speechlm/trainer/titan_trainer.py
Masao-Someki: I checked the code, maybe wandb is not mandatory?

espnet2/speechlm/requirement.txt
Masao-Someki: @jctian98 I think you can move these in the pyproject.toml like

```
speechlm=[
  "huggingface-hub",
  ...
]
```

test/espnet2/speechlm/trainer/conftest.py
Masao-Someki: I think it’s better to use the actual package. Coding AIs often do this, but it hide errors caused by package version issue.

Masao-Someki (commented):
Thank you, I added several comments!

espnet2/speechlm/trainer/titan_trainer.py
jctian98: imported, but may not be in real use.
I guess installing it may not cause trouble for users

espnet2/speechlm/trainer/titan_trainer.py
jctian98: as long as you set the mode to "disable," wandb will not have any behavior.

espnet2/speechlm/trainer/titan_trainer.py
jctian98: LLM training is usually with this scheduler; I can add a doc

test/espnet2/speechlm/trainer/conftest.py
sw005320: OK.
If you think so, then, I suggest @jctian98 to avoid using a hucky import or override technique...
I also had the same experience with the coding agent and did not accept their proposal changes.

jctian98: I can try to use the actual packages but that may not always work, mainly because:
(1) some packages need login, like wandb
(2) some packages are GPU-only, like liger-kernels, or torchtitan-related. they even need a distributed environment.

Otherwise I can make the exact packages.

sw005320: Make sense.
We can use this as a rule.

jctian98:
Update the PR and wait for CI results:
(1) update the pyproject.tolm; the installation of the SpeechLM dependency is optional now.
(2) add docstring to LR scheduler
(3) Revise the test scripts. (3.1) humanfriendly is installed and tested; (3.2) torchtitian is installed; test it if there are GPUs; otherwise, the test would fall back to stubs; (3.3) deepspeed is not tested as it's GPU only.

pyproject.toml
```diff
@@ -119,6 +119,23 @@ spk = [
...
+# Contains only packages not already in the base "dependencies" above.
+# GPU-only runtime deps (torchtitan, liger-kernel) are listed here so that
+# `pip install espnet[speechlm]` pulls them, but they are not imported at
+# module load time in the hot path, so CPU-only installs still succeed.
+speechlm = [
+  "psutil",
+  "transformers==4.57.1",
+  "huggingface-hub",
+  "joblib",
+  "wandb",
```
sw005320: @Masao-Someki, given the popularity of `wandb`, we may add it like tensorbaord.
(@jctian98, we can handle it ourselves. So, you don't have to do any action)

jctian98: Do you want me to keep it for now and updated by @Masao-Someki  later, or do you want me to remove it right now in the [speechlm] group?

Masao-Someki: I think we can simply ask @jctian98 to move the wandb to the main dependency, as we have this conversation in this PR here and it would be a simple modification.
I think we don't have to have additional PR for that.

jctian98: good to me. I made a change.

jctian98:
I'm glad to revise accordingly if we have more comments; otherwise, maybe we can merge this PR first?

sw005320:
Thanks!
Let's move to the next PRs.
@wanchichen and @Masao-Someki, please provide a timely review.

PR #6425, opened 2026-04-20
Update ESPnet2 LoRA adapter integration for extended LoRA backends

wangpuup:
## What did you change?

This PR updates the current LoRA adapter creation in `espnet2/layers/create_adapter_fn.py` so ESPnet can select different LoRA variants backends.
---

This change makes the adapter selections more flexible for experiments with multiple LoRA variants, including SSVD, DoRA, PiSSA, and SVFT.

The updated implementation adds backend selection through the argument `adapter_type`, with supported values:
- `lora`
- `ssvd`
- `dora`
- `pissa`
- `svft`

The default behavior remains unchanged and still uses vanilla LoRA.

---

## Is your PR small enough?

---

## Additional Context

The corresponding extended LoRA library used for these backends is available at: `https://github.com/wangpuup/LoRA` (branch: `ssvd`)

Install with:
```bash
pip install git+https://github.com/wangpuup/LoRA@ssvd
```

### DIFF
--- a/espnet2/layers/create_adapter_fn.py
+++ b/espnet2/layers/create_adapter_fn.py
@@ -90,10 +90,17 @@ def create_lora_adapter(
     dropout_rate: float = 0.0,
     target_modules: List[str] = ["query"],
     bias_type: Optional[str] = "none",
+    adapter_type: str = "lora",
+    **kwargs,
 ):
-    """Create LoRA adapter for the base model.
+    """Create LoRA adapter or its variants for the base model.
 
-    See: https://arxiv.org/pdf/2106.09685.pdf
+    See: LoRA: https://arxiv.org/pdf/2106.09685.pdf
+         SSVD: https://arxiv.org/pdf/2509.02830
+         SSVD-O: https://arxiv.org/pdf/2601.12600
+         DoRA: https://arxiv.org/pdf/2402.09353
+         SVFT: https://arxiv.org/pdf/2405.19597
+         PiSSA: https://arxiv.org/pdf/2404.02948 
 
     Args:
         model (torch.nn.Module): Base model to be adapted.
@@ -108,14 +115,19 @@ def create_lora_adapter(
             "none" means not training any bias vectors;
             "all" means training all bias vectors, include LayerNorm biases;
             "lora_only" means only training bias vectors in LoRA adapted modules.
-
+        adapter_type (str): Backend type for torch.nn.Linear LoRA modules.
+            Defaults to "lora".
+        **kwargs: Additional backend-specific keyword arguments. For example,
+            rotation_ratio: the rotation ratio for SSVD. Only used when adapter_type is "ssvd".
 
     """
 
     if not is_lora_available:
         raise ImportError(
-            "Requiring loralib. Install loralib following: "
-            "https://github.com/microsoft/LoRA"
+            "Requiring loralib. Install the extended loralib backend from "
+        "https://github.com/wangpuup/LoRA (branch: ssvd), "
+        "for example with: "
+        "`pip install git+https://github.com/wangpuup/LoRA@ssvd`"
         )
 
     is_traget_module_exists = False
@@ -134,7 +146,7 @@ def create_lora_adapter(
         parent_module, target_name, target_module = get_submodules(model, key)
         if not isinstance(target_module, lora.LoRALayer):
             new_module = create_new_lora_module(
-                target_module, rank, alpha, dropout_rate
+                target_module, rank, alpha, dropout_rate, adapter_type, **kwargs
             )
             replace_module(parent_module, target_name, target_module, new_module)
         else:
@@ -222,7 +234,12 @@ def create_new_houlsby_module(target_module: torch.nn.Module, bottleneck: int):
 
 @typechecked
 def create_new_lora_module(
-    target_module: torch.nn.Module, rank: int, alpha: int, dropout_rate: float
+    target_module: torch.nn.Module,
+    rank: int,
+    alpha: int,
+    dropout_rate: float,
+    adapter_type: str = "lora",
+    **backend_kwargs,
 ):
     """Create a new lora module for the given target module."""
     bias = hasattr(target_module, "bias") and target_module.bias is not None
@@ -235,13 +252,34 @@ def create_new_lora_module(
             lora_alpha=alpha,
         )
     elif isinstance(target_module, torch.nn.Linear):
-        new_module = lora.Linear(
+        adapter_type_lower = adapter_type.lower()
+        linear_backend_registry = {
+            "lora": getattr(lora, "Linear", None),
+            "ssvd": getattr(lora, "SSVDLinear", None),
+            "dora": getattr(lora, "DoraLinear", None),
+            "pissa": getattr(lora, "PiSSALinear", None),
+            "svft": getattr(lora, "SVFTLinear", None),
+        }
+        available_linear_backends = {
+            name: backend
+            for name, backend in linear_backend_registry.items()
+            if backend is not None
+        }
+        if adapter_type_lower not in available_linear_backends:
+            raise ValueError(
+                f"Unsupported adapter_type '{adapter_type}' for torch.nn.Linear. "
+                f"Available types: {sorted(available_linear_backends.keys())}. "
+                "Ensure the requested backend exists in installed loralib."
+            )
+        backend_cls = available_linear_backends[adapter_type_lower]
+        new_module = backend_cls(
             target_module.in_features,
             target_module.out_features,
             bias=bias,
             r=rank,
             lora_alpha=alpha,
             lora_dropout=dropout_rate,
+            **backend_kwargs,
         )
     else:
         raise ValueError(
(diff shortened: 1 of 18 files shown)


sw005320:
Thanks, @wangpuup.
I just assigned @Masao-Someki as a reviewer since we also want to use this extension in the espnet3 implementation.

espnet2/layers/create_adapter_fn.py
```diff
@@ -108,14 +115,19 @@ def create_lora_adapter(
...
     """
 
     if not is_lora_available:
         raise ImportError(
-            "Requiring loralib. Install loralib following: "
-            "https://github.com/microsoft/LoRA"
+            "Requiring loralib. Install the extended loralib backend from "
+            "https://github.com/wangpuup/LoRA (branch: ssvd), "
+            "for example with: "
+            "`pip install git+https://github.com/wangpuup/LoRA@ssvd`"
```
Masao-Someki: I am not sure if it would be fine to have additional dependency on your repository.
@sw005320 What do you think on this part?

Masao-Someki (commented):
Thank you!
The code looks OK to me, but I am still not sure if it is okay to have additional dependency on your repo..

sw005320:
@wangpuup, can you also add an example recipe of using this and the corresponding results? That would be more useful.

wangpuup:
> @wangpuup, can you also add an example recipe of using this and the corresponding results? That would be more useful.

@sw005320  Sounds good, I'll add it.

wangpuup:
Hi @Masao-Someki ,

There is one remaining CI error in the Ubuntu configuration test. The error seems to come from importing `s3prl` while validating the existing config `egs2/americasnlp22/asr1/conf/train_asr_transformer.yaml`:

```python
File "/home/runner/work/espnet/espnet/tools/venv/lib/python3.10/site-packages/s3prl/upstream/byol_s/byol_a/common.py", line 20, in <module>
      torchaudio.set_audio_backend("sox_io")
  AttributeError: module 'torchaudio' has no attribute 'set_audio_backend'
  Failed to import Flash Attention, using ESPnet default: No module named 'flash_attn'
  Error: S3PRL is not properly installed.
  Please install S3PRL: cd ${MAIN_ROOT}/tools && make s3prl.done
  Error: Process completed with exit code 1.
```
Do you have any suggestions on how I should handle this? Thank you!

espnet2/layers/create_adapter_fn.py
sw005320: I agree with @Masao-Someki 

@wangpuup, there are several options
1. Since there are not so many additions, you may include them inside the network in espnet
2. You can keep them as an external library, but it should be easily usable and maintainable by users and by ESPnet developers by moving this libary as default with some tests

Considering the importance of this package and to promote its usage, my suggestion is 1.

sw005320:
@wangpuup, can you add a unit test to go through this change?
(This means that this package should be included in the default espnet in the default installation, or move the lora models inside espnet)

espnet2/layers/lora/layers.py
sw005320, line 1: If you copy some parts from another repo, please add this reference and also follow their license appriprately, e.g., add a license header for corresponding files

sw005320:
@wangpuup Thanks for the update, and apologies for the direct push — I put a commit on your branch (a51b1322c76ce2f045630b87f7caa92fa675ad7c) so CI would exercise it, but please review it as you would any patch and push back on anything you disagree with.

I traced the actual ESPnet ordering — `AbsTask.build_model` → `create_adapter` (which ends in `model.eval()`) → `init_param` → `Trainer` train/validate/save → inference — and found five bugs that CI was green on. Every new test fails before the commit and passes after.

### Correctness fixes

- **DoRA corrupted the pretrained weight.** `train(mode=True)` never undid the eval-time merge. The comment said `forward` recomputes the decomposition each call, but `forward` reads `self.weight`, which the merge had already overwritten. ESPnet validates after every epoch, so the frozen weight drifted once per epoch — I measured 0.80 after three train/eval cycles, with inference and resume off by 0.26. The DoRA update `m·(W+BA)/‖W+BA‖` can't be inverted from the adapter parameters alone (`‖W+BA‖_row` is lost), so DoRA no longer merges at all.
- **DoRA's `m_initialized` was a plain bool**, so it didn't survive a checkpoint round trip and `apply_m()` overwrote the *trained* magnitude with the row norms of the loaded weight at inference. It's a buffer now.
- **PiSSA factorized the random weight.** `__init__` called `_pissa_factorize()` and set `pissa_initialized=True`, but `replace_module` copies the pretrained weight in *after* construction and `init_param` is loaded later still. The delta is zero at init either way, which is why the tests passed, but `A0`/`B0` spanned a random subspace rather than the principal one. Factorization is lazy now, like SSVD's. (CodeRabbit flagged this too, but its suggested fix — calling `_pissa_factorize()` from `train()` — would have hit the same problem, since the create-time `model.eval()` also runs before `init_param`.)
- **PiSSA's `train()` never called `nn.Linear.train()`**, so `self.training` and the `lora_dropout` child stayed in train mode under `eval()`.
- **`save_strategy: adapter_only` silently saved an empty dict for SVFT/SSVD.** `lora_state_dict` and `mark_only_lora_as_trainable` matched on the `"lora_"` substring; SVFT trains `m_entries`/`gate` and SSVD trains `s`/`gate`/`K_vec`. Both now walk `LoRALayer` modules instead.

SVFT/SSVD also restore the pretrained weight on unmerge (it's exactly `u @ diag(s_pre) @ v`), and all four new backends now track `merged` while their factors are still uninitialized, so the create-time `model.eval()` leaves them in the same state vanilla LoRA ends up in. That makes the re-merge loop in `s2t_inference.py` provably dead, so I removed it — it also only covered s2t, not the other inference entry points.

### Two calls that are really yours

- **SSVD's `s` is now zero-initialized.** `apply_svd()` was kaiming-initializing it, and since `get_sigma()` applies `s · sigmoid(gate)` with `gate = 0`, the layer differed from the pretrained weight by up to 0.38 before a single training step. Zero-init is the usual PEFT invariant and `s` still gets gradient (`∂σ/∂s = 0.5`), but this changes initialization — **the SSVD numbers in the README were produced with the random `s`.** Could you confirm which the paper intends, and whether the results table still holds? If the random init is deliberate I'll revert that hunk.
- **`apply_rotation` builds `A = I + skew`**, i.e. the linearized Cayley transform, so the rotation isn't actually orthogonal. The comment reads intentional, but orthogonality is usually the point of the construction — worth a sanity check against the paper.

### Out of scope for this PR

The `load_pretrained_model.py` change turned unmatched `init_param` keys into a `RuntimeError` for *every* recipe, not just PEFT ones. It's a useful diagnostic — you clearly hit the `encoder.embed.out.0` rename — but as a hard error it breaks any config that loads a partially matching checkpoint. I downgraded it to a warning here; please open it as its own PR so recipe owners see it.

### Smaller follow-ups I left alone

- `tools/installers/install_lora.sh` still runs `pip install loralib`, and `lora.done` is still in `tools/Makefile`, `ci/install.sh` and `ci/install_macos.sh`. Dead now that the layers are vendored, but removing it touches the CI image config so it deserves its own PR.
- `rank: 0` in `peft_tuning_owsm_svft.yaml` / `_ssvd.yaml` makes `decoder.embed.0` a no-op — `LoraEmbedding(r=0)` builds no adapter — so those two configs adapt one fewer module than the lora/dora/pissa ones. `create_new_lora_module` now warns about this; you may want per-backend target lists.
- I dropped the leftover `pytest.importorskip("loralib")` in `test/espnet2/layers/test_create_adapter.py`, which was skipping the LoRA adapter test now that loralib is no longer a dependency.

Tests added: 7 layer-level, 2 for the state-dict helpers, and a 5-backend end-to-end test that walks create → `init_param` → train → save → inference → resume. Locally: 244 passed across `test/espnet2/layers` and `test/espnet2/torch_utils`; black/isort/flake8/pycodestyle clean.

wangpuup:
@sw005320  Thanks for fixing the bugs!

- I added the license headers: the `LoRALayer` / `Embedding` / `Linear` classes and the helpers in `utils.py` are adapted from `microsoft/LoRA` and now carry the Microsoft copyright / MIT license notice. The DoRA / PiSSA / SVFT / SSVD backends are implemented based on the respective papers; I added the paper references in the module and class docstrings.

- Regarding `save_strategy: adapter_only`: this is indeed a bug that I had planned to fix in the next commit. Thanks for helping with this!

- Regarding `SSVD's s is now zero-initialized`: I re-ran the full MyST recipe with zero initialization and it reproduces the results, so I'm keeping the zero initialization.

- Regarding `apply_rotation builds A = I + skew`: the linearized Cayley map is intentional. In Sections III-A and III-B of the paper, we describe two constraints: a strict orthogonal constraint (exact Cayley transform) and an approximate orthogonal constraint (the current linearized Cayley map), with the linearized map being noticeably faster. We compared the performance of these two mappings in Table VI and found the difference to be small in our experiments, so the linearized Cayley map is used by default. Since the paper also defines the strictly orthogonal variant, I added a `rotation_map` option (`linear` = published default, `cayley` = exact Cayley transform `(I-S)⁻¹(I+S)`). Both use the same parameters, so checkpoints are interchangeable. I also verified both settings again on MyST.

- Regarding `out of scope`: agreed. I'll propose this separately in another PR.

PR #6342, opened 2026-01-19
[espnet3-16] Add demo stage

Masao-Someki:
## What did you change?

Added demo stage.
- New stages `pack_demo`, `upload_demo` for BaseSystem.
- Added templates for demo.
- Currently unable to run space, as espnet3 is not released yet, but after merging #6331 we should be able to run space by changing the requirements of the demo.

---

## Why did you make this change?

Automatic generation of models are very important for easily open-source and easily sharing the research outcome.
Since it uses the code and configs for the inference stage, ESPnet3 can generate demos for not only espnet-2-based pretrained models, but also user-defined custom models.

---

## Is your PR small enough?

No, it is larger than 20 file changes.
This is because it adds `espnet3/demo/` package and corresponding unit tests, also adding new CI job for UI tests that requires various files including recipe-side changes and templates.

---

## Additional Context

This PR should be merged after #6339.

* #6133
* #6327
* #6328
* #6329
* #6331
* #6339

PR #6434, opened 2026-05-07
[espnet3-16.1] Add integrration test for demo stage

Masao-Someki:
## What did you change?
Added the demo integration test.

  This PR mainly:
  - adds `ci/test_demo_ui.py`
  - updates `ci/test_integration_espnet3_publication.sh`
  - adds integration-test configs and assets under `test_utils/`
  - adds test model pack fixtures for default and custom demo publication cases

---

Since we had large amount of changes with this PR, I've split this PR into 2 PRs.
This PR focuses on the integration test of the demo stage.

---

## Is your PR small enough?

yes
This PR changes 12 files with 491 insertions and 6 deletions.

---

## Additional Context

* #6133
* #6342


sw005320:
@NewGamezzz, please review this PR

ci/test_integration_espnet3_publication.sh
```diff
@@ -0,0 +1,196 @@
...
+    Path(sys.argv[1]),
+    config_name="publication.yaml",
+    default_package="egs3.TEMPLATE.asr",
+    resolve=True,
+)
+print(getattr(getattr(config, "upload_model", None), "hf_repo", ""))
+PY
+}
+
+python3 -m pip install -e '.[asr]'
```
NewGamezzz, line 115: Do we need to install `gradio` and `playwright` here? I think we also need to install Chromium (`playwright install chromium`) too because `test_demo_ui.py` need to use them.

ci/test_integration_espnet3_publication.sh
```diff
@@ -0,0 +1,196 @@
...
+    check_args+=(--model-tag "${hf_repo}")
+fi
+
+PACK_DIR="${pack_dir}" python3 "${cwd}/ci/test_integration_espnet3_publication_check.py" \
+    "${check_args[@]}"
+
+${python} run.py \
+    --stages pack_demo \
+    --training_config "${training_config}" \
+    --demo_config conf/demo_integration_default.yaml
```
NewGamezzz: Do we need to copy `demo_integration_default.yaml` from `test_utils` first? It doesn't exist in the original directory and appears to be located only in `test_utils`.

egs3/mini_an4/asr/conf/demo.yaml
```diff
@@ -0,0 +1,37 @@
...
+ui:
+  title: mini_an4 ASR demo
+
+##########################################################
+#                   UPLOAD DEMO SETTING                  #
+##########################################################
+# `upload_demo` uploads the packed demo directory to a Hugging Face Space by
+# default.
+upload_demo:
+  hf_repo: espnet/minian4_asr_integration_test
```
sw005320: Would it be possible to get the name from the recipe name? 
We should not leave this for the user.
I'm concerned about the override.

Masao-Someki: Thank you, it is fixed in the previous PR!

sw005320:
LGTM
Please merge it after your work is done and the CI check does not have an issue

PR #6436, opened 2026-05-13
[openbeats-1] BEATs encoder and tokenizer modules

jaeyeonkim99:
## What did you change?

Ported the BEATs encoder and tokenizer modules from the closed PR #6052.

- Added iterative-pretraining support (masked sequence prediction, EMA teacher path) to `BeatsEncoder`.
- Added BEATs tokenizer modules: `BeatsTokenizer` (trained), `BeatsRandomTokenizer` (random projection), and an `AudioTokenizer` wrapper.
- Added unit tests for the encoder, tokenizers, and shared utils.

This PR contains only the BEATs model components. Task class, training CLI, and the iterative pretraining stage runner are in the follow-up `[openbeats-2]` (#6437). Per-corpus recipes will follow in separate PRs after that.

## Why did you make this change?

BEATs is a widely-used audio SSL backbone. The original PR #6052 became too large to review and merge. Re-porting it as smaller PRs to keep each change reviewable. This one is for the encoder and tokenizer modules, and the follow-up `[openbeats-2]`(#6437) covers the training stack (task class, CLIs, stage runner). Per-corpus recipes will follow in separate PRs.

---

## Is your PR small enough?

No . 10 files, ~1800 lines. The encoder and tokenizer modules are inter-dependent and need to be merged together.

---

## Additional Context

* #6052 

This PR is the original PR for OpenBEATs implementation.

PR #6052, opened 2025-03-02
Audioverse, BEATs pre-training and Dasheng audio encoder

Shikhar-S:
## What?

1. Top level runner for evaluating any BEATs-like checkpoint on various audio tasks.
Currently all the code relevant to this functionality is in `egs2/audioverse/v1/run.sh`
2. Pre-training code for BEATs, along with data preparation scripts for additional data sources.
3. Inference code for [Dasheng](https://arxiv.org/pdf/2406.06992) audio encoder 

## Why?

To efficiently run multiple evaluation tasks and pre-train BEATs like audio encoder models 

## See also

Can run recipes from multiple previous PRs:
1. https://github.com/espnet/espnet/pull/5967
2. https://github.com/espnet/espnet/pull/5977
4. https://github.com/espnet/espnet/pull/5998
5. https://github.com/espnet/espnet/pull/6004
6. https://github.com/espnet/espnet/pull/6006
7. https://github.com/espnet/espnet/pull/6028

PR #6437, opened 2026-05-13
[openbeats-2] BEATs pretraining 

jaeyeonkim99:
## What did you change?

Added the training stack on top of the encoder and tokenizer modules from `[openbeats-1]` (#6436 ).

- Added `BeatsPretrainModel` and `BeatsTask` for iterative pretraining.
- Added training CLIs (`beats_train`, `beats_tokenizer_train`) and a Python fbank extraction CLI (`compute_fbank_feats`).
- Added `egs2/TEMPLATE/ssl1/beats.sh` to drive the `iter0 → tok1 → iter1 → tok2 → ...` loop, with a unified DeepSpeed / non-DeepSpeed checkpoint converter and an optional auto-overlay of corpus fbank stats onto the training configs.
- Added shared TEMPLATE helpers used by the stage runner: audio tokenization driver, fbank stats overlay.
- Added unit tests for the pretraining model and task.

The code has been tested with a small recipe (ESC-50). Per-corpus recipes will follow in separate PRs.

---

Continuation of the BEATs port. Adds code needed to run the iterative pretraining loop on top of the modules from #6436. The full loop has been smoke-tested on ESC-50 in both DeepSpeed (2 GPU) and non-DeepSpeed (1 GPU) modes.

---

## Is your PR small enough?

No. 15 files, ~2500 lines. These are training modules that need to be merged together for the stack to be runnable.

---

## Additional Context

* #6052
* #6436 

This PR should be merged after #6436. Per-corpus recipes will follow in separate PRs.


sw005320:
@popcornell, can you review this PR?

espnet2/beats/encoder.py
```diff
@@ -1095,6 +1378,88 @@ def forward(
...
+                    attn_mask_ = torch.zeros_like(attn_mask) + attn_mask_rel_pos
+                    attn_mask_ = attn_mask_.masked_fill(
+                        attn_mask.logical_not(), float("-inf")
+                    )
+                    attn_mask = attn_mask_
+                else:
+                    attn_mask = attn_mask_rel_pos
+
+            attn = F.scaled_dot_product_attention(
+                query=q,
```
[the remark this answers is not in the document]
Shikhar-S: @jaeyeonkim99 I am not confident about this suggestion. Could you please recheck? The test `test_flash_attn` in `test_beats_encoder.py` should pass, because when first implementing I made sure that the values in different version of supported attention are equal using that test.

egs2/TEMPLATE/ssl1/beats.sh
```diff
@@ -0,0 +1,690 @@
...
+            if "${use_corpus_fbank_stats}"; then
+                _stats_file="${data_feats}/${train_set}/fbank_stats.txt"
+                if [ ! -f "${_stats_file}" ]; then
+                    log "WARNING: ${_stats_file} not found; skipping fbank-stats overlay."
+                else
+                    for _cfg in "${train_config}" "${tokenizer_train_config}" "${tokenizer_inference_config}"; do
+                        [ -z "${_cfg}" ] && continue
+                        [ ! -f "${_cfg}" ] && continue
+                        log "Overlaying corpus fbank stats from ${_stats_file} into ${_cfg}"
+                        ${python} pyscripts/utils/overlay_fbank_stats.py \
```
Shikhar-S, line 295: nit: we probably do not need the new utility file? overlay is possible via inline bash here.

Shikhar-S (commented):
Thanks for working on this @jaeyeonkim99
Added a comment, but not critical.

espnet2/beats/encoder.py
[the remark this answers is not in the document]
popcornell: no this suggestion is actually wrong

espnet2/beats/espnet_model.py
```diff
@@ -0,0 +1,340 @@
...
+        if self.contrastive_loss_weight != 0:
+            contrastive_loss = (
+                self.contrastive_loss_weight
+                * self.patch_contrastive_loss(unmasked_patch_emb)
+            )
+            # Remove cls
+            unmasked_patch_emb = unmasked_patch_emb[:, 1:, :]
+            restore_ids = restore_ids[:, 1:] - 1
+            kept_mask = kept_mask[:, 1:]
+            patch_len = torch.cat([patch_len, patch_len], dim=0)
```
popcornell, line 123: why here we double the patch_len ?

espnet2/beats/espnet_model.py
```diff
@@ -0,0 +1,340 @@
...
+            acc_unmask: scalar
+        """
+        logits = logits.transpose(1, 2)  # (Batch, codebook_size, n_patch)
+        loss = self.loss_function(logits, target)
+        loss = loss * masked  # do not count loss for unmasked patches
+        loss = loss.sum()
+
+        padding_mask = make_pad_mask(speech_lengths).to(loss.device)
+        unmasked = ~masked & ~padding_mask  # not masked and not padded
+        masked = masked & ~padding_mask  # masked and not padded, no-op
```
popcornell: why we need such no-op ?

espnet2/beats/espnet_model.py
```diff
@@ -0,0 +1,340 @@
...
+    def _calc_beats_tokenizer_loss(
+        self,
+        output: torch.Tensor,
+        target: torch.Tensor,
+        lengths: torch.Tensor,
+        encoder_dict: dict,
+    ):
+        cos_sim = F.cosine_similarity(target, output, dim=-1)
+        pad_mask = make_pad_mask(lengths).to(cos_sim.device)  # can optimize
+        cos_sim[pad_mask] = 1.0
```
popcornell: we maybe should avoid in-place here. 

```python
cos_sim = F.cosine_similarity(target, output, dim=-1)
pad_mask = make_pad_mask(lengths).to(cos_sim.device)
cos_loss = ((1 - cos_sim) * (~pad_mask)).sum() / lengths.sum()
```

espnet2/beats/espnet_model.py
popcornell: ok it is not a no-op ignore my comment

espnet2/beats/espnet_model.py
```diff
@@ -0,0 +1,340 @@
...
+        preds = logits.argmax(dim=1)
+        with torch.no_grad():
+            corr_masked = ((preds == target) * masked).sum()
+            corr_unmask = ((preds == target) * unmasked).sum()
+
+            count_unmask = unmasked.sum()
+            count_masked = masked.sum()
+            acc_m = corr_masked / (count_masked + 1e-10)
+            acc_u = corr_unmask / (count_unmask + 1e-10)
+            n_uniq_pred = torch.unique(preds, return_counts=False).numel()
```
popcornell: shouldnt we avoid padding positions ? 

```python
valid = masked | unmasked
n_uniq_pred = torch.unique(preds[valid], return_counts=False).numel()
vocab_cov_pred = n_uniq_pred / logits.shape[1]
n_uniq_pred_masked = torch.unique(
preds[masked], return_counts=False).numel()
n_uniq_pred_unmask = torch.unique(preds[unmasked], return_counts=False).numel()
n_uniq_tgt_masked = torch.unique(target[masked], return_counts=False).numel()
n_uniq_tgt = torch.unique(target[valid], return_counts=False).numel()
vocab_cov_tgt = n_uniq_tgt / logits.shape[1]
```

espnet2/speechlm/tokenizer/beats_tokenizer.py
```diff
@@ -0,0 +1,525 @@
...
+            stride=160,
+            bias=False,
+        )
+        seed = config.seed
+        self.random_projection_quantizer = RandomProjectionQuantizer(
+            config.embed_dim,
+            codebook_size=config.quant_n,
+            codebook_dim=config.quant_dim,
+        )
+        self._initialize(seed)
```
popcornell: if we set the seed after the random_proj then it is ignored ?

egs2/TEMPLATE/ssl1/beats.sh
```diff
@@ -0,0 +1,690 @@
...
+train_encoder() {
+    iteration=$1
+    ssl_exp="${expdir}/beats_iter${iteration}_${ssl_tag}"
+
+    log "Training encoder for iteration ${iteration}..."
+
+    _opts=""
+    [ -n "${train_config}" ] && _opts+="--config ${train_config} "
+
+    if [ "${num_splits_ssl}" -gt 1 ]; then
```
popcornell: At the second iteration, this if is skipped because `.done` exists. 
However afterwards at the second iteration it is not guaranteed that e.g. ../split4/target_iter1_tok exists no ? It should crash ?
@Shikhar-S was this tested before ?

espnet2/asr/encoder/beats_encoder.py
popcornell: shouldn t be weight_only=False ?

egs2/TEMPLATE/asr1/pyscripts/feats/dump_audio_tokens.py
```diff
@@ -0,0 +1,157 @@
...
+        config_path,
+        waveform_input=waveform_input,
+    )
+
+    # (3) Tokenizer loop
+    codec_writer = kaldiio.WriteHelper(wspecifier)
+    audio_reader = kaldiio.ReadHelper(rspecifier)
+
+    buffer, length_buffer, key_buffer = [], [], []
+    wav_reader_len = len(open(rspecifier.split(":")[1]).readlines())
```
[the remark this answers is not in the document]
popcornell: @Shikhar-S I think this bug is legit no ? 

If rspecifier = "ark:foo.ark" it will open a ark file in binary mode and then readlines is on binaries

espnet2/asr/encoder/beats_encoder.py
jaeyeonkim99: Fixed in the parent PR #6436 ([17e7633](https://github.com/espnet/espnet/pull/6436/commits/17e7633800fae4b67487dcf472fae4e5e70cf21d)).

espnet2/beats/espnet_model.py
jaeyeonkim99: Yes, this line isn't a no-op. It filters out padding before the unique count. Removed the misleading comment in [3261681](https://github.com/espnet/espnet/pull/6437/commits/32616819f652bb85dbd7eabf0a4807c92b81218d).

espnet2/beats/espnet_model.py
jaeyeonkim99: Applied your suggested rewrite. Fixed in [3261681](https://github.com/espnet/espnet/pull/6437/commits/32616819f652bb85dbd7eabf0a4807c92b81218d).

egs2/TEMPLATE/ssl1/beats.sh
jaeyeonkim99: I split the guard in two: `.done_common` for invariant files (`speech.scp`, `speech_shape`, `target_shape.word`) throughout the iteration, `.done_iter${iteration}` for the per-iter target. So it will check the target token every iteration and split them. Fixed in [6645e79](https://github.com/espnet/espnet/pull/6437/commits/6645e79ff773878fe33b3edbfd137e705335b678).

espnet2/speechlm/tokenizer/beats_tokenizer.py
jaeyeonkim99: You are right. Moved `_initialize(seed)' before the quantization constructor in [7b4373c](https://github.com/espnet/espnet/pull/6436/commits/7b4373c3c3de79db277f85be9e0b80b15ee70015) (parent PR #6436).

egs2/TEMPLATE/ssl1/beats.sh
jaeyeonkim99: I just kept as a script for the idempotency check. It tolerance-compares current vs new fbank values and skips rewriting when equal. Also, it is hard to do cleanly inline since it requires bc/awk for float comparison in bash.

espnet2/beats/espnet_model.py
jaeyeonkim99: When `contrastive_loss_weight != 0`, an extra unmasked branch is forwarded to provide contrastive positives/negatives, so the effective patch length doubles.

espnet2/beats/espnet_model.py
jaeyeonkim99: Thanks for pointing out! Applied your suggestion. Fixed in [3261681](https://github.com/espnet/espnet/pull/6437/commits/32616819f652bb85dbd7eabf0a4807c92b81218d).

egs2/TEMPLATE/asr1/pyscripts/feats/dump_audio_tokens.py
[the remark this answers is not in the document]
jaeyeonkim99: Refactored to a `flush()` helper plus a partial-batch flush after the loop, removing the up-front line count (which also crashed on binary ark files). Fixed in [3261681](https://github.com/espnet/espnet/pull/6437/commits/32616819f652bb85dbd7eabf0a4807c92b81218d).

espnet2/beats/encoder.py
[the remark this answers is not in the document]
jaeyeonkim99: The `1/alpha` in the non-flash path is a numerical-stability trick, not a semantic scaling. It's applied to `q` before the matmul and then cancelled by the `* alpha` on `attn_weights`
  after the max-subtraction, so the non-flash path is equivalent to the flash path's `F.scaled_dot_product_attention(q, k, v)`. Not a bug.

test/espnet2/beats/test_encoder.py
```diff
@@ -2,9 +2,14 @@
...
 
-from espnet2.asr.encoder.beats_encoder import BeatsConfig, BeatsEncoder  # noqa
+from espnet2.asr.encoder.beats_encoder import (
+    BeatsEncoder,
+    BeatsPretrainingPredictor,
+    MultiheadAttention,
+)
 
 is_torch_1_12_1_plus = V(torch.__version__) >= V("1.12.1")
+is_torch_2_plus = V(torch.__version__) >= V("2.0.0")
```
sw005320: We don't need this anymore

test/espnet2/beats/test_espnet_model.py
```diff
@@ -0,0 +1,73 @@
...
+
+from espnet2.asr.encoder.beats_encoder import BeatsEncoder, BeatsPretrainingPredictor
+from espnet2.beats.espnet_model import BeatsPretrainModel, BeatsTokenizerPretrainModel
+from espnet2.speechlm.tokenizer.beats_tokenizer import (
+    BeatsTokenizer,
+    BeatsTokenizerPretrainingPredictor,
+)
+
+is_torch_1_12_1_plus = V(torch.__version__) >= V("1.12.1")
+
```
sw005320: We don't need this

sw005320: Please remove the related part and simplify your code

espnet2/beats/generate_beats_checkpoint.py
sw005320, line 1: I think this kind of checkpoint processing can be done by prepared functions in a trainer.
Or is it designed for some special situations?
(I see that this handles the checkpoint for the multiple trainers)

egs2/TEMPLATE/ssl1/beats.sh
sw005320, line 1: This script is too long.
Is it possible to split some of the functions into the other files and make the stage simpler?

```
setup_common_vars
_best_ckpt_from_run_dir
generate_checkpoint
train_encoder
train_tokenizer
tokenizer_inference
```

sw005320 (commented):
I have several comments

test/espnet2/beats/test_encoder.py
jaeyeonkim99: Removed the guard. Fixed in [7f50c61](https://github.com/espnet/espnet/pull/6437/commits/7f50c61d25ee35fb90a9e7bce746ce91ee3c411e).

test/espnet2/beats/test_espnet_model.py
jaeyeonkim99: Removed the guard. Fixed in [7f50c61](https://github.com/espnet/espnet/pull/6437/commits/7f50c61d25ee35fb90a9e7bce746ce91ee3c411e).

egs2/TEMPLATE/ssl1/beats.sh
jaeyeonkim99: Done. Moved the six stage-7 helpers (`setup_common_vars`, `_best_ckpt_from_run_dir`, `generate_checkpoint`, `train_encoder`, `train_tokenizer`, `tokenizer_inference`) into a sibling `egs2/TEMPLATE/ssl1/beats_utils.sh`, sourced by `beats.sh` (path resolved via `readlink -f` so the recipe symlinks still work). `beats.sh` is down from 713 to 447 lines and the stage bodies are now just the helper calls. Fixed in [93d5ad3](https://github.com/espnet/espnet/pull/6437/commits/93d5ad3cf0f09acd7e3b8dc1be30852a9e935b86).

espnet2/beats/generate_beats_checkpoint.py
jaeyeonkim99: It's BEATs-specific and not covered by the trainer's checkpoint utilities. It (1) extracts only the `encoder.` submodule out of a full pretraining state dict (which also holds the prediction head, and the frozen teacher during tokenizer training), (2) normalizes three container formats of DeepSpeed (`module`), Lightning (`state_dict`), and plain ESPnet and (3) repackages the weights plus the cleaned BEATs `cfg` into the standalone `.pt` that `BeatsEncoder(beats_ckpt_path=...)` loads between iterations. The logic is BEATs-specific, so keeping it separate seems cleaner than folding it into the trainer.

egs2/TEMPLATE/ssl1/beats.sh
jaeyeonkim99: For context: the other SSL recipes (`ssl1/ssl.sh`, `hubert1/hubert.sh`) keep everything inline in a single ~600–720-line script and handle iterative training in-line.  Still, for our case, I think splitting the file is preferable as it can keep the stages simpler and per-iteration logic reusable.

jaeyeonkim99:
While reproducing the AS-2M BEATs training end-to-end, hit the bug during tokenizer training that VQ codebook collapses to a single entry under DeepSpeed bf16.

Root cause: in `EmbeddingEMA` the codebook (`weight`, `cluster_size`, `embed_avg`) was an `nn.Parameter(requires_grad=False)` updated in place by kmeans/EMA. Under DeepSpeed bf16 such a parameter gets an fp32 master copy (zero-init, never in the optimizer) that overwrites the in-place EMA writes every step, so the codebook stays at its zero initialization. Fixed in [bc13775](https://github.com/espnet/espnet/pull/6437/commits/bc13775f46fbc43291931a748744023c7e10ce4a).

sw005320:
I think this PR is in good shape now.
@jaeyeonkim99, is your development done?
If so, I'll merge it after the CI test.

Can you move to the recipe?

sw005320 (approved):

PR #6438, opened 2026-05-13
Refactor file check conditional to handle missing files and remove train/val directory name assumption

G-Thor:
The current version of this script silently fails if utt2gender is not present in the data dir due to a falsy return value in sort_data_dir() triggering pipefail.

Refactoring the file existence check as a proper if clause fixes this issue.

## What did you change?

I refactored a file existence check which was causing silent failures when spk2gender was not included in the data dir.
I also made a change to how the rawdir variable is determined, opting to check the /raw/org/ directory first and falling back to /raw/ if /raw/org/ does not exist. Before, there was a simple check for whether the data dir was named exactly "train" or "val"

---

The implicit falsy return value of sort_data_dir was causing silent pipeline failures when spk2gender was not present in the directory.
The explicit dirname check was also failing on my end due to my training set being named "tr_no_dev" rather than "train".
---

## Is your PR small enough?

yes


sw005320:
Thanks a lot!

PR #6439, opened 2026-05-13
Clean dependencies

Masao-Someki:
## What did you change?

Clean up pyproject.toml to reduce dependencies

---

For smaller size of dependency tree

---

## Is your PR small enough?

yes

---

## Additional Context

-


Masao-Someki:
- setuptools: Removed from dependencies. It is only needed in `build-system.requires`, not at runtime.
- configargparse: Removed from dependencies. The repo uses its own local implementation in `espnet2.utils.config_argparse`, not the external package.
- packaging: Removed from dependencies. It is expected to come from higher-level dependencies.
- scipy: Removed from dependencies. It is expected to come via librosa.
- numpy: Removed from dependencies. It is expected to come via higher-level dependencies such as librosa.
- PyYAML: Removed from dependencies. It is expected to come via hydra-core or lightning.
- soundfile: Removed from dependencies. It is expected to come via librosa or lhotse.
- omegaconf: Removed from dependencies. It is expected to come via hydra-core.
- datasets: Removed from dependencies. I could not find actual usage in the repo.
- torch_optimizer: Removed from dependencies. It only adds optional extra optimizers; standard torch.optim still works without it.
- sentencepiece: Removed from base dependencies and moved into task groups.
- sentencepiece moved to: asr, tts, asr2, s2st, st, s2t, and speechlm.
- lhotse: Removed from base dependencies and moved to speechlm, since its usage was limited to SpeechLM-related code.
- nltk: Removed from base dependencies and moved into task groups.
- nltk moved to: mt, st, and s2st, since it is used for BLEU-related MT/ST logic.
- mt: Added as a new optional dependency group so sentencepiece and nltk can live in a task-specific group.
- tokenizer: I created it temporarily, then removed it. The final layout keeps groups task-based only.
- torch_complex: Kept in base dependencies. Removing it would require code changes in espnet2, because the code still directly uses `torch_complex.ComplexTensor` and related APIs.
- einops: Kept in base dependencies. Removing it would also require code changes in espnet2, because it is still used directly in both gan_codec code and ASR state-space code.

Masao-Someki:
@Fhrozen Do you have any idea why we got error with the warp-rnnt installation? I didn't touch the torch-related items (except for `torch_optimizer` for additional optimizers)

Fhrozen:
@Masao-Someki, I am reviewing the issue. Maybe a setuptools version issue.

sw005320:
@Masao-Someki, Copilot suggests that the following can solve the issue. Can you try it?
```
diff --git a/tools/installers/install_warp-transducer.sh b/tools/installers/install_warp-transducer.sh
index c707d8db78..NEW 100755
--- a/tools/installers/install_warp-transducer.sh
+++ b/tools/installers/install_warp-transducer.sh
@@ -45,7 +45,7 @@ git clone https://github.com/ljn7/warp-transducer.git
     (
         set -euo pipefail
-        cd pytorch_binding && python3 -m pip install --no-build-isolation -e .
+        cd pytorch_binding && python3 -m pip install --no-build-isolation .
     )
 )

**Comment by sw005320** [comment]
@Masao-Someki, this PR seems to have several CI issues.
Can you fix them?

**Comment by Fhrozen** [comment]
From the failed CI, I got that:
- Configparse should be installed at the top level (https://github.com/espnet/espnet/actions/runs/26838251409/job/79137919792?pr=6439):
```
test/espnet2/tasks/test_abs_task.py:3: in <module>
      import configargparse
E   ModuleNotFoundError: No module named 'configargparse'
```
- Pyworld is not compatible with the new versions of pip/setuptools; the last update was last year, and the project seems abandoned. (We may need to clone and update)
- It seems that a dependency is missing for test_integrations (https://github.com/espnet/espnet/actions/runs/26838251420/job/79143104595?pr=6439). The output says that the dump has failed, so maybe a package like resampy, librosa, or a similar one that is used for feat dumping is missing (need to test locally).

PR #6441, opened 2026-05-15
Update MFA installation with warning messages

### LINKED ISSUE #5985: Cannot find `_kalpy` required by MFA
**Describe the bug**
I am trying to apply MFA to the JVS corpus. I installed MFA with `make mfa.done`.
Then I modified `egs2/ljspeech/tts1/local/run_mfa.sh` to make it work on JVS.
The error log (see below) says that it cannot find `_kalpy`.
I checked `tools/venv/lib/python3.10/site-packages`, and indeed it is not installed.

I see in `tools/installers/install_mfa.sh` that MFA is installed through pip, but not conda, as the official homepage recommends.
I am guessing this is the reason why kalpy is not properly installed, because when I ran `tools/installers/install_mfa.sh`, I did not see kalpy get installed either.
The official homepage of kalpy https://pypi.org/project/kalpy-kaldi/ also recommends to install by conda.
I tried to install MFA in my local conda environment, and `mfa validate` runs successfully (which means kalpy is also properly installed and found).

I am truly grateful to whoever wrote this function in ESPnet, but I am just wondering how you got it working...???

**Basic environments:**
 - OS information: Linux
 - python version: python3.10
 - espnet version: latest
 - pytorch version: 2.3.0, but does not matter anyway

**Task information:**
 - Task: TTS
 - Recipe: jvs/tts1
 - ESPnet2

**Error logs**
```
# mfa validate -j 12 --clean --skip_acoustics -t data/local/mfa/tmp data/local/mfa/corpus japanese_mfa japanese_mfa --brackets "" 
# Started at Tue Dec 17 17:24:26 JST 2024
#
  File "/data/group1/z44476r/Experiments/espnet/tools/venv/bin/mfa", line 5, in <module>
    from montreal_forced_aligner.command_line.mfa import mfa_cli
  File "/data/group1/z44476r/Experiments/espnet/tools/venv/lib/python3.10/site-packages/montreal_forced_aligner/__init__.py", line 4, in <module>
    import montreal_forced_aligner.acoustic_modeling as acoustic_modeling
  File "/data/group1/z44476r/Experiments/espnet/tools/venv/lib/python3.10/site-packages/montreal_forced_aligner/acoustic_modeling/__init__.py", line 7, in <module>
    from montreal_forced_aligner.acoustic_modeling.base import AcousticModelTrainingMixin  # noqa
  File "/data/group1/z44476r/Experiments/espnet/tools/venv/lib/python3.10/site-packages/montreal_forced_aligner/acoustic_modeling/base.py", line 11, in <module>
    from _kalpy.gmm import AccumAmDiagGmm
ModuleNotFoundError: No module named '_kalpy'
```

Fhrozen:
Fix #5985

## What did you change?

This pull request introduces improvements to error handling and compatibility checks for the MFA installation process and its dependencies. The main changes ensure users are properly informed about missing dependencies and incompatible Python versions, helping to prevent common setup issues.

Dependency and error handling improvements:

* Added a try/except block to the import of `pyopenjtalk` in `mfa_format.py`, raising a clear error message if the module is missing and instructing users how to install it.

Installation and compatibility checks:

* Updated `install_mfa.sh` to check the Python version before installation. If Python 3.11 or newer is detected, a warning is printed about the missing `imp` module and the need to use Python 3.10 or an appropriate workaround.
* Modified `install_mfa.sh` to prevent installation via pip, exiting with an error message that pip-installed MFA is not compatible with ESPnet and recommending conda installation instead.


pyproject.toml
```diff
@@ -47,6 +47,13 @@ dependencies = [
     "matplotlib",
     "nltk>=3.4.5",
 
+    # utils related. Removed from versa.
+    "pysptk",
+    "resampy",
+    "mir-eval",
+    "fastdtw",
+    "pystoi",
+
```
sw005320: We want to minimize the library installation in the main.
All of them seem to be related to some specific applications (e.g., evaluation), and they should be moved to the appropriate place.

Fhrozen: >We want to minimize the library installation in the main.

Yes, I notice that from #6439. 

>All of them seem to be related to some specific applications (e.g., evaluation), and they should be moved to the appropriate place.

As you mentioned, they are related to specific applications, but these are called in some CI tests, like `test_utils`, and `test_integration`. These libraries were initially installed with Versa, but the last update of Versa is moving these libraries to `Versa [audio,external]`.
I can add a specific label like `[integration]` or `[external]` and put them down.
I was planning to add the libraries here and reorder the location in #6439.
Please let me know which suits you better.

sw005320: Then, how about installing Versa [audio,external] instead of vanilla Versa?

I added @ftshijt so that he notices this issue.

Fhrozen: The problem with `versa[external]` is that it installs many incompatible libraries, like:
- "espnet @ git+https://github.com/ftshijt/espnet.git@espnet_inference",
- "openai-whisper @ git+https://github.com/openai/whisper.git",
- "s3prl",

Given the case, I would recommend installing `versa[audio]` and then adding the missing `fastdtw` to the pyproject.toml.

sw005320: We can ask @ftshijt to revisit his design.

ftshijt: Thanks. I checked the Versa extras design and will move `fastdtw` from `external` to `audio`, since it is used by the audio/evaluation metric stack and does not belong with heavier external integrations like Whisper, s3prl, or the ESPnet fork. With that change, ESPnet should be able to depend on `versa[audio]` instead of adding these libraries directly to the main dependencies or using `versa[audio,external]`. 

Will that work for here?

Fhrozen: Yes @ftshijt, Thank for the support.

pyproject.toml
```diff
@@ -46,6 +46,7 @@ dependencies = [
     "espnet_model_zoo",
     "matplotlib",
     "nltk>=3.4.5",
+    "pysptk",
```
sw005320: Why does `pysptk` remain?
Where was it used?

I don't think this is necessary as a basic library.

Fhrozen: It was also installed with Versa, but it seems that it is no longer required.
In the case of ESPnet, it is employed by:
- https://github.com/espnet/espnet/blob/master/egs2/TEMPLATE/asr1/pyscripts/utils/evaluate_mcd.py
- https://github.com/espnet/espnet/blob/master/egs2/TEMPLATE/asr1/pyscripts/utils/evaluate_f0.py
- https://github.com/espnet/espnet/blob/master/utils/generate_wav_from_fbank.py

which are evaluated in `test_utils`.

sw005320: They are TTS evaluation-related, and we should not include them for the main.
Would it be possible to move this under `tts`?
I'm sure they are only used for TTS

sw005320: @Fhrozen, can you find a solution?
If this pysptk issue is solved, we can merge this PR.

sw005320: How about like this?

image

sw005320: @Masao-Someki, what do you think?
I think it is natural to have test_utils specific installation dependencies.

sw005320: @Fhrozen and @Masao-Someki, any comments?

Fhrozen: I will remove this and pull from master once #6445 is merged

Masao-Someki: Sorry for the late response, I think `test_utils` looks good!

sw005320: @Fhrozen, #6445 was done.

sw005320:
@Fhrozen, is this ready to be merged?
If so, please go ahead

PR #6418, opened 2026-04-15
[espnet3-17] Add Librispeech-100h recipe

Masao-Someki:
## What did you change?

Add `egs3/librispeech_100/asr` recipe

---

## Why did you make this change?

To have an example project on how to write espnet3 project

---

## Is your PR small enough?

TBD / This depends on dependent PRs
I will update this part once #6342 becomes stable.

---

## Additional Context

* #6133
* #6327
* #6328
* #6329
* #6331
* #6339
* #6390
* #6391
* # 6392

PR #6443, opened 2026-05-18
[espnet3-20] Bugfix for parallel processing

Masao-Someki:
## What did you change?

- Redesigned `BaseRunner` into a resumable file-based execution. Completed shards are tracked via `.done` files.
- Removed the `parallel_map` / `parallel_for` public APIs. These in-memory parallel utilities are no longer exposed;
- Renamed several functions for consistency: `make_local_gpu_cluster` → `build_local_gpu_cluster`, `make_client` → `build_client`.
- Expanded docstrings.

---

- parallel_for/map caused major slowdowns in espnet3 workloads.
- The in-memory dask parallel concentrated network communication on the parent job.
- That made performance depend too much on having a very strong parent CPU.
- The file-based runner/provider path scales better and matches the espnet2 execution model.
- To avoid keeping a slower and easier-to-misuse API, I standardized espnet3 parallel execution on the file-based path only.

---

## Is your PR small enough?

15 file changed, with 1,769 additions and 1,422 deletions. 

---

## Additional Context

* #6133 
* #6418


Masao-Someki:
For the parallel execution in ESPnet3, here's how the parallel processing works with this PR.

  1. `_plan_shards` — split indices into N shard specs (N = n_workers or 1 locally)
  2. `_write_manifest` — record the plan to `manifest.json`
  3. `_filter_pending_shards` — skip shards already marked done (resume support)
  4. Dispatch — `_run_local` (sequential) or `_run_parallel_dask` (Dask client.map)
  5. Per shard — `init_state` → `forward` × items → `finalize_state` → write done
  6. `merge` (shard_dirs) — aggregate results on the driver
 
espnet3_baserunner_unified_en_fixed

**What subclasses override**

  The worker-side hooks (`open_writers`, `write_record`, `close_writers`) and the
  driver-side `merge` are the only extension points. Both existing runners follow
  the same pattern:

  | Runner | Worker side | Driver side (`merge`) |
  |---|---|---|
  | `InferenceRunner` | streams results into per-shard `.scp` files | cat shard SCPs into `output_dir/` |
  | `CollectStatsRunner` | accumulates sum/sq/count; writes `_shape` files | aggregate stats + cat shape files |

  `forward` is always a `@staticmethod` on the subclass — everything else stays in the hooks.

espnet3/components/data/collect_stats.py
sw005320, line 1: Can someone review this?
@wangyou?

sw005320:
This has a large change and requires a careful review by a team member.

Masao-Someki:
For the design part, I have been discussed with @popcornell in detail. So if @popcornell have time it would be great if you could check this PR..!!

popcornell:
> For the design part, I have been discussed with @popcornell in detail. So if @popcornell have time it would be great if you could check this PR..!!

On it sorry. Many many PRs for me to review these days

espnet3/components/data/collect_stats.py
```diff
@@ -337,6 +525,205 @@ def forward(
...
+        write_feats = writers["_write_feats"]
+        shard_dir: Path = writers["_shard_dir"]
+
+        if write_feats:
+            stats, shape_info, feats = result
+        else:
+            stats, shape_info = result
+            feats = None
+
+        sum_acc = state.setdefault("sum", {})
```
Lakoc: what about usign defaultdict here and in the state?

sum_acc = state.setdefault("sum", defaultdict(float))
sq_acc = state.setdefault("sq", defaultdict(float))
count_acc = state.setdefault("count", defaultdict(int))
for feat_key, agg in stats.items():
    sum_acc[feat_key] += agg["sum"]
    sq_acc[feat_key] += agg["sq"]
    count_acc[feat_key] += agg["count"]

Masao-Someki: Thanks!
My concern is like `sum` and `sq` `torch.Tensor`. I'm not very familiar with `defaultdict`, I'm not sure whether it would be reasonable to use it here...

ci/test_demo_ui.py
```diff
@@ -0,0 +1,227 @@
+from __future__ import annotations
```
Lakoc, line 1: Is it necessary to include this into the PR?

ci/test_integration_espnet3_publication.sh
```diff
@@ -24,6 +23,69 @@ gen_dummy_coverage() {
     ${python} empty.py
 }
 
+write_demo_test_model_pack_default() {
```
Lakoc, line 26: dtto

espnet3/components/data/collect_stats.py
```diff
@@ -102,59 +144,18 @@ def collect_stats_batch(
...
-                if len_batch is not None:
-                    L = int(len_batch[b_idx])
-                    seq = seq[:L]
-                else:
-                    seq = seq[None]
-
-                if not isinstance(seq, np.ndarray):
-                    seq = np.asarray(seq)
-                w[uid] = seq
+def _build_collate_fn(dataloader_config):
```
Lakoc, line 104: Moving docstrings into separate PR would perhaps make this PR smaller

espnet3/components/data/collect_stats.py
```diff
@@ -172,6 +173,21 @@ def _build_collate_fn(dataloader_config):
 
 
 def _build_dataset(config: DictConfig):
```
Lakoc, line 121: what about putting shard_idx as kwarg to emphasise the intended usage?

espnet3/components/data/collect_stats.py
```diff
@@ -245,21 +360,39 @@ def __init__(
...
+            dataloader_config: Dataloader config, including optional
+                ``collate_fn`` settings.
+            mode: Dataset split name such as ``train`` or ``valid``.
+            task: ESPnet task name. When set, the model is resolved through the
+                espnet2 task bridge.
+            shard_idx: Optional shard index applied to shardable datasets.
+            params: Extra config values merged into the provider config. This is
+                typically used for flags such as ``write_collected_feats``.
+        """
         config = OmegaConf.create({})
```
Lakoc: what about?

config = OmegaConf.create({
    "model_config": model_config,
    "dataset_config": dataset_config,
    "dataloader_config": dataloader_config,
    "mode": mode,
    "task": task,
    "shard_idx": shard_idx,
    **(params or {}),
})
super().__init__(config)

espnet3/components/data/collect_stats.py
```diff
@@ -337,6 +525,205 @@ def forward(
...
+        count_dict: Dict[str, int] = defaultdict(lambda: 0)
+        for shard_dir in shard_dirs:
+            shape_keys.update(
+                (shard_dir / "shape_keys.txt").read_text(encoding="utf-8").splitlines()
+                if (shard_dir / "shape_keys.txt").exists()
+                else []
+            )
+            feat_keys_written.update(
+                (shard_dir / "feat_keys_written.txt")
+                .read_text(encoding="utf-8")
```
Lakoc: I would suggest to move to separate function to simplify code as this block of code is used multiple times
eg.  path.read_text(encoding="utf-8").splitlines() if path.exists() else []

espnet3/components/data/collect_stats.py
```diff
@@ -337,6 +525,205 @@ def forward(
...
+        Returns:
+            dict: Aggregated totals with keys ``"sum"``, ``"sq"``, and
+            ``"count"``, each mapping feature key to its accumulated value.
+            Shape files and optional collected-feat SCP files are also
+            concatenated into ``output_dir / mode``.
+        """
+        shape_keys: set = set()
+        feat_keys_written: set = set()
+        stats_keys: set = set()
+        sum_dict: Dict[str, Any] = defaultdict(lambda: 0)
```
Lakoc: defaultdict(int) -> 0 is default value

espnet3/components/data/collect_stats.py
```diff
@@ -337,6 +525,205 @@ def forward(
...
+            encoding="utf-8",
+        )
+        for key in stats_keys:
+            np.savez(
+                shard_dir / f"{key}_stats.npz",
+                count=state["count"][key],
+                sum=state["sum"][key],
+                sum_square=state["sq"][key],
+            )
+        return None
```
Lakoc: Is it necessary to return None? that's default behaviour of void function

espnet3/components/data/collect_stats.py
```diff
@@ -337,6 +525,205 @@ def forward(
...
+            writers: Writer state returned by :meth:`open_writers`. Updated
+                in-place with running sums, shape file handles, and optional
+                feature writers.
+            result: Return value of :meth:`forward` — either
+                ``(stats, shape_info)`` or ``(stats, shape_info, feats)``.
+            state: Persistent in-memory accumulator for ``sum``, ``sq``,
+                and ``count`` across batches.
+            **env: Additional environment keys (ignored).
+        """
+        writers["_state"] = state
```
Lakoc: is it necessary to pass state into writers? If its only for close_writers, wouldnt passing the state directly be cleaner and remove some potential side effects?

espnet3/parallel/base_runner.py
```diff
@@ -1,156 +1,133 @@
...
+    shard-local files under ``output_dir`` and supports resume by reusing shard
+    state when a previous run completed some splits.
 
     Subclass contract:
         - Implement ``@staticmethod forward(idx, dataset, model, **env) -> Any``
           without capturing ``self``. ``idx`` may be a single index or a batch
           of indices depending on ``batch_size``.
         - Provide an :class:`EnvironmentProvider` that builds the required env
           (e.g., dataset/model) for local and worker executions.
+        - Override ``open_writers`` / ``write_record`` / ``close_writers`` /
```
Lakoc: Where merge_state is located? Please update if older docstring.

espnet3/parallel/base_runner.py
```diff
@@ -202,254 +179,389 @@ def forward(idx: int | Iterable[int], dataset, model, **env) -> Any:
...
+                chunks.append(chunk)
+            start += size
+        return [
+            {"shard_id": shard_id, "items": chunk}
+            for shard_id, chunk in enumerate(chunks)
+        ]
 
-    Args:
-        spec_path (str): Path to the shard spec JSON file.
+    def _filter_pending_shards(
```
Lakoc, line 336: Is it fine that this is not atomic?

espnet3/systems/base/inference_runner.py
```diff
@@ -192,13 +334,159 @@ def forward(idx, dataset=None, model=None, **kwargs):
...
+    def __call__(self, indices: Iterable[int]) -> Any:
+        """Run inference, write SCP outputs, and validate output formats.
+
+        Args:
+            indices (Iterable[int]): Dataset indices to run inference on.
+
+        Returns:
+            Any: ``None`` when all results are written to SCP files on disk
+            (the normal case), or a flat list of validated output dicts if
+            the base ``merge`` returns a list.
```
Lakoc: what do you mean here as the base merge? BaseRunner doesnt have the merge method, and merge always returns {} or fails on 443?

espnet3/systems/base/inference_runner.py
```diff
@@ -2,14 +2,101 @@
...
+def _iter_outputs(result: Any) -> List[Dict[str, Any]]:
+    if isinstance(result, list):
+        outputs: List[Dict[str, Any]] = []
+        for item in result:
+            outputs.extend(_iter_outputs(item))
+        return outputs
+    return [result]
+
+
+def _materialize_output_value(
```
Lakoc, line 40: since this is called frequently, isnt the nested import slowing down the code significantly?
dtto for  import torch, is this part intended to be used also without access to numpy or is it mostly instance of str, int...?

Lakoc (commented):
I went through these files and left some comments on the issues I spotted. That said, my familiarity with the core runner/provider mechanics here is limited, so please treat these as flags rather than conclusions.

popcornell:
I may be wrong however I think there may be a bug right now if you resume with a different number of workers. 
Say you do the processing with 2 workers. It creates the two folders and then puts done into each of them. 
Then later the job crashes for whatever reason. 
Now I restart the job with 4 workers. It will see two folders with done and will create another two then merge everything. Will it crash ? it may also just duplicate data by error no ? I think this latter option could happen right now depending on the particular processing and there must be one way to avoid it.

ci/test_demo_ui.py
Masao-Someki: Oh thanks, seems like some formatting has included.. let me remove them

Masao-Someki: No, some files should not be in this PR, let me remove those

espnet3/components/data/collect_stats.py
Masao-Someki: Hmm, thanks.
Since this is a new function, and in espnet3 each function should have detailed docstrings, I'd prefer to include one unless there’s a very strong reason we can’t.

espnet3/components/data/collect_stats.py
Masao-Someki: I think this would be difficult..?
Because the function is called from the `setup` function inside `build_worker_setup_fn`, which is used on workers.

That means it would need to be captured by the pickled worker setup function. In that case, putting everything inside the config would be much easier.

espnet3/components/data/collect_stats.py
Masao-Someki: Oh but wait a second, it seems I included so many docstrings here...
You're right, I will remove those for this PR, we should add those in document PR

espnet3/components/data/collect_stats.py
Masao-Someki: Thanks, will fix with this!

espnet3/components/data/collect_stats.py
Masao-Someki: Thanks, this is a good point, and state should be directory passed to `close_writer`. I will fix it!

espnet3/parallel/base_runner.py
Masao-Someki: Thanks, and this is a very good point. I will add process to create lock file, and confirm different process won't take the same worker directory.

espnet3/systems/base/inference_runner.py
Masao-Someki: It won't slow down (because python has cache import) but the code look not good.
I will simply move it to module-level import.

espnet3/systems/base/inference_runner.py
Masao-Someki: Sorry and thanks, for this one I will change the merge function to return None (returning dict was the issue in the first place)

Masao-Someki:
@sw005320 Sorry for the confusion, this PR should be merged earlier than 3-17 #6418

egs3/mini_an4/asr/src/inference.py
```diff
@@ -1,4 +1,34 @@
-"""Inference output helpers for Mini AN4 ASR recipes."""
+"""Inference helpers for Mini AN4 ASR recipes."""
```
Lakoc: Why this was added?

espnet3/components/data/collect_stats.py
Lakoc: yeah, that is more like cosmetic change

espnet3/components/data/collect_stats.py
Lakoc: thanks, much cleaner now

espnet3/components/data/collect_stats.py
Lakoc: I see, LGTM then

espnet3/components/data/collect_stats.py
```diff
@@ -337,6 +525,205 @@ def forward(
...
+            for key in (
+                (shard_dir / "stats_keys.txt").read_text(encoding="utf-8").splitlines()
+                if (shard_dir / "stats_keys.txt").exists()
+                else []
+            ):
+                stats_keys.add(key)
+                data = np.load(shard_dir / f"{key}_stats.npz")
+                sum_dict[key] += data["sum"]
+                sq_dict[key] += data["sum_square"]
+                count_dict[key] += data["count"]
```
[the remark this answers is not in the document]
Lakoc: @Masao-Someki what about this one?

espnet3/systems/base/inference.py
```diff
@@ -9,80 +9,10 @@
 
 from espnet3.parallel.parallel import set_parallel
 from espnet3.systems.base.inference_runner import _load_output_fn
-from espnet3.utils.writer_utils import write_artifact
 
 logger = logging.getLogger(__name__)
 
 
-def _flatten_results(results):
```
Lakoc, line 17: should this be part of pr? probably fine to have, better to split to different PR?

espnet3/systems/base/inference_runner.py
Lakoc: you are right, looks better now

Lakoc (approved):
LGTM. Splitting the PR further would help, and there are a few cosmetic changes to make.

Lakoc:
> I may be wrong however I think there may be a bug right now if you resume with a different number of workers. Say you do the processing with 2 workers. It creates the two folders and then puts done into each of them. Then later the job crashes for whatever reason. Now I restart the job with 4 workers. It will see two folders with done and will create another two then merge everything. Will it crash ? it may also just duplicate data by error no ? I think this latter option could happen right now depending on the particular processing and there must be one way to avoid it.

@Masao-Someki is this currently fixed via shard_subdir?

espnet3/parallel/base_runner.py
```diff
@@ -202,254 +122,316 @@ def forward(idx: int | Iterable[int], dataset, model, **env) -> Any:
...
-            List[Any] | None: Results for local/parallel modes, or ``None`` in
-            async mode.
+        par_config = get_parallel_config()
+        if not shards:
+            return []
+
+        provider_setup = (
+            self.provider.build_worker_setup_fn()
+            + self.provider.build_worker_setup_fn()
+        )
```
NewGamezzz: If `env` is set to slurm, this becomes a problem. You are trying to add two functions.

NewGamezzz:
> Thanks, and this is a very good point. I will add process to create lock file, and confirm different process won't take the same worker directory.

One case I was wondering about is what happens if a job is killed mid-run (e.g., due to a SLURM timeout or scancel) and the shard lock files are left behind.

If I'm understanding the flow correctly, the lock is only removed after the shard finishes successfully. If the process is terminated before that, the stale lock would remain. Then, on the next run with resume enabled, the worker would be prevented from resuming because the lock file still exists.

I haven't tested this yet, so please treat this as a question.

egs3/mini_an4/asr/src/inference.py
Masao-Someki: It is because I noticed that Transducer side it didn't returned the hypothesis with text, but list of integers.
This class is simply to return the decoded result. Previously we didn't had this because the typecheck wasn't enough and the list of index was saved with string.

espnet3/systems/base/inference.py
Masao-Someki: Yeah I think this should be in this PR.
Previously, infer() collected inference results in memory and wrote the final outputs afterward. 
This change moves that output-writing responsibility to the write/merge hooks in InferenceRunner/BaseRunner, so shard-local outputs are written incrementally and merged at the end.

espnet3/parallel/base_runner.py
Masao-Someki: Oh thanks, this is a bug. I will remove it

Masao-Someki:
Regarding Samuele’s point about running in parallel, having some tasks fail, and then rerunning with a different `n_parallel`: I added an error for this case. For now, `n_workers` should be the same as in the initial run.

However, since we save `manifest.json`, which includes the index information assigned to each worker, we should be able to support parallel processing only for the remaining items. I think that should be handled in a separate PR. As a tentative patch, I added an error here.

test/egs3/mini_an4/asr/test_inference.py
NewGamezzz, line 1: Do you need to wire this test into CI (ci/test_python_espnet3.sh)?

egs3/mini_an4/asr/src/inference.py
sw005320: Upon the discussion with @Masao-Someki, we will implement it on espnet2 to avoid to have this kind of file

egs3/mini_an4/asr/conf/inference_transducer.yaml
```diff
@@ -46,7 +46,7 @@ dataset:
 #                    MODEL AND OUTPUT                    #
 ##########################################################
 model:
-  _target_: espnet2.bin.asr_transducer_inference.Speech2Text
+  _target_: egs3.mini_an4.asr.src.inference.TransducerInferenceWrapper
```
sw005320: please change it to use espnet2 transducer as we discussed

test/egs3/mini_an4/asr/test_inference.py
Masao-Someki: Thanks! I discussed this with Shinji, and I'll remove the inference wrapper for the transducer as well as this test script.

Masao-Someki:
@sw005320 if this looks okay to you, please merge this PR!

sw005320 (approved):

sw005320:
Thanks!
This change looks very good to me!

PR #6405, opened 2026-04-01
Add SOT multi-talker ASR with Whisper

cyhuang-tw:
## What did you change?

Add Serialized Output Training (SOT) for multi-talker ASR using Whisper encoder/decoder with tiktoken tokenization.

**Core pipeline (`espnet2/`):**
- `SOTWhisperModel`: extends `ESPnetASRModel` for SOT training
- `SOTWhisperPreprocessor`: tiktoken-based tokenizer handling timestamps and speaker change tokens, with support for reusing existing BPE tokens as special tokens
- `SOTBeamSearch`: beam search with probability-based timestamp forcing and speaker separator preservation
- `SOTConstraintScorer`: enforces valid SOT output structure (timestamp pairing, non-decreasing order, separator handling)
- `sot_postprocess`: repetition truncation for hallucination prevention
- `SOTASRTask`: task registration with native Whisper encoder/decoder

**AMI SOT recipe (`egs2/ami/sot_asr1/`):**
- Pipeline following ESPnet `asr.sh` convention (stages 1/5/11/12/13)
- Lhotse CutSet to Kaldi-format data preparation
- meeteval-based utterance-group cpWER evaluation
- Configs for Whisper-small and Whisper-tiny (testing)

**Whisper encoder/decoder fixes:**
- Encoder: derive `n_mels` from model instead of importing removed `N_MELS` constant, fixing v3/turbo compatibility (128 vs 80 mel bins)
- Decoder: handle positional embedding overflow for sequences exceeding 448 tokens

**Unit tests:** 22 tests covering model, preprocessor, task, and postprocessor.

---

## Why did you make this change?

SOT (Serialized Output Training) enables multi-talker ASR by serializing multiple speakers' transcripts into a single output sequence with speaker change tokens. This approach allows standard encoder-decoder models like Whisper to handle multi-talker speech without requiring separate diarization.

---

## Is your PR small enough?

This PR adds 34 files with ~3,800 lines. While above the typical guideline, this is a new recipe with a new task type that includes core pipeline components, an AMI recipe, and comprehensive unit tests. The components are tightly coupled (the recipe depends on the task, model, preprocessor, inference, and scorer), making it impractical to split without creating circular dependencies between PRs.

---

## Additional Context

- **Reference**: [Adapting Multi-Lingual ASR Models for Handling Multiple Talkers](https://arxiv.org/abs/2305.18747)

PR #6448, opened 2026-05-24
Add AMI SOT multi-talker ASR recipe with Whisper

cyhuang-tw:
## What did you change?

Adds a new Serialized Output Training (SOT) recipe for multi-talker ASR on
the AMI meeting corpus, using openai-whisper as the encoder/decoder backbone.
The model emits a single transcript with all speakers concatenated in FIFO
(speaker-start-time) order and separated by a `<sc>` token.

Recipe (`egs2/ami/sot_asr1/`):
- Training via the stock `asr.sh` pipeline (`run.sh --stage ...`).
- Inference against a checkpoint bundle via `run.sh --inference_model <dir>`.
- README with reproduction instructions, dependencies, and results.

Small espnet2 additions (required by the recipe):
- `espnet2/text/whisper_token_id_converter.py`: add `predict_timestamps`
  option to omit `<|notimestamps|>` from the decoder prefix.
- `espnet2/train/preprocessor.py`: thread `predict_timestamps` through
  `CommonPreprocessor_multi`.

A Whisper-small checkpoint trained with this recipe is available at
[`espnet/multi-talker-whisper-small-ami`](https://huggingface.co/espnet/multi-talker-whisper-small-ami).
Utterance-group-level results on AMI SDM test: cpWER 27.95% / DER 9.84%.

---

Extends ESPnet's SOT multi-talker ASR support to (a) the AMI meeting corpus
(a real conversational-meeting benchmark) and (b) Whisper as the underlying
encoder/decoder backbone, complementing the existing Transformer-based SOT
recipe at \`egs2/librimix/sot_asr1\`. Users can either train from scratch via
the stock \`asr.sh\` pipeline or reproduce the published numbers by decoding
against the released checkpoint.

---

## Is your PR small enough?

Yes — 16 files, less than 1000 lines added.

---

## Additional Context

- Supersedes (now closed) #6405.
- Released checkpoint: https://huggingface.co/espnet/multi-talker-whisper-small-ami
- Reference: *Adapting Multi-Lingual ASR Models for Handling Multiple Talkers* (https://arxiv.org/abs/2305.18747)

### DIFF
--- a/egs2/ami/sot_asr1/local/decode.py
+++ b/egs2/ami/sot_asr1/local/decode.py
@@ -1,15 +1,14 @@
 #!/usr/bin/env python3
 """Decode an ESPnet-format Whisper SOT checkpoint via ``openai-whisper``.
 
-The released ``model.pth`` is in ESPnet state-dict naming
+The released ``model.pth`` uses ESPnet state-dict naming
 (``encoder.encoders.*``, ``decoder.decoders.*``). We remap those keys back to
 openai-whisper's naming in memory and delegate inference to whisper's
 ``transcribe()`` pipeline (temperature fallback, compression-ratio /
 log-prob thresholds, KV cache, ``ApplyTimestampRules``).
 
-For SOT multi-talker output, the ``????`` separator is the regular BPE pair
-``????`` (token id 25629). It is rewritten to ``<sc>`` in the output to match
-the reference text format used by ``evaluate_sot.py`` / ``compute_der.py``.
+The SOT speaker-change separator emitted by the model is rewritten to
+``<sc>`` in the output text so downstream tooling can split on it cleanly.
 
 Usage::
 
@@ -27,17 +26,15 @@ from collections import OrderedDict
 from pathlib import Path
 from typing import Dict
 
-import numpy as np
-import soundfile as sf
 import torch
 import whisper
 
 logging.basicConfig(level=logging.INFO, format="%(asctime)s %(levelname)s: %(message)s")
 logger = logging.getLogger("decode")
 
-# tiktoken-multilingual encoding of "????" — emitted by the SOT model as the
-# speaker-change marker between speaker blocks.
-SOT_SEPARATOR_STR = "????"
+# SOT speaker-change separator: token id of a regular BPE pair in the
+# tiktoken multilingual vocabulary (already present in the base 51865-token
+# vocab). The string form is resolved at runtime via the tokenizer.
 SOT_SEPARATOR_ID = 25629
 LOG_EVERY = 100
 
@@ -46,14 +43,12 @@ LOG_EVERY = 100
 # SOT-aware replacement for whisper.decoding.ApplyTimestampRules
 # ---------------------------------------------------------------------------
 # Whisper's stock ``ApplyTimestampRules.apply`` enforces global timestamp
-# pairing and forbids any text after a closing timestamp, which prevents the
-# SST/FIFO model from emitting ``????`` between speaker blocks (the trained
-# pattern is ``<|t_end|> ???? <|t_start|>``). The patch below scopes the
-# pairing rules to the *current speaker block* (tokens after the last
-# separator) so the separator becomes a valid choice right after a closing
-# timestamp; this is a simplified port (no spk_count / spk_id auxiliary
-# tokens) of ``WhisperTimeStampLogitsProcessorCustom`` from the TS-ASR-Whisper
-# paper code (``src/models/dicow/utils.py``).
+# pairing and forbids any text token after a closing timestamp. That prevents
+# the SOT/FIFO model from emitting the speaker-change separator between
+# speaker blocks (its trained pattern is ``<|t_end|> <sep> <|t_start|>``).
+# The patch below scopes the pairing rules to the *current speaker block*
+# (tokens after the last separator) so the separator becomes a valid choice
+# right after a closing timestamp.
 
 
 def _install_sot_separator_patch() -> None:
@@ -153,8 +148,8 @@ def remap_espnet_to_whisper(
 ) -> "OrderedDict[str, torch.Tensor]":
     """Strip ESPnet wrapper prefixes (``encoder.encoders.``, ``decoder.decoders.``)
     and merge ``ExpandedTokenEmbedding`` back into a single flat tensor that
-    ``whisper.model.TextDecoder`` expects. ``????`` was trained as plain BPE
-    (id 25629, already in the base 51865 vocab), so the optional ``add_emb``
+    ``whisper.model.TextDecoder`` expects. The SOT separator was trained as a
+    plain BPE pair already in the base vocabulary, so the optional ``add_emb``
     row is unused and only ``ori_emb`` is kept."""
     out: "OrderedDict[str, torch.Tensor]" = OrderedDict()
     for k, v in espnet_state.items():
@@ -182,7 +177,7 @@ def load_whisper_model_from_espnet(
     # Do NOT call model.half() here even with fp16=True; whisper's transcribe()
     # casts inputs/cache internally and pre-halving conflicts with that.
     logger.info(f"Loading ESPnet weights from {espnet_pth}")
-    espnet_sd = torch.load(espnet_pth, map_location="cpu", weights_only=False)
+    espnet_sd = torch.load(espnet_pth, map_location="cpu", weights_only=True)
     missing, unexpected = model.load_state_dict(
         remap_espnet_to_whisper(espnet_sd), strict=False
     )
@@ -245,13 +240,15 @@ def main():
                 utts.append((parts[0], parts[1]))
     logger.info(f"Total utterances: {len(utts)}")
 
-    # Tokenizer to re-decode segment token IDs with timestamps preserved
-    # (needed for the text_sot output consumed by the DER evaluator).
+    # Tokenizer used to (a) re-decode segment token IDs with timestamps
+    # preserved for the text_sot output, and (b) resolve the SOT separator's
+    # string form from its token id (kept out of source for portability).
     sot_tokenizer = whisper.tokenizer.get_tokenizer(
         multilingual=True,
         task="transcribe",
         language=args.language,
     )
+    sep_str = sot_tokenizer.decode([SOT_SEPARATOR_ID])
 
     def _build_text_sot(result: dict) -> str:
         parts = []
@@ -259,7 +256,7 @@ def main():
             toks = list(seg.get("tokens", []))
             if toks:
                 parts.append(sot_tokenizer.decode_with_timestamps(toks))
-        return "".join(parts).replace(SOT_SEPARATOR_STR, " <sc> ")
+        return "".join(parts).replace(sep_str, " <sc> ")
 
     t_start = time.time()
     fail_count = 0
@@ -269,9 +266,7 @@ def main():
     ):
         for i, (uid, path) in enumerate(utts):
             try:
-                audio, _ = sf.read(path)
-                if audio.dtype != np.float32:
-                    audio = audio.astype(np.float32)
+                audio = whisper.load_audio(path)  # 16kHz float32 mono
                 result = model.transcribe(
                     audio,
                     language=args.language,
@@ -280,15 +275,14 @@ def main():
                     fp16=args.fp16,
                     word_timestamps=False,
                     condition_on_previous_text=False,
-                    # Disable max_initial_timestamp so the model can emit a
-                    # later first timestamp (matches paper-side HF generate()
-                    # behaviour). The default 1.0s would force <|0.00|> even
-                    # when speech starts >1s into the segment, inflating
-                    # false-alarm DER on short/silent-prefix utts.
+                    # Allow the first emitted timestamp to be anywhere in the
+                    # segment. The default 1.0s would force <|0.00|> even when
+                    # speech starts >1s into the segment, inflating
+                    # false-alarm DER on short/silent-prefix utterances.
                     max_initial_timestamp=None,
                 )
                 text = " ".join(
-                    result["text"].strip().replace(SOT_SEPARATOR_STR, " <sc> ").split()
+                    result["text"].strip().replace(sep_str, " <sc> ").split()
                 )
                 f_text.write(f"{uid} {text}\n")
                 f_text_sot.write(f"{uid} {_build_text_sot(result)}\n")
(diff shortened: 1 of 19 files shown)


egs2/ami/sot_asr1/local/decode.py
```diff
@@ -0,0 +1,312 @@
...
+import torch
+import whisper
+
+logging.basicConfig(level=logging.INFO, format="%(asctime)s %(levelname)s: %(message)s")
+logger = logging.getLogger("decode")
+
+# SOT speaker-change separator: token id of a regular BPE pair in the
+# tiktoken multilingual vocabulary (already present in the base 51865-token
+# vocab). The string form is resolved at runtime via the tokenizer.
+SOT_SEPARATOR_ID = 25629
```
simpleoier: Instead of hardcoding this token ID as an integer, let's derive it dynamically from the BPE model so the code doesn't break if we change models or the vocabulary updates or accidentally change the value.

egs2/ami/sot_asr1/local/decode.py
```diff
@@ -0,0 +1,312 @@
...
+# right after a closing timestamp.
+
+
+def _install_sot_separator_patch() -> None:
+    import whisper.decoding as _wd
+
+    if getattr(_wd.ApplyTimestampRules, "_sot_patched", False):
+        return
+
+    SEP = SOT_SEPARATOR_ID
```
simpleoier: Why not simply use `SOT_SEPARATOR_ID`?

egs2/ami/sot_asr1/local/prepare_sot.py
```diff
@@ -0,0 +1,294 @@
...
+
+from lhotse import CutSet
+
+logging.basicConfig(
+    level=logging.INFO,
+    format="%(asctime)s %(levelname)s: %(message)s",
+)
+logger = logging.getLogger(__name__)
+
+SPEAKER_CHANGE_TOKEN = " ???? "
```
simpleoier: Is the ` ???? ` correct?

simpleoier (commented):
LGTM! Thanks! I left some minor comments.

cyhuang-tw:
@simpleoier Thank you very much for your feedback. I have updated the code to derive the speaker-change separator from the model config file instead of hard-coding it. The data preparation and model configs now use consistent settings.

simpleoier (approved):
Thanks!

egs2/ami/sot_asr1/local/decode.py
```diff
@@ -0,0 +1,331 @@
...
+
+
+def load_whisper_model_from_espnet(
+    espnet_pth: Path, whisper_model_size: str, device: str
+) -> whisper.model.Whisper:
+    """Initialize the openai-whisper architecture matching ``whisper_model_size``,
+    then overwrite its weights with the converted ESPnet checkpoint."""
+    logger.info(f"Initializing whisper.{whisper_model_size} architecture...")
+    model = whisper.load_model(whisper_model_size, device="cpu")
+    # Do NOT call model.half() here even with fp16=True; whisper's transcribe()
```
LiChenda: Perhaps it would be better to output an INFO or a WARN here rather than a comment?

egs2/ami/sot_asr1/local/prepare_sot.py
```diff
@@ -0,0 +1,306 @@
...
+    if not getattr(cut, "recording", None):
+        return None
+
+    audio_path = next(
+        (src.source for src in cut.recording.sources if src.type == "file"), None
+    )
+    if audio_path is None:
+        return None
+
+    is_segment = cut.start > 0.01 or abs(cut.duration - cut.recording.duration) > 0.01
```
LiChenda: Is it better to change '0.01' to a configurable parameter?

LiChenda (approved):
I left a few comments. Overall, this PR looks good!

sw005320:
@cyhuang-tw, please reflect @LiChenda's comments

cyhuang-tw:
Thank you for your feedback, and sorry for the late reply. I have updated the code accordingly.

sw005320 (commented):
- Can you add speaker count accuracy?
- Please include the scoring scripts
- I think your updated code will affect pre-trained models, especially for the token dictionary. If so, you need to re-train the model (or directly edit the token ID on the model side) to be compatible with the recipe here.

egs2/ami/sot_asr1/run.sh
```diff
@@ -0,0 +1,75 @@
...
+        outdir="${inference_model}/${decode_out}/${tset}"
+        echo "[run.sh] Decoding ${tset} -> ${outdir}"
+        python local/decode.py "${inference_model}" \
+            --whisper_model "${whisper_model}" \
+            --wav_scp "data/${tset}/wav.scp" \
+            --out_subdir "${decode_out}/${tset}" \
+            ${fp16_flag}
+    done
+    exit 0
+fi
```
sw005320, line 76: What's this?
why you need two inference modes?

cyhuang-tw:
Thanks! The latest commits addressed all the points:
1. Speaker-count accuracy added in `local/evaluate_sot.py`.
2. Scoring scripts included and wired into run.sh.
3. We verified the released checkpoint stays compatible with the recipe, so no re-training is needed.
4. Removed the confusing two inference modes in the scripts.

sw005320 (commented):
Thanks @cyhuang-tw — the espnet2 side is clean and the scoring code is more careful than most recipe scorers I have read. I verified the cpWER arithmetic by exercising `edit_counts` and `group_cpwer` directly (Levenshtein backtracking, the square-padding that turns unmatched reference blocks into deletions, and the Hungarian assignment over `editdistance.eval` all check out), the RTTM field order matches the SCTK spec, and the `_sot_apply` timestamp-pairing logic faithfully mirrors upstream `ApplyTimestampRules` rescoped to the current speaker block. `cmd.sh` and the three `conf/*.conf` files are byte-identical to `egs2/TEMPLATE/asr1/`, which `ci/test_configuration_espnet2.sh` requires. No hardcoded personal or cluster paths anywhere, and nothing downloads or executes remote content.

On the espnet2 changes: `predict_timestamps` defaults to `False` so existing behaviour is preserved, the `sot_sequence` / `sot_sequence_including_notimestamps` switch is exactly the `<|notimestamps|>` prefix, and the only caller of `CommonPreprocessor_multi` (`espnet2/tasks/enh_s2t.py:475`) uses keyword arguments, so inserting the parameter mid-signature breaks nothing. `prepare_sot.py` does emit `<|0.00|>`-style tokens, so `predict_timestamps: true` in the config is coherent with the data. The one gap is test coverage: `test/espnet2/text/test_whisper_token_id_converter.py` has a natural home for this in `test_tokens2ids`, and right now nothing pins the new prefix.

The recipe itself has blocking problems.

---

## Blocking: `./run.sh` cannot start

**1. `path.sh`, `db.sh` and the `scripts` symlink are missing.** `egs2/TEMPLATE/asr1/setup.sh:44` defines the required set as `asr.sh path.sh db.sh scripts pyscripts steps utils`; this recipe ships only four of the seven. `asr.sh` runs under `set -e`, and:

- `asr.sh:287` does `run_args=$(scripts/utils/print_args.sh $0 "$@")` before option parsing. With no `scripts/`, this aborts immediately.
- `asr.sh:296` then does `. ./path.sh`, which would also fail — so `MAIN_ROOT`, the python environment and `tools/sctk/bin` never reach `PATH`.

`egs2/librimix/sot_asr1/` has all seven. Running `egs2/TEMPLATE/asr1/setup.sh` on the directory fixes this.

**2. `local/data.sh` does not exist.** `asr.sh:573-576` invokes `local/data.sh ${local_data_opts}` unconditionally at stage 1, and `run.sh` passes neither `--skip_data_prep true` nor a default `--stage`. A bare `./run.sh` therefore dies at stage 1. The README instead documents calling `local/prepare_sot.py` by hand. Either add a `local/data.sh` that wraps `prepare_sot.py`, or default `run.sh` to `--skip_data_prep true`.

**3. The documented training command skips stages it depends on.** README lines 63-66 give `./run.sh --stage 11 --stop_stage 11` under the heading "data prep already done". Stage 11 reads `${_asr_train_dir}/feats_type` (written by stage 3), the token list from stage 5, and the shape files from stage 10. A first-time reader following the README hits `cat: dump/raw/train/feats_type: No such file or directory`. The workable range is roughly `--stage 3 --stop_stage 11`.

---

## Correctness bugs

**4. `local/prepare_sot.py:38-39` silently truncates reference segments.**

```python
if merged and (sup.start - merged[-1]["end"]) <= max_timestamp_pause:
    merged[-1]["end"] = seg["end"]
```

The new end is assigned rather than maxed. Since supervisions are sorted by `start`, any same-speaker supervision nested inside a longer one shortens the merge. Verified: supervisions `A=[0,10]` and `B=[1,2]` produce a single segment `[0, 2]`, so eight seconds of reference speech disappear from the written timestamps. `score_der.py` then builds a reference RTTM missing that speech and scores the model's correct hypothesis as false alarm. Fix is `merged[-1]["end"] = max(merged[-1]["end"], seg["end"])`.

**5. `local/decode.py:284-290` writes window-relative timestamps into `text_sot`.** `_build_text_sot` renders `decode_with_timestamps(seg["tokens"])` and never reads `seg["seek"]` or `seg["start"]`. Whisper advances `seek` and re-decodes whenever `single_timestamp_ending` is false, which is normal after a temperature-fallback retry; segments from the second iteration carry timestamps relative to the new 30 s window, and whisper compensates only in the float `start`/`end` fields. So for a 20 s cut whose first decode ends at 6 s, the second block's `<|0.00|>` is written verbatim and `score_der.py` places that speaker at 0 s. DER is silently wrong. Offset by `seg["seek"] * HOP_LENGTH / SAMPLE_RATE`, or build the RTTM from `seg["start"]`/`seg["end"]`.

**6. `local/decode.py:88` pins EOT to a hardcoded logit.**

```python
if just_after_sep:
    logits[k, :ts0] = float("-inf")
    logits[k, eot] = 0.0
```

Line 108 restores the separator correctly with a saved `sep_orig[k]`; this one substitutes an arbitrary absolute `0.0` that competes directly against the timestamp logits. On any step where every timestamp logit is negative, EOT wins and decoding stops right after the speaker-change token, dropping every remaining speaker in the group. Two things compound it: the `continue` on line 91 does not protect the row, because the adaptive-sampling block is a second, unconditional `for k` loop that re-masks `logits[k, :ts0]` (EOT included) whenever timestamp confidence is high; and allowing EOT directly after a separator contradicts the trained `<sep> <|t_start|>` pattern. Either save an `eot_orig[k]` or leave EOT masked.

**7. `local/score_der.py:38,108` takes the first DER line, not the aggregate.** `_DER_RE` has no `(ALL)` anchor, and `write_rttm` emits one RTTM "file" per utterance group, so `md-eval.pl` prints a per-file block for each group before the final `*** ... for ALL ***` block. `re.search` returns the first. `egs2/TEMPLATE/asr1/scripts/utils/score_der.sh:39-41` deliberately greps and sorts to avoid exactly this. The reported DER would be that of one arbitrary utterance group, and it would look plausible. I could not run `md-eval.pl` to confirm the output shape, so please check against real output — but the missing anchor is unambiguous in the source.

**8. `local/decode.py:320-324` exits 0 after a completely failed run.** The blanket `except Exception` writes an empty hypothesis and continues, and `main()` returns `None`, so `sys.exit(main())` succeeds even when every utterance failed. `run.sh:69-73` then scores unconditionally and reports 100 % cpWER. A CUDA OOM or a wrong `--wav_scp` root produces a plausible-looking `cpwer.json` instead of a non-zero exit.

**9. The `by_num_speakers` buckets differ between the two scorers.** `score_der.py:65-68` counts a reference speaker only if its block yielded a paired, positive-length timestamp, while `evaluate_sot.py:85-86` counts any block with non-empty cleaned text. `prepare_sot.py:73-78` deliberately omits the closing timestamp on a cut-truncated final segment, so such a block is counted by cpWER but not by DER. The per-speaker-count columns in README lines 136-144 are therefore over different group sets.

**10. Smaller items.** `evaluate_sot.py:157` returns 0.0 cpWER for an empty-reference group (`group_cpwer([], ["x y"])` gives `ins=2, ref_len=0` → 0.0); aggregate cpWER is unaffected, only `per_utt.json` misleads. `prepare_sot.py:60-78` keys the "unfinished" flag to the last merged segment rather than the last emitted one, so the flag silently does nothing when the final segment has empty text. `score_der.py:102-113` never checks `proc.returncode`.

---

## Consistency

**11. `run.sh:38,93` breaks on bash 3.2.** `asr_args=()` expanded as `"${asr_args[@]}"` under `set -u` raises `unbound variable` on macOS's system bash. Fine on Linux CI. Use `${asr_args[@]+"${asr_args[@]}"}`.

**12. README line 118 claims `temperature=0.0`.** `decode.py:301-314` never passes `temperature`, so whisper uses its default fallback tuple — which the module docstring and `run.sh:13-15` both correctly advertise as a feature. The README contradicts them and makes the reported numbers non-reproducible as written.

**13. `run.sh:55-76` silently discards `asr_args` in inference mode**, so `--stage` is ignored without warning when `--inference_model` is set.

**14. `local/score_sot.sh:5` tells the reader to source `path.sh` first**, but the script never does and no `path.sh` exists (finding 1). In practice `score_der.py`'s `find_md_eval()` fallback to `<repo>/tools/sctk/bin/md-eval.pl` saves it.

---

## On the `????` speaker-change symbol

I initially read this as mojibake and was wrong — the README documents it as a deliberate base-vocabulary Whisper BPE token, and the design is internally consistent. `decode.py:290,316` rewrites it to `<sc>` on both outputs, and `evaluate_sot.py:49` / `score_der.py:35` both declare `_SEP_VARIANTS = ("<sc>", "????")` and normalize before splitting (verified by execution). `decode.py:255-259` hard-fails if the symbol does not encode to exactly one id, so a wrong symbol can never silently degrade decoding.

One hazard falls out of it, though. `run.sh:85` hardcodes `--sot_asr false`, which is correct **only** for a base-vocabulary separator: with `--sot_asr false`, `whisper_export_vocabulary.py:86` writes exactly 51865 lines, so the embedding is never expanded. `<sc>` is the default in every espnet2 API that touches this (`whisper_tokenizer.py:46`, `whisper_token_id_converter.py:29`, `whisper_export_vocabulary.py:27`) and the README's own prose describes the output in terms of `<sc>`, so switching the config to `<sc>` is the natural move for a user — and it yields token id 51865 against a 51865-row embedding, i.e. an opaque device-side assert on the first batch. Nothing ties the two settings together. A comment at `run.sh:85`, or an assertion that the configured symbol encodes to a single existing id, would prevent that.

Also worth confirming before merge, since the whole design rests on it: `whisper.tokenizer.get_tokenizer(multilingual=True).encode("????")` should return a single id.

Minor: `decode.py:290` replaces `" ???? "` with `" <sc> "`, producing doubled spaces. Harmless downstream.

---

The three failing `test_python_espnet2` jobs are HuggingFace decoder timeouts, unrelated to this diff.

PR #6459, opened 2026-06-05
[espnet3]: Sortformer diarization

popcornell:
@Masao-Someki  here is a simple porting of NeMo sortformer to ESPnet3. 

Synthetic data via https://github.com/popcornell/FastMSS/


sw005320:
@popcornell, one option is to move all (or most) sortformer-related files under `espnet3/systems/diar` instead of under `espnet2`
We want to have new codes to be under espnet3.
We also want to make each system more independent and modularized without via espnet2
(Actually, I'm thinking of even moving some of the existing files in espnet2 under espnet3)

egs3/diar_sortformer_cache/conf/inference.yaml
```diff
@@ -0,0 +1,81 @@
...
+#       --inference_config conf/inference.yaml \
+#       --metrics_config conf/metrics.yaml
+#
+# `model_file` defaults to the best trained checkpoint. To evaluate NVIDIA's
+# converted pretrained weights instead, point it at the .pth produced by
+#   python -m espnet2.diar.sortformer.convert_hf_sortformer \
+#       --hf_model nvidia/diar_sortformer_4spk-v1 --out exp/sortformer_4spk.pth
+
+dataset:
+  _target_: espnet3.components.data.data_organizer.DataOrganizer
```
Masao-Someki: You don't have to put this here, it will be ported from default config at `TEMPLATE/`

egs3/diar_sortformer_cache/conf/inference.yaml
```diff
@@ -0,0 +1,81 @@
...
+  recipe_dir: ${recipe_dir}
```
Masao-Someki: This one also you don't have to write

egs3/diar_sortformer_cache/conf/inference.yaml
```diff
@@ -0,0 +1,81 @@
...
+output_artifacts:
+  hyp:
+    type: npy
+  ref:
+    type: npy
+
+provider:
+  _target_: espnet3.systems.base.inference_provider.InferenceProvider
+runner:
+  _target_: espnet3.systems.base.inference_runner.InferenceRunner
```
Masao-Someki: These lines also you don't have to.
Basically you have to write it here when
1. you have your custom provider/runner
2. want to edit args like batch size for inference runner like
```
runner:
  batch_size: 4
```

egs3/diar_sortformer_cache/conf/training.yaml
```diff
@@ -0,0 +1,170 @@
...
+#     torch.save(m.encoder.state_dict(), 'exp/nest_encoder.pt')"
+
+task: espnet3.systems.diar.task.SortformerDiarizationTask
+
+##########################################################
+#               STAGE 1: DATA PREPARATION                #
+##########################################################
+# Generates FastMSS 90 s, 3-8-spk simulated meetings and builds AMI SDM cuts
+# (train + dev + test). Set the FastMSS checkout and a noise dir (WHAM/MUSAN).
+data_prep:
```
Masao-Someki: Is this a new config you added? (If not sorry we need to fix it in the espnet3 side)
We want to avoid unnecessary abbreviations and preferring full words, unless the abbreviated form is already widely established. Unless the original name is exceptionally long and the full form would become cumbersome to type or use repeatedly, I think readability is usually better with the full word.

([reference](https://masao-someki.github.io/espnet_draft_home_page/espnet3/contributing/naming-conventions.html#avoid-unnecessary-abbreviations))

espnet3/systems/diar/sortformer/inference_bin.py
```diff
@@ -0,0 +1,70 @@
...
+PyTorch-Lightning ``.ckpt`` (``model.`` prefix is stripped automatically).
+"""
+
+from typing import Optional, Union
+
+import numpy as np
+import torch
+import yaml
+
+from espnet3.utils.task_utils import get_espnet_model
```
Masao-Someki, line 31: Hmm, I'm not sure using espnet3 from espnet2 is a good idea. If I remember correctly, we plan to eventually archive espnet2, so touching espnet2 for this might not be worth it.

Masao-Someki (commented):
Thanks for the PR!
I have 2 comments overall,
1. Please check this [docstring guide](https://masao-someki.github.io/espnet_draft_home_page/espnet3/contributing/docstring-guide.html) to rewrite the docstrings. Current one is not sufficient. You can ask coding agent with [this prompt](https://masao-someki.github.io/espnet_draft_home_page/espnet3/contributing/docstring-guide.html#suggested-coding-agent-prompt-snippet)
2. I think touching `espnet2` is not a good idea. I think you can move model definition scripts under `espnet3/systems/diar/models/softformer`.

espnet3/systems/diar/sortformer/inference_bin.py
popcornell: i moved it into recipe most of the code !

popcornell:
@sw005320 I moved most of the recipe-specific code into the recipe. Common code such as diarization metric is in `espnet3/` now. 
Once we have reservation (I know it is for SLT but still...) I would like to try a large scale training (we have already DiariZen dataset + I want to use https://github.com/popcornell/FastMSS/ for additional 20k hours of synthetic data) so we can release the model as one of the first models in ESPnet3. 
I think we can reach SotA for diarization. 

@Masao-Someki, should I move also specific Sortformer code to the recipe ? I think so right ? It can always been imported in ESPnet3 in also other recipes.

egs3/diar_sortformer_cache/readme.md
```diff
@@ -0,0 +1,105 @@
...
+
+Model/library code is under `espnet3/systems/diar/sortformer/` and
+`espnet3/systems/diar/sortformer/model.py`; the task is
+`espnet3.systems.diar.task.SortformerDiarizationTask`. The 8-speaker loss uses
+**Hungarian PIL + arrival-time-argsort ATS** (brute-force `S!` is infeasible at
+8 speakers).
+
+## Stages
+
+```bash
```
popcornell, line 24: This is optional. I need to benchmark performance with XEUS, if it is good we can remove NEST dependency.

Masao-Someki:
Hi, based on our discussion on system naming rules, could you change the system name to `sortformer`?

PR #6469, opened 2026-06-13
[espnet3-21] Bugfix for Publication stage

Masao-Someki:
## What did you change?

Fixed a publication-stage bug that affected packed ESPnet3 models when they were used outside the original recipe directory.

The changes were:
- Changed packed-model loading to use `chdir`, following the approach used in ESPnet2 and the model zoo, so relative paths resolve correctly from the packaged bundle.

---

I first found this issue when a demo failed after being uploaded to Hugging Face Spaces.

While investigating, I found that some packed models still depended on files like `data/` or `dataset/` in the original environment, so they could fail after being moved to a different directory.

This change makes that failure easier to catch in integration tests and fixes the path resolution issue in packaged-model loading.

---

## Is your PR small enough?

yes

---

## Additional Context

* #6133 
* #6338 
* #6342


Masao-Someki:
ESPnet2 tests gets error with rate limit, espnet3 gets error with Dask version. These should be fixed with merging #6470 and #6467.

sw005320:
Thanks!
I'll move to the next one.

PR #6486, opened 2026-06-22
[espnet3-22] Add `clone` command in espnet3

Masao-Someki:
## What did you change?

I added a `clone` command so that people can start from existing recipe. You can clone the recipe under `egs3/` with:

```
$ espnet3 clone <recipe> --project <project-name>
```

Very detailed docstring is in [the code](https://github.com/Masao-Someki/espnet/blob/583e4875f3e8f524e16d7a69ff6f66d8034536b2/espnet3/cli/clone/command.py#L125)

This feature was mentioned in the [homepage](https://masao-someki.github.io/espnet_draft_home_page/espnet3/), in the "Clone a recipe, run in minutes" section.

---

Basically `egs3/` can also be a collection of official starter kits with this feature.
I thought this might also improve contributor's motivation, since their recipes would serve as actual starting points for others to build on.

When people wants to finetune some model, then they can first clone a recipe with that model, change `dataset/`, modify `src/` little to run fine-tuning, simple and quick.

---

## Is your PR small enough?

Yes
10 files, 3 main files to add this feature.

---

## Additional Context

* #6133 
* [ESPnet3 homepage](https://masao-someki.github.io/espnet_draft_home_page/espnet3/), in the "Clone a recipe, run in minutes" section.


Masao-Someki:
This feature has been tested with the integration test.
The integration test now first clones the mini-an4 recipe, then runs the tests.

sw005320:
Please fix the CI issue.
Can you also find someone as a reviewer?

espnet3/cli/clone/command.py
```diff
@@ -0,0 +1,349 @@
...
+from __future__ import annotations
+
+import argparse
+import logging
+import shutil
+from pathlib import Path
+
+logger = logging.getLogger(__name__)
+
+_INCLUDE = {"conf", "src", "dataset", "run.py", "readme.md", "path.sh"}
```
NewGamezzz: Do we actually need to copy `path.sh`? This can cause an error when the user clones the recipe and places it outside of the espnet3 recipe. Since the path.sh contains a relative path, it becomes invalid.

Masao-Someki: Thanks, and true path.sh might not be required as it depennds on how user innstalls espnet..!
I will remove it

ci/test_integration_espnet3.sh
```diff
@@ -7,15 +7,26 @@ set -euo pipefail
...
+
 python3 -m pip install -e '.[asr]'
 
-cd ./egs3/mini_an4/asr || exit
+clone_workdir=$(mktemp -d)
+espnet3 clone mini_an4/asr --project "${clone_workdir}/recipe"
+cd "${clone_workdir}/recipe" || exit
 gen_dummy_coverage
 echo "==== [ESPnet3] ASR ===="
 source path.sh
```
NewGamezzz, line 43: If you remove it, don't forget to update the test script so that it generates the correct `path.sh`. Otherwise, the test won't run successfully.

Masao-Someki: Thank you, I will fix it

espnet3/cli/clone/command.py
```diff
@@ -0,0 +1,371 @@
...
+    must therefore be filtered through the copytree ignore callback rather
+    than only at this top level.
+    """
+    dest.mkdir(parents=True)
+    for item in _INCLUDE:
+        source = src / item
+        if not source.exists():
+            continue
+        if source.is_symlink():
+            continue
```
NewGamezzz: Why did you avoid copying symbolic link files? This can cause an issue when a config file is a symbolic link to another file. For example, in the `mini_an4/asr` recipe, `training.yaml` is a symbolic link, so the clone command skips this file and causes the CI error.

espnet3/cli/clone/command.py
```diff
@@ -0,0 +1,371 @@
...
+
+logger = logging.getLogger(__name__)
+
+_INCLUDE = {"conf", "src", "dataset", "run.py", "readme.md"}
+
+_DESCRIPTION = """\
+Copy an egs3 recipe to a new directory so you can customise it without
+touching the original.
+
+Clones conf/, src/, dataset/, run.py, readme.md, and path.sh.
```
NewGamezzz, line 19: Please also fix the docstring. The current version still says that the `clone` command will copy `path.sh`.

espnet3/cli/clone/resolver.py
```diff
@@ -0,0 +1,83 @@
...
+    """
+    parts = recipe.strip("/").split("/")
+    if len(parts) != 2:
+        raise ValueError(
+            f"Recipe must be in '<dataset>/<task>' format, got: {recipe!r}"
+        )
+    dataset, task = parts
+    egs3 = _get_egs3_root()
+    recipe_path = egs3 / dataset / task
+    if not recipe_path.exists():
```
NewGamezzz, line 59: If the user accidentally provides a relative path (such as `../espnet3`), it will also pass this check. We should reject the relative path for this check.

espnet3/cli/clone/command.py
Masao-Someki: Thank you.
This was initially skipped because copying them as-is left absolute/relative paths in the original recipe. I added change so that it now convert those paths into relative path from the recipe root (and should work)

espnet3/cli/clone/resolver.py
Masao-Someki: Thanks, you're right, this is an issue. I added validation to ensure that the resolved recipe path is actually within egs3/ before accepting it.

Masao-Someki:
Hi @NewGamezzz if this looks ok to you, please approve the review!

NewGamezzz (approved):
Thanks. I took a look at it and looks good to me.

sw005320 (approved):

Masao-Someki:
Let me check, it seems one of the unit test got timeout issue.

espnet3/cli/clone/resolver.py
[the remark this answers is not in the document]
sw005320: @Masao-Someki, this must be considered.

test/espnet3/cli/test_clone.py
```diff
@@ -0,0 +1,587 @@
...
+
+
+def test_resolve_recipe_raises_value_error_for_no_slash():
+    with pytest.raises(ValueError, match="<dataset>/<task>"):
+        resolve_recipe("mini_an4")
+
+
+def test_resolve_recipe_raises_value_error_for_too_many_parts():
+    with pytest.raises(ValueError, match="<dataset>/<task>"):
+        resolve_recipe("a/b/c")
```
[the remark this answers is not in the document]
sw005320: ditto

espnet3/cli/clone/command.py
```diff
@@ -0,0 +1,349 @@
...
+    demo_yaml = dest / "conf" / "demo.yaml"
+    if demo_yaml.exists():
+        with demo_yaml.open("a", encoding="utf-8") as f:
+            f.write(
+                f"\n# Corpus/system identity injected by espnet3 clone.\n"
+                f"ui:\n"
+                f"  title: {corpus_system} demo\n"
+                f"upload_demo:\n"
+                f"  hf_repo: espnet/{corpus_system}_${{exp_tag}}\n"
+            )
```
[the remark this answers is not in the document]
sw005320: I don't know how it is serious, but please check it

egs3/mini_an4/asr/dataset/builder.py
```diff
@@ -115,7 +118,11 @@ def prepare_source(self, recipe_dir: str | Path, **_kwargs) -> None:
         dataset_root = recipe_root / _CFG["dataset_path"]
         archive = (recipe_root / _CFG["archive_path"]).resolve()
         if not archive.exists():
-            raise FileNotFoundError(f"Archive not found: {archive}")
+            url = _CFG.get("archive_url", "")
+            if not url:
+                raise FileNotFoundError(f"Archive not found: {archive}")
+            logger.info("Downloading archive | url=%s -> %s", url, archive)
+            urllib.request.urlretrieve(url, str(archive))
```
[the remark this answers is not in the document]
Masao-Someki: Since this is a recipe-specific builder, this does not introduce a meaningful security boundary here.

egs3/mini_an4/asr/dataset/config.yaml
```diff
@@ -2,6 +2,7 @@ builder:
   dataset_path: downloads
   data_path: data
   archive_path: ./downloads.tar.gz
+  archive_url: https://github.com/espnet/espnet/raw/refs/heads/master/egs3/mini_an4/asr/downloads.tar.gz
```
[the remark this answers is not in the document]
Masao-Someki: This IT test intentionally exercises the recipe's remote archive download path. The job is already network-dependent for environment setup, and copying the in-tree archive would remove coverage of `prepare_source()` download behavior.

Masao-Someki:
@sw005320 If this looks to you, let's merge this PR for `espnet3 clone` command!

sw005320:
One architectural question: could `clone` avoid hard-coding the ESPnet3 recipe structure and instead derive what to copy from the ESPnet3 recipe policy/schema?

My concern is that the ESPnet3 recipe policy is still evolving. If the policy changes later (e.g., required/optional directories or file conventions), a hard-coded file list in `clone` could continue generating projects based on the old policy unless `clone` is updated separately.

Ideally, I think the policy should be the single source of truth, and `clone` should consume that policy. This would also allow us to use the same policy for a future recipe validator/CI, so that we can change the ESPnet3 policy in one place and have both `clone` and validation automatically follow it.

I don't think we necessarily need to fully implement the policy framework in this PR, but I would like to avoid designing `clone` around a hard-coded recipe structure that we may have to maintain separately.

Masao-Someki:
I agree. Since an ESPnet3 recipe is already an importable egs3 package, clone should be able to reproduce that package directly.

I will simplify this by copying the recipe tree as-is, this should reduce maintenance and also newly added recipe files can be cloned automatically.

Masao-Someki:
Oh it seems coding agent made a huge mistake, let me fix it

PR #6487, opened 2026-06-23
[espnet3-23] Bugfix for preprocessor

Masao-Someki:
## What did you change?

DataOrganizer: auto-sets the train flag per split (True for train, False for valid/test) when an AbsPreprocessor is detected. Type detection uses _partial_=True to avoid calling __init__ before the flag is known.
  - Added _recursive_: false to TEMPLATE yamls so Hydra does not pre-instantiate the preprocessor sub-config.
  - Removed the manual flag override in lightning_module.collect_stats().
  - Removed train: true from recipe configs.
  - Added 6 unit tests.

---

DataOrganizer already knows which split it is building, so there is no reason to hard-code train: true in the config or manually override it at runtime.

---

## Is your PR small enough?

Yes
10 file changed, around 200 lines of change

---

## Additional Context

* #6133


espnet3/components/data/data_organizer.py
```diff
@@ -176,14 +177,70 @@ def __init__(
...
+                    train_preprocessor = instantiate(
+                        {**preprocessor_cfg, "train": True}
+                    )
+                    valid_preprocessor = instantiate(
+                        {**preprocessor_cfg, "train": False}
+                    )
+            else:
+                # Custom preprocessor from config: instantiate normally.
+                train_preprocessor = instantiate(preprocessor_cfg)
+                valid_preprocessor = train_preprocessor
```
Emrys365: For custom preprocessor, it might be better to also preserve an interface for differentiating train/valid modes.

espnet3/components/data/data_organizer.py
```diff
@@ -176,14 +177,70 @@ def __init__(
...
+            else:
+                train_preprocessor = preprocessor_cfg
+            valid_preprocessor = preprocessor_cfg
+            valid_preprocessor.train = False
+        else:
+            # Already-instantiated custom callable (non-AbsPreprocessor).
+            assert callable(preprocessor_cfg), "Preprocessor should be callable."
+            is_espnet_preprocessor = False
+            train_preprocessor = preprocessor_cfg
+            valid_preprocessor = preprocessor_cfg
```
Emrys365, line 336: ditto

Emrys365 (approved):
LGTM. Thanks a lot, @Masao-Someki ! I only left minor comments (in addition to Gemini code reviews).

Masao-Someki:
Hi @Emrys365 it would be great if you could check the updates, I basically reflected gemini and your review, and added a new integration test config to check whether different preprocessor setting for train/valid would work.

espnet3/components/data/data_organizer.py
```diff
@@ -176,46 +243,162 @@ def __init__(
...
+            # Already-instantiated AbsPreprocessor: deepcopy for the train
+            # split; keep the original as valid/test so callers that inspect the
+            # instance (e.g. test-only setups) see side-effects.
+            if train is not None:
+                train_preprocessor = copy.deepcopy(preprocessor_cfg)
+                train_preprocessor.train = True
+            else:
+                train_preprocessor = preprocessor_cfg
+            valid_preprocessor = preprocessor_cfg
+            valid_preprocessor.train = False
```
Emrys365: If `train` is None, will the line `valid_preprocessor.train = False` overwrites the `train` attribute of `train_preprocessor`? I'm not sure if this is expected.

espnet3/components/data/data_organizer.py
```diff
@@ -176,46 +243,162 @@ def __init__(
...
-
-                datasets.append(dataset)
-                transforms.append((transform, self.preprocessor))
-            return CombinedDataset(
-                datasets,
-                transforms,
-                use_espnet_preprocessor=is_espnet_preprocessor,
+        preprocessor_cfg = preprocessor
+        if preprocessor_cfg is None:
+            is_espnet_preprocessor = False
```
Emrys365: It seems that the `is_espnet_preprocessor` variable can be safely deleted now, since it is not actually used.

espnet3/utils/task_utils.py
```diff
@@ -5,14 +5,27 @@
...
+    if not isinstance(preprocess_config, dict) or "_target_" not in preprocess_config:
+        return False
+    partial_preprocessor = instantiate(preprocess_config, _partial_=True)
+    return (
+        hasattr(partial_preprocessor, "func")
+        and isinstance(partial_preprocessor.func, type)
+        and issubclass(partial_preprocessor.func, AbsPreprocessor)
+    )
+
+
```
Emrys365, line 28: In [`espnet3/components/data/data_organizer.py`](https://github.com/Masao-Someki/espnet/blob/e83fbe1f52344ab742e81645c631ba63661649df/espnet3/components/data/data_organizer.py#L257), it is assumed that `preprocess_config` can be one of:
+ None
+ dict or DictConfig
+ AbsPreprocessor

Would it be better to also consider these possibilities in this function?

espnet3/utils/task_utils.py
```diff
@@ -71,10 +83,17 @@ def save_espnet_config(
...
     if dataset_config is not None and "preprocessor" in dataset_config:
-        preprocess_config = dataset_config.pop("preprocessor")
-        if "_target_" in preprocess_config:
-            preprocess_config.pop("_target_")
-        default_config.update(preprocess_config)
+        preprocess_config = dataset_config["preprocessor"]
+        has_split_configs = isinstance(preprocess_config, dict) and any(
+            key in preprocess_config and isinstance(preprocess_config[key], dict)
+            for key in ("train", "valid", "test")
+        )
```
Emrys365, line 90: Maybe `isinstance(xxx, (dict, DictConfig))` is more robust?

Emrys365 (approved):
Sorry for the late response. I totally missed the message...

The updated code generally looks good! I just left a few minor comments.

espnet3/components/data/data_organizer.py
Masao-Someki: Thanks! I also realized that we can remove `use_espnet_preprocessor` safely. I will update on this part and try to make it clear on this if-else block..

espnet3/components/data/data_organizer.py
Masao-Someki: Thank you, that is true it overrides, it was a bug (but no issue as when train is None, validation set is also None)
I will update this to clearer snippet:

```
        elif isinstance(preprocessor_cfg, AbsPreprocessor):
            # Keep the original instance for non-training splits so callers can
            # inspect its side effects.
            if train is not None:
                train_preprocessor = copy.deepcopy(preprocessor_cfg)
                train_preprocessor.train = True
                valid_preprocessor = preprocessor_cfg
                valid_preprocessor.train = False
            else:
                train_preprocessor = do_nothing
                valid_preprocessor = do_nothing
            test_preprocessor = preprocessor_cfg
            test_preprocessor.train = False
```

espnet3/utils/task_utils.py
Masao-Someki: Thanks!
I think the answer is no, because this function is used to check if the preprocessor config is for AbsPreprocessor when saving config for experiment. (`exp/<exp_tag>/config.yaml`)

Since the `preprocessor_config` is a resolved config, it's a python `dict` object.

espnet3/utils/task_utils.py
Masao-Someki: Thanks, as mentioned above it should be dict object, so `DictConfig` is not required!

Issue #6294, opened 2025-11-10
Batch type bug


fsepteixeira:
When running the default recipe for VoxCeleb, with conf/train_ECAPA_wavlm_joint.yaml, I get the following error at the start of the 2nd epoch:

2025-11-07 03:42:09,503 (trainer:324) INFO: 2/20epoch started. Estimated time to finish: 1 day, 13 hours and 30 minutes                                                                                                                 
Traceback (most recent call last):                                                                                                                                                                                                            
  File "<frozen runpy>", line 198, in _run_module_as_main                                                                                                                                                                                     
  File "<frozen runpy>", line 88, in _run_code                                                                                                                                                                                                
  File "espnet/espnet2/bin/spk_train.py", line 27, in <module>                                                                                                                                                                   
    main()                                                                                                                                                                                                                                    
  File "espnet/espnet2/bin/spk_train.py", line 23, in main                                                                                                                                                                       
    SpeakerTask.main(cmd=cmd)                                                                                                                                                                                                                 
  File "espnet/espnet2/tasks/abs_task.py", line 1281, in main                                                                                                                                                                    
    cls.main_worker(args)                                                                                                                                                                                                                     
  File "espnet/espnet2/tasks/abs_task.py", line 1638, in main_worker                                                                                                                                                             
    cls.trainer.run(                                                                                                                                                                                                                          
  File "espnet/espnet2/train/trainer.py", line 347, in run                                                                                                                                                                       
    iterator=train_iter_factory.build_iter(iepoch),                                                                                                                                                                                           
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^                                                                                                                                                                                            
  File "espnet/espnet2/iterators/category_iter_factory.py", line 115, in build_iter                                                                                                                                              
    raise ValueError(f"Unsupported batch_type: {self.batch_type}")                                                                                                                                                                            
ValueError: Unsupported batch_type: folded 

After checking the batch types in category_iter_factory.py, I tried to put as an argument in the conf file
batch_type: catbel

Which raised the following error:
2025-11-10 11:35:16,170 (abs_task:2051) INFO: [train] mini-batch sizes summary: N-batch=116177, mean=32.0, min=32, max=32
Traceback (most recent call last):
  File "<frozen runpy>", line 198, in _run_module_as_main
  File "<frozen runpy>", line 88, in _run_code
  File "espnet/espnet2/bin/spk_train.py", line 27, in <module>
    main()
[...]

PR #6490, opened 2026-06-29
[Fix] CategoryIterFactory built with unmapped 'folded' batch_type (#6294)

yushuosun:
## Motivation

Category-based iteration (e.g. VoxCeleb `spk1` recipes) is built with the wrong `batch_type` when the config uses the legacy default `folded` (#6294).

## Root cause

In `espnet2/tasks/abs_task.py`, `build_category_iter_factory` maps `folded -> catbel` for the epoch-1 sampler:

```python
type=(iter_options.batch_type if iter_options.batch_type != "folded" else "catbel"),
```

but then constructs the `CategoryIterFactory` with the **unmapped** value:

```python
batch_type=iter_options.batch_type,
```

So the factory receives `folded` instead of the intended `catbel`.

## Modifications

`espnet2/tasks/abs_task.py`: apply the same `folded -> catbel` normalization to the `batch_type=` argument passed to `CategoryIterFactory`, matching the mapping already used ~18 lines above for the sampler `type=`.

## Duplicate-check

- Track A — `gh api 'repos/espnet/espnet/pulls?state=open&per_page=100' --paginate` grepped for `6294` / `batch_type` → no open PR references it.
- Track B — issue #6294 has no in-progress claim.

### DIFF
--- a/espnet2/tasks/abs_task.py
+++ b/espnet2/tasks/abs_task.py
@@ -2086,7 +2086,7 @@ class AbsTask(ABC):
                 seed=args.seed,
                 num_iters_per_epoch=iter_options.num_iters_per_epoch,
                 sampler_args=sampler_args,
-                batch_type=iter_options.batch_type,
+                batch_type=(iter_options.batch_type if iter_options.batch_type != 'folded' else 'catbel'),
                 shuffle=iter_options.train,
                 num_workers=args.num_workers,
                 collate_fn=iter_options.collate_fn,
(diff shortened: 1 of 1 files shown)


sw005320:
Thanks for the report and for making a PR, @yushuosun!
@Jungjee and @Emrys365, can you check it?

Emrys365 (approved):
I think this is a bug because the batch sampler should be consistently configured across epochs.

sw005320:
OK, thanks, @yushuosun for the PR and @Emrys365 for reviewing it!

Jungjee:
Thanks @yushuosun for the fix. This was a bug.

PR #6494, opened 2026-07-15
Add TER metric to espnet3

Nithish-Sri-Ram:
## What did you change?

Added a TER (token error rate) metric to the espnet3 ASR system (`espnet3/systems/asr/metrics/ter.py`), following the existing `cer.py`/`wer.py` pattern, plus tests and a commented config entry in the TEMPLATE metrics.yaml (kept commented for now since it needs a recipe-specific bpemodel, happy to uncomment or adjust it if you'd prefer).

espnet3's ASR system reported WER and CER but not TER, while espnet2's Stage 13 scoring reports all three (`char`→CER, `word`→WER, `bpe`→TER). This adds parity. TER tokenizes ref and hyp with the SentencePiece model and scores like WER over the resulting subword tokens (consistent with how espnet3's CER/WER use jiwer). Approved by @sw005320 in #6492 (comment) and @Masao-Someki in #6133 .

## Is your PR small enough?
Yes - 1 new source file + tests + one config entry.

## Additional Context
TER is BPE-specific and opt-in per recipe (needs a `bpemodel`), matching espnet2, which only computes `ter` when a bpemodel exists.


espnet3/systems/asr/metrics/ter.py
```diff
@@ -0,0 +1,107 @@
...
+        Args:
+            text: Input text.
+
+        Returns:
+            Space-joined subword tokens, or a placeholder to avoid empty inputs.
+        """
+        cleaned = self.cleaner(text).strip()
+        if not cleaned:
+            return "."
+        return " ".join(self.tokenizer.text2tokens(cleaned))
```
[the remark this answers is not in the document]
Nithish-Sri-Ram: Thanks! Applied the _tokenize fix - it now falls back to the "." placeholder when tokenization yields no tokens, matching the empty-text handling.

On the sentencepiece lazy-import suggestions: sentencepiece is a core dependency (sentencepiece==0.2.1 in [project.dependencies], not an optional extra), so it's always installed and the top-level import can't fail collection, this matches the existing test/espnet2/text/test_sentencepiece_tokenizer.py, which imports it at top level too. jiwer is the genuinely optional one ([asr] extra), and that's already guarded. So I've kept the sentencepiece imports as they are.

sw005320:
Thanks, @Nithish-Sri-Ram!

@Masao-Someki, please review this PR.

Nithish-Sri-Ram:
Hi @Masao-Someki, after this is done - I'd like to take on something bigger in the ESPnet3 migration. Is there a component or task you'd like me to work on? Happy to take direction.

sw005320:
@Masao-Someki, any update?

egs3/TEMPLATE/asr/conf/metrics.yaml
```diff
@@ -43,3 +43,13 @@ metrics:
...
+
+# TER (token error rate) scores on the model's subword tokens, so it needs the
+# recipe's trained SentencePiece model. It is recipe-specific and off by default.
+# To enable it, uncomment below and point `bpemodel` at your trained model:
+# - metric:
+#     _target_: espnet3.systems.asr.metrics.ter.TER
+#     bpemodel: exp/bpe_5000/bpe.model  # path to the trained SentencePiece model
+#     ref_key: ref        # reference text key -> `<test_name>/ref.scp`
+#     hyp_key: hyp        # hypothesis text key -> `<test_name>/hyp.scp`
+#     clean_types:        # optional TextCleaner pipeline
```
Masao-Someki, line 55: Could you remove the comment out and turn it on?

Masao-Someki (commented):
Thank you @Nithish-Sri-Ram, I've checked this PR and looks good!
One point, let's remove the `#` in the config so that we compute TER as a default setting.

egs3/TEMPLATE/asr/conf/metrics.yaml
Nithish-Sri-Ram: @Masao-Someki  Sounds good, one thing first, bpemodel isn't auto-filled in like exp_tag/exp_dir are, so if it points to a path that doesn't exist, measure crashes before WER/CER even get written (not just TER failing).

Want me to turn it on as-is uncommented (and treat editing the path as required, like the readme already says), or should I first make bpemodel auto-fill from the training config so it works without manual edits?

Masao-Someki: Sorry for late replay. And you're right, let's keep it as commented-out.

Originally, this was designed to be more flexible so that one config could refer to values from another config more freely. So previously we could refer to the training config value from that metrics config.

However, we ran into several issues during config resolution, and after spending a fair amount of time on it, we ended up handling the required values in a more hard-coded way..

I'd like to revisit this part and see if we can make cross-config value propagation more general and robust in future PR.

Nithish-Sri-Ram: Sounds good, keeping it commented for now. Happy to help with the cross-config propagation piece whenever you want to pick it up

sw005320:
@Masao-Someki, if you're OK with this change, please approve it.

Masao-Someki (approved):

PR #6503, opened 2026-07-27
Add Bagpiper SpeechLM recipe (egs2/bagpiper/speechlm1), plus FSDP-gradient / Liger-loss / weight-only-checkpoint fixes

jctian98:
## What did you change?

**Main topic — a new recipe: `egs2/bagpiper/speechlm1`.**
Bagpiper is a multimodal **SpeechLM**: a Qwen3-8B(-Base) backbone extended to both
*understand* audio (speech / sound / music → text) and *generate* audio
(text → speech / sound / music), trained jointly with text-only data. The recipe
ships a minimal, production-oriented layout — the three `launch_bagpiper_stage{1,2,3}_*.sh`
drivers (length-stats → train → inference), the base train configs, the pretrain/SFT
inference configs, the env skeleton (`cmd.sh` / `db.sh` / `path.sh` / `local/path.sh` / `utils`),
and a `README.md`.

**Plus three `espnet2/speechlm` fixes that the recipe depends on:**

1. **lm_head / stream-emb gradients were not reduced under FSDP** (`loss.py`, `parallel.py`).
   The `lm_head` weight is consumed directly in the fused loss rather than through its
   module's `forward`, so FSDP2 never armed its gradient reduce-scatter — each rank silently
   kept only its local slice (tagged `Replicate`), dropping the other ranks' contributions.
   We now consolidate with `full_tensor(grad_placements=[Partial("avg")] * mesh.ndim)` so the
   backward reduces (mean) the gradient across the DP mesh; one `Partial("avg")` per mesh dim
   covers both FSDP (1-D) and HSDP (2-D) meshes. The tied stream embeddings get the same fix in
   `parallel.py`.

2. **Liger fused-loss version compatibility** (`loss.py`).
   `LigerFusedLinearCrossEntropyFunction.apply(...)` returns a variable-length tuple across
   Liger releases (older: `loss, z_loss, accuracy`; newer releases append more values). We slice
   `[:3]` at both call sites so we take exactly the three values we use, instead of unpacking
   errors on newer Liger.

3. **Weight-only checkpoint init on the Titan backend** (`titan_trainer.py`).
   Training can now be initialized from a single consolidated `.pt` weight file — the same file
   used for inference — via
   `set_model_state_dict(..., StateDictOptions(full_state_dict=True, broadcast_from_rank0=True, strict=False))`,
   instead of requiring a full distributed/optimizer checkpoint.

---

We want to contribute the Bagpiper recipe so the community can train and run the multimodal
SpeechLM. The three fixes are prerequisites for it to train and load correctly:
(1) otherwise the tied output/embedding gradients are under-counted under FSDP/HSDP;
(2) makes the fused-loss path robust across Liger versions;
(3) lets a single published checkpoint be reused for both inference and continued training.

---

## Is your PR small enough?

17 files, ~1.27k lines — under the 20-file limit, slightly over the 1000-line guideline.
The bulk is a self-contained new recipe directory (no existing code is touched); the three
library fixes total ~60 lines. Happy to split the recipe and the fixes into separate PRs if
maintainers prefer — noting the recipe won't train/load correctly without the fixes.

---

## Additional Context

The public stage-2 pretrained checkpoint (Qwen3-8B-Base backbone) is on the Hugging Face Hub
dataset `JinchuanTian/Bagpiper_checkpoints`; `conf/train_stage2_qwen3_base.yaml` reconstructs the
model behind it. See the recipe `README.md` for environment (flash-attention-3, TorchTitan),
the three-stage workflow, and inference usage.


jctian98:
ignore this PR, it will not be really merged. just for tentative code/resources sharing.

will start other PRs to merge content with clearer divisions.

PR #6506, opened 2026-08-04
Update VERSA links to new org name (shinjiwlab -> wavlab-speech)

ColebyPearson:
The versa repo moved from the shinjiwlab org to wavlab-speech. These 9 links still point at the old name and only work because GitHub redirects renamed orgs. install_versa.sh actually clones from the old URL, so seemed worth updating before that redirect ever breaks.

Found while fixing some dead links in versa's own docs (wavlab-speech/versa#88).


sw005320 (approved):

sw005320:
Thanks for catching the error!

PR #6507, opened 2026-08-04
Add Nahuatl S2T recipe (egs2/nahuatl/s2t1) — OWSM v4 fine-tuning

eric102004:
## What did you change?

Added a new recipe `egs2/nahuatl/s2t1` that fine-tunes OWSM v4 (`espnet/owsm_v4_medium_1B`) on a Nahuatl speech corpus. The recipe delegates to the standard `TEMPLATE/s2t1/s2t.sh` and adds:

- `local/data_prep.py` + `local/data.sh` — convert the (locally prepared) HuggingFace-format dataset to Kaldi-style data dirs, including the `text.prev` / `text.ctc` streams the OWSM S2T objective expects, and merge all splits.
- `local/init_new_tokens.py` — extend the OWSM v4 BPE token list with three region tags and patch the pretrained checkpoint's embedding / output layers to match.
- `conf/tuning/train_owsm_v4_nahuatl.yaml` + `conf/decode_owsm.yaml` — training and decoding configs.
- `run.sh` (a thin wrapper over `s2t.sh`), plus `debug_train.sh` / `full_train.sh` launch helpers and a `test_data_prep.py` unit test.

---

Nahuatl is a low-resource indigenous Mexican language with no existing ESPnet recipe. This adds a reproducible fine-tuning pipeline for OWSM v4 on a 3-dialect Nahuatl corpus (Hidalgo, Orizaba-Zongolica, Zacatlan-Tepetzintla). The underlying recordings are **not publicly distributed**; they are long-form field recordings with time-aligned transcriptions (Transcriber `.trs` / ELAN `.eaf`) that we convert locally into a HuggingFace-readable `{audio, text}` dataset. Each utterance is tagged with a dialect token (`<nah_hid>`, `<nah_ozg>`, `<nah_ztp>`) so a single model can condition on region. Recordings are split per region into train / validation / test (roughly 80 / 10 / 10). The split is speaker-disjoint: each recording is an interview between a primary consultant and a recurring field interviewer, so we keep only the primary consultant's turns and assign all of a consultant's recordings to the same split — no speaker's voice appears in more than one split.

---

## Is your PR small enough?

Yes — 17 files changed, 742 insertions, 0 deletions (a self-contained new recipe; no existing files are touched).


eric102004:
@coderabbitai full review

eric102004:
@coderabbitai full review

eric102004:
@coderabbitai full review

egs2/nahuatl/s2t1/.gitignore
sw005320, line 1: please remove this

egs2/nahuatl/s2t1/debug_train.sh
sw005320, line 1: please remove the debug files

eric102004:
@coderabbitai full review

sw005320 (commented):
Thanks @eric102004 — a Nahuatl recipe is a genuinely valuable addition, and the care in the data split is the right kind of care: keeping only the primary consultant's turns and assigning all of a consultant's recordings to one split is exactly what stops a speaker leaking across splits. The dialect conditioning is a reasonable design too.

My concern is that this is fine-tuning, and it has grown much larger than the other fine-tuning recipes it sits next to. Measured against the other `s2t1` recipes:

| recipe | `run.sh` | extra top-level scripts | `local/` |
| --- | --- | --- | --- |
| `allsstar_eng/s2t1` | 26 | — | `data.sh`, `download_and_untar.sh`, `path.sh` |
| `must_c_v2/s2t1` | 36 | — | `data.sh`, `data_prep.py`, `path.sh` |
| `owsm_v3.1/s2t1` | 38 | — | `path.sh` |
| `multimed_st/s2t1` | 91 | — | 6 files, incl. OWSM v4 asset prep and `score.sh` |
| **this PR** | **166** | **`debug_train.sh`, `full_train.sh`, `decode.sh`** | **6 files, incl. 206 lines of pytest** |

The PR description calls `run.sh` "a thin wrapper over `s2t.sh`". At 166 lines it is four times the length of any other one. Most of that is not Nahuatl-specific — it is working around `s2t.sh`, and in one case `s2t.sh` already supports what is being worked around.

## The `feats_stats.npz` workaround can be two flags

This is the biggest one. `run.sh` currently splits the `s2t.sh` invocation in two so it can drop a symlink to OWSM's `feats_stats.npz` between stage 10 and stage 11, because `collect_stats` would otherwise overwrite it through the symlink. That is about thirty lines of code and explanation, and it is fragile — it depends on `aggregate_stats_dirs` write ordering.

`s2t.sh` already takes `--s2t_stats_dir` (line 100), and stage 11 reads the statistics straight from it:

```bash
# egs2/TEMPLATE/s2t1/s2t.sh:1330
_opts+="--normalize=global_mvn --normalize_conf stats_file=${s2t_stats_dir}/train/feats_stats.npz "
```

So pointing it at OWSM's stats directory and skipping stage 10 does the same thing with nothing to overwrite:

```bash
./s2t.sh --s2t_stats_dir "${owsm_stats_dir}" --skip_stages 10 ...
```

One `s2t.sh` call, no symlink, no ordering assumption.

## Three launcher scripts that `s2t.sh` already replaces

`debug_train.sh`, `full_train.sh` and `decode.sh` are 120 lines between them, and no other `s2t1` recipe has anything like them. `s2t.sh` already has `--stage` / `--stop_stage`, and your own `run.sh` header documents the debug path:

```
max_epoch=1 num_iters_per_epoch=20 log_interval=1 bash run.sh --stage 11 --stop_stage 11
```

so `debug_train.sh` duplicates a line that is already in the file. `decode.sh` has a real reason to exist — `--lang_sym` is global per run, so the three regions have to be decoded separately — but that belongs inside `run.sh` as a loop over `test_sets`, not as a fourth entry point a user has to discover.

## The tests belong under `test/`

`local/test_data_prep.py` and `local/test_data_prep_collision.py` are 206 lines of pytest living inside a recipe. ESPnet's CI collects from `test/`, so these never run — they are documentation that looks like a safety net. Either move them under `test/espnet2/` where CI will run them, or drop them and let `local/data_prep.py` be reviewed on its own.

## There is no README

Every comparable recipe has one — `allsstar_eng`, `multimed_st`, `owsm_v3.1` all do. It matters more here than usual, because the corpus is not publicly distributed. Right now the only description of how to obtain and shape the data is a comment block at the top of `run.sh`, including a six-argument `init_new_tokens.py` command with `../../../../model_cache/` paths that a reader has to copy by hand. That should be a README section, and the paths should come from `db.sh` like every other corpus location in `egs2/`.

## A question rather than a request: are the new tokens needed?

`local/init_new_tokens.py` is 185 lines, and it is the root of several other complications — the patched BPE model, the token-list existence check in `run.sh`, and the manual prerequisite step. It exists to add `<nah_hid>`, `<nah_ozg>`, `<nah_ztp>` to the vocabulary and resize the embedding and output layers.

OWSM's S2T objective already has a `text.prev` stream for conditioning, and `local/data_prep.py` already generates it. Did you try putting the dialect tag in the prompt instead of the vocabulary? If that works, `init_new_tokens.py`, the BPE patch and the checkpoint surgery all go away, and the recipe becomes an ordinary fine-tune. If it was tried and did not work, that is worth a sentence in the README — it is the first thing the next reader will wonder.

---

None of this is about the modelling or the data work, which look sound. A recipe this size will be read as the reference for "how do I fine-tune OWSM on my own language", so it is worth getting it down to the shape of the ones beside it.

eric102004:
@coderabbitai full review

egs2/nahuatl/s2t1/debug_train.sh
eric102004: Done in 6b4fd339. Removed the launch wrappers (`debug_train.sh`, `full_train.sh`) and the recipe-local `.gitignore`. Training now runs through `run.sh` (debug via the `max_epoch` / `num_iters_per_epoch` env overrides documented in its header) and evaluation through `decode.sh`; scheduler and account options are passed at submit time.

PR #6509, opened 2026-08-05
[espnet3-24] Bugfixed AverageCheckpointsCallback

Masao-Someki:
## What did you change?

Fixed `AverageCheckpointsCallback` so that it correctly averages checkpoints whose state-dict keys do not have a `model.` prefix. The callback now compares checkpoint keys with the current model's `state_dict()` keys and raises a clear error when they are incompatible.

Added tests for prefix-free checkpoints and key mismatches.

---

`ESPnetLightningModule.state_dict()` exposes the model's keys without a `model.` prefix, while the callback previously always removed that prefix. As a result, valid model parameters could be dropped when averaging checkpoints.

---

## Is your PR small enough?

yes
2 files  with 139 additions and 14 deletions

---

## Additional Context

* #6133 
* #6418


sw005320 (approved):

NewGamezzz (commented):
This PR looks good to me. I have no major concerns, only one minor suggestion about the error message.

espnet3/components/callbacks/default_callbacks.py
```diff
@@ -161,12 +169,26 @@ def on_validation_end(self, trainer, pl_module):
...
+                    new_avg_state_dict = {
+                        (k.removeprefix("model.") if k.startswith("model.") else k): v
+                        for k, v in avg_state_dict.items()
+                    }
+                    if set(new_avg_state_dict.keys()) != reference_model_keys:
+                        raise KeyError(
+                            "Averaged checkpoint keys do not match the "
+                            "current model's state_dict keys.\n"
+                            f"Expected: {reference_model_keys}\n"
+                            f"Got: {set(new_avg_state_dict.keys())}"
```
NewGamezzz: I think showing only the mismatched keys (missing/unexpected keys) would be more useful for debugging. Printing the full expected and got keys can be noisy/messy, in my opinion.

sw005320: @Masao-Someki, can you respond to this?

Masao-Someki: Oh sorry and thanks!
True, I'll fix that part

NewGamezzz (approved):
Thanks! This PR looks good to me now. Btw, there is an issue with CI, but I think it does not come from our code.

sw005320:
thanks, @Masao-Someki!

PR #6513, opened 2026-08-07
Fix `data_aug_effects` nesting in `egs3/librispeech_100` e-branchformer config

### LINKED ISSUE #6512: `egs3/librispeech_100`: `data_aug_effects` has an extra nesting level and crashes training
**Describe the bug**

`egs3/librispeech_100/asr/conf/tuning/training_e_branchformer.yaml` nests `data_aug_effects` one level too deep, so training crashes on the first augmented utterance.

```yaml
    data_aug_effects:
      - - [0.5, "speed_perturb", {"factor": 0.9}]
        - [0.5, "speed_perturb", {"factor": 1.1}]
```

This is a one element list. `DataAugmentation.__init__` reads the first element of each entry as its sampling weight (`espnet2/layers/augmentation.py:83`), so `effect_probs` becomes `([0.5, "speed_perturb", {"factor": 0.9}],)` and `weighted_sample_without_replacement` then evaluates `1 / w` on a list (`espnet2/layers/augmentation.py:28`).

The entry matches neither documented form. Per the `DataAugmentation` docstring, type1 is `[weight, name, kwargs]` and type2 is `[weight, [type1, ...]]`. Here the two effects are wrapped in a list with no leading weight, so it reads either as a type2 entry missing its group weight, or as two type1 entries nested one level too deep. Either reading gives a valid fix.

**Basic environments:**
- OS information: `Linux 6.8.0-124-generic #124~22.04.1-Ubuntu SMP PREEMPT_DYNAMIC Tue May 26 21:05:19 UTC x86_64`
- python version: `3.10.16 (main, Dec  6 2024, 19:59:16) [Clang 18.1.8]`
- pytorch version: `pytorch 2.6.0+cu124`
- Git hash: `216aac90f47b6fd872c51bd55d7d9a974f15d1a7`
  - Commit date: `Wed Aug 5 09:53:27 2026 -0400`

**Task information:**
- Task: ASR
- Recipe: `egs3/librispeech_100`
- ESPnet3

**To Reproduce**

```
cd egs3/librispeech_100/asr
python run.py --stages train --training_config conf/tuning/training_e_branchformer.yaml
```

Or without the dataset:

```python
import numpy as np, yaml
from espnet2.layers.augmentation import DataAugmentation

conf = "egs3/librispeech_100/asr/conf/tuning/training_e_branchformer.yaml"
effects = yaml.safe_load(open(conf))["dataset"]["preprocessor"]["data_aug_effects"]

aug = DataAugmentation(effects)
print("effect_probs:", aug.effect_probs)
aug(np.zeros(16000, dtype=np.float32), 16000)
```

**Error logs**

```
effect_probs: ([0.5, 'speed_perturb', {'factor': 0.9}],)
  File "repro.py", line 10, in <module>
    aug(np.zeros(16000, dtype=np.float32), 16000)
  File "espnet2/layers/augmentation.py", line 96, in __call__
    for effect in weighted_sample_without_replacement(
  File "espnet2/layers/augmentation.py", line 28, in weighted_sample_without_replacement
    v = [rng.random() ** (1 / w) for w in weights]
  File "espnet2/layers/augmentation.py", line 28, in <listcomp>
    v = [rng.random() ** (1 / w) for w in weights]
TypeError: unsupported operand type(s) for /: 'int' and 'list'
```

I will open a PR with a fix.

nagohachi:
## What did you change?

Removed one level of nesting from `data_aug_effects` in `egs3/librispeech_100/asr/conf/tuning/training_e_branchformer.yaml`.

```diff
     data_aug_effects:
-      - - [0.5, "speed_perturb", {"factor": 0.9}]
-        - [0.5, "speed_perturb", {"factor": 1.1}]
+      - [0.5, "speed_perturb", {"factor": 0.9}]
+      - [0.5, "speed_perturb", {"factor": 1.1}]
```

---

The entry was a one element list, so `DataAugmentation` read a list as the sampling weight and training crashed (see #6512):

```
TypeError: unsupported operand type(s) for /: 'int' and 'list'
```

Two fixes are possible: add the missing type2 group weight, or drop the nesting and use two type1 entries. I chose the latter because this recipe leaves `data_aug_num` at its default `[1, 1]`, so exactly one effect is applied per utterance and the two forms are equivalent: 0.9 or 1.1 at 50/50, combined with `data_aug_prob: 0.667` for the usual 3-way speed perturbation. The type2 grouping would only matter if `data_aug_num` were raised above 1, so the flat form says the same thing with less structure.

If you would rather keep the grouped form to guard against that, I am happy to switch to `[1.0, [[0.5, ...0.9], [0.5, ...1.1]]]` instead.

After the fix, `effect_probs` is `(0.5, 0.5)` and augmentation runs without error.

---

## Is your PR small enough?

Yes. 1 file, 2 lines changed.

---

## Additional Context

Fixes #6512
The recipe was added in #6418.

### DIFF
--- a/egs3/librispeech_100/asr/conf/tuning/training_e_branchformer.yaml
+++ b/egs3/librispeech_100/asr/conf/tuning/training_e_branchformer.yaml
@@ -49,8 +49,8 @@ dataset:
     fs: 16000
     data_aug_prob: 0.667
     data_aug_effects:
-      - - [0.5, "speed_perturb", {"factor": 0.9}]
-        - [0.5, "speed_perturb", {"factor": 1.1}]
+      - [0.5, "speed_perturb", {"factor": 0.9}]
+      - [0.5, "speed_perturb", {"factor": 1.1}]
 
 create_dataset:
   recipe_dir: ${recipe_dir}


sw005320:
Thanks, @nagohachi
@Masao-Someki, please check

Masao-Someki (approved):
Thank you, this looks good to me!

Masao-Someki:
Thank you @nagohachi !!

PR #6514, opened 2026-08-07
[espnet3-recipe][f5-tts 1] Add general component for TTS recipe

NewGamezzz:
## What did you change?

This PR is the first PR for the F5-TTS LibriTTS recipe. It adds the ESPnet3 infrastructure required for a TTS recipe, without introducing a TTS model or recipe yet.

---

The TTS system and template configurations are needed as the foundation for implementing the TTS recipes.

---

## Is your PR small enough?

This PR contains 14 new files with 2,308 lines of changes. Four of them are test files with 906 line changes. If you want to split the PR further, feel free to tell. I am happy to do so.

---

## Additional Context

This is **part 1 of 3**. The parts are separated so each is independently reviewable, and each depends on the one before it:

1. **This PR** - ESPnet3 TTS infrastructure (`espnet3/systems/tts`, EMA callbacks, `egs3/TEMPLATE/tts` configs).
2. **Next** (https://github.com/espnet/espnet/pull/6515) - the F5-TTS model in `espnet2` (DiT backbone, CFM, vocoder mel front-end, tokenizer, and a `LinearWarmupDecayLR` scheduler).
3. **Last** (https://github.com/espnet/espnet/pull/6516) - the `egs3/libritts/tts` recipe that ties them together, with LibriSpeech-PC cross-sentence evaluation.

Nothing here references F5-TTS specifically; `TTSSystem` and `EMACallback` are usable by any ESPnet3 TTS recipe.

Vendored file provenance: `espnet3/components/callbacks/vendored_ema.py` is copied from `ema_pytorch/ema_pytorch.py` in [lucidrains/ema-pytorch](https://github.com/lucidrains/ema-pytorch) (MIT).

### DIFF
--- a/egs3/TEMPLATE/tts/conf/inference.yaml
+++ b/egs3/TEMPLATE/tts/conf/inference.yaml
@@ -1,0 +1,31 @@
+##########################################################
+#                   Inference SETTING                    #
+##########################################################
+# Quick usage:
+#   python run.py --stages infer --inference_config conf/inference.yaml
+#
+# Shared default merged under every TTS recipe's inference config.
+
+recipe_dir: .
+exp_tag:
+exp_dir: ${recipe_dir}/exp/${exp_tag}
+data_dir: ${recipe_dir}/data
+inference_dir: ${exp_dir}/${self_name:}
+
+dataset:
+
+parallel:
+  env: local
+  n_workers: 1
+
+model:
+
+input_key:
+batch_size:
+output_fn:
+output_artifacts:
+
+provider:
+  _target_: espnet3.systems.base.inference_provider.InferenceProvider
+runner:
+  _target_: espnet3.systems.base.inference_runner.InferenceRunner
--- a/espnet3/components/callbacks/ema.py
+++ b/espnet3/components/callbacks/ema.py
@@ -2,9 +2,9 @@
 
 from __future__ import annotations
 
+import lightning.pytorch as pl
 import torch
 import torch.distributed as dist
-import lightning.pytorch as pl
 
 from espnet3.components.callbacks.vendored_ema import EMA
 
--- a/espnet3/components/callbacks/ema.py
+++ b/espnet3/components/callbacks/ema.py
@@ -10,8 +10,7 @@ from espnet3.components.callbacks.vendored_ema import EMA
 
 
 class EMACallback(pl.Callback):
-    """
-    ESPnet3's EMA callback system.
+    """ESPnet3's EMA callback system.
 
     - Updates once per true optimizer step (not per micro-step).
     - Swaps EMA weights in for validation/test, restores afterward.
--- a/espnet3/components/callbacks/ema.py
+++ b/espnet3/components/callbacks/ema.py
@@ -22,7 +22,27 @@ class EMACallback(pl.Callback):
 
         Args:
             decay: EMA decay rate, forwarded to ``EMA`` as ``beta``.
-            **ema_kwargs: Extra keyword arguments forwarded to ``EMA``.
+            **ema_kwargs: Extra keyword arguments forwarded to ``EMA``
+                (``espnet3.components.callbacks.vendored_ema.EMA``).
+                Note that ``model``, ``beta``, and ``include_online_model``
+                are already set by this callback. The remaining parameters:
+                ``ema_model`` (default: ``None``),
+                ``update_after_step`` (default: ``100``),
+                ``update_every`` (default: ``10``),
+                ``inv_gamma`` (default: ``1.0``),
+                ``power`` (default: ``2 / 3``),
+                ``min_value`` (default: ``0.0``),
+                ``param_or_buffer_names_no_ema`` (default: ``set()``),
+                ``ignore_names`` (default: ``set()``),
+                ``ignore_startswith_names`` (default: ``set()``),
+                ``allow_different_devices`` (default: ``False``),
+                ``use_foreach`` (default: ``False``),
+                ``update_model_with_ema_every`` (default: ``None``),
+                ``update_model_with_ema_beta`` (default: ``0.0``),
+                ``forward_method_names`` (default: ``()``),
+                ``move_ema_to_online_device`` (default: ``False``),
+                ``coerce_dtype`` (default: ``False``),
+                ``lazy_init_ema`` (default: ``False``).
         """
         self.decay = decay
         self.ema_kwargs = ema_kwargs
--- a/espnet3/components/callbacks/ema.py
+++ b/espnet3/components/callbacks/ema.py
@@ -1,0 +1,135 @@
+"""Callback for Exponential Moving Average (EMA) of model weights."""
+
+from __future__ import annotations
+
+import torch
+import torch.distributed as dist
+import lightning.pytorch as pl
+
+from espnet3.components.callbacks.vendored_ema import EMA
+
+
+class EMACallback(pl.Callback):
+    """
+    ESPnet3's EMA callback system.
+
+    - Updates once per true optimizer step (not per micro-step).
+    - Swaps EMA weights in for validation/test, restores afterward.
+    - Saves under 'ema_model_state_dict'.
+    """
+
+    def __init__(self, decay: float = 0.9999, **ema_kwargs):
+        """Configure the callback.
+
+        Args:
+            decay: EMA decay rate, forwarded to ``EMA`` as ``beta``.
+            **ema_kwargs: Extra keyword arguments forwarded to ``EMA``.
+        """
+        self.decay = decay
+        self.ema_kwargs = ema_kwargs
+        self.ema: EMA | None = None
+        self._backup: dict | None = None
+        self._last_global_step: int = 0
+
+    def setup(self, trainer: pl.Trainer, pl_module: pl.LightningModule, stage: str):
+        """Create the EMA copy of the model at the start of ``fit``."""
+        # Initialize EMA on the main process only
+        if stage == "fit" and trainer.is_global_zero:
+            self.ema = EMA(
+                pl_module.model,
+                beta=self.decay,
+                include_online_model=False,
+                **self.ema_kwargs,
+            ).to(pl_module.device)
+
+    def on_train_start(self, trainer, pl_module):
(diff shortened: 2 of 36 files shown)


egs3/TEMPLATE/tts/conf/inference.yaml
```diff
@@ -0,0 +1,31 @@
...
+
+input_key:
+batch_size:
+output_fn:
+output_artifacts:
+
+provider:
+  _target_: espnet3.systems.base.inference_provider.InferenceProvider
+runner:
+  _target_: espnet3.systems.base.inference_runner.InferenceRunner
```
Masao-Someki, line 31: If these are the same as the default config, you can simply remove it

espnet3/components/callbacks/ema.py
```diff
@@ -0,0 +1,134 @@
...
+    - Saves under 'ema_model_state_dict'.
+    """
+
+    def __init__(self, decay: float = 0.9999, **ema_kwargs):
+        """Configure the callback.
+
+        Args:
+            decay: EMA decay rate, forwarded to ``EMA`` as ``beta``.
+            **ema_kwargs: Extra keyword arguments forwarded to ``EMA``.
+        """
```
Masao-Someki, line 46: Since we want the docstrings to be more informative, we want more detailed docstrings.
Could you add reference to the ``EMA``, such as adding module path to the class?

espnet3/components/callbacks/vendored_ema.py
```diff
@@ -0,0 +1,449 @@
+"""Vendored EMA implementation (removes the ``ema_pytorch`` dependency).
```
Masao-Someki, line 22: For this file I think it might be good to add the dependency in the tts group

espnet3/systems/tts/remove_long_short_runner.py
```diff
@@ -0,0 +1,99 @@
...
+        state: Dict[str, Any],
+        **env,
+    ) -> None:
+        """Append one ``forward`` result (or batch of results) to the shard file."""
+        records = result if isinstance(result, list) else [result]
+        for record in records:
+            writers["results"].write(json.dumps(record) + "\n")
+
+    def merge(self, shard_dirs: List[Path]) -> List[Dict[str, Any]]:
+        """Concatenate shard results and restore manifest (``idx``) order."""
```
Masao-Someki: Could you add an example in the docstring that what kind of data will be actually written in the results.jsonl?
2-3 lines of example would be nice.

espnet3/systems/tts/system.py
```diff
@@ -0,0 +1,466 @@
...
+            training_config.remove_long_short.min_wav_duration: Minimum
+                duration in seconds
+            training_config.remove_long_short.max_wav_duration: Maximum
+                duration in seconds
+            training_config.remove_long_short.save_path: Directory to save
+                filtered manifests
+            training_config.remove_long_short.splits: List of splits to
+                process (train, valid, test)
+            training_config.remove_long_short.manifest_paths: Optional dict
+                of split to manifest path (default: data/manifest/{split}.tsv)
```
Masao-Someki: I thought it might be good to just list the config args name like `min_wav_duration`, and add one example of the config.

So it would be like:
```
Configuration should include:
  - ``min_wav_duration``: ...
  - ``max_wav_duration``: ...

Example:

  .. code-block:: yaml

      remove_long_short:
        min_wav_duration: 2
        max_wav_duration: 30
```

espnet3/systems/tts/system.py
```diff
@@ -0,0 +1,466 @@
...
+        Raises:
+            RuntimeError: If required configuration is missing or manifest
+                files not found.
+        """
+        self._reject_stage_args("remove_long_short", args, kwargs)
+        logger.info(
+            "TTSSystem.remove_long_short(): starting long-short utterance removal"
+        )
+
+        rls_cfg = self.training_config.get("remove_long_short", None)
```
Masao-Someki: Please add comment to explain `rls`, (remove long short?), and change `cfg` to `config`. We try not to not allow shortened word as much as possible.
Maybe we can consider renaming it with different wordings

espnet3/systems/tts/system.py
```diff
@@ -0,0 +1,466 @@
...
+        )
+
+    def create_token_list(self, *args, **kwargs):
+        """Create token list from training data.
+
+        This stage processes the training manifest to extract unique
+        tokens from the text transcriptions and saves them to a token
+        list file.
+
+        Configuration should include:
```
Masao-Someki: ditto
Let's add example config file in the docstring

espnet3/systems/tts/system.py
```diff
@@ -0,0 +1,466 @@
...
+            raise RuntimeError(
+                "training_config.create_token_list.save_path must be set "
+                "for create_token_list stage."
+            )
+        filename = tl_cfg.get("filename", None)
+        if filename is None:
+            raise RuntimeError(
+                "training_config.create_token_list.filename must be set "
+                "(e.g. 'tokens.txt'); save_path is the output directory."
+            )
```
Masao-Someki: @NewGamezzz I realized we have multiple these lines. (extracting value from config, and if not exists raise error)
Do you think it's good to have a general function to do this in the BaseSystem, so that user can run the single function with the field key and error message?

Masao-Someki (commented):
Thanks @NewGamezzz , I added comments.
I think the `RemoveLongShort` stage would be a good example of runner/provider use-case!

egs3/TEMPLATE/tts/conf/inference.yaml
NewGamezzz: This config file is the default config (TEMPLATE) for the TTS recipe. So, I think it is ok to put it here.

espnet3/components/callbacks/vendored_ema.py
NewGamezzz: Since EMA can be used by other tasks, should we also add to other groups?

espnet3/components/callbacks/ema.py
NewGamezzz: Thanks. I have added the exact arguments of the EMA and the path to the EMA class.

espnet3/systems/tts/remove_long_short_runner.py
NewGamezzz: Done. I have added the example to the comment.

espnet3/systems/tts/system.py
NewGamezzz: Done. I shorten the comment to include only the arg name and added an example.

espnet3/systems/tts/system.py
NewGamezzz: I change every `rls_cfg` variables to `remove_long_short_config` to avoid the confusion. If you think that it is too long, may be `filtered_config` can be a good candidate?

espnet3/systems/tts/system.py
NewGamezzz: Follow the same format as `remove_long_short` function. Shorten the comment to include only the arg name and add an example config.

espnet3/systems/tts/system.py
NewGamezzz: I added `_get_required_config` function inside `BaseSystem`. It returns the value of that key if exists. Otherwise, it raises an error.

NewGamezzz:
Thanks @Masao-Someki for the comments. I have reflected on it and fix the code accordingly. Can you take a look at it? 

I will create a PR adding the document for `RemoveLongShort` stage as an example of runner/provider.

egs3/TEMPLATE/tts/conf/inference.yaml
Masao-Someki: Oh sorry, yeah true. Then this looks good to me!

espnet3/components/callbacks/vendored_ema.py
Masao-Someki: Yes, it's fine to add it to another groups. You can check the current pyproject.toml, for some package it is listed in multiple groups

espnet3/systems/tts/system.py
Masao-Someki: Thanks, I think `remove_long_short_config` should be fine.

Masao-Someki (approved):
Thanks @NewGamezzz, LGTM!

Masao-Someki:
I confirmed with Shinji that there's no problem with the system directory structure! Let's merge this and move to the next one!

espnet3/components/callbacks/ema.py
sw005320, line 1: I want to make sure this passes the test.
Can you answer it?

espnet3/components/callbacks/vendored_ema.py
```diff
@@ -0,0 +1,449 @@
...
+
+Copy of ``ema_pytorch/ema_pytorch.py`` from lucidrains/ema-pytorch
+(https://github.com/lucidrains/ema-pytorch), which is a single self-contained,
+torch-only file. Reformatted to this repo's style (black, 88-column lines);
+the EMA logic is unchanged so checkpoints written/read via
+``EMA.state_dict()`` (keys ``ema_model.*`` + ``initted`` + ``step``) stay
+compatible.
```
sw005320, line 29: Is it OK in terms of the license?
If it is not Apache 2.0 (ESPnet's default), it's better to add the license header from the original repo.
(We should not change the license)

espnet3/components/callbacks/ema.py
sw005320: I confirmed it in test/espnet3/components/callbacks/test_ema.py

espnet3/components/callbacks/vendored_ema.py
NewGamezzz: Thanks. I checked the original repo. Their code is MIT licensed. Based on my understanding, we should be able to use/copy their code as long as we add the license in the header. I have added the original MIT license at the header of this file. https://github.com/lucidrains/ema-pytorch/blob/main/LICENSE

sw005320 (approved):

Masao-Someki (approved):
Thanks @NewGamezzz !

sw005320:
Thanks, @NewGamezzz!

PR #6515, opened 2026-08-07
[espnet3-recipe][f5-tts 2] Add F5-TTS models and Helpers

NewGamezzz:
## What did you change?

This PR is the second PR for the F5-TTS LibriTTS recipe. It adds F5-TTS - a flow-matching, non-autoregressive TTS model with zero-shot voice cloning - to `espnet2`, so it can be selected with `tts: f5tts` alongside the existing models. No recipe is added here; that is part 3.

**Ported from the F5-TTS reference implementation**
([SWivid/F5-TTS](https://github.com/SWivid/F5-TTS)):

- `espnet2/tts/f5/backbones/dit.py`, `cfm.py`, `modules.py` - the DiT backbone and the conditional flow-matching objective and sampler.
- `espnet2/tts/f5/utils.py`, `espnet2/text/f5_pinyin.py` - the helpers and the zh+en pinyin tokenization those modules need.

**Self-contained copies that avoid new dependencies:**

- `espnet2/tts/f5/rotary.py` - the two rotary-embedding symbols F5 uses from `x_transformers`, so that package is not required.
- `espnet2/tts/f5/solvers.py` - the fixed-step `euler` / `midpoint` ODE steps F5's sampler uses, reproducing `torchdiffeq.odeint` for this case in ~60 lines.

**Written for espnet2:**

- `espnet2/tts/f5/f5tts.py` - `AbsTTS` adapter that makes the ported model selectable from the TTS task.
- `espnet2/tts/f5/inference.py` - inference engine a recipe's `infer` stage builds, handling reference-conditioned synthesis and vocoder decoding.
- `espnet2/tts/feats_extract/vocoder_mel.py` - `AbsFeatsExtract` wrapper around F5's own mel, so training features stay bit-compatible with the neural vocoder.
- `espnet2/text/f5_preprocessor.py` - preprocessor entry for the above tokenizer.
- `espnet2/schedulers/linear_warmup_decay.py` - `LinearWarmupDecayLR`, an `AbsBatchStepScheduler` implementing the linear warmup and linear decay schedule F5 trains with.

---

It adds the core component for the f5-tts recipe. 

---

## Is your PR small enough?
After merging the first part, it will contain 16 file changes with 3,074 line changes. Most of the codes are a direct copy of the official f5-tts repo.

---

## Additional Context

This is **part 2 of 3**. Each part is independently reviewable and depends on the one before it:

1. **Part 1** (https://github.com/espnet/espnet/pull/6514) - ESPnet3 TTS infrastructure: `espnet3/systems/tts`, EMA callbacks, `egs3/TEMPLATE/tts` configs.
2. **This PR** - the F5-TTS model in `espnet2`: DiT backbone, CFM, vocoder mel front-end, tokenizer, and the `LinearWarmupDecayLR` scheduler.
3. **Next** (https://github.com/espnet/espnet/pull/6516) - the `egs3/libritts/tts` recipe tying them together, with LibriSpeech-PC cross-sentence evaluation.

Nothing in this PR is recipe-specific; the model is usable from any espnet2 TTS recipe once registered.

Upstream sources, for reviewers diffing the port:
- F5-TTS: https://github.com/SWivid/F5-TTS (MIT)
- Rotary embedding: `x_transformers/x_transformers.py` in [lucidrains/x-transformers](https://github.com/lucidrains/x-transformers) (MIT)
- ODE steps: equivalent to `torchdiffeq.odeint` for the fixed-step methods F5 uses

TODO:
- [x] Recheck the license of F5-TTS code
- [x] Merge the first pr

## Summary by CodeRabbit

* **New Features**
  * Added F5-TTS reference-conditioned text-to-speech generation.
  * Added Chinese-English pinyin tokenization and preprocessing.
  * Added Vocos and BigVGAN vocoder support.
  * Added configurable linear warmup and decay learning-rate scheduling.
  * Added checkpoint loading, batching, text chunking, cross-faded output, and configuration conversion utilities.

* **Bug Fixes**
  * Improved reference-audio duration handling and generated-output trimming.
  * Corrected mel-feature frame length calculations.
  * Added validation for invalid scheduler settings and mismatched batch lengths.

* **Tests**
  * Added extensive coverage for F5-TTS, inference, vocoding, tokenization, sampling, scheduling, and model components.


espnet3/systems/tts/f5_tts/__init__.py
Masao-Someki, line 1: Let's add citation for the F5TTS paper!

espnet3/systems/tts/f5_tts/builder.py
```diff
@@ -0,0 +1,161 @@
+"""Build the F5-TTS model for ESPnet3's Hydra instantiation path.
+
+Replaces the ``espnet2.tasks.tts.TTSTask.build_model`` route: F5-TTS is an
+ESPnet3 model and is no longer registered in ``espnet2/tasks/tts.py``, so the
+training config reaches it through ``_target_`` instead of ``task:``.
+
+The returned object is still ``espnet2.tts.espnet_model.ESPnetTTSModel``, which
```
Masao-Someki: Actually if you set `task` in the training config, then it will still use espnet2's task class to build the model.
For espnet3 models, we can ignore the task (i.e, left task unset), then model initialization doesn't use the espnet2's task model, or collect-stats.

Note that you need to be responsible to implement/select `collate_fn` if you're not using espnet2 task.

espnet3/systems/tts/f5_tts/builder.py
```diff
@@ -0,0 +1,161 @@
...
+from typing import Any, Dict, List, Optional, Sequence, Union
+
+from omegaconf import DictConfig, ListConfig, OmegaConf
+
+from espnet2.tts.espnet_model import ESPnetTTSModel
+from espnet3.systems.tts.f5_tts.f5tts import F5TTS
+from espnet3.systems.tts.f5_tts.vocoder_mel import VocoderMelSpec
+
+
+def _plain(value: Any) -> Any:
```
Masao-Someki: Let's move this to `espnet3.utils.config_utils.py`, perhaps under a more general name like `convert_to_dict`?

When I was building ESPnet3, I initially thought we wouldn't need it, but I think it would be fine to have a utility like this after all.

espnet3/systems/tts/f5_tts/builder.py
```diff
@@ -0,0 +1,161 @@
...
+        list such as ``frac_lengths_mask: [0.7, 1.0]`` arrives as a real
+        ``list`` and is coerced by the component's own signature rather than
+        being stored as a ``ListConfig``.
+    """
+    if isinstance(value, (DictConfig, ListConfig)):
+        return OmegaConf.to_container(value, resolve=True)
+    return value
+
+
+def build_f5_tts_model(
```
Masao-Someki: As noted above, we don't need this function if `task` is left unset. The `task` flag is only needed for backward compatibility, for users who may want to use ESPnet2 models in the future.

espnet3/systems/tts/f5_tts/f5tts.py
```diff
@@ -0,0 +1,284 @@
...
+
+        Note:
+            The span masked for prediction is drawn at random each call
+            (``frac_lengths_mask``), so the loss is stochastic: two calls on the
+            same batch differ unless the RNG is reseeded.
+        """
+        text = self._remap_text_padding(text, text_lengths)
+        loss, _cond, _pred = self.cfm(feats, text=text, lens=feats_lengths)
+        stats = dict(loss=loss.detach())
+        weight = feats.new_tensor(feats.size(0))
```
Masao-Someki: You can return None here

espnet3/systems/tts/f5_tts/inference.py
```diff
@@ -0,0 +1,522 @@
...
+        self.target_sample_rate = target_sample_rate
+        self.nfe_step = nfe_step
+        self.cfg_strength = cfg_strength
+        self.sway_sampling_coef = sway_sampling_coef
+        self.speed = speed
+        self.target_rms = target_rms
+        self.cross_fade_duration = cross_fade_duration
+        self.seed = seed
+
+        cfg = OmegaConf.to_container(OmegaConf.load(train_config), resolve=True)
```
Masao-Someki: For `OmegaConf.load` you can think of using `espnet3.utils.config_utils.load_config_with_defaults`. It merges the default config under `TEMPLATE/`, which is usually not listed in the training config.

espnet3/systems/tts/f5_tts/inference.py
```diff
@@ -0,0 +1,522 @@
...
+                    "`pip install vocos`."
+                ) from e
+            if vocoder_path:
+                vocoder = Vocos.from_hparams(f"{vocoder_path}/config.yaml")
+                state = torch.load(
+                    f"{vocoder_path}/pytorch_model.bin", map_location="cpu"
+                )
+                vocoder.load_state_dict(state)
+            else:
+                vocoder = Vocos.from_pretrained("charactr/vocos-mel-24khz")
```
Masao-Someki: Let's move these default values to `__init__.py`. They’re difficult to put under `TEMPLATE/` because they are specific to F5-TTS.

For example, we could define them, such as `espnet3.systems.tts.f5_tts.VOCOS_DEFAULT_MODEL`, so future developers and researchers can easily find which default model we're using.

espnet3/systems/tts/f5_tts/inference.py
```diff
@@ -0,0 +1,522 @@
...
+        elif vocoder_name == "bigvgan":
+            try:
+                import bigvgan
+            except ImportError as e:
+                raise ImportError(
+                    "bigvgan is required for vocoder_name='bigvgan'. See "
+                    "https://github.com/NVIDIA/BigVGAN."
+                ) from e
+            repo = vocoder_path or "nvidia/bigvgan_v2_24khz_100band_256x"
```
Masao-Someki: ditto

espnet3/systems/tts/f5_tts/inference.py
```diff
@@ -0,0 +1,522 @@
...
+            vocoder = bigvgan.BigVGAN.from_pretrained(repo, use_cuda_kernel=False)
+            vocoder.remove_weight_norm()
+        else:
+            raise ValueError(f"Unsupported vocoder: {vocoder_name!r}.")
+        return vocoder.to(self.device).eval()
+
+    # -------------------------------------------------------------- inference
+
+    def _tokenize(self, text: str) -> np.ndarray:
+        return self._tokenize_fn(text)
```
Masao-Someki: I just wondered if we need this function, we can just rename `_tokenize_fn` to `_tokenize`

espnet3/systems/tts/f5_tts/preprocessor.py
```diff
@@ -0,0 +1,93 @@
...
+    ):
+        """Configure the preprocessor.
+
+        Args:
+            vocab_file: F5's ``vocab.txt`` (e.g.
+                ``Emilia_ZH_EN_pinyin/vocab.txt``); line number = token id,
+                space at index 0.
+            text_name: Sample dict key holding the raw transcript.
+            train: Accepted for the collect_stats train/valid toggle
+                (no-op here).
```
Masao-Someki: Sorry you can remove this, we fixed this issue in #6523

espnet3/components/schedulers/linear_warmup_decay.py
Masao-Someki, line 1: Let's move this to `espnet3.components.schedulers.linear_warmup_decay.py`

espnet3/systems/tts/f5_tts/vocoder_mel.py
```diff
@@ -0,0 +1,157 @@
...
+        self.mel = MelSpec(
+            n_fft=n_fft,
+            hop_length=hop_length,
+            win_length=win_length,
+            n_mel_channels=n_mels,
+            target_sample_rate=fs,
+            mel_spec_type=mel_spec_type,
+        )
+
+    def output_size(self) -> int:
```
Masao-Someki, line 80: Maybe adding `@property`? (because in espnet3, for the normal functions we don't allow noun as a function name..)

Masao-Someki (commented):
Thank you @NewGamezzz !!

I think we should avoid modifying files that were ported directly from the original F5-TTS repository, so I didn’t review those files in detail..

espnet3/systems/tts/f5_tts/__init__.py
NewGamezzz: Thanks! I added the citation in the header.

espnet3/systems/tts/f5_tts/builder.py
NewGamezzz: Got it. I changed the comment.

espnet3/systems/tts/f5_tts/builder.py
NewGamezzz: Thanks. I have moved it to `config_utils.py`

espnet3/systems/tts/f5_tts/f5tts.py
NewGamezzz: Done.

espnet3/systems/tts/f5_tts/inference.py
NewGamezzz: Thanks. That's right. I have changed to `load_config_with_defaults`.

espnet3/systems/tts/f5_tts/inference.py
NewGamezzz: Thanks. I have moved these to `_init_.py`.

espnet3/systems/tts/f5_tts/inference.py
NewGamezzz: Done.

espnet3/systems/tts/f5_tts/inference.py
NewGamezzz: That's correct. This is redundant. I have removed it and changed the name to `_tokenize`.

espnet3/systems/tts/f5_tts/preprocessor.py
NewGamezzz: Done.

espnet3/components/schedulers/linear_warmup_decay.py
NewGamezzz: Thanks. I have moved to the components directory now.

espnet3/systems/tts/f5_tts/vocoder_mel.py
NewGamezzz: Done.

espnet3/systems/tts/f5_tts/builder.py
NewGamezzz: I may misunderstand this part. With task left unset, this function is what builds the model, since `model._target_` points at it to build the model. Removing it would leave nothing to construct the model from.

Earlier I did use task to build the model, but that required registering F5-TTS in `espnet2/tasks/tts.py`. Now that the model lives in `espnet3/systems/tts/f5_tts/`, registering it there would make espnet2 import from espnet3.

NewGamezzz: Note that this builder still returns the same `ESPnetTTSModel` from espnet2, so it does not change the result compared to building through the task. I trained 10 steps and both give identical gradients and weights.

espnet3/systems/tts/f5_tts/backbones/__init__.py
sw005320, line 1: is "backbones" correct name?

espnet3/systems/tts/f5_tts/builder.py
Masao-Someki: OK, thanks. I think I understand where the confusion came from.

For a new model in ESPnet3, the model does not have to be an `ESPnetTTSModel`. Basically, any class that inherits from `torch.nn.Module` should be fine, as long as it follows the expected interface... e.g., `forward()` should return `loss`, `stats`, and `weight`.
So I thought you can define `F5TTS` directly as the model and point `model._target_` to it.

sw005320 (commented):
Is the naming convention under the `f5_tts` appropriate?
We want to have a balance of flexiblity and consistency.

@Masao-Someki, please comment it.

espnet3/systems/tts/f5_tts/backbones/__init__.py
Masao-Someki: Thank you, and this is a good point.

@NewGamezzz, do you plan to add more backbone architectures under `backbones/`? If not, I think it would be clearer to remove the directory and move `dit.py` directly under `f5_tts/`.

espnet3/systems/tts/f5_tts/inference.py
```diff
@@ -0,0 +1,526 @@
...
+    if not waves:
+        return np.zeros(1, dtype=np.float32)
+    if len(waves) == 1:
+        return waves[0]
+    if cross_fade_duration <= 0:
+        return np.concatenate(waves)
+    final_wave = waves[0]
+    for next_wave in waves[1:]:
+        prev_wave = final_wave
+        n = int(cross_fade_duration * sample_rate)
```
Masao-Someki: We should make it clear what the `n` means I think?
Could you change this to more informative name.

espnet3/systems/tts/f5_tts/inference.py
```diff
@@ -0,0 +1,526 @@
...
+        self.target_sample_rate = target_sample_rate
+        self.nfe_step = nfe_step
+        self.cfg_strength = cfg_strength
+        self.sway_sampling_coef = sway_sampling_coef
+        self.speed = speed
+        self.target_rms = target_rms
+        self.cross_fade_duration = cross_fade_duration
+        self.seed = seed
+
+        cfg = OmegaConf.to_container(
```
Masao-Someki: In ESPnet3, we don't allow abbreviated variable names such as `cfg`. Please rename this to `config`.

espnet3/systems/tts/f5_tts/inference.py
```diff
@@ -0,0 +1,526 @@
...
+            load_config_with_defaults(train_config), resolve=True
+        )
+        fe_conf = (cfg.get("model") or {}).get("feats_extract_conf") or {}
```
Masao-Someki: ditto
please check the other variables in this file!

espnet3/systems/tts/f5_tts/inference.py
```diff
@@ -0,0 +1,526 @@
...
+            else:
+                raw = ckpt
+        return {
+            k.replace("ema_model.", "", 1): v
+            for k, v in raw.items()
+            if k not in ("initted", "step")
+        }
+
+    @staticmethod
+    def _log_load(tag: str, ckpt_path: str, missing, unexpected) -> None:
```
Masao-Someki: Maybe `_log_model_loading` might be clearer?

espnet3/systems/tts/f5_tts/backbones/__init__.py
NewGamezzz: I agree. I don't plan to add it right now. I will delete `backbones` directory and move `dit.py` out.

Masao-Someki (commented):
I added a few comments about variable naming. Let's avoid using abbreviated variable names in ESPnet3.

espnet3/systems/tts/f5_tts/f5tts.py
```diff
@@ -0,0 +1,372 @@
...
+            checkpoint_activations=checkpoint_activations,
+            dropout=dropout,
+        )
+        self.cfm = CFM(
+            transformer=backbone,
+            sigma=sigma,
+            audio_drop_prob=audio_drop_probability,
+            cond_drop_prob=condition_drop_probability,
+            num_channels=self.mel_dim,
+            mel_spec_kwargs=convert_to_dict(mel_spectrogram_kwargs) or {},
```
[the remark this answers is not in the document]
NewGamezzz: There is no valid reason for the two to differ. Therefore, we dropped `mel_spectrogram_kwargs` entirely. F5TTS now derives CFM's mel from feats_extract_config, so a mismatch can no longer be configured.

espnet3/systems/tts/f5_tts/inference.py
NewGamezzz: Thanks. It is the length of `cross_fade_duration` in samples. I changed it to `cross_fade_samples`.

espnet3/systems/tts/f5_tts/inference.py
NewGamezzz: Thanks. I replaced all abbreviations with their full names and applied the same convention to all variables (except those in the ported files).

Note that conventional terms such as EMA, ODE, etc. remain abbreviated. If you prefer to use the full names for these terms as well, please let me know.

espnet3/systems/tts/f5_tts/inference.py
NewGamezzz: Done!

espnet3/systems/tts/f5_tts/inference.py
NewGamezzz: agree. I changed it to `_log_model_loading`.

espnet3/systems/tts/f5_tts/vocoder_mel.py
```diff
@@ -0,0 +1,160 @@
...
+
+        Returns:
+            ``n_mels``.
+
+        Note:
+            :class:`~espnet3.systems.tts.f5_tts.f5tts.F5TTS` reads this to size
+            its backbone, which is why the recipe states the mel dimension only
+            once, in ``feats_extract_conf``.
+        """
+        return self.n_mels
```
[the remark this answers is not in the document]
NewGamezzz: @Masao-Someki I will need to change this to a method. This is because I implemented `VocoderMelSpec` as a subclass of `AbsFeatExtractor`, which treats `output_size` as a method.

If you prefer `output_size` to be a property, I can stop inheriting `VocoderMelSpec` from `AbsFeatExtractor`. Since our model is not an `ESPnetTTSModel`, we can simply ignore this requirement.

espnet3/systems/tts/f5_tts/builder.py
NewGamezzz: Got it! I removed the builder and implemented `extract_feat` and `collect_feat` directly in `F5-TTS`.

espnet3/systems/tts/f5_tts/inference.py
Masao-Someki: Thanks, I think EMA, ODE, etc should be fine.

espnet3/systems/tts/f5_tts/vocoder_mel.py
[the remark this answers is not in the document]
Masao-Someki: OK, let’s implement this in the method. This is something we can’t handle within ESPnet3 because this class inherits from an ESPnet2 model.

test/espnet3/systems/tts/f5_tts/__init__.py
Masao-Someki, line 1: Is there a unit test for solvers.py?

pyproject.toml
```diff
@@ -87,6 +87,7 @@ asr = [
 tts = [
   "pyworld>=0.3.4",
   "pypinyin<=0.44.0",
+  "rjieba",
```
Masao-Someki, line 103: Sorry where is this used? Do we need this dependency?

Masao-Someki (commented):
Thanks @NewGamezzz , I have 2 comments!

pyproject.toml
NewGamezzz: This library is used for Chinese tokenization. It was used in `pinyin.py`. https://github.com/espnet/espnet/pull/6515/changes#diff-04ec569e1f5acc05bca4e0a3093c2d7c6cb35b6c1754f87918aebd2ed3a644f2R81

test/espnet3/systems/tts/f5_tts/__init__.py
NewGamezzz: It was tested inside `test_sampling.py` https://github.com/espnet/espnet/pull/6515/changes#diff-e15c2314409fab96b0ad14c54e446d264f0d49783e93235ce3db397ecf53eb7cR49-R67. Since it is mainly used for sampling, I decided to put it there. Do you want me to move it out and create a new test file?

Masao-Someki (approved):
Thanks @NewGamezzz LGTM!

Masao-Someki:
Hi @NewGamezzz (and @sw005320 and @popcornell),

I discussed the structure for each architecture under `systems/` with @popcornell, and I think it would be better to introduce a `models/` directory under each system.

The main reason is to avoid mixing architecture-specific implementations (e.g., F5-TTS for this PR) with architecture-independent/common utilities (e.g., metrics) at the same directory level as we add more models in the future.

For example:

```text
espnet3/systems/tts/
    system.py
    task.py
    metrics/
    models/
        f5_tts/
        vits/
        ...

espnet3/systems/asr/
    system.py
    task.py
    metrics/
    tokenizers/
    models/
        encoder_decoder/
        encoder_ctc/
        rnn_transducer/
        granite/
        ...
```

Model-specific functionality can then stay inside each model package. For example, if a model requires a specific loss, decoding/inference logic, or helper modules, we can define them under `models/<model>/`.

This also fits the ESPnet3 model interface well, since the model `forward()` itself returns the loss. If there is a loss shared by multiple models, we can still place it at the system level, e.g., `systems/tts/losses/`. On the other hand, formulation-specific losses such as CTC or transducer losses can remain within the corresponding model implementation.

So my preference would be to standardize on `systems/<system>/models/<model>/`.

This will require some changes to the current directory structur, configs and tests, so sorry for the extra work. I think it is better to make this structural change now, before more models and tasks start following the current layout.

sw005320:
- what is `encoder_ctc`? We can just call `ctc`
- `encoder_decoder` can be anything, and `aed` should be fine (attention-based)
- `rnn_transducer` --> `rnnt` should be fine

Masao-Someki:
Thank you, for asr models I agree.

We can refer to HF transformers' directory structure for this discussion. Please refer to [src/transformers](https://github.com/huggingface/transformers/tree/main/src/transformers). I think we should keep each `systems/<system>` directory structured similarly to this.

NewGamezzz:
Got it. I will move F5-TTS under the `espnet3/systems/tts/models/f5_tts` directory.

Just one more question: is this structure intended only for architectures implemented specifically in ESPnet3?

For example, my EnCodec LibriTTS recipe uses the ESPnet2 task to initialize EnCodec, so I don’t currently have an implementation under `espnet3/systems/codec/models/encodec`. In this case, should I leave it as is and refer to the ESPnet2 implementation from config file, or should I also move the relevant code from ESPnet2 to `espnet3/systems/codec/models/encodec/`?

popcornell:
I think what Shinji is suggesting makes sense. 
We should add the convention somewhere in CONTRIBUTION documentation. 

Anyway, regarding this PR, is there something we can move to espnet3/components and outside the recipe to avoid boilerplate ? 

solvers.py technically can be used by xany flow-matching model; 
rotary.py --? positional encoding no ? 
vocoder_mel.py —> BigVGAN/Vocos mel extraction also can be reusable by other vocoder-based TTS

Masao-Someki:
> Just one more question: is this structure intended only for architectures implemented specifically in ESPnet3?

@NewGamezzz Thanks, and yes, for espnet2 models we can just keep referring from the config. This is for a new model implemented in espnet3.

@popcornell Thank you!!
> We should add the convention somewhere in CONTRIBUTION documentation.

I'll add that in the document PR.

> Anyway, regarding this PR, is there something we can move to espnet3/components and outside the recipe to avoid boilerplate ?

`espnet3/components` is not for model definition. It's for handling the workflow like trianing, evaluating, publishing, etc. Even if we want to move files outside of the `f5-tts` directory, we should still keep it under `systems/tts/`. 
If we want RotaryEmbedding in ASR for example, then we can copy it in `systems/asr`.

I think for now it's fine, if we have another TTS model that uses modules from f5-tts, then we can move it under `systems/tts/`

NewGamezzz:
@Masao-Someki I have moved the codes to `espnet3/system/tts/models/f5_tts` and also moved the test files under the models directory. Could you please recheck it and let me know if everything looks good to you?

Masao-Someki (approved):
Thank you @NewGamezzz! LGTM!

Masao-Someki:
Hi, sorry for the back-and-forth and confusion. Would you mind moving the F5-TTS implementation to `systems/f5tts/`?

The idea is to keep the implementation self-contained and make it easier for others to extend your work. For example, by adding ONNX export or support for another task like as voice editing(?). We can technically put `voice_editing_system.py` under `systems/f5tts/`..

NewGamezzz:
@Masao-Someki Done. I have moved it to `system/f5_tts`. Note that I used `f5_tts` instead of `f5tts`, following the naming in the official repository. But if you prefer `f5tts`, I can change it.

Masao-Someki (approved):

Masao-Someki:
@NewGamezzz Thank you and sorry `f5_tts` should be correct!
But I think we should change the `f5tts.py` to `f5_tts.py`, sorry I aproved it but could you change it?

NewGamezzz:
@Masao-Someki Oh, sorry. I forgot to change it. I have changed it now. Thanks!

Masao-Someki:
Sorry for back-and-forth, based on our discussion on system naming rules, could you change the system name to `f5tts`?

PR #6497, opened 2026-07-22
[espnet3-recipe] Add LibriTTS recipe for codec models

NewGamezzz:
## What did you change?

Add a LibriTTS codec recipe to ESPnet3, migrated from the espnet2 recipe (egs2/libritts/codec1).

---

## Why did you make this change?

This PR migrates the espnet2 LibriTTS codec recipe to ESPnet3 so codec models (e.g., Encodec) can be trained with the ESPnet3 workflow.

---

## Is your PR small enough?

This PR contains 26 file changes. However, some of them are just config files.

---

## Additional Context

This PR was tested by training an Encodec model on 4x V100 GPUs. Results on LibriTTS test-clean:
| Metrics      | Scores |
| ----------- | ----------- |
|  utmos        |             3.9874|
|  dns_overall     |          3.1150|
 | dns_p808      |            3.7710|

I modified espnet3/components/data/dataloader.py to fix a multi-GPU deadlock caused by an unequal number of batches per rank. With ChunkIterFactory, ranks are not guaranteed to yield the same number of chunks: a rank that receives longer audio splits it into more chunks. When one rank runs out of batches while another still has some, the ranks disagree on when to enter validation and training deadlocks. To fix this, I added a DistributedSyncIterator that performs an all-reduce each step to check whether any rank has exhausted its samples; if so, all ranks stop the epoch together and enter validation in sync. I also removed the per-rank batch size check, since ESPnet3 uses the same batch size on every rank and the check is no longer needed.

To be discussed: I am thinking of moving the gan module to the base directory because other recipes, such as VITS in the TTS task, can use it. We can also move versa metric to the espnet3 directory too.

PR #6517, opened 2026-08-08
[espnet3] Add BaseIterator with DDP epoch-end sync for dataloaders

NewGamezzz:
## What did you change?

Add a BaseIterator to fix the DDP epoch-end deadlock (such as ChunkIterFactory)

- Add `espnet3/components/data/iterator.py` with `BaseIterator`, the base class for the per-epoch iterators returned by `DataLoaderBuilder`. When `torch.distributed` is initialized, it all-reduces a 1-element has-next flag with MIN before yielding each batch, so every rank ends the epoch at the same batch index. Without distributed it passes batches through unchanged, and it delegates `__len__` to the wrapped iterator.
- `DataLoaderBuilder._build_iter_factory` now always returns a `BaseIterator` wrapping `iter_factory.build_iter(epoch)`.
- Drop the `len(batch) >= world_size` check, which rejected the mandatory single-utterance batches that espnet2's `ChunkIterFactory` emits. Since espnet3 strides whole batches across ranks (batches[rank::world_size]), the actual batch size per GPU is the same as specified in the config - unlike espnet2's sequence iterator, which splits each batch element-wise across GPUs so that the total batch size across GPUs matches the config - so this check becomes unnecessary for espnet3.

Iter factories like espnet2's `ChunkIterFactory` emit a data-dependent number of batches per rank, so under DDP each rank ends its epoch at a different step. Lightning then moves the exhausted rank into validation while the others are still training, and the two sides issue mismatched collectives that deadlock until the NCCL watchdog kills the job.

This is the same fix espnet2 uses in `espnet2/train/trainer.py` (all-reducing an `iterator_stop` flag across ranks) https://github.com/espnet/espnet/pull/6497#discussion_r3726313490.

## Is your PR small enough?

Yes, it contains 4 file changes with 271 line changes.

### DIFF
--- a/espnet3/components/data/iterator.py
+++ b/espnet3/components/data/iterator.py
@@ -3,7 +3,7 @@
 import torch
 
 
-class BaseIterator:
+class EpochSyncIterator:
     """Base class for the per-epoch iterators returned by DataLoaderBuilder.
 
     Wraps the iterator produced by an espnet2-style iter factory
@@ -21,6 +21,35 @@ class BaseIterator:
     flag with MIN, so the epoch ends on every rank at the same batch index.
 
     Without ``torch.distributed``, batches are yielded unchanged.
+
+    Args:
+        iterator (Iterable): This rank's per-epoch batch iterator, typically
+            the return value of ``iter_factory.build_iter(epoch)``. It is
+            iterated once per ``__iter__`` call, and ``__len__`` is delegated
+            to it, so it must be sized for ``len()`` to work.
+
+    Note:
+        Subclasses should override :meth:`generate` rather than ``__iter__``,
+        so that the epoch-end synchronization stays in place.
+
+    Note:
+        The synchronization costs one all-reduce of a 1-element tensor per
+        batch, which is negligible next to the gradient synchronization that
+        every distributed training step already performs.
+
+    Examples:
+        Without ``torch.distributed``, the wrapper is a pass-through:
+
+        >>> list(BaseIterator([{"speech": 0}, {"speech": 1}]))
+        [{'speech': 0}, {'speech': 1}]
+
+        Under DDP, ``DataLoaderBuilder`` wraps the iter factory's iterator so
+        that every rank stops at the shortest rank's batch count::
+
+            iter_factory = ChunkIterFactory(dataset, batches=batches, ...)
+            iterator = BaseIterator(iter_factory.build_iter(epoch))
+            for batch in iterator:  # same number of steps on every rank
+                ...
     """
 
     def __init__(self, iterator):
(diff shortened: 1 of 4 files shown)


NewGamezzz:
@Emrys365 Could you take a quick look at this? I created a separate PR for the ChunkIterFactory fix. The core concept remains the same as the previous version, but I created a dedicated `BaseIterator` class for this file. Does this approach look good to you, or do you think there’s a better way to fix it?

espnet3/components/data/iterator.py
```diff
@@ -0,0 +1,68 @@
...
+    ChunkIterFactory emit a data-dependent number of batches per rank,
+    so under DDP each rank would otherwise end its epoch
+    at a different step. Lightning then moves the exhausted rank into
+    validation while the others are still training, and the two sides issue
+    mismatched collectives that deadlock until the NCCL watchdog kills the
+    job. Before yielding each batch, all ranks all-reduce a 1-element has-next
+    flag with MIN, so the epoch ends on every rank at the same batch index.
+
+    Without ``torch.distributed``, batches are yielded unchanged.
+    """
```
Masao-Someki, line 53: Could you add Args and Example section in this docstring?

espnet3/components/data/iterator.py
```diff
@@ -0,0 +1,68 @@
...
+    def __len__(self):
+        """Return the number of batches of the wrapped iterator.
+
+        In distributed runs this is an upper bound: the epoch ends earlier on
+        every rank when the first rank runs out of batches.
+        """
+        return len(self._iterator)
+
+    def generate(self):
+        """Yield this rank's batches."""
```
Masao-Someki: For any function that is not a private function, please add Args, Returns, Raises (optional), Examples, Notes (optional) sections.

espnet3/components/data/iterator.py
```diff
@@ -0,0 +1,68 @@
+"""Base per-epoch batch iterator for ESPnet3."""
+
+import torch
+
+
+class BaseIterator:
```
Masao-Someki: Hmm, I’m not sure if `BaseIterator` is the best name here. In ESPnet, `Base*` usually seems to imply something like an `ABC` that is intended to be inherited.

I’d prefer a name that describes what this class actually does, such as `DistributedSyncIterator`. Do you have any thoughts or other naming suggestions?

Masao-Someki (commented):
Thank you, @NewGamezzz! Basically, this PR looks good to me. We generally want to include examples in docstrings, and I also have one comment about the naming.

I was initially concerned that the per-batch `all_reduce` might increase inter-node communication and slow down multi-node training. However, since it only communicates a very small tensor, and distributed training already involves synchronization at each training step, I think the additional overhead should be quite small in practice.

espnet3/components/data/iterator.py
NewGamezzz: Thanks. I have added more details in the comment.

espnet3/components/data/iterator.py
NewGamezzz: Thanks. I have added comments to all the public functions.

espnet3/components/data/iterator.py
NewGamezzz: I think `EpochSyncIterator` is a good candidate because this wrapper is mainly responsible for ensuring that each epoch has the same number of batches.

Note that an alternative way to fix this is modifying `AbsIterFactory` in `espnet2/iterators/abs_iter_factory.py`. We can replace `build_iter` of `AbsIterFactory` with `__iter__`  of `EpochSyncIterator` and add an abstract function that every subclass implements instead.

I did not go this way because espnet2's own Trainer already all-reduces an iterator_stop flag [(espnet2/train/trainer.py:570-580)](https://github.com/espnet/espnet/blob/master/espnet2/train/trainer.py#L567-L581), so espnet2 DDP runs would synchronize twice per batch.

Masao-Someki: I agree not implement in the AbsIterFactory.
Let's use the `EpochSyncIterator`!

Masao-Someki (approved):

sw005320:
Thanks, @NewGamezzz!

PR #6518, opened 2026-08-10
Lhotse-based training for a LibriSpeech 100h baseline example

Marvin84:
## What did you change?

This PR adds an example for Lhotse-based training for the LibriSpeech 100h baseline.

The baseline was developed before the current version was available in main, so it was based on Masao's branch with some adjustments for this example.

The main changes are in the Lhotse-related files, particularly the builder and dataset implementations. Some functions are also reused from the LibriSpeech 100h examples in Masao's branch.

This implementation is not intended to be a final or necessarily optimal solution. There may be other ways to structure or handle some parts of the code, and I do not expect the current implementation to be perfect.

However, I have tested the training and confirmed that it runs successfully.

---

Allowing Lhotse-based data preparation

---

## Is your PR small enough?
 Yes

---

## Additional Context
None

## Summary by CodeRabbit

- **New Features**
  - Added Lhotse-based preparation and loading for the LibriSpeech 100-hour dataset.
  - Added ASR training configurations with E-Branchformer, BPE tokenization, augmentation, mixed precision, and validation settings.
  - Added dataset statistics collection for feature normalization.
  - Added configurable inference batching, defaulting to four utterances.

- **Improvements**
  - Dataset samples support position, numeric identifier, or utterance ID access.
  - Dataset preparation validates required LibriSpeech splits.

- **Changes**
  - Removed demo packaging, user interface, and upload configuration.


sw005320:
Could you fix the CI issues?
https://github.com/espnet/espnet/actions/runs/31374470193/job/93410506483?pr=6518

Masao-Someki:
Hi @Marvin84, thanks for making the PR!

I noticed that some files still contain Git conflict markers, such as `>>>>>>> ee641f5eb0 (start lbs100h baseline)`.
Could you please re-check the affected files, resolve any remaining conflict markers, and make sure the codebase is in a finalized state?

Marvin84:
>  Git conflict markers, such as `>>>>>>> ee641f5eb0 (start lbs100h baseline)`. 

Sorry about that, I overlooked the config files.
I just pushed a commit and will follow up to see if all tests run correctly.

egs3/librispeech_100/asr/conf/tuning/training_e_branchformer.yaml
```diff
@@ -18,6 +18,7 @@
 # increase `batch_bins` by the same factor when GPU memory permits. For example,
 # halve `accumulate_grad_batches` and double `batch_bins`.
 
+
```
Masao-Someki: This is fine, but let's minimize the diff!

egs3/librispeech_100/asr/conf/demo.yaml
```diff
@@ -17,7 +17,23 @@
 # - a local packed model directory created by `pack_model`
 # - a Hugging Face model repo tag usable by `InferenceModel.from_pretrained()`
 # Relative local paths are resolved from the packed demo directory at runtime.
-# Left unset here: the default `exp/${exp_tag}/model_pack` already matches
-# this recipe's layout.
+
```
Masao-Someki: I think we don't need this change, because this PR is to add lhotse support?

egs3/librispeech_100/asr/conf/inference.yaml
```diff
@@ -25,30 +25,24 @@
 # - `name` becomes `${inference_dir}/<name>/`
 # - `data_src` is omitted, so ESPnet3 loads
 #   `${recipe_dir}/dataset/__init__.py`
-# - `data_src_args` is passed directly to `LibriSpeech100Dataset(...)`
```
Masao-Someki, line 28: Same as the demo.yaml, I think we don't need to change this file

egs3/librispeech_100/asr/conf/metrics.yaml
```diff
@@ -3,6 +3,7 @@
 ##########################################################
 # Common usage:
 #   python run.py --stages measure \
+
```
Masao-Someki: Same as `egs3/librispeech_100/asr/conf/tuning/training_e_branchformer.yaml`, let's minimize the diff!

egs3/librispeech_100/asr/conf/training.yaml
```diff
@@ -0,0 +1,210 @@
+##########################################################
```
Masao-Someki, line 1: For espnet3, we want to keep the file names to noun. So we can rename this as `training.yaml`
Also if this does not have change with the current config, I think we can remove this file?

egs3/librispeech_100/asr/conf/train_with_lhotse.yaml
```diff
@@ -0,0 +1,212 @@
+##########################################################
```
Masao-Someki, line 1: Is this basically the same as the `egs3/librispeech_100/asr/conf/tuning/training_e_branchformer.yaml`?
Sorry I could not find the major difference..

egs3/librispeech_100/asr/dataset/__init__.py
```diff
@@ -1,8 +1,10 @@
 """LibriSpeech 100h dataset module."""
 
-from egs3.librispeech_100.asr.dataset.builder import (
```
Masao-Someki, line 3: Not all the people uses lhotse, so let's create a new flag in the `egs3/librispeech_100/asr/dataset/config.yaml` where we want to use lhotse or not.

egs3/librispeech_100/asr/dataset/data_utils.py
```diff
@@ -0,0 +1,6 @@
+from enum import Enum
```
Masao-Someki, line 1: Do we need the enum class?
I think you can just use list of str like `["recordings", "supervisions"]`, which is more simple.

egs3/librispeech_100/asr/dataset/lhotse_builder.py
```diff
@@ -0,0 +1,251 @@
...
+
+from __future__ import annotations
+
+import os
+import subprocess
+from importlib import resources
+from pathlib import Path
+from typing import Iterable
+
+from lhotse import CutSet, RecordingSet, SupervisionSet
```
Masao-Someki: Let's have try-error here?
Not everyone uses lhotse, so I wanted to make it an optional dependency.

You can refer to `espnet2/text/cleaner.py`

egs3/librispeech_100/asr/dataset/lhotse_builder.py
```diff
@@ -0,0 +1,251 @@
...
+    config_resource = resources.files(__package__).joinpath("lhotse_config.yaml")
+    with resources.as_file(config_resource) as config_path:
+        return load_config_with_defaults(str(config_path), resolve=False)["builder"]
+
+
+_CFG = _load_builder_config()
+_REQUIRED_SPLITS = {str(split) for split in _CFG["required_splits"]}
+
+
+def iter_source_candidates(
```
Masao-Someki: Let's simply use `egs3/librispeech_100/asr/dataset/builder.py:resolve_librispeech_root`

egs3/librispeech_100/asr/dataset/lhotse_builder.py
```diff
@@ -0,0 +1,251 @@
...
+    )
+
+
+def missing_required_splits(source_root: Path) -> list[str]:
+    """Return required split names that are missing from ``source_root``."""
+    return [
+        str(split)
+        for split in _CFG["required_splits"]
+        if not (source_root / str(split)).is_dir()
+    ]
```
Masao-Someki: For these I think we can import from `egs3/librispeech_100/asr/dataset/builder.py`

egs3/librispeech_100/asr/dataset/lhotse_builder.py
```diff
@@ -0,0 +1,251 @@
...
+
+
+class LibriSpeech100LhotseBuilder(DatasetBuilder):
```
Masao-Someki, line 47: I think you can inherit LibriSpeech100Builder, like:

```
class LibriSpeech100LhotseBuilder(LibriSpeech100Builder):
```

I think this will make this class much simpler.

egs3/librispeech_100/asr/dataset/lhotse_builder.py
```diff
@@ -0,0 +1,251 @@
...
+        dataset_dir.mkdir(parents=True, exist_ok=True)
+
+        manifests = prepare_librispeech(
+            corpus_dir=source_root,
+            dataset_parts=_REQUIRED_SPLITS,
+            num_jobs=num_jobs,
+        )
+
+        for split, split_manifests in manifests.items():
+            for element in LhotseElement:
```
Masao-Someki, line 134: This `LhotseElement` I think we can replace with a list

egs3/librispeech_100/asr/dataset/lhotse_config.yaml
```diff
@@ -0,0 +1,17 @@
+builder:
```
Masao-Someki, line 1: Let's add one flag in the `egs3/librispeech_100/asr/dataset/config.yaml`, so that anyone who wants to use the lhotse dataset, they can activate it in the config.
Such as `builder.use_lhotse`

egs3/librispeech_100/asr/src/utils/collect_stats_local.py
```diff
@@ -0,0 +1,204 @@
+from __future__ import annotations
```
Masao-Someki, line 1: Is there any reason that we need to create collect-stats files under recipe? Do we need any lhotse-specific process for collect-stats?

egs3/librispeech_100/asr/src/utils/collect_stats_local.py
```diff
@@ -0,0 +1,204 @@
...
+
+    Returns:
+        If write_collected_feats:
+            Tuple[stats, shape_info, feats]
+        else:
+            Tuple[stats, shape_info]
+    """
+
+    items = [dataset[i] for i in idxs]
+    uids = [d.pop("utt_id", i) for i, d in enumerate(items)]
```
Masao-Someki, line 37: In this implementation you add `utt_id` only when collect-stats, but removing it here? Sorry I am not sure why this is required?

egs3/librispeech_100/asr/src/utils/collect_stats_parallel.py
```diff
@@ -0,0 +1,210 @@
+# espnet3/collect_stats/collect_parallel.py
```
Masao-Someki, line 1: ditto, why we need this under recipe?

egs3/librispeech_100/asr/src/collect_stats.py
```diff
@@ -0,0 +1,119 @@
+from __future__ import annotations
```
Masao-Someki, line 1: ditto

egs3/librispeech_100/asr/src/librispeech_system.py
```diff
@@ -0,0 +1,113 @@
...
+from omegaconf import DictConfig, OmegaConf
+
+from egs3.librispeech_100.asr.src.collect_stats import collect_stats
+from espnet3.parallel.parallel import set_parallel
+from espnet3.systems.asr.system import ASRSystem
+
+logger = logging.getLogger(__name__)
+
+
+class LBSSystem(ASRSystem):
```
Masao-Someki: Ususally in espnet3 we don't allow short variable name. So this should be `LibriSpeechSystem`, but I am not sure why we need a new system here..
Basically `espnet3/systems/asr/system.py` should be fine I think..

egs3/librispeech_100/asr/run.py
```diff
@@ -6,6 +6,8 @@
 )
 from espnet3.systems.asr.system import ASRSystem
 
+# from egs3.librispeech_100.asr.src.librispeech_system import LBSSystem
```
Masao-Someki: If this is not used, I think we can remove the system file?

Masao-Someki (commented):
Thank you @Marvin84, I added comments on this PR!

egs3/librispeech_100/asr/src/utils/collect_stats_local.py
Marvin84: Hi, when performing `collect_stats` through some old code in `espnet2` this was required. Probably, there is now a version incompatibility. I can eliminate this, but I would need to retest the pipeline again.

egs3/librispeech_100/asr/src/utils/collect_stats_parallel.py
Marvin84: As above, this is certainly not needed in `espnet3` implementation.

egs3/librispeech_100/asr/conf/train_with_lhotse.yaml
Marvin84: There are minor differences in the `dataset` definition

egs3/librispeech_100/asr/dataset/__init__.py
Marvin84: Ok, please let me know what flag you would like to add and where.

egs3/librispeech_100/asr/dataset/data_utils.py
Marvin84: I personally think that `enums` are a good practice. They are easily extendible with more functionalities and avoid using pure strings.

egs3/librispeech_100/asr/dataset/lhotse_builder.py
Marvin84: Regarding `missing_required_splits `, this is using a global variable `_CFG` that might differ from the one used within `egs3/librispeech_100/asr/dataset/builder.py`. How do we handle this?

egs3/librispeech_100/asr/dataset/lhotse_builder.py
Marvin84: This will create an intermediate dependency on the standard Librspeech class. If this is what you intend to do, I can change it. However, I belive it would be a cleaner solution if this inheritated the `DatasetBuilder` directly.

egs3/librispeech_100/asr/dataset/lhotse_builder.py
Marvin84: See my comment above regarding good practice. To be discussed.

egs3/librispeech_100/asr/dataset/lhotse_config.yaml
Marvin84: See above. To be discussed.

egs3/librispeech_100/asr/src/utils/collect_stats_local.py
Marvin84: No, this was just something I used before I added lhotse.

egs3/librispeech_100/asr/src/librispeech_system.py
Marvin84: I believe this is again because of old usage of `collect_stats`, I will eliminate this class. This is not relevant for `lhotse` integration.

egs3/librispeech_100/asr/conf/train_with_lhotse.yaml
```diff
@@ -0,0 +1,212 @@
...
+
+create_dataset:
+  recipe_dir: ${recipe_dir}
+
+
+dataset:
+  _target_: espnet3.components.data.data_organizer.DataOrganizer
+  train:
+    - name: train-clean-100
+      data_src: librispeech_100/asr
```
Masao-Someki, line 31: Let's use `data_src` to point to the LhotseDataset class module here, like `data_src: egs3.librispeech_100.asr.dataset.lhotse`?
Or you can set something like:

```
lhotse_data_src: egs3.librispeech_100.asr.dataset.lhotse

dataset:
  train:
    - data_src: ${lhotse_data_src}
      data_src_args:
        split: train-clean-100
  valid:
    - data_src: ${lhotse_data_src}
      data_src_args:
        split: dev-clean
```

egs3/librispeech_100/asr/conf/train_with_lhotse.yaml
```diff
@@ -0,0 +1,212 @@
...
+    train: true
+    token_type: bpe
+    token_list: ${tokenizer.save_path}/tokens.txt
+    bpemodel: ${tokenizer.save_path}/bpe.model
+    text_cleaner:
+    fs: 16000
+    data_aug_effects:
+      - - 0.6
+        - - [0.5, "speed_perturb", {"factor": 0.9}]
+          - [0.5, "speed_perturb", {"factor": 1.1}]
```
Masao-Someki, line 64: Would you check the `egs3/librispeech_100/asr/conf/tuning/training_e_branchformer.yaml`, actually we fixed this config in #6513

egs3/librispeech_100/asr/conf/data_build_config.yaml
Masao-Someki, line 1: Do we need this config file? Also we need to remove user-specific path like `/scratch/work/raissit1/dev/` if we need to include this in espnet master.

egs3/librispeech_100/asr/conf/demo.yaml
Masao-Someki, line 1: I think we can remove this diff also?
I think the values can be ported from the default config at `TEMPLATE/asr/conf/`

Maybe we can keep `upload_demo.hf_repo`

egs3/librispeech_100/asr/conf/inference.yaml
```diff
@@ -26,29 +26,24 @@
 # - `data_src` is omitted, so ESPnet3 loads
 #   `${recipe_dir}/dataset/__init__.py`
 # - `data_src_args` is passed directly to `LibriSpeech100Dataset(...)`
-#
-# Dataset splits:
-#   test:
-#     - split="dev-clean" from `LibriSpeech100Dataset`
-#     - split="dev-other" from `LibriSpeech100Dataset`
-#     - split="test-clean" from `LibriSpeech100Dataset`
-#     - split="test-other" from `LibriSpeech100Dataset`
```
Masao-Someki, line 35: Please don't remove comments, we always want to make espnet3 documents/comments to be very informative!
You can add more information on the comment, that is always welcome!

egs3/librispeech_100/asr/conf/inference.yaml
```diff
@@ -26,29 +26,24 @@
...
+#    - name: dev-clean
+#      data_src_args:
+#        split: dev-clean
+#
+#    - name: test-clean
+#      data_src_args:
+#        split: test-clean
+#    - name: test-other
+#      data_src_args:
+#        split: test-other
```
Masao-Someki, line 46: I think this was for debugging? Could you remove this diff?

egs3/librispeech_100/asr/conf/training.yaml
Masao-Someki, line 1: I just want to confirm whether we need this file. Could you share why it needs to be separate from `egs3/librispeech_100/asr/conf/tuning/training_e_branchformer.yaml`?

egs3/librispeech_100/asr/dataset/data_utils.py
Masao-Someki: I agree that an enum would make this easier to extend, but I'm not sure whether we actually expect `LhotseElement` to be extended in the future.

At the moment, it seems to be used only as a collection of fixed string values such as `recording` and `supervision`. If we don't expect more element types to be added, I think a simple list or tuple would be easier to read.

egs3/librispeech_100/asr/dataset/lhotse_builder.py
Masao-Someki: I think most of the differences in the Lhotse preparation are in the `build` function and the functions related to it, right?

For example, I think we can share the same `prepare_source` implementation. If we need to change that logic in the future, the same change should also be reflected in the Lhotse builder.

egs3/librispeech_100/asr/dataset/lhotse_builder.py
Masao-Someki: Thanks. I added a comment above as well, but I feel that using a list of strings would be simpler and easier to read and understand...

egs3/librispeech_100/asr/dataset/lhotse_builder.py
```diff
@@ -0,0 +1,207 @@
...
+                split_manifests[element.value].to_file(
+                    dataset_dir / f"{split}_{element.value}.{file_format}"
+                )
+
+    def load_cutsets(
+        self,
+        split: str,
+        dataset_dir: str | Path | None = None,
+    ) -> CutSet:
```
Masao-Someki, line 143: This function is not used by `DatasetBuilder`, it is specific to the `Dataset` class. Please move it to the `Dataset` class.
Also, I don't think this function will be called from outside the Dataset class, so we can make it a private method.

egs3/librispeech_100/asr/src/utils/collect_stats_local.py
Masao-Someki: Got it, then could you remove this file?

egs3/librispeech_100/asr/src/utils/collect_stats_parallel.py
Masao-Someki: Thanks, then could you remove this file?

egs3/librispeech_100/asr/src/collect_stats.py
Masao-Someki: Could you also remove this file?

egs3/librispeech_100/asr/dataset/__init__.py
Masao-Someki: I just realized that we can specify the dataset module through `data_src`, so we don't need an additional flag!

Could you simply remove this diff? We want to keep the original implementation as the default dataset class, and we can specify the Lhotse dataset module directly in the training config when needed.

Masao-Someki (commented):
Thank you @Marvin84 , I have added some additional comments!

egs3/librispeech_100/asr/src/utils/collect_stats_parallel.py
Marvin84: I removed the whole `utils` folder, given that it has nothing to do with this PR

PR #6536, opened 2026-08-26
Report GPU-only `espnet2/sds` tests as skipped rather than passed

amirlankalm:
## What did you change?

Ten of the twelve test functions under `test/espnet2/sds` began with a bare `return`
instead of a skip marker. They now use `pytest.mark.skipif`.

```diff
+@pytest.mark.skipif(not torch.cuda.is_available(), reason="requires a GPU")
 def test_forward():
-    if not torch.cuda.is_available():
-        return  # Only GPU supported
     asr_model = ESPnetASRModel(
```

Nine files, +13/-20. `import pytest` is added where it was not already imported, the
marker is placed above any existing decorators so the markers stay together, and the
`pytest.importorskip()` calls inside the affected function bodies are untouched.

---

A bare `return` makes pytest record the test as **passed**. `test_python_espnet2` runs on
`ubuntu-latest`, which has no GPU, so on every PR these tests report success while
asserting nothing, and `espnet2/sds` looks covered when it is not.

That is how the audio problem in #6527 went unnoticed:
`test/espnet2/sds/end_to_end/test_mini_omni_e2e.py::test_forward` was reporting a pass on
CI for a `forward()` that inserted silence at every chunk boundary and dropped the last
256 ms of every response.

On a machine without CUDA, before and after:

```
before:  9 passed, 2 skipped
after:   2 passed, 9 skipped
```

with each skip now labelled:

```
SKIPPED [1] test/espnet2/sds/asr/test_espnet_asr.py:7: requires a GPU
SKIPPED [1] test/espnet2/sds/asr/test_owsm_asr.py:7: requires a GPU
SKIPPED [2] test/espnet2/sds/asr/test_owsm_ctc_asr.py:7: requires a GPU
SKIPPED [2] test/espnet2/sds/asr/test_whisper_asr.py:16: requires a GPU
SKIPPED [1] test/espnet2/sds/end_to_end/test_mini_omni_e2e.py:11: requires a GPU
SKIPPED [1] test/espnet2/sds/llm/test_hugging_face_llm.py:9: requires a GPU
SKIPPED [1] test/espnet2/sds/tts/test_chat_tts.py:7: requires a GPU
```

The two remaining passes are the tests that genuinely run on CPU,
`utils/test_chat.py` and `vad/test_webrtc_vad.py`.

**Nothing runs that did not run before, and nothing stops running.** Only the reported
status changes, from a pass that means nothing to a skip that means what it says. The
intent is that "green" stops implying "exercised" for this package.

---

## Is your PR small enough?

Yes. 9 files, +13/-20.

---

## Additional Context

```
pycodestyle <the 9 files>          # exit 0
pre-commit run --files <the 9>     # all hooks pass, including isort on the new imports
pytest test/espnet2/sds            # 2 passed, 9 skipped
```

`flake8` output is unchanged: 19 pre-existing `D100`/`D103` docstring notices before and
after.

Two things I did not do here, to keep this to one concern:

- I did not add CPU-safe replacements for the coverage these tests are not providing.
  #6527 and #6528 each add one for the specific behaviour they change
  (`test_mini_omni_e2e_forward.py`, `test_mini_omni_e2e_init.py`), stubbing the client and
  the decoder so they need no checkpoint, GPU or ffmpeg and run in about two seconds. That
  pattern would extend to the ASR, LLM and TTS wrappers, but it is a lot more code and
  seemed better as its own PR if you want it.
- I did not touch the `--execution-timeout 10.0` in `ci/test_python_espnet2.sh`. Worth
  noting that a real mini-omni forward pass takes about 16 s on CPU for a 2 s clip, so
  these tests could not run under the current timeout even on a GPU runner. If you would
  like them exercised somewhere, that is the other thing that would need to change.

Following up on your note to go ahead with the additional PRs.


amirlankalm:
The red check here, `test_python_espnet2 (3.12, 2.11.0, false, 6.0.0)`, is not from this
change. The six failures are all in `test/espnet2/gan_tts/{jets,vits}` and are Hugging Face
rate limiting:

```
E  urllib.error.HTTPError: HTTP Error 403: rate limit exceeded
E  KeyError: 'Authorization'
```

Nothing under `test/espnet2/sds` failed. The run summary is
`6 failed, 56882 passed, 812 skipped`, and the same job passes on the other matrix entries.

sw005320:
Could you apply this to the other parts of the test?

amirlankalm:
Sure, on it.

amirlankalm:
Done, added the other 27 across asr, enh, slu, hubert and beats.

amirlankalm:
Unrelated to this, evaluate_pseudomos hits the GitHub API rate limit inside torch.hub. All six test_python_espnet2 jobs pass.

sw005320:
Thanks!
Since we only support >=2.9.1 (https://github.com/espnet/espnet), instead of keeping the PyTorch version condition, we can remove it from the source.

amirlankalm:
Yeah that makes sense, removed them. 25 markers plus 6 flag defs and imports that went unused with them. Left the compound ones (`not is_python_3_8_plus or not is_torch_1_7_plus`) and the inline `if is_torch_1_9_plus: ... else: ...` branches alone since dropping those changes behaviour and they're outside this PR, there's ~96 of them across 21 more files if you want a follow-up. btw pyproject still says `torch>=2.3.1`.

sw005320 (approved):
Thanks a lot!
After the CI check, I'll merge this PR

PR #6538, opened 2026-08-26
Add an `sds` extra so `espnet2/sds` can be installed

amirlankalm:
## What did you change?

Added an `sds` extra to `pyproject.toml` covering what `espnet2/sds` actually imports.

```toml
sds = [
  "pydub",
  "snac",
  "litgpt==0.4.3",
  "openai-whisper",
  "webrtcvad-wheels",
  "gradio",
  "transformers",
]
```

Nothing else is touched, and the extra is not added to `all` (see the last note below).

---

There is currently no supported way to install `espnet2/sds` from the repo. Following
CONTRIBUTING 4 and then importing the mini-omni backend:

```
$ pip install -e ".[test]"
$ python -c "from espnet2.sds.end_to_end.mini_omni.inference import OmniInference"
ModuleNotFoundError: No module named 'whisper'
```

An AST scan of `espnet2/sds/**` against `pyproject.toml`, `ci/install.sh` and
`tools/Makefile`:

| package | imported by | declared before this PR |
|---|---|---|
| `litgpt` | `end_to_end/mini_omni/litgpt/__init__.py`, which imports the real package | no |
| `snac` | `end_to_end/mini_omni/inference.py` | no |
| `openai-whisper` | `end_to_end/mini_omni/inference.py`, `asr/whisper_asr.py` | only via `tools/Makefile: whisper.done` |
| `webrtcvad` | `vad/webrtc_vad.py` | no |
| `pydub` | `end_to_end/mini_omni_e2e.py` | only transitively, through `gradio` |
| `gradio` | `espnet_model.py`, at module level | only the `test` extra |
| `transformers` | `llm/hugging_face_llm.py` | only the `all` and `speechlm` extras |

`gradio` and `transformers` are included because `ESPnetSDSModelInterface.__init__`
raises `ImportError` without gradio, and `HuggingFaceLLM` needs transformers, so the
module is unusable without them even though they are reachable through other extras.

The working version set has so far only existed in the `requirements.txt` of an external
Hugging Face Space (`Siddhant/Voice_Assistant_Demo`), last modified in February 2025, which
`egs2/TEMPLATE/sds1/README.md` points at under Troubleshooting. That is easy to drift from
silently.

**Why `litgpt` is pinned.** `litgpt` 0.5.13 declares `huggingface-hub<1.4`, while
`gradio` 6.26.0 declares `huggingface-hub<2.0,>=1.16.0`. Those cannot both be satisfied, so
an unpinned entry does not fail loudly, it makes pip quietly downgrade gradio instead.
`litgpt==0.4.3` is the release the vendored mini-omni code was adapted from and the last one
with no upper bound on `huggingface-hub`, so it removes the clash rather than working around
it.

---

## Is your PR small enough?

Yes. 1 file, +18.

---

## Additional Context

I verified `litgpt==0.4.3` rather than assuming it from the Space's pin:

```
litgpt 0.4.3 with huggingface-hub 1.28.0
  from espnet2.sds.end_to_end.mini_omni.inference import OmniInference   ok
  from ...mini_omni.litgpt.model import GPT, Config                      ok
  from ...mini_omni.litgpt import Tokenizer                              ok
  MiniOmniE2EModel(device="cpu") constructed in 5.7 s
```

so the whole vendored stack imports and the model builds and loads its weights on the
pinned version.

The complete set is installed together in my environment at the versions this extra
resolves to, and is self-consistent:

```
pydub 0.25.1 · snac 1.2.1 · litgpt 0.4.3 · openai-whisper 20250625
webrtcvad-wheels 2.0.14 · gradio 6.26.0 · transformers 5.15.1 · huggingface-hub 1.28.0

$ pip check
No broken requirements found.
```

`webrtcvad-wheels` rather than `webrtcvad` because it ships prebuilt wheels; the original
needs a compiler. The Space lists both.

`ChatTTS` is deliberately excluded. `espnet2/sds/tts/chat_tts.py` imports it lazily inside
`ChatTTSModel.__init__`, so it is only needed when that one TTS backend is selected, and it
is not needed for the end-to-end path.

Two things I want to check with you rather than decide:

- **I did not add `sds` to the `all` extra.** Doing so would make `ci/install.sh`
  (`pip install -e ".[test,doc,all]"`) install these, which would in turn let the
  `espnet2/sds` tests actually run instead of being skipped for missing imports. That seemed
  worth having, but it also puts a `litgpt==0.4.3` pin into every CI install, so I would
  rather you decide.
- **#6439 "Clean dependencies" is open** and reworking the base `dependencies` list. This
  change only adds a block under `[project.optional-dependencies]`, so I do not expect a
  textual conflict, but if you would prefer this to land after #6439 I am happy to rebase.

Following up on your note to go ahead with the additional PRs; this is the second of the two
I mentioned, alongside #6536.


sw005320 (approved):

sw005320:
Thanks, @amirlankalm!

amirlankalm:
> Thanks, @amirlankalm!

happy to help!

PR #6540, opened 2026-08-27
Add ESPnet-Sidon feature predictor recipe

Dahee96:
## What did you change?

Added an ESPnet recipe for Sidon under `egs2/libritts_r/sidon`.

---

To provide a Sidon reproduction recipe in ESPnet.

---

## Is your PR small enough?

Yes
---

## Summary by CodeRabbit

* **New Features**
  * Added a complete LibriTTS-R speech-enhancement workflow, from data preparation and training through inference and evaluation.
  * Added Sidon enhancement modeling with frozen w2v-BERT 2.0 features and optional LoRA adaptation.
  * Added audio degradation simulation, resampling, room-impulse-response generation, and chunked waveform restoration.
  * Added evaluation with DNSMOS, NISQA, speaker similarity, WER, PESQ, STOI, and VERSA metrics.
  * Added support for local and cluster-based execution environments.


egs2/libritts_r/enh1/local/score.py
sw005320, line 1: It's good to use versa
@ZhuoyanTao can help

egs2/libritts_r/enh1/run.sh
sw005320, line 1: This looks good

espnet2/main_funcs/collect_stats.py
```diff
@@ -55,6 +55,8 @@ def collect_stats(
                 for name in batch:
                     if name.endswith("_lengths"):
                         continue
+                    if not isinstance(batch[name], torch.Tensor):
```
sw005320: what's this?
why we need this?

espnet2/bin/enh_inference_sidon.py
sw005320, line 1: How are they different from `enh_inference`?
Do we need to create a new inference?

espnet2/bin/enh_train_sidon.py
sw005320, line 1: ditto
How are they different from `enh_train`?
Do we need to create a new inference?

espnet2/enh/sidon_model.py
sw005320, line 1: ditto
How are they different from `enh_model`?
Do we need to create a new inference?

espnet2/tasks/sidon.py
sw005320, line 1: ditto

sw005320 (commented):
- it's better to change the directory name. `rstr1`? Or it seems that it is almost equivalent to `enh` and we can use `enh1`?

espnet2/main_funcs/collect_stats.py
Dahee96: The modification was originally needed because the Sidon collate function added nested dictionaries, noisy_speech_ssl and speech_ref1_ssl, to each batch. The original collect_stats.py assumes that each batch entry is a tensor and therefore failed on those dictionaries.  
However, SSL inputs are not needed during statistics collection. I have reverted the common collect_stats.py modification and updated SidonTask so that it does not construct the SSL inputs or load the w2v-BERT models when collect_stats=True.

espnet2/tasks/sidon.py
Dahee96: I considered integrating Sidon into `EnhancementTask`. However, `EnhancementTask` and `ESPnetEnhancementModel` assume an encoder–separator–decoder architecture trained with waveform- or spectral-domain enhancement losses.  
Sidon instead uses a frozen w2v-BERT teacher and a LoRA-adapted student. The student predicts the clean teacher representation at layer 8, and the training loss is computed directly in the SSL feature domain. It has no waveform decoder during training. Its collate function also needs to construct paired noisy/clean w2v-BERT inputs after applying the Sidon degradation pipeline.  
Supporting this through `EnhancementTask` would require adding Sidon-specific model selection, preprocessing, teacher–student feature extraction, and loss handling to the common enhancement task. For this reason, I implemented a separate `SidonTask` to keep the existing enhancement pipeline unchanged.

espnet2/enh/sidon_model.py
Dahee96: `sidon_model.py` defines the actual teacher–student feature predictor. This cannot be represented directly by the current `ESPnetEnhancementModel`, because the Sidon model has no conventional separator or waveform decoder during training. It compares the student and frozen-teacher representations in the w2v-BERT feature space. Therefore, a dedicated model class is required even though the final application is speech enhancement.

espnet2/bin/enh_inference_sidon.py
Dahee96: `enh_inference_sidon.py` is required because inference combines an ESPnet feature-predictor checkpoint with the separately distributed official Sidon TorchScript vocoder. The existing `enh_inference.py` expects a single `ESPnetEnhancementModel` and directly produces waveforms through its encoder–separator–decoder pipeline. Sidon instead produces intermediate SSL features and converts them to 48-kHz speech using an external vocoder, so the existing inference interface cannot be reused without substantial Sidon-specific branching.

egs2/libritts_r/enh1/run.sh
Dahee96: I renamed the recipe directory from `sidon` to `enh1`

egs2/libritts_r/enh1/local/score.py
ZhuoyanTao: Hi @Dahee96 — following up on @sw005320's suggestion to use versa for scoring. I took a look at the recipe and put together a proposal for stage 8 below.

From the code:

- `enh_inference_sidon.py` already writes `wav.scp` alongside the wav dir, so we can use it directly instead of globbing the wav directory
- The inference output is at 48 kHz while LibriTTS-R references are 24 kHz. Versa's scorer automatically downsamples pred+gt to the lower of their two rates before any metric runs (pass 1: 48→16 kHz via noisy gt; pass 2: 48→24 kHz via clean ref), then each metric resamples further to its required rate internally. No `fs` override needed in the yamls
- Enhancement always has the noisy input (it's the model's input), but a clean reference is only available when you have paired data (e.g. synthetically degraded speech). So metrics are split into two passes: pass 1 passes the noisy input as versa's `--gt` argument (always available), pass 2 uses a clean reference and is gated on `--ref_wav_scp`

---
Therefore, I propose we add two config files for versa in conf/, and add wiring with versa via a new stage 8 at the end of run.sh

`conf/versa_enh.yaml` — always runs. Ref-free metrics (DNSMOS, UTMOS, WER) ignore `--gt`; speaker uses it for SpkSim vs degraded input, matching the Sidon paper metric.

```yaml
- name: espnet_wer
  model_tag: default
  beam_size: 5
  text_cleaner: whisper_basic

- name: pseudo_mos
  predictor_types: ["dnsmos", "utmos"]

# SpkSim vs noisy input — measures speaker preservation under restoration.
- name: speaker
  model_tag: default
```

`conf/versa_enh_ref_based.yaml` — only when clean reference is available.

```yaml
- name: pesq

- name: stoi

- name: speaker
  model_tag: default
```
Note: speaker here computes speaker similarity between output and clean reference (when we have one). Might be redundant.

---

### Stage 8 addition to `run.sh`

Add `ref_wav_scp=` to the header variables alongside the other defaults:

```bash
ref_wav_scp=   # optional: set to clean reference wav.scp to enable PESQ/STOI
```

Then append:

```bash
if [ ${stage} -le 8 ] && [ ${stop_stage} -ge 8 ]; then
    for test_set in ${test_sets}; do
        log "Stage 8: versa scoring (${test_set})"
        _inf_dir=${expdir}/inference_${test_set}
        _eval_dir=${_inf_dir}/scoring/versa_eval
        mkdir -p "${_eval_dir}"

        _pred_scp="${_inf_dir}/wav.scp"   # written by enh_inference_sidon
        _noisy_scp="data/${test_set}_16k/wav.scp"

        _nj=$(( nj < $(wc -l < "${_pred_scp}") ? nj : $(wc -l < "${_pred_scp}") ))

        _split_pred=""
        for n in $(seq "${_nj}"); do _split_pred+="${_eval_dir}/pred.${n} "; done
        utils/split_scp.pl "${_pred_scp}" ${_split_pred}

        _split_noisy=""
        for n in $(seq "${_nj}"); do _split_noisy+="${_eval_dir}/noisy.${n} "; done
        utils/split_scp.pl "${_noisy_scp}" ${_split_noisy}

        _text_opt=""
        if [ -f "data/${test_set}/text" ]; then
            _text_opt="--text data/${test_set}/text"
        fi

        # Pass 1 (always): WER + DNSMOS/UTMOS + SpkSim-vs-noisy.
        # Ref-free metrics ignore --gt; speaker uses it for SpkSim vs degraded input.
        ${decode_cmd} JOB=1:"${_nj}" "${_eval_dir}/versa.JOB.log" \
            ${python} -m versa.bin.scorer \
                --pred "${_eval_dir}/pred.JOB" \
                --score_config conf/versa_enh.yaml \
                --cache_folder "${_eval_dir}/cache" \
                --gt "${_eval_dir}/noisy.JOB" \
                ${_text_opt} \
                --output_file "${_eval_dir}/result.JOB.txt" \
                --io soundfile
        ${python} pyscripts/utils/aggregate_eval.py \
            --logdir "${_eval_dir}" --scoredir "${_eval_dir}" --nj "${_nj}"
        log "Versa results: ${_eval_dir}/avg_result.txt"

        # Pass 2 (optional): PESQ + STOI + SpkSim-vs-ref, only with clean reference.
        if [ -n "${ref_wav_scp}" ] && [ -f "${ref_wav_scp}" ]; then
            _rb_dir="${_inf_dir}/scoring/versa_ref"
            mkdir -p "${_rb_dir}"
            _split_gt=""
            for n in $(seq "${_nj}"); do _split_gt+="${_rb_dir}/gt.${n} "; done
            utils/split_scp.pl "${ref_wav_scp}" ${_split_gt}
            ${decode_cmd} JOB=1:"${_nj}" "${_rb_dir}/versa.JOB.log" \
                ${python} -m versa.bin.scorer \
                    --pred "${_eval_dir}/pred.JOB" \
                    --score_config conf/versa_enh_ref_based.yaml \
                    --cache_folder "${_rb_dir}/cache" \
                    --gt "${_rb_dir}/gt.JOB" \
                    --output_file "${_rb_dir}/result.JOB.txt" \
                    --io soundfile
            ${python} pyscripts/utils/aggregate_eval.py \
                --logdir "${_rb_dir}" --scoredir "${_rb_dir}" --nj "${_nj}"
            log "Versa ref-based results: ${_rb_dir}/avg_result.txt"
        else
            log "Skipping ref-based metrics (no --ref_wav_scp provided)"
        fi
    done
fi
```

To run:

```bash
# Without clean reference
./run.sh --stage 8 --stop_stage 8

# With clean LibriTTS-R reference
./run.sh --stage 8 --stop_stage 8 --ref_wav_scp data/test-clean/wav.scp
```

---

Longer-term: this versa stage would ideally land in `egs2/TEMPLATE/enh1/enh.sh` the same way `tts.sh` has it, so all enhancement recipes benefit. Happy to help draft that as a follow-up.

Dahee96: @ZhuoyanTao Thanks for the suggestion. I've added VERSA scoring in Stage 8.

PR #6542, opened 2026-08-27
Make `MiniOmniE2EModel` generation settings configurable

amirlankalm:
## What did you change?

`MiniOmniE2EModel` now exposes the generation settings that `OmniInference.run_AT_batch_stream`
already accepts, and uses the ones it was already storing.

```diff
     def __init__(
         self,
         device: str = "cuda",
         dtype: str = "float16",
+        stream_stride: int = 4,
+        max_tokens: int = 2048,
+        temperature: float = 0.9,
+        top_k: Optional[int] = 1,
+        top_p: float = 1.0,
     ):
...
-        self.stream_stride = 4
-        self.max_tokens = 2048
+        self.stream_stride = stream_stride
+        self.max_tokens = max_tokens
+        self.temperature = temperature
+        self.top_k = top_k
+        self.top_p = top_p
...
     def forward(
         self,
         array: np.ndarray,
         orig_sr: int,
+        temperature: Optional[float] = None,
+        top_k: Optional[int] = None,
+        top_p: Optional[float] = None,
     ) -> Tuple[str, bytes]:
...
-        stream_stride = 4
-        max_tokens = 2048
+        temperature = self.temperature if temperature is None else temperature
+        top_k = self.top_k if top_k is None else top_k
+        top_p = self.top_p if top_p is None else top_p
 
         with tempfile.NamedTemporaryFile(suffix=".wav", delete=False) as f:
             f.write(data_buff)
             audio_generator = self.client.run_AT_batch_stream(
-                f.name, stream_stride, max_tokens
+                f.name,
+                self.stream_stride,
+                self.max_tokens,
+                temperature=temperature,
+                top_k=top_k,
+                top_p=top_p,
             )
```

**Defaults are unchanged**, so behaviour is identical for existing callers: stride 4, 2048
tokens, `temperature=0.9`, `top_k=1`, `top_p=1.0`.

---

Two halves of the same gap.

**1. The sampling parameters could not be set, and the effective behaviour is greedy.**
`forward()` called `run_AT_batch_stream` without them, so its own defaults always applied,
and one of those is `top_k=1`. In the vendored sampler
(`mini_omni/litgpt/generate/base.py`):

```python
if top_k is not None:
    v, i = torch.topk(logits, min(top_k, logits.size(-1)))
    logits = torch.full_like(logits, float("-inf")).scatter_(-1, i, v)
...
probs = torch.nn.functional.softmax(logits / temperature, dim=-1)
```

With `top_k=1` every logit but one becomes `-inf`, so the softmax is one-hot and the
`multinomial` draw is deterministic. `temperature=0.9` has no effect. Measured on CPU with
the first 2 s of `test_utils/ctc_align_test.wav`, two runs at the shipped defaults:

```
run 1: It sounds like you're asking about the sale of a hotel. Could you please provide ...
run 2: It sounds like you're asking about the sale of a hotel. Could you please provide ...
```

byte-identical. With `top_k` raised, the same input gives different and still fluent
responses:

```
top_k=5             : I'm sorry, but I can't help with that. It's important to respect ...
top_k=20            : I'm sorry, but I'm not able to answer that question. Is there ...
top_k=20, temp=1.2  : It sounds like you might be looking for some details about the sale ...
top_p=0.9           : Could you please provide more details about what you're looking to ...
```

Five distinct outputs across six runs, and none of them degenerate. So the capability is
there in the client; it was just not reachable through the wrapper.

**2. `stream_stride` and `max_tokens` were dead.** They are assigned in `__init__`, but both
`warmup()` and `forward()` declared identical locals and used those, so the attributes were
never read. Setting them had no effect. This is the same shape as the `device` argument in
#6528.

I made the sampling parameters overridable per call in `forward()` rather than only at
construction, because the case that needs them is drawing several candidate responses for a
single input, which means varying them between calls on one model instance. Constructor
values act as the defaults.

---

## Is your PR small enough?

Yes. 2 files, +210/-9: 60 changed lines in `mini_omni_e2e.py` and a new 150-line test file.

---

## Additional Context

Five CPU-only tests, stubbing the client and the SNAC decoder and replacing pydub's
`AudioSegment`, so they need no checkpoint, no GPU and no ffmpeg and run in under a second.
They assert on what reaches `run_AT_batch_stream`.

Against this branch all five pass. Against `master`, **four fail and
`test_defaults_are_unchanged` still passes**, which is the point: the defaults really are
untouched, only the ability to change them is new.

Also checked end to end through the public class on CPU, rather than only against the stub:

```
default, called twice          -> identical text
forward(..., top_k=20, temperature=1.2) -> different text, still fluent
model.top_k after that call    -> still 1, the override did not persist
```

**One design decision worth flagging.** `None` is meaningful to the underlying sampler,
where it disables top-k filtering entirely, and `forward` also needs a way to say "not
overridden". I used `None` for the latter and documented that top-k can only be disabled on
the constructor, rather than introducing a sentinel object, because the signature stays
readable and the lost case is narrow: per-call disabling while the constructor holds a
numeric `top_k`, which a large `top_k` expresses just as well. If you would rather have a
sentinel so `None` keeps one single meaning throughout, say so and I will switch it.

```
pycodestyle espnet2/sds/end_to_end/mini_omni_e2e.py <test file>   # exit 0
pre-commit run --files <both>                                     # all hooks pass
pytest test/espnet2/sds                                           # 18 passed
```

`flake8` count is unchanged at 7, all pre-existing `D` docstring notices on
`mini_omni_e2e.py`; the new test file adds none.

Environment:

```
- python version: 3.12.8
```

Two notes on scope:

- I did **not** change the default `top_k`. Making the demo stochastic by default is a
  behaviour change for everyone and is your call, not mine. This only makes the choice
  reachable, and documents that `top_k=1` means greedy, which is easy to miss sitting next
  to `temperature=0.9`.
- This touches the same part of `forward()` as #6527, so whichever of the two lands second
  will want a small rebase. Happy to do it in either order.

The reason I went looking: I want to sample several candidate responses per turn so that
turn-level preference pairs can be built from the model's own output, and there is currently
no way to do that through the public class. Raising it as a toolkit gap rather than carrying
a patch downstream.


amirlankalm:
Both red checks here are unrelated to the change.

`check_installable_on_macos (3.10, 2.9.1, true)` fails in "install espnet" before any of
this code runs, resolving Intel's conda channel:

```
NameResolutionError("HTTPSConnection(host='software.repos.intel.com', port=443):
Failed to resolve 'software.repos.intel.com' ([Errno 8] nodename nor servname provided...)")
CondaHTTPError: HTTP 000 CONNECTION FAILED
```

`codecov/patch` reports 0% because the espnet2 coverage upload had not arrived when the
report was generated. Its own comment shows `test_python_espnet2` at `6 -> 0` uploads and
`test_integration_espnet2` at `66 -> 0`, which is also why the project figure appears to
drop by 50 points. Measured locally against the diff, none of the added executable lines
are uncovered:

```
$ pytest --cov=espnet2.sds.end_to_end.mini_omni_e2e test/espnet2/sds/end_to_end/
added executable lines uncovered: 0
```

The `__init__` assignments are exercised by `test_mini_omni_e2e_init.py` from #6528, and
`forward` by the new tests in this PR.

sw005320 (approved):

sw005320:
This seems to be reproducible and comes from your change https://github.com/espnet/espnet/actions/runs/33166478198/job/98832991875?pr=6542
Can you take a look at it?

amirlankalm:
Yeah that's mine. This branch predates #6527 so the forward tests weren't on it, my local run never saw them. They build the model with `__new__` and `forward` now reads `self.temperature/top_k/top_p`, so the stub was missing them. Merged master in and set them on the stub plus `**kwargs` on the stub client, 26 pass.

sw005320:
Thanks!

PR #6547, opened 2026-08-28
Make kaldiio an optional dependency

### LINKED ISSUE #6529: About kaldiio license compatibility with ESPnet production use
The existing discussion in #5549 raised this question, but it was closed as stale without a clarification. I am opening a new issue because the concern remains relevant to current ESPnet versions.

The current [kaldiio license](https://github.com/nttcslab-sp/kaldiio/blob/master/LICENSE) appears to restrict use to internal evaluation and limit redistribution, while [ESPnet still lists kaldiio as a dependency](https://github.com/espnet/espnet/blob/master/pyproject.toml).

I am concerned that this creates uncertainty for publicly distributed software.

sw005320:
## What did you change?

Moved `kaldiio` out of the default install and behind a new `kaldi` extra.

- **`pyproject.toml`**: dropped `kaldiio>=2.18.0` from `[project].dependencies` and added a dedicated `kaldiio` extra. It is pulled in by `all` and by `test`, so CI still exercises every Kaldi code path. `tools/setup.py` gets the same split.
- **New helper `espnet2/utils/kaldiio_utils.import_kaldiio()`**: imports `kaldiio` on demand, and otherwise raises an `ImportError` that says how to install it.
- **Lazy call sites**: every module-level `import kaldiio` in `espnet2/`, `utils/` and `pyscripts/audio/format_wav_scp.py` now happens inside the Kaldi branch via that helper, so importing `espnet2` no longer requires `kaldiio`. This covers `ESPnetDataset`'s `kaldi_loader`, `IterableESPnetDataset`'s `load_kaldi`, the legacy `KaldiReader`/`KaldiWriter`/`CMVN`, the feature-extraction CLIs, and the SpeechLM `KaldiAudioReader` (which keeps failing fast in `__init__`, now with the shared message).
- **`doc/installation.md`**: documented the new extra.

Kaldi-format-only recipe scripts under `egs2/` (`dump_codec.py`, `extract_spk_embed.py`, `mfa_format.py`, `compute_vad.py`, …) keep importing `kaldiio` directly, since they cannot do anything without it.

---

Closes #6529. The `kaldiio` license restricts use and redistribution, which creates uncertainty for anyone publicly distributing software that depends on ESPnet. `kaldiio` is only needed to read/write Kaldi `ark`/`scp` files, so it does not need to be a hard dependency of `pip install espnet`.

After this change, `pip install espnet` and the default `raw`-feature recipe path work without `kaldiio`; users who need Kaldi-format features run `pip install "espnet[kaldiio]"`.

---

## Is your PR small enough?

Yes — 18 files, +126/−35.

---

## Additional Context

`flake8` and `isort` pass on all touched files. `black --check` flags `espnet2/speechlm/dataloader/multimodal_loader/audio_loader.py`, but that is pre-existing on `master` (local black version skew) and unrelated to this diff.

One judgment call worth a reviewer's opinion: `tools/Makefile` runs `pip install -e ".."` with no extras, so the source-install recipe environment no longer picks up `kaldiio`. I left it opt-in and documented rather than changing it to `".[kaldiio]"`, but that one-line change would keep the Kaldi-specific recipe stages working out of the box if we prefer that.

### DIFF
--- a/doc/installation.md
+++ b/doc/installation.md
@@ -227,6 +227,7 @@ We also have [prebuilt Kaldi binaries](https://github.com/espnet/espnet/blob/mas
         | `st`       | Speech Translation            |
         | `s2t`      | Speech to Text (e.g., OWSM)   |
         | `spk`      | Speaker recognition           |
+        | `kaldi`    | Kaldi ark/scp I/O (`kaldiio`) |
         | `dev`      | Code formatting and linting   |
         | `test`     | Unit test dependencies        |
         | `doc`      | Documentation generation      |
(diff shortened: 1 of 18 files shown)


Masao-Someki (approved):
LGTM!

PR #6548, opened 2026-08-28
Fix the broken CommonVoice downloads (covost2, cvss, commonvoice, open_li52, open_li110, owsm_v3)

### LINKED ISSUE #5675: Can not download covost2, could you please update it with new link?
**Describe your question**
The link for downloading the covost2 is not working.
local/download_and_untar_commonvoice.sh: error executing wget https://voice-prod-bundler-ee1969a6ce8178826482b88e843c335139bd3fb4.s3.amazonaws.com/cv-corpus-4-2019-12-10/es.tar.gz
 
It looks like the link above is not existing anymore. Could you please update it with the new link? thanks

sw005320:
## What did you change?

Replaced the hard-coded CommonVoice download URLs in every recipe that fetches CommonVoice with a shared helper, `egs2/commonvoice/asr1/local/download_commonvoice.sh` (symlinked into the other recipes), which

- does nothing if the corpus is already unpacked in the target directory,
- downloads a signed link passed as `--cv_data_url` (single-language recipes),
- unpacks an archive the user placed in or next to the target directory by hand,
- and otherwise exits with instructions on how to obtain the data (the CommonVoice website, or the gated HuggingFace mirror `mozilla-foundation/common_voice_<version>`).

It also normalizes the `cv-corpus-<version>/<lang>/` layout of the current archives to what each recipe's data preparation expects, so nothing downstream had to change.

Recipes touched: `covost2` (asr1/st1/mt1), `cvss` (asr1/st1/s2st1), `commonvoice`, `open_li52`, `open_li110`, `owsm_v3`.

Bugs found in the touched download stages and fixed here:

- `download_and_untar.sh` compared the archive size against a never-set `$filesize`, which deleted an already downloaded archive, and derived the output filename from the URL, which breaks for signed URLs carrying a query string (now `wget -O`).
- `cvss/st1` fetched the CVSS archive from a hard-coded `cvss_t_v1.0/` path, which 404s for the default `version=c`, and unpacked it outside the directory stage 2 reads.
- `open_li110` called `local/voxforge/download_and_untar.sh`, which does not exist (the script is `local/commonvoice/download_and_untar.sh`).
- `owsm_v3` looped over the literal string `langs`, expanded `${lang}` in the URL before the loop, and called a `local/download_and_untar.sh` that does not exist in that recipe.
- `open_li52` removed `${COMMONVOICE}/<lang>.tar.gz` after unpacking, which would now delete an archive the user had downloaded by hand.

---

Fixes #5675. The CommonVoice download URLs in these recipes are dead: both the old `voice-prod-bundler-*.s3.amazonaws.com` links and the current `mozilla-common-voice-datasets` bucket return **403** (verified for cv-corpus-4, 5.1 and 9.0). Mozilla only hands out signed, time-limited links after the license is accepted on https://commonvoice.mozilla.org/en/datasets, so there is no replacement URL to swap in — the recipes have to let the user supply the data.

An already downloaded corpus is detected and left alone, so existing setups keep working. The CoVoST2 translation TSVs on `dl.fbaipublicfiles.com` are still available and are downloaded as before.

---

## Is your PR small enough?

23 files and ~290 lines, so slightly over the 20-file guideline — 10 of the files are one-line symlinks to the shared helper. It can be split at the commit boundary if preferred: the first commit fixes `covost2` + `cvss` (the recipes in the issue, 16 files), the second applies the same helper to `commonvoice`, `open_li52`, `open_li110` and `owsm_v3`.

---

## Additional Context

Tested locally against a sandboxed recipe layout with synthetic archives (no CI coverage for these data stages): missing data prints the instructions and exits 1 without leaving stray directories; a manually placed archive is unpacked and flattened; a re-run is a no-op; a signed URL with an `?X-Amz-...` query string downloads correctly; a legacy flat `<lang>.tar.gz` still works; a multi-language loop with one language missing continues past the failure; and an existing `${COMMONVOICE}/cv-corpus-5.1-2020-06-22/<lang>` install is skipped rather than re-downloaded.

Not addressed here: `egs2/cvss/asr1/local/data.sh` calls `local/cv_data_prep.pl`, which does not exist in that recipe (pre-existing, unrelated to the download).

### DIFF
--- a/egs2/commonvoice/asr1/local/data.sh
+++ b/egs2/commonvoice/asr1/local/data.sh
@@ -12,12 +12,13 @@ stage=0       # start from 0 if you need to start from data preparation
 stop_stage=100
 SECONDS=0
 lang=en # en de fr cy tt kab ca zh-TW it fa eu es ru tr nl eo zh-CN rw pt zh-HK cs pl uk
+# CommonVoice is not distributed through a public URL anymore. Set this to the
+# (time-limited, signed) link obtained after accepting the terms on
+# https://commonvoice.mozilla.org/en/datasets to download it automatically,
+# or see the message printed by local/download_commonvoice.sh.
+cv_data_url=
 
  . utils/parse_options.sh || exit 1;
-
-# base url for downloads.
-# Deprecated url:https://voice-prod-bundler-ee1969a6ce8178826482b88e843c335139bd3fb4.s3.amazonaws.com/cv-corpus-3/$lang.tar.gz
-data_url=https://voice-prod-bundler-ee1969a6ce8178826482b88e843c335139bd3fb4.s3.amazonaws.com/cv-corpus-5.1-2020-06-22/${lang}.tar.gz
 
 log() {
     local fname=${BASH_SOURCE[1]##*/}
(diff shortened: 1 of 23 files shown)


Masao-Someki (approved):
Thank you, LGTM!

PR #6555, opened 2026-08-29
Build the `dialog_eval` scoring models once instead of per utterance

amirlankalm:
## What did you change?

Every helper in `pyscripts/utils/dialog_eval` calls its `*_setup` functions inside the
function that scores a **single utterance**, so the models are rebuilt on every call.

- `ASR_WER.py` and `TTS_intelligibility.py` each construct **three ASR systems**
  (ESPnet, OWSM, Whisper) per call
- `TTS_speech_quality.py` builds the three MOS predictors plus SHEET-SSQA per call

The two WER helpers also request the same three ASR systems as each other, so the demo
holds two independent copies of them.

Every argument passed to those setups is a constant (`model_tag="default"`,
`beam_size=1`, `text_cleaner="whisper_en"`, `use_gpu=True`), so the results can be built
once and shared. `model_cache.py` holds one `lru_cache`d accessor per model and the
eight setup call sites now go through it.

The `versa` import stays inside the cached functions, so a missing versa still raises at
the point of use with the existing message rather than at import time.

Behaviour is unchanged: same models, same arguments, same returned strings. This only
removes repeated construction from the per-utterance path.

Verified with a stub `versa` that counts calls, since the helpers hardcode
`use_gpu=True` and therefore cannot run on a CPU-only machine. Over five scoring calls:

```
before: each of the five setups ran 5 times   (25 model constructions)
after:  each ran once (5), metrics still ran 5 times, output byte-identical
```

## Is your PR small enough?

Yes. 4 files, +107/-59.

## Additional context

No test is included because `test/` currently has no home for `egs2` pyscripts, the tree
only covers `espnet2`, `espnet3` and `egs3`. Happy to add one if you would like a
location for it.

Two things noticed while doing this and deliberately **not** changed, since they are
separate concerns:

- `use_gpu=True` is hardcoded in all of these helpers, so they cannot run on CPU at all.
- The helpers return formatted strings rather than numbers, so a programmatic consumer
  has to re-parse them.

Environment:

```
- python version: 3.12.8
```


sw005320 (approved):
LGTM, thanks!

PR #6557, opened 2026-08-29
Run the espnet2 unit tests under pytest-xdist

sw005320:
## Why

`test python` is the longest step of `test_python_espnet2`, averaging **21.6 min** across the six matrix cells of a master run (measured on run 33166270901), and it runs single threaded on a 4 vCPU runner.

Unlike splitting a job across more matrix cells, this reduces **both** wall clock and total compute — it does not add jobs competing for the organization-wide 60-concurrent-job limit.

## Change

Run the suite under `pytest-xdist` with one worker per vCPU. Three things are needed to make that *safe* rather than merely parallel:

**1. `--dist loadfile`, not the default `--dist load`**

Keeps every test in a file on the same worker. The default scatters individual tests across workers, which repeats and races module-level fixtures and module-level model construction. espnet has plenty of both.

**2. `OMP_NUM_THREADS=1` / `MKL_NUM_THREADS=1`**

torch defaults to one thread per core. Four workers × four threads on a four vCPU runner is sixteen threads on four cores — that oversubscribes the box and makes the suite *slower*, which is the usual reason `-n auto` disappoints.

**3. The s3prl checkpoint cache is warmed serially before the workers fork**

`test/espnet2/layers/test_create_adapter*.py` build an `S3prlFrontend` **at import time**, so each worker would otherwise download `hubert_base_ls960.pt` to the same path simultaneously.

The warm-up is non-fatal, and files under 1 MB are deleted afterwards, so a truncated download is not what `actions/cache` stores for the next run. That last part matters: the HTML error page huggingface.co returns under load is ~3 kB, and it is what broke collection in #6553's CI:

```
E   _pickle.UnpicklingError: Weights only load failed ...
E   Unsupported operand 60          # 0x3C == '<'
!!!!!!!!!!!!!!!!!!!! Interrupted: 1 error during collection !!!!!!!!!!!!!!!!!!!!
```

Without the size check, #6554's cache could have preserved that bad file indefinitely.

## Unchanged

`pytest-cov` combines worker data on its own, so `coverage report` / `coverage xml` need no changes, and `COVERAGE_CORE=sysmon` on python 3.12 still applies per worker.

`test/espnet3` is deliberately left single threaded — its `test python` step averages 2.4 min, which is not worth the extra surface.

## What still needs confirming

**The speedup has to come from CI on this PR.** The suite cannot be timed meaningfully outside a runner, and whether every test file is xdist-safe is exactly the sort of thing that only shows up when it runs. A realistic expectation is 2.5–3x on the pytest step rather than a clean 4x, given `--dist loadfile` granularity and uneven file sizes.

If individual files turn out not to be parallel-safe, the fix is `@pytest.mark.xdist_group` or serialising the affected file, not abandoning the approach.

### DIFF
--- a/ci/test_python_espnet2.sh
+++ b/ci/test_python_espnet2.sh
@@ -11,6 +11,11 @@ if python3 -c "import sys; exit(0 if sys.version_info >= (3,12) else 1)"; then
     export COVERAGE_CORE=sysmon
 fi
 
+# One BLAS/OMP thread per worker: 4 xdist workers each spawning 4 threads on a
+# 4 vCPU runner oversubscribes the box and makes the suite slower, not faster.
+export OMP_NUM_THREADS=1
+export MKL_NUM_THREADS=1
+
 exclude="egs2/TEMPLATE/asr1/utils,egs2/TEMPLATE/asr1/steps,egs2/TEMPLATE/tts1/sid,doc,tools,test_utils/bats-core,test_utils/bats-support,test_utils/bats-assert"
 
 # flake8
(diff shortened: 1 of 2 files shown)


sw005320:
Closing: CI showed the xdist change does not work for this suite.

**No speedup.** `512 failed, 56415 passed, 801 skipped in 1382.38s (23:02)` — the `test python` step took 23 min against a 21.6 min baseline for the whole step. These tests are dominated by torch operations that already saturate the runner's 4 vCPUs, so splitting them across 4 worker processes divides the same CPU and adds process overhead. Dropping the `OMP_NUM_THREADS=1` pin would only trade that for 16 threads on 4 cores.

**512 failures**, all of one shape, in `test_stft_encoder.py`, `test_transformer_separator.py` and `test_frontend.py`:

```
RuntimeError: Function 'PowBackward0' returned nan values in its 0th output.
RuntimeError: Function 'SqrtBackward0' returned nan values in its 0th output.
```

Backward passes where an STFT magnitude reaching exactly zero makes `sqrt`/`log` produce NaN. Those tests are not seeded and depend on ambient global RNG state, so a different execution order reaches inputs that today's order happens to avoid. xdist did not break them, it exposed them — worth knowing before any future change to how the suite is split.

The one independently useful piece — fetching the s3prl checkpoint before pytest and deleting truncated downloads so #6554's cache cannot store an HTML error page — is split out into #6562.

PR #6560, opened 2026-08-29
Drop install targets and config that nothing uses

sw005320:
Applies the same test that found festival in #6551: **installed by CI, referenced nowhere outside the install scripts themselves.**

## Dead make targets

| target | every reference in the repo | install time |
|---|---|---|
| `mwerSegmenter` | `ci/install.sh`, `ci/install_macos.sh` | 2 s |
| `nkf` | `ci/install.sh`, `ci/install_macos.sh` | 1 s |
| `py3mmseg` | `ci/install.sh`, `ci/install_macos.sh` | 4 s |
| `muskits` | the install scripts, plus prose in two HF README templates — no code path | 9 s |

`moses` and `sph2pipe` look similar but are **kept**: `moses` is used by `egs2/TEMPLATE/asr1/pyscripts/utils/simuleval_agent.py` and `espnet3/publication/demo/session.py`, and `sph2pipe` by the kaldi utils under `egs2/TEMPLATE` that the integration tests call.

## tensorboardx

Its own comment says what it was for:

```bash
# NOTE(kamo): tensorboardx is used for chainer mode only
python3 -m pip install tensorboardx
```

Nothing imports it, and the real dependency in `pyproject.toml` is `tensorboard>=1.14`, a different package. The `matplotlib` cache warm-up that shared its timer block is unrelated and is kept.

## chainer

`ci/install.sh` never read `CHAINER_VERSION` — zero occurrences — and nothing imports chainer. Yet the repository still carried:

- a `chainer-version: [6.0.0]` matrix dimension, with one value, in eight jobs
- `chainer-version` as a **required** input of `prepare-environment`
- `CHAINER_VERSION` exported by four workflows
- `python3 -m pip uninstall -y chainer` in two CI scripts, against a package nothing installs

All removed.

## The doc extra

`ci/install.sh` installed `.[test,doc,all]`, so sphinx and its five companions went into every one of the ~100 CI jobs. Only `publish_doc` needs them, and it runs `ci/doc.sh`, which now installs `.[doc]` itself.

## Honest accounting

This is worth roughly **20 s per job — about 35 runner-minutes of the current 1851**. Small. There is no second festival: after #6551 the install profile is flat, the largest single item being pyopenjtalk at 52 s.

The reason to do it now is that these are about to be baked into a prebuilt CI image, and dead configuration is harder to reason about later than it is to delete today.

## One thing to note when reviewing

Dropping the `chainer-version` matrix dimension **changes job names**:

```
test_python_espnet2 (3.10, 2.9.1, false, 6.0.0)  ->  (3.10, 2.9.1, false)
```

The required status checks introduced in #6553 are the `ci_gate_*` jobs, whose names do not depend on the matrix, so this is safe. It would not have been before #6553.

### DIFF
--- a/.github/actions/prepare-environment/action.yml
+++ b/.github/actions/prepare-environment/action.yml
@@ -11,9 +11,6 @@ inputs:
   pytorch-version:
     required: true
     type: string
-  chainer-version:
-    required: true
-    type: string
   use-conda:
     required: true
     type: string
@@ -37,7 +34,6 @@ runs:
       env:
         ESPNET_PYTHON_VERSION: ${{ inputs.python-version }}
         TH_VERSION: ${{ inputs.pytorch-version }}
-        CHAINER_VERSION: ${{ inputs.chainer-version }}
         USE_CONDA: ${{ inputs.use-conda }}
         HF_TOKEN: ${{ inputs.hf-token }}
       run: |
--- a/.github/workflows/ci_on_macos.yml
+++ b/.github/workflows/ci_on_macos.yml
@@ -49,7 +49,6 @@ jobs:
         env:
           ESPNET_PYTHON_VERSION: ${{ matrix.python-version }}
           TH_VERSION: ${{ matrix.pytorch-version }}
-          CHAINER_VERSION: 6.0.0
           USE_CONDA: ${{ matrix.use-conda }}
           # FIXME(kamo): clang is used by default, but I don't know how to use "-fopenmp" with clang
           WITH_OMP: OFF
--- a/.github/workflows/ci_on_ubuntu.yml
+++ b/.github/workflows/ci_on_ubuntu.yml
@@ -61,7 +61,6 @@ jobs:
   #     matrix:
   #       python-version: ["3.10", "3.12"]
   #       pytorch-version: [2.9.1, 2.10.0, 2.11.0]
-  #       chainer-version: [6.0.0]
   #       use-conda: [false]
   #   steps:
   #     - uses: actions/checkout@master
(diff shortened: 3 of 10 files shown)


sw005320:
Pushed a fix. Three of the four removals were wrong.

CI failed in all six `test_python_espnet2` cells and on debian12:

```
ImportError while importing test module 'test/espnet2/fileio/test_score_scp.py'
E   ModuleNotFoundError: No module named 'miditoolkit'
```

`miditoolkit` is installed by `tools/installers/install_muskits.sh`, and `espnet2/fileio/score_scp.py` imports it.

The test I used to decide these were dead grepped for each tool's **own name**. But recipes and library code reference what a tool *provides*, not the name of the script that installs it. Checking what each installer actually puts on the system gives a different answer:

| tool | provides | real user |
|---|---|---|
| muskits | `miditoolkit` | `espnet2/fileio/score_scp.py` |
| py3mmseg | `mmseg` | `egs2/hkust/asr1/local/hkust_segment.py` |
| nkf | `nkf` binary | `egs2/csj/asr1/local/*` |

All three restored. Only `mwerSegmenter` survives the corrected test — it appears nowhere but `.gitignore`, `.pre-commit-config.yaml`, `tools/Makefile`, `tools/extra_path.sh` and the install scripts.

Worth noting the same check was applied to festival in #6551, where it holds up: festival provides the `festival` binary and the phonemizer festival backend, and `phoneme_tokenizer.py` uses `backend="espeak"` exclusively.

What remains here is `mwerSegmenter`, `tensorboardx` (nothing imports it; the real dependency is `tensorboard`), the chainer configuration, and moving the `doc` extra to `ci/doc.sh` — about 10 s per job rather than the 20 s I claimed. The case for it is tidiness ahead of building a prebuilt CI image, not speed. Happy to close it instead if that is not worth a review.

PR #6561, opened 2026-08-29
Stop one flaky matrix cell from cancelling the rest, and drop an unused conda channel

sw005320:
Three rerun cycles in a row have gone to flakes with the same shape: a macOS conda flake on #6551, an s3prl download flake on #6553, and the macOS conda flake again on #6560.

## 1. `fail-fast: false`

The matrices use the default `fail-fast: true`, so one failing cell cancels its siblings. On #6560 that turned a single flake into what looked like a broken change:

```
failure    check_installable_on_macos (3.10, 2.9.1, true)    <- the actual flake
cancelled  check_installable_on_macos (3.10, 2.9.1, false)   <- collateral
failure    ci_gate_macos
```

The `use-conda=false` cell was perfectly healthy and does not even use conda. `gh pr checks` renders cancelled as `fail`, so this reads as two failures.

Set on the matrices in `ci_on_ubuntu`, `ci_on_macos`, `ci_on_windows`.

**This is a trade-off, not a free win.** A genuinely broken change now runs all 72 integration cells instead of stopping at the first failure. That seems the right side to err on: a full CI cycle currently takes between 145 min and 3.5 h, so discovering failures one at a time costs far more wall clock than the extra cells cost in runner minutes. Happy to restrict it to the small matrices if reviewers would rather keep the early exit on the 72-cell one.

## 2. The Intel conda channel is unused on the platform it keeps breaking

`tools/setup_miniforge.sh` prepends the channel unconditionally:

```bash
conda config --prepend channels https://software.repos.intel.com/python/conda/
```

But the only thing installed from it is twelve lines further down:

```bash
if [ "${unamem}" = "x86_64" ]; then
    conda install -y mkl=2024.0
fi
```

The macOS runners are **osx-arm64**. They never install anything from that channel — and yet when DNS for `software.repos.intel.com` fails on a runner, the whole job dies:

```
CondaHTTPError: HTTP 000 CONNECTION FAILED for url
<https://software.repos.intel.com/python/conda/noarch/repodata.json>
NameResolutionError: Failed to resolve 'software.repos.intel.com'
```

Gating the channel on the same `x86_64` condition as the install that needs it leaves x86_64 behaviour byte-for-byte unchanged, and removes a single point of failure that osx-arm64 was getting nothing from.

The host resolves fine from outside CI, so these are transient runner-side DNS failures rather than the channel having gone away — which is exactly why they are worth engineering around instead of waiting out.

### DIFF
--- a/.github/workflows/ci_on_macos.yml
+++ b/.github/workflows/ci_on_macos.yml
@@ -27,6 +27,7 @@ jobs:
     runs-on: macOS-latest
     if: github.event.pull_request.draft == false
     strategy:
+      fail-fast: false
       matrix:
         python-version: ["3.10"]
         pytorch-version: [2.9.1]
--- a/.github/workflows/ci_on_ubuntu.yml
+++ b/.github/workflows/ci_on_ubuntu.yml
@@ -91,6 +91,7 @@ jobs:
     if: |
       github.event.pull_request.draft == false
     strategy:
+      fail-fast: false
       max-parallel: 20
       matrix:
         python-version: ["3.10", "3.12"]
@@ -140,6 +141,7 @@ jobs:
     if: |
       github.event.pull_request.draft == false
     strategy:
+      fail-fast: false
       max-parallel: 20
       matrix:
         python-version: ["3.10", "3.12"]
@@ -178,6 +180,7 @@ jobs:
     if: |
       github.event.pull_request.draft == false
     strategy:
+      fail-fast: false
       max-parallel: 20
       matrix:
         python-version: ["3.10"]
@@ -209,6 +212,7 @@ jobs:
     if: |
       github.event.pull_request.draft == false
     strategy:
+      fail-fast: false
       max-parallel: 20
       matrix:
         python-version: ["3.10", "3.12"]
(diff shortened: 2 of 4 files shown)


sw005320:
A concrete data point for the `fail-fast: false` half of this PR, from #6559 a few minutes ago.

One job failed, on an apt flake that has nothing to do with that PR — `packages.microsoft.com` returning 403 for a repository preinstalled in the runner image, which espnet installs nothing from:

```
E: Failed to fetch https://packages.microsoft.com/repos/azure-cli/dists/noble/InRelease  403  Forbidden
E: The repository '...azure-cli noble InRelease' is no longer signed.
##[error]Process completed with exit code 100
```

`gh run rerun --failed` then re-queued **27 jobs**, not one:

```
queued:27  success:73
```

26 healthy jobs had been cancelled by the default `fail-fast`, so their work was thrown away and has to be redone. With `fail-fast: false` the rerun would have been a single job.

The PR description frames the trade-off as 'a broken change now runs all 72 integration cells'. That cost is real, but it only applies to genuinely broken changes. This cost applies every time any external dependency hiccups, which in the last day has been the Intel conda channel twice, a huggingface.co rate limit once, and now apt.

PR #6563, opened 2026-08-30
Add a prebuilt CI image, build-and-measure only

sw005320:
## Why

`Make environment` runs **8.0 min in each of ~101 jobs** per run — 13.5 runner-hours, **44% of the 1851 runner-minutes** a pull request now costs. It is the largest remaining item by a wide margin, and every job rebuilds exactly the same thing.

A prebuilt image replaces that with a pull.

## What this PR does — and deliberately does not do

It adds the image and the workflow that builds it. **No CI job is switched over.**

The reason to land the measurement first: **the size of the image decides whether the approach is worth pursuing at all.** GitHub runners keep no Docker layers between jobs, so every job pays a full pull.

| if the image is… | pull costs | `Make environment` becomes | worth it? |
|---|---|---|---|
| 4–6 GB | 1–2 min | ~2.5 min | yes — about a third of total CI compute |
| 15 GB | 5+ min | ~6 min | marginal; a different approach is needed |

That number is currently a guess. Guessing it is not a good basis for converting 101 jobs.

The build runs **on pull requests without pushing**, so size and build time land in the job summary even from a fork, where `packages: write` is not available. Pushes to GHCR happen only from master, the nightly schedule, or manual dispatch — never per pull request.

## What to look at

The job summary reports:

```
| tag               | py3.10-th2.9.1-<hash> |
| uncompressed size | ???? MiB              |
| build time        | ?m ??s                |
```

plus a `docker history` breakdown so we can see which layer dominates.

## Two details worth reviewing

**The Dockerfile is two stages.** Deleting the source in the same stage that copied it would leave it in the layer below and shrink nothing. The final stage takes only `tools/` — the part that took 8 minutes to build.

**The image deliberately contains a broken editable install**, pointing at a source tree the final stage does not have. Jobs will run `pip install -e . --no-deps` against their own checkout. That is what makes the tested code the pull request's code rather than whatever was baked into the image — and it is fast, because every dependency is already present.

## How a job would use it, later

```yaml
container:
  image: ghcr.io/espnet/espnet-ci:py3.10-th2.9.1-<hash>
steps:
  - uses: actions/checkout@v5
  - run: pip install -e . --no-deps
  - run: ./ci/test_python_espnet2.sh
```

The tag carries a hash of `ci/install.sh`, `docker/ci.dockerfile`, `pyproject.toml`, `tools/Makefile` and `tools/installers/**`. A job can compare that hash against its own checkout and **fall back to `ci/install.sh`** when a pull request changes any of those files. Fast path when the image matches, correct path always.

## Scope

One matrix cell (`3.10` / `2.9.1`). Widening to the full 2 python × 3 torch is worth doing once these numbers exist.

## Summary by CodeRabbit

* **New Features**
  * Added a prebuilt Docker-based CI environment with required system, Python, and PyTorch dependencies.
  * Added automated image builds triggered on demand, nightly, and by relevant changes.
  * CI images are tagged, size-checked, and published automatically for eligible builds.
  * Configured CI to use the prebuilt Python environment by default.

### DIFF
--- a/.github/workflows/build_ci_image.yml
+++ b/.github/workflows/build_ci_image.yml
@@ -1,0 +1,120 @@
+name: build ci image
+
+# Builds the prebuilt CI environment described in docker/ci.dockerfile.
+#
+# This is deliberately NOT run per pull request. It runs when the files that
+# determine the image's contents change on master, plus nightly to catch
+# upstream package drift, plus on demand. CI jobs only ever pull.
+#
+# Pull requests build the image without pushing it, which is what makes this
+# measurable from a fork: the size and build time are reported in the summary.
+
+on:
+  workflow_dispatch:
+  schedule:
+    # 03:00 UTC daily - upstream wheels move even when this repository does not.
+    - cron: "0 3 * * *"
+  push:
+    branches:
+      - master
+    paths:
+      - "ci/install.sh"
+      - "docker/ci.dockerfile"
+      - "pyproject.toml"
+      - "tools/Makefile"
+      - "tools/installers/**"
+      - ".github/workflows/build_ci_image.yml"
+  pull_request:
+    branches:
+      - master
+    paths:
+      - "ci/install.sh"
+      - "docker/ci.dockerfile"
+      - "pyproject.toml"
+      - "tools/Makefile"
+      - "tools/installers/**"
+      - ".github/workflows/build_ci_image.yml"
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.sha }}
+  cancel-in-progress: ${{ github.event_name == 'pull_request' }}
+
+permissions:
+  contents: read
+  packages: write
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        include:
+          # One cell for now. The full matrix is 2 python x 3 torch; widening it
+          # is only worth doing once the numbers from this one are known.
+          - python-version: "3.10"
+            pytorch-version: "2.9.1"
+    steps:
+      - uses: actions/checkout@v5
+
+      - name: Free disk space
+        run: |
+          echo "::group::before"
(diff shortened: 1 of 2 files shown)


sw005320:
Both findings were correct and are fixed in e121737.

**1. `ci/install.sh` failures were masked.** `A && B && C || true` groups as `(A && B && C) || true`, so the trailing `|| true` on the cleanup covered the install too. A failed install would have produced a *successful* build of an image with a partial environment — the worst possible outcome for something ~101 jobs are meant to depend on. Verified the grouping directly:

```
old:  f && (true || true) && true || true   -> exit 0   (failure swallowed)
new:  f && { true || true; } && { true || true; }  -> exit 1   (failure propagates)
```

Each cleanup now tolerates only its own failure.

**2. The venv was not on PATH.** `python:3.10-bookworm` sets `ENV PATH /usr/local/bin:$PATH` and installs its own interpreter there, so `pip` and `python` in the final stage resolved to the base image rather than the copied venv.

This one is worth spelling out because it would have broken the whole design: the `pip install -e . --no-deps` each job runs — the step that points the editable install at the job's own checkout — would have installed espnet into the *system* interpreter, leaving it missing from the venv the tests actually run under. Fixed with `ENV PATH="/espnet/tools/venv/bin:${PATH}"`.

Notably this would **not** have been caught by this PR's own CI, which only builds the image and never runs anything inside it. It would have surfaced as a confusing failure in the follow-up PR that switches a job over.

---

For context, the measurement this PR exists for came back on the previous run:

| | |
|---|---|
| uncompressed size | **5,307 MiB** |
| build time | **10m 57s** |

`docker history` breakdown: `COPY /espnet/tools` is 4.15 GB, the CI apt layer 409 MB, and the `python:3.10-bookworm` base the remaining ~890 MB. That is at the low end of the 4–6 GB range the design assumed, so the approach looks viable. Pull time from GHCR is the next thing to measure.

sw005320:
@coderabbitai review

PR #6566, opened 2026-08-30
Measure one job running in the prebuilt CI image

sw005320:
## The one number still missing

#6563 established that the image is **5307 MiB** and takes **11 min** to build. What decides whether converting ~101 jobs is worth it is the number that is still unknown: **how long a runner takes to pull it.**

GitHub runners keep no layers between jobs, so every job pays a full pull, against the **8.0 min `Make environment`** it replaces.

| pull time | job cost becomes | total CI compute |
|---|---|---|
| 1–2 min | ~2.5 min | **−31%** |
| 3–4 min | ~4.5 min | −19% |
| 5+ min | ~6 min | not worth the machinery |

## How this measures it

Rather than convert `test_python_espnet2` and find out the hard way, this adds a **second job** running the same tests as its `3.10 / 2.9.1` cell, inside the image, so the two appear side by side in the same run.

The number to read is the **`Initialize containers`** step, which is the pull.

It is deliberately **not** in `ci_gate_ubuntu`'s `needs` and is `continue-on-error: true`, so it cannot block a merge while it is being evaluated. Both jobs are meant to be deleted once the question is answered, whichever way it goes.

## Three things this tests beyond the timing

**Whether the tests behave the same at all.** The interpreter changes from an `actions/setup-python` build on Ubuntu 24.04 to `python:3.10-bookworm` on Debian 12. `ci_on_debian12` already passes, which makes this plausible rather than certain — and it is exactly the kind of thing that would be miserable to discover after converting every job.

**Whether the editable install really works.** `pip install -e . --no-deps` against the checkout is the step that makes the pull request's code the code under test rather than whatever was baked in. The step prints where `espnet` was imported from, and how long it took.

**Whether a fork can pull a private GHCR package.** The package is private, so the job passes `credentials`. A fork PR gets a read-only `GITHUB_TOKEN`; whether that is enough is untested. If it is not, the image has to be published publicly — which #6565 is the prerequisite for, since the image currently contains kaldiio and its licence forbids redistribution.

## Two implementation details

**The tag is computed, not hardcoded** — from the same file list, in the same order, as `build_ci_image.yml`. This branch changes none of those files, so it resolves to the image already published from master.

**`tools/` gets linked in.** A checkout has `tools/`'s tracked files but none of the build output — no venv, no `activate_python.sh`, no `sctk`. A step links in whatever the image has and the checkout does not, which leaves tracked files alone and needs no list to maintain.

```bash
for path in /espnet/tools/*; do
  name=$(basename "$path")
  [ -e "tools/${name}" ] || ln -s "$path" "tools/${name}"
done
```

## Merge order with #6565

`pip install -r ci/no_redistribute.txt` is tolerated failing, because that file arrives with #6565. Before it merges, the image still contains kaldiio and the line is a no-op; after, the file exists and the install is needed. Either order works.

## Summary by CodeRabbit

* **Tests**
  * Improved container test diagnostics by reporting Python and pip resolution, the active interpreter environment, and the checkout location.
  * Updated installation validation to verify the available ESPnet2 and ESPnet3 components.
  * Added reporting of the ESPnet2 module location to simplify test result analysis.
  * Existing Ubuntu CI checks remain unchanged.

### DIFF
--- a/.github/workflows/ci_on_ubuntu.yml
+++ b/.github/workflows/ci_on_ubuntu.yml
@@ -209,23 +209,18 @@ jobs:
           ls -la tools/ | head -20
 
       - name: Which interpreter
-        # The first attempt installed espnet successfully and then could not
-        # import it, which means pip and python did not agree on an
-        # interpreter. Container jobs also default to `sh`, not bash, and
-        # GitHub composes its own PATH for them, so neither can be assumed.
+        # Kept from the debugging round because it is nearly free and answers
+        # the first question anyone will ask if a container job misbehaves.
+        # It has already earned its place: it showed that PATH is fine and
+        # python, python3 and pip all resolve to the venv, which is what ruled
+        # out the interpreter mismatch this was written to look for.
         shell: bash
         run: |
           echo "PATH=$PATH"
           echo "which python  : $(command -v python || echo none)"
-          echo "which python3 : $(command -v python3 || echo none)"
           echo "which pip     : $(command -v pip || echo none)"
-          for exe in python python3 /espnet/tools/venv/bin/python; do
-            command -v "$exe" >/dev/null 2>&1 || continue
-            echo "--- $exe ---"
-            "$exe" -c 'import sys; print(" executable:", sys.executable); print(" prefix:", sys.prefix)'
-          done
-          echo "--- checkout ---"
-          pwd && ls -d espnet espnet2 2>/dev/null
+          python -c 'import sys; print("executable:", sys.executable); print("prefix:", sys.prefix)'
+          echo "checkout: $(pwd)"
 
       - name: Point the editable install at this checkout
         # The image's editable install refers to a source tree that is not in
@@ -246,7 +241,11 @@ jobs:
             echo "ci/no_redistribute.txt not on this branch yet (arrives with #6565)"
           fi
           echo "reinstall took $((SECONDS - start))s"
-          "${VENV}/python" -c 'import espnet, espnet2; print("espnet from", espnet.__file__)'
+          # The distribution is named espnet, but the importable packages are
+          # espnet2 and espnet3; there is no top-level espnet module. Asserting
+          # on the distribution name is what made the previous run look like a
+          # broken install when it was not.
+          "${VENV}/python" -c 'import espnet2, espnet3; print("espnet2 from", espnet2.__file__)'
 
       - name: test python
         run: ./ci/test_python_espnet2.sh
--- a/.github/workflows/ci_on_ubuntu.yml
+++ b/.github/workflows/ci_on_ubuntu.yml
@@ -152,6 +152,76 @@ jobs:
           path: ~/.cache/s3prl
           key: ${{ runner.os }}-s3prl-v1
 
+  # ---------------------------------------------------------------------------
+  # Experiment. Runs the same tests as the 3.10 / 2.9.1 cell of
+  # test_python_espnet2, but inside the prebuilt image from
+  # docker/ci.dockerfile, so the two can be compared directly.
+  #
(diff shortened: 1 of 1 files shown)


sw005320:
First run produced the number this PR exists for, plus one surprise.

## The number

```
Initialize containers   62s     <- the image pull
Run actions/checkout    3s
Link prebuilt tools     0s
```

**62 seconds**, against the **8.0 min** `Make environment` it replaces. That is at the good end of the range the design assumed, and better than the 2.5 min per job I had estimated — closer to ~1.2 min.

The other open question also answered itself: the **private GHCR package pulled fine from a fork pull request** using the read-only `GITHUB_TOKEN` and `credentials`. Publishing the image publicly is therefore not a prerequisite for this to work, which is a good thing given #6565.

## The surprise

pip installed espnet successfully — it even found and replaced the copy baked into the image:

```
Obtaining file:///__w/espnet/espnet
Successfully built espnet
  Attempting uninstall: espnet
    Found existing installation: espnet 202604
Successfully installed espnet-202604
```

and the very next command could not import it:

```
ModuleNotFoundError: No module named 'espnet'
```

That only happens when `pip` and `python` are not the same interpreter.

Two things make that possible in a container job and I accounted for neither:

- **The default shell is `sh`, not bash.** The log shows `shell: sh -e {0}`. That is also why `reinstall took 0s` — `$SECONDS` is not POSIX.
- **GitHub composes its own PATH for container steps** rather than simply inheriting the image's, so the `ENV PATH=/espnet/tools/venv/bin:...` added in #6563 is not something to rely on inside a job.

Pushed 9b7ff3b: call the venv's interpreter by absolute path via `python -m pip`, which cannot disagree with itself, and add a step that prints what `python`, `python3` and `pip` actually resolve to. If the diagnosis is right the install now works; if it is wrong, the next run says why rather than leaving it to inference.

This is exactly the class of thing the experiment was for. Converting all ~101 jobs first and hitting it afterwards would have been considerably worse.

sw005320:
Closing: the experiment answered its questions, and the numbers now live in #6569 where they inform an actual change. As the description said, both jobs were meant to be deleted once the question was settled.

## What it established

**The pull is cheap.** `Initialize containers` measured 62s, 80s and 81s across three runs, against the **510s** `Make environment` in the same run's normal job.

**The whole job is faster, by less than hoped.** Same run, same tests, same matrix cell:

| | container | normal | |
|---|---|---|---|
| environment | 81s | 510s | −7.2 min |
| editable install | 8s | — | +8s |
| `test python` | 1789s | 1602s | **+3.1 min** |
| **job total** | **1883s** | 2133s | **−4.2 min** |

The tests ran **11.7% slower** inside the container. Net is −4.2 min per job, not the −6.5 min assumed, so roughly **−23%** of total CI compute rather than −31%. Still the largest single lever, and one sample — the normal job's `test python` ranges from a 21.6 min average to 26.7 min in this run, so some of it may be noise. Converting the six cells will produce six comparisons per run, which beats another one-off measurement.

**A fork can pull the private package.** `credentials` with the read-only `GITHUB_TOKEN` worked, three times. Publishing the image publicly is not a prerequisite — which is fortunate, since the organisation blocks it and #6565 established the image could not have been published as it was.

**The editable install works.** `pip install -e . --no-deps` takes 8s and the tests run against the pull request's code.

## What it cost to learn, and was worth it

Two of my own errors surfaced here rather than after converting 101 jobs:

- Asserting on `import espnet`, when the repository has `espnet2/` and `espnet3/` and no `espnet/` at all. That made a working install look broken.
- Diagnosing it as a PATH problem in a container job. The diagnostic step disproved that outright: PATH is composed correctly and `python`, `python3` and `pip` all resolve to the venv.

Finding either of those with every job converted would have been considerably worse.

PR #6567, opened 2026-08-30
fix(state_spaces): DownSample.forward passes a stray argument, raising TypeError on every call

Anai-Guo:
## Problem

`DownSample.forward` passes five positional arguments to a four-parameter
function, so it raises `TypeError` on every call
(`espnet2/asr/state_spaces/pool.py:67`):

```python
def downsample(x, stride=1, expand=1, transposed=False):   # line 20
...
class DownSample(SequenceModule):
    def forward(self, x):
        return downsample(x, self.stride, self.expand, False, self.transposed)   # line 67
```

```
TypeError: downsample() takes from 1 to 4 positional arguments but 5 were given
```

The stray literal `False` sits in the `transposed` slot and pushes
`self.transposed` past the end of the signature.

## Sibling baseline

`UpSample` in the same file is the same class with the same four constructor
parameters and the same four stored attributes, and `upsample` has the identical
signature to `downsample`. Its `forward` gets it right:

| | forward |
|---|---|
| `UpSample` (line 177) | `upsample(x, self.stride, self.expand, self.transposed)` — 4 args |
| `DownSample` (line 67) | `downsample(x, self.stride, self.expand, False, self.transposed)` — 5 args |

`self.transposed` is only ever read in `forward`, so passing a hardcoded `False`
in its place would also make the constructor's `transposed` argument dead.
`UpSample` is the intended shape.

## Reachability

`DownSample` is a user-selectable pooling layer — it is the `"sample"` entry of
the pool registry at `espnet2/asr/state_spaces/pool.py:384`, so any config that
selects it fails on the first forward pass.

## Fix

```diff
-        return downsample(x, self.stride, self.expand, False, self.transposed)
+        return downsample(x, self.stride, self.expand, self.transposed)
```

## Verification

`downsample`, `upsample`, and both `forward` bodies were extracted from the real
file with `ast` and executed on CPU against a `(2, 8, 6)` tensor with
`stride=2, expand=1, transposed=True`:

```
UpSample.forward            -> OK, shape (2, 8, 12)
DownSample.forward          -> TypeError: downsample() takes from 1 to 4 positional arguments but 5 were given
DownSample(patched).forward -> OK, shape (2, 8, 3)
matches downsample(x, 2, 1, transposed=True): True
```

The patched output is identical to calling `downsample` directly with
`transposed=self.transposed`, and `UpSample` is untouched.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed an issue with downsampling during speech recognition processing, improving compatibility with the current downsampling behavior.
  * Improved handling of transposed layouts, expansion settings, and empty inputs during sampling operations.

* **Tests**
  * Added coverage for downsampling and upsampling across supported layouts, strides, expansion settings, and registry-based sampling.


sw005320 (approved):

sw005320:
Thanks!
Could you add a unit test?

Anai-Guo:
Added `test/espnet2/asr/state_spaces/test_pool.py` (b3102c3).

It covers `DownSample` and `UpSample` over `stride`/`expand` on both `transposed` layouts, pins each `forward` to the module-level function it delegates to, and asserts that `transposed` actually reaches `downsample()` rather than being replaced by a literal (the `expand` axis is the observable difference).

Verified locally with torch 2.5.1 / einops 0.8.2:

- on the parent commit (fix not applied): 12 failed, 8 passed — every `DownSample` case fails with `TypeError: downsample() takes from 1 to 4 positional arguments but 5 were given`, and the 8 `UpSample` cases pass as the sibling baseline.
- with the fix: 20 passed.

`black` and `isort` (profile=black) report no changes; `flake8 --extend-ignore=D` is clean on both files.

sw005320:
Thanks a lot, @Anai-Guo!

PR #6571, opened 2026-08-30
Exclude zero padding from utterance_mvn variance

Kayvan-Zahiri:
`utterance_mvn` subtracts the utterance mean from the whole padded tensor, so the padded frames end up holding `-mean` instead of zero. With `norm_vars=True` those frames get squared into the variance sum, which inflates the standard deviation by an amount that depends on how much padding the batch happened to carry.

A 6-frame utterance padded to 10 frames comes out with a per-channel std of about 0.35 instead of 1.0, and normalizing the same utterance on its own gives different numbers.

The function already knows about this: the `norm_means=False` branch re-masks `y` before computing its variance, and `GlobalMVN.forward` re-masks after subtracting its mean. This restores the zero padding right after the mean subtraction in the `norm_means=True` branch.

Inputs with no padding are bit-identical to before. Two tests added to `test/espnet2/layers/test_utterance_mvn.py`; both fail without the change.

### DIFF
--- a/espnet2/layers/utterance_mvn.py
+++ b/espnet2/layers/utterance_mvn.py
@@ -72,6 +72,12 @@ def utterance_mvn(
 
     if norm_means:
         x -= mean
+        # mean is broadcast over the padded frames too, so restore the zero padding
+        # before it reaches the variance below or the caller.
+        if x.requires_grad:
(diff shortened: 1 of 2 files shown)


sw005320 (approved):
Thanks for your catch, @Kayvan-Zahiri!

PR #6574, opened 2026-08-31
Bake the integration jobs' dependencies into the image

sw005320:
Groundwork for converting `test_integration_espnet2` — 72 jobs, and the bulk of what a CI run now costs. Two things those jobs need that the image does not yet have.

## `libjpeg-dev`

Comparing what `.github/actions/install-system-dependencies` installs against the image's own apt list, that is the **only** difference:

```
action:  cmake libsndfile1-dev bc sox ffmpeg libjpeg-dev
image:   cmake libsndfile1-dev bc sox ffmpeg ...
```

With it, a job running in this image **does not need to call apt at all**.

That is worth more than the seconds it saves. The runner image ships an apt repository at `packages.microsoft.com` that espnet installs nothing from, and when it answers 403 the whole job dies in `apt-get update`:

```
E: Failed to fetch https://packages.microsoft.com/repos/azure-cli/dists/noble/InRelease  403  Forbidden
##[error]Process completed with exit code 100
```

That has already failed a run of this work once — and it would be **72 chances per run** once the integration jobs are converted.

## Kaldi

`ci/install.sh` removes `tools/kaldi` at the end of the build, so this runs `ci/install_kaldi.sh` in the final stage rather than inheriting it from the builder.

Per job that is a `git clone` plus a GitHub Releases download. Across 72 integration jobs it is 72 more chances of exactly the kind of network failure that has cost this work several reruns already — sctk timing out, huggingface rate limiting, apt 403ing.

The build asserts `src/featbin` ended up non-empty rather than looking for a particular binary, since the tarball's contents are not ours to assume.

## `ci/install_kaldi.sh` joins the tag hash

It now determines what is in the image, so it belongs in the hash and the rebuild triggers. Same hole #6569 closed for `ci/no_redistribute.txt`, and the rule is worth stating plainly:

> Anything that changes the image's contents belongs in the hash, or the image silently stops matching what asked for it.

## Nothing uses this yet

The workflow still calls `install-system-dependencies` and `ci/install_kaldi.sh` per job, and keeps working unchanged. Removing those steps happens in the conversion PR, once the image carrying this is published.

That ordering is deliberate: the image has to exist before anything resolves to it.

### DIFF
--- a/.github/workflows/build_ci_image.yml
+++ b/.github/workflows/build_ci_image.yml
@@ -19,6 +19,7 @@ on:
       - master
     paths:
       - "ci/install.sh"
+      - "ci/install_kaldi.sh"
       - "ci/no_redistribute.txt"
       - "ci/report_unlicensed.py"
       - "docker/ci.dockerfile"
@@ -31,6 +32,7 @@ on:
       - master
     paths:
       - "ci/install.sh"
+      - "ci/install_kaldi.sh"
       - "ci/no_redistribute.txt"
       - "ci/report_unlicensed.py"
       - "docker/ci.dockerfile"
@@ -79,7 +81,7 @@ jobs:
       - name: Resolve image tag
         id: tag
         run: |
-          hash=${{ hashFiles('ci/install.sh', 'ci/no_redistribute.txt', 'docker/ci.dockerfile', 'pyproject.toml', 'tools/Makefile', 'tools/installers/**') }}
+          hash=${{ hashFiles('ci/install.sh', 'ci/install_kaldi.sh', 'ci/no_redistribute.txt', 'docker/ci.dockerfile', 'pyproject.toml', 'tools/Makefile', 'tools/installers/**') }}
           echo "tag=py${{ matrix.python-version }}-th${{ matrix.pytorch-version }}-${hash:0:12}" >> "$GITHUB_OUTPUT"
 
       - name: Build
--- a/.github/workflows/ci_on_ubuntu.yml
+++ b/.github/workflows/ci_on_ubuntu.yml
@@ -98,9 +98,13 @@ jobs:
       hash: ${{ steps.hash.outputs.hash }}
     steps:
       - uses: actions/checkout@v5
+      # The list below is duplicated in build_ci_image.yml and the two must
+      # match exactly, or the image is published under one tag and requested
+      # under another. Checked rather than trusted.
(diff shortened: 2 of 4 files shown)


sw005320:
Both findings looked at. The first was real and is fixed in f6e2838; the second I am not taking, with reasoning.

## 1. Hash input lists — real bug, fixed

Adding `ci/install_kaldi.sh` to `build_ci_image.yml` left the identical list in `ci_on_ubuntu.yml`'s `resolve_ci_image` untouched:

```
build_ci_image.yml   ci/install.sh, ci/install_kaldi.sh, ci/no_redistribute.txt, ...
ci_on_ubuntu.yml     ci/install.sh,                      ci/no_redistribute.txt, ...
```

Merged as it stood, the build would have published one tag while every container job asked for another, and **nothing would have been pullable** — every converted job, which is `test_python_espnet2` today and three more once #6573 lands.

Worth naming the mistake precisely: the commit being reviewed claimed computing the tag *"rather than hardcoding keeps the two from drifting apart"*, while putting the list in two files. Computing it does not help if the inputs are duplicated.

So beyond fixing the list, `ci/check_image_hash_inputs.py` now parses both workflows and fails if they differ. `resolve_ci_image` runs it **before** using the hash, so the failure lands on whoever changed one and not the other.

Verified against this exact bug — reverting the fix:

```
The image-tag hash inputs disagree.
  .github/workflows/build_ci_image.yml:
    ['ci/install.sh', 'ci/install_kaldi.sh', 'ci/no_redistribute.txt', ...]
  .github/workflows/ci_on_ubuntu.yml:
    ['ci/install.sh', 'ci/no_redistribute.txt', ...]
```

## 2. `packages: write` on pull request builds — not taking it

Sound in principle. Three things make it not worth the restructuring here:

- GitHub does not allow conditional `permissions:`, and there is **one job**. Scoping it properly means splitting build from publish — and shipping a 5 GB image between two jobs.
- The `Log in to GHCR` and `Push` steps are already gated on `github.event_name != 'pull_request'`, so the token is never *used* on a pull request.
- A fork pull request receives a **read-only** token regardless of what the `permissions:` block requests. The exposure is same-repository branches, whose authors already have write access to the package.

Happy to revisit if the build/publish split becomes worthwhile for other reasons.

sw005320:
@coderabbitai review

PR #6575, opened 2026-08-31
Fall back to building the environment when the image is not published

sw005320:
## The problem #6574 ran into

It changed `docker/ci.dockerfile` and `ci/install_kaldi.sh`, so it resolved to a tag that only exists **once it merges**. All six containerised cells died:

```
Error response from daemon: manifest unknown
##[error]Docker pull failed with exit code 1
```

It was merged on the strength of the image build passing and every other job being green. That was a reasonable call once. **It is not a process** — and this recurs by design: `pyproject.toml` is one of the hash inputs, so every dependency bump lands in the same hole.

## The fix

`resolve_ci_image` now asks the registry whether the tag is actually there:

```yaml
outputs:
  hash: ...
  available: ${{ steps.probe.outputs.available }}
```

and each job's container image becomes that answer:

```yaml
container:
  image: ${{ needs.resolve_ci_image.outputs.available == 'true'
             && format('ghcr.io/{0}-ci:py{1}-th{2}-{3}', ...) || '' }}
```

**An empty container value makes the job run on the runner instead.** The documentation does not say this — it says only that a job *without* a `container` runs on the host — so it was verified directly rather than assumed:

| probe job | `container:` | result |
|---|---|---|
| `empty_container` | `''` | `in_container=NO`, `Ubuntu 24.04.4 LTS` |
| `real_container` | `alpine:3.20` | `in_container=YES`, `Alpine Linux v3.20` |

The control job rules out "the expression simply wasn't evaluated".

The steps then branch on the same flag:

```yaml
- uses: ./.github/actions/use-prebuilt-environment
  if: needs.resolve_ci_image.outputs.available == 'true'
- name: Make environment
  if: needs.resolve_ci_image.outputs.available != 'true'
  uses: ./.github/actions/prepare-environment
```

This is the shape originally described when the image was introduced in #6563: **fast path when the image matches, correct path always.**

## Why all six variants are probed together

Those files decide every variant, so the tags appear and disappear as a set. Being conservative costs a slow run of a rare pull request; being wrong costs `manifest unknown` on every containerised job, which is the thing this exists to prevent.

The probe also emits a `::notice::` explaining why a run is slow, so the next person does not have to work it out.

## How this gets tested

**This PR exercises the fast path** — it touches none of the hash inputs, so the image exists and the jobs should pull it as usual.

**The slow path is exercised by the next PR that does** touch them. If this works, that PR simply runs slower instead of failing. That is the whole point.


sw005320:
@coderabbitai review

sw005320:
@coderabbitai full review

sw005320:
This turned out to matter more than the description says. It is not only a pull request problem — **master hits it too**, and did, an hour ago.

Merging #6574 started `ci on ubuntu` and `build ci image` at the same time. The CI run resolved to the new hash and went looking for an image the build had not produced yet:

```
ci on ubuntu     02:50:47Z   ->  Initialize containers: manifest unknown
build ci image   03:05:44Z   ->  03:29:28Z  (published, 39 min later)
```

All 15 containerised cells failed on master. Nothing was wrong with the change — the same cells pass on this branch now that the image exists, which is what the 123 green checks here are.

So the shape of the problem is:

| | without this PR | with it |
|---|---|---|
| PR changing image inputs | 6+ jobs fail on `manifest unknown` | falls back, runs slower, passes |
| **master merge of such a PR** | **every containerised job fails** | falls back, runs slower, passes |

The second row is the one worth fixing. A merge that turns master red for the length of an image build, every time someone bumps a dependency, is not something to handle by explaining it each time.

No revert needed for #6574: master's code is fine, the image is published (`py3.12-th2.10.0-d419f710fb90` and its five siblings), and the next master run will pull it normally.

PR #6585, opened 2026-08-31
Give the other three CI workflows read-only tokens too

sw005320:
#6583 did this for `ci_on_ubuntu.yml`. CodeRabbit reached the same conclusion for
`ci_on_windows.yml` independently on #6584, and macOS and debian12 are in the
same position: no `permissions:` block, so a job that checks out and runs pull
request code holds a token reporting `Contents: write`, `Actions: write`,
`PullRequests: write`, `Packages: write` and more.

Two permissions each, and only two:

| permission | used by |
|---|---|
| `contents: read` | `actions/checkout`, `actions/cache` |
| `pull-requests: read` | `joerick/pr-labels-action` |

**No `packages: read`** — unlike `ci_on_ubuntu.yml`, none of these pulls a GHCR
image, and `ci_on_debian12.yml`'s `debian:12` comes from Docker Hub, which does
not authenticate with `GITHUB_TOKEN`.

**No `id-token`** — it is not granted today, and none of the three uploads to
codecov.

## Verified

- all three parse; `permissions` resolves as intended, both jobs intact in each
- none of the three references `codecov` or `ghcr.io`
- merges cleanly with #6579, #6582, #6583 and #6584

## Not in here

The `publish_*.yml` workflows and `claude.yml` also have no `permissions:` block.
Those genuinely need write, so working out the minimum for each is a real review
rather than a mechanical one, and it should not ride along with three read-only
CI workflows.

### DIFF
--- a/.github/workflows/ci_on_debian12.yml
+++ b/.github/workflows/ci_on_debian12.yml
@@ -22,6 +22,22 @@ concurrency:
   group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.sha }}
   cancel-in-progress: ${{ github.event_name == 'pull_request' }}
 
+# The default on this repository is almost entirely write - a job's GITHUB_TOKEN
+# reports Contents: write, Actions: write, PullRequests: write, Packages: write
+# and more. This job checks out a pull request's code and runs it, next to
+# third-party actions, and writes nothing.
+#
+#   contents       checkout, and actions/cache
+#   pull-requests  joerick/pr-labels-action
+#
+# No packages: read here - unlike ci_on_ubuntu.yml these pull no GHCR image
+# (ci_on_debian12.yml's debian:12 comes from Docker Hub, which does not use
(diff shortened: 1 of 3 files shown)


sw005320:
@coderabbitai same SHA-pinning finding, sixth PR in a row — so let me stop
deferring it with a paragraph each time and give you a date instead.

**It goes in as its own PR as soon as #6586 merges.** The dependency is real
rather than an excuse: #6586 is deciding *which* checkout ref these workflows
use at all (`@master` turned out to be a branch frozen in July 2020, so nine
workflows have been running v2-era code). Pinning `actions/checkout` before that
lands would either conflict with #6586 or pin the stale ref.

And doing everything except checkout is not a coherent intermediate — that is the
partial state I have been arguing against, where a reader cannot tell audited
from unaudited. One PR, every reference, tags kept as trailing comments so later
bumps stay reviewable.

The `permissions:` change in this PR is orthogonal to that and does not depend
on it.

PR #6593, opened 2026-09-01
Pin every remaining action to a commit SHA

sw005320:
#6591 pinned the 12 steps that are handed a secret, on the argument that a
mutable ref there lets whoever moves the tag read the secret. This finishes the
job: the other **49 steps, across 12 more actions**.

The reason is weaker per step but not absent — a moved tag runs code nobody here
reviewed, with the workspace and whatever else the job holds, and these jobs check
out and execute pull request code.

| action | steps |
|---|---|
| `actions/checkout@v7` | 21 (12 in `ci_on_ubuntu.yml`) |
| `actions/cache` + `/restore` + `/save` | 12 |
| `actions/setup-python@v6,v7` | 7 |
| `joerick/pr-labels-action@v1.0.9` | 5 |
| `docker/setup-qemu-action`, `docker/setup-buildx-action` | 2 |
| `actions/delete-package-versions`, `actions/upload-artifact` | 2 |
| `openjournals/openjournals-draft-action@master` | 1 |

**The last one was not even a tag.** `openjournals/openjournals-draft-action@master`
is a branch — though unlike `actions/checkout@master`, which #6586 removed, this
one is the repository's default branch and current as of 2024, so it was stale in
form rather than in content.

Waited for #6581 because 19 of the checkout references were at v5 and 2 at v7;
pinning first would have meant redoing all of them.

Also pinned: the reference inside the commented-out `setup_environment` job. Same
reasoning as #6586 — uncommenting it later should not reintroduce an unpinned ref.

## Two holes in the checker, both fixed here

- **`_workflows()` globbed only `.github/workflows`**, so the composite action under
  `.github/actions` was never examined. `prepare-environment/action.yml` has an
  `actions/setup-python` step, which was therefore unpinned *and* unreported.
- The check only applied to steps receiving a secret. It now applies to all of them.

After this: **61 steps pinned, 0 unpinned.**

## The annotated-tag trap

Four of the actions resolved so far are *annotated* tags, where
`git/ref/tags/<tag>` returns the SHA of the tag object rather than a commit —
Actions rejects that. They have to be dereferenced through `git/tags/<sha>`:

| action | ref SHA | actual commit |
|---|---|---|
| `codecov/codecov-action@v7` | `a99c28d3` | `fb8b3582` |
| `anthropics/claude-code-action@v1` | `923714f1` | `833fb0f8` |
| `peaceiris/actions-gh-pages@v4` | `329bcc8f` | `84c30a85` |
| `actions/delete-package-versions@v5` | `25ad4af5` | `e5bc658c` |

Pasting the ref SHA would have broken those workflows in a way that *looks* like a
correct pin.

## Verified

Reverting a pin is reported, in both a workflow and a composite action:

```
ci_on_ubuntu.yml: resolve_ci_image: actions/checkout@v7 is not pinned to a commit SHA
action.yml: runs: actions/setup-python@v6 is not pinned to a commit SHA
```

`black`, `pycodestyle --config=setup.cfg` clean.

## Correction

On #6591 I said 57 references remained. **It is 49.** The 69 I derived that from
was a raw grep over `.github/` including the `.eol` archives; counted through the
YAML it is 61 total, 12 of which #6591 had already pinned.

Tags are kept as trailing comments so a later bump stays reviewable.

### DIFF
--- a/.github/actions/prepare-environment/action.yml
+++ b/.github/actions/prepare-environment/action.yml
@@ -23,7 +23,7 @@ inputs:
 runs:
   using: "composite"
   steps:
-    - uses: actions/setup-python@v6
+    - uses: actions/setup-python@ece7cb06caefa5fff74198d8649806c4678c61a1 # v6
       if: ${{ inputs.os-version == 'ubuntu' }}
       with:
         python-version: ${{ inputs.python-version }}
--- a/.github/workflows/build_ci_image.yml
+++ b/.github/workflows/build_ci_image.yml
@@ -67,7 +67,7 @@ jobs:
     outputs:
       matrix: ${{ steps.v.outputs.matrix }}
     steps:
-      - uses: actions/checkout@v7
+      - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7
       - run: python3 ci/check_ci_image_config.py
       - id: v
         run: echo "matrix=$(python3 ci/image_variants.py matrix)" >> "$GITHUB_OUTPUT"
@@ -83,7 +83,7 @@ jobs:
       # rebuilding the same environment in every job.
       matrix: ${{ fromJSON(needs.variants.outputs.matrix) }}
     steps:
-      - uses: actions/checkout@v7
+      - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7
 
       - name: Free disk space
         run: |
@@ -157,7 +157,7 @@ jobs:
         # to an older tag if it does not change the inputs.
         if: github.event_name != 'pull_request'
         continue-on-error: true
-        uses: actions/delete-package-versions@v5
+        uses: actions/delete-package-versions@e5bc658cc4c965c472efe991f8beea3981499c55 # v5
         with:
           owner: ${{ github.repository_owner }}
           package-name: espnet-ci
--- a/.github/workflows/build_ci_image.yml
+++ b/.github/workflows/build_ci_image.yml
@@ -56,7 +56,6 @@ concurrency:
 
 permissions:
   contents: read
-  packages: write
 
 jobs:
   # The grid lives in ci/image_variants.json so that this workflow, the probe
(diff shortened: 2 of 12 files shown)


sw005320:
All four taken. Each verified against the actual code before changing anything, and each verified again by breaking the thing it checks.

## 1. `persist-credentials` — 20 sites

Confirmed nothing depends on the persisted token first: `.github/` contains **no `git push`, `commit`, `tag` or submodule**, no run step uses the `gh` CLI, and `peaceiris/actions-gh-pages` takes its `github_token` as an explicit input rather than reading `.git/config`.

**`claude.yml` keeps the default and is exempt.** Its job holds `contents: write` precisely so the action can commit — turning credentials off there could break it, and it only runs when someone types `@claude`, so the breakage would surface at the worst moment. Your own list of sites excluded it too, which is the same conclusion.

That exemption is the rule the new check uses: `persist-credentials: false` unless the job may write contents. It is self-documenting rather than a hardcoded filename.

## 2. `packages: write` scoping

Correct — it was workflow-wide, so `variants` had it, and `variants` only runs `ci/image_variants.py`. Moved to the `build` job.

## 3. The message overclaimed

Right, and this matters more than its size suggests: the success line is the thing someone trusts when the check passes. Now `every third-party action is pinned to a SHA`.

## 4. `check_no_duplicate_keys` still read `_workflows()`

This is the sharpest of the four. `check_actions_pinned` reads `_action_files()` and swallows a parse failure with `continue` **on the assumption that the duplicate-key check will report it**. It would not have, for the composite actions — so a malformed `.github/actions/*/action.yml` would have been silently unvalidated by both. Fixed, and the `continue` now has a comment saying what it is relying on.

## Verified

| broken input | message |
|---|---|
| checkout without the option | `publish_doc.yml: generate_and_deploy_doc: checkout without persist-credentials: false` |
| composite action unpinned | `action.yml: runs: actions/setup-python@v6 is not pinned to a commit SHA` |
| duplicate key in a composite | `.github/actions/prepare-environment/action.yml: duplicate key 'a' at line 44` |

and `claude.yml` still passes, exempt on `contents: write`.

The persist-credentials note is written out once in `ci_on_ubuntu.yml` with a one-line reference in the other eight, rather than nine copies of the same paragraph.

PR #6597, opened 2026-09-01
Install k2 again, after four years of installing nothing

sw005320:
`k2.done` has been completing in **3 seconds** and installing nothing:

```
[WARNING] k2=1.10.dev20211112 for pip doesn't provide pytorch=2.9.1 binary.
          Skip k2-installation
```

So `import k2` always fails, and the two `use_k2` blocks in
`ci/test_integration_espnet2.sh` have **never executed**. That is why `use_k2`
appears in no CI log — something I noted on #6579 and put down to k2 not being
installed, rather than asking why.

## Two things were broken

Fixing either alone would not have helped.

1. The version was pinned to **`1.10.dev20211112`, from November 2021**, whose
   guard refuses any torch `>= 1.10.1`.
2. The index it passed to `-f`, `https://k2-fsa.org/nightly/`, now **returns 404**
   — so a version bump on its own would have resolved nothing.

## The version, resolved rather than asserted

`1.24.4.dev20260625` is the one release covering every pair
`ci/image_variants.json` builds. Checked with pip and the real platform tags
against `https://k2-fsa.github.io/k2/cpu.html`:

```
torch2.9.1   cp312  Successfully downloaded k2
torch2.9.1   cp313  Successfully downloaded k2
torch2.10.0  cp312  Successfully downloaded k2
torch2.10.0  cp313  Successfully downloaded k2
torch2.11.0  cp312  Successfully downloaded k2
torch2.11.0  cp313  Successfully downloaded k2
```

Wheels are `manylinux_2_27/2_28`; the GLIBC floor check stays.

**Where they come from is worth knowing:** the index is `k2-fsa.github.io`, but the
files live on Hugging Face under `csukuangfj2/k2` — the maintainer's own account.
That is k2's documented channel rather than a mirror chosen here, and it is a
supply-chain fact rather than a hidden one.

## conda cannot work

The `k2-fsa` conda channel stops at `1.24.3.dev20230508` with nothing for torch 2.9
or later. `conda_k2_version` is now empty on purpose and that branch skips with the
reason spelled out. **debian12 is the affected environment** — it sets
`USE_CONDA: true` — so k2 stays absent there and present everywhere else.

## What was deleted

The pytorch and CUDA ladders in the middle of this script enumerated what k2 1.10
offered in 2021 and were keyed on that exact version string, so bumping the version
turned all of them into dead code. The index is the authority on what exists; pip
consults it directly.

## The behaviour change that matters

**A missing wheel is now a failure, not a skip.** Adding a torch version k2 has not
caught up with will break the build and name the combination, instead of quietly
removing test coverage. The error message says what to do about it.

## Expect this to be red

The `use_k2` blocks will run for the first time in four years. If they fail, that
is code nobody has exercised since 2021 — not a regression from this change. I
would rather find out than keep the silence.

### DIFF
--- a/.github/actions/use-prebuilt-environment/action.yml
+++ b/.github/actions/use-prebuilt-environment/action.yml
@@ -38,4 +38,4 @@ runs:
         # importable, and once by landing in the system interpreter instead of the
         # venv while the log said it had succeeded. An install log is not
         # evidence; an import is.
-        "${VENV}/python" -c 'import espnet2, espnet3, kaldiio, k2; print("espnet2 from", espnet2.__file__); print("k2", k2.__version__)'
+        "${VENV}/python" -c 'import espnet2, espnet3, kaldiio, k2; print("espnet2 from", espnet2.__file__); print("k2 from", k2.__file__)'
--- a/.github/actions/use-prebuilt-environment/action.yml
+++ b/.github/actions/use-prebuilt-environment/action.yml
@@ -34,4 +34,8 @@ runs:
         # from PyPI, exactly as the non-container path does.
         "${VENV}/python" -m pip install -r ci/no_redistribute.txt
 
-        "${VENV}/python" -c 'import espnet2, espnet3, kaldiio; print("espnet2 from", espnet2.__file__)'
+        # k2 is here because it spent four years being "installed" without being
+        # importable, and once by landing in the system interpreter instead of the
+        # venv while the log said it had succeeded. An install log is not
+        # evidence; an import is.
+        "${VENV}/python" -c 'import espnet2, espnet3, kaldiio, k2; print("espnet2 from", espnet2.__file__); print("k2", k2.__version__)'
--- a/ci/test_integration_espnet2.sh
+++ b/ci/test_integration_espnet2.sh
@@ -160,15 +160,22 @@ if [ "${task}" == "asr_transducer" ] || [ "${task}" == "all" ]; then
         fi
     fi
 
-    # k2 is not installed in CI, so neither of these runs today - "use_k2" does
-    # not appear in any current log. Noting it because they are the one place the
-    # split leaves a latent cross-task dependency: they decode at stage 12 only,
-    # and the model they load (feats_normalize=utterance_mvn,
-    # extract_feats_in_collect_stats=false) is trained by the asr_misc task, not
-    # this one. Whoever installs k2 will have to train it here first.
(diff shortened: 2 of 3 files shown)


sw005320:
Pushed `36a73caf2`. **The first version of this PR reported success and installed
nothing usable** — worth writing down, because every symptom looked like good news.

All six image builds passed and the log said:

```
Successfully installed graphviz-0.21 k2-1.24.4.dev20260625+cpu.torch2.9.1 ...
```

One line gave it away:

```
Requirement already satisfied: filelock in /usr/local/lib/python3.13/site-packages
```

`/usr/local`, not `./venv`. `install_k2.sh` called bare `pip install`, and on
python 3.13 there is **no `venv/bin/pip`** — `ensurepip --upgrade` runs there
because its bundled pip is newer than the pinned 25.2 — so `pip` resolved to the
system interpreter on PATH. k2 went into the wrong python, and the image's final
stage, which copies only `/espnet/tools`, discarded it.

Same defect I fixed in `docker/ci.dockerfile` two commits ago. This script has
line 46 using `python3 -m pip` and line 156 using `pip`.

## Three symptoms, each of which read as reassuring

| observation | what it actually meant |
|---|---|
| Image size unchanged: 7476 MiB vs master's 7476 | not "k2 is small" — the 3 GB of CUDA wheels it dragged in were thrown away with the rest |
| The CPU torch survived | only because the replacement landed elsewhere |
| build green | `import k2` still failed, so `use_k2` still never ran |

## Fixed

`python3 -m pip`, and `--no-deps`. The wheel declares `torch==2.9.1`, and
resolving that pulls the default PyPI torch — the **CUDA** build — over the CPU
one this environment installs deliberately. I read the wheel metadata rather than
guessing what `--no-deps` would drop:

```
Requires-Dist: torch==2.9.1
Requires-Dist: graphviz
Requires-Dist: typing_extensions; python_version < "3.8"
```

`graphviz` is now installed explicitly.

## Two verifications added

An install log has now twice failed to mean what it said, so:

- `install_k2.sh` imports k2 and prints its version and file, failing with a hint
  to check `sys.executable`.
- `use-prebuilt-environment` does the same per job, next to the existing
  espnet2/espnet3/kaldiio check — so a broken image is caught where it is used,
  not inferred from a green build.

Restoring the k2 import guards in espnet2 is the obvious follow-up and is
deliberately not here: it belongs after CI demonstrates k2 is importable, not
alongside the change trying to make it so.

PR #6598, opened 2026-09-01
Batch beam search over utterances for ASR and S2T inference

sw005320:
## What did you change?

Added `UttBatchBeamSearch`, a beam search that batches over **utterances** as well as hypotheses, and wired it into `asr_inference.py` / `s2t_inference.py` behind `--batch_size > 1`, which previously raised `NotImplementedError("batch decoding is not implemented")`.

**`espnet2/legacy/nets/batch_beam_search_utt.py` (new)**

Keeps a fixed `(n_utt, beam)` grid of hypotheses laid out as `b * beam + k`, which is the layout `CTCPrefixScoreTH` already assumes (it recovers the utterance index as `flat_index // beam`).

- Finished hypotheses are **masked out** of the grid rather than removed, because `CTCPrefixScoreTH` requires a constant number of hypotheses per utterance. This is equivalent to what `BatchBeamSearch` does, since `search()` always returns exactly `beam_size` new hypotheses regardless of how many ended.
- `maxlen`, `minlen`, end detection and termination are tracked per utterance, so utterances of different lengths decode together.
- On the first step only beam 0 is active, reproducing the single-hypothesis start of `BeamSearch`; otherwise the first `topk` would return `beam` copies of the same token.
- The hypothesis bookkeeping (`yseq`, `score`, per-scorer `scores`) is vectorized instead of looping over hypotheses in Python.

**Supporting changes**

- `CTCPrefixScorer.batch_init_state` now accepts a padded batch plus its lengths. A new `batch_select_state` reorders a whole grid in one call through `CTCPrefixScoreTH.index_select_state`, replacing a per-hypothesis Python loop. `select_state` learned to unpack the resulting batched state.
- `BaseTransformerDecoder.batch_score` takes an optional `xs_mask`, passed to `forward_one_step` as `memory_mask`, so padded encoder frames are excluded from the cross attention. Without it, batching utterances of different lengths silently changes the results. Scorers that do not accept `xs_mask` (e.g. `TransformerMDDecoder`) log a warning once and behave as before.
- `Speech2Text.batch_decode()` in both bins; `inference()` writes one result per key instead of assuming `keys[0]`.
- ASR `batch_decode` sorts the batch by decreasing length (required by the RNN encoder's `pack_padded_sequence`) and restores the caller's order afterwards.
- **Bug fix:** `batch_size` was missing from `speech2text_kwargs` in both bins, so it never reached `Speech2Text` at all.

Batching is refused with an explicit message when any full scorer is not a `BatchScorerInterface`, and for the transducer, HuggingFace, streaming, time-sync, partially-AR, Enh+ASR and multi-speaker paths.

`utils/benchmark_batch_beam_search.py` is a synthetic benchmark for the numbers below. Happy to drop it or move it elsewhere if `utils/` is the wrong home.

---

`BatchBeamSearch` vectorizes the beam axis but decodes one utterance at a time. A beam-10 decode of a single utterance is far too small to fill a modern GPU, so most of the accelerator sits idle during inference.

Measured on MPS with a 6-layer 512-dim 2048-FF 8-head decoder, vocab 5000, beam 10, CTC weight 0.3, 16 utterances, 30 output tokens (a constant max output length, i.e. no early exit — the least favourable case for batching):

Encoder lengths 150–300 frames:

| | ms/utt | speedup |
|---|---:|---:|
| `BatchBeamSearch` (one utterance at a time) | 3240 | 1.00x |
| `UttBatchBeamSearch` batch=1 | 2274 | 1.42x |
| batch=2 | 1299 | 2.49x |
| batch=4 | 1070 | 3.03x |
| batch=8 | 892 | 3.63x |
| batch=16 | 752 | **4.31x** |

Uniform 300 frames reaches 4.81x at batch=8. The `batch=1` row is the vectorized bookkeeping alone, independent of batching. On CPU the gain is only ~1.5x, as expected — this is an accelerator-side optimization.

Memory grows roughly linearly with the batch size: with 1500-frame encoder outputs, beam 10 and dim 512, about 250–270 MB per unit of batch size (the replicated encoder output and the CTC `r` tensor).

---

## Is your PR small enough?

Not quite: 9 files, +1836 / -177. Of that, 719 lines are the new beam search module and 453 are new tests; the edits to existing non-test files are 387 lines, most of which is re-indentation in the two `inference()` loops. I can split the core module + scorer changes from the CLI wiring if you would rather review them separately, though the first half would then be unused until the second lands.

---

## Additional Context

**Testing**

- `test/espnet2/legacy/test_batch_beam_search_utt.py` (new): equivalence against `BatchBeamSearch` run one utterance at a time, over CTC weight {0, 0.5, 1.0} x LM {0, 0.5} x beam {1, 3} x `normalize_length` x float32/float64, plus uniform-length, `maxlenratio` and per-utterance-primer cases. Token sequences and scores match.
- Mutation check: disabling the decoder's `xs_mask` makes 30 of these tests fail, so the padding mask is genuinely covered.
- `test_asr_inference.py` / `test_s2t_inference.py`: `batch_decode` vs. per-utterance decoding, input-order restoration, and the rejection paths.
- End to end, `--batch_size 1` and `--batch_size 3` produce byte-identical output files on 5 utterances of different lengths with 2-best and CTC.

**Known limitations**

- `ScoreFilter` in `s2t_inference.py` still implements `batch_score` as a Python loop over hypotheses, so its cost grows from `beam` to `n_utt * beam` calls per step. S2T will see less of the speedup above until it is vectorized. I can do that in a follow-up.
- A done utterance stays in the grid until the whole batch finishes, so a long straggler wastes some compute. `CTCPrefixScoreTH` fixes its batch size at construction, so shrinking the grid mid-decode is not possible without changes there.
- Pre-existing, unrelated: `CTCPrefixScoreTH` raises `IndexError` when asked to score a prefix longer than the encoder output. S2T hypotheses start with a 4-token prompt, so S2T + CTC decoding can hit this today on `master` whenever `end_detect` does not stop the search first.

### DIFF
--- a/espnet2/asr/decoder/transformer_decoder.py
+++ b/espnet2/asr/decoder/transformer_decoder.py
@@ -265,6 +265,7 @@ class BaseTransformerDecoder(
         states: List[Any],
         xs: torch.Tensor,
         return_hs: bool = False,
+        xs_mask: torch.Tensor = None,
     ) -> Tuple[torch.Tensor, List[Any]]:
         """Score new token batch.
 
@@ -273,7 +274,11 @@ class BaseTransformerDecoder(
             states (List[Any]): Scorer states for prefix tokens.
             xs (torch.Tensor):
                 The encoder feature that generates ys (n_batch, xlen, n_feat).
-
+            return_hs (bool): Whether to return the decoder hidden states.
+            xs_mask (torch.Tensor): Non-padding mask of `xs` (n_batch, 1, xlen).
+                Needed when `xs` holds utterances of different lengths, as in
+                utterance-batched beam search; without it the cross attention
+                also attends to the padded encoder frames.
 
         Returns:
             tuple[torch.Tensor, List[Any]]: Tuple of
@@ -297,11 +302,21 @@ class BaseTransformerDecoder(
         ys_mask = subsequent_mask(ys.size(-1), device=xs.device).unsqueeze(0)
         if return_hs:
             (logp, hs), states = self.forward_one_step(
-                ys, ys_mask, xs, cache=batch_state, return_hs=return_hs
+                ys,
+                ys_mask,
+                xs,
+                memory_mask=xs_mask,
+                cache=batch_state,
+                return_hs=return_hs,
             )
         else:
             logp, states = self.forward_one_step(
-                ys, ys_mask, xs, cache=batch_state, return_hs=return_hs
+                ys,
+                ys_mask,
+                xs,
+                memory_mask=xs_mask,
+                cache=batch_state,
+                return_hs=return_hs,
             )
 
         # transpose state of [layer, batch] into [batch, layer]
--- a/espnet2/bin/asr_inference.py
+++ b/espnet2/bin/asr_inference.py
@@ -28,6 +28,7 @@ from espnet2.asr.transducer.beam_search_transducer import Hypothesis as TransHyp
 from espnet2.fileio.datadir_writer import DatadirWriter
 from espnet2.legacy.nets.batch_beam_search import BatchBeamSearch
 from espnet2.legacy.nets.batch_beam_search_online_sim import BatchBeamSearchOnlineSim
+from espnet2.legacy.nets.batch_beam_search_utt import UttBatchBeamSearch
 from espnet2.legacy.nets.beam_search import BeamSearch, Hypothesis
 from espnet2.legacy.nets.beam_search_timesync import BeamSearchTimeSync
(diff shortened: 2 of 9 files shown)


sw005320:
Superseded by #6599, which does the same thing without a second beam search class: `BatchBeamSearch` itself now takes either a `(T, D)` encoder output or a batched `(n_utt, T, D)` one, since decoding a single utterance is the `n_utt == 1` case of decoding a batch.

Closing this one so review can focus on #6599. The branch is kept for reference.

PR #6599, opened 2026-09-01
Decode a batch of utterances in one beam search

sw005320:
Supersedes #6598 (now closed), which added the same capability as a separate `UttBatchBeamSearch` class. Same tests and same measurements, folded into `BatchBeamSearch` itself.

## What did you change?

Decoding one utterance is the `n_utt == 1` case of decoding a batch of them, so `BatchBeamSearch` handles both instead of there being a second class.

`BatchBeamSearch.forward` takes either a `(T, D)` encoder output and returns an n-best list exactly as before, or a batched `(n_utt, T, D)` one plus `x_lengths` and returns one n-best list per utterance. `x_lengths` comes after `pre_x` in the signature, so every existing positional call keeps working. Beam pruning is one expression in both cases:

```python
weighted_scores.view(n_utt, -1).topk(self.beam_size, dim=-1)
```

Hypotheses of the same utterance are kept contiguous, `b * n_hyp_per_utt + k`, which is the layout `CTCPrefixScoreTH` already assumes — it recovers the utterance index of a hypothesis as `flat_index // n_hyp_per_utt`.

**Where the two cases genuinely differ.** `CTCPrefixScoreTH` requires the same number of hypotheses for every utterance in the batch, so once several utterances are decoded together a finished hypothesis has to be flagged in `active` and left in place rather than removed. With a single utterance that constraint does not apply, so the batch is compacted after every step exactly as it is on master today. That is not two algorithms, it is one optimization that only a single utterance can afford — and it is what keeps `BatchBeamSearchOnlineSim`, `BatchBeamSearchOnline` and `PartiallyARBeamSearch` working unchanged. All three drive `search` / `post_process` themselves and depend on the running set shrinking (`len(running_hyps) == 0` in `batch_beam_search_online_sim.py:220`, `self._batch_select(self.running_hyps, [0])` in `batch_beam_search_online.py:250`).

`BatchHypothesis` gained `n_utt`, `active` and `done`, all with defaults, so the subclasses that build one field by field are unaffected.

**Supporting changes**

- `CTCPrefixScorer.batch_init_state` accepts a padded batch plus its lengths. A new `batch_select_state` reorders a whole batch through `CTCPrefixScoreTH.index_select_state` instead of looping over hypotheses in Python; it is used whenever the hypothesis count is unchanged, which is always true for a batch of utterances and true for a single utterance except right after some hypotheses ended.
- `BaseTransformerDecoder.batch_score` takes an optional `xs_mask`, passed to `forward_one_step` as `memory_mask`, so padded encoder frames are excluded from the cross attention. Without it, batching utterances of different lengths silently changes the results. Scorers that do not accept it (e.g. `TransformerMDDecoder`) log a warning once and behave as before. `score_full` only passes the new arguments when they are set, so subclasses that override it with the old signature are unaffected.
- `Speech2Text.batch_decode()` in `asr_inference.py` / `s2t_inference.py`, `--batch_size > 1` no longer raises, and `inference()` writes one result per key instead of assuming `keys[0]`. ASR sorts the batch by decreasing length (required by the RNN encoder's `pack_padded_sequence`) and restores the caller's order afterwards.
- **Bug fix:** `batch_size` was missing from `speech2text_kwargs` in both bins, so it never reached `Speech2Text` at all.

Batching is refused with an explicit message when a full scorer is not a `BatchScorerInterface`, and for the transducer, HuggingFace, streaming, time-sync, partially-AR, Enh+ASR and multi-speaker paths.

---

A beam-10 decode of a single utterance is far too small to fill a modern GPU, so most of the accelerator sits idle during inference.

Measured with `utils/benchmark_batch_beam_search.py` on MPS: 6-layer 512-dim 2048-FF 8-head decoder, vocab 5000, beam 10, 16 utterances of 150–300 frames, 30 output tokens with a constant max output length (no early exit, the least favourable case for batching). Median of three runs each; the baseline is master run through the same script.

| | ms/utt | vs. master |
|---|---:|---:|
| master, one utterance at a time | 4329 | 1.00x |
| this PR, one utterance at a time | 2362 | 1.83x |
| this PR, batch=2 | 1573 | 2.75x |
| this PR, batch=4 | 796 | 5.44x |
| this PR, batch=8 | 593 | 7.30x |
| this PR, batch=16 | 565 | **7.66x** |

Two independent things are in that table, and the second row separates them. Decoding is faster even without batching, because profiling a batched decode on the GPU showed the model itself was 2% of the wall clock and nearly all the rest was Python walking device tensors one element at a time — every read a device-to-host synchronisation, and there are `n_utt * beam` of them per scorer per step. Fixing that (the three per-hypothesis loops in `CTCPrefixScoreTH.__call__`, the id loop in `_select_states`, the per-slot reads in `post_process`, and a `torch.tensor` rebuilt every step in `LengthBonus.batch_score`) is what the 1.83x row is. Batching then multiplies it.

**On a trained model, with real audio.** The table above isolates the beam search: random weights, synthetic encoder outputs, a fixed output length. To see what it is worth end to end I ran the LibriSpeech E-Branchformer from this recipe's README (`asapp/e_branchformer_librispeech`, 148.9M params) over real test-clean audio, beam 10, CTC 0.3, `maxlenratio=0.0` so end detection behaves normally:

| batch | RTF | vs. batch_size=1 | transcripts identical to batch_size=1 |
|---:|---:|---:|:---:|
| 1 | 0.556 | 1.00x | — |
| 2 | 0.449 | 1.24x | 8/8 |
| 4 | 0.333 | 1.67x | 8/8 |
| 8 | 0.303 | **1.84x** | 8/8 |

Two things worth saying about this.

The transcripts came out identical at every batch size, on this model and this data, despite the encoder length difference described under Known limitations. That is reassuring but it is not a guarantee — it is eight utterances on one model.

And 1.84x is a long way from 7x. The encoder is not the reason: it was 3-10% of the time, so the beam search really is what dominates. The gap is the straggler effect. These utterances run 2.1-25.1 s, so `maxlen` differs by an order of magnitude across a batch, and the batch keeps stepping until its longest utterance is done while the finished slots are still scored. On top of that the CTC prefix score costs O(T) per step with T the padded length of the batch, so one long utterance inflates every step for everyone. Bucketing utterances by length before batching is what would recover this, and it is not something this PR does.

Numbers on this machine vary by up to ~20% run to run; the master baseline was stable within 4%. On CPU the batching gain is only ~1.5x, as expected — filling an idle accelerator is the point.

What is left in the profile is the `for t in range(start, end)` recursion over encoder frames inside the CTC prefix scorer, which is sequential by construction.

Memory grows roughly linearly with the batch size: with 1500-frame encoder outputs, beam 10 and dim 512, about 250–270 MB per unit of batch size, from the replicated encoder output and the CTC `r` tensor.

---

## Is your PR small enough?

Not quite: 9 files, +1762 / -295. `batch_beam_search.py` accounts for most of it, and 452 of the added lines are tests. I can split the scorer/decoder changes from the beam search rewrite if that helps review, though the halves are not independently useful.

---

## Additional Context

**Testing**

- `test/espnet2/legacy/test_batch_beam_search_reference.py` (new) pins single-utterance decoding against `reference_batch_beam_search.py`, a byte-for-byte frozen copy of the implementation this PR replaces (`ec5632b16`). It asserts the same tokens, total score, per-scorer scores and n-best ordering with `rtol=atol=0`, over 221 configurations (CTC weight x LM type x six length-bound regimes x length normalization x float32/float64), plus `hyp_primer` and `return_hs`. Rewriting `BatchBeamSearch` in place removes the code it replaced, so without this the guarantee would only be transitive through the non-batch `BeamSearch`.
- That test bites: reverting the `<eos>`-appended-twice behaviour at maxlen fails 122 cases, moving the minlen boundary by one fails 36, changing the length-normalization divisor fails 28, disabling the pre-beam fails 45. Disabling the vectorized CTC state selection changes nothing, confirming `index_select_state` and the per-hypothesis path agree to the bit.
- `test/espnet2/legacy/test_ctc_prefix_score_reference.py` (new) pins the vectorized `CTCPrefixScoreTH` against a frozen copy of the loop-based one, at `rtol=atol=0`, over seven shapes with and without pre-beam pruning and with uneven utterance lengths. The beam search parity tests cannot cover that file: both sides import it, so a change there would move the reference too. Three mutations confirm it bites.
- `test/espnet2/legacy/test_batch_beam_search_memory.py` (new) guards the two ways beam search goes wrong without producing a wrong hypothesis. It measures the bytes the search holds — running plus finished hypotheses, counted by distinct storage so a view costs whatever it is a view of — and asserts the shape of the growth rather than an absolute number, so it needs no accelerator: doubling the output length must at most double the memory (measured 1.4–1.7x), doubling the utterances must at most double it (measured exactly 2.00x), a finished hypothesis must own its state outright, and every scorer must see all `n_utt * beam` hypotheses in a single call per step. All four bite under mutation.
- A mini_an4 asr1 integration case decodes with `--batch_size 2`, checking only that it runs and writes a result for every key (see the encoder-length note below for why the output is not compared).

- The existing `test_batch_beam_search.py` compares the legacy `BeamSearch` against `BatchBeamSearch`; that is the regression guard for the `n_utt == 1` path and it passes unchanged.
- `test/espnet2/legacy/test_batch_beam_search_batched.py` (new): batched decoding against the same class run one utterance at a time, over CTC weight {0, 0.5, 1.0} x LM {0, 0.5} x beam {1, 3} x `normalize_length` x float32/float64, plus uniform-length, `maxlenratio` and per-utterance-primer cases. Token sequences and scores match.
- Mutation check: disabling the decoder's `xs_mask` makes 30 of these fail, so the padding mask is genuinely covered.
- `test_asr_inference.py` / `test_s2t_inference.py`: `batch_decode` vs. per-utterance decoding, input-order restoration, and the rejection paths.
- `test_partially_AR.py` and the `BatchBeamSearchOnlineSim` / `Speech2TextStreaming` tests pass unchanged.
- Across `test/espnet2/legacy`, `test/espnet2/lm`, `test/espnet2/asr/decoder` and the inference bins, the set of failing tests is identical before and after (all are missing optional dependencies in my environment).
- End to end through `asr.sh`, `--batch_size 2` decodes the mini_an4 test sets and writes a result for every key. It does **not** reproduce `--batch_size 1` exactly there; see the note on encoder length below.
- End to end with an encoder that has no convolutional subsampling, `--batch_size 1` and `--batch_size 3` produce byte-identical output files on 5 utterances of different lengths with 2-best and CTC.

**Review feedback addressed** (commit `33c0496`)

- The batched retry decodes only the utterances that produced nothing, so a per-utterance `hyp_primer` has to be narrowed to match. It was not, and `_primers` raised `ValueError: got N hyp primers for M utterances`, failing the whole batch. Reachable from S2T, which sets per-utterance primers. Fixed, with a regression test that reproduces it when the fix is reverted.
- `encode` raises `TooShortUttError` for a whole minibatch when a single utterance is too short, and both inference scripts then wrote the placeholder n-best list for every key. They now retry the batch one utterance at a time and only replace the ones that really fail.
- `forward` is a template method, so `init_hyp` and `post_process` are called the way they were before utterance batching whenever a single utterance is decoded. An override written against the old signature keeps working.
- Moved the device/dtype placement in `s2t_inference.py` out of the `partial_ar` conditional. This is pre-existing on master — `PartiallyARInference` never received `--ngpu` / `--dtype` — but this PR touches that block, so it is fixed here.

**A memory bug this turned up**

Scorer states are slices of `(n_utt * beam, ...)` tensors, so a hypothesis's state is a view onto the running batch. A hypothesis that has ended is held until the whole batch finishes, so it pinned a storage `n_utt * beam` times larger than it needs — once for every step at which any hypothesis ended. On a 4-utterance, beam-5 decode each ended hypothesis kept 20x more memory alive than its state occupies, for both the CTC and the decoder states.

This exists on master too, where the factor is the beam size; batching utterances multiplies it by the number of utterances, which is why it is fixed here. Nothing scores an ended hypothesis again, so copying its state out of the shared storage costs nothing and the frozen-reference cases still pass unchanged.

**Known limitations**

- `ScoreFilter` in `s2t_inference.py` still implements `batch_score` as a Python loop over hypotheses, so its cost grows from `beam` to `n_utt * beam` calls per step. S2T will see less of the speedup above until it is vectorized; happy to do that in a follow-up.
- **Batched decoding does not reproduce `--batch_size 1` exactly for encoders with convolutional subsampling.** The beam search itself is exact — given the same encoder output it returns the same hypotheses and scores, which is what the unit tests check. The encoder is what differs: `Conv1dSubsampling2` (and its siblings) subsample the padding mask with `x_mask[:, :, :-2:1][:, :, :-2:2]`, which slices relative to the *padded* width of the batch, so an utterance's `enc_olens` depends on the longest utterance next to it. Measured on mini_an4 with `input_layer: conv1d2`, one utterance of 36800 samples:

  | batch | enc_olens |
  |---|---|
  | alone, or padded alone | 142 |
  | next to a 46400-sample utterance | 144 |
  | next to a copy of itself | 142 |

  This is pre-existing behaviour, unrelated to this PR — the same happens during training — but it is what makes batched decoding differ. The extra frames are real audio that the single-utterance path truncates, so batching is arguably closer to correct, but the results are not identical. Bucketing utterances by length reduces the effect. Filed separately as #6600.
- A finished utterance stays in the batch until the whole batch is done, so a long straggler wastes some compute. `CTCPrefixScoreTH` fixes its batch size at construction, so shrinking mid-decode is not possible without changes there.
- Pre-existing and unrelated: `CTCPrefixScoreTH` raises `IndexError` when asked to score a prefix longer than the encoder output. S2T hypotheses start with a 4-token prompt, so S2T + CTC decoding can hit this on master today whenever `end_detect` does not stop the search first.


sw005320:
@coderabbitai review

espnet2/legacy/nets/batch_beam_search.py
```diff
@@ -390,34 +844,391 @@ def post_process(
...
+                active[lo:hi] = [False] * n_hyp_per_utt
+            elif maxlenratio == 0.0 and end_detect(
+                [h.asdict() for h in ended_per_utt[b]], i
+            ):
+                logger.info(f"end detected at {i}")
+                done[b] = True
+                active[lo:hi] = [False] * n_hyp_per_utt
+
+        done = torch.tensor(done, dtype=torch.bool)
+        if n_utt == 1:
```
chenehk, line 905: This case applies to both non-batch input [T, D] and batch input with batch=1 [1, T, D], which will crash if we don't keep the fixed number of hypotheses. I think we can either squeeze the data with [1, T, D] -> [T, D], or drop this line.

chenehk (commented):
I found a potential crash. It's quite a corner case, and other than this, it looks good to me!

espnet2/legacy/nets/batch_beam_search.py
sw005320: Confirmed, and in one of the two cases it was worse than a crash.

Instrumenting the search on a `(1, T, D)` input shows the encoder tensor pinned at `beam` rows while the hypothesis set is 1, then 2: `[(1, (3, 4, 256)), (2, (3, 4, 256))]`. With a padding mask that raises -- `The size of tensor a (7) must match the size of tensor b (21)` -- but without one it silently returned the correct answer, because the replicated memory rows are identical and `view()` folds them into the time axis, where attention over the repeated copy gives the same weighted sum. It was working by accident, which is why the tests did not catch it.

It is also reachable in ordinary use rather than only by hand: `--batch_size N` produces a batch of one whenever the last chunk of a dataset is odd.

Fixed in `a542c23`, taking your second option rather than squeezing at the entry: with one utterance the encoder output is not replicated at all, and `search` expands it to the current width the way the `(T, D)` path already does, with a one-row mask expanded alongside. That leaves the compaction invariant intact for `BatchBeamSearchOnlineSim` and `BatchBeamSearchOnline`, which rely on the running set shrinking. `(1, T, D)` now reproduces `(T, D)` exactly, padded or not, with a regression test for both; reverting the fix fails it.

Thank you -- this was a real gap in my testing.

chenehk (approved):
LGTM!

PR #6601, opened 2026-09-01
Retry the kaldi binary download the way it actually fails

sw005320:
`test_integration_espnet2 (3.12, 2.9.1, lm)` died in `Install Kaldi`:

```
Connecting to release-assets.githubusercontent.com|185.199.110.133|:443... connected.
Unable to establish SSL connection.
##[error]Process completed with exit code 4.
```

The download already had `wget --tries=3`, **which reads like a retry and is not
one here.** GitHub's release assets redirect to
`release-assets.githubusercontent.com`, and it is the handshake there that breaks.
wget's `--tries` does not open a new connection for that, so one failed handshake
ends the step while the flag suggests three attempts were made.

## Not a new solution

`install_ffmpeg.sh` had already solved this, as a local function: `--tries=1` per
attempt so each is a fresh connection, exponential backoff, and a `tar tf` check
because a mirror can answer 200 with an HTML error page that only fails later in
`tar`.

Rather than copy 25 lines, that function moves to
`tools/installers/download_with_retry.sh` and both callers source it. **Duplicating
it is the shape of defect this CI keeps producing** — two hash-input lists, four
python declarations, `tools/setup.py` — and a retry helper drifting between two
copies would be the same story again.

## One behaviour detail moved with it

The local version hardcoded `--no-check-certificate`. The shared one does not,
because kaldi's download is from GitHub and has never disabled certificate
verification. Both ffmpeg call sites now pass it explicitly, so ffmpeg behaves
exactly as before and kaldi does not silently inherit a weaker check.

Also replaced `[ ! -e X ] && wget ...` with an explicit `if`. The `&&` form does
not abort under `set -e` when the file exists, which is fine where it sits, but it
is one subtlety fewer.

## Verified

Each branch against a `wget` stub — after first getting the stub wrong, since `-O`
is not at `$3` and the output path has to be parsed out of `"$@"`:

| case | result |
|---|---|
| first attempt succeeds | rc=0, 1 attempt |
| fails then succeeds | rc=0, **2 attempts** |
| 200 with an HTML error page | rc=1, 3 attempts, reports "not a valid archive" |
| handshake fails every time | rc=1, 3 attempts |

Merges cleanly with #6596 and #6597.

## Context

This is the **fifth** distinct infrastructure flake this week, after apt 403,
chocolatey 503, `torch.hub`'s anonymous GitHub API limit, and Hugging Face 429.

### DIFF
--- a/ci/install_kaldi.sh
+++ b/ci/install_kaldi.sh
@@ -26,7 +26,12 @@ set -euo pipefail
 # shellcheck source=tools/installers/download_with_retry.sh
 . ./tools/installers/download_with_retry.sh
 
-if [ ! -e ubuntu16-featbin.tar.gz ]; then
+# Reuse an existing archive only if it really is one. Skipping the download on
+# `-e` alone would hand a partial file straight to the tar below, bypassing the
+# validation this change exists to add.
+if [ ! -e ubuntu16-featbin.tar.gz ] \
+        || ! tar tf ubuntu16-featbin.tar.gz > /dev/null 2>&1; then
+    rm -f ubuntu16-featbin.tar.gz
     download_with_retry \
         https://github.com/espnet/kaldi-bin/releases/download/v0.0.1/ubuntu16-featbin.tar.gz \
         ubuntu16-featbin.tar.gz
--- a/ci/install_kaldi.sh
+++ b/ci/install_kaldi.sh
@@ -12,6 +12,24 @@ set -euo pipefail
 
 # download pre-built kaldi binary
 # TODO(karita) support non ubuntu env
-[ ! -e ubuntu16-featbin.tar.gz ] && wget --tries=3 https://github.com/espnet/kaldi-bin/releases/download/v0.0.1/ubuntu16-featbin.tar.gz
+#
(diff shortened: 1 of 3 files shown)


sw005320:
@coderabbitai correct, and it defeats the point of the change it is guarding. The
whole reason for the helper is that a download can arrive corrupt; checking `-e`
and skipping means the one file *most likely* to be corrupt — a leftover from a
failed attempt — is the one that never gets checked.

Took the suggestion, and fixed the same thing one level down as well.

**The caller** now reuses the archive only when `tar tf` says it is one, and
removes it otherwise.

**The helper** now removes its output after a failed attempt. `wget -O` truncates
the target before writing, so a failure leaves a partial or empty file behind for
whatever looks next. Fixing it there means the next caller does not have to
rediscover this — which is the argument for having extracted the helper in the
first place.

Verified both:

| case | result |
|---|---|
| helper, every attempt fails | 3 attempts, **no output file left** |
| guard, file absent | downloads |
| guard, file present but corrupt | removes it and downloads |
| guard, file present and valid | reuses it |

The third row is the hole you reported.

PR #6605, opened 2026-09-02
[espnet3] Make EpochSyncIterator re-iterable so chunk-factory training survives past epoch 0

NewGamezzz:
## What did you change?

This PR makes `EpochSyncIterator` re-iterable and aligns the epoch numbering with the espnet2 convention. Without these fixes, training with `ChunkIterFactory` silently stops after the first epoch, and `shuffle: true` crashes at epoch 0. The changes are as follows:

1. Renamed `iterator.py` to `epoch_sync_iterator.py` so the module name matches the class it contains.
2. `EpochSyncIterator` now stores a zero-argument factory `partial(iter_factory.build_iter, epoch)` instead of a built iterator, and calls it once per `__iter__` so every pass gets a fresh iterator. This matters because `ChunkIterFactory.build_iter` returns a one-shot generator rather than a re-iterable loader. Lightning iterates the dataloader more than once per epoch (a setup/prefetch pass plus the epoch loop), which closed the shared generator and left every epoch after the first empty. A re-iterable source (e.g. a DataLoader) also still works.
3. `DataLoaderBuilder` now passes `self.epoch + 1` to `build_iter`, since espnet2 iter factories count epochs from 1 while Lightning's `current_epoch` counts from 0. This fixes the `ValueError: Seed must be between 0 and 2**32 - 1` crash with `shuffle: true` at epoch 0. Note: because the shuffle seed depends on the epoch number, a run resuming from a checkpoint trained before this fix will see **a different batch order at the same epoch**.
4. `__len__` caches its result and hands the pass it builds for the length probe to the next `__iter__`, so the sequence-factory path costs two `build_iter` calls per epoch instead of three.

---

The current `EpochSyncIterator` only works with re-iterable iterator such as `SequenceIterFactory`, which returns `DataLoader`. With `ChunkIterFactory`, which returns a generator, training works only for the first epoch. From epoch 1 on, the PyTorch Lightning trainer calls `iter()` on the train loader twice per epoch and discards the first pass
(https://github.com/Lightning-AI/pytorch-lightning/blob/21ac84116633b8c629da03c41e6d6ec2cfb15241/src/lightning/pytorch/loops/training_epoch_loop.py#L240-L246). Every epoch after the first then trains 0 batches while the run exits 0.

A simple snippet replicating the issue:

```python
class MiniDataLoader:
    """SequenceIterFactory.build_iter's return: a RE-ITERABLE container.

    Like torch's DataLoader: holds the data, and every iter() call
    manufactures a brand-new internal cursor.
    """

    def __init__(self, batches):
        self.batches = batches

    def __iter__(self):
        return iter(list(self.batches))   # fresh cursor per call

    def __len__(self):
        return len(self.batches)          # sized (matters in real Lightning)


class OldEpochSyncIterator:
    """The OLD EpochSyncIterator, unchanged: stores whatever build_iter returned."""

    def __init__(self, source):
        self._source = source

    def __iter__(self):
        yield from self._source


def chunk_build_iter():
    """ChunkIterFactory.build_iter: generator function -> ONE-SHOT cursor."""
    yield from ["batch1", "batch2", "batch3"]


def sequence_build_iter():
    """SequenceIterFactory.build_iter: returns a RE-ITERABLE container."""
    return MiniDataLoader(["batch1", "batch2", "batch3"])


def one_epoch_like_lightning(loader):
    pass1 = iter(loader)   # FitLoop.setup_data()
    next(pass1)            # the fetcher's prefetch pull
    pass1 = iter(loader)   # on_run_start -> discards + closes started pass 1
    print(f"epoch trains on {list(pass1)}")


one_epoch_like_lightning(OldEpochSyncIterator(chunk_build_iter()))     # []
one_epoch_like_lightning(OldEpochSyncIterator(sequence_build_iter()))  # full pass
```

---

## Is your PR small enough?

Yes! It contains 5 file changes with 488 changed lines.

---

## Additional Context

This has been tested with espnet3's LibriTTS codec recipe (https://github.com/espnet/espnet/pull/6497). I trained the model for 3 epoch on a small subset of LibriTTS with 2 V100. The per-step losses are bit-identical to the old code. **Note that this is test with shuffle: False**


espnet3/components/data/epoch_sync_iterator.py
```diff
@@ -94,7 +130,17 @@ def generate(self):
             >>> list(NonEmptyIterator([[0], [], [1]]))
             [[0], [1]]
         """
-        yield from self._iterator
+        yield from self._new_pass()
+
+    def _new_pass(self):
```
Masao-Someki: PLease change this to verb, so it can be `_get_new_pass`

espnet3/components/data/epoch_sync_iterator.py
```diff
@@ -47,32 +57,57 @@ class EpochSyncIterator:
...
-        return len(self._iterator)
+        if self._length is None:
+            # Build the pass outside the try: only the len() probe may be
+            # read as "unsized" - a TypeError raised inside the factory is a
+            # real bug and must propagate.
+            new_pass = self._new_pass()
+            try:
+                self._length = len(new_pass)
+            except TypeError:
+                self._length = _UNSIZED
```
Masao-Someki, line 109: I thought we could change it to None? Do we need to use an object class here?

Masao-Someki (commented):
Thanks @NewGamezzz I checked it, I had one request and one question.

espnet3/components/data/epoch_sync_iterator.py
NewGamezzz: Thanks. I have changed it.

espnet3/components/data/epoch_sync_iterator.py
NewGamezzz: I changed `_UNSIZED` to `-1` instead. `None` cannot be used here because I already used it to represent that the length is not probe yet. On the first `__len__` call, `EpochSyncIterator` builds a fresh iterator to get its length, then caches the result. If `_UNSIZED` were also stored as None, it will force `EpochSyncIterator` to rebuild the iterator every time we call `__len__`.

sw005320 (commented):
Thanks @NewGamezzz — I went through this and I think the direction is right. The root-cause analysis is accurate: `ChunkIterFactory.build_iter` is a generator function, so what the old code stored was a one-shot generator, which cannot survive the two `iter()` calls Lightning makes per epoch. Taking a zero-argument callable and calling it once per `__iter__` is the right fix, and keeping `generate()` as the extension point so subclasses inherit the DDP sync is a nice touch.

I also confirmed that `reload_dataloaders_every_n_epochs = 1` is forced for the iter-factory path in `espnet3/components/trainers/trainer.py`, so the "one instance covers exactly one epoch" assumption behind the `__len__` cache and the epoch baked into the `partial` does hold.

No blockers from me. Three comments below.

---

### 1. The `self.epoch + 1` change deserves a mention in the PR description

This is a second, independent fix bundled into a PR about re-iterability. It is a legitimate one — `SequenceIterFactory` computes `divmod(num_iters_per_epoch * (epoch - 1), N)` and seeds `RandomState(real_epoch - 1 + seed)`, so espnet2 assumes a 1-based epoch and Lightning's 0-based `current_epoch` blows up. You even added `test_epoch0_with_shuffle_and_num_iters_per_epoch_does_not_crash` for it, which is good.

But it is not mentioned in the description, and it does not sit well with the claim that the per-step losses are bit-identical to the old code. The shuffle seed is `epoch + seed`, so shifting the epoch by one changes the batch order. I suspect the recipe you validated with has shuffle disabled. Could you add this fix to the description, and note that resuming an existing checkpoint will see a different data order than before?

Related: `_maybe_shard_dataset` still uses the 0-based `self.epoch` for shard scheduling, so the two epoch conventions now coexist in the same class. Binding it to a local like `espnet2_epoch = self.epoch + 1`, or adding a one-line comment, would save the next reader a lookup.

### 2. `build_iter` now runs up to 3x per epoch on the sequence path

Per epoch the calls are: one for the length probe, one for Lightning's `setup_data`, and one for the re-`iter()` in `on_run_start`. Previously it was called once, at builder time.

For `ChunkIterFactory` this is free, since calling a generator function does not execute its body. For `SequenceIterFactory` each call runs `sampler.generate()` and shuffles the whole batch list, so this is O(corpus) three times per epoch instead of once.

Handing out the pass you already built for the length probe would bring that back to 2:

```python
def _get_new_pass(self):
    if not callable(self._source):
        return self._source
    if self._pending_pass is not None:
        pending, self._pending_pass = self._pending_pass, None
        return pending
    return self._source()
```

with `__len__` stashing its probe into `_pending_pass`. Correctness is preserved in the abandoned-prefetch case: once the handed-out pass is closed, the next one is built fresh. Not a blocker, but worth considering before this hits large corpora.

### 3. The callable/iterable dual mode is callable-only in production

`DataLoaderBuilder` only ever passes a `partial`; the iterable branch is exercised solely by tests and docstrings. There is also a theoretical mismatch if someone passes an iterable that happens to define `__call__`. Keeping the iterable path for backward compatibility is fine, but a one-line comment saying that is why it is there would help.

---

### CI

The espnet3 jobs all pass. The six `test_python_espnet2` failures are nltk resource `LookupError`s in the phoneme tokenizer tests, and this diff touches no espnet2 code. The same job passes on more recent PRs, so a rebase onto master and a rerun should clear it.

CodeRabbit's point about caching a factory-raised `TypeError` as "unsized" is already addressed by building the pass outside the `try`, and both of @Masao-Someki's comments look resolved.

NewGamezzz:
Thanks @sw005320, I have reflected on your comments. Here is my response to each:

1. I have updated the PR description to cover this fix. You are right about the bit-identical claim. The equivalence test was run with shuffle disabled, so it validates the re-iterability change only, and I have made that explicit in the description. I also added `espnet2_epoch = self.epoch + 1` with a comment, so the espnet2 (1-based) and Lightning (0-based) epoch conventions are now explicitly separated in DataLoaderBuilder.
2. Agreed, and implemented as suggested. `__len__` now stashes the pass it builds for the length probe into `_pending_pass`, and `_get_new_pass` hands it out once before falling back to the factory, so the sequence path is back to two `build_iter` calls per epoch. A handed-out pass is never reused, so the abandoned-prefetch case still gets a fresh pass.
3. I have added a comment in `_get_new_pass` noting the plain-iterable branch is kept for backward compatibility and that `DataLoaderBuilder` always passes a callable.

Masao-Someki (approved):
Thanks, @NewGamezzz!

sw005320:
Thanks, @NewGamezzz!

PR #6608, opened 2026-09-02
Fix ProcessContext import in AbsTask

ondatk68:
## What did you change?
Moved the `ProcessContext` import outside the `wandb` import exception handler in `espnet2/tasks/abs_task.py`.

Before:
```python
  try:
      import wandb
  except Exception:
      wandb = None

      from torch.multiprocessing.spawn import ProcessContext
```

After:
```python
  try:
      import wandb
  except Exception:
      wandb = None

  from torch.multiprocessing.spawn import ProcessContext
```
---

This bug was introduced in https://github.com/espnet/espnet/pull/6399, which removed the PyTorch version guard around the ProcessContext import.
Before this change was merged, `ProcessContext` was defined regardless of whether wandb was available:
```python
  try:
      import wandb
  except Exception:
      wandb = None
  
  if V(torch.__version__) >= V("1.5.0"):
      from torch.multiprocessing.spawn import ProcessContext
  else:
      from torch.multiprocessing.spawn import SpawnContext as ProcessContext
```
After the version guard was removed, the remaining import was unintentionally left indented inside the except Exception block for the wandb import. Then `ProcessContext` is not defined when wandb is installed and imports successfully.

---

## Is your PR small enough?
yes

---


sw005320 (approved):
Good catch!
Thanks!

PR #6609, opened 2026-09-02
Stop declaring the three forked packages, which PyPI refuses

sw005320:
## What is wrong

espnet has not been published to PyPI since **202511**. Both `v.202604` tags and the `v.202609` tag failed at the upload step:

```
ERROR HTTPError: 400 Bad Request from https://upload.pypi.org/legacy/
400 Can't have direct dependency: ctc-segmentation @
    git+https://github.com/espnet/ctc-segmentation.git@9b9ea1d ; extra == "asr"
```

Trusted publishing was **not** the problem — authentication succeeded and PyPI rejected the *metadata*. `pyproject.toml` declared three dependencies as PEP 508 direct references, and PyPI refuses any distribution whose metadata contains one:

| | declared in | pinned to |
|---|---|---|
| `ctc-segmentation` | extra `asr` | `9b9ea1d` |
| `g2p_en` | extra `tts` | `master` |
| `s3prl` | extra `s2st` | `572af70` |

All three are forks under `github.com/espnet` because upstream stopped maintaining them, so switching to the PyPI releases is not an option.

## Why removing them is safe

`tools/Makefile` already installs all three, at pinned commits, and has since before this broke:

```make
espnet.done: pytorch.done conda_packages.done lightning.done
	# Install additional packages that cannot be included in setup.py due to package build issues.
	... pip install git+https://github.com/espnet/g2p.git@053dfa9
	... pip install git+https://github.com/espnet/ctc-segmentation.git@9b9ea1d
```

plus `s3prl.done` → `installers/install_s3prl.sh`, which `ci/install.sh` requests explicitly. The declarations in `pyproject.toml` were duplication — and *drifting* duplication, since pyproject pinned `g2p_en` at `@master` while the Makefile pins `053dfa9`.

## This is a regression, not a design

At `v.202511`, `setup.py` carried this line where the two used to be:

```python
# Moving g2p_en and ctc-segmentation to makefile
```

The `setup.py` → `pyproject.toml` migration put them back as direct references. Nothing recorded *why* they had been moved, so every release for the next five months failed for a reason nobody could see. The comment this PR adds is that record.

## Why nothing caught it, and what now does

The build succeeds, and **`twine check` reports `PASSED` on the rejected artifact** — verified by rebuilding with one reference restored:

```
Checking dist2/espnet-202609-py3-none-any.whl: PASSED
Checking dist2/espnet-202609.tar.gz: PASSED
→ direct references in metadata: 1
  Requires-Dist: ctc-segmentation @ git+https://... ; extra == "asr"
```

Only PyPI's upload endpoint objects, and that is five months downstream of the mistake. So the rule becomes the tenth invariant in `ci/check_ci_image_config.py`, which runs on every PR, and `doc/make_release.py`'s preflight reuses that function rather than restating it. It matches on the PEP 508 `@ <url>` separator rather than on `git+`, because PyPI rejects `https://` archives and `file://` paths just the same.

## Verification

- Each reference restored in turn → checker names the file, line and section, exits 1
- An `https://` archive reference added → also caught
- `torch>=2.9.1`, `espnet_tts_frontend`, `pypinyin<=0.44.0`, `extras-pkg[all]>=1.0` added → no false positives
- `publishable_metadata()` run from another directory → still finds `pyproject.toml`
- Built wheel and sdist: **112 `Requires-Dist` entries, 0 direct references**, `Requires-Python: <3.14,>=3.12`
- `black --check`, `pycodestyle --max-line-length=88`: clean

## After this merges

PyPI never accepted 202609, so the version number is still free. Re-running `publish_python_package.yml` against a tag that includes this commit should upload `espnet 202609`.

### DIFF
--- a/ci/check_ci_image_config.py
+++ b/ci/check_ci_image_config.py
@@ -47,7 +47,16 @@ Every job must also have a permissions block in scope, at the workflow or the
 job level. Inheriting the repository default means inheriting write on almost
 everything, and nothing fails to say so.
 
-And seventh, the workflow files must have no duplicate mapping keys. PyYAML
+Seventh, pyproject.toml must declare no direct references - a dependency
+written "name @ git+https://..." rather than as a version range. PyPI refuses any
+distribution whose metadata contains one, with 400 Can't have direct dependency,
+and nothing else in the repository looks: the build succeeds, every test passes,
+and the failure surfaces only when a release tag is pushed. Three of them arrived
+with the setup.py to pyproject.toml migration and went unnoticed for five months,
+because the release before it had deliberately moved the same three packages out
+to tools/Makefile and nothing recorded why.
+
+And eighth, the workflow files must have no duplicate mapping keys. PyYAML
 accepts them and lets the last one win, so writing a second env: block into a
 step silently discards the first - which is exactly what nearly dropped
 GITHUB_TOKEN from the two steps that need it for torch.hub.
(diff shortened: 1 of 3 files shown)


sw005320:
Superseded. Removing the three declarations weakens `pip install espnet[asr]`, which is not an acceptable trade.

The three forks are being published to PyPI under their own names instead - `espnet-ctc-segmentation`, `espnet-g2p-en`, `espnet-s3prl` - so `pyproject.toml` can declare them as ordinary version requirements. A replacement PR will do that, and will keep the direct-reference invariant added here.

Findings worth keeping from this one:

- `twine check` reports PASSED on the artifact PyPI rejects, so nothing local catches a direct reference before a tag is pushed.
- At `v.202511`, `setup.py` carried the line `# Moving g2p_en and ctc-segmentation to makefile`. The `setup.py` to `pyproject.toml` migration reintroduced them as direct references, which is what broke publishing for five months.
- `ctc-segmentation` needs no ESPnet-specific patch at all: upstream `master` already replaced `oldest-supported-numpy` with `numpy` in its build requirements and imports cleanly on Python 3.12. Only the 1.7.4 release is broken.

PR #6612, opened 2026-09-02
Raise max-parallel from 20 to 30, where it actually binds

sw005320:
## The cap is not the account's

Only **21 jobs** were running on a pull request against an allocation of 60. The limiter is `max-parallel: 20`, set on every matrix in `ci_on_ubuntu.yml` since [#5360](https://github.com/espnet/espnet/pull/5360) in July 2023.

**Seven of the eight are no-ops.** The matrices are:

| job | jobs | `max-parallel: 20` |
|---|---|---|
| **`test_integration_espnet2`** | **65** | **binds — 4 waves of 20** |
| `test_configuration_espnet2` | 9 | no effect |
| `test_python_espnet2` / `test_python_espnet3` / `test_integration_espnet3` | 6 each | no effect |
| `test_utils_espnet2` | 2 | no effect |
| `test_shell_espnet2` / `test_import` | 1 each | no effect |

Every test job here depends on nothing but `process_labels` and `resolve_ci_image`, so they all become eligible together. The small matrices drain in the first few minutes, and the run then spends most of its length with 20 integration jobs running and 45 waiting.

## Why 30, and not 60

That phase is throughput-limited, until it isn't. Measured on [run 33629114132](https://github.com/espnet/espnet/actions/runs/33629114132): 65 jobs, ~1170 minutes of compute, 18 minutes on average, **38 for the longest**.

| `max-parallel` | integration phase, wall clock |
|---|---|
| 20 (today) | ~59 min |
| **30** | **~39 min** |
| 40 | ~38 min — the longest single job is now the floor |

So past about 31 there is nothing left to divide. 30 captures essentially all of the available gain.

Peak demand across the workflow becomes `30 + 9 + 6 + 6 + 6 + 2 + 1 + 1 = 61`, plus 3 from the macOS run, 2 from Windows and 1 from Debian. Slightly over 60, so a few jobs queue at the start until the small matrices finish — which costs nothing, and only matters while several PRs are open simultaneously.

## Applied to all eight

Not only to the one that binds, so the number means one thing throughout — no single matrix takes more than half the allocation — and a matrix that later grows past 20 is not silently throttled by a value nobody remembers setting.

## Notes

- `.github/workflows/**` is **not** among the prebuilt image's hash inputs (`ci/install.sh`, `ci/install_kaldi.sh`, `ci/no_redistribute.txt`, `docker/ci.dockerfile`, `pyproject.toml`, `tools/Makefile`, `tools/installers/**`), so this does not force an image rebuild and this PR's own CI runs on the published image.
- `python3 ci/check_ci_image_config.py` exits 0; the workflow still parses.
- Wall-clock figures come from a single run and are not comparable across days — on a busy fleet, queue wait dominates. The compute total and the longest-job figure are what the choice of 30 rests on, and those are environment-independent.

### DIFF
--- a/.github/workflows/ci_on_ubuntu.yml
+++ b/.github/workflows/ci_on_ubuntu.yml
@@ -95,7 +95,7 @@ jobs:
   #   if: |
   #     github.event.pull_request.draft == false
   #   strategy:
-  #     max-parallel: 20
+  #     max-parallel: 30
   #     matrix:
   #       python-version: ["3.12", "3.13"]
   #       pytorch-version: [2.9.1, 2.10.0, 2.11.0]
@@ -208,7 +208,7 @@ jobs:
       github.event.pull_request.draft == false
     strategy:
       fail-fast: false
-      max-parallel: 20
+      max-parallel: 30
       matrix: ${{ fromJSON(needs.resolve_ci_image.outputs.matrix) }}
     # Everything Make environment used to build is already here. Measured on
     # #6566: 81s to pull against 510s to build, and the whole job 4.2 min
@@ -312,7 +312,7 @@ jobs:
       github.event.pull_request.draft == false
     strategy:
       fail-fast: false
-      max-parallel: 20
+      max-parallel: 30
       matrix:
         python-version: ["3.12", "3.13"]
         pytorch-version: [2.9.1]
@@ -372,7 +372,7 @@ jobs:
       github.event.pull_request.draft == false
     strategy:
       fail-fast: false
-      max-parallel: 20
+      max-parallel: 30
       matrix:
         python-version: ["3.12"]
         pytorch-version: [2.9.1]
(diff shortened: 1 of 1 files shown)


sw005320:
Closing. The measurements here are sound but the premise is not, and the day this was opened disproved it. Recording both so the next person does not repeat the work.

## What was measured, and still holds

`max-parallel: 20` is set on eight matrices in `ci_on_ubuntu.yml`, and **seven of them are no-ops**. The matrices are 65, 9, 6, 6, 6, 2, 1 and 1 jobs, so only `test_integration_espnet2` is ever throttled — and it is throttled hard: 65 jobs twenty at a time.

Every test job here depends on nothing but `process_labels` and `resolve_ci_image`, so they all become eligible together, the small matrices drain in the first few minutes, and the run then spends most of its length with 20 integration jobs running and 45 waiting.

Measured on [run 33629114132](https://github.com/espnet/espnet/actions/runs/33629114132): 65 jobs, about 1170 minutes of compute, 18 minutes on average and **38 for the longest**.

| `max-parallel` | integration phase, wall clock |
|---|---|
| 20 | ~59 min |
| 30 | ~39 min |
| 40 | ~38 min — the longest single job is now the floor |

So past about 31 there is nothing left to divide, and 60 buys exactly what 40 does.

## Why the premise is wrong

This PR said the cost "only matters while several pull requests are open at once, which is not how this repository is usually busy." Measured a few hours later, on the day this was opened:

```
in_progress = 60 / 60,  queued = 733
```

across ten concurrent pull requests, none of them this one. When the account's concurrency limit is the binding constraint, raising per-matrix parallelism changes nothing — the run simply waits for slots at a different place. Peak demand would have gone from 53 to 63 jobs against a cap of 60, so the extra ten would have queued.

The days when 30 helps are the days when CI is not busy, which are the days the wait does not matter.

## What actually reduces the integration phase

Not parallelism — the phase is throughput-limited, so the lever is total compute, not how it is spread. Two things measured the same day:

- **39% of all CI compute ran after the first failure was already known** (eight runs, 12,964 minutes). Mechanisms for acting on that were explored and rejected in #6620, which records why.
- Cancelling one's own run on seeing a failure, done by hand on three pull requests, freed 20 of the 60 slots immediately.

Neither needs `max-parallel` touched.

## If someone revisits this

The number to change is on `test_integration_espnet2` alone; the other seven are decoration and setting them is only useful so that a matrix which later grows past 20 is not silently throttled by a value nobody remembers setting. And the ceiling is the longest single job, which was 38 minutes when this was measured — check that first, because it moves whenever the task split in `ci/test_integration_espnet2.sh` changes.

PR #6616, opened 2026-09-02
Remove the remaining dead torch 1.x version conditions from the tests

amirlankalm:
## What did you change?

Follow-up to #6536, where you asked for the PyTorch version conditions to be
removed from the source rather than turned into skips. That PR removed the ones
it already touched; this one does the rest.

**Scope is `is_torch_1_*_plus` only.** Those are dead under either floor: the
maintainer statement of `>=2.9.1` and the `torch>=2.3.1` currently declared in
`pyproject.toml`. **`is_torch_2_6_plus` and `is_torch_2_9_plus` are deliberately
left alone**, because they are *not* dead under 2.3.1, so which floor governs
decides their fate. Happy to take those out too if you confirm 2.9.1 is the
number, and `pyproject.toml` probably wants updating either way.

67 sites in 23 files, edited by shape rather than deleted uniformly:

- compound conditions keep their live half
  `not is_python_3_8_plus or not is_torch_1_7_plus` -> `not is_python_3_8_plus`
  `not is_torch_1_8_plus or is_torch_2_9_plus` -> `is_torch_2_9_plus`
- `if FLAG: A else: B` -> `A`
- where the condition is always **false**, the `else` branch is what survives
  instead (one site, in `test_asr_transducer_inference.py`)
- `X if FLAG else Y` -> `X`, and `arg=FLAG` -> `arg=True`
- skips whose condition can never fire are removed with their bodies
- 28 flag definitions and 20 now-unused imports, including two `distutils`
  imports, which also helps since distutils is gone on Python 3.12+

Every one of these flags is always `True`, so a correct edit must change nothing.
Verified by running all 21 collectable files before and after and diffing
per-test outcomes: **2454/2454 identical**, including every skip reason and
count. The only textual difference was source line numbers moving inside skip
messages because the files got shorter.

The 14 failures and 9 errors in that run are pre-existing and reproduce on the
base commit. `test_espnet_model.py` and `test_espnet_model_tse.py` could not be
collected locally because an optional enhancement dependency is missing, so their
changes rest on CI rather than on my run.

## Is your PR small enough?

23 files, +81/-221, one concern.

## Additional context

Noticed and **not** changed, since fixing it would alter behaviour and belongs in
its own PR: in `test_houlsby_adapter_layer.py` the conditions read
`is_s3prl_available and not is_torch_1_8_plus or not is_python_3_8_plus`. Since
`and` binds tighter, the s3prl term was already inert and simplifies away exactly.
It looks like `not is_s3prl_available` was intended.

Environment:

```
- python version: 3.12.8
```

### DIFF
--- a/test/espnet2/asr/decoder/test_whisper_decoder.py
+++ b/test/espnet2/asr/decoder/test_whisper_decoder.py
@@ -2,7 +2,6 @@ import sys
 
 import pytest
 import torch
-from packaging.version import parse as V
 
 from espnet2.asr.decoder.whisper_decoder import OpenAIWhisperDecoder
 
@@ -12,12 +11,11 @@ pytest.importorskip("whisper")
 
 # NOTE(Shih-Lun): needed for `persistent` param in
 #                 torch.nn.Module.register_buffer()
-is_torch_1_7_plus = V(torch.__version__) >= V("1.7.0")
 is_python_3_8_plus = sys.version_info >= (3, 8)
 
 
 @pytest.mark.skipif(
-    not is_python_3_8_plus or not is_torch_1_7_plus,
+    not is_python_3_8_plus,
     reason="whisper not supported on python<3.8, torch<1.7",
 )
 @pytest.fixture()
@@ -30,7 +28,7 @@ def whisper_decoder(request):
 
 
 @pytest.mark.skipif(
-    not is_python_3_8_plus or not is_torch_1_7_plus,
+    not is_python_3_8_plus,
     reason="whisper not supported on python<3.8, torch<1.7",
 )
 @pytest.mark.timeout(50)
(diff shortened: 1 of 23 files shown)


sw005320:
Super thanks!
This really helps to simplify our codes!!!

amirlankalm:
No problem. If 2.9.1 is the real floor I'll take out `is_torch_2_6`/`2_9` too and fix pyproject.

sw005320 (approved):

sw005320:
Please fix the CI error.

amirlankalm:
Fixed. Two F401s I introduced: `torch` in `test_asr_transducer_inference.py` was only used by the `is_torch_1_5_plus` definition, and `ComplexTensor` in `test_dc_crn_separator.py` only by the else branch of the ternary, so removing the dead conditions left both imports unused.

`flake8 --extend-ignore=D test/espnet2` is 0 for the whole tree now.

My local check was pycodestyle plus pre-commit, neither of which runs pyflakes, so the F codes went unchecked. Installed flake8 and re-ran my other open branches, they are clean.

PR #6617, opened 2026-09-02
Remove the temporary wav that `MiniOmniE2EModel` writes per turn

amirlankalm:
## What did you change?

`warmup()` and `forward()` write the user turn to a `NamedTemporaryFile` with
`delete=False`, and nothing in the package ever removes it, so every call leaves
a wav behind. `sds1` runs as a long-lived Gradio app, so it grows without bound:

```
 2 s user turn ->  62.5 KB per call     1000 turns =  64 MB
 5 s user turn -> 156.3 KB per call     1000 turns = 160 MB
10 s user turn -> 312.5 KB per call     1000 turns = 320 MB
```

`delete=False` is necessary here rather than a mistake, which is why the fix is
not simply to flip it. `run_AT_batch_stream` is a **generator**, so it does not
open the path until it is drained, and that happens after the `with` block has
already closed the file. So the file genuinely has to outlive the block, and can
only be unlinked once the generator is exhausted:

```diff
     with tempfile.NamedTemporaryFile(suffix=".wav", delete=False) as f:
         f.write(data_buff)
+        wav_path = f.name
+    try:
         audio_generator = self.client.run_AT_batch_stream(
-            f.name,
+            wav_path,
             ...
         )
-    ans = []
-    while True:
-        ...
+        ans = []
+        while True:
+            ...
+    finally:
+        os.unlink(wav_path)
```

The `finally` also covers a failed generation, which previously leaked with
certainty. Behaviour is otherwise unchanged: same file written, same path passed
to the client, same values returned.

## Is your PR small enough?

Yes. 1 production file, +19/-13, plus a test file.

## Additional context

Three CPU-only tests. They stub the client and the SNAC decoder and point
`tempfile` at a directory of their own so the count is unambiguous, which means
no checkpoint, GPU or ffmpeg is required. The stub client opens the path *inside*
the generator body exactly like the real one, so the ordering constraint above is
genuinely exercised rather than assumed.

Against `master` all three fail with `AssertionError: 1 temp file(s) left behind`.
Against this branch `test/espnet2/sds` is 14 passed, 13 skipped.

Environment:

```
- python version: 3.12.8
```

### DIFF
--- a/espnet2/sds/end_to_end/mini_omni_e2e.py
+++ b/espnet2/sds/end_to_end/mini_omni_e2e.py
@@ -1,5 +1,6 @@
 import base64
 import io
+import os
 import tempfile
 from typing import Optional, Tuple
 
@@ -130,15 +131,19 @@ class MiniOmniE2EModel(AbsE2E):
 
         with tempfile.NamedTemporaryFile(suffix=".wav", delete=False) as f:
             f.write(data_buff)
+            wav_path = f.name
+        try:
             audio_generator = self.client.run_AT_batch_stream(
-                f.name,
+                wav_path,
                 self.stream_stride,
                 self.max_tokens,
                 temperature=self.temperature,
                 top_k=self.top_k,
                 top_p=self.top_p,
             )
-        _ = [k for k in audio_generator]
+            _ = [k for k in audio_generator]
+        finally:
+            os.unlink(wav_path)
 
     def forward(
         self,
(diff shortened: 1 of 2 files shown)


amirlankalm:
Good catch on the write path, that was a real hole. `NamedTemporaryFile` creates the file before `write()` runs, so a failed write left the partial wav behind and neither `finally` was reached.

Switched to writing inside a `TemporaryDirectory` instead, which covers a failed write, a failed generation and the normal path with one mechanism, and drops `delete=False` and the explicit unlink. Added the regression test you asked for.

Verified three ways: on master all 4 fail, on the previous commit 3 pass and the write-failure one fails, on this commit all 4 pass.

sw005320 (approved):

PR #6619, opened 2026-09-02
Import the versa metrics from the modules that define them

### LINKED ISSUE #6618: dialog_eval metrics fail to import against current versa
## Describe the bug

`pyscripts/utils/dialog_eval/` imports ten names from `versa` top-level. On a fresh
install none of them resolve, so `TTS_psuedomos`,
`handle_espnet_TTS_intelligibility` and `handle_espnet_ASR_WER` all raise before
doing any work:

```
ImportError: cannot import name 'pseudo_mos_metric' from 'versa'
```

Those back the "TTS Speech Quality", "TTS Intelligibility" and "ASR WER" options
in the `sds1` demo, so three of the four evaluation modes are unusable.

| imported name | status at current versa |
|---|---|
| `pseudo_mos_setup`, `pseudo_mos_metric` | exist, moved to `versa.utterance_metrics.pseudo_mos` |
| `sheet_ssqa_setup`, `sheet_ssqa` | exist, moved to `versa.utterance_metrics.sheet_ssqa` |
| `espnet_wer_setup`, `espnet_levenshtein_metric` | not present |
| `owsm_wer_setup`, `owsm_levenshtein_metric` | not present |
| `whisper_wer_setup`, `whisper_levenshtein_metric` | not present |

versa has moved to a class-based API and now exports 56 `*Metric` classes
(`EspnetWerMetric`, `ASRMatchMetric`, ...), so the four MOS/SSQA helpers only
need a new import path, while the six WER helpers have no function-style
equivalent left.

## Why it drifted without anyone noticing

Two things, both worth fixing separately from the imports:

1. `tools/installers/install_versa.sh` records the commit it was written against
   in a comment, `3e73ea43659baffb5cd42a8e5946cb659ad27535`, but then does a
   plain `git clone` with no `checkout`, so installs track versa HEAD.
2. `dialog_eval` is imported nowhere under `test/` or `ci/`, so no job ever
   exercises these modules and no CI failure was produced when the API moved.

## To Reproduce

```python
>>> from versa import pseudo_mos_metric
ImportError: cannot import name 'pseudo_mos_metric' from 'versa'
```

or through the package:

```python
>>> from pyscripts.utils.dialog_eval.TTS_speech_quality import TTS_psuedomos
>>> TTS_psuedomos((24000, some_int16_audio))
Error: Versa is not properly installed.
ImportError: cannot import name 'pseudo_mos_metric' from 'versa'
```

## Expected behavior

The three evaluation modes return scores rather than raising.

## Two ways forward

**Pin the installer** to the commit already recorded in the comment. One line,
but that commit needs `pysptk` and `pkg_resources`, so it has its own trouble on
Python 3.12 and it freezes the toolkit against an old versa.

**Migrate `dialog_eval` to the current versa API.** More work, and the six WER
helpers need mapping onto the class-based metrics, but it tracks where versa is
actually going.

I would lean toward migrating, plus a small import smoke test for
`dialog_eval` so this cannot drift silently again. Happy to do either, I just did
not want to pick the API direction unilaterally.

## Environment

```
- python version: 3.12.8
- versa: wavlab-speech/versa HEAD (unpinned, per install_versa.sh)
```

amirlankalm:
Fixes #6618.

## What did you change?

Every helper in `dialog_eval` does `from versa import ...`, and none of those ten
names are re-exported at versa top level any more. So `TTS_psuedomos`,
`handle_espnet_TTS_intelligibility` and `handle_espnet_ASR_WER` all raise before
doing any work, which is three of the four evaluation modes in the `sds1` demo.

The functions are all still there with unchanged signatures, they just moved into
submodules, so this is an import path change and nothing else:

| names | now imported from |
|---|---|
| `pseudo_mos_setup`, `pseudo_mos_metric` | `versa.utterance_metrics.pseudo_mos` |
| `sheet_ssqa_setup`, `sheet_ssqa` | `versa.utterance_metrics.sheet_ssqa` |
| `espnet_wer_setup`, `espnet_levenshtein_metric` | `versa.corpus_metrics.espnet_wer` |
| `owsm_wer_setup`, `owsm_levenshtein_metric` | `versa.corpus_metrics.owsm_wer` |
| `whisper_wer_setup`, `whisper_levenshtein_metric` | `versa.corpus_metrics.whisper_wer` |

No call sites change. `espnet_levenshtein_metric(wer_utils, pred_x, ref_text, fs=16000)`
is exactly what `ASR_WER.py` already passes, so there is no behaviour to reason
about, only where the name comes from.

Verified against versa HEAD. Before, importing any of them fails:

```
ImportError: cannot import name 'pseudo_mos_metric' from 'versa'
```

After, all ten resolve and the code reaches the predictors:

```
10/10 resolve
TTS_psuedomos -> AssertionError: Torch not compiled with CUDA enabled
```

That second error is a **separate** problem and is deliberately out of scope
here: `use_gpu=True` is hardcoded in `model_cache.py` and in the two metric calls
in `TTS_speech_quality.py`, so on a CPU-only install these still fail once past
the import. I have that fix ready as a one-line-per-site follow-up if you want
it, it just seemed like a different concern from the imports.

## Is your PR small enough?

Yes. 4 files, imports only.

## Additional context

No test, for the same reason as #6555: `test/` has no home for `egs2` pyscripts,
the tree only covers `espnet2`, `espnet3` and `egs3`. A small contract test
asserting these ten names exist at those module paths would stop this drifting
silently again, and I am happy to add it if you tell me where it should live.

Worth noting separately, `tools/installers/install_versa.sh` records the commit
it was written against in a comment but then does a plain `git clone`, so installs
track versa HEAD. That is what let the API move without anything failing.

Environment:

```
- python version: 3.12.8
```

### DIFF
--- a/egs2/TEMPLATE/asr1/pyscripts/utils/dialog_eval/ASR_WER.py
+++ b/egs2/TEMPLATE/asr1/pyscripts/utils/dialog_eval/ASR_WER.py
@@ -62,11 +62,9 @@ def handle_espnet_ASR_WER(
          Whisper CER: 6.50"
     """
     try:
-        from versa import (
-            espnet_levenshtein_metric,
-            owsm_levenshtein_metric,
-            whisper_levenshtein_metric,
-        )
+        from versa.corpus_metrics.espnet_wer import espnet_levenshtein_metric
+        from versa.corpus_metrics.owsm_wer import owsm_levenshtein_metric
+        from versa.corpus_metrics.whisper_wer import whisper_levenshtein_metric
     except Exception as e:
         print("Error: Versa is not properly installed.")
         raise e
--- a/egs2/TEMPLATE/asr1/pyscripts/utils/dialog_eval/TTS_intelligibility.py
+++ b/egs2/TEMPLATE/asr1/pyscripts/utils/dialog_eval/TTS_intelligibility.py
@@ -66,11 +66,9 @@ def handle_espnet_TTS_intelligibility(
         Whisper CER: 6.50
     """
     try:
-        from versa import (
-            espnet_levenshtein_metric,
-            owsm_levenshtein_metric,
-            whisper_levenshtein_metric,
-        )
+        from versa.corpus_metrics.espnet_wer import espnet_levenshtein_metric
+        from versa.corpus_metrics.owsm_wer import owsm_levenshtein_metric
+        from versa.corpus_metrics.whisper_wer import whisper_levenshtein_metric
     except Exception as e:
         print("Error: Versa is not properly installed.")
         raise e
--- a/egs2/TEMPLATE/asr1/pyscripts/utils/dialog_eval/TTS_speech_quality.py
+++ b/egs2/TEMPLATE/asr1/pyscripts/utils/dialog_eval/TTS_speech_quality.py
@@ -46,10 +46,8 @@ def TTS_psuedomos(TTS_audio_output: Tuple[int, np.ndarray]) -> str:
         sheet_ssqa: 4.03
     """
     try:
-        from versa import (
-            pseudo_mos_metric,
-            sheet_ssqa,
-        )
+        from versa.utterance_metrics.pseudo_mos import pseudo_mos_metric
+        from versa.utterance_metrics.sheet_ssqa import sheet_ssqa
     except Exception as e:
         print("Error: Versa is not properly installed.")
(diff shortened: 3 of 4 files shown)


sw005320 (approved):

sw005320:
Thanks!
After CI is finished, it will be merged.

PR #6620, opened 2026-09-02
Cancel the rest of a pull request's run when a fast job fails

sw005320:
## What did you change?

When one of the four **fast, deterministic** jobs fails on a pull request, it now cancels the rest of the run: `test_python_espnet2`, `test_utils_espnet2`, `test_shell_espnet2`, `test_import`.

Measured over the last eight `ci_on_ubuntu` runs that ran to completion — 12,964 minutes of compute in total:

| branch | result | total | after the first failure |
|---|---|---|---|
| master | FAIL | 2732m | **1448m (53%)** |
| ctc-window-margin | FAIL | 1226m | **1187m (97%)** |
| fix/pypi-direct-deps | FAIL | 2631m | **2418m (92%)** |
| five others | pass | ~1200m each | — |

Three of eight failed, and **39% of all compute in the sample ran after the first failure was already known.** The two worst were over 90%, and both failed early — in jobs like these four.

A failure in a fast deterministic job means the pull request is broken rather than unlucky, and the integration matrix behind it is otherwise ~1200 minutes spent on a known answer.

## Two things this deliberately is not

**Not `fail-fast: true`.** That gives up the thing this matrix is for. The grid is python × pytorch *coverage*, so "3.13 failed" and "everything failed" are different answers and only the second is a quick fix. Losing the distinction means fix → re-run → discover the next cell → repeat. It would also have saved 33% here rather than 39%.

**Not every job.** Cancelling from the integration jobs would amplify flakes. Today alone, lxml published a new version two minutes before a job resolved it, with only the sdist up, taking one job down; the same class has hit apt, chocolatey, `torch.hub` and Hugging Face in this repository before. With `fail-fast: false` kept on the long matrix, one flake still leaves every other result standing and exactly one job to re-run.

Worth knowing for the flake case anyway: **`gh run rerun --failed` re-runs cancelled jobs as well as failed ones**, so a cancelled run resumes rather than restarts. Cancelling costs latency, not compute.

**Pull requests only** — on master the full picture is worth the compute.

## Is your PR small enough?

Yes — one file, one step plus a permissions block on four jobs.

## Additional Context

`wget` rather than `curl` or `gh`: the CI image has wget and neither of the others (`docker/ci.dockerfile` installs `unzip wget` in the final stage). The request is `|| echo`'d, so a cancel that cannot happen does not stack a second failure on top of the real one.

**Verified:**

- The wget invocation exactly as written, run against an already-completed run, returns `HTTP/1.1 409 Conflict` — not 400, 401 or 404 — so the method, headers, body and URL are right, and the `||` branch swallows it without failing the step.
- Each of the four jobs carries the step as its **last** one, gated on `failure() && github.event_name == 'pull_request'`; no other job does.
- `python3 ci/check_ci_image_config.py` exits 0; the workflow parses.

**Not verified end to end:** that a job's own `GITHUB_TOKEN` with `actions: write` is accepted for cancelling its own run. That is the combination GitHub documents for this endpoint, but it cannot be exercised without a job that actually fails — the next broken pull request will settle it.

The job-level `permissions` blocks repeat `contents`, `packages` and `pull-requests` read, because a job-level block **replaces** the workflow-level one rather than adding to it; dropping them would have taken away checkout and the image pull.

`.github/workflows/` is not an image-hash input, so this runs on the published image rather than rebuilding it.

### DIFF
--- a/.github/workflows/ci_on_ubuntu.yml
+++ b/.github/workflows/ci_on_ubuntu.yml
@@ -203,6 +203,14 @@ jobs:
 
   test_python_espnet2:
     runs-on: ubuntu-latest
+    # Same grants as the workflow default, plus actions: write, which the
+    # cancel step below needs. A job-level block replaces the workflow one
+    # rather than adding to it, so the read grants are repeated here.
+    permissions:
+      contents: read
+      packages: read
+      pull-requests: read
+      actions: write
     needs: [process_labels, resolve_ci_image]
     if: |
       github.event.pull_request.draft == false
@@ -304,9 +312,44 @@ jobs:
         with:
           path: ~/.cache/s3prl
           key: ${{ runner.os }}-s3prl-v1
+      # This job is fast and deterministic, so a failure here means the pull
+      # request is broken rather than unlucky - and the integration matrix
+      # behind it is then roughly 1200 minutes of compute spent on a known
+      # answer. Measured over eight runs that ran to completion, 39% of all
+      # CI compute ran after the first failure was already known; the two
+      # worst were over 90%, and both failed early, in jobs like this one.
+      #
+      # Only from the fast jobs. The integration matrix keeps fail-fast:
+      # false, so a network flake there still leaves every other result
+      # standing and one job to re-run. `gh run rerun --failed` re-runs
+      # cancelled jobs too, so if this was a flake, cancelling costs
+      # latency rather than compute.
+      #
+      # Pull requests only: on master the full picture is worth the compute.
+      #
+      # wget, not curl or gh: the CI image has wget and neither of those.
+      - name: Cancel the rest of this run
+        if: failure() && github.event_name == 'pull_request'
+        env:
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        run: |
+          wget -q -O- --method=POST --body-data='' \
+            --header="Authorization: Bearer ${GH_TOKEN}" \
+            --header='Accept: application/vnd.github+json' \
+            "${GITHUB_API_URL}/repos/${GITHUB_REPOSITORY}/actions/runs/${GITHUB_RUN_ID}/cancel" \
+            && echo 'cancel requested' \
+            || echo 'cancel request failed; leaving the run alone'
 
   test_utils_espnet2:
     runs-on: ubuntu-latest
+    # Same grants as the workflow default, plus actions: write, which the
+    # cancel step below needs. A job-level block replaces the workflow one
+    # rather than adding to it, so the read grants are repeated here.
+    permissions:
+      contents: read
+      packages: read
+      pull-requests: read
+      actions: write
     needs: [process_labels, resolve_ci_image]
     if: |
       github.event.pull_request.draft == false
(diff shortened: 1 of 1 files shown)


sw005320:
Closing. The waste is real, but every mechanism available to fix it is either ineffective here or not worth the complexity it adds to `ci_on_ubuntu.yml`. Recording the measurements so the next person considering this does not have to redo them.

## The waste is real

Last eight `ci_on_ubuntu` runs that ran to completion, 12,964 minutes of compute:

| branch | result | total | after the first failure |
|---|---|---|---|
| master | FAIL | 2732m | 1448m (53%) |
| ctc-window-margin | FAIL | 1226m | 1187m (97%) |
| fix/pypi-direct-deps | FAIL | 2631m | 2418m (92%) |
| five others | pass | ~1200m each | — |

**39% of all compute in the sample ran after the first failure was already known.**

## Why this PR does not fix it

The permission dump from this PR's own `test_python_espnet2 (3.12, 2.10.0)` job — the job carrying `permissions: actions: write`:

```
GITHUB_TOKEN Permissions
Actions: read
Contents: read
Metadata: read
Packages: read
PullRequests: read
```

`actions: write` was requested and **`Actions: read` was granted**. A `pull_request` from a fork gets a read-only token regardless of the `permissions` block, so the cancel call returns 403, the `|| echo` swallows it, and nothing is cancelled. It works for same-repository branches and never for forks — and this PR comes from one, so the step could not have worked here even once.

That makes it a step that reports success while doing nothing, which is the failure mode this repository produces most often and the one `ci/check_ci_image_config.py` exists to catch.

## Why the simpler alternative does not work either

Making the expensive jobs `needs:` the cheap ones would need no token, no API call and no extra steps. But of the six failing runs with real compute, **`test_import` succeeded in all six**:

| branch | first failure | at | test_import |
|---|---|---|---|
| fix-dialog-eval-versa-imports | `test_python_espnet2 (3.13, 2.10.0)` | 82m | success |
| espnet3/fix/epochsync_reiterable | `test_python_espnet2 (3.13, 2.10.0)` | 77m | success |
| recipe/vggsound-cls1 | `test_python_espnet2 (3.12, 2.11.0)` | 74m | success |
| fix-sds-mini-omni-tempfile-leak | `test_python_espnet2 (3.13, 2.10.0)` | 65m | success |
| remove-dead-torch-version-conditions | `test_python_espnet3 (3.13, 2.11.0)` | 41m | success |
| batch-beam-search-unified | `test_python_espnet2 (3.13, 2.9.1)` | 44m | success |

Gating on `test_import` would have caught none of them. Gating on `test_python_espnet2` means the integration matrix waits up to 103 minutes before it starts, which is worse than the thing it saves.

Job timings from three successful runs, for whoever needs them next:

| job | count | median | max | last completion, from run start |
|---|---|---|---|---|
| `test_import` | 1 | 5m | 6m | 6m |
| `test_shell_espnet2` | 1 | 13m | 22m | 23m |
| `test_utils_espnet2` | 2 | 4m | 17m | 33m |
| `test_configuration_espnet2` | 9 | 15m | 31m | 45m |
| `test_python_espnet2` | 6 | 26m | 56m | 68m |
| `test_integration_espnet2` | 65 | 14m | 52m | 182m |

## The third option, also rejected

A separate workflow on `check_run: completed` would get a writable token even for fork pull requests, and would add nothing to `ci_on_ubuntu.yml`. But `check_run` cannot be filtered by job name in the trigger, so it fires once per job completion — roughly **100 extra workflow runs per CI run**. With the account's 60-job concurrency limit already saturating (733 jobs queued earlier today), that trades compute for queue pressure.

## What actually helps, with no new machinery

Cancelling your own run when you see a failure. Done by hand today on three pull requests, it freed 20 of the 60 slots immediately. That is etiquette rather than a mechanism, so if it is worth encoding it belongs in `CONTRIBUTING` or as one declarative rule in `.mergify.yml` — not as thirty lines in each of four jobs here.

PR #6621, opened 2026-09-03
Follow cuda availability instead of hardcoding `use_gpu=True` in `dialog_eval`

amirlankalm:
**Stacked on #6619**, which has to land first for any of this to be reachable. The
diff against `master` will show that PR's changes too until it merges.

## What did you change?

Every scorer in `dialog_eval` passes `use_gpu=True`: the five setup calls in
`model_cache.py` and the two metric calls in `TTS_speech_quality.py`. versa
asserts on a CUDA build when told to use the GPU, so on a CPU-only install these
raise:

```
```

which means the demo's "TTS Speech Quality", "TTS Intelligibility" and "ASR WER"
modes cannot run at all without a GPU.

One cached helper now reports availability and all eight sites read it:

```python
@lru_cache(maxsize=None)
def use_gpu() -> bool:
    import torch

    return torch.cuda.is_available()
```

So a GPU is used when there is one and skipped when there is not. On a GPU box
behaviour is unchanged.

Verified on a CPU-only machine, which was not possible before either fix, since
the imports failed first:

```
use_gpu() = False
predictors loaded via model_cache: ['dnsmos', 'plcmos', 'utmos']
utmos 1.748   dns_overall 1.079   dns_p808 2.108   plcmos 3.949
```

That is 3 of the 4 metrics in `TTS_psuedomos`.

## Is your PR small enough?

Yes. 2 files, 8 call sites plus one helper.

## Additional context

**Not verified locally: SHEET SSQA.** It additionally needs `s3prl`, and
`ci/install.sh` flags that as the forked build which breaks CI, so I did not
install it. That path rests on the same one-line change as the others but I have
not run it.

If you would rather this were an explicit setting than an auto-detect, say so
and I will take it from a config or an environment variable instead. Auto-detect
seemed closest to the existing behaviour, which was simply assuming a GPU.

Environment:

```
- python version: 3.12.8
```

### DIFF
--- a/egs2/TEMPLATE/asr1/pyscripts/utils/dialog_eval/ASR_WER.py
+++ b/egs2/TEMPLATE/asr1/pyscripts/utils/dialog_eval/ASR_WER.py
@@ -62,11 +62,9 @@ def handle_espnet_ASR_WER(
          Whisper CER: 6.50"
     """
     try:
-        from versa import (
-            espnet_levenshtein_metric,
-            owsm_levenshtein_metric,
-            whisper_levenshtein_metric,
-        )
+        from versa.corpus_metrics.espnet_wer import espnet_levenshtein_metric
+        from versa.corpus_metrics.owsm_wer import owsm_levenshtein_metric
+        from versa.corpus_metrics.whisper_wer import whisper_levenshtein_metric
     except Exception as e:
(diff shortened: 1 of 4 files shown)


sw005320 (approved):
Thanks!

PR #6623, opened 2026-09-03
Use safe_torch_load in handle_finetuned_checkpoint

sw005320:
## What did you change?

`handle_finetuned_checkpoint` in `espnet2/beats/generate_beats_checkpoint.py` now loads through `safe_torch_load` instead of `torch.load(..., weights_only=False)`.

The same file already imports `safe_torch_load` and uses it in `average_checkpoints`, but this one call site was left behind. The path it loads comes from the config's `encoder_conf.beats_ckpt_path`, and the function is reachable from this module's own CLI, so a BEATs pretrained checkpoint obtained from a third party was being deserialized through the unrestricted pickle unpickler.

`safe_torch_load` tries `weights_only=True` first and refuses to fall back without an explicit opt-in, which is the behavior every other checkpoint load in espnet2 already has.

This was the last explicit `weights_only=False` in `espnet2/` and `espnet3/` outside `safe_torch_load` itself.

### Behavior change to be aware of

BEATs checkpoints carry a non-tensor `cfg` entry, so `weights_only=True` will not succeed on them. Loading one now requires `ESPNET_ALLOW_UNSAFE_TORCH_LOAD=1`, or the interactive confirmation on a TTY. That is the same opt-in `average_checkpoints` in this file has already required since it was converted, so the two paths in the module now behave consistently rather than one silently accepting what the other refuses.

## Is your PR small enough?

Yes. Two files, 15 lines added.

## Additional Context

Adds the matching regression test. The file already had the equivalent pair for `average_checkpoints`, so this follows that pattern:

```
test_average_checkpoints_requires_explicit_opt_in
test_average_checkpoints_accepts_env_opt_in
test_handle_finetuned_checkpoint_requires_explicit_opt_in   <- new
```

`black` and `flake8` are clean on both files.

### DIFF
--- a/espnet2/beats/generate_beats_checkpoint.py
+++ b/espnet2/beats/generate_beats_checkpoint.py
@@ -157,7 +157,7 @@ def handle_finetuned_checkpoint(checkpoint, config):
         raise FileNotFoundError(
             f"Beats pretrained checkpoint {beats_pt_ckpt_path} does not exist."
         )
-    pt_ckpt = torch.load(beats_pt_ckpt_path, map_location="cpu", weights_only=False)
+    pt_ckpt = safe_torch_load(beats_pt_ckpt_path, map_location="cpu")
     if "cfg" not in pt_ckpt:
         raise ValueError(
             f"Pretrained checkpoint {beats_pt_ckpt_path} does not contain 'cfg' key."
--- a/test/espnet2/beats/test_generate_beats_checkpoint.py
+++ b/test/espnet2/beats/test_generate_beats_checkpoint.py
@@ -1,6 +1,7 @@
 import pytest
 import torch
 
+import espnet2.torch_utils.safe_torch_load as safe_torch_load_module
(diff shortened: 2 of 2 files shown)


sw005320:
Thanks @coderabbitai — valid, and it applied to the pre-existing test too, so I fixed both in 80b31ab.

You are right that `monkeypatch.delenv` only closes one of the two doors. `safe_torch_load` refuses on `not (env_opt_in or _confirm_unsafe_interactively(path))`, and that second call returns early only when `sys.stdin.isatty()` is false. Pytest's default capture makes it false in CI, which is why this was passing, but under `pytest -s` on a terminal the test would have blocked on `input()` rather than asserting anything.

Both refusal tests now go through a shared `_no_opt_in` helper that patches `_confirm_unsafe_interactively` to return `False` alongside deleting the env var. I extended it to `test_average_checkpoints_requires_explicit_opt_in` as well, since that one had the identical gap and leaving it would be odd.

PR #6624, opened 2026-09-03
Project the encoder output once per utterance in the transformer decoder

sw005320:
Follow-up to #6599. Independent of #6610, but measured with its CTC window on, since that is the configuration that will ship and it is where the decoder becomes the larger share of a step.

## What did you change?

The transformer decoder's source attention projected the encoder output through `linear_k` and `linear_v` at **every decoding step, in every layer, for every hypothesis**, although that tensor never changes during a decode. With the beam replicated over the batch that was the largest single matmul of a step: `n_utt * beam` rows of `T` frames, six layers deep, once per output token.

- `MultiHeadedAttention.project_kv(key, value)` returns exactly the `(k, v)` that `forward_qkv` would, and `forward(..., kv=(k, v))` attends with a pair computed earlier. The pair may have one row per query, or one row per *group* of consecutive queries -- the utterance-major layout of `BatchBeamSearch`, where hypothesis `b * beam + i` belongs to utterance `b`. In the grouped case the `beam` queries of an utterance are folded into the query-time axis and attend to a single copy of the projections, so nothing is replicated over the beam. It is the same matmul in a different blocking; the frozen-reference beam search tests of #6599 still pass at `rtol=atol=0`.
- `BaseTransformerDecoder.memory_kv(memory)` computes the per-layer projections once and caches them while the same encoder output is being decoded. The cache is keyed on the identity of the tensor that owns the storage plus its version counter, and a weak reference drops it the moment that tensor is garbage collected. A later utterance whose encoder output lands at the same address therefore cannot see stale projections, and an in-place edit invalidates them. A stride-0 memory -- a `(T, D)` output expanded over the hypotheses, which is what the single-utterance search hands over -- is projected once from its first row. Both `score` (plain `BeamSearch`) and `batch_score` use it, so the non-batch fallback benefits too.
- `BatchBeamSearch.score_full` hands a decoder that declares `accepts_shared_memory` a strided `(n_utt, T, D)` view instead of the replicated `(n_utt * beam, T, D)` tensor. Every other scorer keeps the replicated layout, as does `TransformerMDDecoder`, which has its own `forward_one_step`. `DecoderLayer.forward` takes an optional `memory_kv` and only forwards it when set, so the other users of the layer (`mlm_decoder`, the legacy decoder) are untouched.

No public signature changes: the new arguments are keyword-only with defaults, and `batch_score` still accepts the old `(n_batch, T, D)` memory.

It removes work that is provably redundant, and the amount removed grows with `beam * T * D^2`, which is exactly where decoding is slowest. What it buys in wall clock depends on what else is in the step.

**Projection count**, LibriSpeech E-Branchformer, four utterances, hook on `linear_k` of the first layer's source attention:

| | before | after |
|---|---:|---:|
| batch_size 1 (109 decoder calls) | 109 | 4 |
| batch_size 4 (37 decoder calls) | 37 | 1 |

**Where the time went**, same four utterances on MPS, timers with a device synchronisation around each call:

| | before | after |
|---|---:|---:|
| source attention, batch_size 1 | 3.7 s | 1.4 s |
| source attention, batch_size 4 | 2.4 s | 0.8 s |
| decoder `batch_score`, batch_size 1 | 9.3 s | 6.3 s |
| decoder `batch_score`, batch_size 4 | 4.4 s | 2.7 s |

**End to end**, trained models on real audio, beam 10, CTC 0.3, `--ctc_window_margin 100` (#6610) on both sides. MPS numbers are medians of three interleaved runs, because this machine drifts by up to 20% with temperature; a baseline run followed by a branch run is not a fair comparison here. Transcripts identical to the baseline in every run.

| | baseline | this PR | |
|---|---:|---:|---:|
| LibriSpeech test-clean, 8 utts, MPS, batch_size 1 | 23.6 s | 24.7 s | 0.95x |
| LibriSpeech test-clean, 8 utts, MPS, batch_size 4 | 14.1 s | 12.7 s | 1.11x |
| AISHELL-1 test, 25 utts, MPS, batch_size 1 | 46.1 s | 48.0 s | 0.96x |
| AISHELL-1 test, 25 utts, MPS, batch_size 4 | 24.0 s | 22.1 s | 1.09x |
| LibriSpeech test-clean, 8 utts, CPU, batch_size 1 | 27.4 s | 25.2 s | 1.09x |
| LibriSpeech test-clean, 8 utts, CPU, batch_size 4 | 22.8 s | 17.2 s | 1.33x |

Read it this way. At `batch_size 1` on this laptop's GPU a decoding step is bound by kernel launches -- the windowed CTC recursion alone is a few hundred of them -- and the two projections it saves per layer are a few percent of the step, so the result is a wash within noise. At `batch_size 4` the projections are four times larger and the saving shows, 1.1x on MPS and 1.33x on CPU, where the FLOPs count. On a server GPU with `--batch_size 8` or more, a larger beam, or long-form audio, the share this removes is larger still: on a synthetic 8 x 800-frame batch profiled while preparing #6610 the decoder was 75% of the step and this projection most of it. I could not measure that configuration here.

What this does not yet do: the other scorers still receive the replicated encoder output, so the `(n_utt * beam, T, D)` tensor is still materialized. Dropping it once every full scorer accepts the shared layout is a follow-up; that is where the ~250 MB per unit of batch size measured in #6599 would go.

## Is your PR small enough?

4 library files, +269 / -17; 4 test files, +248. Yes.

## Additional Context

**Testing**

- `test/espnet2/legacy/test_attention.py` (new): `project_kv` equals `forward_qkv` bit for bit; attending with `kv` equals attending without it, with and without a mask, `time1` of 1 and 3, SDPA and default path; grouped projections equal replicated ones with no mask, a per-query mask and a per-group mask; a per-position `(batch, time1, time2)` mask falls back to expansion; an indivisible group count raises.
- `test_transformer_decoder.py`: the decoder projects once per memory across calls; recomputes after an in-place edit and for another tensor of the same shape; the cache is gone after the memory is garbage collected; a shared `(n_utt, T, D)` memory scores like the replicated one with and without a padding mask, across two steps; an expanded memory is projected once from one row regardless of how many hypotheses it is expanded over, and `score` reuses the same projections.
- `test_batch_beam_search_batched.py`: the beam search hands the decoder one encoder row per utterance and the LM one per hypothesis.
- `test_batch_beam_search_memory.py`: the vectorization guard now expects `n_utt` memory rows for the decoder; the hypothesis count it checks is unchanged.
- The frozen-reference tests from #6599 (221 configurations at `rtol=atol=0`), the batched parity tests, the online, time-sync, partially-AR and `Speech2Text` tests pass unchanged. Across every `test/espnet2` file that touches attention, decoders or beam search, the set of failing tests is identical before and after (all missing optional dependencies on this machine).
- Real-model check: LibriSpeech E-Branchformer and AISHELL conformer, 16 and 50 utterances, batch sizes 1 and 4, transcripts identical to master with and without the CTC window.

**Design notes**

- Why a cache inside the decoder rather than a new argument through `batch_score`: the streaming subclasses (`BatchBeamSearchOnline`, `BatchBeamSearchOnlineSim`) override `score_full` with the old signature and would bypass any new argument. Keyed on the memory tensor, they get the reuse without changes and cannot get a stale result.
- Why a weak reference rather than an explicit reset: `BatchBeamSearchOnline` never calls `batch_init_state` on the decoder, so no hook at the start of a decode is reached by every caller. Tying the cache lifetime to the tensor's lifetime needs no cooperation from the caller.
- `self.attn` keeps its `(batch, head, time1, time2)` layout in the grouped case, so anything reading attention weights after a step sees what it saw before.

### DIFF
--- a/espnet2/asr/decoder/transformer_decoder.py
+++ b/espnet2/asr/decoder/transformer_decoder.py
@@ -4,7 +4,8 @@
 """Decoder definition."""
 
 import logging
-from typing import Any, List, Sequence, Tuple
+import weakref
+from typing import Any, List, Optional, Sequence, Tuple
 
 import torch
 from typeguard import typechecked
@@ -188,6 +189,65 @@ class BaseTransformerDecoder(
             return (x, intermediate_outs), olens
         return x, olens
 
+    # `batch_score` accepts an encoder output with one row per utterance
+    # rather than one per hypothesis (utterance-major, see
+    # `BatchBeamSearch.score_full`), so the beam search need not replicate it.
+    accepts_shared_memory = True
+
+    def memory_kv(
+        self, memory: torch.Tensor
+    ) -> List[Tuple[torch.Tensor, torch.Tensor]]:
+        """Return the per-layer key/value projections of ``memory``, cached.
+
+        The encoder output does not change while an utterance is decoded, yet
+        :meth:`forward_one_step` used to project it again at every step, in
+        every layer, for every hypothesis. The projections are computed once
+        here and reused for as long as the same tensor is being decoded.
+
+        The cache is keyed on the identity of the tensor that owns the storage
+        (``memory._base`` for a view) together with its version counter, and
+        it is dropped the moment that tensor is garbage collected. A later
+        utterance whose encoder output happens to land at the same address can
+        therefore never see stale projections, and an in-place modification
+        invalidates them too.
+
+        A memory whose rows are stride-0 copies of a single row -- the way the
+        single-utterance beam search expands a ``(T, D)`` encoder output over
+        its hypotheses -- is projected once, and the attention broadcasts the
+        result over the queries.
+
+        Args:
+            memory: Encoder output ``(n_rows, T, D)``.
+
+        Returns:
+            One ``(k, v)`` pair per decoder layer, each
+            ``(n_rows or 1, n_head, T, d_k)``.
+
+        """
+        base = memory._base if memory._base is not None else memory
+        shared_rows = memory.dim() == 3 and memory.size(0) > 1 and memory.stride(0) == 0
+        src = memory[:1] if shared_rows else memory
+        key = (
+            base._version,
+            tuple(src.shape),
+            # the stride of a size-1 leading axis is arbitrary and changes with
+            # the number of hypotheses the memory was expanded over
+            src.stride()[1:] if src.size(0) == 1 else src.stride(),
+            src.storage_offset(),
+            src.dtype,
+            src.device,
+        )
+        cached = getattr(self, "_memory_kv_cache", None)
+        if cached is not None and cached[0] == key and cached[1]() is base:
+            return cached[2]
+
+        kv = [layer.src_attn.project_kv(src, src) for layer in self.decoders]
+        # NOTE: a weakref, not a strong one: the cache must not keep the
+        # encoder output alive, and it dies with it
+        ref = weakref.ref(base, lambda _: setattr(self, "_memory_kv_cache", None))
+        self._memory_kv_cache = (key, ref, kv)
+        return kv
+
     def forward_one_step(
         self,
         tgt: torch.Tensor,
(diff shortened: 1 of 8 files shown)


sw005320:
Closing for now: on the hardware I can measure, the saving is a few percent of a batch_size=1 step, which is bound by kernel launches from the CTC recursion rather than by these matmuls. It only shows at batch_size>=4 (1.1x MPS, 1.33x CPU). Keeping the branch; will revisit together with a fix for the CTC recursion, which is where the time goes.

PR #6631, opened 2026-09-04
Add automatic Claude Code review for pull requests

jctian98:
## What did you change?

- **New workflow `.github/workflows/claude-code-review.yml`**: runs the official `code-review` plugin of `anthropics/claude-code-action` on pull requests and posts findings as inline review comments.
- **`.github/workflows/claude.yml`**: restore the `issue_comment` trigger so that `@claude ...` written in a PR conversation works (see "Why" for the reason it never did).
- **New `CLAUDE.md`**: repository layout, everyday commands, and the coding / recipe / test / PR conventions taken from `CONTRIBUTING.md`, `setup.cfg`, `pyproject.toml` and the `ci/` scripts. The review plugin checks PRs against it; it also serves anyone using Claude Code on the repo.
- Both workflows authenticate with `CLAUDE_CODE_OAUTH_TOKEN`, a one-year token generated with `claude setup-token` from a maintainer's Claude subscription (repository secret added 2026-09-04). Runs therefore draw on that subscription's usage limits instead of API billing.
- A `claude-review` label was created on the repository (no file change).

### Who gets reviewed

| PR author | When the review runs |
|---|---|
| Owner, org member, collaborator | Automatically on open, reopen, and when marked ready for review |
| Anyone else (incl. first-time contributors) | When a maintainer adds the `claude-review` label |
| Any PR | Manually: Actions tab, "Claude Code Review", "Run workflow", enter the PR number |

Draft PRs and bot-authored PRs (dependabot, Copilot) are skipped. Follow-up questions to a review go through `@claude ...` comments handled by `claude.yml`.

---

Two findings from the run history motivated the design (the `ANTHROPIC_API_KEY` mentioned below is the old, now unused secret):

1. **The managed "Claude Code Review" service was enabled on this repo from 2026-03-28 to 2026-06-07.** All 39 things it posted were the same notice: "This pull request is from a fork — automated review is disabled." It never reviewed a single PR, because every PR it saw came from a fork.
2. **`claude.yml` has never had a successful run** (0 successes, 39 failures since March). GitHub withholds secrets from `pull_request`, `pull_request_review` and `pull_request_review_comment` runs of fork PRs, so the action could obtain neither `ANTHROPIC_API_KEY` nor an OIDC token (`Unable to get ACTIONS_ID_TOKEN_REQUEST_URL`). `issue_comment`, the trigger that *does* receive secrets for fork PRs, was commented out. 26 of the last 30 PRs here came from forks, including maintainers' own.

The standard review workflow from the Claude Code docs triggers on `pull_request`, which would therefore skip almost every ESPnet PR. This PR uses `pull_request_target` instead, following the action's own security guidance (https://github.com/anthropics/claude-code-action/blob/main/docs/security.md):

- The workflow file and the checkout come from `master`; **the PR's code is never checked out or executed**. The review reads the diff through the GitHub API.
- `GITHUB_TOKEN` permissions are read-only except `pull-requests: write` for posting the review.
- Claude's tools are limited to the inline-comment tool and the read-only `gh pr view` / `gh pr diff` commands declared by the skill. No shell beyond that.
- `allowed_non_write_users: "*"` is required for the action to run on PRs whose author has no write access; access control is done by the job-level `if:` (member / label gate), not by GitHub write permission.
- Residual risk is prompt injection through the diff text, which can at worst produce a misleading comment or spend API budget. The label gate keeps unknown contributors from triggering runs on their own.

---

## Is your PR small enough?

yes (3 files, +242 / -2)

---

## Additional Context

- **This workflow only takes effect after merge**: `pull_request_target` runs the version of the file on `master`, so CI on this PR cannot exercise it. The first real test is the next PR opened after merging, or a manual run from the Actions tab with any open PR number.
- **Credential**: the workflows no longer reference `ANTHROPIC_API_KEY`; that secret (added 2026-03-28, never used by a successful run) can be deleted. To move to API billing later, add a Console key with a spend cap as `ANTHROPIC_API_KEY` and swap the credential line, as the comment in the workflow describes. The subscription token expires after one year and then has to be regenerated with `claude setup-token`.
- **Usage**: the plugin runs a handful of model calls per review (triage, summary, two compliance agents, two bug agents, verification), all counted against the subscription's usage limits. Reviews are not repeated on every push; the plugin stops when Claude has already commented on a PR.
- **Tuning knobs**: change `types:` to add `synchronize` for review-on-every-push; edit the `author_association` list or the label name in the `if:`; add `--model ...` to `claude_args`.
- `@claude` mentions still require the commenter to have write access (default of the action).
- Docs: https://code.claude.com/docs/en/github-actions


sw005320:
@jctian98, [coderabbitai](https://github.com/apps/coderabbitai) is very good in general.
Do you see additional benefits of using Claude Code reviews?
My concern is using the same backbone, which may results in less diverse reviews.

PR #6633, opened 2026-09-04
Derive Conv*Subsampling masks from per-utterance lengths

### LINKED ISSUE #6600: Encoder output length depends on the longest utterance in the batch (Conv*Subsampling mask slicing)
## Summary

An encoder built on `Conv1dSubsampling*` / `Conv2dSubsampling*` returns a different `olens` for the same utterance depending on **which other utterances share its batch**. The padding mask is subsampled by slicing relative to the padded width of the batch, rather than as a function of each utterance's own length.

I ran into this while adding batched beam search (#6599): decoding a minibatch cannot reproduce `--batch_size 1` because the encoder itself disagrees about how long the encoder output is. Filing it separately because it is pre-existing and independent of that PR.

## Reproduction

No data or trained model needed:

```python
import torch
from espnet2.asr.encoder.transformer_encoder import TransformerEncoder

torch.manual_seed(0)
enc = TransformerEncoder(
    input_size=8, output_size=4, attention_heads=2, linear_units=4,
    num_blocks=1, input_layer="conv1d2",
).eval()

short, long_ = 288, 363
feats = torch.randn(2, long_, 8)
with torch.no_grad():
    _, olens_alone, _ = enc(feats[1:2, :short], torch.tensor([short]))
    _, olens_batch, _ = enc(feats, torch.tensor([long_, short]))
    _, olens_pair, _ = enc(feats[:, :short], torch.tensor([short, short]))

print(olens_alone.tolist(), olens_batch.tolist(), olens_pair.tolist())
```

```
feats_lengths = 288 alone                     -> [142]
feats_lengths = [363, 288] batched together   -> [180, 144]
feats_lengths = [288, 288] batched together   -> [142, 142]
```

The 288-frame utterance is 142 frames long on its own and 144 frames long next to a 363-frame one.

## Why

`Conv1dSubsampling2.forward` ([subsampling.py#L249](https://github.com/espnet/espnet/blob/master/espnet2/legacy/nets/pytorch_backend/transformer/subsampling.py#L249)):

```python
x = self.conv(x)                                   # Conv1d(k=3,s=1) then Conv1d(k=3,s=2)
...
x_mask = x_mask[:, :, :-2:1][:, :, :-2:2]
```

The two slices mirror the two convolutions, but they trim from **the end of the tensor**, and the tensor is as wide as the longest utterance in the batch. `make_pad_mask` builds the mask at width `max(ilens)`, so what gets trimmed depends on the batch:

| | mask width | after `[:-2:1]` | after `[:-2:2]` | `True` count for the 288-frame row |
| --- | ---: | ---: | ---: | ---: |
| alone | 288 | 286 | 142 | **142** — the last two real frames are trimmed away |
| next to a 363-frame utterance | 363 | 361 | 180 | **144** — only padding is trimmed |

So `olens` is not a function of `ilens` alone. For the shortest utterance in a batch the trim lands on padding; for an utterance decoded alone it lands on real frames.

Note which side is "wrong": the extra two frames in the batched case are **real audio** that the single-utterance path discards. Batching is arguably the more correct of the two.

The same pattern appears in every subsampling class in that file: `Conv1dSubsampling1/2/3`, `Conv2dSubsampling`, `Conv2dSubsampling1/2/6/8`.

## Consequences

- **Inference.** With `--batch_size > 1` (now possible for the attention/CTC beam search via #6599) an utterance decodes differently than it would on its own, and differently again depending on how the batch is bucketed. Not a large difference in practice, but not reproducible either.
- **Training.** The same utterance contributes a different number of encoder frames depending on its batch, so the CTC loss denominator and the attention context change with batch composition. Bucketing by length (`batch_type: sorted`/`folded`) makes this mostly consistent, which is probably why it has gone unnoticed.
- **CTC alignment / segmentation** tools that batch will inherit the same drift.

## What a fix would involve

Making the mask trim per-utterance, e.g. deriving `olens` from `ilens` with the same formula the convolutions use — for `Conv1dSubsampling2`, `((ilens - 2) - 2) // 2 + 1` — instead of slicing the padded mask.

This is not a free change:

- It alters the encoder output length of existing models, so decoding results move for every recipe using a conv subsampling front-end. Published results and released models would need re-checking.
- The per-utterance formula agrees with today's behaviour for the **longest** utterance in a batch, and disagrees by one or two frames for shorter ones, so the effect on any given recipe depends on its batching.

Given that, I did not touch it in #6599 — I documented the consequence there instead. Happy to prepare the fix if you think the correction is worth the churn, or to leave it documented if not.

## Environment

espnet master (`ec5632b16`), torch 2.13.0, CPU.

modelpath-dev:
## What did you change?

`Conv*Subsampling` used to shrink the pad mask by slicing from the end of the padded batch (`x_mask[:, :, :-2:1][:, :, :-2:2]`, and the same pattern in the other seven classes). That makes `olens` for a given utterance depend on the longest row.

The mask is now rebuilt from each row's own input length using the convolution length formula. All eight classes go through one helper. Encoders already do `olens = masks.sum(1)`, so they pick this up without further changes.

The longest utterance in a batch is unchanged; that path already matched the formula. A shorter one now matches the length it has when decoded alone.

Filed as #6600 after #6599. With batched beam search, `--batch_size > 1` cannot reproduce `--batch_size 1` because the encoder itself disagrees about how long the output is.

The issue's snippet, `input_layer="conv1d2"`:

| feats_lengths | olens before | olens after |
|---|---|---|
| 288 alone | [142] | [142] |
| [363, 288] | [180, 144] | [180, 142] |
| [288, 288] | [142, 142] | [142, 142] |

This will move training and decode for shorter utterances in a mixed-length batch by a frame or two. Length-bucketed recipes (`batch_type: sorted` / `folded`) barely see it. Published numbers from `--batch_size 1` are unaffected. I think that is the right trade: `olens` is a function of `ilens` again.

Fixes #6600.

## Is your PR small enough?

Yes. Two files.

## Additional Context

`test/espnet2/legacy/test_subsampling.py` now checks, for every Conv* class, that a 288-frame utterance has the same `olens` alone, next to a 363-frame one, and next to a copy of itself. The `Conv1dSubsampling2` case pins the numbers from the issue. A second test runs the encoder snippet from #6600.

`pytest test/espnet2/legacy/test_subsampling.py`

### DIFF
--- a/espnet2/legacy/nets/pytorch_backend/transformer/subsampling.py
+++ b/espnet2/legacy/nets/pytorch_backend/transformer/subsampling.py
@@ -49,6 +49,32 @@ def check_short_utt(ins, size):
     return False, -1
 
 
+def _conv_out_length(length, kernel_size, stride, padding=0, dilation=1):
+    """Time-axis output length of a convolution (PyTorch Conv1d/Conv2d formula)."""
+    return (length + 2 * padding - dilation * (kernel_size - 1) - 1) // stride + 1
+
+
+def _subsample_mask(x_mask, out_time, convs):
+    """Rebuild the pad mask from each utterance's own input length.
+
+    ``convs`` is a sequence of ``(kernel_size, stride)`` matching the time-axis
+    convolutions of the module. Slicing the padded batch
+    (``x_mask[:, :, :-k:s]``) trims relative to ``max(ilens)``, so a short
+    utterance reports a different ``olens`` depending on its batch mates
+    (issue 6600).
+
+    """
+    if x_mask is None:
+        return None
+    olens = x_mask.squeeze(1).sum(dim=-1)
+    for kernel_size, stride in convs:
+        olens = _conv_out_length(olens, kernel_size, stride)
+    olens = olens.clamp(min=0, max=out_time)
+    frames = torch.arange(out_time, device=x_mask.device)
+    mask = frames.unsqueeze(0) < olens.unsqueeze(1)
+    return mask.to(dtype=x_mask.dtype).unsqueeze(1)
+
+
 def _upgrade_legacy_subsampling_state_dict(state_dict, prefix):
     """Remap legacy nn.Sequential keys for subsampling modules."""
     w_new = prefix + "out.weight"
@@ -144,7 +170,7 @@ class Conv1dSubsampling1(torch.nn.Module):
         b, c, t = x.size()
         x = self.out(x.transpose(1, 2).contiguous())
         if x_mask is not None:
-            x_mask = x_mask[:, :, :-2:1][:, :, :-2:1]
+            x_mask = _subsample_mask(x_mask, x.size(1), ((3, 1), (3, 1)))
 
         if prefix_embeds is not None:
             x = torch.cat([prefix_embeds, x], dim=1)
@@ -246,7 +272,7 @@ class Conv1dSubsampling2(torch.nn.Module):
         b, c, t = x.size()
         x = self.out(x.transpose(1, 2).contiguous())
         if x_mask is not None:
-            x_mask = x_mask[:, :, :-2:1][:, :, :-2:2]
+            x_mask = _subsample_mask(x_mask, x.size(1), ((3, 1), (3, 2)))
 
         if prefix_embeds is not None:
             x = torch.cat([prefix_embeds, x], dim=1)
@@ -348,7 +374,7 @@ class Conv1dSubsampling3(torch.nn.Module):
         b, c, t = x.size()
         x = self.out(x.transpose(1, 2).contiguous())
         if x_mask is not None:
-            x_mask = x_mask[:, :, :-2:1][:, :, :-4:3]
+            x_mask = _subsample_mask(x_mask, x.size(1), ((3, 1), (5, 3)))
 
         if prefix_embeds is not None:
             x = torch.cat([prefix_embeds, x], dim=1)
@@ -450,7 +476,7 @@ class Conv2dSubsampling(torch.nn.Module):
         b, c, t, f = x.size()
         x = self.out(x.transpose(1, 2).contiguous().view(b, t, c * f))
         if x_mask is not None:
-            x_mask = x_mask[:, :, :-2:2][:, :, :-2:2]
+            x_mask = _subsample_mask(x_mask, x.size(1), ((3, 2), (3, 2)))
 
         if prefix_embeds is not None:
             x = torch.cat([prefix_embeds, x], dim=1)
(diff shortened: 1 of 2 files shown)


sw005320 (commented):
Thanks @modelpath-dev — the diagnosis is right and the fix is the correct shape: rebuilding the mask from each row's own length instead of slicing the padded batch makes `olens` a function of `ilens` again, and routing all eight classes through one helper is better than eight near-identical edits. The test that checks 288-alone against 288-next-to-363 against 288-next-to-288 is exactly the right assertion. I also agree with the trade-off call in the description — a frame or two of movement for short utterances in mixed-length batches is worth `olens` being well-defined.

**There is a ninth class with the same bug, in a sibling file, and it is live.**

`espnet2/legacy/nets/pytorch_backend/transformer/subsampling_without_posenc.py:60-61`:

```python
for k, s in zip(self.kernels, self.strides):
    x_mask = x_mask[:, :, : -k + 1 : s]
```

Same construction: the slice bound is relative to the padded batch length, so the surviving count depends on the longest row. For `kernels=[3, 3], strides=[2, 2]` this expands to `x_mask[:, :, :-2:2][:, :, :-2:2]` — character-for-character what `Conv2dSubsampling` does at line 453 of the file you are fixing.

I reproduced it with the three configurations the encoders actually construct:

| `input_layer` | kernels / strides | 288 alone | 288 padded to 363 | conv formula |
| --- | --- | --- | --- | --- |
| `conv2d` | `[3,3]` / `[2,2]` | 71 | **72** | 71 |
| `conv2d6` | `[3,5]` / `[2,3]` | 47 | **48** | 47 |
| `conv2d8` | `[3,3,3]` / `[2,2,2]` | 35 | **36** | 35 |

Same one-frame overshoot as the `conv1d2` case in #6600.

This is not dead code. `Conv2dSubsamplingWOPosEnc` is the embed layer for both streaming encoders:

- `espnet2/asr/encoder/contextual_block_transformer_encoder.py:108,113,118`
- `espnet2/asr/encoder/contextual_block_conformer_encoder.py:113,118,123`

and both derive lengths the way your description assumes — `olens = masks.squeeze(1).sum(1)`, at `contextual_block_transformer_encoder.py:251,365` and `contextual_block_conformer_encoder.py:269,383`. So they inherit the bug today and would inherit the fix for free, exactly like the encoders you already covered.

Worth folding into this PR rather than a follow-up, I think. The class takes `kernels` and `strides` as constructor arguments rather than hard-coding them, so it should drop straight into the same helper — arguably more cleanly than the eight fixed-geometry classes, since the loop already has `(k, s)` in hand. A test row for at least the `conv2d` config would keep it honest.

**Also checked and clean:** `RNNP` in `espnet2/legacy/nets/pytorch_backend/rnn/encoders.py:85` computes `ilens = [int(i + 1) // sub for i in ilens]` per row, so it never depended on the batch maximum. A repository-wide search for stride-slicing of masks (`[:, :, :-N:M]`) turns up only `subsampling.py` and `subsampling_without_posenc.py`, so with that second file this should be the complete set.

modelpath-dev:
@sw005320 
Shinji, you merged 6633 and asked for the ninth class in subsampling_without_posenc.py. I'll land that next. The 6660 table already showed olens was necessary and not sufficient.

modelpath-dev:
@sw005320  Ninth class is up in [https://github.com/espnet/espnet/pull/6670](url) 
Same helper, with the conv2d / conv2d6 / conv2d8 rows from your review.

PR #6637, opened 2026-09-07
[Phonological Tokenizer 1/7] Add differentiable k-means and tokenizer class

ondatk68:
## What did you change?

This is 1/7 of a series of PRs related to the Phonological Tokenizer. 
This PR covers the implementation of differentiable k-means and tokenizer.

**Implementation files**

| File | Role |
|---|---|
| `espnet2/tok/output.py` | Defines the tokenizer output.|
| `espnet2/tok/quantizer/abs_quantizer.py` | Defines `AbsSpeechTokenizerQuantizer`to support potential discretization methods other than differentiable k-means in the future. |
| `espnet2/tok/quantizer/differentiable_kmeans.py` | Implements `DifferentiableKMeans`. |
| `espnet2/tok/tokenizer.py` | Defines a `SpeechTokenizer` consisting of a frontend and a quantizer. The frontend is expected to use S3PRL specified through the ASR frontend, while the quantizer is expected to use differentiable k-means. Also supports freezing the tokenizer during the early stage of training, where only the downstream ASR and vocoder are trained.|

**Test files**

| File | Role |
|---|---|
| `test/espnet2/tok/quantizer/test_differentiable_kmeans.py` | Tests centroid loading, nearest-centroid encoding, padding masks, hard assignments and gradient propagation, feature-dimension validation, temperature behavior, and both distance types. |
| `test/espnet2/tok/test_tokenizer.py` | Tests frontend/quantizer composition and the transition from frozen tokenization to trainable tokenization. |

**Package initialization files**

| File | Role |
|---|---|
| `espnet2/tok/__init__.py` | Declares the speech tokenizer package. |
| `espnet2/tok/quantizer/__init__.py` | Declares the speech quantizer package. |
| `test/espnet2/tok/__init__.py` | Declares the speech tokenizer test package. |
| `test/espnet2/tok/quantizer/__init__.py` | Declares the quantizer test package. |

---

Joint optimization of a speech frontend and downstream objectives requires discrete assignments that preserve a gradient path. 

---

## Is your PR small enough?

yes

---

### DIFF
--- a/espnet2/tok/__init__.py
+++ b/espnet2/tok/__init__.py
@@ -1,0 +1,1 @@
+"""Trainable speech tokenizers and their auxiliary objectives."""
--- a/espnet2/tok/output.py
+++ b/espnet2/tok/output.py
@@ -1,0 +1,28 @@
+"""Shared output types for trainable speech tokenizers."""
+
+from dataclasses import dataclass
+from typing import Optional
+
+import torch
+
+
+@dataclass
+class SpeechTokenizerOutput:
+    """Outputs of a trainable speech tokenizer.
+
+    Args:
+        continuous: Continuous frontend features of shape ``(B, T, D)``.
+        assignment: Hard straight-through cluster assignments of shape
+            ``(B, T, K)``.  The forward values are one-hot during training,
+            while gradients propagate through their soft counterparts.
+        token_ids: Integer cluster indices of shape ``(B, T)``.
+        lengths: Valid token lengths of shape ``(B,)``.
+        soft_assignment: Optional soft cluster assignments of shape
+            ``(B, T, K)`` for analysis and auxiliary regularization.
+    """
+
+    continuous: torch.Tensor
+    assignment: torch.Tensor
+    token_ids: torch.Tensor
+    lengths: torch.Tensor
+    soft_assignment: Optional[torch.Tensor] = None
--- a/espnet2/tok/quantizer/__init__.py
+++ b/espnet2/tok/quantizer/__init__.py
@@ -1,0 +1,1 @@
+"""Quantizers for trainable speech tokenization."""
--- a/espnet2/tok/quantizer/abs_quantizer.py
+++ b/espnet2/tok/quantizer/abs_quantizer.py
@@ -1,0 +1,36 @@
+"""Abstract interface for speech-tokenizer quantizers."""
+
+from abc import ABC, abstractmethod
+
+import torch
+
+from espnet2.tok.output import SpeechTokenizerOutput
+
+
+class AbsSpeechTokenizerQuantizer(torch.nn.Module, ABC):
+    """Define the common contract for differentiable speech quantizers.
+
+    Implementations are expected to provide differentiable assignments for
+    training, hard integer token IDs for inference, and the corresponding
+    sequence lengths.
+    """
+
+    @property
+    @abstractmethod
+    def num_clusters(self) -> int:
+        """Return the number of discrete clusters."""
+        raise NotImplementedError
+
+    @abstractmethod
+    def forward(
+        self, features: torch.Tensor, feature_lengths: torch.Tensor
+    ) -> SpeechTokenizerOutput:
+        """Quantize continuous features with a differentiable assignment."""
+        raise NotImplementedError
+
+    @abstractmethod
+    def encode(
+        self, features: torch.Tensor, feature_lengths: torch.Tensor
+    ) -> SpeechTokenizerOutput:
+        """Quantize continuous features deterministically."""
+        raise NotImplementedError
--- a/espnet2/tok/quantizer/differentiable_kmeans.py
+++ b/espnet2/tok/quantizer/differentiable_kmeans.py
@@ -1,0 +1,226 @@
+"""Differentiable k-means quantization."""
+
+from pathlib import Path
+from typing import Union
+
+import joblib
+import numpy as np
+import torch
+import torch.nn.functional as F
+from typeguard import typechecked
+
+from espnet2.tok.output import SpeechTokenizerOutput
+from espnet2.tok.quantizer.abs_quantizer import (
+    AbsSpeechTokenizerQuantizer,
+)
+
+
+class DifferentiableKMeans(AbsSpeechTokenizerQuantizer):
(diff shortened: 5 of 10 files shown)


sw005320 (commented):
Thanks @ondatk68 — the abstraction is clean, the `SpeechTokenizerOutput` dataclass documents its shapes properly, and separating `forward` (differentiable) from `encode` (deterministic) on the quantizer contract is the right split for what this needs to do. Splitting the work into seven PRs is appreciated.

Two things are worth settling now rather than after the remaining six land, because both are about where this code sits relative to what espnet2 already has, and that gets expensive to move later.

## 1. Say in the code why this is not the codec task

The first question any reviewer will ask is why a new quantizer hierarchy exists when `espnet2/gan_codec/` already has one. I went through it, and I think you are right to keep this separate — but nothing in the PR says so, so the question will keep being asked.

The argument, as far as I can reconstruct it: `AbsGANCodec` makes `decode()` abstract (`espnet2/gan_codec/abs_gan_codec.py:39`), and `ResidualVectorQuantizer` pairs `encode` with `decode` (`residual_vq.py:130,147`) because a codec is an autoencoder — waveform in, waveform out, trained with reconstruction and adversarial losses. `SpeechTokenizer` has no decoder, no reconstruction objective and no discriminator; it consumes an `AbsFrontend` and emits units for a downstream ASR and vocoder. Folding it into `gan_codec` would force a meaningless `decode()`.

Could you put that in the `espnet2/tok/__init__.py` or `AbsSpeechTokenizerQuantizer` docstring, in a sentence or two? Something that names `gan_codec` explicitly and says what makes this different. `abs_quantizer.py`'s docstring currently motivates the abstraction ("support potential discretization methods other than differentiable k-means") but not its existence alongside the codec one.

## 2. The relationship to the existing tokenizer classes needs sorting out

This is the one I would like resolved before PR 2. There are now going to be several things in espnet2 called a tokenizer, and the boundaries are not obvious from the names.

**`espnet2/speechlm/tokenizer/abs_tokenizer.py`** — `AbsTokenizer`, whose docstring describes the opposite direction from yours: "transform the LM-generated tokens into the corresponding targets. E.g., Speech Codec codes -> waveform". Its `forward` is decorated `@torch.no_grad()`. So on paper it is a de-tokenizer, and gradient flow — the entire point of your class — is explicitly excluded. That is a real distinction, and it is worth stating.

**`espnet2/beats/audio_tokenizer.py:34`** — but `AudioTokenizer` subclasses that same `AbsTokenizer` and goes the *other* way: `encode(wavs, wav_lens)` returns `(codes, code_lengths)` (`:105`). Same direction as yours, and almost the same output shape as your `token_ids` plus `lengths`. So the existing base class is already being used for both directions, which weakens the "different direction" defence.

**`espnet2/beats/tokenizer.py:62`** — `BeatsTokenizer` is the closest existing analogue by role: an SSL encoder followed by a `NormEMAVectorQuantizer` with `kmeans_init=True` (`:93`), exposing `encode()` that returns quantized features plus indices (`:130,149`). Encoder plus k-means-initialised quantizer producing discrete units is, structurally, what `SpeechTokenizer` does. The novelty in your work is that the assignment stays differentiable — which is a real difference, not a cosmetic one, but it is a difference in one property rather than in kind.

I am not asking you to inherit from any of these. Straight-through assignments with a trainable frontend genuinely do not fit a `@torch.no_grad()` base class, and I would rather have a clean hierarchy than a forced one. What I would like is for the PR to answer, in the code:

- Why `AbsSpeechTokenizerQuantizer` does not extend or replace `speechlm.tokenizer.AbsTokenizer`.
- How a user picks between `SpeechTokenizer`, `AudioTokenizer` and `BeatsTokenizer`. Right now three classes with near-identical names take audio and return discrete units, and nothing tells you which to reach for.
- Whether `BeatsTokenizer` could eventually be expressed as an `AbsSpeechTokenizerQuantizer` implementation, or whether the two stay separate permanently. Not a request to do it in this PR — but if the answer is "they converge later," the abstraction should be shaped for that now, and if it is "never," a comment saying so saves the next person the investigation.

A short "Relation to existing tokenizers" section in the PR description would cover the reviewing side, and one or two docstring sentences would cover the reader who arrives at the file cold in six months.

## Smaller

`SpeechTokenizer.__init__` reaches for the quantizer's feature dimension with `getattr(quantizer, "feature_dim", None)`, twice, including in the error message. Since every implementation must have it for that check to be meaningful, making `feature_dim` an abstract property on `AbsSpeechTokenizerQuantizer` — as `num_clusters` already is — would turn a silent `None` mismatch into a clear contract violation at class definition time.

sw005320:
@ondatk68, please reflect the review comments by me and [coderabbitai](https://github.com/apps/coderabbitai)

ondatk68:
Thank you for the review. I pushed four focused commits:

1. https://github.com/ondatk68/espnet/commit/3ce88b2b9451d75fd214d6bb98dbe419c7471d11 validates feature-length dtype and `[0, T]` bounds in both training and deterministic encoding, with boundary tests.
2. https://github.com/ondatk68/espnet/commit/5400582f9f74f0f9037e4055bcc530c8d121eef4 makes `feature_dim` part of the abstract quantizer contract.
3. https://github.com/ondatk68/espnet/commit/587575109826850a99c85c0deea4cb7453ed11f2 documents why `SpeechTokenizer` does not implement the waveform-decoding `AbsGANCodec` interface.
4. https://github.com/ondatk68/espnet/commit/2b8684f3e291d34b790d8c61e294ba400af56427 documents how the SpeechLM, AudioTokenizer, and BEATs classes differ, and adds an optional quantized-feature output for future VQ-style training.

I think ``BeatsTokenizer`` could also be reproduced using ``SpeechTokenizer`` by using ``BeatsEncoder`` as the frontend and ``quantize_layer`` and ``NormEMAVectorQuantizer`` as the quantizer. Since ``BeatsTokenizer`` is not implemented as an ``AbsFrontend``, some additional modifications would likely be required, though.

However, ``BeatsTokenizer`` was originally designed for BEATs pretraining, and its training objective is quite different. Therefore, I do not think there is a strong need to unify ``BeatsTokenizer`` and ``SpeechTokenizer``.

PR #6645, opened 2026-09-07
Bagpiper-Integration-1: Installation

jctian98:
## What did you change?

Add `espnet2/speechlm/INSTALL.md` with a short installation path: choose a PyTorch CUDA build, install `pip install -e ".[speechlm]"`, then install FlashAttention for the GPU, with an FA3 example for H100/H800.

Keep SpeechLM dependencies in `pyproject.toml`. Update Transformers from 5.10.1 to 5.14.1 and add `flash-linear-attention>=0.2.2`, `causal-conv1d>=1.6.2`, and `tilelang>=0.1.12`, each guarded by `sys_platform == 'linux'`. Preserve the Linux guards on TorchTitan and Liger Kernel. The PR contains no separate requirements list.

Provide a concise installation entry point for Bagpiper using ESPnet's existing SpeechLM extra. A single dependency declaration avoids conflicting package versions, while the guide covers the machine-specific PyTorch and FlashAttention installation choices.

## Is your PR small enough?

Yes: 2 files, 49 added lines and 10 removed lines.

## Additional Context

This is the first PR in the Bagpiper integration series, covering installation.

Validation: TOML parsing and whitespace checks passed. Installation and training were not run.


sw005320 (commented):
Thanks @jctian98 — a short entry point for Bagpiper that skips the full ESPnet install is a reasonable thing to want, and the FlashAttention section is genuinely useful; that part cannot be expressed as a declared dependency and belongs in prose.

The `requirement.txt` is the part I would like to reconsider. ESPnet already declares SpeechLM's dependencies in `pyproject.toml`, and a second list beside it will drift. I compared the two, and it is already inconsistent on day one.

## What the 18 lines actually are

| | count | which |
| --- | --- | --- |
| Already declared | 13 | `psutil`, `huggingface-hub`, `joblib`, `polars`, `duckdb` are in the `speechlm` extra; `lhotse`, `pyyaml`, `typeguard`, `humanfriendly`, `librosa`, `wandb`, `datasets` are in the base `dependencies` |
| Direct conflict | 1 | `transformers` |
| Genuinely new | 3 | `flash-linear-attention`, `causal-conv1d`, `tilelang` |
| Silently loses a platform marker | 2 | `torchtitan`, `liger-kernel` |

So 15 of 18 lines either duplicate or contradict something that already exists, and only 3 are specific to this work.

## The two that will actually bite

**`transformers` is pinned to two different versions.**

```
pyproject.toml [speechlm]   transformers==5.5.4
requirement.txt             transformers==5.14.1
```

Both are exact pins, so they cannot both hold. Anyone who does `pip install -e ".[speechlm]"` and then follows INSTALL.md step 2 gets the second pin silently overwriting the first — no error, no warning, and a `transformers` that the rest of the SpeechLM code was not pinned against. If 5.14.1 is what Bagpiper needs, that is a real change to make, but it should be made once in `pyproject.toml`, where the people who depend on 5.5.4 will see it in review.

**`torchtitan` and `liger-kernel` lose their `sys_platform` guard.**

`pyproject.toml` carries them as:

```toml
# torchtitan (distributed training) and liger-kernel (CUDA Triton
# kernels) are Linux+CUDA only. Leave them off Windows/macOS so
# `pip install espnet[all]` doesn't spend minutes backtracking
# through incompatible versions and then fail.
"torchtitan; sys_platform == 'linux'",
"liger-kernel; sys_platform == 'linux'",
```

`requirement.txt` lists both unconditionally, so running it on macOS or Windows walks straight into the failure that comment was added to prevent.

## Keep INSTALL.md

I would keep the file. Two of its three steps cover things a dependency list cannot express:

- **Step 1**, choosing a CUDA build of PyTorch, is a user decision about their machine.
- **Step 3**, FlashAttention, *cannot* go in `pyproject.toml`. `git+https://...#subdirectory=hopper` is a direct reference, and PyPI rejects the whole distribution with `400 Can't have direct dependency` when one is present — that is what kept ESPnet off PyPI from v.202604 to v.202609, and `ci/check_ci_image_config.py::check_no_direct_references` now fails the build on one. The per-GPU backend choice is not expressible declaratively either.

That is exactly the material a prose install guide should carry.

## Suggested shape

1. Add the three genuinely new packages to the `speechlm` extra in `pyproject.toml`. If they are Linux+CUDA only like their neighbours, give them the same `; sys_platform == 'linux'` marker.
2. Decide the `transformers` version in one place. Bumping `5.5.4` to `5.14.1` is a call worth making explicitly rather than by shadowing.
3. Drop `espnet2/speechlm/requirement.txt`.
4. Keep INSTALL.md, with step 2 becoming `pip install -e ".[speechlm]"`. Steps 1 and 3 stay as written.

That leaves one place where SpeechLM's dependencies are declared, and an install guide that explains only the parts a declaration cannot.

sw005320 (commented):
Thanks for reworking this — dropping `requirement.txt`, moving the three packages into the `speechlm` extra with platform markers, and keeping INSTALL.md for the parts a declaration cannot express is exactly right. Apologies for one thing in my earlier comment: I read `pyproject.toml` from a stale checkout and quoted `transformers==5.5.4`; master is on `==5.10.1`, so the bump here is 5.10.1 → 5.14.1, not 5.5.4 → 5.14.1.

The CI failure is one package, and it takes everything else down with it.

## Root cause

All 95 failures are the same `build` job failing and cascading. `causal-conv1d` publishes **no wheels — sdist only**, and its `setup.py` requires `nvcc`:

```
UserWarning: causal_conv1d was requested, but nvcc was not found.
torch.__version__  = 2.14.0+cu130
  File "<string>", line 185, in <module>
NameError: name 'bare_metal_version' is not defined
```

`bare_metal_version` is only assigned inside the nvcc branch, so on a CUDA-less image the warning is followed by a hard `NameError`. That is a bug in the package, but the effect here is what matters:

**it fails during `Getting requirements to build wheel`** — metadata generation, before anything is installed. So `pip install -e ".[...]"` aborts, `ci/install.sh` exits 1, `ci.dockerfile:40` fails, the image never builds, and every job that needs it fails.

`sys_platform == 'linux'` cannot prevent this. CI *is* Linux. The marker guards the platform, not the presence of CUDA, and there is no environment marker for "has nvcc".

Worth noting separately: the traceback shows `torch 2.14.0+cu130` inside pip's isolated build environment. `causal-conv1d`'s build requires torch, and the isolated env resolves it independently of the `torch>=2.9.1,<2.12` pin in `pyproject.toml`.

## ESPnet already has the pattern for this

Earlier in the same log, `flash-attn` hits the identical `CUDA_HOME` error — and the build continues:

```
Flash Attention failed to install, trying without building CUDA
Failed to install flash attention. Skipping...
```

because it goes through `tools/installers/install_flash_attn.sh`, which tolerates failure, rather than being declared in `pyproject.toml`. CUDA-only source packages are installed opportunistically; only things that can resolve everywhere are declared.

## What I would change

I checked all three on PyPI:

| package | latest | distributions | verdict |
| --- | --- | --- | --- |
| `causal-conv1d` | 1.7.0 | **sdist only** | move out of `pyproject.toml` |
| `tilelang` | 0.1.14 | 4 wheels + sdist (`cp39-abi3` for manylinux x86_64, aarch64, macOS arm64) | fine to keep |
| `flash-linear-attention` | 0.5.2 | `py3-none-any` wheel | fine to keep |

So only `causal-conv1d` needs to move. Either add a `tools/installers/install_causal_conv1d.sh` alongside the flash-attn one, or put it in INSTALL.md step 3 next to FlashAttention — it belongs to the same category of "needs a CUDA toolchain, choose per your GPU," which is the material that justifies keeping INSTALL.md at all.

`tilelang` ships a manylinux wheel, so `sys_platform == 'linux'` on it is doing real work and can stay. `flash-linear-attention` is pure Python; its marker is harmless but is not buying anything, in case you would rather drop it.

jctian98:
This PR may have more changes before the training code and recipes settle down. Don't merge right now.

jctian98:
Ok to merge for now; we can add some patches if we find some issues for the three PRs.

sw005320 (approved):

PR #6646, opened 2026-09-08
Bagpiper-Integration-2: Trainer

jctian98:
SpeechLM currently replaces Hugging Face Qwen3 MoE experts with custom grouped-matmul and expert-parallel implementations. This change uses Transformers' native `grouped_mm` backend and removes SpeechLM's EP/DeepEP support.

- Configure native experts for both pretrained loading and pipeline-parallel replicas.
- Preserve the memory-efficient router loss and capture native router logits in PP, including with activation checkpointing.
- Delete `grouped_moe.py`, EP-specific sharding/prefetch/gradient-clipping branches, and their obsolete tests. Reject configurations that request EP.
- Keep unit coverage on CPU, organize tests by production source file, and preserve the native GPU dispatch comparison as opt-in coverage under `test_utils/speechlm/`.

The installation update is tracked separately in #6645. Validation used its dependencies in a fresh Python 3.12 conda environment: PyTorch 2.11.0+cu128, Transformers 5.14.1, and TorchTitan 0.2.2.

Validation:

- 122 related CPU tests passed with CUDA hidden and Hugging Face offline, including native-expert identity/configuration, router-loss gradients, PP activation checkpointing, and trainer behavior.
- The opt-in H100 BF16 test verifies two native grouped-MM calls and compares outputs, input gradients, and parameter gradients against eager experts. It passed with `torch.nn.functional.grouped_mm` and with that API temporarily hidden to exercise the `torch._grouped_mm` dispatch used by older PyTorch. PyTorch 2.9 itself was not installed for this check.
- Existing Transformers mocks are isolated from lazy-module replacement, so model and trainer tests pass together.
- Black, isort, and `git diff --check` passed.

Run the optional GPU check explicitly from the repository root:

```bash
python -m pytest -o addopts= test_utils/speechlm/test_native_moe_cuda.py
```

The diff includes deletion of the custom MoE implementation and movement of existing tests into source-specific files. Full Bagpiper training has not been run; the existing SpeechLM fused-loss return-value mismatch with Liger 0.8.2 is outside this MoE/EP change.

PR #6647, opened 2026-09-08
Bagpiper-Integration-3: Recipe

jctian98:
## What did you change?

Add training recipes that run the current SpeechLM trainer from prepared dataset manifests and length statistics:

- Bagpiper warmup as the `conf/train.yaml` default, with pretraining and SFT configurations under `conf/tuning/`.
- Bagpiper-TTS supervised fine-tuning through its default `conf/train.yaml`.
- A shared `torchrun` launcher with single-node and multi-node support, configurable GPU counts, and DCP initialization.
- READMEs explaining the training-only layout, environment and scheduler behavior, stage purposes and schedules, token/runtime estimates, and checkpoint recovery. All four YAMLs document their Hopper-specific attention default.

Data preparation is outside these recipes. The configurations use Qwen3-8B-Base, Xcodec, the Qwen3-Omni audio encoder, and TorchTitan FSDP2. Training schedules are adjustable starting settings rather than an exact reproduction of the original DeepSpeed runs.

Provide ESPnet-style `run.sh`, `conf/`, and README entry points for Bagpiper and Bagpiper-TTS without the machine-specific launch scripts used in local experiments.

## Is your PR small enough?

Yes: 14 files changed, 618 insertions and 1 deletion.

## Additional Context

This is the recipe part of the Bagpiper integration, alongside installation #6645 and trainer #6646. The branch starts independently from master and contains only recipe/template changes. The installation and trainer integration must be available when this recipe is merged.

Validation completed:

- Shell syntax and help output for both recipe entry points.
- All four YAML configurations parsed and checked against current trainer settings.
- 16 launcher checks covering canonical defaults, training CLI compatibility, quoted paths/specifiers, single-node and multi-node arguments, initialization, invalid inputs, and omission of empty specifiers.
- `PYTHONHASHSEED=random` is overridden with `0`; independent Python workers agree on dataset hashes. The renamed YAMLs preserve their training settings.
- `git diff --check` and symlink/documentation link checks.

Full training has not been run. The current Titan loader accepts DCP directories for training initialization; the released Hub `base.pt` / `model.pt` files are not yet supported by this path. An existing Liger 0.8.2 return-value mismatch in the SpeechLM loss also remains to be addressed in the trainer integration.


sw005320 (commented):
Thanks @jctian98 — the launcher is careful (every argument validated with a clear message, `PYTHONHASHSEED` pinned so resampling agrees across ranks, `--resume-path` checked for `.metadata` rather than just existing), and the stage table plus the resume-vs-reinitialize distinction are the two things people actually get wrong. CI is green at 134 passing.

Two structural comments and a few smaller ones.

## 1. Say why this recipe is shaped differently from every other one

`egs2/bagpiper/speechlm1` and `egs2/bagpiper_tts/speechlm1` contain `README.md`, `conf/`, `run.sh`, and symlinks to `train.sh` and `utils`. Nothing else. Every other recipe in `egs2/` — including `egs2/TEMPLATE/speechlm1` itself, which ships a `setup.sh` for exactly this — also carries `cmd.sh`, `path.sh`, `db.sh`, `local/`, `scripts`, `pyscripts`, `steps`, and `conf/{slurm,queue,pbs}.conf`.

I think that is the right call for SpeechLM. But it is surprising enough that it should be written down, because three of the consequences change how a user has to operate:

- **No `path.sh`.** `train.sh` sets `PYTHONPATH` to the repo root and runs `python=python`, whatever is active. Every other recipe sources `path.sh` and gets `tools/venv`. So this recipe *requires* the user to have activated the SpeechLM environment from `INSTALL.md` first, and silently uses the wrong interpreter if they have not.
- **No `cmd.sh`.** There is no `${cuda_cmd}`, so nothing goes through `run.pl` / `slurm.pl`. Launch is `torchrun` directly, and multi-node means running the same command by hand on every node with a different `--node-rank`. That is a real operational difference from `--ngpu` on a normal recipe.
- **No `local/data.sh`.** Data preparation is out of scope, which the README does say — but it is what makes this not a recipe in the usual `run.sh --stage 1` sense.

`egs2/TEMPLATE/speechlm1/README.md` is the natural home for this, and it is currently five lines. While you are there, the one sentence about `speechlm.sh` deserves expanding: two drivers now live side by side in that directory — `speechlm.sh` calling `espnet2.bin.speechlm_train`, and the new `train.sh` calling `espnet2.speechlm.bin.train` — and a reader needs to know which one applies to them and why both exist.

## 2. The per-stage description should carry the numbers

The stage table gives configuration, initialization, and trainable components, which is a good start. But the schedule lives only in the YAML, and it is what a reader most needs in order to judge whether they can run this:

| stage | `max_step` | `lr` | `warmup_steps` | `grad_accum` | `min_lr_ratio` |
| --- | --- | --- | --- | --- | --- |
| warmup | 10,000 | 5e-4 | 100 | 4 | **1.0** |
| pretrain | 600,000 | 1e-4 | 5,000 | 4 | 0.3 |
| sft | 50,000 | 1e-5 | 1,000 | **1** | 0.1 |

Things a reader cannot currently get from the README:

- **Why the example resume paths are `step_10000` and `step_600000`.** They are each stage's `max_step`, but the README presents them as bare paths, so it reads like an arbitrary checkpoint rather than "the end of the previous stage."
- **`min_lr_ratio: 1.0` in warmup means the learning rate is constant** after 100 steps — no decay at all, unlike the other two stages. That is a deliberate choice for a stage that is only aligning new embeddings, and it is worth one sentence.
- **SFT uses `gradient_accumulation_steps: 1` against 4 elsewhere**, so its effective batch is a quarter of the others at the same `batch_size: 8192`. Nothing in the README hints at this.
- **What each stage is *for*.** The table says what is trainable; it does not say that warmup exists to align the expanded embeddings and audio adaptor to a frozen backbone, that pretraining learns the joint model, and that SFT specialises to dialogue. One line each would carry a lot.
- **Scale.** 600,000 steps at 8192 tokens per GPU with `grad_accum: 4` is a very large run, and the README gives no way to estimate it. Even a rough "this was N GPU-days on M H100s" would tell a reader whether to attempt it — and the caveat that these are starting settings rather than the DeepSpeed reproduction would land better next to a number.

## Smaller things

**`run.sh` defaults to the middle stage.** `egs2/bagpiper/speechlm1/run.sh` hardcodes `--train-config conf/train_pretrain.yaml --output-dir exp/pretrain`, but every documented invocation overrides both, and the documented first step is warmup. Someone who runs `./run.sh` with only data arguments starts pretraining from a bare Qwen3-8B-Base with no warmup checkpoint, and nothing says that is wrong. Either default to `train_warmup.yaml`, or drop the two defaults so the stage has to be named explicitly. (The `bagpiper_tts` one defaults to its only config, so it is fine.)

**The PR links will go stale.** README lines 13-15 tell the reader to combine this with #6645 and #6646. That is useful right now and confusing after all three merge. Consider phrasing it as a version requirement instead, or add it to a merge checklist.

**Empty specifiers are always passed.** `train.sh` sends all four `--*-specifier` flags unconditionally, so a user who supplies only registered specifiers still passes `--train-unregistered-specifier ""`. The shell-side validation correctly requires one of each pair, but it is worth confirming that `espnet2.speechlm.bin.train` treats an empty string the same as an absent flag.

**FlashAttention-3 is the shipped default in all four YAMLs.** Both READMEs tell the reader to change it off Hopper, which is right — but as the default, an A100 user gets a runtime failure rather than a fallback. A short comment at the two `attn_implementation` lines in each YAML would catch people who copy the config before reading the README.

sw005320:
@jctian98, please also reflect coderabbitai's reviews

sw005320 (approved):
LGTM
After the CI passes, I'll merge this PR.

Please go to the next one!

Issue #6644, opened 2026-09-07
Docs site: 17 Dependabot alerts blocked behind a vuepress upgrade that OOMs the docs build


sw005320:
## The problem

17 Dependabot alerts are open against `doc/vuepress/package-lock.json` — 13 × `vite`, 3 × `js-yaml`, 1 × `esbuild` — and none of them can be fixed on their own.

`package.json` exact-pins `vuepress` while caret-ranging its plugins, so `vuepress-plugin-search-pro` floats to rc.59, whose peer is `vuepress@2.0.0-rc.18`:

```
Found: vuepress@2.0.0-rc.14                    (exact pin, root devDependency)
Could not resolve: peer vuepress@2.0.0-rc.18   from vuepress-plugin-search-pro@2.0.0-rc.59
```

`npm audit fix` aborts with `ERESOLVE`. That is why Dependabot — which is enabled here and has merged 16+ other PRs against this same directory — left exactly these behind. Clearing them requires moving the whole vuepress stack at once.

## How urgent this is: not very

Everything here is `scope=development`. This tree only builds the documentation site in `ci/doc.sh` and ships in no released espnet artifact, so none of these are reachable by anyone installing the package. Please do not treat this as a security emergency — it is dependency hygiene, and the reason it is worth doing is that the pins keep drifting further out of date, not that users are exposed.

The `nanoid` alerts from the same tree were fixed separately in #6640.

## What is already done

Branch **`fix/vuepress-rc31-upgrade`** (kept, please do not delete) — see #6641 for the full write-up.

The dependency work is complete and verified:

```
@vuepress/bundler-vite      2.0.0-rc.14  -> 2.0.0-rc.31
vuepress                    2.0.0-rc.14  -> 2.0.0-rc.31
vuepress-theme-hope         2.0.0-rc.51  -> 2.0.0-rc.109
vuepress-plugin-search-pro  ^2.0.0-rc.51 -> removed, replaced by @vuepress/plugin-slimsearch 2.0.0-rc.134
vue                         ^3.4.31      -> ^3.5.40
sass-embedded                            -> ^1.104.0
```

- `npm audit`: **0 vulnerabilities** (from 36). All 17 alerts would close.
- The lockfile regenerates byte-identically from `package.json` alone.
[...]

PR #6641, opened 2026-09-07
Upgrade the VuePress docs stack to the rc.31 line

sw005320:
Clears the **17 remaining Dependabot alerts** against `doc/vuepress/package-lock.json` (11 × `vite`, 3 × `js-yaml`, 1 × `esbuild`, plus 2 more `nanoid` advisories). Follow-up to #6640, which fixed only the one `nanoid` alert that could be fixed in isolation.

## Why they could not be fixed one at a time

`package.json` exact-pins `vuepress` while caret-ranging its plugins, so `vuepress-plugin-search-pro` floated to rc.59 — whose peer is `vuepress@2.0.0-rc.18` — and resolution aborted:

```
Found: vuepress@2.0.0-rc.14                    (exact pin, root devDependency)
Could not resolve: peer vuepress@2.0.0-rc.18   from vuepress-plugin-search-pro@2.0.0-rc.59
```

`npm audit fix` cannot get past that, which is why Dependabot — which *is* enabled here and has merged 16+ other pull requests against this same directory — left these behind. Meanwhile the rc.14 bundler pins `vite ~5.3.1` and `esbuild ~0.21.5`, and `gray-matter` pulls `js-yaml` 3.x. Every remaining alert traces to that one stale pin.

## Change

One coherent version set instead:

```
@vuepress/bundler-vite      2.0.0-rc.14  -> 2.0.0-rc.31
vuepress                    2.0.0-rc.14  -> 2.0.0-rc.31
vuepress-theme-hope         2.0.0-rc.51  -> 2.0.0-rc.109
vuepress-plugin-search-pro  ^2.0.0-rc.51 -> removed
@vuepress/plugin-slimsearch              -> 2.0.0-rc.134
vue                         ^3.4.31      -> ^3.5.40
```

`search-pro` is **replaced, not bumped**: its own latest still peers on rc.18, so it cannot coexist with rc.31. `slimsearch` is its successor here — theme-hope rc.109 peer-declares `@vuepress/plugin-slimsearch@2.0.0-rc.134` and types `plugins.searchPro` as `never` (*"@deprecated Use `plugins.slimsearch` instead"*).

Resulting tree: **vite 8.2.2**, **esbuild gone entirely** (vite 8 uses rolldown 1.2.7), **js-yaml 3.15.2** under gray-matter (the advisories cover `<= 3.15.0`). `npm audit`: **0 vulnerabilities, down from 36.**

## theme.ts migrations

rc.109 types four of the options this config used as `never`. All four are migrated:

[...]

Issue #1686, opened 2020-03-17
results are not stable | very slow


vcjob:
Hello everyone!

I am using one of the ASR egs. When I try to decide which beam-size, ctc-weight and lm-weight should I use, but while repeating inference with the same parameters, results are different each time!
This is how the model is being loaded:
```
from espnet.asr.pytorch_backend.asr import load_trained_model
model, train_args = load_trained_model(model_dir)
model.to('cuda')                                                                                                     
model.eval()

```
How to make it return the same results each time? I use beam search with size 2 and always get the 1st result out of it.
-------
Another question I have is, why it is so slow on wavs longer ~15 secs? What can I do about it? I mean, `encoded = model.encode(torch.as_tensor(fbank).cuda())` is really slow, especially on CPU.

Thank you a lot!

sw005320:
> Hello everyone!
> 
> I am using one of the ASR egs. When I try to decide which beam-size, ctc-weight and lm-weight should I use, but while repeating inference with the same parameters, results are different each time!
> This is how the model is being loaded:
> 
> ```
> from espnet.asr.pytorch_backend.asr import load_trained_model
> model, train_args = load_trained_model(model_dir)
> model.to('cuda')                                                                                                     
> model.eval()
> ```
>
> How to make it return the same results each time? I use beam search with size 2 and always get the 1st result out of it.

Thanks for the report.
Could you let us know more details?
We might make some part non-deterministic (we try not to do it as much as possible).

> Another question I have is, why it is so slow on wavs longer ~15 secs? What can I do about it? I mean, `encoded = model.encode(torch.as_tensor(fbank).cuda())` is really slow, especially on CPU.

We have several issues in the decoder but I have never had serious issues in the encoder. Could you let us know how long it would take for the encoder calculations for ~15 sec utterances?

vcjob:
> > Hello everyone!
> > I am using one of the ASR egs. When I try to decide which beam-size, ctc-weight and lm-weight should I use, but while repeating inference with the same parameters, results are different each time!
> > This is how the model is being loaded:
> > ```
> > from espnet.asr.pytorch_backend.asr import load_trained_model
> > model, train_args = load_trained_model(model_dir)
> > model.to('cuda')                                                                                                     
> > model.eval()
> > ```
> > 
> > 
> > How to make it return the same results each time? I use beam search with size 2 and always get the 1st result out of it.
> 
> Thanks for the report.
> Could you let us know more details?
[...]

PR #6654, opened 2026-09-09
Raise the docs build heap to 14 GB (vuepress rc.31 stack)

MannXo:
## What did you change?

Builds on `fix/vuepress-rc31-upgrade` (the rc.31 dependency upgrade from #6641, unchanged here) and adds one more commit: raises the docs build's V8 heap from 8192 MB to 14336 MB in `ci/doc.sh`.

#6644 tracks 17 Dependabot alerts that can only close together, by moving the whole vuepress stack to rc.31 at once (the dependency work itself is done and verified — 0 vulnerabilities, lockfile regenerates byte-identically). The blocker has been that the docs build OOMs on that stack: rendering the site's 1856 pages needs more than 2x the memory the rc.14 stack needed, and 8192 MB (the previous attempt) wasn't enough either — the same GC pattern recurred, pinned against the new ceiling instead (7942 → 7960 MB before the same "Reached heap limit" abort).

`ubuntu-latest` runners have 16 GB total, so there's headroom above 8 GB that hasn't been tried. This PR is that next step: if 14 GB clears it, the memory requirement was simply higher than first assumed and this is the fix. If it still OOMs, the failure mode itself is informative — a plain OOM-kill with no V8 heap-limit message would mean the render genuinely needs more than the runner has, which points at the render phase itself (there's a `chunks larger than 1024 kB after minification` warning from the vite/rolldown bundler in the same build that's worth checking next) rather than at the ceiling.

## Is your PR small enough?

The lockfile diff is large (dependency version bumps + updated integrity hashes, byte-identical to a real `npm install`) but that's the min-necessary result of the rc.31 upgrade, not something split further. Outside the lockfile, this PR's own new commit touches one file, one export line.

## Additional Context

Related: #6644, #6641, #6640.

This PR needs the `Documentation` label for `generate_and_deploy_doc` to actually run and verify the build — I don't have permission to add labels from a fork PR, so flagging it here per #6644's own note.


sw005320:
Closing and immediately reopening this on purpose — nothing is wrong with the PR, this is a CI retrigger. Reopening in a moment.

The reason: `generate_and_deploy_doc` **skipped** on the run that's currently green, so the 14 GB heap change has not actually been exercised yet.

```
publish doc run started : 2026-09-09T12:58:55Z
Documentation label added: 2026-09-09T12:59:49Z   (54s too late)
```

`get_labels` therefore saw no label, `is_doc` came out `false`, and the build job was skipped. A skipped job doesn't fail its workflow, so the run reports `success` — which is easy to misread as "the docs build passed". It didn't run at all.

A plain workflow re-run won't fix this either: `joerick/pr-labels-action` reads labels from the event payload (`github.context.payload?.pull_request?.labels`), and a re-run replays the original payload. A fresh `pull_request` event is needed. `reopened` is one of the default trigger types, so close+reopen produces one with the label now present — and saves you having to push an empty commit from the fork.

Thanks for picking this up, and sorry for the notification noise.

sw005320:
The retrigger worked — `generate_and_deploy_doc` actually ran this time. It failed, but the way it failed answers the question the PR set out to ask, and the answer is worth more than a pass would have been. Thanks for picking this up.

## 14 GB was applied, and it still aborted

```
- Rendering 1853 pages
Mark-Compact 13918.2 (14379.4) -> 13906.1 (14383.4) MB, 19196.32 ms, current mu = 0.017
Mark-Compact 13922.8 (14383.6) -> 13909.2 (14387.1) MB, 24630.76 ms, current mu = 0.007
FATAL ERROR: Reached heap limit Allocation failed - JavaScript heap out of memory
                                                                      (exit 134)
```

The ceiling took effect — old space grew to 14379 MB. And it is a clean V8 heap-limit abort, not the kernel OOM-kill you flagged as the other possibility, so the runner had the physical memory. On that specific question your PR body was right to distinguish the two, and it came out on the V8 side.

## The GC numbers are the actual finding

A **19-second** full mark-compact reclaimed **12 MB**. The next one spent **24.6 seconds** to reclaim 13.6 MB. `current mu = 0.007` means the process is spending **99.3 % of its time in GC**.

Reclaiming 12 MB from a 14 GB heap means essentially everything on it is still reachable. That is not a workload that needs more memory — it is unbounded retention. Something holds a reference to every page it renders.

Three ceilings, same shape:

| heap limit | peak before abort | render phase before abort |
|---|---|---|
| ~4 GB (node default) | 3982 / 4120 MB | ~40 s |
| 8 GB | 7942 / 8236 MB | ~99 s |
| **14 GB** | **13918 / 14379 MB** | **~333 s** |

It fills whatever it is given and dies at the wall. 24 GB would do the same thing, more slowly. **Raising `--max-old-space-size` is a dead end**, and that is now established rather than suspected — which is exactly what this PR was for.

## Suggestion

Drop the `ci/doc.sh` change. At 14336 it does not fix the build and it stretches the time-to-failure from 40 s to over 5 minutes, so CI pays five times as much to learn the same thing. With that commit gone the branch is the rc.31 dependency work alone, which is #6641 again — so I would close this PR too and let #6644 carry the state. Nothing is lost: `fix/vuepress-rc31-upgrade` stays put, and the dependency side of it is still sound (0 vulnerabilities, lockfile regenerates byte-identically, `theme.ts` typechecks clean).

I will fold this result into #6644 so the next person does not re-run the experiment.

## Where I would look next

The retention is in the render phase, so the question is what accumulates per page. The rigorous route is a heap profile on the real content — `node --heap-prof` around `vuepress-vite build`, which needs `ci/install.sh` + `ci/doc.sh` locally to generate the actual sphinx / notebook / recipe pages. Nobody has done that yet.

Cheaper first pass is bisecting the plugins: build with `plugins.slimsearch: false`, then with the redirect and sitemap plugins off, and see which one stops the growth. Fair warning that slimsearch is a weaker suspect than it looks — I built 1856 synthetic prose pages locally with it enabled and peaked at 2.4 GB — so do not anchor on it.

One more caveat from my own attempts, recorded in #6641: synthetic corpora do not model these docs. At the code-block density where they discriminate, the **rc.14** stack is worse than rc.31 — more memory, slower. Any reproduction needs the real generated content.

MannXo:
Agreed on all of it, and thank you for running the retrigger rather than just reading the skip as a pass.

The GC numbers settle it. A 19-second mark-compact reclaiming 12 MB out of 14 GB means essentially nothing on the heap is collectable, so the process is not short of memory, it is holding everything. `current mu = 0.007` is the same statement from the other side. A larger ceiling buys proportionally more time in the death spiral and nothing else, which your three-point table shows directly: the peak sits between 96.4% and 96.8% of the limit every time, so the build fills whatever it is handed. That is retention, not headroom, and my PR's premise was wrong.

Closing it. Your read of what is left is exact: this branch is your three commits from `fix/vuepress-rc31-upgrade` plus one commit of mine that touches `ci/doc.sh` and nothing else, so with that commit dropped there is no diff here that is not already #6641. Better for #6644 to carry the state than for two PRs to.

Two notes for whoever picks up the retention hunt, neither of them a result.

The heap profile has not been done and I have not done it. I looked for a matching upstream report before guessing and did not find one: vuepress/core#1678 is a heap limit abort in the *compile* phase rather than render, and vuepress/core#1686's `writeTemp` cache fix targets stale state across dev-server restarts, so neither describes this. That is a null result, not a lead.

The one number worth carrying forward is the per-page figure. 13918 MB live across 1853 pages is about 7.5 MB retained per rendered page, which is far too large for accumulated HTML output and points at something structural being held per page rather than a growing string or array. That is consistent with your slimsearch caveat, since a search index over this corpus would not reach that scale either, and it suggests the bisect will be more informative if it watches the growth *rate* per page rather than only whether the build survives.

Happy to keep going on #6644 if it is useful. I would want to do the real thing, `ci/install.sh` plus `ci/doc.sh` with `node --heap-prof` around the build, rather than another cheap probe, since the last cheap probe is what we just spent five minutes of CI learning from. Say the word if you would rather leave it with you.

sw005320:
Please keep going — and yes, do the real thing rather than another cheap probe. That is exactly the right call, and `ci/install.sh` + `ci/doc.sh` with `node --heap-prof` around the build is the step #6644 names as the one nobody has taken.

Three things in your closing comment are worth more than the PR would have been if it had passed.

The **7.5 MB retained per rendered page** is the first number anyone has produced that constrains what the cause can be. You are right that it is too large for accumulated HTML — and it also rules out the slimsearch theory more firmly than my own caveat did, since an index over this corpus is nowhere near that scale per page. That figure belongs in the issue, not on a closed PR, so I am carrying it over.

The **upstream search** is a real contribution even though it found nothing. vuepress/core#1678 being a compile-phase abort and #1686 being a dev-server cache fix are both exactly the kind of near-miss that someone would otherwise re-find and mistake for a lead. A recorded null result saves that.

And **watching the growth rate per page rather than survival** is a better experimental design than what I proposed. My bisect suggestion only asked whether the build lives, which is a one-bit answer per CI cycle. Rate-per-page gives you a signal from a run that still dies, and from far fewer pages than the full set.

One correction to my own review while you are picking this up: I claimed the regression came from shiki 1.10 → 4.4, and the measurements did not support that. Synthetic corpora turned out not to model these docs at all — at the code-block density where they discriminate, the rc.14 stack is *worse* than rc.31, more memory and slower. Both the highlighter-language restriction and the prismjs swap measured as no help. So the cause is genuinely open; do not let my framing narrow where you look.

Assigning you on #6644 and consolidating the state there now.

PR #6655, opened 2026-09-09
Add seoul corpus ASR recipe

KimHaeri:
## What did you change?
Added a new ESPnet2 ASR recipe, `egs2/seoul_corpus/asr1`, for [Seoul Corpus (Korean Corpus of Spontaneous Speech)](https://www.openslr.org/113/): one hour of interview speech from each of 40 native Seoul-Korean speakers. 

---

Seoul Corpus is one of the open corpora of a spontaneous conversational Korean, and ESPnet had no recipe for it.
It also requires some specific data preprocessing. The labels are provided as Praat TextGrids, and the interviewer's speech is recorded on the same audio channel as the interviewee's, so it needs to be excluded by tag.

One setting carries the result:
- `--min_wav_duration 1.0` on train/valid. Without it the recipe does not converge usefully: 9.4% of the training utterances are a single syllable, so the loss can be driven down by learning the label prior instead of the acoustics.

---

## Is your PR small enough?
Yes

---

## Additional Context
Final results on test set:
| | Corr | Sub | Del | Ins | Err |
| --- | --- | --- | --- | --- | --- |
| CER | 85.2 | 10.5 | 4.4 | 2.7 | 17.6 |
| WER | 68.3 | 26.7 | 5.0 | 2.8 | 34.5 |
| TER | 74.8 | 18.0 | 7.2 | 2.1 | 27.3 |

- An external LM is configured but off. The reported CER is from the ASR model alone. You can enable it with `./run.sh --use_lm true`, but it is untrained and unevaluated in this PR, and `lm_weight` would need tuning.


egs2/seoul_corpus/asr1/README.md
sw005320, line 1: - Please upload the model to the HF hub and add a link to this README.md
- "* **Non-speech intervals are dropped**: `<SIL>`, `<NOISE>`, `<VOCNOISE>`,
  `<LAUGH>`, `<UNKNOWN>`, `<PRIVATE.INFO>`, and `<IVER>` (the interviewer
  talking). `<LAUGH-그래서>` means the speaker said 그래서 while laughing, so it
  becomes plain `그래서`. Any interval still holding an unrecognised `<...>` tag
  is dropped rather than trained on with a hole in its transcript.": We should keep them during training/inference and drop them during scoring. This is a normal practice. 
  - when you handle them, please make them as a special token in the sentencepiece so that it will not be tokenized.

KimHaeri: Yes, that's true. I mistakenly missed that point. I'll fix it and retrain the model. I'll also upload the model trained with corrected transcripts and tokenizer. Thank you for the comment!

sw005320:
Please also fix the CI error https://github.com/espnet/espnet/actions/runs/34371628294/job/102534018330?pr=6655

sw005320:
@KimHaeri, any update?

KimHaeri:
> @KimHaeri, any update?

I'm running an experiment now, and I'll update it once I get the results.

PR #6658, opened 2026-09-10
Decode espnet3 test sets in batches by default

sw005320:
Follow-up to #6599, which added utterance-batched beam search to `Speech2Text` and the espnet2 CLI. This makes espnet3 use it.

## What did you change?

**Batched inference in espnet3 did not run.** `InferenceRunner` already chunks the dataset when `batch_size` is set and hands the model one list per input key, but `Speech2Text.__call__` only took a single utterance. The runner caught the resulting type error and told the user to set `batch_size` to None. It gave that same advice for an out-of-memory error, where it is the wrong remedy.

- `Speech2Text.__call__` accepts a list of utterances and decodes it with `batch_decode`, which now pads a list itself (a padded tensor plus lengths is still accepted). When the configured search cannot decode a batch -- a non-batch scorer forced the plain `BeamSearch`, or a streaming, transducer, time-synchronous, partially-AR or multi-speaker model -- the list is decoded one utterance at a time with a single warning. A batched config therefore works with every model; only the speed differs.
- **Out of memory is handled where it is understood.** A batch that does not fit is split in halves and retried, with a warning that names the utterance lengths and suggests lowering the batch size or sorting the data by length. A single utterance that does not fit even on its own raises a `RuntimeError` naming its length and the options: more memory, the CPU, a smaller beam, splitting the audio. `is_out_of_memory_error` in `device_funcs` recognises CUDA's `torch.OutOfMemoryError`, MPS's plain `RuntimeError`, and an OOM re-raised as the cause of another exception; `release_accelerator_memory` returns the cached blocks before the retry.
- The espnet3 runner uses the same test to report an OOM with the dataset indices, the input lengths and the batch size, instead of the generic "set batch_size to None", which is kept for every other error.
- The ASR template and the `librispeech_100` and `mini_an4` recipes set `batch_size: 4`, and their `build_output` returns one dict per item when called with a batch. The transducer config keeps single-utterance decoding (its `Speech2Text` is a different class).

espnet3 is the recipe layer users will pick up, and decoding one utterance at a time leaves most of a GPU idle. Measured for #6599 and afterwards, real models on MPS and a Colab T4, batch 4 against batch 1: 1.2x to 2.7x depending on the corpus and device, with batch 8 losing part of that to stragglers when the data is not sorted by length. Four is the default that is safe everywhere; 8-16 pays off once the runner can sort a test set by length, which it cannot yet because dataset items carry no duration until the audio is loaded.

The OOM behaviour is the second half of "on by default": a default that can crash a decode on the one long utterance in a test set is not a default. Now it costs one warning and a retry, and the failure that remains, an utterance that does not fit alone, says exactly which one and why.

## Is your PR small enough?

Library: 3 files, +170 / -10. Recipes and template: 6 files, +50. Tests: 3 files, +147. Yes.

## Additional Context

**Testing**

- `test_asr_inference.py` (espnet2): a list of equal-length utterances decodes exactly like the utterances one by one, and a padded tensor is still accepted; a batch whose encoder raises OOM above two utterances is retried as 3 + 2 then 2 + 1 and returns all five results, with a warning naming the lengths; a single utterance that always OOMs raises the named error on both the batched and the fallback path, and a non-OOM error passes through untouched; a config whose decoder is not a batch scorer (the default RNN test config) decodes a list one at a time and warns once.
- `test_inference_runner.py` (espnet3): an OOM is reported with input lengths and batch size, chained to the original error; the generic message is unchanged for other failures.
- `test_asr_inference.py` (espnet3): the template `build_output` handles a batch and a single item.
- `ci/test_integration_espnet3.sh` runs the mini_an4 recipe with `conf/inference.yaml`, which now decodes in batches of four, so the whole path is exercised end to end in CI.

**Not in this PR**

- Length sorting in the runner, which is what would make batch 8-16 worthwhile. Dataset items expose no duration without loading the audio; a `duration` field in the manifest would enable it.
- A too-short utterance (`TooShortUttError`) inside a batch: espnet2's CLI retries it one at a time and writes a placeholder, espnet3 does not yet.

### DIFF
--- a/egs3/TEMPLATE/asr/conf/inference.yaml
+++ b/egs3/TEMPLATE/asr/conf/inference.yaml
@@ -104,9 +104,16 @@ input_key: speech
 #
 #idx_key: utt_id
 
-# Optional batch size.
-# Set to `null` if your model or `output_fn` only supports single-item decoding.
-# batch_size: 1
+# Number of utterances decoded per model call. Consecutive dataset items form a
+# batch and are passed to the model as a list per input key.
+# `espnet2.bin.asr_inference.Speech2Text` decodes such a list in one beam
+# search, which is 1.5-2.5x faster than one utterance at a time on a GPU.
+# A long utterance next to short ones pads them all, so 4 is a safe default;
+# 8-16 pays off when the test set is roughly sorted by length. A batch that does
+# not fit in memory is split in halves and retried with a warning; a single
+# utterance that does not fit raises an error that names its length.
+# Set to `null` for a model or `output_fn` that only supports one item per call.
+batch_size: 4
 
 # Optional recipe-provided formatter that converts raw model outputs into dicts such as:
 #   {"utt_id": "...", "hyp": "..."}
--- a/egs3/TEMPLATE/asr/src/inference.py
+++ b/egs3/TEMPLATE/asr/src/inference.py
@@ -2,7 +2,14 @@
 
 
 def build_output(data, model_output, idx):
-    """Build a dict of outputs for SCP writing."""
+    """Build the output dict(s) for SCP writing.
+
+    Called with one dataset item, its model output and its index, or, when
+    the inference config sets `batch_size`, with a list of each, in which case
+    one dict per item is returned.
+    """
+    if isinstance(data, list):
+        return [build_output(d, o, i) for d, o, i in zip(data, model_output, idx)]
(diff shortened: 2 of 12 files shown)


sw005320:
@chenehk, I finished the batch decoding implementation in espnet3. Please test it.

PR #6659, opened 2026-09-10
Voxtlm v1 update

RuiRuihigh:
## What did you change?

Fixes and additions to the `voxtlm_v1/lm1` recipe ([VoxtLM](https://arxiv.org/abs/2309.07937), arXiv:2309.07937) and the shared `TEMPLATE/lm1` it's built on, found while reproducing the paper's D_Bal / OPT-350M setup end-to-end.

| Area | Change |
|---|---|
| `TEMPLATE/lm1/lm.sh` | **Correctness fix**: each speech task (asr/tts/speechlm) previously trained its **own** k-means model, all writing to the same `${km_dir}` and silently overwriting each other — tasks ended up with mutually incompatible discrete-token vocabularies despite sharing the same `0..nclusters-1` token id range downstream. Now builds one shared, frame-balanced pool across all tasks and trains a single k-means model on it (new `--kmeans_balanced_frames` option). Also made per-task labeling resumable via a completion marker, so a crashed/killed run doesn't redo tasks that already finished. |
| `TEMPLATE/asr1/pyscripts/utils/learn_kmeans.py` | **Memory fix**: feature loading always materialized a Python list of every utterance's features before `np.concatenate`-ing them (~2x final array size in peak RAM). Now does a two-pass load (scan shapes, then fill one pre-sized array) and skips a redundant copy for the single-shard case. Output is numerically identical, just lower peak memory (traded for one extra sequential disk read). |
| `voxtlm_v1/lm1/local/data_vctk.sh` + new `local/vctk/data_prep_0.92.sh` | **Data-version fix**: this recipe's TTS data prep expects CSTR VCTK Corpus **v0.92** (`wav48_silence_trimmed/` layout), but upstream's shared VCTK downloader fetches the older `wav48/`-layout corpus, which doesn't match. Now auto-downloads the correct v0.92 release directly when not already present locally, verifies the archive, and extracts it to the expected layout. |
| `voxtlm_v1/lm1/local/librilight/cut_by_vad.py` | **Correctness fix**: a single VAD-detected segment with no long pause could run 30+ minutes uncut, blowing up memory in anything doing a full-sequence forward pass downstream. Now chops any over-length VAD segment into `target_len_sec`-sized chunks instead of saving it whole. |
| `voxtlm_v1/lm1/local/data.sh` | **New feature**: `--data_config bal/3m/set` and `--speechlm_train_size`/`--textlm_train_size` options to reproduce the paper's D_Bal/D_3M/D_Set data scales via reproducible seeded subsampling, without touching ASR/TTS data. |
| `voxtlm_v1/lm1/conf/train_transformer_opt350.yaml` | **Config fix**: added a missing `model: lm_multitask` line — without it, `espnet2.tasks.lm.build_model` instantiates the single-task `ESPnetLanguageModel` instead of the multitask model VoxtLM needs, and crashes on the multitask-only `model_conf` fields (`lsm_weight`, `sos_syms`, `eos_sym`). |
| `voxtlm_v1/lm1/README.md` | **Docs**: corpus download-vs-symlink behavior, a concrete/verified per-stage command + setup table (kmeans memory routing, A100 config + OPT weight offline caching for training, sequence-length filtering to avoid O(L²) self-attention OOM, running eval stages without an A100, splitting ASR WER by test-clean/test-other), and a short list of known gaps vs. the paper's full eval suite (sWUGGY/sBLIMP, TTS CER/MOSNet). |

---

Reproducing VoxtLM's D_Bal / OPT-350M-initialized results with this recipe hit several bugs and gaps that would block anyone else trying to run it end-to-end:

- a **k-means correctness bug** affecting *any* multi-task speech-LM recipe built on `TEMPLATE/lm1`, not just this one
- an **OOM** in k-means fitting on large pools
- a **VCTK version mismatch** between the downloader and what the recipe expects
- a **missing multitask config field** that crashes training outright
- **no way to reproduce** the paper's specific data scales (D_Bal/D_3M/D_Set)

Fixing these in the shared recipe/template (rather than working around them locally) means the next person reproducing this paper — or building another multitask speech-LM recipe on `TEMPLATE/lm1` — doesn't hit the same issues.

---

## Is your PR small enough?

**Yes** — 11 files changed, 472 insertions / 36 deletions.

---

## Additional Context

- **Paper**: [VoxtLM: Unified Decoder-Only Models for Consolidating Speech Recognition/Synthesis and Speech/Text Continuation Tasks](https://arxiv.org/abs/2309.07937) (arXiv:2309.07937)
- **Environment**: developed and verified on a SLURM cluster. SLURM partition/account configs are site-specific and intentionally not included in this PR — see the README's *"A note on SLURM config files"* for what to set up.
- **Verification**: checked against the paper's D_Bal-scale, OPT-350M-initialized setup —
  - ASR WER: **6.7%** (test-clean) / **17.3%** (test-other)
  - textLM / speechLM perplexity: **27.80** / **24.93**
  - (checkpoint selected by validation accuracy, before overfitting set in — per the training curve)


sw005320:
@RuiRuihigh, please deal with [coderabbitai](https://github.com/apps/coderabbitai)'s review first

egs2/voxtlm_v1/lm1/local/vctk/data_prep_vctk0.92.sh
sw005320, line 1: egs2/voxtlm_v1/lm1/local/vctk/data_prep_0.92.sh --. egs2/voxtlm_v1/lm1/local/vctk/data_prep_vctk0.92.sh

.gitignore
```diff
@@ -73,6 +73,8 @@ egs2/**/**/asset
 egs2/**/**/checkpoint
 egs2/**/**/flagged_data_points
 egs2/**/**/hub
+egs2/voxtlm_v1/lm1/conf/slurm.conf
+egs2/voxtlm_v1/lm1/myrun.sh
```
sw005320: Please exclude this or include this pattern with a regular expression.

egs2/voxtlm_v1/lm1/conf/tuning/train_transformer_opt350.yaml
```diff
@@ -7,6 +7,7 @@ lm: transformer_opt
 lm_conf:
     opt_name: facebook/opt-350m
 
+model: lm_multitask
```
sw005320, line 10: Is it true?

RuiRuihigh: Original espnet/egs2/voxtlm_v1/lm1/conf/train_transformer_size768_e12.yaml file use this settint. I follow its setting. Besides, if I don't set the model to lm_multitask, there will be some mistakes.

sw005320:
@RuiRuihigh, please fix the CI error in the integration test

sw005320:
Thanks, @RuiRuihigh!

PR #6661, opened 2026-09-11
Add MELD speech emotion recognition recipe and CLSSystem for ESPnet3

itoten:
# Add a MELD speech emotion recognition recipe (`egs3/meld/cls`) and `CLSSystem`

Ports the merged ESPnet2 recipe `egs2/meld/cls1` to ESPnet3. Adds `CLSSystem` (`BaseSystem` + `remove_long_short` + `prepare_labels`), `egs3/meld/cls`, `egs3/TEMPLATE/cls`, and five classification metrics (WA / UA / Macro F1 / mAP / AUC). Every stage from `create_dataset` to `upload_demo` has been run.

This is still in progress. Not done yet:

- `egs3/meld/cls/readme.md`
- unit tests for the changes in `espnet3/systems/cls/{task,espnet_model,system}.py` and `espnet3/utils/task_utils.py` (the metrics are covered by `test/espnet3/systems/cls/metrics/test_metrics.py`)
- an integration test for the new system (extending `ci/test_integration_espnet3.sh`)

I ran into a number of problems along the way and made a call on each of them. **I would appreciate a review of whether those calls are the right ones**, which is why this is a draft.

Results reproduce ESPnet2. Best epoch and early-stopping epoch match, and the three validation metrics are within ±0.4.

| | WA | UA | Macro F1 | mAP | AUC |
|---|---|---|---|---|---|
| ESPnet2 | 50.81 | 28.00 | 27.75 | 28.24 | 70.31 |
| ESPnet3 | 52.72 | 25.44 | 26.17 | 28.30 | 70.05 |

ESPnet2's scoring stage reports only `mean_acc` / `mAP` / `mean_auc`, so UA and Macro F1 had to be computed by hand for the ESPnet2 PR. `measure` writes all five to `metrics.json`.

---

## What this adds

```
espnet3/systems/cls/
├── task.py           ESPnet3's own CLSTask
├── espnet_model.py   ClassificationModel (subclass of the ESPnet2 model)
├── system.py         CLSSystem
└── metrics/          five metrics

egs3/TEMPLATE/cls/    five configs + run.py + src
egs3/meld/cls/        the recipe itself
```

`task.py` starts from ESPnet2's `CLSTask` but lives on the ESPnet3 side, following the intent of reducing the dependency on ESPnet2.

`CLSSystem` in `system.py` adds two stages to `BaseSystem`. `remove_long_short` is the same one TTS uses; `prepare_labels` corresponds to stage 4 of ESPnet2's `cls.sh` and builds `token_list` from the training manifest.

`metrics/` agrees with ESPnet2's `pyscripts/utils/cls_score.py` to two decimal places on WA / mAP / AUC when both are run over the same SCP files.

`egs3/TEMPLATE/cls/` is new: there was no template for cls. It is based on the ASR one.

---

### 1. `freeze_param` is not applied, and training OOMs

**Problem:** ESPnet2 applies `freeze_param` in `main_worker()` (`espnet2/tasks/abs_task.py:1423`), which ESPnet3 replaces with Lightning, so it never runs. WavLM stays trainable and the run asks for 45.4 GB and dies.

**What I did:** Added a subclass of `ESPnetClassificationModel` at `espnet3/systems/cls/espnet_model.py` that applies the freeze in `__init__`. My understanding is that ESPnet3 keeps its stage code out of model internals, so the model owns this. `task.py` builds that class through `model_choices` — the ESPnet2 task defines `model_choices` but hardcodes the class instead of using it, so I matched what `ASRTask` does. Trainable parameters then match ESPnet2 at 1.42 M of 95.80 M.

> This is my reading of the guidance I was given. Please correct me if it is not
> what was intended.

### 2. `remove_long_short` lives under `systems/tts/`

**Problem:** CLS needs the same filtering, but the provider and runner sit in `systems/tts/`, which would mean importing TTS from `systems/cls/`.

**What I did:** Moved `remove_long_short_{provider,runner}.py` to `systems/base/`. No behavioural change. `TTSSystem`'s import and the test location follow.

### 3. The packages CLS needs are not declared

**Problem:** `collect_stats` failed with `ModuleNotFoundError: s3prl`. Looking into it:

- `espnet-s3prl` is only under the `s2st` extras, although SSL frontends are used by ASR and CLS too
- `torcheval` appears nowhere in `pyproject.toml`, yet `espnet2/cls/espnet_model.py:58` raises `ImportError` without it
- there is no `cls` extras group

So `pip install espnet` cannot build a CLS model.

**What I did:** Installed them by hand to keep going. Adding a `cls` extras group looks like the real fix — happy to include that here.

### 4. A path-valued `token_list` survives into the saved config

**Problem:** ESPnet2 replaces the path with the token list inside `build_model()` (`# Overwriting token_list to keep it as "portable".`) and dumps the config afterwards. ESPnet3 calls `save_espnet_config()` before the model is built (`training.py:101-105`), so the path is kept. `exp/<tag>/config.yaml` then holds an absolute path from the training machine and the model cannot be loaded anywhere else.

**What I did:** Inline the token list in `save_espnet_config()`, just before it writes.

> Fixing this in `pack_model` would be reasonable too, but `exp/<tag>/config.yaml`
> is read with `yaml.safe_load` (`abs_task.py:2482`), so `${recipe_dir}` is not
> interpolated and rewriting the path cannot work.

### 5. `infer` returns stale results after the model changes (unresolved)

**Problem:** After retraining, `infer` and `measure` reproduced all ten metrics from the previous run to two decimal places. `is_shard_done()` (`parallel/base_runner.py:194`) skips a shard whenever `split.N/done` exists, so the stage had only re-concatenated the old SCP files. The marker is there to resume an interrupted run, but it does not notice that the model changed, and the stage still logs `[DONE]` — no error, no warning.

`train` behaves the same way: rerun without clearing `exp_dir` and it resumes from `last.ckpt`, finishing immediately once `max_epochs` is reached.

**What I did:** Moved the experiment directory aside before each run. `pack_model` has `allow_overwrite`; `train` and `infer` have no equivalent.

### 6. `.gitignore` hides a new recipe's `dataset/` directory

**Problem:** `egs*/*/*/data*` (line 54) matches `egs3/<recipe>/<task>/dataset/`, so `git add` silently skips the recipe's dataset module. Existing recipes are unaffected because they were tracked before the rule was added, so this only shows up when creating a new recipe.

**What I did:** Used `git add -f`. Narrowing the pattern to `data` or `data_*` looks like the fix.

---

## Problems and what I did: `pack_model` through `upload_demo`

I have not been able to verify this section myself. It is written up almost verbatim from what Claude Code found. Some of it may be factually wrong, and some of it may belong in separate issues.

### 7. Demo audio never reaches the model

**Problem:** `DefaultAudioUI.build_input` (`publication/demo/assets.py:93`) creates `gr.Audio` without `type=`, so Gradio's default yields a `(sample_rate, samples)` tuple, and `session.py` passes it through unchanged. Inference classes annotate `Union[torch.Tensor, np.ndarray]`, so typeguard rejects it. `Speech2Text.__call__` carries the same annotation, so **ASR demos should fail the same way**. `ci/test_demo_ui.py` uses a stub whose expected output is `"speech=1"`, so the model boundary is never exercised.

**What I did:** Converted the value in the recipe's `src/app.py` (unwrap the tuple, downmix, cast to float32, resample to 16 kHz) rather than change shared behaviour.

### 8. Spaces are pinned to the packing interpreter's Python

**Problem:** `pack_demo` writes `python_version` from `sys.version_info` (`demo/packing.py:446`). The development setup installs Python 3.11 while `requires-python` is `>=3.12,<3.14`, so a Space packed from an environment built by following the guide cannot install espnet.

**What I did:** Pinned it through `pack.readme_context.python_version`.

### 9. Published configs keep the training machine's local paths (unresolved)

**Problem:** `_rewrite_paths_for_bundle` (`publication_utils.py:149`) rewrites only paths that were copied into the bundle, so `data_dir`, `manifest_paths` and `save_path` are published as they are. It does not affect inference, but it exposes the training machine's layout, and the bundled `conf/training.yaml` cannot be reused for training.

**What I did:** Nothing yet. Dropping the data-prep sections when publishing, or warning on paths that point outside the bundle, both seem worth considering.

---

## Is your PR small enough?

No.


egs3/meld/cls/dataset/builder.py
Masao-Someki, line 1: Could you make any public functions private (i.e., prefix them with `_`) if they are not used outside this file?

egs3/meld/cls/conf/demo.yaml
```diff
@@ -0,0 +1,59 @@
...
+model:
+  # Point at the published model repo rather than the local `model_pack`: a
+  # local directory is symlinked into the bundle and `upload_demo` resolves it,
+  # which would push the same 366 MB to the Space as well.
+  dir_or_tag: espnet/meld_cls_wavlm_base_plus
+  # The packed bundle carries this recipe's `dataset/` and `src/` modules, and
+  # loading it imports them.
+  trust_user_code: true
+
+ui:
```
Masao-Someki, line 21: I think you can add description file here?
You can write your own or just specify template file

itoten:
## Does CLS need `remove_long_short` at all?

Item 2 above asks where this stage should live. Having built ten more classification recipes since, I think the better question is whether CLS should have it.

I added it because `egs2/*/cls1/run.sh` passes `--min_wav_duration 0.1 --max_wav_duration 20`. Those are the ESPnet2 **ASR** defaults (`egs2/TEMPLATE/asr1/asr.sh:60-61`), not values chosen for emotion recognition. `egs3/TEMPLATE/asr` has no such stage, so CLS is the only task family that drops data by default.

Two reasons to doubt it:

- **The memory case is already covered.** Every recipe uses `type: folded` with `fold_lengths: [500]`, so a long utterance shrinks its own batch instead of causing an OOM.
- **The threshold costs real data at the boundary.** Across eleven recipes it drops 23 utterances, all in MELD: 14 under 0.1 s (2-4 WavLM frames, arguably corrupt) and 9 over 20 s — four of those are 20.10-20.98 s. MELD's training split is 9,988; this recipe trains on 9,969, which nobody comparing against published MELD numbers can tell without reading the recipe.

Proposal, if you agree: drop it from `DEFAULT_STAGES`, point `split_manifest_paths` at `manifest/`, and keep the stage runnable via `--stages remove_long_short` for recipes that need to screen out corrupt audio. The code stays in `systems/base/` since TTS uses it.

This changes MELD's numbers here, so I would rather hear your view first. If you prefer keeping it on, I would at least set the bounds per corpus from the measured durations instead of inheriting the ASR defaults.

sw005320:
> ## Does CLS need `remove_long_short` at all?
> Item 2 above asks where this stage should live. Having built ten more classification recipes since, I think the better question is whether CLS should have it.
> 
> I added it because `egs2/*/cls1/run.sh` passes `--min_wav_duration 0.1 --max_wav_duration 20`. Those are the ESPnet2 **ASR** defaults (`egs2/TEMPLATE/asr1/asr.sh:60-61`), not values chosen for emotion recognition. `egs3/TEMPLATE/asr` has no such stage, so CLS is the only task family that drops data by default.
> 
> Two reasons to doubt it:
> 
> * **The memory case is already covered.** Every recipe uses `type: folded` with `fold_lengths: [500]`, so a long utterance shrinks its own batch instead of causing an OOM.
> * **The threshold costs real data at the boundary.** Across eleven recipes it drops 23 utterances, all in MELD: 14 under 0.1 s (2-4 WavLM frames, arguably corrupt) and 9 over 20 s — four of those are 20.10-20.98 s. MELD's training split is 9,988; this recipe trains on 9,969, which nobody comparing against published MELD numbers can tell without reading the recipe.
> 
> Proposal, if you agree: drop it from `DEFAULT_STAGES`, point `split_manifest_paths` at `manifest/`, and keep the stage runnable via `--stages remove_long_short` for recipes that need to screen out corrupt audio. The code stays in `systems/base/` since TTS uses it.
> 
> This changes MELD's numbers here, so I would rather hear your view first. If you prefer keeping it on, I would at least set the bounds per corpus from the measured durations instead of inheriting the ASR defaults.

@itoten, too-long and too-short sentences are still harmful (too short: probably less meaningful, and too long: efficiency, even if we fold the data, respectively), but we can relax the max value from 20 s  to 30 s.

egs3/meld/cls/dataset/builder.py
```diff
@@ -0,0 +1,423 @@
...
+
+
+class MELDBuilder(DatasetBuilder):
+    """Prepare and build MELD assets for ESPnet3 recipes."""
+
+    def is_source_prepared(
+        self,
+        recipe_dir: str | Path,
+        source_dir: str | Path | None = None,
+        **_kwargs,
```
Masao-Someki, line 241: could you use `**kwargs`?

egs3/meld/cls/dataset/builder.py
```diff
@@ -0,0 +1,423 @@
...
+            (data_root / relpath).is_file()
+            for relpath in _CFG["manifest_paths"].values()
+        )
+
+    def build(
+        self,
+        recipe_dir: str | Path,
+        source_dir: str | Path | None = None,
+        **_kwargs,
+    ) -> None:
```
Masao-Someki, line 322: For this format conversion I thought we could use `espnet3.parallel` for parallel processing!

egs3/TEMPLATE/cls/run.py
```diff
@@ -0,0 +1,236 @@
...
+    # -----------------------------------------
+    logger.info("System: %s", system_cls.__name__)
+    logger.info("Requested stages: %s", args.stages)
+    logger.info("Resolved stages: %s", stages_to_run)
+
+    # Guardrail: ensure required configs exist for requested stages
+    pretrain_stages = {
+        "create_dataset",
+        "collect_stats",
+        "train",
```
Masao-Someki, line 188: I think you can add `prepare_labels` stage here, so that we can add warning when people tries to run that stage without providing any config.

espnet3/systems/base/remove_long_short_provider.py
Masao-Someki, line 1: Please copy it to the cls system! We allow repetition between systems

espnet3/systems/base/remove_long_short_runner.py
Masao-Someki, line 1: Please copy it to the cls system! We allow repetition between systems

espnet3/systems/cls/metrics/auc.py
```diff
@@ -0,0 +1,99 @@
...
+    """Compute the macro-averaged one-vs-rest AUC from class probabilities.
+
+    The area under the ROC curve is computed per class and then macro
+    averaged, matching the ``mean_auc`` column of espnet2's ``cls_score.py``.
+
+    This metric needs ``score.scp`` in addition to ``ref.scp``, which the
+    default ``ref_key``/``hyp_key`` fallback cannot express. Declare the
+    inputs in ``conf/metrics.yaml``::
+
+        - metric:
```
Masao-Someki, line 30: Could you format this code block so that we can render with sphinx with code block?

espnet3/systems/cls/metrics/macro_f1.py
```diff
@@ -0,0 +1,87 @@
...
+    single_labels,
+)
+
+
+class MacroF1(BaseMetric):
+    """Compute the macro-averaged F1 score for a classification dataset.
+
+    The F1 score of each class is computed independently and then averaged
+    without weighting by support, so rare classes are not drowned out.
+    """
```
Masao-Someki, line 21: I think we can add similar docstrings with auc.py like config example.

espnet3/systems/cls/metrics/scoring_utils.py
Masao-Someki, line 1: Please add examples to all docstrings

espnet3/systems/cls/metrics/wa.py
```diff
@@ -0,0 +1,64 @@
...
+from espnet3.components.metrics.base_metric import BaseMetric
+
+
+class WA(BaseMetric):
+    """Compute weighted accuracy (WA) for a classification dataset.
+
+    WA is the fraction of utterances whose predicted label equals the
+    reference label. It is the same quantity espnet2's ``cls_score.py``
+    reports as ``mean_acc``; the "mean" there averages one shared accuracy
+    value across classes and is a no-op.
```
Masao-Someki, line 17: Ditto let's have a lot of examples!

espnet3/systems/tts/system.py
```diff
@@ -16,10 +16,10 @@
 from espnet2.text.build_tokenizer import build_tokenizer
 from espnet2.text.cleaner import TextCleaner
 from espnet3.parallel.parallel import set_parallel
+from espnet3.systems.base.remove_long_short_provider import RemoveLongShortProvider
+from espnet3.systems.base.remove_long_short_runner import RemoveLongShortRunner
```
Masao-Someki, line 20: Please move this to cls system

test/espnet3/systems/base/test_remove_long_short.py
Masao-Someki, line 1: Please delete this diff

Masao-Someki (commented):
Thank you @itoten , please check the comments!

Masao-Someki:
Please also check CodeRabbit’s review! I think it’s a useful AI reviewer, and it’s worth revisiting the points it raised.

Masao-Someki:
Hi, based on our discussion on system naming rules, could you change the system name to `esp2_cls`?

ci/test_integration_espnet3.sh
```diff
@@ -13,7 +13,7 @@ gen_dummy_coverage(){
     ${python} empty.py
 }
 
-python3 -m pip install -e '.[asr]'
+python3 -m pip install -e '.[asr,cls]'
```
Masao-Someki: You can add `pip install -e '.[cls]'` right before the cls integration test starts, so it's line 53 or 54

egs3/meld/esp2_cls/conf/training.yaml
```diff
@@ -0,0 +1,164 @@
...
+#
+#     # NOTE(kamo): Not applying to test_sets to keep original data
+#
+# Filtering the test split would change what the reported scores are measured
+# on, so `splits` lists train and valid only.
+# The TEMPLATE default of one worker starts a cluster without overlapping
+# anything. MELD converts 13,708 clips in create_dataset, so it shards them.
+parallel:
+  env: local
+  n_workers: 8
```
Masao-Someki: Let's change this to 1, we always want to have very default value for these things

egs3/meld/esp2_cls/dataset/builder.py
```diff
@@ -185,30 +190,32 @@ def _move_first_existing(candidates: list[Path], target: Path) -> None:
     )
 
 
-def _unpack_archive(source_root: Path, archive: Path) -> None:
-    """Unpack the MELD archive into the layout this recipe expects.
+def _build_source_layout(staging: Path, archive: Path) -> None:
+    """Extract the archive under ``staging`` in the layout this recipe expects.
```
Masao-Someki: What is the rationale behind renaming source_root to staging? Since it doesn't have `*_dir` notation, it can be difficult to predict what is `staging` by just reading the code.. Basically we want to have `*_dir` for directory name/path.

Masao-Someki:
Thank you @itoten I have reviewed the new commits, and added 2 comments! After fixing these reviews I think this PR looks good to me.
But this PR becomes unexpectedly large, maybe in the future we should ask contributors to separate PRs, one for actual PR and one for integration test..

itoten:
@Masao-Someki 

> You can add pip install -e '.[cls]' right before the cls integration test starts, so it's line 53 or 54

Added it there.

> Let's change this to 1, we always want to have very default value for these things

Changed it to 1. I also checked the other values in this recipe, and none of them need to change.

> What is the rationale behind renaming source_root to staging? Since it doesn't have *_dir notation, it can be difficult to predict what is staging by just reading the code.. Basically we want to have *_dir for directory name/path.

There was no particular reason for the rename — changed staging to staging_dir.

espnet3/systems/esp2_cls/audio_conversion_runner.py
```diff
@@ -0,0 +1,142 @@
...
+from espnet3.parallel.base_runner import BaseRunner
+
+logger = logging.getLogger(__name__)
+
+
+class AudioConversionRunner(BaseRunner):
+    """Runner for converting source clips to WAV in parallel.
+
+    Conversion dominates ``create_dataset`` on corpora distributed as video:
+    MELD ships 13,708 MP4 clips, which takes over an hour one at a time. Each
```
Masao-Someki, line 18: I think this runner can be used by another recipes, let's avoid to include recipe-specific note!

espnet3/publication/demo/packing.py
```diff
@@ -184,6 +184,9 @@ def upload_demo(system) -> None:
             repo_type=repo_type,
             folder_path=str(demo_dir),
             delete_patterns=delete_patterns,
+            # Stage logs are written into the bundle directory and name the
+            # build host and its paths, which a public Space must not carry.
+            ignore_patterns=["*.log"],
```
Masao-Someki, line 189: Sorry could you remove this? Logs like training log is very useful if uploaded with a model to know what kind of environment do they use to train the model.

Masao-Someki (commented):
Thank you @itoten, I added 2 comments, other part I think it looks good to me!

PR #6662, opened 2026-09-11
Load the pretrained-vocoder config with yaml.safe_load

sw005320:
## What did you change?

`espnet2/tts/utils/parallel_wavegan_pretrained_vocoder.py` read the vocoder's `config.yml` with the full, unsafe YAML loader:

```python
config = yaml.load(f, Loader=yaml.Loader)   ->   config = yaml.safe_load(f)
```

Plus a regression test.

`yaml.Loader` honours tags such as `!!python/object/apply`, so a `config.yml` shipped alongside a vocoder checkpoint executes arbitrary code the moment it is parsed. Unlike `FullLoader`, this is not gated on any PyYAML version.

The reachable path is the ordinary one: `espnet2/bin/tts_inference.py` takes a `vocoder_tag`, passes it to `download_pretrained_model`, and constructs this class from whatever bundle comes back. So the threat model matches the checkpoint-deserialization case fixed earlier — a user points ESPnet at a pretrained model from a hub or a collaborator, and the loader runs code on their behalf.

`build_model_from_file` in `espnet2/tasks/abs_task.py` already used `yaml.safe_load`; this vocoder loader was the outlier among the model-loading paths.

## Is it a drop-in?

Yes. The loader reads scalar fields (`config["sampling_rate"]`) and passes the mapping to `parallel_wavegan.utils.load_model`. Real vocoder bundles ship plain YAML, which `safe_load` parses identically — verified against PyYAML 6.0.1:

```
safe_load("sampling_rate: 24000\nformat: hdf5\n")          -> {'sampling_rate': 24000, 'format': 'hdf5'}
safe_load(... !!python/object/apply:os.system ['echo x'])  -> ConstructorError
```

A config carrying a python tag now raises `yaml.constructor.ConstructorError` (a `yaml.YAMLError`) instead of executing.

## Test

`test/espnet2/tts/utils/test_parallel_wavegan_pretrained_vocoder.py` covers both directions — a config with a python tag is refused, and a plain config still loads and reaches `load_model` unchanged. The optional `parallel_wavegan` dependency is stubbed, because the import inside `__init__` has to succeed before the config read is reached.

**The test has not been run locally** — torch is not installed in the environment this was written in, so CI is its first execution. `black`, `isort` and `flake8` are clean.

## Not changed here

`Loader=yaml.Loader` also appears in `espnet2/utils/nested_dict_action.py`, `utils/change_yaml.py` and three `egs2` recipe scripts. Those parse the user's own CLI input or recipe-local files rather than a downloaded bundle, so they are a different risk class; keeping this PR to the one reachable model-loading path makes it easier to review and to back-port. Worth a separate cleanup.

### DIFF
--- a/espnet2/tts/utils/parallel_wavegan_pretrained_vocoder.py
+++ b/espnet2/tts/utils/parallel_wavegan_pretrained_vocoder.py
@@ -34,7 +34,7 @@ class ParallelWaveGANPretrainedVocoder(torch.nn.Module):
             dirname = os.path.dirname(str(model_file))
             config_file = os.path.join(dirname, "config.yml")
         with open(config_file) as f:
-            config = yaml.load(f, Loader=yaml.Loader)
+            config = yaml.safe_load(f)
         self.fs = config["sampling_rate"]
         self.vocoder = load_model(model_file, config)
         if hasattr(self.vocoder, "remove_weight_norm"):
--- a/test/espnet2/tts/utils/test_parallel_wavegan_pretrained_vocoder.py
+++ b/test/espnet2/tts/utils/test_parallel_wavegan_pretrained_vocoder.py
@@ -1,0 +1,59 @@
+import sys
+import types
+
+import pytest
+import torch
+import yaml
+
+from espnet2.tts.utils.parallel_wavegan_pretrained_vocoder import (
+    ParallelWaveGANPretrainedVocoder,
+)
+
+
+def _stub_parallel_wavegan(monkeypatch):
+    """Stand in for the optional ``parallel_wavegan`` dependency.
+
+    The loader imports it inside ``__init__``, so that import has to succeed
+    before the config read -- the thing under test -- is reached at all.
+    """
+    captured = {}
+
+    def load_model(model_file, config):
+        captured["config"] = config
+        return torch.nn.Module()
+
+    utils = types.ModuleType("parallel_wavegan.utils")
+    utils.load_model = load_model
+    package = types.ModuleType("parallel_wavegan")
+    package.utils = utils
+    monkeypatch.setitem(sys.modules, "parallel_wavegan", package)
+    monkeypatch.setitem(sys.modules, "parallel_wavegan.utils", utils)
+    return captured
+
+
+def test_config_with_python_tag_is_refused(tmp_path, monkeypatch):
+    """A vocoder bundle must not execute code through its config.
+
+    ``yaml.Loader`` honours tags such as ``!!python/object/apply``, so a
+    downloaded vocoder could run arbitrary code at load time. ``safe_load``
+    refuses them.
+    """
+    _stub_parallel_wavegan(monkeypatch)
+    (tmp_path / "config.yml").write_text(
+        "sampling_rate: 24000\n"
+        "payload: !!python/object/apply:os.system ['echo pwned']\n"
+    )
+
+    with pytest.raises(yaml.YAMLError):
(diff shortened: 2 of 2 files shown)


sw005320:
For the record now that the advisory is public: this fixes GHSA-rvjf-ppg9-9rwg, reported by @ChristianTeroerde on 2026-06-15. Shipped in `202609.post2`, on PyPI as of today.

Apologies to @ChristianTeroerde for the 88 days. The report was accurate and specific, and the delay was entirely on our side. The advisory has no comment thread left now that it is published, so the thanks go here instead.

Worth recording that the report's framing shaped more than this one line. It pointed out that `build_model_from_file` in `espnet2/tasks/abs_task.py` already used `yaml.safe_load` and that this vocoder loader was *the outlier* — which prompted a sweep of the rest of the repository in #6664. That turned up one more in shipped code: `espnet2/utils/nested_dict_action.py`, the argparse action behind `--*_conf` on every espnet2 task, was calling plain `eval()` on its value. Both fixes are in `202609.post2`.

The published advisory lists the affected range as `<= 202609.post1` rather than the `<= 202604` on the original report, since 202609 and 202609.post1 shipped after it was filed and were still affected. The reporter's CVSS of 8.8 with `UI:R` was kept unchanged — it was the right call, and more accurate than the 9.8 that auto-fill had produced on a sibling advisory.

ChristianTeroerde:
Hi Shinji,

Thank you for fixing this, publishing the advisory, and crediting my report. I appreciate your acknowledgment of the delay and the explanation.

I’m glad the comparison with the existing `safe_load` path helped prompt the wider review and uncover the additional issue in `nested_dict_action.py`. It’s good to see both fixes shipped in `202609.post2`. Thanks also for updating the affected version range and retaining the 8.8 rating with user interaction required.

One follow-up: has a CVE ID been requested for GHSA-rvjf-ppg9-9rwg? If not, could you please request one through GitHub?

Best,
Christian

sw005320:
Hi Christian,

Yes — a CVE has been requested for GHSA-rvjf-ppg9-9rwg, and I have just re-confirmed it rather than taking my own word for it. Thank you for asking; the check was worth running.

Being precise about what I can and cannot tell you: the REST API exposes only `cve_id` on an advisory, with no field indicating that a request is in flight, so "requested" is not directly observable. I re-issued the request and it was accepted, so the request is definitely in now. What I cannot establish retroactively is whether the earlier one registered — if both did, GitHub de-duplicates them.

Your question also turned up something I had not checked. GHSA-jcfw-5whq-98j4, the tar-slip advisory published on 2026-09-04, still had no CVE a week on, and I had not requested one for it in this round at all. I have now requested CVEs for all three published advisories:

- GHSA-rvjf-ppg9-9rwg — yours
- GHSA-64f6-3gqc-r926 — `torch.load` deserialization
- GHSA-jcfw-5whq-98j4 — `pack_funcs.unpack` tar-slip

GitHub is the CNA here, so there is no separate filing; the IDs will attach to the advisories automatically once assigned. If any of them is still showing no CVE in a couple of weeks, please do prod me again — that turned out to be the useful thing about your message.

Best,
Shinji

ChristianTeroerde:
Hi Shinji,

Thank you for checking this so carefully and making sure the CVE requests are in place. I really appreciate the transparency, the fixes, and the public acknowledgment.

I’m glad my follow-up also helped get the other advisories covered. It’s been a pleasure working with you on this. I’ll keep an eye on the advisory and check back in a couple of weeks if the ID hasn’t appeared.

Thanks again for following through!

Best,
Christian

sw005320:
Thanks, @ChristianTeroerde!

PR #6667, opened 2026-09-11
Summarise an ended hypothesis for end detection once, not at every step

sw005320:
Found while checking batched decoding for the SPGISpeech recipe (#6656) on top of #6658.

## What did you change?

Every search called `end_detect([h.asdict() for h in ended_hyps], i)` at every decoding step. `end_detect` only needs each hypothesis's length and score, and an ended hypothesis never changes, but `asdict` reads `yseq.tolist()`, `float(score)` and one `float` per scorer, each a device-to-host synchronisation, for **every ended hypothesis at every step**. The cost of a step therefore grew with the number of hypotheses that had ended so far, i.e. with the utterance length and with the batch size.

`BeamSearch.end_detected(ended_hyps, i)` computes a `{yseq, score}` summary once per hypothesis and caches it. The cache is shared by every call, because `BatchBeamSearch.post_process` calls it once per utterance within a step and the utterances must not evict each other; an entry is dropped once its hypothesis has not been seen for a whole step, which frees the cache at the end of a decode and also copes with a caller that builds a fresh list every step (`PartiallyARBeamSearch`). The entry holds the hypothesis itself, so a recycled `id()` cannot alias a new one. All five searches use it: `BeamSearch`, `BatchBeamSearch`, `BatchBeamSearchOnline`, `BatchBeamSearchOnlineSim`, `PartiallyARBeamSearch`. `end_detect` in `e2e_asr_common` is untouched.

Profile of a `batch_size=16` decode, model-zoo SPGISpeech conformer, beam 20, 32 utterances sorted by length, MPS. Per decoding step:

| | master | this PR |
|---|---:|---:|
| `Hypothesis.asdict` calls, whole decode | 10,368 | 0 (2,311 one-time summaries, one per ended hypothesis) |
| `post_process` | 166-283 ms | 45 ms |
| of which end detection | 138-224 ms | 17 ms |
| whole decode of the 32 utterances | 26.8-45.0 s | 20.7 s |
| transcripts | | identical, 32/32 |

The master range is because this laptop's timings drift by 1.5-2x with heat and memory pressure between runs; the call count is the reliable number. What remains in end detection is the one-time conversion of each hypothesis as it ends (two synchronisations each); folding those into the single per-step read `post_process` already does for `is_eos` would remove them too, and is a follow-up.

At `batch_size 1` the same thing happens on a smaller scale: the ended list grows over the utterance, and every step re-reads all of it, so long utterances pay most.

## Is your PR small enough?

5 library files, +45 / -12, plus one test. Yes.

## Additional Context

**Testing.** `test_beam_search.py::test_end_detected_converts_each_ended_hypothesis_once`: two utterances decoded side by side over eleven steps, one or two hypotheses ending per step; `end_detected` returns exactly what `end_detect` over `asdict` returns, every hypothesis of both utterances stays cached (neither evicts the other), the next step reuses the very same summary objects without a single `tolist`, and hypotheses absent for a whole step are forgotten. The frozen-reference tests of #6599 (221 configurations at `rtol=atol=0`), the batched parity tests, the online and partially-AR tests pass unchanged: the decision `end_detect` makes is computed from the same numbers as before.


sw005320:
Both review findings addressed in b88641f.

- **Cache scope.** Correct, and it was worse than a lost optimisation: with one utterance per call evicting the others, the summaries were rebuilt every step after all, which is why the branch profile showed 94 ms in end detection against 34 ms for the prototype. The cache is now shared across calls and an entry is dropped only once its hypothesis has not been seen for a whole step. Re-profiled at batch 16: end detection 17 ms/step (master 138-224 ms), `post_process` 45 ms (master 166-283 ms), whole decode 20.7 s against 26.8-45.0 s, transcripts identical 32/32. PR description updated with these numbers.
- **Test.** It now decodes two utterances side by side, asserts every hypothesis of both stays cached, that the following step reuses the same summary objects with zero `tolist` calls, and that hypotheses absent for a step are forgotten.

sw005320:
@chenehk, can you also review this PR?

chenehk (approved):
I've reviewed the implementation and the test. Looks good to me!

sw005320:
@chenehk, thanks!

PR #6670, opened 2026-09-13
fix(asr): rebuild Conv2dSubsamplingWOPosEnc mask from ilens

modelpath-dev:
## Why

`Conv2dSubsamplingWOPosEnc` still sliced the pad mask from the padded batch width. A 288-frame utterance reported 72 frames next to a 363-frame one and 71 frames alone. The other eight Conv* classes already rebuild the mask from each row's own length through `_subsample_mask`. This class is the embed layer for both contextual-block encoders, so they inherited the same leak.

## Scope

- `Conv2dSubsamplingWOPosEnc.forward` now calls `_subsample_mask` with `zip(self.kernels, self.strides)`.
- Tests pin the three encoder configs from the #6633 review (`conv2d` 71, `conv2d6` 47, `conv2d8` 35) and a `ContextualBlockTransformerEncoder` path with `block_size=0`.

Out of scope. Time-convolution padding on contextual-block layers, which #6660 left as-is.

## Tradeoffs

Importing `_subsample_mask` from the sibling file is a private import. A new shared module would own the helper more cleanly and would also move the eight existing call sites. This PR keeps the helper where #6633 put it.

## Blast Radius

Streaming contextual-block ASR recipes that use `input_layer=conv2d|conv2d6|conv2d8`. Short utterances in a mixed-length batch can move by one frame. The longest row and any decode with `batch_size=1` stay the same.

## Verification

```
PYTHONPATH=. python -m pytest -o addopts= \
  test/espnet2/legacy/test_subsampling.py::test_woposenc_olens_is_independent_of_batch_mates \
  test/espnet2/legacy/test_subsampling.py::test_woposenc_contextual_block_encoder_olens \
  test/espnet2/legacy/test_subsampling.py::test_olens_is_independent_of_batch_mates \
  test/espnet2/legacy/test_subsampling.py::test_issue_6600_transformer_encoder_repro
```

13 passed. Live 288/363 table after the fix: conv2d 71/71, conv2d6 47/47, conv2d8 35/35.


sw005320:
Verified locally on the PR head (4bc3281):

- `_subsample_mask` consumes `convs` once per call and the PR builds a fresh `zip` on every `forward`, so repeated calls give the same lengths (checked: `[90, 71]` twice on the same batch).
- The streaming path is untouched by construction: `forward_infer` in both contextual-block encoders calls the embed layer with `x_mask=None`, which still returns `(x, None)`. Only `forward_train`, i.e. training and non-streaming decoding, sees the new lengths.
- The contextual-block **conformer**, which the new test does not cover, inherits the fix for all three geometries: 288 alone / 288 beside 363 gives 71/71 (`conv2d`), 47/47 (`conv2d6`), 35/35 (`conv2d8`), matching the conv formula.
- `test_subsampling.py`, `test_contextual_block_transformer_encoder.py` and the conformer encoder tests pass (65 passed, 16 skipped).

With this the set from the #6633 review is complete: every mask-slicing subsampling class in the tree now derives `olens` from `ilens`. CodeRabbit's note about test placement is a style nit; not blocking from my side.

modelpath-dev:
Thanks Shinji.
  Pushed a small follow-up: split the tests out the way CodeRabbit suggested, and added a conformer olens case for the path you checked. Commit is 2f3d45d.

sw005320 (approved):
Thanks, @modelpath-dev!

PR #6671, opened 2026-09-14
Add ESPnet3 SVS system and ACE-Opencpop VISinger 2 recipe

Ting-Justin-Jiang:
## What did you change?

Add the ESPnet3 singing voice synthesis recipe: `egs3/acesinger/svs`,
followed `egs2/acesinger/svs1` (VISinger 2 on ACE-Opencpop).

ESPnet3 had no SVS support, so the PR adds three things:

1. `espnet3/systems/svs/`: the SVS system, GAN training (generator and
   discriminator take turns, as in ESPnet2's `GANTrainer`), and the usual
   SVS metrics (MCD, log-F0 RMSE, semitone accuracy, V/UV error).
2. `egs3/TEMPLATE/svs/`: the SVS template, laid out following the ASR template.
3. `egs3/acesinger/svs/`: the recipe itself. The data comes from
   `espnet/ace-opencpop-segments` on the Hub.

The model, preprocessor, collate function and inference class are directly from the ESPnet2

Two small changes to existing ESPnet3 code: GAN models need their own
training step, so a system can now pass one in; and the stats stage can save
the features it extracts (fbank and pitch), so training does not recompute
them. Both mirror what the egs2 recipe already does.

To have an SVS recipe on ESPnet3. ACE-Opencpop already has a working egs2
recipe and its segmented data is on the Hub, so it is the natural first one.

## Is your PR small enough?

No: 39 files, about 2,500 lines. Most of it is configs (the VISinger 2 config
is 230 lines), the template copied from `egs3/TEMPLATE/asr`, the dataset
package and tests. I can split it into the system/template part and the
recipe part if preferred.

## Additional Context

Results

| dataset | MCD | log-F0 RMSE | Semitone ACC (%) | VUV error (%) |
| --- | --- | --- | --- | --- |
| test | 6.10 | 0.183 | 56.7 | 9.2 |


sw005320 (commented):
- Please replace the metrics by calling VERSA. @ftshijt, please help.
- @Masao-Someki, please instruct for our recent module changes. This will affect his design including systems and test

egs3/acesinger/svs/dataset/__pycache__/__init__.cpython-312.pyc
Masao-Someki, line 1: Please remove this file

egs3/acesinger/svs/dataset/__pycache__/builder.cpython-312.pyc
Masao-Someki, line 1: ditto

egs3/acesinger/svs/dataset/builder.py
```diff
@@ -0,0 +1,178 @@
...
+                    row["transcription"],
+                    int(row["singer"]),
+                    _format_label(row),
+                    _format_score(row),
+                )
+            )
+    return rows
+
+
+class ACEOpencpopBuilder(DatasetBuilder):
```
Masao-Someki, line 101: Simple question, is there any reason we want to convert HF dataset into wav file? I thought we could just use it for training as well!

egs3/acesinger/svs/dataset/dataset.py
```diff
@@ -0,0 +1,251 @@
...
+    obj = json.loads(score)
+    return int(obj["tempo"]), obj["note"]
+
+
+# ---------------------------------------------------------------------------
+# Public dataset class
+# ---------------------------------------------------------------------------
+
+
+class ACEOpencpopDataset(TorchDataset):
```
Masao-Someki, line 109: This dataset class as well I think it could be simpler to use HF dataset

egs3/TEMPLATE/svs/conf/metrics.yaml
Masao-Someki, line 1: I think we can have default metrics with the one we have in the `systems/`

egs3/TEMPLATE/svs/conf/training.yaml
```diff
@@ -0,0 +1,107 @@
...
+##########################################################
+# Single optimizer path:
+optimizer:
+scheduler:
+scheduler_interval:
+scheduler_monitor:
+
+# Multiple-loss / multiple-optimizer path (e.g. GAN-based SVS):
+optimizers:
+schedulers:
```
Masao-Someki, line 61: Let's add a template optimizer/scheduler config?

egs3/TEMPLATE/svs/conf/training.yaml
```diff
@@ -0,0 +1,107 @@
...
+
+##########################################################
+#                OTHER TRAINING SETTING                  #
+##########################################################
+best_model_criterion:
+
+# Optional random seed used by collect_stats/train.
+seed:
```
Masao-Someki, line 69: I think it would be fine just setting default values here also?

espnet3/systems/base/training.py
```diff
@@ -74,8 +76,16 @@ def collect_stats(config: DictConfig) -> None:
...
+def train(config: DictConfig, module_cls=None) -> None:
+    """Run the training loop.
+
+    Args:
+        config: Training configuration.
+        module_cls: ``ESPnetLightningModule`` subclass that wraps the model;
+            the base module when ``None``. Systems whose models need a
+            different step, such as the GAN generator/discriminator turns of
+            SVS, pass their own class here.
+    """
```
Masao-Someki, line 88: (just comment) Thanks, this looks nice. Could you add examples on module_cls? For example, we can have 2 code snippets with and without module_cls

espnet3/systems/svs/metrics/singing.py
```diff
@@ -0,0 +1,267 @@
...
+    Example:
+        ```yaml
+        metrics:
+          - metric:
+              _target_: espnet3.systems.svs.metrics.singing.SingingMetrics
+              num_workers: 16
+            inputs:
+              wav: wav
+              ref: ref
+        ```
```
Masao-Someki, line 158: Could you format this block so that we can render it from sphinx?

test/espnet3/systems/base/test_training.py
Masao-Someki, line 1: Could you also add unit test to confirm we can use the custom lightning modules?

Masao-Someki (commented):
Thanks!
For several libraries like pyworld, you can add it in the pypeoject as svs group. Please include that diff in this PR!

espnet3/systems/svs/metrics/singing.py
Masao-Someki, line 1: Could you split the metrics file into individual metrics? We want to make it consistent, and usually other recipes we have one class for each metrics, like `WER`, `PESQ`, etc.
We should be able to split this class to MCD, LogF0RMSE, SemitoneAcc, VUVErr. Or if you want to use versa, that is also fine.

Masao-Someki (commented):
I added one comment on metrics

espnet3/systems/svs/metrics/singing.py
sw005320: We should use VERSA
@ftshijt, could you suggest which configuration we should use?

Masao-Someki: @Ting-Justin-Jiang you might check [this file](https://github.com/NewGamezzz/espnet/blob/d3839bc5c13a6f6aa5c742b2fac33f35b023945a/egs3/libritts/tts/src/metrics/versa.py) from Thanapat.
It's in draft PR and we're still not reviewed this file, but he implemented the versa metrics.

Masao-Someki:
Hi, based on our discussion on system naming rules, could you change the system name to `esp2_svs`?

Ting-Justin-Jiang:
Thank you @Masao-Someki and @sw005320 for reveiwing. I will incorporate your feedbacks in system design and evaluation metrics.

PR #6673, opened 2026-09-14
Fix mask indexing and batching in partially AR inference

### LINKED ISSUE #6669: Four defects in partially_AR_model.py surfaced by CodeRabbit on the line-ending PR
While #6668 normalised the line endings of `espnet2/asr/partially_AR_model.py`, CodeRabbit reviewed the (unchanged) code and reported four findings. #6668 is line endings only, so they are recorded here for a separate fix. The code has been this way since the file was added; `git diff --ignore-space-at-eol` on #6668 is empty for this file.

The one marked critical is a wrong loop variable (`i` instead of `m` in `_get_mask_idx`), which would make every mask in a batch resolve to the position of the first one; that needs to be verified against a real partially-AR decode before it is fixed, since `test_partially_AR.py` passes as things stand.

### `espnet2/asr/partially_AR_model.py` line 150

_🎯 Functional Correctness_ | _🟠 Major_ | _⚡ Quick win_

**`max_mask_parallel` is overwritten on the first call and keeps a tensor value.**

`mask_num` is a 0-dim tensor from line 143. Line 150 stores `mask_num + 1` into the instance attribute, so the sentinel `-1` is lost after the first utterance. Every later utterance reuses the first utterance's mask count as the batch cap. An utterance with more masks is then split into batches without the user requesting it, which makes decoding results depend on input order.

Use a local variable and an `int`.

### `espnet2/asr/partially_AR_model.py` line 156

_🩺 Stability & Availability_ | _🟠 Major_ | _⚡ Quick win_

**The batch loop runs one empty iteration when `mask_num` divides evenly.**

`range((mask_num // self.max_mask_parallel) + 1)` always adds a trailing iteration. If `mask_num` is a multiple of `max_mask_parallel`, that iteration gets `bs_iter == mask_num` and `max_iter == 0`. No mask is registered, and line 175 calls the beam search with an empty `masks` list. `PartiallyARBeamSearch.init_hyp` then fails on `assert len(self.masks) > 0, "add_mask must be called before init_hyp"`. Example: `max_mask_parallel=2` with 4 masks.

Use ceiling division.

### `espnet2/asr/partially_AR_model.py` line 160

_🎯 Functional Correctness_ | _🔴 Critical_ | _⚡ Quick win_

**Wrong loop variable: every mask in a batch resolves to the same position.**

The inner loop iterates `m`, but line 160 passes the outer batch index `i` to `_get_mask_idx`. Ruff also reports `m` as unused. With the default `max_mask_parallel=-1` the outer loop runs a single iteration with `i=0`, so all registered masks use the position of the first mask token. `prev_tokens`, `next_token`, and the primer are then identical for every mask, and `forward` fills all masked slots from the same context.

Pass the per-mask index instead.

</details>

### `espnet2/asr/partially_AR_model.py` line 172

_🎯 Functional Correctness_ | _🟠 Major_ | _⚡ Quick win_

**Pass a scalar EOS value on both branches.**

When `mask_idx` is last, `[self.eos]` makes `torch.LongTensor([m[1]])` a `(1, 1)` tensor. `init_hyp` assigns it to the scalar `eoses[i]`, which can fail because in-place assignment cannot expand the target.

```diff
                 next_token = (
-                    yseq_with_mask[0, mask_idx + 1].tolist()
+                    int(yseq_with_mask[0, mask_idx + 1])
                     if mask_idx < len(yseq_with_mask[0]) - 1
-                    else [self.eos]
+                    else self.eos
                 )
```

_Source: Linters/SAST tools_

meryemsakin:
## What did you change?

Fixes the four defects reported in #6669 in `espnet2/asr/partially_AR_model.py`, plus one related defect found while verifying them:

1. **Wrong loop variable**: `_get_mask_idx(yseq_with_mask, i)` → `m`. Previously every mask in a batch was registered with the context of the first mask.
2. **Empty trailing batch**: the batch loop now steps with `range(0, mask_num, max_mask_parallel)`, so e.g. `max_mask_parallel=2` with 4 masks no longer runs an empty batch that hits `assert len(self.masks) > 0`.
3. **`max_mask_parallel` mutation**: the `-1` default is resolved into a local int per call; the attribute is no longer overwritten with the first utterance's tensor mask count.
4. **EOS shape**: a trailing mask passes `self.eos` instead of `[self.eos]`, matching the scalar used on the other branch.
5. **Hypotheses written into unmerged masks** (not in the issue): beam search returns one hypothesis per *merged* mask span, but they were written back into the *unmerged* `y_in`. With adjacent masks, the next span's hypothesis landed in the leftover slot of the previous span and a `<mask>` token remained in the output. The result now collapses only consecutive mask tokens (repeated confident tokens are kept).

Adds `test/espnet2/asr/test_partially_ar_masks.py` using the real `TransformerDecoder` and `PartiallyARBeamSearch` with controlled CTC probabilities. It checks the registered mask contexts and EOS values, caps of -1/1/2/3 (including evenly divisible counts), repeated calls on the same instance, and hypothesis placement with merged masks and repeated tokens.

---

Reported in #6669. The critical finding was to be verified against a real partially-AR decode before fixing, so I checked each one:

- **Tests**: the new tests fail on `master` (10 failed) and pass with this change. `test_partially_AR.py` still passes.
- **Real speech**: `espnet/shinji-watanabe-librispeech_asr_train_asr_transformer_e18_raw_bpe_sp_valid.acc.best` on LibriSpeech `5703-47212-0000`, CPU, threshold 0.99, beam 5, `max_mask_parallel=-1`, same encoder output for every version:
  - `master`: all 8 masks are registered with the identical context; output keeps 1 `<mask>` token. Text: `... SCARLET SMOKE ANDTACK THE INVERASHI ANDL ... DURING AND SUMMER MONTHS AND ANDD UP AND DOWN ... BETWEEN INVERASH ANDI ANDL AND CRIAN`
  - this PR: each mask gets its own context; no `<mask>` left. Text: `... SCARLET SMOKESTACK THE INVERASHIEL ... DURING THE SUMMER MONTHS PLIED UP AND DOWN ... BETWEEN INVERASHIEL AND CRIANAN`
- The EOS change is for consistency: I could not reproduce the reported assignment failure with torch 2.11, where `eoses[i] = torch.LongTensor([[eos]])` still works.

This is a single-utterance check, not a WER benchmark.

---

## Is your PR small enough?

Yes. 2 files, +118/−9.

---

## Additional Context

- Fixes #6669
- Local checks: black 26.5.1, isort, pycodestyle, flake8 on the changed files. `test_partially_ar_masks.py`, `test_partially_AR.py`, `legacy/test_batch_beam_search.py`, and `legacy/test_beam_search.py` gave 37 passed and 150 skipped (CUDA and unsupported dtype/decoding combinations) on macOS arm64, Python 3.13, torch 2.11.
- Not measured: GPU and corpus-level WER.


sw005320 (approved):
Thanks a lot, @meryemsakin!

PR #6498, opened 2026-07-23
[openbeats-3] BEATs AudioSet-2M pretraining recipe

jaeyeonkim99:
## What did you change?

Added the `egs2/audioset/ssl1` recipe that runs the BEATs iterative pre-training loop (from #6437) on AudioSet-2M.

- `run.sh` driving `beats.sh` over the `iter0 encoder → iter1 tokenizer → iter1 encoder` loop.
- Training configs: `conf/beats_base.yaml` (encoder), `conf/tok_beats_base.yaml` (VQ tokenizer), `conf/as2m_inf.yaml` (tokenizer inference), and the matching DeepSpeed configs.
- `local/data.sh` + `local/data_prep_as2m.py` (multiprocessing AudioSet manifest prep).
- `README.md` with setup, usage, and downstream RESULTS.
- Released the pre-trained encoder/tokenizer checkpoints on HuggingFace ([iter0](https://huggingface.co/jaeyeonkim99/BEATs-base-AS2M-iter0), [iter1](https://huggingface.co/jaeyeonkim99/BEATs-base-AS2M-iter1), [tokenizer](https://huggingface.co/jaeyeonkim99/BEATs-tokenizer-AS2M-iter1)).

---

## Why did you make this change?

The per-corpus recipe promised in #6437. It gives a runnable, reproducible AudioSet-2M BEATs pre-training recipe on top of the merged training stack, with released checkpoints and downstream numbers so the iterative pre-training can be reproduced end-to-end.

---

## Is your PR small enough?

Mostly. 22 files, ~780 lines, all recipe scripts/configs (no core code); most of the file count is standard recipe symlinks and config files that must ship together.

---

## Additional Context

* #6052
* #6436
* #6437

PR #6674, opened 2026-09-15
Add BEATs SSL recipe for ESPnet3

jaeyeonkim99:
## What did you change?

Ports the BEATs iterative pre-training recipe from #6498 (`egs2/audioset/ssl1`) to ESPnet3.

- **`espnet3/systems/ssl/`**: `BeatsSystem` runs one BEATs iteration per `run.py` invocation (`training_config.iteration`), using the standard stage names:
  - `train_tokenizer`: no-op at iteration 0. At iteration N > 0 it distills the VQ tokenizer from the iteration N-1 encoder and exports `beats_tokenizer_iter<N>.pt`.
  - `infer`: tokenizes the corpus (random projection at iteration 0, trained tokenizer after that) into index-keyed `target.scp` + `target_shape`.
  - `collect_stats` / `train`: trains the encoder and exports `beats_encoder_iter<N>.pt`, averaged over the top-K checkpoints.
  - Supporting modules: `BeatsTokenizationModel`, `BeatsTargetReader` (byte-offset index, so 2M-utterance target files are not loaded into every worker), `export_beats_checkpoint`.
- **`egs3/TEMPLATE/ssl/`**: runner plus default encoder / tokenizer / tokenization configs. The runner copies `iteration`, `teacher_ckpt_path`, fbank stats, `waveform_input` and data paths from the training config into the other two configs.
- **`egs3/audioset/ssl/`**: AudioSet-2M recipe, override-only configs. The dataset builder is ported from `local/data_prep_as2m.py` plus the egs2 duration filter. The dataset can also read existing egs2 fbank dumps (`feats_path`).
- **`egs3/mini_an4/ssl/`** + `ci/test_integration_espnet3.sh`: tiny BEATs config that runs iteration 0 and iteration 1 end to end in CI.
- **`espnet3/systems/base/training.py`**: `freeze_param` support, with the same semantics as ESPnet2 (needed to freeze the tokenizer teacher).
- **`espnet2/tasks/beats.py`**: `BeatsTokenizerTask` builds the teacher with the tokenizer's `fbank_mean`/`fbank_std` instead of the library defaults. There is no change for existing egs2 configs.
- **`espnet2/beats/tokenizer.py`**: `NormEMAVectorQuantizer` now checks for an initialized process group on every all-reduce instead of once at construction. Lightning builds the model before DDP starts, so the codebook EMA previously used per-rank statistics; egs2 (process group initialized before model build) behaves as before.
- **`espnet3/utils/stages_utils.py`**: `train_tokenizer` gets the same per-rank stage-log handling as `train`, so multi-GPU ranks do not rotate rank 0's log.

Mapping from egs2 → ESPnet3:

| egs2 `beats.sh` | ESPnet3 |
|---|---|
| stages 1–3 (data prep, format, `--max_wav_duration 11`) | `create_dataset` |
| stage 4 (fbank dump) | on-the-fly fbank (`waveform_input: true`), or reuse dumps via `feats_path` |
| stage 5 / `tokenizer_inference` | `infer` |
| stage 6 | `collect_stats` |
| stage 7 (`train_tokenizer`, `train_encoder`, `generate_checkpoint`) | `train_tokenizer`, `train` |
| DeepSpeed ZeRO-0 bf16 + `WarmupDecayLR` | Lightning DDP `bf16-mixed` + `PiecewiseLinearWarmupLR` (same curve; final LR held after 400K steps) |
| global `batch_bins: 420000` on 2 GPUs | per-device `batch_bins: 210000` (ESPnet3 does not split batches across ranks) |

---

Brings the AudioSet-2M BEATs pre-training recipe (and the OpenBEATs pipeline behind it) to ESPnet3.

Porting also surfaced an egs2 config issue. Stage 4 writes the corpus fbank statistics into the configs correctly, but in `conf/beats_base.yaml` / `conf/tok_beats_base.yaml` they sit inside `beats_config` / `tokenizer_config`. `BeatsEncoder` only takes `fbank_mean`/`fbank_std` as constructor arguments (`encoder_conf`), so those values are stored but never used. As a result, egs2 encoder/tokenizer training and teacher loading normalized with the library defaults (15.29130 / 5.90532), and only iteration-0 tokenization (`conf/as2m_inf.yaml`, top-level keys) used the AudioSet values (15.66439 / 6.38312). The released egs2 checkpoints are consistent with the defaults that downstream recipes use, so this PR does not change the egs2 configs.

In ESPnet3 the AudioSet statistics are set once and passed as `encoder_conf` arguments to the encoder, tokenizer, teacher and tokenization. Checkpoints trained with this recipe therefore need `encoder_conf.fbank_mean/fbank_std` set for downstream fine-tuning (documented in the recipe readme). The training-log comparison below uses the egs2 effective values so the comparison is like-for-like.

---

## Verification
- Tokenization equivalence: ESPnet3 `BeatsTokenizationModel` vs egs2 iteration-0 targets on 200 AudioSet eval clips: 99.98% token agreement (179/200 identical; remaining diffs are CPU vs GPU float ties).
- Training-log comparison vs the egs2 run `beats_iter0_base.as2m_v2`: same fbank dumps, iter0 targets, shape files and global batch (2×A40, bf16-mixed, first 2,848 steps)

  | batch | loss (egs2 / ESPnet3) | acc_mask | n_uniq_pred_msk | count_masked / GPU | lr |
  |---|---|---|---|---|---|
  | 356 | 3.305 / 3.250 | 0.534 / 0.535 | 2.66 / 2.59 | 5.245e4 / 52,452 | 1.272e-5 / 1.275e-5 |
  | 712 | 3.055 / 3.053 | 0.541 / 0.541 | 2.16 / 2.15 | 5.245e4 / 52,451 | 1.817e-5 / 1.820e-5 |
  | 1068 | 3.026 / 3.015 | 0.548 / 0.550 | 2.67 / 2.93 | 5.245e4 / 52,452 | 2.362e-5 / 2.365e-5 |
  | 1424 | 2.953 / 2.950 | 0.565 / 0.566 | 6.25 / 6.62 | 5.245e4 / 52,451 | 2.907e-5 / 2.910e-5 |
  | 1780 | 2.868 / 2.862 | 0.581 / 0.583 | 9.82 / 9.61 | 5.245e4 / 52,451 | 3.452e-5 / 3.455e-5 |
  | 2136 | 2.797 / 2.789 | 0.596 / 0.598 | 11.16 / 11.09 | 5.243e4 / 52,452 | 3.997e-5 / 4.000e-5 |
  | 2492 | 2.736 / 2.724 | 0.608 / 0.611 | 13.61 / 13.49 | 5.243e4 / 52,450 | 4.543e-5 / 4.546e-5 |
  | 2848 | 2.672 / 2.688 | 0.620 / 0.617 | 16.39 / 16.04 | 5.245e4 / 52,412 | 5.088e-5 / 5.091e-5 |

## Is your PR small enough?

No — 51 files / ~3.8K lines in 13 feature commits, of which ~1.1K are tests and ~0.9K are configs/readmes. The system, TEMPLATE, recipe and CI recipe depend on each other, so they are submitted together.

---

## Additional Context

- #6437, #6498


egs3/audioset/ssl/conf/inference.yaml
```diff
@@ -0,0 +1,18 @@
...
+    - name: train
+      data_src_args:
+        split: train
+        recipe_dir: ${recipe_dir}
+    - name: valid
+      data_src_args:
+        split: eval
+        recipe_dir: ${recipe_dir}
+
+batch_size: 144
```
Masao-Someki, line 18: Just for curiosity, is there any reason on using batch_size of 144?

Masao-Someki: If this comes from your environment, I think it would be nice to add comments on how many gpus or what GPU you used for this evaluation

egs3/audioset/ssl/conf/training_iter1.yaml
```diff
@@ -0,0 +1,16 @@
...
+# BEATs iteration 1 on AudioSet-2M: the tokenizer is distilled from the
+# iteration-0 encoder, and the encoder is re-trained on its targets.
+#
+#   python run.py --stages train_tokenizer infer train \
+#       --training_config conf/training_iter1.yaml \
+#       --train_tokenizer_config conf/training_tokenizer.yaml \
+#       --inference_config conf/inference.yaml
+defaults:
+  - training
+  - _self_
```
Masao-Someki, line 13: You can remove this

egs3/audioset/ssl/dataset/builder.py
```diff
@@ -0,0 +1,290 @@
...
+            )
+
+    def is_built(self, recipe_dir: str | Path, **_kwargs) -> bool:
+        """Return whether both manifests exist."""
+        data_root = Path(recipe_dir).resolve() / _CFG["data_path"]
+        return all(
+            (data_root / path).is_file() for path in _CFG["manifest_paths"].values()
+        )
+
+    def build(
```
Masao-Someki, line 222: Could you replace the `ProcessPoolExecutor.map` call with ESPnet3’s Provider / Runner pattern?

Please do not create a nested `ProcessPoolExecutor` inside runner workers; the runner should own scheduling and sharding.

A lightweight implementation could look like this:

```python
from typing import Any, Callable

from espnet3.parallel.base_runner import BaseRunner
from espnet3.parallel.env_provider import EnvironmentProvider


class ClipProvider(EnvironmentProvider):
    def __init__(self, config, examples: list[AudioSetExample]):
        super().__init__(config)
        self.examples = examples

    def build_env_local(self) -> dict[str, Any]:
        return {"examples": self.examples}

    def build_worker_setup_fn(self) -> Callable[[], dict[str, Any]]:
        examples = self.examples

        def setup() -> dict[str, Any]:
            return {"examples": examples}

        return setup


class PrepareClipRunner(BaseRunner):
    @staticmethod
    def forward(
        idx: int,
        examples: list[AudioSetExample],
        **_kwargs,
    ) -> dict:
        ok, num_samples = _prepare_clip(examples[idx])
        return {
            "idx": idx,
            "ok": ok,
            "num_samples": num_samples,
        }

    # Implement open_writers(), write_record(), and merge() to write
    # shard-local JSONL files and return records sorted by "idx".
```

Then `build()` becomes roughly:

```python
set_parallel(config.parallel)

runner = PrepareClipRunner(
    provider=ClipProvider(config, examples),
    batch_size=256,
    output_dir=data_root / "prepare_clips",
    shard_subdir=split,
    resume=False,
)

results = runner(range(len(examples)))

rows = [
    (
        f"{utt_id_prefix}-{result['idx']}",
        examples[result["idx"]].audio_path,
        result["num_samples"],
    )
    for result in results
    if result["ok"]
    and min_samples < result["num_samples"] < max_samples
]
```

For example, Slurm can be enabled without changing the provider or runner:

```yaml
parallel:
  env: slurm
  n_workers: 32
  options:
    cores: 1
    memory: 4GB
    walltime: "01:00:00"
```

egs3/audioset/ssl/dataset/builder.py
```diff
@@ -0,0 +1,290 @@
...
+        source_root = resolve_source_root(source_dir)
+        missing = [
+            str(source_root / entry["csv_path"])
+            for entries in _CFG["segment_lists"].values()
+            for entry in entries
+            if not (source_root / entry["csv_path"]).is_file()
+        ]
+        if missing:
+            raise FileNotFoundError(
+                "AudioSet segment lists are missing: " + ", ".join(missing)
```
Masao-Someki, line 212: I think it's also helpful to mention that we couldn't fing audioset and user need to download the audioset.

egs3/mini_an4/ssl/conf/training_iter1.yaml
```diff
@@ -0,0 +1,10 @@
+##########################################################
+#                    TRAINING SETTING                    #
+##########################################################
+# Tiny BEATs iteration 1 for CI.
+defaults:
+  - training
+  - _self_
```
Masao-Someki, line 7: You can remove this _self_

egs3/TEMPLATE/ssl/conf/inference.yaml
```diff
@@ -0,0 +1,91 @@
...
+exp_tag:
+exp_dir: ${recipe_dir}/exp/${exp_tag}
+data_dir: ${recipe_dir}/data
+dataset_dir: ${data_dir}
+# Must equal `training_config.target_dir`.
+inference_dir: ${exp_dir}/targets
+
+iteration: 0
+fbank_mean: 15.29130
+fbank_std: 5.90532
```
Masao-Someki, line 31: Could you add a comment on these values?

egs3/TEMPLATE/ssl/conf/training.yaml
```diff
@@ -0,0 +1,234 @@
...
+trainer:
+  accelerator: auto
+  devices: ${num_device}
+  num_nodes: ${num_nodes}
+  precision: bf16-mixed
+  accumulate_grad_batches: 1
+  check_val_every_n_epoch: 1
+  gradient_clip_val: 1.0
+  log_every_n_steps: 356
+  max_epochs: 56
```
Masao-Someki, line 222: Are these values should be default? I think maybe this is specific to openbeats?

egs3/TEMPLATE/ssl/conf/training_tokenizer.yaml
```diff
@@ -0,0 +1,183 @@
...
+recipe_dir: .
+data_dir: ${recipe_dir}/data
+ssl_tag: base
+exp_tag: beats_tokenizer_iter${iteration}_${ssl_tag}
+exp_dir: ${recipe_dir}/exp/${exp_tag}
+stats_dir: ${recipe_dir}/exp/stats_${ssl_tag}
+dataset_dir: ${data_dir}
+
+fbank_mean: 15.29130
+fbank_std: 5.90532
```
Masao-Someki, line 37: Should we keep these values as default values for all other ssl models..? I thought we could set it to some defaule value like 1 or 0?

egs3/TEMPLATE/ssl/conf/training_tokenizer.yaml
Masao-Someki, line 1: Could you confirm if all values could be the default values for other ssl models? We want to remove beats specific values from the TEMPLATE/ config

egs3/TEMPLATE/ssl/readme.md
Masao-Someki, line 1: For SSL?

egs3/TEMPLATE/ssl/run.py
```diff
@@ -0,0 +1,210 @@
...
+# Canonical stage order for one BEATs iteration. Execution always follows this
+# order: the tokenizer (iteration > 0) must exist before `infer` writes targets,
+# and targets must exist before `collect_stats`/`train` build the dataset.
+DEFAULT_STAGES: List[str] = [
+    "create_dataset",
+    "train_tokenizer",
+    "infer",
+    "collect_stats",
+    "train",
+]
```
Masao-Someki, line 35: I think we can expose one stage `pretrain`? You can implement a function in the system class like:

```
def pretrain(..):
  self.create_dateset()
  self.train_tokenizer()
  self.infer()
  self.collect_stats()
  self.train()
```

and we can just expose `pretrain`, `infer`?, `measure`, and publication/demo(?) stages.

espnet3/systems/ssl/__init__.py
Masao-Someki, line 1: Let's change system from ssl to openbeats

espnet3/systems/ssl/checkpoint_export.py
Masao-Someki, line 1: Hmm, I think we can simply add a callback after training, for example using `on_validation_end`.

You can create a new PyTorch Lightning callback and specify it in the config like this:

```yaml
trainer:
  ...
  callbacks:
    - _target_: espnet3.systems.openbeats.callbacks.ConvertCheckpoint
      args: value
      args2: value
```

Any callbacks defined under `trainer.callbacks` are executed after the default callbacks, so we can assume that the averaged checkpoints have already been created by that point.

espnet3/systems/ssl/system.py
Masao-Someki, line 1: As mentioned we can create `pretrain` method (=stage) to execute multiple stages. 
We also want to have measure, and publication-related stages.

espnet3/systems/ssl/system.py
```diff
@@ -0,0 +1,347 @@
...
+    def iteration(self) -> int:
+        """Return the BEATs iteration selected by ``training_config.iteration``."""
+        if self.training_config is None:
+            return 0
+        iteration = int(self.training_config.get("iteration", 0) or 0)
+        if iteration < 0:
+            raise ValueError(f"training_config.iteration must be >= 0: {iteration}")
+        return iteration
+
+    def get_encoder_checkpoint_path(self) -> Path:
```
Masao-Someki, line 129: ESPnet3 assumes any public functions in system class is a stage. So please change this to a private function `_get_encoder_checkpoint_path`..

espnet3/systems/ssl/system.py
```diff
@@ -0,0 +1,347 @@
...
+        """Return the portable encoder checkpoint path of this iteration."""
+        return Path(self.training_config.exp_dir) / (
+            f"beats_encoder_iter{self.iteration}.pt"
+        )
+
+    def get_tokenizer_checkpoint_path(self) -> Path:
```
Masao-Someki, line 135: ditto

espnet3/systems/ssl/system.py
```diff
@@ -0,0 +1,347 @@
...
+            publication_config=publication_config,
+            stage_log_mapping={
+                "train_tokenizer": "train_tokenizer_config.exp_dir",
+                **(stage_log_mapping or {}),
+            },
+            demo_config=demo_config,
+        )
+
+    @property
+    def iteration(self) -> int:
```
Masao-Someki, line 120: I think we can change this to `_get_iteration()`?

espnet3/systems/ssl/system.py
```diff
@@ -0,0 +1,347 @@
...
+
+    Args:
+        output_path: Token list file to write.
+        codebook_size: Number of tokenizer codebook entries.
+
+    Returns:
+        Path: ``output_path``.
+
+    Raises:
+        ValueError: If an existing file does not match ``codebook_size``.
```
Masao-Someki, line 302: Please include example code snippets for public functions. For example, showing what the first three lines of an output file would look like would make it much easier for readers to understand the expected behavior.

espnet3/systems/ssl/system.py
```diff
@@ -0,0 +1,347 @@
...
+                "training_config.model.token_list must be set.",
+            )
+        )
+        codebook_size = int(
+            model_config.encoder_conf.beats_config.get("codebook_vocab_size", 1024)
+        )
+        write_token_list(token_list, codebook_size)
+
+
+def write_token_list(output_path: str | Path, codebook_size: int) -> Path:
```
Masao-Someki, line 287: If this is not used from other components, I think we can just make it a private helper so that we can change it to `_write_token_list`

espnet3/systems/ssl/system.py
```diff
@@ -0,0 +1,347 @@
...
+    # per-process: a shared one is truncated by the next writer while an
+    # earlier rank is still writing it.
+    tmp_path = output.with_name(f"{output.name}.{os.getpid()}.tmp")
+    tmp_path.write_text("\n".join(expected) + "\n", encoding="utf-8")
+    tmp_path.replace(output)
+    logger.info("Wrote token list (%d entries): %s", len(expected), output)
+    return output
+
+
+def write_target_shape(target_path: str | Path, output_path: str | Path) -> Path:
```
Masao-Someki, line 325: ditto

espnet3/systems/ssl/target_reader.py
```diff
@@ -0,0 +1,115 @@
...
+            raise FileNotFoundError(
+                f"Target file not found: {self.target_path}. Run the `infer` stage "
+                "for this iteration first."
+            )
+        self.num_items = int(num_items)
+        self._offsets, self._lengths = self._build_index()
+        self._fd = None
+        self._fd_pid = None
+
+    def _build_index(self) -> tuple[np.ndarray, np.ndarray]:
```
Masao-Someki, line 59: Please add simple comment on how this could be used, what it would do

espnet3/systems/ssl/target_reader.py
```diff
@@ -0,0 +1,115 @@
...
+                "The targets were produced for a different dataset."
+            )
+        return offsets, lengths
+
+    def __len__(self) -> int:
+        """Return the number of indexed targets."""
+        return self.num_items
+
+    def __getitem__(self, idx: int) -> str:
+        """Return the space-separated token ids for dataset item ``idx``."""
```
Masao-Someki, line 100: I think we can add a concrete example here. For instance, if we pass `1`, it would be helpful to show what output we expect to get.

espnet3/systems/ssl/tokenization_model.py
```diff
@@ -0,0 +1,180 @@
...
+
+        Args:
+            speech: One waveform/fbank array, or a list of them for batched
+                inference. Items in a batch may have different lengths; they
+                are zero-padded and the padding is masked out.
+
+        Returns:
+            np.ndarray | List[np.ndarray]: ``int64`` code ids of shape
+            ``(num_patches,)`` for a single item, or one such array per item
+            when ``speech`` is a list.
```
Masao-Someki, line 136: Could you add example snippet here?

espnet3/systems/ssl/system.py
```diff
@@ -0,0 +1,347 @@
...
+                f"Teacher checkpoint not found: {teacher}. Run the `train` stage "
+                f"for iteration {self.iteration - 1} first."
+            )
+        logger.info(
+            "Training BEATs tokenizer | iteration=%d teacher=%s exp_dir=%s",
+            self.iteration,
+            teacher,
+            config.exp_dir,
+        )
+        trainer = run_training(config)
```
Masao-Someki, line 190: This run_training is basically the same as `super().train()`, right? Is there any specific reason we need to import the `train()` and use it?

espnet3/utils/stages_utils.py
```diff
@@ -149,7 +153,7 @@ def run_stages(
             log_dir = stage_log_dirs.get(stage) or stage_log_dirs.get("default")
             filename = f"{stage}.log"
 
-            if stage == "train":
+            if stage in _TRAINING_STAGES:
```
Masao-Someki, line 156: Thanks this looks nice, for future support I think we can process `train_*` stages here

Masao-Someki (commented):
Thank you @jaeyeonkim99!!
Could you add `metrics.yaml` in the recipe? Otherwise I think we cannot compute metrics from the trained model..

Also could you change the system name to `systems/openbeats`, or `systems/beats`? 
I think ESPnet3 ssl (beats) implementation is not for general ssl model, but for beats specifically.
Then we can keep many of the template values or readme as-is. In espnet3 we allow each systems does not have to be task-oriented. For example, we have `f5_tts` system, `knn_vc` system, which is to implement the F5TTS or knn-vc model.

Masao-Someki:
Hi, based on our discussion on system naming rules, could you change the system name to `beats`?

PR #6675, opened 2026-09-16
Recommend maxlenratio=0 in the OWSM v4 README

amirlankalm:
## What did you change?

One sentence in `egs2/owsm_v4/s2t1/README.md`. It recommended `maxlenratio=1.0` and called it the default; the default of `Speech2Text` is `0.0`. The sentence now says to keep the default of `0.0` and to avoid smaller positive values, and explains why.

---

`maxlenratio=0` and `1.0` give the same maximum output length for OWSM (the number of encoder frames, 374 for the 30 s window), but the beam search only runs its end detection (Eq. 50 of Watanabe et al., 2017) when `maxlenratio == 0`. With `1.0` every utterance is decoded to step 374: once the transcript has ended, the other beams keep producing punctuation over the padded silence, and on CPU that is about 97% of the decoding time.

Measured on 100 LibriSpeech test-clean utterances sampled from the Open ASR Leaderboard (13.8 min of audio, Apple M4, 4 threads, OWSM v4 base, beam 5, batch 1): `maxlenratio=1.0` decodes in 994 s (RTFx 0.8), `maxlenratio=0` in 94 s (RTFx 8.8), transcripts identical on all 100 utterances. Five more sets of the same sample (test-other, AMI, VoxPopuli, Earnings-22, Common Voice) are also identical between the two settings.

Discussed with @sw005320: README only, no code change.

---

## Is your PR small enough?

Yes, one line.

---

## Additional Context

The sampled leaderboard sets and the measurement scripts are in a follow-up PR to this recipe.


sw005320 (approved):

sw005320:
Thanks, @amirlankalm!

PR #6676, opened 2026-09-16
Add an Open ASR Leaderboard test to the OWSM v4 recipe

amirlankalm:
## What did you change?

Three scripts under `egs2/owsm_v4/s2t1/local/` and a README section.

- `prepare_hf_asr_leaderboard.py` streams a test set of [hf-audio/open-asr-leaderboard](https://huggingface.co/datasets/hf-audio/open-asr-leaderboard), takes a seeded shuffle over a buffer of rows and keeps the first N (default 100) utterances of at most 30 s, dropping empty and "ignore time segment in scoring" references as the leaderboard's own loader does. It writes a Kaldi-style data directory (`wav.scp`, `text`, `utt2dur`, `utt2spk`, `spk2utt`) plus an `info.json` with the sampling parameters and duration statistics. The Hub copies are stored longest-first, so "the first N rows" would be the N longest utterances; the seeded shuffle avoids that and is deterministic (two independent runs picked the same 100 utterances on all six sets).
- `eval_hf_asr_leaderboard.py` decodes such a directory with `Speech2Text` (batched through `batch_decode` from #6599 when `--batch_size > 1`, optionally sorted by duration with `--sort`), normalizes references and hypotheses with `TextCleaner(["whisper_en"])` (the leaderboard's `EnglishTextNormalizer`), and reports the corpus WER from `jiwer` and RTFx = audio seconds / decoding seconds, timing only the decode calls after one untimed warm-up batch. It writes a summary json, a per-utterance jsonl, and prints one markdown row. `--maxlenratio` defaults to 0, following #6675.
- `run_hf_asr_leaderboard.sh` runs several `config:split` pairs with one decoding configuration and collects `exp/hf_asr_leaderboard/summary.md`.

---

Suggested by @sw005320 as a laptop-sized leaderboard test for the OWSM v4 recipe, to work on CPU inference speed: one configuration over six sets of 100 utterances runs in about 5 minutes, so an inference change can be checked for WER and speed on the same sample. The README section reports what plain master gives with the README setting of #6675 (Apple M4, 4 threads, decoding time only):

| set | utterances | audio (min) | mean length (s) |
| --- | ---: | ---: | ---: |
| test-clean | 100 | 13.8 | 8.3 |
| test-other | 100 | 11.8 | 7.1 |
| AMI | 100 | 8.2 | 4.9 |
| VoxPopuli | 100 | 18.6 | 11.2 |
| Earnings-22 | 100 | 13.1 | 7.8 |
| Common Voice | 100 | 12.4 | 7.5 |

| decoding (WER % / RTFx) | test-clean | test-other | AMI | VoxPopuli | Earnings-22 | Common Voice |
| --- | ---: | ---: | ---: | ---: | ---: | ---: |
| beam 5, batch 1, maxlenratio 1.0 (old README advice) | 3.35 / 0.9 | - | - | - | - | - |
| beam 5, batch 1, maxlenratio 0 (default) | 3.35 / 8.8 | 9.31 / 8.2 | 16.77 / 6.2 | 5.37 / 9.4 | 16.61 / 9.9 | 19.39 / 10.9 |
| beam 5, batch 8 sorted, maxlenratio 0 | 3.35 / 9.8 | 9.31 / 9.3 | 16.77 / 7.1 | 5.37 / 7.8 | 16.61 / 9.9 | 19.39 / 12.2 |
| greedy, batch 8 sorted, maxlenratio 0 | 3.57 / 23.1 | 9.78 / 17.5 | 29.63 / 11.6 | 5.69 / 32.7 | 16.50 / 24.4 | 20.81 / 26.0 |
| 370M: beam 5, batch 8 sorted, maxlenratio 0 | 2.32 / 2.7 | 6.31 / 2.2 | 12.65 / 0.6 | 3.96 / 3.5 | 14.03 / 2.9 | 15.15 / 2.3 |

Batch 1 and sorted batch 8 give identical transcripts on all six sets, as do `maxlenratio` 0 and 1.0. Two observations the README spells out: with end detection a batch runs until its slowest utterance is detected as finished, so on CPU sorted batching gains little and loses on VoxPopuli; and on the 370M model over AMI end detection does not fire on some utterances, so 4 of 13 batches ran to the 374-step maximum. Greedy loops on some AMI utterances (a 13-word reference became 131 words), hence its WER there.

---

## Is your PR small enough?

Yes: three new scripts (about 450 lines) and a README section.

---

## Additional Context

- Leaving the streaming iterator early hangs at interpreter exit inside the Hub client's retry loop (reproduced: a 3-row stream that `break`s never exits, 120 s timeout), so the prepare script flushes its outputs and leaves with `os._exit(0)`. Documented in a comment.
- `datasets` is a base dependency; `jiwer` is in the `asr` extra and `openai-whisper` in `sds`. The eval script raises a clear error if the normalizer is missing. The dataset is gated (accept the terms and log in first).
- black / isort / flake8 (incl. hacking) clean; the shell script passes shellcheck with the CI exclusions. Prepared samples and outputs go under `data/` and `exp/`, which the repository already ignores.
- Follow-up to #6675.


sw005320:
please handle [coderabbitai](https://github.com/apps/coderabbitai)'s reviews

amirlankalm:
handled all four in 6596ba5: `--revision` pinned to the commit the numbers were sampled from and recorded in info.json, the runner re-prepares a sample unless n, seed and revision match a complete dir, common voice added to the default sets, and `--batch_size` must be > 0.

PR #6677, opened 2026-09-16
Skip the conv subsampling of a constant padded tail at inference

amirlankalm:
## What did you change?

`Conv2dSubsampling`, `Conv2dSubsampling1/2/6/8` now share one helper, `_conv2d_project`, for the convolution stack and the output projection. At inference it checks whether every utterance in the batch ends in a run of one repeated frame (one comparison of each frame with the last one). If so, the convolutions run only up to the first output frame whose receptive field lies inside the run, and that frame is repeated for the remaining positions. Frames whose receptive field touches real content are computed as before, so the output is the full computation's. Training mode, a non-constant tail, or `module.skip_constant_tail = False` take the usual path. Masks are untouched (they come from the lengths, not from the values).

---

The padding of an inference batch is one repeated frame: zeros that the normalization turns into a fixed log-mel row, or, for OWSM, most of a 30 s window. A convolution over identical frames gives identical outputs, so the work spent on them is redundant. For OWSM v4 base on CPU the three stride-2 convolutions are about a third of the encoder time and mostly run over padding: an 8 s utterance is 27% of the 3001 mel frames. On the conv stack alone an 11 s utterance goes from 108 ms to 39 ms with a maximum output difference of 0.0.

End to end on the Open ASR Leaderboard LibriSpeech test-clean sample (100 utterances, 13.8 min, Apple M4, 4 threads, OWSM v4 base, beam 5, `maxlenratio=0` as in #6675, the scripts of #6676), transcripts identical to the baseline in every row:

| decoding | master | with this PR |
|---|---:|---:|
| batch 1 | RTFx 8.8 (0.94 s/utt) | 9.0 (0.92 s/utt) |
| batch 8, sorted by length | 9.8 (0.84 s/utt) | 10.9 (0.76 s/utt) |

The saving is a fixed 60-70 ms per utterance, so its share grows as the decoder gets faster: with a decoder key/value cache and an exact per-utterance stop (branches, not in this PR) the same change is 1.10x at batch 1 and 1.14x at batch 8.

The E-Branchformer layers still run over the whole padded window; they cannot be cropped, because attention mixes every frame and the model was trained on 30 s inputs (encoding the true length gives WER 3.35 -> 10.2).

Relation to #6633 and #6660: those made the encoder padding-invariant (masks from each utterance's own length, time convolutions inside the layers zero the padded frames). This change is about cost, not invariance, and sits in front of both: the subsampling convolutions still read the padded frames, they just stop computing them once every frame in the run is the same. With #6660 the padding of a shorter utterance in a mixed-length batch is zeros, a constant run, so the shortcut applies there as well.

---

## Is your PR small enough?

Yes: one library file (+54 / -20) and one test file.

---

## Additional Context

- `test/espnet2/legacy/test_subsampling_constant_tail.py`: for all five classes, output equal to the full computation on batches with different constant tails per utterance (with and without a mask, whose subsampled version is unchanged); the shortcut is really taken (the convolutions see a prefix); the full path runs when one utterance has no padding or in training mode; an all-constant input; and the receptive-field arithmetic (the crop always keeps one output frame that lies entirely in the padding).
- Existing subsampling, encoder and s2t inference suites pass (19,635 tests; the one failure in this environment, `test_Speech2Text_quantized`, fails identically on master: macOS torch leaves the quantized engine unset).
- black / isort / flake8 (incl. hacking) clean.
- In general a convolution over a shorter input can differ from the full one in the last floating-point bits if the backend picks another kernel by shape; here the outputs were identical on CPU.


amirlankalm:
done in 8011721: the shortcut now also requires autograd to be off, and the tests run under no_grad and check that grad mode takes the full path.

PR #6679, opened 2026-09-16
Guard each optional package on its own

sw005320:
## What

Three modules import two unrelated optional packages inside a single `try/except ImportError` and set both to `None` when it fires, so whichever one is absent takes the other down with it:

| file | bundled together |
| :--- | :--- |
| `espnet2/text/phoneme_tokenizer.py` | `g2p_en` (English) + `jamo` (Korean) |
| `espnet2/text/cleaner.py` | `tacotron_cleaner` (English) + `jaconv` (Japanese) |
| `espnet2/enh/loss/criterions/time_domain.py` | `ci_sdr` + `fast_bss_eval` |

## Why it matters

The failure is quiet and the message points the wrong way. With `g2p_en` installed and `jamo` not, an English TTS model dies inside `G2p_en.__init__` on:

```
RuntimeError: Please install espnet with `pip install espnet[tts]`
```

— the tts extra being, of course, already installed. This came up while setting up a Colab demo: the time went into looking for a broken `g2p_en` rather than a missing Korean package. The same shape is in `cleaner.py` (a missing `jaconv` disables the English tacotron cleaner) and in `time_domain.py` (whichever metric is absent disables both losses).

## Change

One `try` per package, and the error names the package that is actually missing:

```python
try:
    import g2p_en
except ImportError:
    g2p_en = None

try:
    import jamo
except ImportError:
    jamo = None
```

```
RuntimeError: jamo is not installed: pip install espnet[tts]
```

Behaviour is otherwise unchanged — each feature still raises when its own dependency is absent.

## Test

`test/espnet2/test_optional_import_guards.py` reloads each module with one package hidden from the import machinery and checks the other one survives, plus a case for the error naming what is missing. Against the previous code, five of its seven cases fail (the other two skip here, `ci_sdr` and `fast_bss_eval` not being installed); with this change they pass.

Unrelated to, and independent of, #6678.


sw005320:
Addressed in d9c6833 (plus 1eacedc, tidying a string the pinned black had joined).

The point is taken: the message assertion covered one of the six guards, so the other five could have named the wrong package and still passed - and naming the right package is the whole change. It is now parameterized over all six: `Jaso`/jamo, `G2p_en`/g2p_en, the two cleaner types, and `CISDRLoss`/`SDRLoss`/`SISNRLoss` against ci_sdr and fast_bss_eval.

The English-survives-without-Korean case keeps a test of its own, since it asserts the opposite - that nothing raises.

12 passed, 2 skipped locally. The two skips are the *survivor* checks for ci_sdr and fast_bss_eval, which are not installed here; their message checks run regardless, the package being absent being the condition under test.

PR #6682, opened 2026-09-16
Point the README quick start at OWSM-CTC and show how to publish a demo

sw005320:
## What

README only, +31/−11.

**Quick start.** The example now runs `espnet/owsm_ctc_v4_1B` through `Speech2TextGreedySearch.batch_decode`: pass a path of any length or sample rate, get the text back. Run as written on `test_utils/ctc_align_test.wav`, CPU only, it prints

> The sale of the hotels is part of Holi's strategy to sell off assets and concentrate on property management.

`<st_deu>` and `<nolang>`, mentioned in the paragraph below the snippet, are both in the released model's token list.

**Demos.** The table now leads with the spoken-dialogue recipe behind `egs2/TEMPLATE/sds1`. Below it, the `pack_model → pack_demo → upload_demo` stages that espnet3 ships for Gradio + Hugging Face Spaces appear in the README for the first time; until now `conf/demo.yaml` was the only place they were described.

## Why

The previous quick start used the encoder-decoder OWSM v4 small: read the file yourself, 16 kHz only, beam search of 5 on the CPU. It is also not what people take from the Hub. Over the last thirty days `espnet/owsm_v4_small_370M` had 51 downloads; `espnet/owsm_ctc_v4_1B` had 9,981, and it is the one model in the organisation whose card carries a runnable snippet.

## After review (second commit)

- **Every Hugging Face Space the README linked is in `RUNTIME_ERROR`** — `akhaliq/espnet2_asr` and `akhaliq/ESPnet2-TTS` (linked since 2021), `Siddhant/ESPnet2-SLU`, and `Siddhant/Voice_Assistant_Demo` which the first commit added. The pages return 200, so a link check passes; `GET /api/spaces/<id>` shows `runtime.stage`. All four links are gone; the table now links recipes and Colab notebooks only. Replacing them with org-owned Spaces built by `upload_demo` is the obvious follow-up and needs an HF token.
- The publishing commands take `--training_config`: `conf/demo.yaml` interpolates `${exp_tag}`, so `pack_demo --dry_run` without it fails with `InterpolationKeyError: 'exp_tag'` and `pack_model` refuses to run. Verified with `--dry_run` both ways.
- "one CPU pass" → "one encoder pass per 30 s window"; `batch_decode` chunks longer audio.

## Checked

- `pack_demo` refuses to run without `model.dir_or_tag`, which the template takes from `pack_model`'s output — hence three stages, not two.
[...]

PR #6684, opened 2026-09-16
Give the streaming ASR/ST and classification classes from_pretrained

sw005320:
## What

`Speech2TextStreaming` (ASR and ST) and `Classification` gain the `from_pretrained(model_tag=None, **kwargs)` staticmethod that the other twenty-three inference classes in `espnet2/bin` already have — same body as `asr_inference.py`, name for name. `pack.py` already emits the constructor keys for all three (`asr_*`/`lm_*`, `st_*`, `classification_*`), so packing is untouched.

Not covered, on purpose: `asvspoof_inference.SpeechAntiSpoof` (pack.py has no packing for that task, so no tag could reach it); `lid_inference` and `ps2st_inference` have no class.

## Why

The streaming ASR notebook has to call `ModelDownloader().download_and_unpack()` and spread the result into the constructor by hand, and #6682's README line "every other task uses the same `from_pretrained` pattern" was not quite true until now.

## Tests

- `test_Speech2Text_streaming_from_pretrained` (ASR), `test_Speech2TextStreaming_from_pretrained` (ST), `test_Classification_from_pretrained` — `model_tag=None` on a dry-run config, one call on random audio, as `test_asr_transducer_inference` does.
- `st_inference_streaming` had no test before; its fixture is new (contextual_block_transformer, same block sizes as the streaming ASR test).
- Locally: the three new tests pass; the rest of the three files pass except `transformers`-dependent and quantized tests, which need packages/engines my macOS env lacks.


sw005320:
Added offline tests for the tagged path (`ModelDownloader` faked at the import site inside each method) for all three classes — covers the branch codecov/patch flagged and the point raised in the review.

sw005320:
@coderabbitai full review

PR #6686, opened 2026-09-17
Do not require importing k2 on a torch version k2 has no wheel for

sw005320:
## Every job on torch 2.14.0 that uses a prebuilt image fails

```
ModuleNotFoundError: No module named 'k2'
##[error]Process completed with exit code 1.
```

`tools/installers/install_k2.sh` carries `k2_missing_for="2.14.0"` and skips k2 there **on purpose** — k2 publishes no wheel for that torch at all yet, for any python. The composite action that verifies the environment did not know, and asked for:

```bash
"${VENV}/python" -c 'import espnet2, espnet3, kaldiio, k2; ...'
```

## master is green only by accident of timing

The two have disagreed since 2.14.0 was added in #6678. Until the images for that hash were published, `resolve_ci_image` reported them unavailable, every job fell back to building its own environment, and this step was **skipped entirely**:

```
master acc0acf6 (green): test_python_espnet2 (3.12, 2.14.0) -> success
   4    skipped  Run ./.github/actions/use-prebuilt-environment
   5    success  Make environment
```

The images exist now, so the next run on any branch takes the fast path and hits it. #6685 was simply the first thing to run after publication — five of its jobs failed this way, all on 2.14.0.

**So this is not one PR's problem. Nothing can go green until it is fixed.**

## Fix

The action asks the interpreter which torch it has — rather than trusting the image tag — and reads the gap list out of the installer, so the two cannot say different things:

```bash
torch_version=$("${VENV}/python" -c 'import torch; print(torch.__version__.split("+")[0])')
k2_missing_for=$(sed -n 's/^k2_missing_for="\(.*\)"/\1/p' tools/installers/install_k2.sh)
```

## Verified, five cases against a stubbed interpreter

| case | exit | |
|---|---|---|
| gap version (2.14.0), k2 absent | 0 | passes without k2 |
| gap version with `+cpu` suffix | 0 | the case the version split exists for |
| non-gap version (2.13.0), k2 present | 0 | imports k2 |
| **non-gap version, k2 missing** | **1** | still fails — the check is not weakened into uselessness |
| non-gap version with `+cu126` | 0 | |

My first run of that harness reported the `+cpu` case as *failing*. The stub was echoing the version without applying the split the action relies on — it was testing the stub, not the action. Fixed by running the real expression against a fake `torch` module, and that is the run above.

## Keeping it

`ci/check_ci_image_config.py` already checked that `k2_missing_for` names only built torch versions. It now also requires that, **while the list is non-empty, the action verifying the environment reads it**. Checked by running it against master's copy of the action, which it rejects:

```
.github/actions/use-prebuilt-environment/action.yml: verifies the environment
without reading tools/installers/install_k2.sh's k2_missing_for, which
currently skips k2 for torch 2.14.0.
  Requiring `import k2` on a torch version k2 publishes no wheel for fails
  every job on that part of the grid.
```

The skip is still recorded where it was: `k2_missing_for` remains the one place that says this variant tests less than the others, and the k2 tests remain `pytest.importorskip`.


sw005320:
This seems to be overlapped with https://github.com/espnet/espnet/pull/6681

sw005320:
Closing — this duplicates #6681, which was opened two and a half hours earlier and is already green (100/100).

Same diagnosis and, as it turns out, nearly the same patch: probe `torch.__version__.split("+")[0]`, `sed` the `k2_missing_for` list out of `install_k2.sh`, and assert k2 only when it is expected. I wrote it without checking the open pull requests first. That is on me.

One thing I checked before closing, because the two differ in shape: #6681 writes the loop as `[ "${torch_version}" = "${missing}" ] && expect_k2=false`, and I wondered whether a failing test as the last command of the loop body would abort the step under GitHub's `bash --noprofile --norc -eo pipefail`. Measured both forms against both a gap and a non-gap version — **all four exit 0**. My concern was wrong; #6681's loop is fine as written.

**The one thing here that #6681 does not have** is the other half of the rule in `ci/check_ci_image_config.py`: while `k2_missing_for` is non-empty, the action that verifies the environment must read it. Without that, the two files can drift apart again exactly as they did, and the runs stay green until the images happen to be published.

```
.github/actions/use-prebuilt-environment/action.yml: verifies the environment
without reading tools/installers/install_k2.sh's k2_missing_for, which
currently skips k2 for torch 2.14.0.
```

Happy to add that to #6681, or to send it as a small follow-up once #6681 lands — whichever you prefer. Nothing else here is worth keeping.

PR #6688, opened 2026-09-17
Add kNN-VC voice conversion to espnet3

aalmuarik:
## What did you change?

Added kNN-VC (Baas, van Niekerk and Kamper, Interspeech 2023,
https://github.com/bshall/knn-vc) to espnet3, in three commits:

1. `espnet3/systems/vc/models/knnvc/`: vendored WavLM, the frozen encoder, the kNN
   matcher, the HiFi-GAN vocoder training model, checkpoint conversion for the
   authors' released weights, and `KNNVCModel` for inference.
2. `VCSystem` with a new `prepare_features` stage (a port of their
   `prematch_dataset.py`), `egs3/TEMPLATE/vc/`, and
   `ci/test_integration_espnet3_vc.sh` wired into `ci_on_ubuntu.yml`.
3. `egs3/librispeech_100/vc`, reproducing their setup by default.

72 unit tests, plus a CPU integration test that downloads nothing.

---

To run kNN-VC through espnet3's stages rather than as a standalone repository.

The port is numerically equivalent to the official implementation:

| Check | Result |
|---|---|
| Generator, released `prematch_g_02500000.pt` | max diff 0.0 (bit-exact) |
| WavLM layer-6 features | max diff 3.8e-05 |
| Prematched features | max diff 1.9e-03 |
| End-to-end conversion | correlation 1.000000 |

It also runs end to end on the full corpus: `prepare_features` over 31242 utterances
(37 GB, 2h13m on one A6000), then `train`, `infer` over 200 pairs, and `pack_model`,
with the packed bundle loading from outside the repo. Reproducing the paper's quality
needs training the vocoder for something like the authors' 2.5M steps, which I have
not run.

---

## Is your PR small enough?

Each commit is. Together they are not, and 1650 of the lines in components are the
vendored WavLM.

| Commit | Files | Lines written here |
|---|---|---|
| components | 19 | 2007, plus 1650 of vendored WavLM |
| VCSystem and TEMPLATE | 20 | 2196, of which 369 is the CI script |
| recipe | 15 | 1340 |

Each one stands alone and passes its own tests and lint, and the three branches are
pushed separately. They are in one PR because a cross-fork PR's base has to be a
branch in `espnet/espnet`, so they cannot be stacked until the one below merges. I am
happy to open them one at a time instead.

---

## Additional Context

- **Vendored WavLM.** `espnet-s3prl` is declared only in the `s2st` extra, so it is
  not a dependency espnet3 users have. It also cannot reproduce kNN-VC as installed:
  s3prl layer-norms the input waveform when the checkpoint sets `cfg.normalize`,
  which WavLM-Large does and kNN-VC deliberately skips. That moves layer-6 features
  to cosine 0.89 against this port. `expert.cfg.normalize=False` matches exactly, but
  no espnet config surface reaches that flag. This follows the precedent of
  `espnet3/components/callbacks/vendored_ema.py`.
- **GAN stepping.** The model returns
  `[OptimizationStep(discriminator), OptimizationStep(generator)]`, so the
  discriminator updates first, as in the authors' `hifigan/train.py`. The
  discriminator is frozen during the generator's adversarial pass, otherwise autograd
  fails on the in-place discriminator update. One difference from their script
  remains, documented in the recipe readme: espnet3 computes both losses from a
  single `forward`, so the generator is scored by the discriminator before its
  update, where `hifigan/train.py` re-runs the updated one.
- **Discriminator output kernel.** `HiFiGANPeriodDiscriminator` derives its output
  kernel as `kernel_sizes[1] - 1` and requires an odd `kernel_sizes[1]`, so
  HiFi-GAN's `(3, 1)` cannot be set from config. `KNNVCVocoderModel` rebuilds that
  one convolution instead of changing a module every ESPnet GAN model shares.
- **Separate integration script.** The VC leg is its own
  `ci/test_integration_espnet3_vc.sh` with its own job, like
  `ci/test_integration_espnet3_publication.sh`, rather than an addition to
  `ci/test_integration_espnet3.sh`. That script runs an existing recipe in
  `egs3/mini_an4/asr`; the VC recipe needs LibriSpeech and a 1.2 GB WavLM checkpoint,
  so this one builds a miniature recipe from the mini_an4 audio with a stub encoder
  and a tiny HiFi-GAN instead, and downloads nothing.
- **`measure` ships no VC metric yet.** ASR-WER and SV-EER are left as follow-up and
  noted in the recipe readme. The VC stage list also omits `pack_demo` and
  `upload_demo`, since the shared demo UI cannot express a variable-length reference
  set.

Deviations from the original recipe, and the option that restores each one, are
listed under "Relation to the original recipe" in `egs3/librispeech_100/vc/readme.md`.


aalmuarik:
@Masao-Someki opening this as draft, ready for a first look whenever you have time.

Masao-Someki:
Hi thanks @aalmuarik!!
Just few comments, could you fix these first before I review the entire recipe?
1. We can create knnvc system, so that we can have `espnet3/systems/knnvc/` and you can put metrics, system.py, model definition etc inside.
2. Could you add the integration test on `ci/test_integration_espnet3.sh` after the asr test, and create `egs3/mini_an4/knnvc`?

espnet3/systems/knnvc/matcher.py
```diff
@@ -0,0 +1,105 @@
...
+    Args:
+        source_feats: Query frames, shape ``(N1, dim)``.
+        matching_pool: Matching set frames, shape ``(N2, dim)``.
+
+    Returns:
+        Distance matrix of shape ``(N1, N2)`` in ``[0, 2]``.
+
+    Raises:
+        ValueError: If either input is not 2-D or the feature dims differ.
+    """
```
Masao-Someki, line 33: Let's add an example!

egs3/librispeech_100/knnvc/dataset/dataset.py
```diff
@@ -0,0 +1,368 @@
...
+        return {
+            "speech": _read_audio(source.audio_path),
+            "reference_speech": self._read_reference_audio(target_speaker),
+            "target_speaker": target_speaker,
+            "pair_id": f"{source.utt_id}_to_{target_speaker}",
+            "text": source.text,
+        }
+
+    def __getitem__(self, idx: int) -> Dict[str, Any]:
+        """Return one item according to ``kind`` (see the class docstring)."""
```
Masao-Someki, line 362: Could you add examples on return value based on `self.kind`?

egs3/librispeech_100/knnvc/dataset/dataset.py
Masao-Someki, line 1: Could you add one line of docstrings to private functions starting with `_`?

egs3/TEMPLATE/knnvc/conf/metrics.yaml
```diff
@@ -0,0 +1,22 @@
...
+# stage cannot run until a recipe (or a `--metrics_config` of your own) fills
+# it in with at least one `BaseMetric` target. See the recipe readme.
+
+recipe_dir: .
+exp_tag:
+exp_dir: ${recipe_dir}/exp/${exp_tag}
+
+dataset:
+
+metrics:
```
Masao-Someki, line 22: Just one confirmation, this PR provides training script for the vocoder side, right? What metrics is typically used for vocoder training and I think we should add some?

espnet3/systems/knnvc/model.py
```diff
@@ -0,0 +1,292 @@
...
+
+class KNNVCModel:
+    """Any-to-any voice conversion with kNN-VC, callable per utterance.
+
+    This is the ``model:`` target of the ``infer`` stage. ESPnet3's
+    ``InferenceRunner`` calls it as ``model(**inputs)`` with the dataset fields
+    named by ``inference_config.input_key``; the recipe dataset provides
+    ``speech`` (source utterance), ``reference_speech`` (target speaker's
+    utterances) and ``target_speaker`` (a cache key), see
+    ``egs3/librispeech_100/knnvc``.
```
Masao-Someki, line 39: Let's avoid to mention recipe-specific item in the espnet3/ package

espnet3/systems/knnvc/wavlm_encoder.py
```diff
@@ -0,0 +1,176 @@
...
+from espnet3.systems.knnvc.checkpoint import (
+    load_state_dict_from_path_or_url,
+)
+from espnet3.systems.knnvc.vendored_wavlm import WavLM, WavLMConfig
+
+logger = logging.getLogger(__name__)
+
+WAVLM_LARGE_URL = (
+    "https://github.com/bshall/knn-vc/releases/download/v0.1/WavLM-Large.pt"
+)
```
Masao-Someki, line 28: For these default values, we want to have in `egs3/TEMPLATE/`, so could you move this? I think we don't need to test this default value.

test/espnet3/systems/knnvc/test_wavlm_encoder.py
```diff
@@ -0,0 +1,83 @@
...
+# original kNN-VC implementation) rather than here.
+
+# A minimal WavLM config: two transformer layers is enough for the guards
+# below, which all fail before any module is constructed.
+_TINY_CFG = {"encoder_layers": 2, "encoder_embed_dim": 8}
+
+
+def test_default_checkpoint_is_the_knn_vc_release():
+    assert WAVLM_LARGE_URL.startswith("https://github.com/bshall/knn-vc/releases/")
+    assert WAVLM_LARGE_URL.endswith("WavLM-Large.pt")
```
Masao-Someki, line 42: I think we don't need to test this

test_utils/espnet3/tb/optim_test/version_0/hparams.yaml
Masao-Someki, line 1: ditto

test_utils/espnet3/tb/optim_test/version_0/events.out.tfevents.1789600335.creativeself.476040.0
Masao-Someki, line 1: Is this used from test? If not could you remove this?

Masao-Someki (commented):
Thank you @aalmuarik, the documentations of this PR is quite nice.
Could you add `metrics/` and `metrics.yaml`, otherwise we cannot compute any metrics with this recipes.

test_utils/espnet3/tb/optim_test/version_0/events.out.tfevents.1789600335.creativeself.476040.0
aalmuarik: Removed, together with hparams.yaml. Neither is used by any test. espnet3's
own test suite writes them, they aren't gitignored, and they got swept into a
commit by mistake.

test_utils/espnet3/tb/optim_test/version_0/hparams.yaml
aalmuarik: Removed.

test/espnet3/systems/knnvc/test_wavlm_encoder.py
aalmuarik: Removed.

espnet3/systems/knnvc/wavlm_encoder.py
aalmuarik: Moved. The URL now lives only in egs3/TEMPLATE/knnvc/conf, where it already
was, so the package no longer duplicates it. checkpoint is now a required
argument, and KNNVCModel raises a clear error if it gets neither an encoder
nor a checkpoint.

espnet3/systems/knnvc/model.py
aalmuarik: Removed. The docstring now only describes the fields a conversion dataset has
to provide.

egs3/TEMPLATE/knnvc/conf/metrics.yaml
aalmuarik: Yes, this PR trains the HiFi-GAN vocoder. I added
espnet3/systems/knnvc/metrics/intelligibility.py, which transcribes the
converted audio and scores it against the source transcript. That is the
WER/CER protocol the kNN-VC paper reports, and both recipes now have a
metrics.yaml.

I ran it on 20 dev-clean conversions made with the authors' released vocoder:
2.71% WER and 0.90% CER. The same ASR on the original audio gives 1.48% and
0.32%, so conversion adds about 1.2 points of WER.

That scores the whole conversion, though, not the vocoder on its own. The
usual way to measure the vocoder alone is copy-synthesis: vocode each test
utterance from its own features and compare it to the original with MCD or
F0 RMSE (egs2 has evaluate_mcd.py and evaluate_f0.py). The validation mel L1
loss tracked during training is already one such measure, and it is what
picks the averaged checkpoint. Copy-synthesis needs an inference mode that
skips the kNN matching. Would you like MCD in this PR, or as a follow-up?

Also, mini_an4/knnvc has a metrics.yaml, but I left measure out of the CI
stages. It downloads an ASR model and takes a few minutes, while the rest of
that test runs in about ten seconds without downloads. Should I add it to CI
anyway?

egs3/librispeech_100/knnvc/dataset/dataset.py
aalmuarik: Done. I added the same one-line docstrings to the mini_an4 recipe too.

egs3/librispeech_100/knnvc/dataset/dataset.py
aalmuarik: Added example output for each kind, using real shapes from dev-clean with
the recipe defaults.

PR #6706, opened 2026-09-18
Say what to do when from_pretrained cannot load a published model

sw005320:
## What

I loaded the 19 most-downloaded `espnet/*` models on the Hub with the current espnet and ran one inference on each. 15 worked; 3 failed with messages that did not say what was wrong. Two of those belong here (the third is [espnet_model_zoo#93](https://github.com/espnet/espnet_model_zoo/pull/93)).

**1. A model published with espnet3's `pack_model` fails in the espnet2 constructor.** `Classification.from_pretrained("espnet/meld_cls_wavlm_base_plus")`:

```
TypeError: Classification.__init__() got an unexpected keyword argument 'training_config'
```

The downloader hands back that bundle's `inference_config` / `training_config`, which the espnet2 inference classes do not take — such a model has to be loaded through `espnet3.publication.inference_model.InferenceModel`. The 25 `from_pretrained` sites each held their own copy of the same model_zoo boilerplate; they now call `espnet2.utils.pretrained.download_pretrained()`, which keeps the "install espnet_model_zoo" message in one place and turns the above into:

```
RuntimeError: espnet/meld_cls_wavlm_base_plus was published with espnet3's pack_model;
load it with espnet3.publication.inference_model.InferenceModel.from_pretrained(
'espnet/meld_cls_wavlm_base_plus') instead (trust_user_code=True if the bundle ships its own code).
```

I checked that route works: `InferenceModel.from_pretrained(tag, trust_user_code=True)` builds the model (its own remaining failure is a missing `s3prl`, which that model genuinely needs).

**2. Japanese TTS dies on its first sentence.** `Text2Speech.from_pretrained("espnet/kan-bayashi_jsut_vits_prosody")` loads, then raises a bare `ModuleNotFoundError: No module named 'pyopenjtalk'` from inside `text2tokens`. pyopenjtalk publishes no wheels (sdist only, needs cmake), so it cannot be a dependency of the `tts` extra. Instead:

- the three `import pyopenjtalk` sites go through `_import_pyopenjtalk()`, whose `ImportError` names both install routes (`pip install pyopenjtalk`, `tools/installers/install_pyopenjtalk.sh`) and says why it is in no extra;
- `PhonemeTokenizer.__init__` calls it when a `pyopenjtalk*` `g2p_type` is selected, so a model that cannot possibly speak fails at load time instead of on the first sentence.

## Checked

| | |
|---|---|
| new unit tests (`test_pretrained.py`) | 4 |
| phoneme tokenizer tests | 36 passed |
| every changed module imports | yes (`asr_inference_k2` needs `k2`, as on master) |
| `ci/check_inference_imports.py` in a declared-deps-only venv | OK: 26 modules |
| isort / black / flake8 | clean |

Net −211 lines: 25 copies of the boilerplate become one call each.


sw005320:
@coderabbitai full review

sw005320:
`codecov/patch` reports 64.93% of the diff hit against a 72.45% target. Where the diff is, in added lines:

| | lines |
|---|---|
| `from_pretrained` call sites (one per class) | 26 |
| their import line | 26 |
| `_import_pyopenjtalk()` guards in PhonemeTokenizer | 8 |
| the helper, its tests, and the tokenizer's import site | 91 |

`espnet2/utils/pretrained.py` is at 100% (4 new unit tests). The uncovered lines are the `from_pretrained` bodies: reaching them means downloading a model from the Hub, which the suite does not do — those same lines were uncovered before this PR, as the boilerplate they replace. Each one is now a single call to a tested function instead of eleven untested lines, so the ratio moved even though nothing got less tested.

I can raise the number by monkeypatching the downloader and each class's constructor and asserting the call, but that would assert the plumbing rather than any behaviour. Happy to add it if you would rather see the check green — say the word.

sw005320:
Added the test rather than leaving the check red, since it turned out to be worth having on its own.

`test_from_pretrained_routing.py` derives the list of inference classes from the source (26 of them, in 25 files) and, for each, calls `from_pretrained` with the download and the constructor replaced. It asserts two things: that the class fetches through `download_pretrained` and passes the artifacts to its constructor, and that no tag means no download. Nothing is fetched, so it runs in about 12 seconds.

That is the invariant this PR is built on — one helper, so an espnet3 bundle is recognised in one place — and a new inference class that reintroduces its own boilerplate now fails it. It also covers the 26 call sites, which is where the patch coverage went.

`asr_inference_k2` and `uasr_inference_k2` skip without k2, as elsewhere in the suite. 55 passed, 2 skipped.

PR #6707, opened 2026-09-18
Point the demo links at demos that work, and notice when one rots

sw005320:
## What I found

Checked every demo link in the repository's markdown against the two APIs that can actually tell whether it works. Neither failure mode answers 404, which is why none of this ever turned a check red: Colab renders "Notebook not found" for a moved notebook, and a Space whose container has died still serves its page.

| link | state |
|---|---|
| `akhaliq/espnet2_asr` (ASR demo) | RUNTIME_ERROR, last modified 2021-08-31 |
| `akhaliq/ESPnet2-TTS` (TTS demo) | RUNTIME_ERROR, 2021-10-18 |
| `Siddhant/ESPnet2-SLU` (SLU demo) | RUNTIME_ERROR, 2022-05-24 |
| `Siddhant/Voice_Assistant_Demo` (SDS README) | RUNTIME_ERROR, 2025-02-12 |
| `espnet2_tts_realtime_demo.ipynb` (README + tts1 README) | moved when espnet/notebook was reorganised |

Meanwhile the organisation's own Spaces are up: `espnet/OWSM_V4_Demo` (zero-a10g, 11 likes), `espnet/TTS`, `espnet/cvss-c_es-en_s2st`, `espnet/TheESPnetLeaderBoard`, `espnet/meld_cls_wavlm_base_plus`. The README linked none of them.

## Changes

- ASR and TTS demo links → `espnet/OWSM_V4_Demo` and `espnet/TTS`.
- TTS notebook → `ESPnet2/Demo/TTS/tts_realtime_demo.ipynb` (README and `egs2/TEMPLATE/tts1/README.md`).
- SLU and Voice-Assistant links dropped: those Spaces error, and so do the organisation's copies (`espnet/SingingSDS`, `espnet/Voice_Assistant_Demo`, `espnet/svs` are all RUNTIME_ERROR), so there is nothing to point at yet. Reviving one and restoring the link is a one-line change — worth doing, and the checker below will then keep it honest.
- `ci/check_demo_links.py`: for every notebook link, the path must still exist in `espnet/notebook`'s file list; for every Space link, the runtime stage must not be an error. A link into a Space's *files* (`/tree/main`, how `espnet2/sds/.../mini_omni/README.md` cites adapted code) is deliberately not checked as a demo.
- `.github/workflows/check_demo_links.yml`: daily at 07:00 UTC, one issue edited in place, same shape as `report_broken_workflows.yml`. Not per pull request: a Space that happens to be restarting must not fail someone's build.

## Checked

```
self-check ok
checked 6 notebook links and 2 Space links
$ echo $?
0
```

`ci/check_ci_image_config.py` passes (new workflow: pinned action SHA, `persist-credentials: false`, declared permissions). Two `colab.research.google.com/drive/...` links (speech enhancement, SLU) are personal Drive notebooks and cannot be verified through an API; they are left alone.


sw005320:
@coderabbitai full review

sw005320:
Addressed: an unreadable tracked markdown file now raises `ScanError` instead of being skipped, so the run exits 2 and the step fails rather than closing the report issue. The handle is in a `with` block. Self-check and a real scan still pass (6 notebook links, 2 Space links, exit 0).

sw005320:
@coderabbitai full review

sw005320:
Both addressed:

- `markdown_files()` turns an `OSError` or a failed `git ls-files` into `ScanError`, so git failing exits 2 and fails the step instead of reading as "no broken links" and closing the issue.
- Notebook links keep their ref: `find_links` returns `(ref, path)` pairs and each ref's tree is fetched once and cached, so a link to a tag or a branch is checked there rather than against master. Every link in the repository still uses master, so the scan is unchanged in practice; the self-check covers a `v1.0` link.

Self-check passes and a real scan reports 6 notebook links, 2 Space links, exit 0.

sw005320:
@coderabbitai full review

sw005320:
Addressed. `utils/synth_wav.sh` sent beginners to the repository-root `tts_realtime_demo.ipynb`, which is the ESPnet1 notebook's old path; it points at `ESPnet1/tts_realtime_demo.ipynb` now.

The scanner was the reason that one survived, so it reads tracked `*.sh` as well as `*.md` (7 notebook links now instead of 6). That surfaced three tracked scripts under `egs2` that are symlinks to files not in the repository (`egs2/TEMPLATE/asr1/utils/simple_dict.sh`, `egs2/babel/asr1/local/make_corpus_subset.sh`, `egs2/librispeech/ssl1/hubert.sh`); a dangling symlink holds no text here, so it is skipped, while a real read error still fails the scan. Those three look like leftovers worth a separate look.

Self-check passes, scan exits 0.

sw005320:
@coderabbitai full review

sw005320:
Addressed. The symlink's target is `os.stat`'d explicitly: `ENOENT`/`ENOTDIR` is the case this skip is for (the three egs2 scripts pointing outside the repository), and every other `OSError` raises `ScanError` so the file is never silently dropped. `os.path.exists()` had made a permission error and a missing target indistinguishable.

Self-check passes, scan reports 7 notebook links and 2 Space links, exit 0.

sw005320:
@coderabbitai full review

sw005320:
Right — the script cannot run at all, so repairing its link was the wrong fix. Removed, along with its ASR twin:

| | why it is dead |
|---|---|
| `utils/synth_wav.sh` | refuses to start outside `egs/<corpus>/tts1`; `egs/` is gone from the repository; calls `tts_decode.py`, also gone; referenced by nothing |
| `utils/recog_wav.sh` | same shape: wants `egs/tedlium2/asr1`, calls `asr_recog.py`; referenced by nothing |

640 lines of ESPnet1 residue, and both would have sent anyone who found them to a directory that does not exist.

The checker keeps reading tracked `*.sh` even though no script carries a demo link now — a script's help text is exactly where the next one will appear.

**Related residue, left for a separate decision.** 13 more files under `utils/` are referenced nowhere in the repository, most of them ESPnet1-era JSON tooling:

```
clean_corpus.sh  concat_json_multiref.py  convert_fbank.sh  divide_lang.sh
eval_source_separation.sh  feats2npy.py  generate_wav.sh  json2sctm.py
result2json.py  score_bleu.sh  score_sclite_case.sh  score_sclite_wo_dict.sh
show_result.sh  speed_perturb.sh
```

`docker/build.sh` also still refers to `egs/…`. Happy to open that as its own pull request if you want it cleared.

sw005320:
@coderabbitai full review

sw005320:
One more, found while verifying that the OWSM-CTC demo actually runs before it is published: the upload command in its README could not have worked.

```
hf upload spaces/espnet/owsm-ctc-v4 ... --repo-type space
```

huggingface_hub reads a `spaces/` prefix only inside an `hf://` URI. As a plain argument it is part of the id, and `validate_repo_id` rejects the second slash: *Repo id must be in the form 'repo_name' or 'namespace/repo_name'*. The id is `espnet/owsm-ctc-v4`, with `--repo-type space` saying what it is; `hf upload` then creates the Space with the gradio SDK if it does not exist.

The demo itself is verified: loads `espnet/owsm_ctc_v4_1B` (151 languages, 25 translation targets), transcribes the example clip in about 11 s on this laptop's CPU and translates it into Japanese.

sw005320:
Addressed: `NO_APP_FILE` is in `BROKEN_STAGES` now. It is worth having for a specific reason — it is the state an upload that sent the wrong directory leaves behind, so it is the failure most likely to follow publishing the OWSM-CTC demo, and the scan would have called it clean.

sw005320:
@coderabbitai full review

sw005320:
Addressed, and widened a little: `IncompleteRead` is caught in `_get_json` as you describe, and `main()` now turns *anything* unforeseen into status 2 with the traceback, so no future exception can be read as "the links are broken".

```
$ python3 -c '...m.scan = raise RuntimeError("boom"); sys.exit(m.main())'
exit 2 | scan failed: unexpected error
```

Self-check passes, real scan reports 6 notebook links and 2 Space links, exit 0.

sw005320:
@coderabbitai full review

sw005320:
Addressed: `space_stage` raises `ScanError` when the Hub's answer carries no runtime stage, instead of returning "UNKNOWN" — which was not in `BROKEN_STAGES`, so a Space whose state could not be read passed as working. A Space that is gone still comes back `DELETED`.

```
runtime missing  -> ScanError: espnet/x: the Hub's answer carries no runtime stage
Space gone       -> DELETED
```

Self-check passes, real scan reports 6 notebook links and 2 Space links, exit 0.

sw005320:
@coderabbitai full review

sw005320:
Addressed: both issue-maintenance blocks `set -euo pipefail`, so a failed `gh issue list` cannot leave `existing` empty and have the step file a duplicate.

Worth noting that `report_broken_workflows.yml` — which this workflow is modelled on — has the same two pipelines without `pipefail`. Same latent duplicate-issue bug. I left it alone to keep this PR to one subject; happy to send it as a one-line follow-up, or to fold it in here if you prefer.

YAML parses, `ci/check_ci_image_config.py` passes.

sw005320:
@coderabbitai full review

sw005320:
Addressed. `blob/<ref>/<path>` has nothing marking where the ref ends, so `split_ref` moves the boundary rightwards until a ref resolves: `master` still costs one request (cached across links), a branch like `feature/demo` is found on the second candidate, and only a link whose ref does not exist at any split fails the scan. The ref is URL-encoded, so a slash is not read as a path separator.

`notebook_paths` now returns `None` for "no such ref" rather than raising, since a 404 there is a candidate being rejected, not a scan failure — a truncated tree or a malformed answer still raises.

The `--self-check` covers the boundary offline: against a repository where only `feature/demo` exists, it asserts the two candidates tried in order and the resulting split, and that a ref which never resolves raises `ScanError`.

```
self-check ok
checked 6 notebook links and 2 Space links
```

sw005320:
@coderabbitai full review

sw005320:
`report_broken_workflows.yml` has the same `pipefail` fix now, in this PR as you asked — it carried the same two `gh issue list | head -1` pipelines.

Then I read the scanner as a whole instead of one comment at a time, and found four more. All of the same shape: the scan could report *clean* or *wrong* rather than admit it had not looked.

1. **It scanned the wrong files when run from anywhere but the repository root.** `git ls-files` lists relative to the working directory, so from `egs2/` it saw only what was below and called the rest fine. It resolves the root with `rev-parse --show-toplevel` and reads from there. Verified by running it from `egs2/`: same 6 notebook links and 2 Space links as from the root.
2. **`split_ref` took the first ref that resolved.** With a slash-containing branch, a shorter ref that also exists would win and a live notebook would be reported gone. The split that *finds the file* wins now; if no split holds it, it is reported against the ref that exists. Two more self-check cases cover this.
3. **Unauthenticated GitHub allows 60 requests an hour per address**, which a shared Actions runner can exhaust — the daily run would then fail on somebody else's traffic. The scan sends the workflow's `GITHUB_TOKEN` to `api.github.com` (read-only, only for the notebook repository's file list). Checked that a bad token fails loudly: `scan failed: … HTTP 401`, exit 2.
4. **The self-check injected its stub by rebinding a module global.** `split_ref` takes the fetcher as an argument instead.

Also rebased onto master so CI runs against the current tree (#6706 and #6708 have landed since this branch was cut).

Full scan after all of it: 6 notebook links, 2 Space links, exit 0; `ci/check_ci_image_config.py` passes.

sw005320:
@coderabbitai full review

sw005320:
All four addressed, in both workflows where they apply.

1. **`head -1` with `pipefail`** — right, and it is the pipefail from the previous round that made it dangerous. `first(...)` inside gh's own `--jq` does the same job with no second process to close the pipe.
2. **Matching on title alone** — both lookups now also require the body to carry the `Maintained by <workflow>` marker. The report writes that marker from the same `MARKER` variable, so the text the workflow looks for and the text it writes cannot drift apart. Checked with jq against a list containing an impostor issue of the same title: ours is selected whichever comes first, nothing is returned when only the impostor exists, and the exit status is 0 in both cases.
3. **Trailing slash on a Space URL** — accepted now. The cases I checked:

   | link | recorded |
   |---|---|
   | `…/spaces/espnet/TTS` | espnet/TTS |
   | `…/spaces/espnet/TTS/` | espnet/TTS |
   | `[x](…/spaces/espnet/TTS/)` | espnet/TTS |
   | `…/spaces/gradio/omni-mini/tree/main` | not a demo link |
   | `…/spaces/espnet/TTS/blob/main/app.py` | not a demo link |

4. **`DELETING`** — added to `BROKEN_STAGES`.

The one failing test, `test_SizedDict_shared`, is a 5-second timeout in a multiprocessing test (`AttributeError: 'ForkAwareLocal' object has no attribute 'connection'`). This PR changes no Python outside `ci/`, so it is not from here; I have re-run the job.

sw005320:
@coderabbitai full review

sw005320:
Correction on the last line: I did not re-run that job in the end — the push above started a fresh CI run, which supersedes it. If `test_SizedDict_shared` times out again there, it is worth a look of its own; it has nothing to touch in this PR.

sw005320:
Both addressed, the first one by turning the list around.

`STOPPED` was missing, and so was `PAUSED` — my own comment claimed a paused Space "wakes up on a visit", which is wrong: only its owner can restart one. Naming the broken stages means anything the Hub adds later, or that I failed to think of, passes as a working demo. So the set is now the stages that *work* — running, still starting, or asleep, since a visit does wake a sleeping Space — and everything else is reported.

```
RUNNING          -> fine            PAUSED         -> reported broken
SLEEPING         -> fine            STOPPED        -> reported broken
                                    DELETING       -> reported broken
                                    SOMETHING_NEW  -> reported broken
```

On the encoding: `find_links` decodes both captures now, since `notebook_paths` re-encodes the ref when it fetches the tree and the tree lists decoded paths. The self-check covers `blob/feature%2Fdemo/a%20b.ipynb` becoming `("feature/demo", "a b.ipynb")`.

sw005320:
@coderabbitai full review

PR #6708, opened 2026-09-18
Assert that every inference class fetches through the shared helper

sw005320:
## What

Follow-up to #6706, which merged before this test was pushed.

#6706 replaced 26 copies of the `espnet_model_zoo` boilerplate with one call to `espnet2.utils.pretrained.download_pretrained()`, so that a bundle packed by espnet3 is recognised in one place instead of failing inside an espnet2 constructor with `unexpected keyword argument 'training_config'`. Nothing held that invariant: a new inference class can reintroduce its own copy and nothing notices.

`test/espnet2/bin/test_from_pretrained_routing.py` derives the class list from the source (26 classes in 25 files) and, for each, calls `from_pretrained` with the download and the constructor replaced. It asserts that the class fetches through `download_pretrained` and hands the artifacts to its constructor, and that no model tag means no download. Nothing is fetched, so it runs in about 13 seconds.

It also covers the 26 rewired call sites, which is where #6706's patch coverage went.

## Checked

```
51 passed, 2 skipped in 13.64s
```

The two skips are `asr_inference_k2` and `uasr_inference_k2` without k2, as elsewhere in the suite. flake8 clean.


sw005320:
@coderabbitai full review

sw005320:
Addressed the first point, and it found something.

The discovery skipped files that did not already contain a `download_pretrained` call — so a class fetching its own way would have been excluded from the test meant to catch exactly that. Membership now comes from the signature: `from_pretrained` takes a `model_tag`.

That admits one class the filter had hidden: `s2st_inference.Speech2Speech.from_pretrained` takes only `vocoder_tag`, so **an S2ST model cannot be loaded from a published tag at all**. Rather than drop it silently, `test_only_s2st_offers_from_pretrained_without_a_model_tag` names it; if that gap is closed the class joins the routing cases and the expectation goes away. 52 passed, 2 skipped.

On the second point — splitting into one test file per `espnet2/bin` module — I have not done it, and I think it would make this test worse rather than better. What it asserts is not a property of any one module but of the set: *every* inference class that accepts a `model_tag` routes through the one helper. Twenty-five files each holding a copy of the same body would duplicate the assertion, and the part that matters most — that the set is complete, so a class added tomorrow is covered without anyone remembering to add a file — cannot be expressed per module at all. The path rule fits tests of a module's behaviour; this one's subject is the boundary between them. Happy to reconsider if a maintainer prefers the split.

sw005320:
@coderabbitai full review

PR #6709, opened 2026-09-18
Keep the demo's card inside the limits the Hub enforces

sw005320:
## What happened

Publishing the OWSM-CTC demo as a Space failed on its first try:

```
Error: Invalid value. Invalid metadata in README.md.
- "short_description" length must be less than or equal to 60 characters long
```

Mine was 67. The Space is up now (`espnet/owsm-ctc-v4`, building at the time of writing), but nothing in this repository could have told anyone in advance: the card lives here and is validated only when someone uploads it — the one moment a demo is being published, which is a poor time to find out.

## Changes

- `short_description` is 54 characters: *ASR, speech translation and language ID in one encoder*.
- `ci/check_demo_links.py` reads the Space cards under a `demo/` directory and reports what the Hub would refuse: the length limits on `short_description` and `title`, a missing `title`/`sdk`/`app_file`, an `app_file` that is not beside the card, and front matter that is not valid YAML. A plain README in a demo directory is not treated as a Space card.

## Checked

The self-check now includes the card that actually failed:

```
self-check ok      # asserts: short_description is 67 characters; the Hub allows 60
checked 6 notebook links and 2 Space links
```

`ci/check_ci_image_config.py` passes, flake8 clean.


sw005320:
@coderabbitai full review

sw005320:
All three addressed, and two were defects in what this PR adds.

| finding | what it did | now |
|---|---|---|
| `app_file` resolved against the working directory | a scan from a subdirectory reported the `app.py` beside the card as missing | resolved from the repository root; a value that climbs out of the directory that becomes the Space is reported rather than followed |
| front matter opened with `---` and never closed | parsed anyway, so a card the Hub reads no metadata from could pass | reported: *front matter is opened with --- but never closed* |
| `title`/`sdk` accepted on truthiness | `title: true`, `sdk: 1` passed | must be non-empty text, as the Hub requires |

A non-string `app_file` was the sharpest of them: `Path(path).parent / True` raises `TypeError`, which under `main()`'s safety net ends the whole scan with status 2 rather than reporting one bad card. It is validated as text first now.

The self-check covers all four cases, plus the card that actually failed the upload this afternoon. Scan from the root and from `egs2/owsm_ctc_v4`: same result, exit 0.

sw005320:
@coderabbitai full review

PR #6710, opened 2026-09-18
Give ESPnet a command line for trying a published model

sw005320:
## Why

Every `espnet2.bin.*_inference` module has a command line already — but it is the one a recipe needs. To transcribe one file today:

```sh
printf 'utt1 audio.wav\n' > audio.scp
python -m espnet2.bin.asr_inference --output_dir out \
    --data_path_and_name_and_type audio.scp,speech,sound --model_tag espnet/...
```

`--output_dir` and `--data_path_and_name_and_type` are both required, so someone who just wants to hear what ESPnet does has to learn the scp format or write Python first. That is the whole distance for a first-time user, and it is the same distance the demos we published this week were meant to close.

## What

```sh
espnet asr audio.wav                       # transcribe, detecting the language
espnet asr audio.wav --language jpn
espnet translate audio.wav --to eng
espnet tts "Hello from ESPnet" -o out.wav
espnet enhance noisy.wav -o clean.wav
espnet models                              # the default model of each command
```

- One console script, `espnet2/bin/cli.py`, beside the existing `espnet-mcp`.
- A default model per task, each from the organization; `--model <tag>` replaces it, `--device` takes cuda or mps.
- Imports happen inside each subcommand, so `--help` and `models` answer without loading torch, and the module imports on a bare inference install — it is in `ci/check_inference_imports.py` now (27 modules).
- A missing file is reported *before* a model is downloaded. `espnet asr nope.wav` is one line on stderr and exit 1, not a 4 GB download followed by a traceback.
- `enhance` writes `clean.spk1.wav`, `clean.spk2.wav` … when the model separates speakers, and `clean.wav` when it does not.

The README's quick start now leads with this, and links [the OWSM-CTC v4 Space](https://huggingface.co/spaces/espnet/owsm-ctc-v4) published today beside it.

## Checked

Run against the real models on this laptop (CPU):

| command | result |
|---|---|
| `espnet asr ctc_align_test.wav` | *The sale of the hotels is part of Holi's strategy to sell off assets and concentrate on property management.* |
| `espnet translate … --to jpn --language eng` | ホテルの販売は、資産を売し、財産管理に集中するため休戦略の一部です。 |
| `espnet tts "Hello from ESPnet"` | wrote a 65 kB wav |
| `espnet enhance …` | wrote a 209 kB wav |
| `espnet asr /nowhere.wav` | `espnet: no such file: /nowhere.wav`, exit 1 |

13 unit tests cover the dispatch without downloading anything: which class is asked for which tag, the language and task tokens built from the flags, the per-speaker output naming, and that a missing file fails before any model is fetched. `ci/check_inference_imports.py` passes in a declared-deps-only venv, including the new module.


sw005320:
@coderabbitai full review

sw005320:
All four addressed. The first was a real defect and a good catch.

**Channels.** `_load_audio` averaged multichannel input to mono, so `espnet enhance` on a multichannel recording threw away exactly what a beamformer exists to use. `SeparateSpeech` takes `(Batch, Nsamples [, Channels])`, so the channels are passed through now, and a test asserts the shape survives. Checked on a real 2-channel file with the default model: writes the enhanced wav as before.

**Output paths.** `-o clean` produced `.spk1.clean`, and `-o runs.v1/clean` produced `runs.spk1.v1/clean` — a different directory. Both branches use `Path` now, and an output with no extension is refused before the model loads, since soundfile cannot infer a format from one either:

```
$ espnet enhance a.wav -o /tmp/clean
espnet: output needs a file extension, e.g. /tmp/clean.wav
```

**"any tag".** Right — each command loads one task's class. The wording is fixed in the module docstring, `--model`'s help, `espnet models` and the README, and a mismatched tag is now explained rather than surfacing as an unexpected-keyword `TypeError`:

```
$ espnet tts "hi" --model espnet/owsm_ctc_v4_1B
espnet: espnet/owsm_ctc_v4_1B does not look like a model for `espnet tts`
(Text2Speech.__init__() got an unexpected keyword argument 's2t_train_config').
Pass --model with a tts model; `espnet models` names the default.
```

**Exit codes.** Asserted exactly, including argparse's 2 for a missing `--to` and 0 for `--help`.

16 tests pass, none of them downloading anything; `ci/check_inference_imports.py` still passes with the CLI in its list.

sw005320:
@coderabbitai full review

sw005320:
Addressed the review. The first point was the important one.

| review point | what changed |
|---|---|
| `_build()` caught every `TypeError` | translates only an unexpected keyword argument the class genuinely does not take, verified against its signature; anything else re-raises untouched |
| `tts` validated its output after loading | validates first, as `enhance` does — `-o out` no longer costs a download |
| filesystem errors escaped raw | a failed write is a `CLIError`: `espnet: cannot write /tmp/gone/x.wav: …` |
| `--device` took any string | the parser accepts `cpu`, `mps`, `cuda`, `cuda:<n>` and rejects the rest with exit 2 |
| `--language` semantics | documented as OWSM's token rather than a generic ASR concept, in the docstring and both help strings |
| tests were mostly mocked dispatch | the command now also runs as a process: `--help` (0), `models` (0), a missing file (1, one line, no traceback), an unknown subcommand (2), a bad device (2) |

On the `TypeError` narrowing specifically, two tests hold the line: a `TypeError` raised from inside model construction, and one naming a keyword the constructor *does* accept, both reach the user unchanged rather than being reported as "wrong model".

29 tests, none downloading anything. `ci/check_inference_imports.py` passes with the CLI in its list.

**Not done, deliberately**: a CI check that each tag in `DEFAULT_MODELS` still exists and still suits its command. It is the right idea — those tags are the CLI's public surface and a rename would break `espnet asr` silently — but it needs the network, so it belongs in the daily `check_demo_links` workflow rather than here, as its own change.

PR #6711, opened 2026-09-18
Two OWSM v4 demos with the same tasks, one per model

sw005320:
## What

Two Spaces, both maintained from this repository:

| | `espnet/owsm-ctc-v4` (published) | `espnet/owsm-v4` (to create) |
|---|---|---|
| model | `owsm_ctc_v4_1B`, encoder-only | `owsm_v4_medium_1B`, encoder-decoder |
| speech recognition | yes | yes |
| speech translation, any-to-any | yes | yes |
| language identification | yes | yes |
| long-form | yes | yes |
| text prompt | no — OWSM-CTC does not support it | yes |
| decoding | one encoder pass per 30 s window | beam search |
| source | `egs2/owsm_ctc_v4/s2t1/demo` | `egs2/owsm_v4/s2t1/demo` |

Both run under ZeroGPU through `@spaces.GPU`, behind a shim so `python app.py` also works locally, where `DEVICE` picks the device.

The v4 app starts from what `espnet/OWSM_V4_Demo` runs today, minus what turned out to be obsolete:

- the block that downloaded the model and then **rewrote its `config.yaml`** to delete every `gradient_checkpoint_layers` key. Current espnet takes that argument (`EBranchformerEncoder`), and the path it patched is the pre-0.1.8 cache location. It also edited a file inside the Hub cache, which is the bug we fixed in the downloader today;
- `espnet @ git+…@v.202412` and a 2024 `espnet_model_zoo` commit, which is why that hack was needed and why the Space cannot pick up a fix — including the one that stops model vocabularies being corrupted with the local cache path;
- the 152-line language table: both apps read the menus from the checkpoint. The language symbols are the three-letter tokens before `<asr>`, the targets the `<st_xxx>` ones, which holds for both model families, so a checkpoint covering more languages needs no edit here.

Language identification differs because the models do: OWSM-CTC names the language in its own output when asked with `<nolang>`, so one short pass gives both the language and the text; the autoregressive model uses `Speech2Language` on the same checkpoint.

## Checked

Both apps end to end on this laptop's CPU, 151 languages and 25 targets each:

| | CTC | v4 |
|---|---|---|
| transcribe + detect | English, *"The sale of the hotels is part of Holi's strategy…"* | English, *"…part of holiday strategy…"* |
| translate to Japanese | ホテルの販売は、資産を売し、財産管理に集中するため休戦略の一部です。 | ホテルの販売はホリデーの戦略の一部です… |
| long-form | text | `[0.0 - 6.54] …` with timestamps |
| text prompt | not offered | *"Holiday Inn"* → *"Holiday's strategy"* (visibly steers it) |

Both cards pass the Hub rules `ci/check_demo_links.py` enforces in #6709.

## What is left to you

`espnet/owsm-v4` does not exist yet, so **nothing in this PR links to it** — the README and the cards name it in code, and those become links once it is up. To create it:

```sh
hf upload espnet/owsm-v4 egs2/owsm_v4/s2t1/demo . --repo-type space
```

then set it to `zero-a10g` in the Space's settings, as `OWSM_V4_Demo` is. Re-uploading `espnet/owsm-ctc-v4` from this branch gives it the same four tasks.


sw005320:
@coderabbitai full review

sw005320:
@coderabbitai full review

sw005320:
Rebased onto master (the conflict was `short_description`, where I kept the line describing the four tasks — 53 characters, inside the limit), and the CI failure is fixed.

**The lint failure was mine**: I checked these files with `--max-line-length 100` rather than the repository's own 88, so two lines slipped through. Both wrapped; `flake8`, `black`, `isort` and the pre-commit hooks all pass at the project's configuration now.

**On the two inline comments** (`maxlenratio = -300` left on the shared object, and the un-trimmed input): both were against the first commit, which was the faithful copy of what the Space runs today. The rewrite in the second commit replaced that file — the current `predict` has no `maxlenratio` assignment at all (it is set once in the constructor), no bare `except`, and trims to the 30 s window before both language identification and decoding, as quoted below. So there is nothing left to restore in a `finally`.

```python
@spaces.GPU
def predict(audio_path, language_label, task_label, long_form, prompt):
    speech = _read(audio_path)
    if language_label == DETECT:
        code = s2l(_pad(speech[: SAMPLE_RATE * WINDOW_SECS]))[0][0].strip()[1:-1]
    ...
        text = s2t(
            _pad(speech[: SAMPLE_RATE * WINDOW_SECS]),
            prompt or "<na>",
            lang_sym=lang_sym,
            task_sym=task_sym,
        )[0][-2]
```

**Also fixed since the last review**: both apps chose their device with `torch.cuda.is_available()` at import. On ZeroGPU that is False — the GPU is attached only while a `@spaces.GPU` function runs — so the models would have loaded onto the CPU on the very hardware bought to run them. `DEVICE` wins if set, then CUDA whenever the runtime is ZeroGPU, then CUDA if torch sees a GPU, else CPU. Checked over all five combinations, and the CTC demo still runs locally on the CPU.

sw005320:
@coderabbitai full review

sw005320:
Four findings, two subjects. One is fixed; on the other I disagree, with a reason.

**The ZeroGPU time limit — fixed, and it mattered.** ZeroGPU gives a `@spaces.GPU` call a slice of GPU time and kills it when the slice ends, so long-form decoding of a long recording would have been cut off rather than answered. Both demos ask for a slice (`@spaces.GPU(duration=120)`) and refuse audio over two minutes:

```
150 s file: Error: This demo takes up to 120 s; that file is 150 s.
            Run the app yourself for longer audio - the model has no such limit.
```

That is what the Space being replaced did with `gr.Audio(max_length=120)`, an argument gradio 6 removed — which is how I dropped the guard. The local shim takes the decorator's arguments too, so `python app.py` still works, and both cards now say the limit is the demo's rather than the model's.

**The `LANGUAGE_NAMES` table — kept in both files, deliberately.** A Space is exactly one directory: `hf upload espnet/owsm-v4 egs2/owsm_v4/s2t1/demo .` uploads that directory as the Space's root, and nothing outside it. A shared module elsewhere in the repository would not be uploaded, so the import would fail on the Hub — the Space would build and then crash on the first request. The two copies are a table of ISO 639-3 names, pure data, identical by construction; the Space this replaces carries the same table inline for the same reason, at 152 lines rather than 117.

What I did remove is the part that could drift: the *list* of languages and translation targets is read from the checkpoint, so a model covering more languages needs no edit. A name missing from the table falls back to its code, so the table being incomplete degrades a label, never the behaviour.

If you would rather not have the duplication at all, the alternative is a package like `langcodes` in both `requirements.txt` — happy to do that instead, though it trades 117 lines of data for a dependency that can break a Space build.

sw005320:
@coderabbitai full review

sw005320:
Two findings.

**The counts — fixed.** "recognises, translates and identifies 151 languages" did read as though translation covered all 151. It is 151 in, 25 out; the README line and both Space cards say that now.

**Loading the models at import rather than inside `@spaces.GPU` — keeping it, on evidence.** `espnet/OWSM_V4_Demo` has run on `zero-a10g` for a year doing exactly this: it builds `Speech2Language`, the autoregressive `Speech2Text` and the CTC model at module scope with `device="cuda"`, and only `predict` is decorated. That is the pattern these two apps inherit, and it is what ZeroGPU's own `spaces` package is there to support — it is imported before torch touches CUDA, which is why the import-time transfer is deferred until a decorated call runs.

Moving construction inside `predict` would also cost something real: the menus are built from the checkpoint's token list at import, so the model would have to load twice, or the interface would have to be built lazily on the first request — and a 1B checkpoint loading inside a time-limited GPU slice eats the slice the request needs.

If it turns out I am wrong about the runtime, the symptom is loud and immediate (the Space errors on its first request rather than misbehaving quietly), and the fix is a lazy global. I would rather match the Space that demonstrably works than change the pattern on a hypothesis — happy to revisit if you have seen it fail.

sw005320:
@coderabbitai full review

sw005320:
Thank you — and you were right that the head had moved. Three of the five describe the first commit, which was the faithful copy of what `espnet/OWSM_V4_Demo` runs; the second commit replaced that file. Checked against the current head:

| your point | current code |
|---|---|
| `maxlenratio = -300` left on a shared object | no assignment anywhere in `predict`; it is set once in the constructor (`maxlenratio=0.0`) and never mutated, so nothing leaks between requests |
| bare `except:` | none in either app (`grep -n 'except:'` is empty) |
| fallback says "first 30s" but passes the whole waveform | there is no fallback branch; the standard path is given `_pad(speech[: SAMPLE_RATE * WINDOW_SECS])`, and `SAMPLE_RATE` is the rate `librosa.load` resampled to, not an assumption |

The two that are live, both fixed:

**The README** no longer presents `espnet/owsm-v4` as hosted — it is described as the source you can run locally or publish, and nothing links to the 404.

**Device precedence** is now written down, in the code and in both cards: `DEVICE` wins, so `DEVICE=cpu` is a local CPU test on a GPU machine; otherwise CUDA when `SPACES_ZERO_GPU` says the runtime is ZeroGPU, or when torch reports a GPU; CPU otherwise. The marker matters — I had inferred ZeroGPU from `spaces` being importable, and that package installs anywhere, so a CPU-only machine with it installed would have chosen CUDA.

**On the duplication**, I think you have identified the real risk and I do not want to hand-wave it. A Space is exactly one directory — `hf upload espnet/owsm-v4 egs2/owsm_v4/s2t1/demo .` sends that directory and nothing else — so a shared module in the repository would not be uploaded and the Space would crash on its first request. What I can do is make drift *fail*: `test/espnet2/bin/test_demo_apps.py` compares the half the two apps share and fails when the copies stop matching, so a fix applied to one must be applied to the other. It also checks that each demo refuses audio longer than the GPU slice it asks for, that `spaces` is imported before torch, and that each card names the model its app loads.

**Reading both files end to end also turned up three of my own**, which is the part of this I should have done before asking for review:

- the CTC demo reported the language as **"asr"** when asked to detect it. I had narrowed the symbol parser to three-letter tokens, and `asr` is three letters. A symbol now counts as the language only if the checkpoint lists it as one.
- long-form decoded the first window twice, once to name the language and once as part of the whole; it skips that pass when the language is known. A long-form request went from 91 s to 13 s here.
- audio over 30 s with Long-form unticked was silently cut to the first window. It still is — that is what the checkbox is for — but the demo says so now.

Verified after all of it, on CPU: CTC detects English and transcribes, translates into Japanese, and decodes long-form; the autoregressive app does the same with timestamps, and the prompt still steers it (*holiday* → *Holiday's*). 9 drift tests pass, pre-commit is clean.

sw005320:
@coderabbitai full review

PR #6712, opened 2026-09-18
Compare paths as paths when checking a Space's app_file

sw005320:
Follow-up to #6709, from a review that landed as it merged.

`check_space_card` decided whether a Space's `app_file` stays inside the directory that becomes the Space by comparing strings:

```python
if not str(target).startswith(str(directory.resolve())):
```

A sibling whose name begins the same way — `egs2/x/demo-copy` beside `egs2/x/demo` — passes that test while being outside the Space, so a card pointing at `../demo-copy/app.py` would have been accepted and the upload would then fail on the Hub for a reason the check exists to catch.

`Path.is_relative_to` compares path components, which is what was meant. The self-check gains that exact case beside the `../outside.py` one it already had.

```
self-check ok
checked 6 notebook links and 3 Space links
```


sw005320:
@coderabbitai full review

PR #6713, opened 2026-09-19
Link the OWSM v4 Space now that it is up

sw005320:
## What

#6711 deliberately named `espnet/owsm-v4` in code rather than linking it: the Space did not exist yet, and the daily `check_demo_links` job would have reported the 404 — correctly. It is up and running now, so the three places that referred to it become links: the README, the CTC demo's description (its sibling), and the v4 demo's own card.

The effect that matters is not the links themselves: both Spaces are now inside the daily check, so if either stops running an issue is filed rather than the demo quietly rotting, which is how the previous five got to 2021.

## Checked

```
checked 6 notebook links and 3 Space links     # exit 0, so all three are running
```

The drift tests still pass, pre-commit is clean.


sw005320:
@coderabbitai full review

PR #6714, opened 2026-09-19
Bump the version to 202610 for the interim release

sw005320:
## What

One line: `version.txt` becomes `202610`. `doc/make_release.py --apply` wrote it and prepared the draft release for `v.202610`; merging this is step 1 of the four the script prints, and the tag must not be created before it lands, or the published package would carry the old version.

## Why the number

The milestone `v.202610` collects **everything merged since `v.202609`** — 80 pull requests, including this week's work on making published models loadable, the `espnet` command line, and the two OWSM demos. The in-flight work stays in `v.202612`, which keeps its 23 open items.

`master` said `202609.post1`, which is *older* than the `202609.post2` on PyPI: that patch release was cut from a branch off the `v.202609` tag with the bump on it, so master never saw the change. `202610` is ahead of both, and matches the month this release goes out — the same way `202204` and `202205` were a month apart.

## Checked

- `v.202610` has 0 open items and 80 closed, which is what the preflight requires.
- The other preflight checks pass: `id-token: write` present, no secret in the publish workflow, no direct reference in `pyproject.toml` — the three that silently kept espnet off PyPI between 202604 and 202609.
- Neither the tag `v.202610` nor that version on PyPI exists yet.

## After this merges

The remaining steps are the script's, in order: fill in the Overview on the draft, publish the draft (**this creates the tag and uploads to PyPI, and cannot be undone**), then check `gh run list --workflow publish_python_package.yml` — both 202604 tags failed there silently.


sw005320:
@coderabbitai full review

PR #6716, opened 2026-09-19
Say 202610 in What's new, and make the next release say its own

sw005320:
`README.md#whats-new` still described 202609 after 202610 shipped, and five pull requests were merged into 202610 with no milestone at all, so the notes never listed them. Both were manual steps that nothing pointed at.

**Now.** The 202610 entry is written from the published release notes, and 202609 moves into `Earlier releases`. (The five milestone-less pull requests, #6714, #6607, #6606, #6604 and #6594, have been assigned to v.202610.)

**Next time**, in `doc/make_release.py`:

- `--apply` writes the What's new entry with a placeholder summary, so the version bump pull request already carries it. Preflight fails while the entry is missing, and again while it still holds the placeholder.
- Preflight fails when merged pull requests since the previous tag have no milestone and were not opened by a bot; they would be missing from the notes.
- Preflight fails when no milestone exists for the release after this one, since that is where the next pull requests go.
- The printed hand-work list now starts with the README summary and ends with creating the following milestone.
- A README whose What's new section is not shaped as expected is reported, not rewritten.

Checked against the real repository: the current release passes the README check, the next one fails without `--apply`, `--apply` writes the entry and demotes 202610, and the milestone search found exactly the five pull requests above before they were fixed and reports clean now.

The automation that keeps the milestone filled in, `.github/workflows/assign_milestone.yml`, is #6717.


sw005320:
Fixed in the last commit. `readme_up_to_date` now validates the section structure first and reports it when the What's new list, the `Earlier releases` block or the `Full history:` line is not there, instead of passing on a bullet found anywhere in the file. The bullet is then looked for only in the current What's new list, so an entry already demoted into `Earlier releases` no longer reads as this release being announced.

Checked on a copy of the README with `Earlier releases` renamed: preflight reports the structure, and `--apply` writes nothing. Checked too that after the next release is added, 202610 stops counting as the announcement.

PR #6717, opened 2026-09-19
Give every human pull request the next release milestone

sw005320:
The release notes are generated from a milestone. A pull request without one is simply not in them, and assigning it was a manual step that everyone forgot, so the 202610 notes had to be reconciled by hand afterwards.

What this adds:

- **On open, reopen or ready-for-review**, the pull request gets the earliest open `v.YYYYMM` milestone, which is the next release. A milestone already set is left alone.
- **Bots are skipped.** A dependency bump or a pre-commit autoupdate is not something anyone writes a release note from.
- **A daily sweep** assigns the ones the event missed, so pull requests opened before this existed are covered too.

It follows `label_pull_request.yml`: `pull_request_target` so fork pull requests get a writable token, no checkout, nothing from the pull request is executed. The companion change to `doc/make_release.py`, which fails preflight when merged pull requests since the last tag have no milestone, is in #6716.

`ci/check_ci_image_config.py` passes.


sw005320:
Both addressed in the last commit.

The milestone is now re-read immediately before the write, for the event path and the sweep alike, so a milestone set between listing and patching is left alone rather than replaced by the default.

`workflow_dispatch` is gone. The daily sweep covers the same need, and dropping it means this job's write permissions cannot be run from a ref someone picks.

PR #6718, opened 2026-09-19
Two fixes the course notebook needed: hydra effects and S2T data augmentation

sw005320:
The ASR tutorial used in CMU 11492/11692/18495 installs espnet from `owaski/espnet -b ls_asr100_2_ctc`, because three things it needs live only there. We are moving those course notebooks into `espnet/notebook` and maintaining them, which means they have to install released espnet, so the three come here.

- **`Speech2TextCTCGreedySearch`** decodes with the CTC branch of an encoder-decoder S2T model such as OWSM v3.1: no decoder, no search, much faster, worse transcripts. That trade is right for a teaching example or a sanity check. `espnet2.bin.s2t_inference_ctc` stays the one for a model trained CTC-only, such as OWSM-CTC. The result carries `None` where a hypothesis would be, and the type alias now says so.
- **`DataAugmentation` accepts the `ListConfig` a hydra config arrives as.** Its elements are `ListConfig` too, so the `isinstance(tup[1], list)` test read every effect wrongly. The fork imported omegaconf at module level; that would put a training dependency on an inference path, which `ci/check_inference_imports.py` forbids, so the import is lazy and keyed on the type's module.
- **`S2TPreprocessor` forwards `data_aug_effects`, `data_aug_num` and `data_aug_prob`** to `CommonPreprocessor`, which has always taken them. Without this, an S2T config asking for speed perturbation trained without it and said nothing.

Tests: CTC greedy decoding returns text with no hypothesis and with blanks and repeats collapsed; `DataAugmentation` built from an `OmegaConf` list reads the probabilities correctly (verified that it fails without the fix); `S2TPreprocessor` ends up with the augmentation its arguments asked for.

`ci/check_inference_imports.py` passes: 27 inference modules import with a bare install.


sw005320:
Both addressed.

`batch_decode` is overridden: it pads and encodes the batch once, then collapses each utterance's CTC path through the same `_decode_single_sample`, so `__call__` and `batch_decode` cannot answer differently for one model. A test asserts they agree, utterance by utterance. The prompt arguments are accepted so a collated batch still passes straight through, and the class docstring says plainly that they are ignored: they condition the decoder, which this class never runs.

The augmentation test now carries only the nested effect. With two effects `DataAugmentation` picks one at random, so it could have passed against the broken version; it also asserts the converted structure rather than only that the call does not raise.

popcornell:
It seems to me that S2TCTCPreprocessor (OWSM-CTC, same file) has the identical problem ?
It inherits _speech_process but its `__init__ `doesn't accept or forward the three data_aug_* kwargs, so an OWSM-CTC config with augmentation fails with the same TypeError.
Maybe we just apply the same change there?

sw005320:
Good catch, and it was the same bug: `S2TCTCPreprocessor` inherits `_speech_process` from `CommonPreprocessor` exactly as `S2TPreprocessor` does, and its `__init__` neither accepted nor forwarded `data_aug_effects`, `data_aug_num` or `data_aug_prob`. An OWSM-CTC config asking for speed perturbation trained without it, the same way.

Fixed in the same shape, and the test is parametrised over both preprocessors now. I checked it the other way round too: removing either fix makes that preprocessor's case fail, so the test is testing the thing.

sw005320:
@popcornell, if it looks OK, can you approve this PR?

Emrys365 (approved):
LSTM

popcornell:
I checked  i think it is ok now

PR #6720, opened 2026-09-19
Add a top-level espnet package: `import espnet` and `espnet.load()`

sw005320:
## Why

`pip install espnet` installs `espnet2`, `espnet3` and `egs3` only. The distribution is
called `espnet`, so the first line a new user writes is `import espnet` — and since
ESPnet1 was removed and nothing took the name back, that raises `ModuleNotFoundError`.

This adds a small top-level `espnet` package with the two entry points that line is
reaching for.

## What

`espnet/__init__.py`:

- **`__version__`** — from `importlib.metadata.version("espnet")`, falling back to
  `"unknown (espnet is not installed; running from a source tree)"` rather than raising
  in a checkout that was never pip-installed.
- **`load(model_tag, task=None, device="cpu", **kwargs)`** — downloads a published model
  and returns the espnet2 inference object for its task:

  ```python
  import espnet
  asr = espnet.load("espnet/owsm_ctc_v4_1B")
  tts = espnet.load("espnet/kan-bayashi_ljspeech_vits")
  enh = espnet.load("espnet/some-tag", task="enh", device="cuda:0")
  ```

  `TASKS` maps `asr` → `Speech2Text`, `s2t` → `Speech2TextGreedySearch`, `tts` →
  `Text2Speech`, `enh` → `SeparateSpeech`, `spk` → `Speech2Embedding`, `diar` →
  `DiarizeSpeech`. Extra keyword arguments go straight to the class.

### How the task is inferred

An explicit `task=` always wins and skips the lookup entirely. Otherwise it is read from
the model's own Hugging Face metadata, because that is the only place that distinguishes
the four tasks whose packed `meta.yaml` all say plain `train_config` / `model_file`
(tts, enh, spk, diar):

1. The repository's `pipeline_tag` and its tags are collected together — older
   repositories set only the latter — and matched against an ordered label table, most
   specific first. Order is what resolves the overlap: espnet publishes speaker models as
   `audio-classification` with a `speaker-verification` / `speaker-recognition` tag beside
   them, so the speaker labels are checked first and bare `audio-classification` is
   deliberately *not* in the table rather than mapped to a guess.
2. ASR and S2T carry the same Hub labels, so those two are split by fetching just
   `meta.yaml` from the repository (one small file, not the checkpoint): an `s2t_`-prefixed
   artifact key means `Speech2TextGreedySearch`, otherwise `Speech2Text`. The two classes
   take differently named constructor arguments, which is exactly what `meta.yaml` records.
3. A tag whose metadata names no task — a local directory, a URL, a hand-uploaded
   repository — raises a `ValueError` that names the tag and lists the valid tasks.

Checked live against six published models: `kan-bayashi_csmsc_vits` → tts,
`voxcelebs12_rawnet3` → spk, `owsm_ctc_v4_1B` → s2t,
`Wangyou_Zhang_universal_train_enh_uses_refch0_2mem_raw` → enh,
`Hoon_Chung_jsut_asr_train_asr_conformer8_raw_char_sp_valid.acc.ave` → asr,
`diar_ami_eend` → diar.

### Import weight

The module body imports the standard library and nothing else. `torch` and `espnet2` are
imported inside `load()`, so `import espnet` stays cheap and no training package can reach
the inference path. A test asserts this in a fresh interpreter, and `espnet` is now in
`ci/check_inference_imports.py`'s list, which also fails loudly if the package ever stops
shipping in the wheel.

### Shared with the CLI rather than copied

Loading a tag, and explaining a tag that was published for another task, are the same job
`espnet <command>` already does. `espnet2/utils/pretrained.py` gains `build_pretrained()`
and `ModelTagError`; `cli._build` is now a wrapper that supplies the command line's own
wording, and `cli.main` catches `ModelTagError` alongside `CLIError` so it still prints one
line instead of a traceback. The existing CLI message is unchanged and its tests pass
untouched.

One difference between the two front ends matters: the CLI passes no constructor arguments
of its own, but `load()` forwards `**kwargs`. An unexpected-keyword `TypeError` is
therefore only translated into "this tag was published for another task" when the caller
did *not* pass that name — otherwise the real `TypeError` stands, so
`espnet.load(tag, task="tts", beam_size=1)` is never blamed on the model.

## Tests

`test/espnet2/bin/test_toplevel_api.py` — they test `espnet/__init__.py`, not
`espnet2/bin`, but that is where CI runs them: `ci/test_python_espnet2.sh` runs
`pytest test/espnet2` and `ci/test_python_espnet3.sh` runs `pytest test/espnet3/`, and
nothing runs a `test/espnet/` directory, so tests there would be collected by no job at
all. `espnet.load` is a front end onto the `espnet2.bin.*_inference` classes, which makes
`test/espnet2/bin` a fair home regardless.

Nothing is downloaded: each test substitutes the inference class that `load` imports.

- `__version__` is a non-empty string, and a checkout that was never pip-installed still
  gets one (the module is re-executed with `importlib.metadata.version` raising).
- Every task in `TASKS` dispatches to its own class, and every entry names a class that
  actually exists — imported outright rather than skipped when missing, since all six are
  entry points `ci/check_inference_imports.py` already requires to import on a bare
  install.
- `device` and extra keyword arguments reach the constructor.
- An explicit `task` wins, and asks the Hub nothing.
- An unknown task is refused with a `ValueError` naming it and every valid task.
- `_hub_labels` merges a repository's `pipeline_tag` with its tags, and a repository
  carrying neither is an empty set rather than `{None}`.
- Each Hub label maps to the right task, including the `audio-classification` +
  `speaker-*` pair.
- ASR vs S2T is split by the packed `meta.yaml` keys; an unfetchable `meta.yaml`, or one
  that is not a mapping, leaves the question open rather than guessing.
- A model the Hub says nothing about asks for the task by name.
- A tag published for another task raises `ModelTagError` naming the offending
  constructor argument — but an argument the *caller* invented is re-raised as the plain
  `TypeError` it is, rather than being blamed on the model.
- `import espnet` in a fresh interpreter pulls in neither `torch` nor any `espnet2` module.

Every executable line of `espnet/__init__.py`, and every line this PR adds to
`espnet2/utils/pretrained.py`, is reached in process.

## Packaging

`[tool.setuptools.packages.find] include = ["espnet*", "egs3*"]` already picks the new
package up — no `pyproject.toml` change was needed. Verified by discovery
(`espnet`, `espnet2`, `espnet3`, `egs3`) and by building the wheel: `espnet/__init__.py`
is in `espnet-202610-py3-none-any.whl`.

`setup.cfg` gains `--cov=espnet` beside the existing `--cov=espnet2 --cov=espnet3`, so the
new package is not invisible to coverage.


sw005320:
Both CodeRabbit findings were real and are fixed in 13cfb4b.

**`build_pretrained` blamed the model for the caller's own argument.** Translating every unexpected-keyword `TypeError` into "this tag was published for another task" was safe for `espnet <command>`, which passes no constructor arguments of its own — but `espnet.load` forwards `**kwargs`, so `espnet.load(tag, task="tts", beam_size=1)` claimed the model had been published with `beam_size`. `from_pretrained` merges the caller's keyword arguments with the downloaded ones and the `TypeError` does not say which set a name came from, so a name the caller passed is no longer translated at all: the real `TypeError` stands and names the key. I did not try to separate the overlapping case (caller and model both supply the key) by re-reading the downloaded keys — that would mean a second downloader round trip on an error path, and re-raising is never a false claim, where the translated message would be.

**`importorskip` in the registry test.** Agreed, replaced with a direct import. All six `TASKS` modules are inference entry points that `ci/check_inference_imports.py` already requires to import on a bare install, so a skip could only hide the typo the test exists to catch.

Added tests for the caller-argument case and for the two metadata paths that leave the task undecided (an unfetchable `meta.yaml`, and one that is not a mapping), which should also close the patch-coverage gap codecov flagged.

sw005320:
Followed up on `codecov/patch` in adf65f7.

It was partly an artifact and partly real. The artifact: codecov's own report said "HEAD has 105 uploads less than BASE", with `test_python_espnet2` at **0** uploads against 6 on base — that is the job that runs these tests, so `espnet2/bin/cli.py` was reading 0% too, and it is covered entirely by the existing `test_cli.py`.

The real part: tracing which lines of `espnet/__init__.py` the suite actually executes **in process** left two, and the coordinator's note is right that the subprocess test contributes nothing to coverage (it exists to assert that a fresh interpreter importing `espnet` pulls in neither `torch` nor any `espnet2` module, which cannot be observed from inside a process that has already imported both).

The two lines are now covered by tests that were worth having anyway:

- **the `__version__` fallback** — it only runs at import time, so the module is re-executed with `importlib.metadata.version` raising `PackageNotFoundError`, the way it does in a checkout that was never pip-installed.
- **`_hub_labels`' success path** — `model_info` is replaced with a stub rather than reached over the network. Every other test monkeypatches `_hub_labels` away entirely, so nothing checked that it merges a repository's `pipeline_tag` with its `tags`. That merge is what lets an espnet speaker model — published as `audio-classification` with a `speaker-verification` tag beside it — be recognised at all. A repository carrying neither is now asserted to be an empty set rather than `{None}`.

Every executable line of `espnet/__init__.py` is now reached in process, and every line this PR adds to `espnet2/utils/pretrained.py` is too (the remaining misses there are in `download_pretrained`, which this PR does not touch).

sw005320:
Merged master in (7a93ec7) — the PR had gone `CONFLICTING`.

Master's `espnet --version` landed in the same import block this branch edits. Resolved by keeping `importlib.metadata`, which master's new `_version()` needs, and dropping `inspect`, which nothing in `cli.py` uses any more now that the unexpected-keyword translation lives in `espnet2/utils/pretrained.py`.

I deliberately did **not** point `cli._version()` at the new `espnet.__version__`, even though they now compute the same string: `_version()` looks the version up on each call, which is exactly what its own tests monkeypatch, while `espnet.__version__` is fixed at import time. Both give the same "source tree" wording when the package is not installed.

The PR diff is unchanged at six files. `test_cli.py` (now including master's four new version tests) and `test_toplevel_api.py` pass together locally.

Status before the merge: `unit_test_espnet2_and_espnet3_on_debian12`, `test_import`, `codecov/patch` and `codecov/project` all green — `codecov/patch` passed as soon as the unit job's upload landed, confirming the earlier red was the missing-upload artifact described above.

PR #6721, opened 2026-09-19
Add a small inference docker image, and a table of which image is which

sw005320:
Since 202610 a bare `pip install espnet` is an inference install - 20 declared
dependencies and an `espnet` console script - so "try a published model" no
longer needs the development image, which carries Kaldi, the recipe tooling and
the training stack in order to run one file through one model. This adds the
small image that follows from that, and a table saying which of the three
images a reader actually wants.

### `docker/inference.dockerfile`

```sh
docker run --rm -v "$PWD:/data" \
    -v "$HOME/.cache/huggingface:/cache/huggingface" \
    espnet/espnet:inference-latest asr /data/audio.wav
```

`espnet` is the ENTRYPOINT, so every subcommand (`asr`, `translate`, `tts`,
`enhance`, `models`) is just an argument. Decisions worth the review:

- **`python:3.12-slim`, two stages.** The compiler lives in the builder, which
  nothing is copied out of but `/opt/venv`. Every inference dependency has a
  manylinux wheel today; the toolchain is there so a transitive sdist does not
  turn into a failed weekly publish.
- **torch from `https://download.pytorch.org/whl/cpu`, not PyPI.** The PyPI
  wheel pulls ~3 GB of `nvidia-*` packages that a CPU image can never use.
  Installing `+cpu` first leaves espnet's `torch>=2.11.0,...` already satisfied,
  so pip does not resolve back. `torchaudio` is pinned separately because its
  releases stop at 2.11.0.
- **No model weights.** One OWSM checkpoint is ~4 GB, the subcommand defaults
  are three different checkpoints, and `--model` takes any tag from
  https://huggingface.co/espnet. Models download on first use into
  `HF_HOME=/root/.cache/huggingface`; the cache mount is documented in the
  dockerfile, in `docker/README.md` and in the top-level README, because without
  it every run downloads the model again.
- **`WORKDIR /data`**, so a relative path means the directory the user mounted.
- **Everything writable is under `/cache`** (`HOME`, `HF_HOME`, and numba's
  cache, since librosa's JIT functions are `@jit(cache=True)`), world-writable
  and sticky. That is so `--user "$(id -u):$(id -g)"` works: without it a Linux
  host gets root-owned transcripts and wavs in the mounted directory, which is
  CodeRabbit's one finding on this PR. An arbitrary uid has no home directory
  here and cannot write under `/root`, so the caches are not under it. The
  documented mount is therefore `-v "$HOME/.cache/huggingface:/cache/huggingface"`.

Expected size: roughly **1.0-1.2 GB uncompressed, ~450 MB to pull**. Estimated,
not measured: the torch `+cpu` cp312 manylinux wheel is 190 MB by
`Content-Length`, the other large wheels come to ~155 MB from PyPI metadata
(llvmlite 60, scipy 35, numpy 17, pandas 11, sklearn 9, numba 4, the rest
small), a wheel expands about 2.2x on install, and `python:3.12-slim` is about
125 MB. For comparison, `espnet/espnet:cpu-latest` is tens of GB.

### Publish workflow

`espnet/espnet:inference-latest` is a third matrix row. The two development rows
differed only in a couple of build args, so rather than give the build step a
conditional, each row now carries its own `dockerfile` and its own `build_flags`
and the step is one `docker build`. Action pins, `permissions:` and
`persist-credentials: false` are untouched. The inference row passes
`--build-arg ESPNET_VERSION="$(cat ../version.txt)"`, so the published image
cannot lag a release even if the dockerfile's own default does.

`python3 ci/check_ci_image_config.py` passes (all fourteen invariants, including
the version-pin scan, which now also sees `TH_VERSION=2.11.0` in the new
dockerfile).

### Docs

- `docker/README.md`: a table of the three images - development cpu/gpu,
  inference, and `espnet/vllm:latest` for serving SpeechLM models - with what
  each is for and the one command each, then the inference image's mounts and
  local build. The vLLM row links #6695, which documents that path;
  `egs2/bagpiper/speechlm1/README.md` is left alone, as it belongs to that PR.
  A cross-link from the Bagpiper page back to this table can follow once #6695
  merges.
- Top-level README: the one `docker run` next to the existing `pip install
  espnet` quick start, four lines, since that section is already the
  "no code, on any audio file" section.
- `doc/docker.md`: one paragraph at the top saying that page is about the
  development images, and where the others are listed.

### What was not verified

**The image was not built.** Docker is not installed on the machine this was
written on, and `hadolint` was not available either, so the dockerfile was
reviewed line by line against `docker/espnet.dockerfile` and
`docker/prebuilt/devel.dockerfile` instead. The first real build will be the
publish workflow, which has a `workflow_dispatch` trigger, so it can be run
without waiting for Monday.

Verified without a build:
- package name, console script and entry point read from `pyproject.toml`
  (`[project.scripts] espnet = "espnet2.bin.cli:main"`) and
  `espnet2/bin/cli.py`; the subcommand list is that parser's own `--help`.
- `espnet==202610` exists on PyPI and matches `version.txt`; its
  `requires-python` is `>=3.12,<3.14`, so the `python:3.12-slim` base is inside
  it.
- `torch-2.11.0+cpu-cp312-cp312-manylinux_2_28_x86_64.whl` and the matching
  torchaudio wheel exist on the PyTorch CPU index (and in aarch64, for anyone
  building on arm).
- the cache path: `espnet_model_zoo`'s default `ModelDownloader` passes
  `cache_dir=None` to `snapshot_download`, so a Hub tag lands under `HF_HOME` -
  which is why that is the mount documented. A tag published outside the Hub
  goes to `~/.cache/espnet_model_zoo` instead, and `docker/README.md` says so.

Nor was `--user "$(id -u):$(id -g)"` exercised; it is reasoned from the
permissions the image sets, not observed.

Still unknown until a build runs: the real image size, whether any transitive
dependency arrives as an sdist (the builder stage is there for that), and
whether `soundfile`'s and `sentencepiece`'s bundled native libraries are enough
on this base, as expected, or whether the final stage needs `libsndfile1`.


sw005320:
Addressed the review finding in b1c7c0626aa.

The container did run as root with both caches under `/root/.cache`, so on a Linux host the wav or transcript it left in the mounted directory was root-owned, and `--user "$(id -u):$(id -g)"` would not have worked either: an arbitrary uid has no home directory in this image and cannot write under `/root`.

Everything the process writes now lives under one world-writable, sticky `/cache` — `HF_HOME` for a Hub tag, `HOME` because `espnet_model_zoo` caches a tag published outside the Hub under `~/.cache/espnet_model_zoo`, and `NUMBA_CACHE_DIR` because librosa's JIT functions are `@jit(cache=True)` and would otherwise try to write beside the installed package. So `--user` now works with no other change, and the documented mount is `-v "$HOME/.cache/huggingface:/cache/huggingface"`; `docker/README.md` and the top-level README say to add `--user` on Linux.

Default is still root, so the plain command works as written on Docker Desktop, where file sharing maps ownership anyway.

PR #6722, opened 2026-09-19
`espnet demo`: the OWSM browser demo as one command

sw005320:
## What

```sh
pip install "espnet[demo]"
espnet demo                     # http://127.0.0.1:7860
espnet demo --model espnet/owsm_ctc_v4_1B --device cuda --port 8000 --share
```

`espnet demo` serves the OWSM app in a browser from the espnet install the
user already has: microphone or upload, the spoken-language menu and the
translation targets read from the checkpoint, long-form decoding, and the
device rule the Spaces use (CUDA when torch sees one, CPU otherwise; `--device`
overrides). It downloads the same `espnet/owsm_ctc_v4_1B` that `espnet asr`
uses, prints the URL, and blocks until Ctrl-C.

Until now, seeing a published model in a browser meant cloning the repo,
finding `egs2/owsm_ctc_v4/s2t1/demo`, installing its requirements and running
`app.py`.

## gradio is not a core dependency

`pip install espnet` must stay enough to load a model, so gradio lives in a new
`[demo]` extra (the `[mcp]` pattern; kept out of `[all]` for the reason `[sds]`
is). `espnet2/bin/demo.py` imports it inside the function that builds the app,
and a missing gradio reads like the CLI's other user errors:

```
$ espnet demo
espnet: gradio is not installed: pip install 'espnet[demo]'
$ echo $?
1
```

`espnet2.bin.demo` is now in `ci/check_inference_imports.py`, so a module-level
gradio import fails CI rather than a user's bare install, and a unit test
asserts the same invariant in a subprocess.

## Shared with the Spaces, but not yet imported by them

`espnet2/bin/demo.py` holds what both Space apps already duplicated: the ISO
639-3 table, the menus a checkpoint describes, reading and padding audio, and
splitting OWSM's leading symbols off its text.

**The two Space apps are untouched.** An earlier revision had them import the
module and raised their `requirements.txt` floor to the release that will carry
it; that puts a version which does not exist yet on a live demo's critical
path. Both Spaces run today and are re-uploaded from this repository by hand
against whatever espnet PyPI serves, so any upload before that release would
fail to resolve and the Space would die. They switch to importing the module
once a release carries it.

What holds the three copies together until then is
`test/espnet2/bin/test_demo_apps.py`, which already fails when the two apps
drift from each other and now also fails when either drifts from
`espnet2.bin.demo`. Importing `app.py` would download a 1B checkpoint, so the
test lifts the constants and helpers out of each app's syntax tree and runs
them on their own, then compares: the six shared constants, both menus built
from a token list, the padding, and the CTC app's symbol splitting.

The shared module pads with `np.pad` where the apps call
`librosa.util.fix_length` - the same zero padding without reaching through
`librosa.util` (and `scipy.ndimage`) for one line, and the drift test asserts
the two produce the same array.

Two deliberate differences from the hosted Spaces: `espnet demo` trims audio
past the two-minute cap and warns, where a Space refuses it outright (locally
there is no ZeroGPU deadline to miss), and it offers no remote example file, so
building the page needs no network.

## Tests

`test/espnet2/bin/test_cli.py` (no model is downloaded; gradio and the
inference class are monkeypatched): `demo` is registered and in `--help`, the
default tag and device reach `from_pretrained`, `--model/--port/--share` reach
`launch`, the device rule is used when `--device` is absent and overridden when
it is not, a missing gradio is a one-line error and exit 1 rather than a
traceback, the dropdowns are built from the checkpoint's tokens, a short
recording is padded to one 30 s window, and audio past the cap is trimmed with
a `gr.Warning` and decoded long-form in the language the user picked.

`test/espnet2/bin/test_demo.py` covers the shared helpers on a token list:
menus, `<unk>`/`<asr>` boundaries, unnamed language codes, symbol splitting,
padding, and that importing the module does not import gradio.
`test/espnet2/bin/test_demo_apps.py` gains the drift comparisons described
above.

`black`, `isort`, `flake8` clean; the three files are 69 passed, 2 skipped
(the padding comparison needs a `scipy.ndimage` that this machine's numpy
cannot load; it runs in CI).

## Not verified here

The page itself was checked by building the real Gradio app (6.28.0) with a
stub model, serving it and fetching it over HTTP - it renders and the Run
handler returns. It has **not** been run against a real checkpoint on this
machine: no model was downloaded, so microphone input, long-form decoding on a
GPU and `--share` are unexercised. The two Space apps are byte-identical to master, so nothing about them
changed; they were not launched (no gradio in the local environment).


sw005320:
Reworked so the live Spaces are not on this PR's critical path.

`egs2/owsm_ctc_v4/s2t1/demo` and `egs2/owsm_v4/s2t1/demo` are back to exactly what master has - both `app.py`, both `requirements.txt` and both `README.md`; `git diff master -- egs2/` is empty. The earlier revision had them import `espnet2.bin.demo` and raised their espnet floor to the release that will carry it, which would have broken either Space the next time it was uploaded, since that release does not exist yet.

`espnet2/bin/demo.py` and the `espnet demo` command are unchanged from the first revision. What keeps the three copies of the shared pieces honest is `test/espnet2/bin/test_demo_apps.py`, which already failed when the two apps drifted from each other and now also fails when either drifts from the module. Importing `app.py` would download a 1B checkpoint, so the test lifts the relevant definitions out of each app's syntax tree and executes them on their own, standing in for the loaded model, then compares:

- the six constants both apps and the module define (`SAMPLE_RATE`, `WINDOW_SECS`, `MAX_SECS`, `DETECT`, `ASR_LABEL`, `LANGUAGE_NAMES`);
- both dropdowns, built from the same token list;
- the padding - `librosa.util.fix_length` in the apps against `np.pad` in the module, asserted to give the same array;
- the CTC app's splitting of OWSM's leading symbols.

The apps switch to importing the module once a release carries it; the PR body now says so.

`black`, `isort` and `flake8` are clean, and `test_cli.py`, `test_demo.py` and `test_demo_apps.py` are 69 passed, 2 skipped locally - the two skips are the padding comparison, which needs a `scipy.ndimage` this machine's numpy cannot load, and which runs in CI.

sw005320:
Thanks @coderabbitai - two of the three applied in 99f84d8.

**`load_gradio` caught every `ImportError`** — agreed, and it was worse than hiding the cause: a gradio that is installed but fails on one of its own dependencies was reported as "gradio is not installed", pointing the user at `pip install espnet[demo]`, which would not fix it. Now only gradio's own `ModuleNotFoundError` answers `None` and anything else is re-raised, with a test for each half.

**`importorskip("scipy.ndimage")` in the padding comparison** — agreed, removed. scipy is a core dependency, so an install where `librosa.util` cannot import it is broken rather than a case to skip. (It does mean the two padding tests fail on my machine, whose conda scipy is built against numpy 1.x while numpy 2.2.6 is installed — `python -c "import scipy.ndimage"` fails there on master too. Equivalence was checked separately against `librosa.util.fix_length`'s implementation for short, exact-length and over-length input.)

**Bounding `gradio` in the `[demo]` extra** — not doing this one. The two `gradio` declarations already in `pyproject.toml`, in `[sds]` and `[test]`, are both unbounded, so a bound here would be the odd one out; and the only versions this app has run on are 6.27.0 (what the Spaces pin) and 6.28.0 (a local smoke test that built the real app, served it and fetched the page). Writing a range around one tested version is the "needlessly narrow" half of the same policy. A floor is worth adding the day a gradio release is known to break it.

sw005320:
> Is `espnet demo` intended to become a generic model demo interface eventually,
> or is it intentionally an OWSM-specific command for now?

The former, and the name was chosen for that. What exists today is one app —
the OWSM one — and the intent is that `espnet demo --model <tag>` grows to pick
the app that fits the tag, the way `espnet asr`/`tts`/`enhance` already pick the
inference class. Making it `espnet owsm-demo` now would mean renaming it the
first time a second app arrives, and a deprecation for a command a few weeks
old.

What I did not want to do in this PR is build the dispatch layer before there is
a second thing to dispatch to. So the split is: the CLI stays thin, the
OWSM-specific work is all in `espnet2/bin/demo.py` (whose docstring says so in
its first line), and a second app arrives as a second module plus the dispatch —
not as a rewrite of this one.

> it would be good to validate that the selected checkpoint is supported by this
> demo and give a clear error otherwise

Done, and it was worse than an unclear error: the failure came from inside the
constructor **after** the 4 GB download. The task is now settled first, from the
model's own Hub metadata — exactly your model → identify task → demo chain,
reusing `espnet._infer_task`, the same resolver `espnet.load()` uses:

```
$ espnet demo --model espnet/kan-bayashi_ljspeech_vits
`espnet demo` serves speech-to-text models, and espnet/kan-bayashi_ljspeech_vits is a tts model. Pass --model with an OWSM tag; `espnet models` names the default.
```

One deliberate hole: a model whose metadata says nothing about its task is let
through rather than refused. The check is there to turn a knowable mistake into
a sentence, not to become a second gate a valid checkpoint has to pass. Both
behaviours are tested, and the refusal test asserts that nothing was fetched.

This is also where the model capability metadata you mention would plug in: the
resolver reads `pipeline_tag`, the Hub tags, and falls back to the `meta.yaml`
inside the repository. Today it answers "which task"; a demo registry would ask
it the same question and get more of an answer.

> `MAX_SECS = 120` — OWSM-specific or general?

General, and the two are now named apart in the source. `WINDOW_SECS = 30` is
the model's: what OWSM was trained on, and a checkpoint trained on another
length would change it. `MAX_SECS = 120` is the demo's: two minutes of audio is
already a wait in a browser, it would be the same number for any model this
command grew to serve, and neither the model nor `espnet asr` has such a limit.

The conflict with master is resolved and the branch is up to date.

PR #6723, opened 2026-09-19
Add TTS, speech enhancement and speaker verification demo Spaces

sw005320:
Three more demo Spaces, built to the shape of the two OWSM ones.

The repository maintains `egs2/owsm_ctc_v4/s2t1/demo` and
`egs2/owsm_v4/s2t1/demo`. Three of ESPnet's headline tasks had no demo anyone
could click — the Spaces the README used to link for them had been erroring
since 2021-2022 — so this adds one each, in the same three-file shape
(`app.py`, a card, `requirements.txt`), with the same ZeroGPU shim before
torch, the same `SPACES_ZERO_GPU` / `DEVICE` rule, and the same
`@spaces.GPU(duration=...)` on the decorated call.

| directory | model | what it does |
|---|---|---|
| [`egs2/ljspeech/tts1/demo`](egs2/ljspeech/tts1/demo) | [`kan-bayashi_ljspeech_vits`](https://huggingface.co/espnet/kan-bayashi_ljspeech_vits) | text in, waveform out, with a speed control; the default model of `espnet tts` |
| [`egs2/universal_se_v1/enh1/demo`](egs2/universal_se_v1/enh1/demo) | [`Wangyou_Zhang_universal_train_enh_uses_refch0_2mem_raw`](https://huggingface.co/espnet/Wangyou_Zhang_universal_train_enh_uses_refch0_2mem_raw) | noise and reverberation out of one or many channels at any rate; the default model of `espnet enhance` |
| [`egs2/voxceleb/spk1/demo`](egs2/voxceleb/spk1/demo) | [`voxcelebs12_rawnet3`](https://huggingface.co/espnet/voxcelebs12_rawnet3) | two recordings in, a cosine and a verdict out |

## Read from the checkpoint, not written down

Following the OWSM demos, where the language menu and the translation targets
come out of the token list:

* **TTS** — the speaker menu appears only when the model takes speaker IDs and
  the language menu only when it takes language IDs (both hidden for
  single-speaker English LJSpeech); the speed slider appears only when the
  model's own decoding config has `alpha`, which VITS, FastSpeech and
  FastSpeech2 have and Tacotron 2 does not. A checkpoint that needs an
  x-vector or a reference recording is refused at startup with the reason,
  rather than failing inside the first request.
* **Enhancement** — the "Process at" menu is the set of sample rates in the
  checkpoint's training categories (`1ch_48k`, `2ch_16k`, …), the channel
  warning fires past the largest channel count those categories name, and
  there is one output player per `num_spk`, so a separation checkpoint reached
  through `ENH_MODEL_TAG` grows the column instead of dropping speakers.
* **Speaker verification** — the sample rate and the length the model was
  trained to score come from its preprocessor config, which is what makes a
  shorter clip worth a warning.

Each caps its input — 500 characters, 30 s, 30 s — and raises a `gr.Warning`
saying what was cut, rather than losing the tail silently.

## Where the verification threshold comes from

`egs2/voxceleb/spk1/README.md` reports, for this checkpoint on VoxCeleb1-O,
target trials at -0.8091 ± 0.1398 with an EER of 0.978%. The recipe scores a
trial as minus the Euclidean distance between L2-normalised embeddings, so a
score `s` is a cosine of `1 - s²/2`; reading the target scores as normal, the
equal error point sits 2.33 standard deviations below their mean, a distance
of 1.135, a cosine of 0.36. That is the slider's starting value. It is one
operating point on one corpus of read English speech, so it is a slider and
not a constant, and the card says so.

## Verified locally, on CPU

Each app was imported with `DEVICE=cpu` and its `predict` run:

* TTS: loads in 3.3 s, says "The quick brown fox…" in 3.3 s of audio at
  22.05 kHz; speed 0.75 gives 4.3 s of audio; 1499 characters of input warn
  and synthesise the first 500.
* Enhancement: mono, stereo and forced-48 kHz all run; on
  `test_utils/ctc_align_test.wav` plus Gaussian noise it lifts SI-SDR from
  -7.5 dB to +7.7 dB.
* Speaker verification: the two halves of `test_utils/ctc_align_test.wav`
  score 0.683 ("Same speaker"), and `ctc_align_test.wav` against
  `st_test.wav` scores -0.063 ("Different speakers"); a 39 s input warns and
  uses its first 30 s.

`test/espnet2/bin/test_demo_apps.py` passes (38 tests), and
`ci/check_demo_links.py` still exits 0 with these files staged.

## The README's demo list is left for a follow-up

`ci/check_demo_links.py` asks the Hub for the runtime stage of every linked
Space. These three do not exist yet, and the Hub answers HTTP 401 for them —
which the checker raises as a `ScanError`, so the daily run would exit 2 and
report a failed scan, not merely a broken link. Adding the links before the
upload would therefore break the checker every day until the upload happened.

Once the three Spaces are running, three lines go next to the OWSM ones in
`README.md`:

```markdown
- Hosted demo: [LJSpeech VITS](https://huggingface.co/spaces/espnet/ljspeech-vits), maintained from [`egs2/ljspeech/tts1/demo`](egs2/ljspeech/tts1/demo) — English text to speech with an end-to-end VITS model
- Hosted demo: [Universal speech enhancement](https://huggingface.co/spaces/espnet/universal-se), maintained from [`egs2/universal_se_v1/enh1/demo`](egs2/universal_se_v1/enh1/demo) — one model for noise, reverberation, one microphone or several, and any sample rate
- Hosted demo: [Speaker verification](https://huggingface.co/spaces/espnet/speaker-verification), maintained from [`egs2/voxceleb/spk1/demo`](egs2/voxceleb/spk1/demo) — two recordings, a cosine similarity and a verdict
```

## Creating and uploading the Spaces

Nothing here was uploaded; that is a manual step. Each Space wants ZeroGPU
hardware (`zero-a10g`), which is a setting on the Space rather than a file:

```sh
hf auth login
hf repo create espnet/ljspeech-vits --repo-type space --space_sdk gradio
hf repo create espnet/universal-se --repo-type space --space_sdk gradio
hf repo create espnet/speaker-verification --repo-type space --space_sdk gradio
# then set each to ZeroGPU (zero-a10g) in its Settings, and from this checkout:
hf upload espnet/ljspeech-vits egs2/ljspeech/tts1/demo . --repo-type space
hf upload espnet/universal-se egs2/universal_se_v1/enh1/demo . --repo-type space
hf upload espnet/speaker-verification egs2/voxceleb/spk1/demo . --repo-type space
```

The Space names are the only thing in this PR that is a choice rather than a
fact: they appear in each `app.py` docstring and each card, so if you would
rather call them something else, say so and I will change all three.


sw005320:
Thanks — three of the four are in, in e8388221e3b.

**Bounded decoding in both audio demos.** `librosa.load` was reading the whole upload before the slice threw most of it away. Both now ask the file header for the duration, warn on that, and decode with `duration=MAX_SECS`, so a 39 s file comes back as exactly 30 s without the rest ever being resampled. Verified locally on CPU: the enhancement demo still handles mono, stereo and forced-48 kHz, and the speaker demo still scores -0.063 across the two `test_utils` speakers.

**The cap test now checks that the cap is applied.** Right that it passed for an app that declares a limit, warns about it and never enforces it — the more misleading of the two failures. It now also requires the input to be measured against the cap in a comparison, and the message check reads the call's arguments rather than the whole call. Mutation-checked: replacing the comparison with `if True:` fails the test.

**Splitting the file into five is the one I'd rather not do.** This test only exists because the five demos are *not* one source file each: a Space is one directory, so the shared parts are copied between apps rather than imported, and the tests compare the copies against each other. `test_the_shared_half_of_the_two_owsm_apps_is_identical` and `test_every_app_keeps_the_same_device_rule` have no meaning inside a single-app module — the device rule getting out of step is silent, the app just runs on the CPU of a machine rented for its GPU. The file predates this PR with the same cross-file shape for two demos; this extends it to five rather than changing the idea. Happy to revisit if a maintainer reads the `test/**` instruction as binding here.

sw005320:
Pointed the SE Space at `kohei0209/tfgridnet_urgent25`, so it runs the model
`espnet enhance` loads after #6733 and the sentence "the model `espnet enhance`
uses by default" stays true.

One thing that needed fixing rather than swapping: the rate menu is read off the
checkpoint's training categories, and this checkpoint spells them `1ch_16000Hz`
where USES spells them `1ch_16k`. The menu came up empty. `_rates()` now reads
both spellings, so either checkpoint fills it — verified on both category lists.

The directory still sits under the USES recipe. The URGENT 2025 recipe that
trained the new default is not upstream (the model card says so too), so there
is nowhere better for it yet; `ENH_MODEL_TAG` still runs the app on USES, and
the README says both things. When that recipe lands, the demo moves next to it.

sw005320:
Reverted the model swap — we are unifying the three front ends on USES instead, so this Space stays on the model it was written around and this PR stays about publishing three Spaces. The branch is back to what it was before that commit.

sw005320:
> are these demo apps intended to work with other compatible checkpoints by
> changing only the model tag?

Yes — that is what `TTS_MODEL_TAG`, `ENH_MODEL_TAG` and `SPK_MODEL_TAG` are for,
and your review found the two places where it was still not true. Both fixed.

**The enhancement app's rate menu.** It reads the training categories off the
checkpoint, and matched only the `1ch_16k` spelling. A recipe that writes
`1ch_16000Hz` instead — the URGENT 2025 one does — got an empty "Process at"
menu. It now reads both, checked against both category lists:

```
USES      -> [48000, 16000]
URGENT25  -> [48000, 44100, 32000, 24000, 22050, 16000, 8000]
```

**The speaker app's threshold.** It was a constant labelled "this model's equal
error point on VoxCeleb1-O", which stops being true the moment `SPK_MODEL_TAG`
points elsewhere — a wrong number *and* a wrong sentence. Unlike the rest, this
one cannot be derived: a threshold is a choice about which errors to make, and
no checkpoint records one. So it is `SPK_THRESHOLD`, and the slider's own label
now says which of the two things it is showing — the default model's measured
operating point, or a starting value for a checkpoint whose own point you have
to find.

What is already read from the checkpoint, for the record: the TTS speaker and
language menus (`use_sids`/`use_lids`, `tts.spks`/`tts.langs`), whether the speed
slider exists at all, and `tts.fs`; the enhancement rates, channel ceiling and
one output player per `num_spk`; the speaker model's sample rate and the crop
length it was trained to score, which is what makes a short recording worth a
warning.

What is deliberately not derived is the prose. Each app's title and description
describe the checkpoint it is published with — "ECAPA-TDNN … 0.86% EER on
VoxCeleb1-O" — because the Space is that model's demo, and a generic sentence
would be worse for every visitor in order to be correct for a developer who set
an environment variable. The README of each directory says the tag is
overridable.

> I don't think the proposed split of test_demo_apps.py into five files is
> necessary

Agreed, and not done. Those tests exist precisely to check the cross-demo
invariants — the ZeroGPU shim and the device rule are compared across all five
apps, and the two OWSM apps against each other. Splitting them per demo would
delete the thing they are for.

sw005320:
The speaker demo now verifies with
[`voxcelebs12_rawnet3`](https://huggingface.co/espnet/voxcelebs12_rawnet3) —
0.739% equal error rate on VoxCeleb1-O against ECAPA-mel's 0.856%.

It is the best of the recipe's models a `pip install` can load, which is the
constraint that decided it. The three that score better —
`ecapa_frozen` (0.638%) and the two that fine-tune WavLM — reach their front
end through S3PRL, and S3PRL's released version calls
`torchaudio.set_audio_backend`, removed in torchaudio 2.1 where espnet pins
2.11. I installed s3prl and watched the load fail:

```
AttributeError: module 'torchaudio' has no attribute 'set_audio_backend'
```

The call is still on s3prl's `main`, so installing from git would not help
either. A Space installs from `requirements.txt` and nothing else, so those
checkpoints would leave this Space failing to start.

**The threshold loses its claim with the model it belonged to.** 0.36 was
ECAPA-mel's equal error point; the recipe table records each model's equal
error *rate*, never the score at which it occurs, so there is nothing to
inherit. The slider now says what it is — a place to start — for every
checkpoint, and `SPK_THRESHOLD` sets it for anyone who has measured one. That
also removes the `DEFAULT_TAG` branch, which existed only to tell the default
apart from a swapped-in checkpoint.

Checked on the app itself: two 3 s crops of one speaker score **0.700**,
against **0.198** for a different recording, in 0.45 s on a CPU.

The conflict with master is also resolved. #6722 rewrote `test_demo_apps.py`
to compare the two OWSM apps against `espnet2/bin/demo.py` while this branch
generalised the same file to five demos; both are kept, with the
`espnet2.bin.demo` comparisons scoped to the two OWSM apps, since the other
three share no decoding with it. 55 tests pass.

PR #6725, opened 2026-09-19
Transcribe as the audio arrives: `espnet asr --live` and `--stream`

sw005320:
`espnet asr file.wav` waits for the whole file and prints once. That is the wrong shape for a demonstration, and impossible for a microphone, so:

```sh
espnet asr --live            # the default input device, until Ctrl-C
espnet asr long.wav --stream # a file, one window at a time
```

Both print each window as it is decoded, through one loop in `espnet2/bin/live.py`; only the source differs.

**sounddevice stays out of the dependencies.** It needs PortAudio, which is a system package, and nothing else in espnet records audio. `--live` imports it when asked and, when it is missing, says what to install on this platform instead of raising `ImportError`. `ci/check_inference_imports.py` still passes: 27 inference modules import with a bare install.

**What this trades.** The windows are decoded independently, so a word straddling a boundary can come out twice or not at all. Plain `espnet asr` reads the whole file and remains the accurate path. The module docstring says this, rather than leaving someone to find it in a transcript.

**Verified.** `espnet asr audio.wav --stream` against `espnet/owsm_ctc_v4_1B` on CPU prints "The sale of the hotels is part of holiday strategy to sell off assets and concentrate on property management." in 47 seconds. Ten tests cover the window regrouping (including the short last window, which would otherwise drop the end of a recording), channel mixing, resampling, the empty transcript, Ctrl-C ending a recording with 130 rather than a traceback, the missing-sounddevice message, and the microphone loop against a stub device. The microphone itself was not recorded from.

This touches `espnet2/bin/cli.py`, as does #6722; whichever lands second needs a trivial rebase in `build_parser`.


sw005320:
Both are right, and both are now fixed. You have named the thing exactly: this
is not a leak, it is audio held without a bound — memory that grows with time
rather than with anything the user chose.

## 1. The microphone queue

`queue.Queue()` → `BoundedBlocks`, holding 30 s and dropping the **oldest** block
when full.

The policy had to be decided rather than defaulted, so it is stated in the
class: a live transcript is worth having *because* it is live, and a gap in it
beats one that is minutes behind. Dropping the newest would keep the transcript
complete and let it fall behind forever, which is the same failure in slower
motion. The callback never blocks and never raises — a callback that waits on
the consumer stalls the device itself.

Every drop is counted and said out loud: once at the first drop, then once per
10 s of dropped audio, and a total when the recording ends, naming a smaller
model or `--device cuda` as the fix. Silently losing speech is not something a
transcript may do.

```
decoding is slower than the audio arrives: dropped 10 s of audio so far
...
23 s of audio was dropped: the decoder could not keep up. A smaller model or --device cuda would have kept it.
```

## 2. `--stream` read the whole file

Also right, and worse than it looked: `sf.read` was followed by
`librosa.resample`, so both the source and the resampled copy were resident.
Measured on 30 minutes of 8 kHz audio, before a single window was decoded:

| | peak RSS |
|---|---|
| before (`sf.read` + `librosa.resample`) | **382 MB** |
| after (block by block) | **0.8 MB** |

It now reads through `sf.SoundFile` one block at a time, mixes to mono and
resamples per block. Per-block resampling with `librosa.resample` would put a
discontinuity at every boundary, so this uses `soxr.ResampleStream`, which
carries its filter state across blocks — the output is the signal the one-shot
call would have produced. That is a test rather than a claim: the streamed
result is correlated > 0.99 with `librosa.resample` of the whole file, at the
same length. soxr is not a new dependency; librosa requires it.

## Tests

Four new ones, none needing PortAudio or a model:

- the queue never exceeds its ceiling, and what survives is the *newest* audio
- the first drop is reported immediately, then sparingly (a line per block would
  be noise)
- `from_file` fails if anything calls `soundfile.read` — reading the file whole
  is the one thing it must not do
- block-wise resampling matches whole-file resampling

`espnet asr audio.wav --stream` still transcribes correctly on
`espnet/owsm_ctc_v3.2_ft_1B`.

PR #6522, opened 2026-08-22
[espnet3-recipe] Add VCTK noisy enhancement recipe

Dahee96:
## What did you change?

Added an ESPnet3 speech enhancement recipe using the VCTK-Noisy dataset.

---

## Why did you make this change?

To add a speech enhancement recipe to ESPnet3.

---

## Is your PR small enough?

yes

---

## Summary by CodeRabbit

* **New Features**
  * Added ESPnet3 speech enhancement support for training, inference, and model packaging.
  * Added a VCTK-Noisy enhancement recipe with Conv-TasNet configuration and end-to-end pipeline commands.
  * Added a Gradio demo for running enhancement models and generating audio outputs.
  * Added SI-SNR, PESQ, and STOI/ESTOI evaluation metrics.

* **Documentation**
  * Added setup, usage, dataset, evaluation, model, and Hugging Face documentation for enhancement recipes.

PR #6726, opened 2026-09-19
fix(espnet3): unblock VCTK enhancement CI with EnhancementTask docstrings

modelpath-dev:
## Summary
- Continues [#6522](https://github.com/espnet/espnet/pull/6522) (VCTK-Noisy ESPnet3 enhancement recipe) from Dahee’s PR head after the recent master merge.
- Adds brief package / class / method docstrings on `espnet3/systems/enh` so `test_python_espnet3` flake8-docstrings passes, matching the existing ASR task style.
- No behavior changes; recipe content is unchanged aside from the docstring CI fix.

## Test plan
- [x] `python3 -m flake8 espnet3/systems/enh/` (exit 0, no D10x)
- [ ] Confirm `test_python_espnet3` is green on this PR
- [ ] Keep [#6522](https://github.com/espnet/espnet/pull/6522) in mind for close/supersede once this lands, or cherry-pick the docstring commit there if preferred

Made with [Cursor](https://cursor.com)


egs3/mini_an4/enh/dataset/builder.py
```diff
@@ -0,0 +1,235 @@
...
+        if not archive.exists():
+            raise FileNotFoundError(f"Archive not found: {archive}")
+        dataset_root.mkdir(parents=True, exist_ok=True)
+        _normalize_downloads_layout(dataset_root)
+        if self.is_source_prepared(recipe_dir=recipe_dir):
+            return
+        subprocess.run(
+            ["tar", "-xzf", str(archive), "-C", str(dataset_root)],
+            check=True,
+        )
```
Emrys365: You can use the off-the-shelf Python utility `extract_targz` instead:

```suggestion
from espnet3.utils.download_utils import extract_targz

...

        extract_targz(archive, dataset_root)
```

egs3/mini_an4/enh/src/inference.py
```diff
@@ -0,0 +1,116 @@
...
+            padded = np.zeros((len(speech_mix), max_len), dtype=np.float32)
+            for i, s in enumerate(speech_mix):
+                padded[i, : len(s)] = s
+            speech_mix = padded
+        else:
+            # Single sample: (T,) -> (1, T)
+            if speech_mix.ndim == 1:
+                speech_mix = speech_mix[np.newaxis, :]
+
+        return self._model(speech_mix)
```
Emrys365: The padded batch sequence should be reverted to the original lengths (as a list) for real applications.

egs3/TEMPLATE/esp2_enh/conf/demo.yaml
Emrys365, line 1: It is better to include the detailed comments in the TEMPLATE recipe, so that
+ contributors can later refer to this for creating new recipes
+ users can refer to this for understanding the usage

egs3/TEMPLATE/esp2_enh/conf/inference.yaml
Emrys365, line 1: ditto

egs3/TEMPLATE/esp2_enh/conf/metrics.yaml
Emrys365, line 1: ditto

egs3/TEMPLATE/esp2_enh/conf/publication.yaml
Emrys365, line 1: ditto

egs3/TEMPLATE/esp2_enh/conf/training.yaml
Emrys365, line 1: ditto

egs3/TEMPLATE/enh/run.py
```diff
@@ -0,0 +1,175 @@
...
+)
+from espnet3.utils.stages_utils import (
+    parse_cli_and_stage_args,
+    resolve_stages,
+    run_stages,
+)
+
+DEFAULT_STAGES: List[str] = [
+    "create_dataset",
+    "train_tokenizer",
```
Emrys365: Since this stage is not needed in speech enhancement, you can remove it.

```suggestion
```

egs3/TEMPLATE/enh/run.py
```diff
@@ -0,0 +1,175 @@
...
+        help="Hydra config for pack/upload stages.",
+    )
+    parser.add_argument(
+        "--demo_config",
+        default=None,
+        type=Path,
+        help="Hydra config for pack_demo/upload_demo stages.",
+    )
+    parser.add_argument("--dry_run", action="store_true")
+    parser.add_argument("--write_requirements", action="store_true")
```
Emrys365: Add the `help` string for each argument to explain their meaninig.

egs3/TEMPLATE/enh/run.py
```diff
@@ -0,0 +1,175 @@
...
+        publication_config=publication_config,
+        demo_config=demo_config,
+    )
+    run_logger.info("System: %s", system_cls.__name__)
+    run_logger.info("Requested stages: %s", args.stages)
+    run_logger.info("Resolved stages: %s", stages_to_run)
+
+    required_configs = {
+        "create_dataset": training_config,
+        "train_tokenizer": training_config,
```
Emrys365:
```suggestion
```

egs3/vctk_noisy/enh/conf/metrics.yaml
Emrys365, line 1: Should add some common speech enhancement metrics here.

egs3/vctk_noisy/enh/conf/publication.yaml
Emrys365, line 1: This config should also be more concrete to guide the users on how to pack and upload a model.

egs3/vctk_noisy/enh/conf/training.yaml
```diff
@@ -0,0 +1,94 @@
...
+        split: train
+        data_path: ${dataset_dir}
+        chunk_length: 32000 #16000
+  valid:
+    - data_src_args:
+        split: valid
+        data_path: ${dataset_dir}
+  preprocessor:
+    _target_: espnet2.train.preprocessor.EnhPreprocessor
+    train: true
```
Emrys365: You can remove this specific argument in the latest codebase. It will be automatically handled in the ESPnet3 framework.

```suggestion
```

egs3/vctk_noisy/enh/conf/training.yaml
```diff
@@ -0,0 +1,94 @@
...
+    nonlinear: relu
+
+  decoder: conv
+  decoder_conf:
+    channel: 256
+    kernel_size: 20
+    stride: 10
+
+  model_conf:
+    loss_type: ci_sdr
```
Emrys365: Can you change this to 'si_snr' as it is more commonly-used and stable in the literature?

egs3/vctk_noisy/esp2_enh/conf/training.yaml
```diff
@@ -0,0 +1,94 @@
...
+        type: unsorted
+        shape_files:
+          - ${stats_dir}/train/feats_shape
+        batch_size: 8
+
+##########################################################
+#                    TRAINER SETTING                     #
+##########################################################
+trainer:
+  gradient_clip_val: 5.0
```
Emrys365, line 105: You might also want to configure the `max_epochs` argument for each concrete recipe.

egs3/vctk_noisy/esp2_enh/README.md
Emrys365, line 1: The `dataset` directory is missing for this recipe.

egs3/vctk_noisy/esp2_enh/README.md
```diff
@@ -0,0 +1,49 @@
...
+- Training split: 26 speakers (all except p226, p287)
+- Validation split: p226, p287
+- Test split: noisy_testset_wav
+
+## Results
+
+| Split | SI-SNR (dB) | PESQ  | STOI  |
+|-------|-------------|-------|-------|
+| test  | 18.73       | 2.551 | 0.937 |
+| valid | 16.15       | 2.270 | 0.865 |
```
Emrys365, line 61: Are the results from this recipe? Could you provide the configuration files that reproduce the results?

egs3/vctk_noisy/enh/src/inference.py
```diff
@@ -0,0 +1,116 @@
...
+            padded = np.zeros((len(speech_mix), max_len), dtype=np.float32)
+            for i, s in enumerate(speech_mix):
+                padded[i, : len(s)] = s
+            speech_mix = padded
+        else:
+            # Single sample: (T,) -> (1, T)
+            if speech_mix.ndim == 1:
+                speech_mix = speech_mix[np.newaxis, :]
+
+        return self._model(speech_mix)
```
Emrys365: If the input `speech_mix` is not batched, would it be better to keep the output the same shape (without a singleton batch dim)?

espnet3/systems/enh/metrics/audio.py
Emrys365, line 1: What is the intention of this file? And why is it placed under the `metrics` directory?

espnet3/systems/enh/metrics/audio.py
```diff
@@ -0,0 +1,25 @@
...
+    Args:
+        path: Path to a waveform file.
+        sample_rate: Target sampling rate. If omitted, keep the original rate.
+
+    Returns:
+        A one-dimensional ``float32`` waveform.
+    """
+    waveform, current_rate = sf.read(str(path), dtype="float32", always_2d=False)
+    if waveform.ndim > 1:
+        waveform = waveform.mean(axis=1)
```
Emrys365: Taking the average across all audio channels may not be a good way. In most cases, you can simply take one of the channels (the reference channel).

egs3/mini_an4/enh/dataset/builder.py
modelpath-dev: Good catch — switched to extract_targz.

egs3/mini_an4/enh/src/inference.py
modelpath-dev: Agreed. Wrapper now pads only for the model call, then returns a list of original-length waveforms per speaker. Same change on the vctk side.

egs3/TEMPLATE/esp2_enh/conf/demo.yaml
modelpath-dev: Done — pulled in the same style of comments as the ASR TEMPLATE (adapted for enh).

egs3/TEMPLATE/esp2_enh/conf/inference.yaml
modelpath-dev: Done.

egs3/TEMPLATE/esp2_enh/conf/metrics.yaml
modelpath-dev: Done.

egs3/TEMPLATE/esp2_enh/conf/publication.yaml
modelpath-dev: Done.

egs3/TEMPLATE/esp2_enh/conf/training.yaml
modelpath-dev: Done.

egs3/TEMPLATE/enh/run.py
modelpath-dev: Removed train_tokenizer from DEFAULT_STAGES and from the required_configs map.

egs3/TEMPLATE/enh/run.py
modelpath-dev: Added help strings for all the config / dry-run flags (same wording pattern as ASR TEMPLATE).

egs3/TEMPLATE/enh/run.py
modelpath-dev: Removed.

egs3/vctk_noisy/enh/conf/metrics.yaml
modelpath-dev: Added SI-SNR / PESQ / STOI explicitly in the vctk metrics override.

egs3/vctk_noisy/enh/conf/publication.yaml
modelpath-dev: Expanded pack_model / upload_model with concrete paths, includes, and a short usage note.

egs3/vctk_noisy/enh/conf/training.yaml
modelpath-dev: Dropped train: true — DataOrganizer sets that per split now.

egs3/vctk_noisy/enh/conf/training.yaml
modelpath-dev: Switched model_conf.loss_type to si_snr, and aligned best_model_criterion with valid/si_snr.

egs3/vctk_noisy/esp2_enh/conf/training.yaml
modelpath-dev: Set trainer.max_epochs: 100 on the vctk recipe.

egs3/vctk_noisy/esp2_enh/README.md
modelpath-dev: Added egs3/vctk_noisy/enh/dataset/ (VCTKNoisyDataset + builder that just validates the DEMAND layout).

egs3/vctk_noisy/esp2_enh/README.md
modelpath-dev: Those numbers were from the egs2 Conv-TasNet recipe, not this one,  I marked that clearly in the README. Will replace the table once we have a real measure run here. Config to reproduce training is conf/training.yaml.

egs3/vctk_noisy/enh/src/inference.py
modelpath-dev: Yep, single (T,) input now comes back as (T,) without the extra batch dim.

espnet3/systems/enh/metrics/audio.py
modelpath-dev: It's just a shared wav loader for the SI-SNR / PESQ / STOI metrics (they all need the same soundfile + resample path). Left it under metrics/ so those classes can import it without reaching into a random utils module, happy to move it if you'd rather.

espnet3/systems/enh/metrics/audio.py
modelpath-dev: Switched to the first channel instead of mean.

egs3/vctk_noisy/esp2_enh/conf/inference.yaml
```diff
@@ -0,0 +1,39 @@
...
+        data_path: ${dataset_dir}
+        inference: true
+
+##########################################################
+#                    MODEL AND OUTPUT                    #
+##########################################################
+model:
+  _target_: src.inference.SeparateSpeechWrapper
+  train_config: ${exp_dir}/config.yaml
+  model_file: ${exp_dir}/last.ckpt
```
Emrys365, line 45: Better to explicitly define the variables `dataset_dir` and `exp_dir` to avoid confusion.

espnet3/systems/enh/metrics/audio.py
Emrys365: I think it is better to place the function in one of the metrics and import it for other metrics. @Masao-Someki What's your opinion on this?

egs3/mini_an4/enh/conf/metrics.yaml
Masao-Someki, line 1: I think we can add all metrics in `metrics/` here!

egs3/mini_an4/enh/dataset/__init__.py
```diff
@@ -0,0 +1,6 @@
+"""Mini AN4 enhancement dataset module."""
+
+from .builder import MiniAn4Builder as DatasetBuilder
+from .dataset import MiniAn4EnhDataset as Dataset
```
Masao-Someki: Please use the full module path here!
So that it would looks like `egs3.mini_an4.enh.dataset.builder` and `egs3.mini_an4.enh.dataset.dataset`

egs3/vctk_noisy/enh/conf/metrics.yaml
Masao-Someki, line 1: If you have the metrics definition in the default (TEMPLATE/) config, you can omit it here.

egs3/vctk_noisy/enh/conf/publication.yaml
Masao-Someki, line 1: We can skip the default configurations here!

egs3/vctk_noisy/esp2_enh/dataset/builder.py
Masao-Someki, line 1: Could you add a little more detail to the docstring, such as the rationale for not downloading VCTK automatically? Technically, we can download it using `espnet3.utils.download_utils`, but doing so may lead to redundant copies of the dataset in the user's storage.

egs3/vctk_noisy/esp2_enh/src/inference.py
```diff
@@ -0,0 +1,109 @@
...
+        Args:
+            speech_mix: A waveform with shape ``(T,)`` or ``(B, T)``, or a list
+                of one-dimensional NumPy waveforms. List inputs may have
+                different lengths and are zero-padded only for the model call.
+
+        Returns:
+            A list with one item per separated speaker.
+            - single ``(T,)`` input -> each speaker is ``(T,)`` (no batch dim)
+            - list input -> each speaker is a list of ``(T_i,)`` arrays
+            - already-batched ``(B, T)`` input -> each speaker is ``(B, T)``
```
Masao-Someki, line 81: Could you add example snippets on how this function works? For example we can have 2 example when we pass a single audio and batched audio!

egs3/vctk_noisy/enh/conf/training.yaml
```diff
@@ -0,0 +1,104 @@
...
+#               VCTK-NOISY TRAINING OVERRIDES            #
+##########################################################
+# Quick usage:
+#   python run.py --stages create_dataset --training_config conf/training.yaml
+#   python run.py --stages collect_stats  --training_config conf/training.yaml
+#   python run.py --stages train          --training_config conf/training.yaml
+#
+# Set dataset_dir to your VCTK-DEMAND root (see README.md).
+
+dataset_dir: /DB/VCTK_DEMAND # /path/to/vctk_noisy
```
Masao-Someki: Could you move this to `dataset/config.yaml`?
We usually want to have dataset-related items under `dataset/`, so that it would become much easier to refer this dataset from the other recipes!

espnet3/systems/enh/metrics/pesq.py
```diff
@@ -0,0 +1,53 @@
...
+from espnet3.components.metrics.base_metric import BaseMetric
+from espnet3.systems.enh.metrics.audio import load_audio
+
+try:
+    from pesq import pesq as pesq_fn
+except ImportError:
+    pesq_fn = None
+
+
+class PESQMetric(BaseMetric):
```
Masao-Someki: Could you rename this class to the metrics name only, so that it can be `PESQ`?
For the other systems we keep this rule, that just use the name of the metrics as a class name and not adding other words.

espnet3/systems/enh/metrics/sisnr.py
```diff
@@ -0,0 +1,49 @@
...
+    target = (
+        np.dot(estimate, reference) / (np.dot(reference, reference) + 1e-8)
+    ) * reference
+    noise = estimate - target
+    return float(
+        10.0 * np.log10((np.dot(target, target) + 1e-8) / (np.dot(noise, noise) + 1e-8))
+    )
+
+
+class SISNRMetric(BaseMetric):
```
Masao-Someki: ditto

espnet3/systems/enh/metrics/stoi.py
```diff
@@ -0,0 +1,61 @@
...
+from espnet3.components.metrics.base_metric import BaseMetric
+from espnet3.systems.enh.metrics.audio import load_audio
+
+try:
+    from pystoi import stoi as stoi_fn
+except ImportError:
+    stoi_fn = None
+
+
+class STOIMetric(BaseMetric):
```
Masao-Someki: ditto

espnet3/utils/stages_utils.py
```diff
@@ -181,13 +181,10 @@ def run_stages(
...
+                # Stages are invoked with no CLI args; any TypeError here is an
+                # internal failure (e.g. stats collation), not a bad-arguments
+                # error from the stage signature.
                 fn()
-            except TypeError as e:
-                log.exception("Stage '%s' failed (bad arguments)", stage)
-                raise TypeError(
-                    f"Stage '{stage}' does not accept CLI arguments; "
-                    "put all settings in the YAML config."
-                ) from e
```
Masao-Someki, line 190: Is there any reason on removing this part?

Masao-Someki (commented):
Thanks @modelpath-dev, I left comments on the PR!!

espnet3/systems/enh/metrics/audio.py
modelpath-dev: Yeah that works for me. I'll fold load_audio into one of the metric modules and import it from the others, then drop the standalone audio.py. Happy to follow whatever Masao prefers on the exact home for it.

egs3/vctk_noisy/esp2_enh/conf/inference.yaml
modelpath-dev: Makes sense, will spell both out explicitly in the inference config so it's obvious what gets resolved where.

egs3/mini_an4/enh/conf/metrics.yaml
modelpath-dev: Sure, I'll list the full set from metrics/ in the mini_an4 override.

egs3/mini_an4/enh/dataset/__init__.py
modelpath-dev: Got it, will switch to the full egs3.mini_an4.enh.dataset.* paths.

egs3/vctk_noisy/enh/conf/metrics.yaml
modelpath-dev: Fair point. I had inlined them after an earlier review ask, but if TEMPLATE already has the defaults I'll drop the duplicate and just inherit.

egs3/vctk_noisy/enh/conf/publication.yaml
modelpath-dev: Yep, I'll trim it down to only the vctk-specific bits.

egs3/vctk_noisy/esp2_enh/dataset/builder.py
modelpath-dev: Will do. Main reason is VCTK-DEMAND is large and people often already have it somewhere, so auto-download into the recipe would just make another copy. I'll put that rationale in the docstring.

egs3/vctk_noisy/esp2_enh/src/inference.py
modelpath-dev: Sure, I'll add short examples for single-utt and batched input in the docstring.

egs3/vctk_noisy/enh/conf/training.yaml
modelpath-dev: Good call, I'll move the dataset-related defaults under dataset/config.yaml so other recipes can reuse it cleanly.

espnet3/systems/enh/metrics/pesq.py
modelpath-dev: Will rename to PESQ to match the other systems.

espnet3/systems/enh/metrics/sisnr.py
modelpath-dev: Same here, will rename to SISNR.

espnet3/systems/enh/metrics/stoi.py
modelpath-dev: Same, will rename to STOI.

espnet3/utils/stages_utils.py
modelpath-dev: Yeah, that was intentional. The old branch caught every TypeError from a stage and rewrote it as "Stage X does not accept CLI arguments", which was misleading when the real failure was inside the stage (in our case collect_stats collation raising TypeError: unhashable type: 'dict'). Stages are already called with no args, so I dropped the special case so the original traceback shows up. Happy to discuss if you'd rather keep a narrower check.

Masao-Someki: OK sounds good, thanks!

modelpath-dev:
Hi @Masao-Someki  
Is there anything else that I need to fix in this issue ?

Masao-Someki:
Hi, based on our discussion on system naming rules, could you change the system name to `esp2_enh`?

espnet3/systems/enh/metrics/audio.py
modelpath-dev: Moved `load_audio` into `sisnr.py`, and PESQ and STOI import it from there. `audio.py` is gone.

egs3/vctk_noisy/esp2_enh/conf/inference.yaml
modelpath-dev: Added `dataset_dir`, `exp_tag` and `exp_dir` at the top of the inference config, with a note on where each one comes from.

egs3/mini_an4/enh/conf/metrics.yaml
modelpath-dev: mini_an4 now runs SI-SNR, PESQ and STOI. They come from the TEMPLATE metrics config, so this file is just a comment now, and the CI step installs pesq and pystoi.

egs3/mini_an4/enh/dataset/__init__.py
modelpath-dev: Switched to the full module paths (`egs3.mini_an4.esp2_enh.dataset.*`). Did the same for vctk_noisy.

egs3/vctk_noisy/enh/conf/metrics.yaml
modelpath-dev: Dropped the duplicate definitions. It inherits them from the TEMPLATE now.

egs3/vctk_noisy/enh/conf/publication.yaml
modelpath-dev: Trimmed it down to the model card path only. That one has to stay because the TEMPLATE has no `src/` of its own, so the default points at a file that doesn't exist.

egs3/vctk_noisy/esp2_enh/dataset/builder.py
modelpath-dev: Added the rationale to the module docstring.

egs3/vctk_noisy/esp2_enh/src/inference.py
modelpath-dev: Added single and batched examples to `SeparateSpeechWrapper.__call__`.

egs3/vctk_noisy/enh/conf/training.yaml
modelpath-dev: Moved the default root into `dataset/config.yaml` as `builder.dataset_dir`. The builder and dataset fall back to it when `dataset_dir` is left empty, so the training and inference configs keep it blank unless you want to point at another copy.

espnet3/systems/enh/metrics/pesq.py
modelpath-dev: Renamed to `PESQ`.

espnet3/systems/enh/metrics/sisnr.py
modelpath-dev: Renamed to `SISNR`.

espnet3/systems/enh/metrics/stoi.py
modelpath-dev: Renamed to `STOI`.

modelpath-dev:
Hi @Masao-Someki , 
Renamed the system to `esp2_enh`, so it's now `espnet3/systems/esp2_enh`, `egs3/TEMPLATE/esp2_enh`, `egs3/mini_an4/esp2_enh` and `egs3/vctk_noisy/esp2_enh`.

I also pushed the changes from the earlier round of review that I hadn't pushed yet, and replied on each thread. The mini_an4 enh run (create_dataset through measure) passes locally with SI-SNR, PESQ and STOI.

espnet3/systems/enh/metrics/audio.py
Masao-Someki: Thanks @modelpath-dev and @Emrys365, and I agree with @Emrys365's idea!

ci/test_integration_espnet3.sh
```diff
@@ -51,9 +51,9 @@ run_with_training_config \
 
 cd "${cwd}" || exit
 
-python3 -m pip install -e '.[enh]'
+python3 -m pip install -e '.[enh]' pesq pystoi
```
Masao-Someki: Please add these dependencies in the `pyproject.toml`

https://github.com/modelpath-dev/espnet/blob/c87339507c99c063142216f2ee48276995c78759/pyproject.toml#L168-L171

egs3/vctk_noisy/esp2_enh/dataset/config.yaml
Masao-Someki, line 1: Please check here:

https://github.com/modelpath-dev/espnet/blob/c87339507c99c063142216f2ee48276995c78759/egs3/librispeech_100/asr/dataset/config.yaml#L1-L3

I think you can set the environment variable to specify the path, and resolve the path like this:

https://github.com/modelpath-dev/espnet/blob/c87339507c99c063142216f2ee48276995c78759/egs3/librispeech_100/asr/dataset/builder.py#L23-L34

egs3/vctk_noisy/esp2_enh/dataset/dataset.py
```diff
@@ -50,12 +53,19 @@ class VCTKNoisyDataset(TorchDataset):
...
         inference: bool = False,
         **_kwargs,
     ) -> None:
         self.split = split
+        if data_path is None:
+            data_path = default_dataset_dir()
+        if data_path is None:
+            raise ValueError(
+                "data_path is required. Set builder.dataset_dir in "
+                "dataset/config.yaml to your VCTK-DEMAND root."
```
Masao-Someki: We should be able to specify the path by environment variable. Please also extend the error message based on that!

Masao-Someki (commented):
Thank you, I checked the latest commit and added messages. After these fix I think this PR looks good to me

ci/test_integration_espnet3.sh
modelpath-dev: Moved them into the `enh` extra, so this line is back to plain `pip install -e '.[enh]'`.

egs3/vctk_noisy/esp2_enh/dataset/config.yaml
modelpath-dev: Thanks for the pointer, that is much cleaner. It names `VCTK_DEMAND` through `source_env_var` now (plus the four dirs it needs), resolved in the builder the same way librispeech_100 does. The /DB path is gone.

egs3/vctk_noisy/esp2_enh/dataset/dataset.py
modelpath-dev: Good catch. The dataset goes through the same resolver now, so the env var works here too, and if neither that nor `dataset_dir` is set the error tells you to set `VCTK_DEMAND` and lists the folders it expects.

modelpath-dev:
Thanks, all three are in. Swapped the hard-coded path for VCTK_DEMAND and updated the README and both conf files to match, and pesq/pystoi now live in the enh extra. Ran the mini_an4 enh recipe here and it still scores fine on all three metrics.

modelpath-dev:
Added unit tests for the esp2_enh system (task wiring, the SI-SNR/PESQ/STOI metrics, and the system class), so the codecov patch check has something to measure. While writing them I found `sisnr.py` imported `resampy`, which only comes with the `egs2` extra, so a plain `pip install -e '.[enh]'` couldn't import the metrics. Switched it to `librosa.resample`, which is already a core dependency and what the espnet2 preprocessor uses.

egs3/vctk_noisy/esp2_enh/README.md
```diff
@@ -0,0 +1,61 @@
...
+
+The numbers below are **reference scores from the ESPnet2 recipe**
+`egs2/vctk_noisy/enh1` (Conv-TasNet), not yet re-measured with this ESPnet3
+recipe. After you train and run `measure` here, replace the table with the
+scores from `metrics.json`.
+
+| Split | SI-SNR (dB) | PESQ  | STOI  |
+|-------|-------------|-------|-------|
+| test  | 18.73       | 2.551 | 0.937 |
+| valid | 16.15       | 2.270 | 0.865 |
```
Masao-Someki, line 61: Sorry where did you get these numbers?

espnet3/systems/esp2_enh/metrics/sisnr.py
Masao-Someki, line 1: I think you can refer to this implementation to use the espnet2-compatible sisnr calculation? I think at least it's better than re-implementing everything from scratch!
`espnet2/enh/loss/criterions/time_domain.py`

Masao-Someki (commented):
Hi thanks @modelpath-dev ! I added one question and one request.
If you trained a model with this recipe, it would be great if you could put any numbers here in the comment!

PR #6728, opened 2026-09-19
One Speech2Text for both kinds of S2T checkpoint

sw005320:
Asked whether best-path CTC decoding could be handled by `Speech2Text` rather than by another variant, and whether `Speech2TextGreedySearch` could come along. It can, and this does it.

## Before

Four ways to decode an S2T model, three of them already in the tree:

| class | model | decoding |
|---|---|---|
| `s2t_inference.Speech2Text` | encoder-decoder | beam search |
| `s2t_inference_ctc.Speech2Text` | CTC-only | CTC prefix beam search |
| `s2t_inference_ctc.Speech2TextGreedySearch` | CTC-only | best path |
| ~~`Speech2TextCTCGreedySearch`~~ | encoder-decoder | best path (proposed in #6718, withdrawn) |

## After

`s2t_inference.Speech2Text` loads whichever model the training config describes. The CTC-only task writes `model: espnet_ctc`, so one key decides which task builds the checkpoint. A CTC-only model has no decoder, so no beam search is built for it, and `__call__` decodes the best CTC path, which is all such a model can do.

`best_path()` is that decoding, exposed on both kinds. An encoder-decoder model has a CTC branch beside its decoder, and reading it directly is the fast, rough transcript that a demonstration or a sanity check wants.

`Speech2TextGreedySearch` keeps only what the base class does not do: a `batch_decode` that takes file paths and cuts long audio into overlapping buffers. Its constructor, `__call__`, `_decode_interctc` and `_decode_single_sample` are gone, inherited instead, and it now refuses a checkpoint with a decoder instead of reading its CTC head by surprise. That is 259 lines lighter.

## Why not the existing options

`ctc_weight=1.0, beam_size=1` does drop the decoder from the search: the scorers left are `ctc` and `scorefilter`. But it is CTC *prefix* search, which scores the whole vocabulary at every frame. Ten seconds of audio, OWSM v3.1 base, CPU, warmed up, best of two:

| | time | transcript |
|---|---|---|
| `best_path()` | 0.9 s | holiday strategy |
| `ctc_weight=1.0, beam_size=1` | 25.3 s | holidays strategy |
| default beam search | 2.7 s | holiday strategy |

Twenty-eight times slower than best path, and nine times slower than the search it was meant to replace, which is not a first look at anything.

## Verified against the published models

- `espnet/owsm_ctc_v4_1B` through `Speech2Text`: `ctc_only` is true, `beam_search` is None, and `__call__` gives the same sentence as `Speech2TextGreedySearch` does.
- `Speech2TextGreedySearch` still decodes it, and `batch_decode("audio.wav")` still runs the long-form path.
- `espnet/owsm_v3.1_ebf_base` decodes unchanged with beam search, and `best_path()` on it takes 0.8 s against 2.7 s.

Five new tests. The three `*_quantized` tests fail on my machine with and without this change: no quantization engine in this PyTorch build.

## What is left for later

`s2t_inference_ctc.Speech2Text` - the CTC prefix beam search - is untouched. Folding it in means deciding what `batch_decode` should mean when the two classes' versions take different arguments and return different shapes, which is a second question and does not have to be answered here.


sw005320:
Done, in `b9e00124c40`. `Speech2TextGreedySearch` is no longer a class anyone should reach for: the public surface is the three methods on `Speech2Text`.

```python
s2t = Speech2Text.from_pretrained("espnet/owsm_ctc_v4_1B")
s2t(speech)                    # one utterance
s2t.best_path(speech)          # the CTC branch, on either kind of checkpoint
s2t.decode_long("audio.wav")   # a whole recording -> [(start, end, text), ...]
```

**`decode_long` covers both kinds and returns one shape.** An encoder-decoder model emits timestamps, so it is decoded segment by segment and returns one entry per utterance; a CTC-only model has none, so it is decoded in overlapping buffers and returns a single entry covering the recording. It takes a path, an array or a tensor, which is what let the file reading in `Speech2TextGreedySearch.batch_decode` disappear rather than move.

**`Speech2TextGreedySearch` is a deprecated forwarder**, about thirty lines: a `DeprecationWarning`, `decode_long_batched_buffered` and `batch_decode` calling `decode_long`. It is not removed outright because the two Spaces this repository maintains, and the usage sections on the Hub model cards, name it. One behaviour changed: its `batch_decode` no longer fills a batch with buffers taken from several recordings. Each recording's own buffers are still batched, which is where the time goes; packing across recordings only helped at the tail of each one.

**`from_pretrained` became a classmethod.** It named `Speech2Text` explicitly, so every subclass was handed a base-class instance. `Speech2TextGreedySearch.from_pretrained(...)` returned a `Speech2Text` and then failed on its own arguments, which is how I found it.

`espnet2/bin/s2t_inference_ctc.py` is 403 lines lighter. Three new tests cover `decode_long` on a CTC-only checkpoint, the deprecated class returning the same text through both of its old entry points, and `from_pretrained` building the class it was called on.

Verified again on `espnet/owsm_ctc_v4_1B`: `decode_long` through the base class and `batch_decode` through the deprecated one return the same text.

Still to come, deliberately in its own pull request: `espnet2/bin/cli.py` calls the deprecated `batch_decode`, and should call `decode_long`. That file is being changed by #6722 and #6725 already, and a third change to it here would collide with both.

test/espnet2/bin/test_s2t_inference_ctc.py
Masao-Someki, line 1: I think we should copy all tests related to the previous implementations to the `s2t_inference.Speech2Text`? For example, we can consider testing the quantization support like `test_Speech2TextGreedy_quantized`

espnet2/bin/s2t_inference.py
```diff
@@ -225,28 +322,71 @@ def __init__(
...
                 ),
-                sos=s2t_model.sos,
-                eos=s2t_model.eos,
-                vocab_size=len(token_list),
-            ),
-        )
+            )
+            weights = dict(
+                decoder=1.0 - ctc_weight,
+                ctc=ctc_weight,
```
Masao-Someki, line 387: I think it would be helpful to log info or warning that setting `ctc_weight=1` does not mean greedy decoding based on the CTC probabilities. I think this part can be confusing for users.

Masao-Someki (commented):
Thank you!
I've searched over the repo, and I think we might need to change the SDS recipe from Speech2TextGreedySearch to `s2t_inference.Speech2Text`. ([espnet2/sds/asr/owsm_ctc_asr.py](https://github.com/sw005320/espnet-1/blob/69f7d271eb64103bc2c505fd27da13e013e25540/espnet2/sds/asr/owsm_ctc_asr.py#L6))

I think we also need to fix the OWSM-V4 demo on HF space once this PR has merged.

sw005320:
Thank you - the SDS recogniser was there, and so was the MCP server. Both are moved.

One thing worth naming, because it would have been a silent regression: they called `Speech2TextGreedySearch(...)(speech)`, and that class always meant best path. In this pull request `__call__` searches, so a straight swap to `Speech2Text` would have turned a dialogue turn into a CTC beam search. They call `best_path()` now, which is the same decoding as before. The MCP server's `batch_decode(path)` became `decode_long(path)` for the same reason.

Checked on `espnet/owsm_ctc_v3.2_ft_1B` on a CPU: `OWSMCTCModel` warms up in 17 s and transcribes in 9 s, with the transcript unchanged.

Two callers are deliberately left behind:

- `espnet2/bin/cli.py`, because #6722 and #6725 are both editing that file; it follows as soon as they land.
- `egs2/owsm_ctc_v4/s2t1/demo/app.py`, the Space, which installs released espnet - it cannot use any of this until a release carries it.

On the Space: agreed, and it is the same for the model cards on the Hub, which name `Speech2TextGreedySearch` in their usage sections. Since 202610 has not been announced widely, the plan Shinji and I settled on is a 202610 patch release carrying this, and then updating the Spaces, the cards and `espnet_model_zoo`'s snippet template together, rather than waiting for 202612.

sw005320:
Right — the forwarders were for code outside the repository, but everything
inside it was still calling them, so `espnet asr` printed a DeprecationWarning
about an interface the user never chose, and the new one had no callers proving
it works. All of them now moved:

| | |
|---|---|
| `espnet2/bin/cli.py` | `asr` and `translate` decode through `Speech2Text.decode_long` |
| `espnet/__init__.py` | `espnet.load(task="s2t")` builds `Speech2Text` |
| `egs2/owsm_ctc_v4/s2t1/demo/app.py` | the Space calls `best_path()` and `decode_long()` |
| `README.md`, `egs2/owsm_ctc_v3.1/s2t1/README.md`, `egs2/powsm_ctc/s2t1/README.md` | the snippets people copy |

What is deliberately left: `espnet2/bin/s2t_inference_ctc.py` itself (the
forwarders), `test/espnet2/bin/test_s2t_inference_ctc.py` (which tests that they
forward), and the module's name in `ci/check_inference_imports.py` (it must stay
importable).

**One user-visible gain beyond the warning.** The CLI could only take a CTC-only
checkpoint. On master:

```
$ espnet asr audio.wav --model espnet/owsm_v3.1_ebf_base
AttributeError: 'Namespace' object has no attribute 'promptencoder'
```

because the tag was loaded by the CTC task whatever it held. With this branch it
transcribes — one class reads the config and builds either model. Checked both
ways on the same file:

- `espnet/owsm_ctc_v3.2_ft_1B` → "The sale of the hotels is part of Holiday's strategy to sell off assets and concentrate on property management"
- `espnet/owsm_v3.1_ebf_base` → the same sentence, lowercased, 12 s vs 31 s

**Why `decode_long` and not `__call__`** in the two commands: on a CTC-only
checkpoint `decode_long` is best-path decoding in overlapping buffers, which is
what `batch_decode` did; on an encoder-decoder one it is a segment-by-segment
beam search on the model's own timestamps. Calling the object instead would run
a CTC prefix beam search on OWSM-CTC — two orders of magnitude slower for
nothing anyone asked for. The top-level API docstring and the recipe READMEs now
say which is which, since `espnet.load(task="s2t")` hands the object straight to
a user.

Also in this push, fixing the two red checks: the routing test replaced the
class on its module, which a `from_pretrained` written as a classmethod never
looks at, and the MCP test's fake still offered `batch_decode`.

Note for after the merge: the `espnet/owsm-ctc-v4` Space serves the app from
this directory, so it needs re-uploading to pick the change up. The three Hub
model cards still show `Speech2TextGreedySearch` and are a separate pass.

popcornell:
Can we add a warning for ctc only and prefix when no lm is used ? Since the gain would be very small. 

```python
if self.ctc_only and lm_weight == 0.0 and ngram_weight == 0.0:
    logging.warning(
        "This checkpoint is CTC-only and no language model or n-gram was "
        "given, so __call__ runs a CTC prefix beam search over the whole "
        "vocabulary with nothing to fuse; it is ~25x slower than best_path() "
        "for a marginal gain. We recommend using best_path() unless you add lm_weight or "
        "ngram_weight."
    )
```

espnet2/bin/s2t_inference.py
sw005320: Added, and it is said once per model rather than once per utterance — it lives
in `__init__`, where the weights are decided, not in a decoding method, so a
loop over a test set prints it once:

```
decoding on the CTC head alone: a prefix beam search over 5 hypotheses. This is
not best-path decoding (also called greedy or argmax decoding), and beam_size=1
would not make it so - a prefix search scores label sequences, summing the frame
paths that collapse to each one, where best-path takes the most likely symbol at
each frame and collapses once. They can disagree, and the search is far slower.
Speech2Text.best_path() is best-path decoding.
```

The point you raised is exactly the one I tried to make explicit: `beam_size=1`
is *not* the escape hatch people assume it is, so the message says so rather
than only naming `best_path()`.

The same distinction is now in two more places a reader can arrive from — the
routing comment where the scorers are chosen, and `__call__`'s docstring. Tested
both that it appears for `ctc_weight=1.0` and not for `ctc_weight=0.3`, and that
decoding three utterances after construction leaves the count at one.

I dropped a "two orders of magnitude slower" clause I had written first: I
measured 3x on OWSM-CTC v4 and over 100x on OWSM v3.1 base, which is too wide a
spread to put a single number on, so it just says "far slower".

PR #6730, opened 2026-09-19
Let batch_decode take a padded numpy array or a list of utterances

amirlankalm:
## What did you change?

`Speech2Text.batch_decode` in `espnet2/bin/asr_inference.py` and `espnet2/bin/s2t_inference.py` accepts the same input kinds as `__call__`: a padded tensor, a padded numpy array, or a list of unpadded utterances (tensors or arrays), which is padded here. `speech_lengths` may be a tensor, a numpy array or a list of ints. Before, `typeguard` rejected a padded 2-D numpy array (a `numpy.ndarray` is not a `Sequence`), and the s2t version accepted only a tensor.

---

Item 6 of @sw005320's handoff notes for the batched decoding work: harmless from the CLI, where the tensors come from the loader, but an odd edge for the Python API, where `__call__` takes a numpy array and `batch_decode` did not.

---

## Is your PR small enough?

Yes: about 30 lines across the two inference modules, plus one test in each test file.

---

## Additional Context

`test_Speech2Text_batch_decode_accepts_numpy_and_lists` in both test files: a padded numpy array with numpy lengths, a list of utterances, and (asr) a Python list of lengths all give the same tokens and scores as the tensor path. Existing `batch_decode` tests pass. black / isort / flake8 (incl. hacking) clean.


amirlankalm:
done in 55d29d3ae: both tests assert the result count before comparing.

PR #6732, opened 2026-09-19
Follow espnet/notebook's new layout

sw005320:
espnet/notebook#54 restructures that repository:

```
Demos/{ASR,SE,SLU,TTS,Others}
Courses/{CMUSpeechTechnology26S,CMUSpeechProcessing23S,CMUSpeechRecognition22F,CMUSpeechRecognition21F}
```

The `ESPnet2` level is gone - a visitor does not need to know what ESPnet2 is to find a demo - and each course directory is `<name><year><term>` rather than four spellings of a term.

Ten links here point at the old paths: seven in `README.md`, one each in `egs2/TEMPLATE/tts1`, `egs2/TEMPLATE/asvspoof1` and `egs2/aishell3/tts2`. `ci/check_demo_links.py` carries two of the old paths in its own fixtures, so those move too, and its self-check passes.

Two smaller things that fall out of the same restructure:

- `doc/index.md`'s second card linked to `notebook/#espnet-ez`. That repository no longer has an ESPnet-EZ section, so the card points at the courses and says what they are.
- `doc/notebook2rst.sh` deleted an `ESPnet1` directory before converting. It is gone too.

**Merge this after espnet/notebook#54.** Until that merges these paths do not exist, and `check_demo_links.py` runs daily against the live repository.


Masao-Someki (approved):
LGTM!

PR #6733, opened 2026-09-19
Enhance with the URGENT 2025 TF-GridNet by default

sw005320:
`espnet enhance` and the MCP server's `enhance` tool named two different
checkpoints — `espnet/Wangyou_Zhang_universal_train_enh_uses_refch0_2mem_raw`
(USES) and `espnet/Wangyou_Zhang_chime4_enh_train_enh_conv_tasnet_raw`, the
latter fixed at 16 kHz — so the same recording came back two different ways
depending on which front end asked. Both now load
[`kohei0209/tfgridnet_urgent25`](https://huggingface.co/kohei0209/tfgridnet_urgent25),
the URGENT 2025 challenge baseline: one model for 8–48 kHz, 8.5 M parameters,
34 MB. #6723 points the `espnet/universal-se` Space at the same checkpoint, so
all three surfaces enhance alike.

### What I measured

Ran both the old and the new default on the same degraded signals, CPU, 4 s of
speech each, SI-SNR against the clean reference:

| condition | noisy in | tfgridnet_urgent25 | USES (previous default) |
|---|---|---|---|
| white noise, 8 kHz | −6.61 dB | **6.43 dB** (5.9 s) | **6.74 dB** (6.7 s) |
| white noise, 16 kHz | −6.43 dB | **6.84 dB** (12.7 s) | **8.12 dB** (15.2 s) |
| white noise, 48 kHz | −6.42 dB | **9.16 dB** (34.3 s) | **10.09 dB** (54.5 s) |
| 3.4 kHz lowpass + clipping + noise, 16 kHz | −4.59 dB | **−2.38 dB** | **−1.26 dB** |

So on these four, USES scores higher on SI-SNR and the new default is
1.1–1.6× faster. These are synthetic degradations scored with a
waveform-level metric, not the URGENT 2025 evaluation, and TF-GridNet here was
trained with an L1 time+magnitude loss rather than an SI-SNR one — a metric it
is not optimised for is the wrong way to rank them. I am reporting the numbers
rather than claiming they support the change; the reason for the change is the
choice of baseline, not these four rows.

### The organisation question

This is the first default outside the `espnet` organisation, and that is a real
cost: the owning account can rename or delete the repository, and every
`espnet enhance` with no `--model` then fails to download.
`test_every_default_names_a_model_in_the_espnet_organisation` became
`test_every_default_names_a_published_model`, which still rejects a typo and
now carries the exception as a named list with its reason — a second one cannot
be added silently, and the entry disappears once the checkpoint is mirrored
into the organisation. **Mirroring it is worth doing** (CC-BY-4.0, so a copy
with credit is allowed).

### Checked

- `espnet enhance noisy.wav -o clean.wav` end to end on the new default.
- `test/espnet2/bin/test_cli.py` — 33 passed.


Emrys365 (approved):

sw005320:
Closing: we are unifying on USES instead. `espnet enhance` and the `universal-se` Space already load it, and my own measurements in the table above had USES ahead on every condition I ran, so swapping them was the wrong direction.

What is still worth doing from this branch is the other half: the MCP server's `enhance` tool loads a CHiME-4 Conv-TasNet, trained on one corpus at one sample rate, while the other two front ends load USES. That goes into its own pull request against USES.

PR #6735, opened 2026-09-19
[espnet3-recipe] add ST system, BLEU metric, and two multi-tokenizer fixes

chenehk:
## What did you change?

Added `espnet3/systems/st/`:

- **`system.py`** — `STSystem`, which trains one SentencePiece model per side. Also overrides `_has_tokenizer` and extends `collect_stats`.
- **`text_case.py`** — ports of `st.sh`'s `tc`/`lc`/`lc.rm` conventions, espnet's `remove_punctuation.pl`, and Moses `normalize-punctuation.perl`.
- **`metrics/bleu.py`** — corpus BLEU via sacreBLEU.

Fixed two things needed by any multi-tokenizer task:

- **`utils/task_utils.py`** — the model config now wins over the preprocessor config for keys it defines.
  - e.g., `model.bpemodel` vs. `preprocessor.bpemodel`
- **`components/data/collect_stats.py`** — also record token-stream shapes (shapes only, no statistics) for batch_bins.

And **`.gitignore`** — un-exclude `egs*/*/*/dataset/`.

---

`STTask` carries two vocabularies — target for the ST decoder and ST-CTC, source for the auxiliary ASR decoder and ASR-CTC — which `ASRSystem`'s single-vocabulary tokenizer stage cannot express. Building a working ST recipe on espnet3 surfaced three blockers:

1. **`STSystem` could train tokenizers but never a model.** `ASRSystem.train` calls `_has_tokenizer()`, which reads `tokenizer.save_path`; a two-vocabulary config stores sides under `tokenizer.tgt`/`tokenizer.src`, so the stage aborted before its first step with `ConfigAttributeError: Missing key save_path`.

2. **The saved config could not reload its own checkpoint.** `save_espnet_config` flattens the preprocessor over the model config at the root of the espnet2 config; they overlap on `token_list`/`token_type`/`bpemodel`, which are scalars in the model but **lists** in a multi-tokenizer preprocessor. `Task.build_model` then read `vocab_size = len(list) = 2`:

   ```
   size mismatch for decoder.output_layer.weight: copying a param with
   shape [4000, 256] ... the shape in current model is [2, 256]
   ```

   Training was unaffected (it builds from `config.model`); `infer` and `pack_model` were broken. Single-vocabulary tasks are immune, since there the two values are identical.

3. **Batching could not see text length.** `collect_feats` returns acoustic features, so a sampler bounded on feats alone let a batch of short utterances grow without limit — 1,732 utterances in one batch, exhausting a 40 GB A100. The collated batch already carries the token streams; the writer tracks shape keys and stats keys separately, so no stats file changes.

`.gitignore`'s `egs*/*/*/data*` also globs `dataset/`, making a new recipe's Python package invisible to `git add`.

---

## Is your PR small enough?

Yes — 8 files, 512 lines.

---

## Additional Context

Prerequisite for the MuST-C ST recipe, submitted separately to keep each reviewable.

No existing behaviour changes for single-vocabulary tasks: the `task_utils` fix is a no-op when the two values agree, and the `collect_stats` change only adds shape files.

No unit tests yet — `normalize_punctuation` in particular is a ~40-rule Perl port whose rule ordering is load-bearing, and would benefit from golden-string tests.


sw005320:
@chenehk, you should be very careful when you introduce the metrics.
- Can we reuse the existing stable implementation?
- Is there any other implementation that we can use in espnet?
- If you really need to implement it yourself, does it really work without any errors in many scenarios in MT?

Since you're not an expert in MT, I don't recommend that you implement it yourself. 

Check https://github.com/search?q=repo%3Aespnet%2Fespnet+BLEU&type=code

chenehk:
> @chenehk, you should be very careful when you introduce the metrics.
>
>Can we reuse the existing stable implementation?
>Is there any other implementation that we can use in espnet?
>If you really need to implement it yourself, does it really work without any errors in many scenarios in MT?
>Since you're not an expert in MT, I don't recommend that you implement it yourself.
>
>Check https://github.com/search?q=repo%3Aespnet%2Fespnet+BLEU&type=code

@sw005320 Yes, I agree! The [espnet3/systems/st/metrics/bleu.py](https://github.com/espnet/espnet/pull/6735/changes#diff-f035b2526002fe8ee92ffdd2d305c0684e0a04a8d7cd3d0578faea6a3f0d8ba4) is basically a wrapper for `sacrebleu`, adding customized text normalization styles. Do you have any suggestions other than making a wrapper in the system like this?

sw005320:
> > @chenehk, you should be very careful when you introduce the metrics.
> > Can we reuse the existing stable implementation?
> > Is there any other implementation that we can use in espnet?
> > If you really need to implement it yourself, does it really work without any errors in many scenarios in MT?
> > Since you're not an expert in MT, I don't recommend that you implement it yourself.
> > Check https://github.com/search?q=repo%3Aespnet%2Fespnet+BLEU&type=code
> 
> @sw005320 Yes, I agree! The [espnet3/systems/st/metrics/bleu.py](https://github.com/espnet/espnet/pull/6735/changes#diff-f035b2526002fe8ee92ffdd2d305c0684e0a04a8d7cd3d0578faea6a3f0d8ba4) is basically a wrapper for `sacrebleu`, adding customized text normalization styles. Do you have any suggestions other than making a wrapper in the system like this?

This is what we used to use.
https://github.com/espnet/espnet/blob/master/egs2/TEMPLATE/st1/st.sh#L1551
Do you have sufficient treatment?

chenehk:
> > > @chenehk, you should be very careful when you introduce the metrics.
> > > Can we reuse the existing stable implementation?
> > > Is there any other implementation that we can use in espnet?
> > > If you really need to implement it yourself, does it really work without any errors in many scenarios in MT?
> > > Since you're not an expert in MT, I don't recommend that you implement it yourself.
> > > Check https://github.com/search?q=repo%3Aespnet%2Fespnet+BLEU&type=code
> > 
> > 
> > @sw005320 Yes, I agree! The [espnet3/systems/st/metrics/bleu.py](https://github.com/espnet/espnet/pull/6735/changes#diff-f035b2526002fe8ee92ffdd2d305c0684e0a04a8d7cd3d0578faea6a3f0d8ba4) is basically a wrapper for `sacrebleu`, adding customized text normalization styles. Do you have any suggestions other than making a wrapper in the system like this?
> 
> This is what we used to use. https://github.com/espnet/espnet/blob/master/egs2/TEMPLATE/st1/st.sh#L1551 Do you have sufficient treatment?

The implementation only reports one BLEU score (after normalization), whereas the original `st.sh` reports both (1) cased (2) normalized + punctuation removed scores. This was fixed at [a3f1367](https://github.com/espnet/espnet/pull/6735/commits/a3f1367c7d882e1e4886cade5d11e2326cf302a2), thank you for pointing this out!

egs3/must_c/st/conf/tuning/train_st_conformer.yaml
```diff
@@ -0,0 +1,324 @@
...
+
+##########################################################
+#                        DATASET                         #
+##########################################################
+# egs2: train_set=train.en-de, train_dev=dev.en-de,
+#       test_sets="tst-COMMON.en-de tst-HE.en-de"
+dataset:
+  train:
+    - data_src: must_c/st
+      data_src_args: &pair_args
```
Masao-Someki, line 47: Could you add a comment on this format?

egs3/must_c/st/conf/inference.yaml
Masao-Someki, line 1: You can drop some configs, if they're the same as default values.

egs3/must_c/st/conf/publication.yaml
Masao-Someki, line 1: ditto

egs3/must_c/st/dataset/builder.py
```diff
@@ -0,0 +1,433 @@
...
+        self,
+        recipe_dir: str | Path,
+        source_dir: str | Path | None = None,
+        tgt_lang: str | None = None,
+        **_kwargs,
+    ) -> bool:
+        """Check whether the required MuST-C splits are available."""
+        recipe_root = Path(recipe_dir).resolve()
+        try:
+            selected = tgt_lang or TGT_LANG
```
Masao-Someki, line 185: I think it's better to add warning here when we select `TGT_LANG` (=default language)

egs3/must_c/st/dataset/builder.py
```diff
@@ -0,0 +1,433 @@
...
+            recipe_dir: Recipe root directory.
+            source_dir: Optional override pointing at the raw corpus root.
+            **_kwargs: Unused extra options for API compatibility.
+
+        Raises:
+            FileNotFoundError: If the language-pair directory or required
+                splits are missing.
+        """
+        recipe_root = Path(recipe_dir).resolve()
+        selected = tgt_lang or TGT_LANG
```
Masao-Someki, line 219: ditto

egs3/must_c/st/dataset/builder.py
```diff
@@ -0,0 +1,433 @@
...
+    import inspect
+
+    parameters = inspect.signature(function).parameters
+    variadic = any(
+        parameter.kind == parameter.VAR_KEYWORD for parameter in parameters.values()
+    )
+    clean = {
+        key: value for key, value in kwargs.items() if variadic or key in parameters
+    }
+    return function(**clean)
```
Masao-Someki, line 352: I am not sure why we need this function? I think we can remove this.
If we have kwargs that is not defined in the original function as args, it will be just ignored

egs3/must_c/st/dataset/builder.py
```diff
@@ -0,0 +1,433 @@
...
+        self,
+        recipe_dir: str | Path,
+        source_dir: str | Path | None = None,
+        tgt_lang: str | None = None,
+        **_kwargs,
+    ) -> None:
+        """No-op build step for raw-directory-backed MuST-C access."""
+        self.prepare_source(
+            recipe_dir=recipe_dir, source_dir=source_dir, tgt_lang=tgt_lang
+        )
```
Masao-Someki, line 264: I think we can remove this function? It does nothing

egs3/must_c/st/dataset/builder.py
```diff
@@ -0,0 +1,433 @@
...
+        for target in targets:
+            lang_pair_root = resolve_source_root(recipe_root, source_dir, target)
+            missing = missing_required_splits(lang_pair_root, target)
+            if missing:
+                raise FileNotFoundError(
+                    f"MuST-C {SRC_LANG}-{target} is missing required splits "
+                    f"{missing} under {lang_pair_root}"
+                )
+
+    def _is_recipe_built(
```
Masao-Someki, line 242: I think we can remove this function

egs3/must_c/st/dataset/builder.py
```diff
@@ -0,0 +1,433 @@
...
+        if cache_root is not None:
+            return all(
+                _cache_has_columns(cache_root / split) for split in _HF_CACHE_SPLITS
+            )
+        return _call_supported(
+            self._is_recipe_built,
+            recipe_dir=recipe_dir,
+            cache=cache,
+            **kwargs,
+        )
```
Masao-Someki, line 278: I think we should avoid this. If we want a huggingface dataset as a cache, then `is_built` should be a function that just check if the cache exists.

egs3/must_c/st/dataset/builder.py
```diff
@@ -0,0 +1,433 @@
...
+            cache=no_cache,
+            **kwargs,
+        ):
+            _call_supported(
+                self._build_recipe_source,
+                recipe_dir=recipe_dir,
+                cache=no_cache,
+                **kwargs,
+            )
+        _build_hf_cache(recipe_dir, cache_root, kwargs)
```
Masao-Someki, line 305: This `build` function will be only executed when `is_built` return False and after `prepare_source`. So you don't have to re-check if cache exists or source is downloaded.

egs3/must_c/st/dataset/builder.py
```diff
@@ -0,0 +1,433 @@
...
+    """Write one HF cache split per required split, skipping complete ones."""
+    import importlib
+    import json
+    import shutil
+
+    import numpy as np
+    from datasets import Dataset as HFDataset
+
+    module = importlib.import_module(__package__)
+    dataset_class = module.Dataset
```
Masao-Someki, line 365: I think you can just import `egs3.must_c.st.dataset.Dataset`. But please write it here, we need to use lazy import otherwise I think it would cause circular import.

egs3/must_c/st/dataset/builder.py
```diff
@@ -0,0 +1,433 @@
...
+    # Deferred: dataset.py imports this module, so a top-level import loops.
+    _read_segment = importlib.import_module(f"{__package__}.dataset")._read_segment
```
Masao-Someki, line 367: ditto

egs3/must_c/st/dataset/builder.py
```diff
@@ -0,0 +1,433 @@
...
+                            "lang_pair": f"{example.src_lang}-{example.tgt_lang}",
+                            "offset": float(example.offset),
+                            "duration": float(example.duration),
+                        }
+                    except Exception as exc:
+                        stream.write(
+                            json.dumps({"raw_index": index, "error": repr(exc)}) + "\n"
+                        )
+
+        HFDataset.from_generator(rows).save_to_disk(str(temporary))
```
Masao-Someki, line 432: I think this is very slow.
Could you split the audio verification code, and use `espnet3.parallel` to parallelize the process? And then with the verified list of wav path we can create HF dataset from the list.

egs3/must_c/st/dataset/config.yaml
```diff
@@ -0,0 +1,42 @@
+builder:
+  source_env_var: MUST_C
+  # Real on-disk location of the raw MuST-C v1 corpus release used to develop
+  # this recipe. Overridden by $MUST_C when that env var is set, or by an
+  # explicit source_dir argument.
+  source_dir_default: /work/hdd/bbjs/shared/corpora/must-c_v1.2
```
Masao-Someki, line 6: Please remove it, you can set `download`

egs3/must_c/st/dataset/dataset.py
Masao-Someki, line 1: Could you revisit this file? I think we assume we create a HF dataset cache in the DatasetBuilder, so you can just load the cache.

egs3/must_c/st/dataset/builder.py
Masao-Someki: But technically we should not use Dataset package in the DatasetBuilder, because DatasetBuilder will just prepare resource, for Dataset class..

egs3/must_c/st/src/inference.py
```diff
@@ -0,0 +1,78 @@
...
+    (``espnet2/bin/st_inference.py``), not inside ``Speech2Text.__call__``;
+    espnet3 calls the class directly, so a single 0.05 s utterance otherwise
+    aborts the whole test set.
+
+    The fallback mirrors egs2 exactly: emit ``" "`` and keep the utterance, so
+    its reference is still scored (as a miss) rather than dropped -- dropping
+    would inflate BLEU by removing the hardest items from the denominator.
+    """
+
+    def __call__(self, *args, **kwargs):
```
Masao-Someki, line 41: Could you add Args (from parent class), Returns, and example when we feed speech and what format of output we get?

egs3/must_c/st/src/inference.py
```diff
@@ -0,0 +1,78 @@
...
+    that flag is off, as spgispeech and librispeech_100 do unconditionally --
+    but integer keys make a bad hypothesis impossible to trace back to a talk.
+    """
+    return {
+        "utt_id": data.get("utt_id", str(idx)),
+        "hyp": model_output[0][0],
+        # `text` is the ST target (test entries use task="st"); `src_text` is
+        # the English side, written so hypotheses can be read against input.
+        "ref": data.get("text", ""),
+        "src": data.get("src_text", ""),
```
Masao-Someki, line 77: Is there any motivation to return the source language text? Is it used for metrics calculation?

egs3/must_c/st/run.py
```diff
@@ -0,0 +1,18 @@
+"""ESPnet3 entry point for MuST-C speech translation.
+
+This recipe reproduces the egs2 ST recipe (``egs2/must_c/st1``) with
+``STSystem``: separate source and target vocabularies, as ``st.sh`` builds.
+"""
+
+from egs3.TEMPLATE.asr.run import (
```
Masao-Someki, line 7: I think you can create a st system

espnet3/components/data/collect_stats.py
```diff
@@ -95,6 +95,29 @@ def collect_stats_batch(
...
+        # record, and a collate fn is free to call it something other than
+        # "<key>_lengths" (espnet3's own test collate returns a bare
+        # "lengths"), so the suffix check above cannot be relied on alone.
+        if tensor.dim() < 2:
+            continue
+        lengths = tensors.get(f"{key}_lengths")
+        for batch_idx, uid in enumerate(list(uids)):
+            length = (
+                int(lengths[batch_idx]) if lengths is not None else int(tensor.shape[1])
+            )
```
Masao-Someki, line 118: I think you can always count the length or at least we should have default value for `lengths` like None, because we should think of someone uses `collect_stats` without espnet's `CommonPreprocessor`.

espnet3/systems/st/system.py
```diff
@@ -0,0 +1,207 @@
+"""ST system: separate source and target vocabularies, as egs2's st.sh builds."""
+
+import logging
+import os
+from importlib import import_module
+from pathlib import Path
+from typing import Iterable, List
+
+from espnet3.systems.asr.system import ASRSystem
```
Masao-Someki, line 9: I think you can start from BaseSystem, I think it won't change anything for this system class.

espnet3/systems/st/system.py
```diff
@@ -0,0 +1,207 @@
...
+from espnet3.systems.asr.tokenizers.sentencepiece import train_sentencepiece
+
+logger = logging.getLogger(__name__)
+
+_SIDES = ("tgt", "src")
+
+
+class STSystem(ASRSystem):
+    """ASRSystem whose tokenizer stage trains TWO SentencePiece models.
```
Masao-Someki, line 18: I think you can mention the difference from ASR system, but since we can start from BaseSystem we might want to change the description here.

espnet3/systems/st/system.py
```diff
@@ -0,0 +1,207 @@
...
+        config = self.training_config.tokenizer
+        sides = [s for s in _SIDES if getattr(config, s, None) is not None]
+        if not sides:
+            return False
+        return all(self._has_tokenizer_side(getattr(config, s)) for s in sides)
+
+    def _gather_texts(self, side: str, side_config) -> tuple[Path, List[str]]:
+        train_path = Path(
+            getattr(side_config, "train_file", "")
+            or f"{self.training_config.data_dir}/train_tokenizer/{side}.txt"
```
Masao-Someki, line 88: Let's avoid hardcoded path, we can just raise error when we don't have that config key.

espnet3/systems/st/system.py
```diff
@@ -0,0 +1,207 @@
...
+            if not reuse:
+                raise RuntimeError(
+                    f"Tokenizer training text already exists: {train_path}"
+                )
+            logger.info("[%s] reusing existing tokenizer text at %s", side, train_path)
+            return train_path, train_path.read_text(encoding="utf-8").splitlines()
+
+        builder_config = side_config.text_builder
+        module_path, func_name = builder_config.func.rsplit(".", 1)
+        builder = getattr(import_module(module_path), func_name)
```
Masao-Someki, line 101: I think you can use hydra here?

espnet3/systems/st/system.py
```diff
@@ -0,0 +1,207 @@
...
+                logger.info(
+                    "[%s] %s -> %s entries, vocab %d",
+                    mode,
+                    path.name,
+                    len(lines),
+                    vocab,
+                )
+
+    def train_tokenizer(self, *args, **kwargs):
+        """Train the target and source SentencePiece models."""
```
Masao-Someki, line 172: Let's add expected configs, and what we get as an example snippet block?

espnet3/systems/st/text_case.py
Masao-Someki, line 1: I think we can rename this file to `normalization.py`

egs3/must_c/st/conf/tuning/train_st_conformer.yaml
```diff
@@ -0,0 +1,324 @@
...
+  # MULTI-tokenizer preprocessor -- exactly what STTask.build_preprocess_fn()
+  # constructs in egs2. `CommonPreprocessor` holds a single tokenizer and a
+  # single `text_name` string; passing it two of anything raises
+  #   TypeError: __init__() got an unexpected keyword argument
+  #              'tokenizer_encode_conf'
+  # Index 0 is the target side (ST decoder / ST-CTC), index 1 the source side
+  # (auxiliary ASR decoder / ASR-CTC), matching `text` / `src_text`.
+  preprocessor:
+    _target_: espnet2.train.preprocessor.MutliTokenizerCommonPreprocessor
+    train: true
```
Masao-Someki, line 82: You can remove this

test/espnet3/systems/st/test_bleu.py
Masao-Someki, line 1: As shinji mentioned we can consider adding sanity check between espnet2 process and espnet3's implementation.

Masao-Someki: For example we can prepare 2 hyps and 1 ref in test_utils, and calculate bleu with the two systems and compare.

test/espnet3/systems/st/test_text_case.py
Masao-Someki, line 1: Let's rename it to test_normalization.py

Masao-Someki (commented):
Thanks @chenehk!
I've added a lot of comments, please take a look and fix it!

Masao-Someki:
Hi, based on our discussion on system naming rules, could you change the system name to `esp2_st`?

egs3/must_c/st/conf/tuning/train_st_conformer.yaml
chenehk: This field is actually [required](https://github.com/espnet/espnet/blob/master/espnet2/train/preprocessor.py#L784). 
Although there are some problems with this code. (e.g., the typo `Mutli` instead of `Multi`) Do you think it would be better if we create our own version?

Masao-Someki: Yeah, the thing is, we need to set it to False in validation, right? We removed it from the config requirement! 

https://github.com/espnet/espnet/blob/0538bb8d923d1d0fb059093ec03bb97a4496f6d6/espnet3/components/data/data_organizer.py#L319-L331

egs3/must_c/st/dataset/builder.py
chenehk: Deleted!

egs3/must_c/st/dataset/builder.py
chenehk: It only does a simple path check now.

egs3/must_c/st/dataset/builder.py
chenehk: It does a simple path check now.

egs3/must_c/st/dataset/dataset.py
chenehk: It's now just loading the cache, one option added though: `min_wav_duration` and `max_wav_duration` are added for length filtering.

egs3/must_c/st/src/inference.py
chenehk: Excluded.

espnet3/systems/st/system.py
chenehk: Fixed

espnet3/systems/st/system.py
chenehk: Fixed to start from `BaseSystem`.

espnet3/systems/st/system.py
chenehk: Fixed.

espnet3/systems/st/system.py
chenehk: Fixed.

espnet3/systems/st/system.py
chenehk: Added.

espnet3/systems/esp2_st/normalization.py
chenehk: Done.

egs3/must_c/st/conf/tuning/train_st_conformer.yaml
chenehk: Fixed! Thank you for the insightful comment!

test/espnet3/systems/esp2_st/test_normalization.py
chenehk: Done

test/espnet3/systems/esp2_st/test_bleu.py
chenehk: Done.

espnet3/components/data/collect_stats.py
chenehk: Fixed.

egs3/must_c/st/dataset/builder.py
chenehk: `Builder` doesn't depend on `Dataset` any more!

chenehk:
@Masao-Someki Thank you for the insightful comments! I learned a lot in the process of applying the changes. They are mostly addressed now. Please kindly review again.

Masao-Someki:
Hi sorry I forgot to mention, would you create an integration test?
You can create one super small recipe for `mini_an4` dataset, and run it from `ci/test_integration_espnet3.sh`.

You can check these files and PRs:
- [ci/test_integration_espnet3.sh](https://github.com/chenehk/espnet/blob/espnet3/st-system/ci/test_integration_espnet3.sh)
- [egs3/mini_an4/](https://github.com/chenehk/espnet/tree/espnet3/st-system/egs3/mini_an4)
- https://github.com/espnet/espnet/pull/6726/changes#diff-03e588e02f42c91b9edbf2812816ee3d684a38a7064bb9174433c9cddafea54c

PR #6736, opened 2026-09-19
[espnet3-recipe] add MuST-C en-de speech translation recipe

chenehk:
## What did you change?

Port of `egs2/must_c/st1` to egs3: conformer ST on MuST-C v1.2 en-de with two SentencePiece vocabularies, matching `run.sh` (`src_nbpe`/`tgt_nbpe` 4000, `src_case=lc.rm`, `tgt_case=tc`).

- **`dataset/`** — reads the raw MuST-C release directly (per-split yaml, `.en`/`.<tgt>` text, per-talk wav sliced by offset/duration), so no Kaldi data preparation is required.
- **`conf/`** — train / infer / measure / pack configs.
- **`src/`** — inference model wrapper and output formatting.

---

To reproduce the egs2 ST recipe in egs3. Three details were needed for fidelity:

- **The 0.1–20 s filter of `st.sh` stage 4** is applied when the Dataset reads its cache, so `collect_stats` sees the same trimmed set the model trains on (train 229,703 → 220,466, dev 1,423 → 1,381). The cache itself stays unfiltered, mirroring `dump/raw/org`, so the bounds can change without a rebuild. Test sets are left whole, as `st.sh` does.

- **`batch_bins` 26,500,000** reproduces egs2's batch across three shape files: 84.8 utterances and 499 s of audio per batch, against egs2's 85 and 499.

- **`sym_space`/`tgt_sym_space`** are set to the SentencePiece word boundary. At espnet2's `<space>` default no vocabulary token matches, so both error calculators rebuilt text with no spaces: `valid/wer` was pinned at exactly 1.000 (a sentence error rate), `valid/cer` understated and `valid/bleu` inflated. Reporting only — no loss uses it, and `best_model_criterion` is `valid/acc`.

`tst-COMMON` keeps 54 utterances too short for conv2d subsampling, because egs2 does not filter test sets. espnet2 guards that in its CLI loop rather than in `Speech2Text.__call__`, so `src/inference.py` reproduces egs2's blank result instead of aborting the test set.

---

## Is your PR small enough?

Files yes (12), lines no (1,602).

A recipe is close to atomic — the dataset package, configs and inference function are interdependent, and splitting further would leave non-functional intermediate states. It is all new files under one directory rather than changes to existing code. The `espnet3` half is already split into a separate PR for this reason. Happy to divide further if you would prefer.

---

## Additional Context

**Depends on** the #6735; merge that first.

**Results** — `valid.acc.ave_10best` at epoch 41 of 80 (training ongoing), beam 10, sacreBLEU `tok:13a`:

| test set | utts | BLEU | BLEU_lc |
|---|---|---|---|
| tst-COMMON | 2,641 | 23.49 | 24.20 |
| tst-HE | 600 | 22.48 | 22.95 |

`egs2/must_c/st1` publishes no results, so there is no like-for-like reference. The nearest is `egs2/must_c_v2/st1`'s `decode_st_conformer` row at 25.7 BLEU — MuST-C **v2** and converged, so indicative only.

**Deviations from egs2**, each marked in the config: no 3× speed perturbation, no Moses tokenizer before SentencePiece, and `num_workers` 4 rather than the inherited 1.

**Known gaps:** results are from an intermediate checkpoint and will be updated at epoch 80; `dataset/config.yaml` carries a cluster-local `source_dir_default`.


chenehk:
I'm merging this into #6735. Closing this PR.

PR #6357, opened 2026-02-06
Add TSE task/system code and LibriMix recipe for ESPnet3

Emrys365:
## What did you change?

1. Added the following recipes
    + `egs3/TEMPLATE/tse`: Added the template recipe for TSE
    + `egs3/librimix/tse`: Added the recipe for TSE on the LibriMix dataset
2. Add the corresponding code for the Target Speaker Extraction (TSE) task
    + `espnet3/systems/tse`: Main code for TSE task
    + `espnet3/publication/demo/session.py`: Add support for loading audio array data (floating) in the demo page
    + `espnet3/utils/download_utils.py`: Add an interface for extracting zip files (used in the LibriMix TSE recipe)
3. Fix some compatibility issues of the TSE-related functions in ESPnet2
    + `espnet2/bin/enh_inference.py`: Now supports input data does not have a batch dimension
    + `espnet2/bin/enh_tse_inference.py`: Now supports input data does not have a batch dimension
    + `espnet2/enh/espnet_model.py`: Fix a bug when intermediate outputs are returned by an SE model for loss computation
    + `espnet2/iterators/chunk_iter_factory.py`: Fix a compatibility issue in ESPnet3
    + `espnet2/train/preprocessor.py`: Fix a compatibility issue with the TSE task in ESPnet3 (for loading string-valued and array-valued enrollment audios)

---

## Why did you make this change?

To accelerate development of ESPnet-3 and enrich its functionality for speech signal processing

---

## Is your PR small enough?

No, because this includes many template and config files for the recipe.

If needed, I can split the TSE system and task code to another PR.

---

## Additional Context

+ https://github.com/espnet/espnet/issues/6133

## Summary by CodeRabbit

* **New Features**
  * Added a complete LibriMix target-speaker extraction workflow for dataset preparation, training, inference, evaluation, packaging, and demos.
  * Added Gradio demos and Hugging Face model/demo publishing support.
  * Added PESQ, STOI/ESTOI, SI-SNR, SDR, DNSMOS, and TSOS evaluation metrics.
  * Added ZIP archive extraction support and reusable TSE configuration templates.

* **Bug Fixes**
  * Added support for unbatched audio inference and improved audio input/output handling.
  * Corrected speaker-count configuration and improved sampling-rate and enrollment handling.
  * Added compatibility with OmegaConf list configurations.

PR #6738, opened 2026-09-19
[espnet3] Bugfix for espnet3.parallel, local environment

Masao-Someki:
## What did you change?

  - Make `parallel.env: local` dispatch `BaseRunner` shards through Dask `LocalCluster` workers.
  - Add `parallel.allow_overwrite_lock` for explicitly replacing stale shard locks.
  - Add local-parallel and stale-lock regression tests

---

`env: local` previously ignored `n_workers` and ran serially, despite the existing LocalCluster backend.

---

## Is your PR small enough?

yes

---

## Additional Context

* #6133
* #6357


Masao-Someki:
@Emrys365 Thank you for let me know about this bug, I have fixed it.
I also noticed we should have a flag to overwrite the `split.N` directories from previous failed trial, so I also commit it here.

espnet3/parallel/base_runner.py
```diff
@@ -336,17 +335,31 @@ def _resolve_shards(self, items: Sequence[Any]) -> List[Dict[str, Any]]:
...
             )
             if self.resume and self.is_shard_done(shard_dir):
                 continue
-            if not self._try_lock_shard(shard_dir):
+            locked = self._try_lock_shard(shard_dir)
+            if not locked:
                 if self.resume and self.is_shard_done(shard_dir):
                     continue
+                if allow_overwrite_lock:
+                    self._get_lock_path(shard_dir).unlink(missing_ok=True)
```
Emrys365, line 360: I think it is better to show a warning here in case that the `allow_overwrite_lock` argument is unintentionally used.

Emrys365 (approved):
LGTM!

PR #6740, opened 2026-09-19
[espnet3-recipe] add an SLU recipe for SLURP

KimHaeri:
## What did you change?

Added `egs3/slurp/slu`, an espnet3 recipe for spoken language understanding on SLURP. It follows `egs2/slurp/slu1` for the task setup and `egs3/librispeech_100/asr` for the recipe layout.

The model is an espnet2 `SLUTask` built through the `task:` field of the training config: a conformer encoder, a deliberation encoder over BERT transcript embeddings, and a `HuggingFaceTransformersPostDecoder`. The intent is predicted as reserved sentencepiece symbols prefixed to the transcript, so one decoded sequence carries both the intent and the words.

Three training configs are provided:

| config | second input to the deliberation encoder |
| --- | --- |
| `conf/tuning/training_conformer.yaml` | none -- the first-pass ASR model |
| `conf/tuning/train_slu_bert_conformer_deliberation.yaml` | ground-truth transcript |
| `conf/tuning/train_slu_bert_conformer_deliberation_asr.yaml` | transcript decoded by the first pass |

The three are stages of one pipeline, not alternatives: the SLU configs initialise their frozen encoder from the first-pass checkpoint, and the third additionally reads hypotheses that the first pass dumped.

Recipe-local pieces:

- **`dataset/`** — validates the corpus layout and writes TSV manifests (`utt_id`, `wav_path`, `intent`, `transcript`) for the four SLURP splits. `SlurpDataset` optionally attaches a `transcript` field, either the reference or hypotheses dumped by a previous inference stage.
- **`src/system.py`** — `SLUSystem(ASRSystem)`, overriding `train_tokenizer`. It reserves the intent labels as sentencepiece user-defined symbols so each stays one token, and writes the transcript token list, the way stage 5 of `egs2/TEMPLATE/slu1/slu.sh` builds both lists together.
- **`src/inference.py`** — splits a decoded sequence back into intent and transcript.
- **`src/metrics.py`** — `IntentAccuracy`, which also dumps the misclassified utterances for error analysis.
- **`src/callbacks.py`** — parameter loading and freezing for the pretrained submodules (`PretrainedParameterLoader`, `ParameterFreezer`), since espnet3's training config has no field for egs2's `--pretrained_model` or `freeze_param`.

Unit tests under `test/egs3/slurp/slu/` cover the manifest builder, the dataset fields, both token lists, inference output parsing and the `System` override (26 cases, +533 lines).

---

espnet3 has no SLU recipe yet, and SLURP is the natural first one because `egs2/slurp/slu1` already establishes the reference setup and numbers to compare against.

---

## Is your PR small enough?

**No** — 25 files, +2773 lines, against the stated ceiling of 20 files / 1000 lines.

The changes are a single new recipe with no modifications to existing files, so the review surface is smaller than the line count suggests: 10 config and `__init__` files, 5 dataset/`src` modules, `run.py` and `path.sh`, and 3 test files (+533 lines) that CI runs.

I did not find a split that leaves both halves meaningful — a recipe without its dataset builder or without its `System` override does not run, and landing the tests separately would mean merging untested code first. Happy to split it differently if reviewers prefer a particular cut.

---

### Where this deliberately differs from egs2

The data preparation and the model architecture match `egs2/slurp/slu1` field for field.
Three things do not match, all of them consequences of espnet3's config surface rather than choices about the task:

- **Vocabulary size.** egs2 tokenizes offline in `local/run_spm.sh`: BPE-500 over the transcripts, the text rewritten as `<intent> <bpe> <pieces>`, then handed to `slu.sh` with `--token_type word`, so the intent survives because the final vocabulary is a word list over already-encoded text (~570 tokens). This recipe does it in one pass inside `train_tokenizer`, with the intents as sentencepiece `user_defined_symbols`, giving 999 tokens — about 1.75x egs2's vocabulary, and a differently segmented subword inventory.
- **Batching and warmup.** egs2 overrides neither, so espnet2's `batch_type: folded` / `batch_size: 20` applies: ~5,994 steps per epoch and ~300k over 50 epochs, of which its `warmup_steps: 25000` is 8%. This recipe batches by `batch_bins`, which gives ~1,222 steps per epoch, so the same 25000 would spend 41% of the run still ramping the learning rate up. The configs scale `warmup_steps` to keep egs2's proportion.
- **Frontend.** egs2 sets no `frontend_conf` and so gets `DefaultFrontend` 32 ms window / 8 ms hop; this recipe uses the 25 ms / 10 ms that the other egs3 recipes set explicitly.

### Results are not included yet

Training runs are still in progress. I will add `README.md` with the numbers and a pre-trained model link in a follow-up PR.


egs3/slurp/slu/conf/inference.yaml
```diff
@@ -0,0 +1,71 @@
...
+##########################################################
+# One shard per worker, merged afterwards. Inference is CPU-bound here --
+# `Speech2Text` defaults to `device: cpu` and this recipe does not override it
+# -- and a single process handles well under one utterance per second at this
+# beam size, so the test set alone would take more than a day serially. Each
+# worker builds its own `Speech2Text` (~440 MB of weights plus the torch
+# runtime), so allow ~2 GB of RAM per worker and keep the count at or below the
+# cores of the node.
+parallel:
+  n_workers: 16
```
Masao-Someki, line 27: For public recipe please set `n_workers` to 1!

Masao-Someki (commented):
I just reviewed half of this PR, could you create a `TEMPLATE/` directory and a system directory under `espnet3/systems`?
I think we can put metrics, callbacks etc inside the systems/ directory.

Masao-Someki:
Hi, based on our discussion on system naming rules, could you change the system name to `esp2_slu`?

KimHaeri:
> I just reviewed half of this PR, could you create a `TEMPLATE/` directory and a system directory under `espnet3/systems`? I think we can put metrics, callbacks etc inside the systems/ directory.

From my understanding of the code structure in `espnet3`, I think it would be better to create `espnet3/systems/slu` directory and move `system.py` and `metrics.py` to this directory.

I also think `callbacks.py` can be moved to `espnet3/components/callbacks`.

So I was wondering what `TEMPLATE/` directory you mentioned is for.

If I misunderstood anything, please let me know!

egs3/slurp/slu/conf/inference.yaml
KimHaeri: I will modify this. Thank you!

KimHaeri:
> Hi, based on our discussion on system naming rules, could you change the system name to `esp2_slu`?

I couldn't quite figure out what you meant by the system name.
Do you mean that we should create `espnet3/systems/esp2_slu` instead of `espnet3/systems/slu`?

Masao-Someki:
Sorry for the confusion, here's more clear description:

> Do you mean that we should create espnet3/systems/esp2_slu instead of espnet3/systems/slu?

Yes, let's create `espnet3/systems/esp2_slu` directory and put `system.py` and other items there!

> From my understanding of the code structure in espnet3, I think it would be better to create espnet3/systems/slu directory and move system.py and metrics.py to this directory.
> I also think callbacks.py can be moved to espnet3/components/callbacks.
> So I was wondering what TEMPLATE/ directory you mentioned is for.
> If I misunderstood anything, please let me know!

Thanks, the `system.py` and `metrics.py` can be placed under `espnet3/systems/esp2_slu`. For the `callbacks.py`, thanks let's put it under `components/`. But before that please rename it so that we consistently have `*Callback` class as callback.

> So I was wondering what TEMPLATE/ directory you mentioned is for.

We have default configuration values under `egs3/TEMPLATE/<system_name>/conf/*.yaml`.
So for example, if we have `egs3/TEMPLATE/esp2_slu/conf/training.yaml`, then all config values in that `training.yaml` will be inserted to your recipe config as a default value. You can just keep the configs that differs from the default config in each recipes.

egs3/slurp/esp2_slu/conf/tuning/train_slu_bert_conformer_deliberation.yaml
```diff
@@ -0,0 +1,239 @@
...
+trainer:
+  gradient_clip_val: 5.0
+  max_epochs: 50
+  precision: bf16-mixed
+  callbacks:
+    - _target_: espnet3.components.callbacks.pretrained_parameters.PretrainedParametersCallback
+      init_param: ${recipe_dir}/exp/training_conformer/valid.acc.ave_10best.pth:encoder:encoder
+
+fit:
+  ckpt_path: last
```
Masao-Someki: I think this doesn't work? maybe `${exp_dir}.last.ckpt`?

egs3/slurp/esp2_slu/conf/tuning/train_slu_bert_conformer_deliberation_asr.yaml
```diff
@@ -0,0 +1,249 @@
...
+trainer:
+  gradient_clip_val: 5.0
+  max_epochs: 50
+  precision: bf16-mixed
+  callbacks:
+    - _target_: espnet3.components.callbacks.pretrained_parameters.PretrainedParametersCallback
+      init_param: ${recipe_dir}/exp/training_conformer/valid.acc.ave_10best.pth:encoder:encoder
+
+fit:
+  ckpt_path: last
```
Masao-Someki: ditto

egs3/slurp/esp2_slu/conf/tuning/training_conformer.yaml
```diff
@@ -0,0 +1,190 @@
...
+
+##########################################################
+#                      FIT SETTING                       #
+##########################################################
+# Resume from the last checkpoint when one exists. Lightning treats
+# `ckpt_path: last` as set-and-forget -- with no checkpoint yet it warns and
+# starts from scratch -- so a run stopped by a scheduler wall clock can be
+# continued by resubmitting the same command.
+fit:
+  ckpt_path: last
```
Masao-Someki: ditto

egs3/slurp/esp2_slu/dataset/builder.py
```diff
@@ -0,0 +1,441 @@
...
+                incomplete. The message names the missing paths.
+        """
+        recipe_root = Path(recipe_dir).resolve()
+        source_root = resolve_source_root(recipe_root, source_dir=source_dir)
+        missing = _find_missing_source_files(source_root)
+        if missing:
+            raise FileNotFoundError(
+                f"SLURP source at {source_root} is incomplete. Missing: "
+                + ", ".join(missing)
+                + "\nSee readme.md for how to obtain the metadata and the audio."
```
Masao-Someki: Could you write the `README.md` mentioning how to obtain the slurp dataset?

egs3/slurp/esp2_slu/run.py
```diff
@@ -0,0 +1,17 @@
+from egs3.TEMPLATE.asr.run import (
```
Masao-Someki: Could you change it to `esp2_slu`?
Let's create esp2_slu system

egs3/TEMPLATE/asr/run.py
Masao-Someki: You can copy run.py to `egs3/TEMPLATE/esp2_slu` system and add this modification!

egs3/TEMPLATE/esp2_slu/conf/metrics.yaml
```diff
@@ -0,0 +1,55 @@
...
+# `--inference_config` provides it. Add `inference_dir` only when you want to
+# score a different inference output directory.
+exp_tag:
+exp_dir: ${recipe_dir}/exp/${exp_tag}
+
+
+##########################################################
+#                    METRIC DEFINITION                   #
+##########################################################
+metrics:
```
Masao-Someki: How about adding the IntentAccuracy here?

egs3/TEMPLATE/esp2_slu/conf/training.yaml
```diff
@@ -0,0 +1,316 @@
...
+##########################################################
+#                  SYSTEM AND PATH SETTING               #
+##########################################################
+num_device: 1
+num_nodes: 1
+
+# The espnet2 task this system is built on. `SLUTask` predicts the intent as a
+# label prefixed to the transcript, and optionally reads a transcript as a
+# second input for its post-decoder.
+task: espnet2.tasks.slu.SLUTask
```
Masao-Someki: You can copy the SLUTask file to `systems/esp2_slu/task.py` and use that file! (Just simple copy is fine)

espnet3/systems/esp2_slu/system.py
```diff
@@ -0,0 +1,242 @@
...
+            save_path,
+            tokenizer_config.vocab_size,
+            character_coverage=float(
+                getattr(tokenizer_config, "character_coverage", 1.0) or 1.0
+            ),
+            model_type=tokenizer_config.model_type,
+            user_defined_symbols=user_defined_symbols,
+        )
+
+    def write_transcript_token_list(self) -> Path | None:
```
Masao-Someki: Is this a new stage for SLU? If not, please add `_` in the beginning to make it a helper function. All public functions of the system class is a stage, and we want to avoid to have any stage-like function!

Masao-Someki (commented):
Thanks @KimHaeri !
Please copy slu task script to `systems/esp2_slu/task.py`, and please setup `egs3/TEMPLATE/esp2_slu/` with run.py and os on.

PR #6741, opened 2026-09-19
[espnet3-recipe] Wrap up for spgispeech-asr, batch_beam_search tested 

chenehk:
## What did you change?

Follow-up to #6656 (merged): training has now run all 35 configured epochs,
and this updates the recipe to match, fixes a builder bug, and keeps
inference working after an upstream default changed.

- **`dataset/config.yaml`** (bugfix): sets `default_source_dir` so
  `builder.py`'s existing fallback (added during review of #6656) actually
  resolves the raw corpus, instead of every caller needing to pass
  `source_dir` explicitly.
- **`src/inference.py`**: `build_output` now handles the batched call
  ESPnet3 makes when `batch_size` is set, recursing over the list like
  `egs3/TEMPLATE/asr/src/inference.py`'s own `build_output`. Needed because
  `egs3/TEMPLATE/asr/conf/inference.yaml`'s default changed from
  `batch_size: null` to `4` in #6658 ("Decode espnet3 test sets in batches
  by default") — without this fix, the recipe would silently fall back to
  serial, one-item-at-a-time decoding.
- **`conf/inference.yaml`**: pins `batch_size: 4` explicitly (rather than
  inheriting the TEMPLATE default silently) and documents the measured
  effect.
- **Reference results updated** for the completed run, in both `readme.md`
  and the training config's comment block.

- The training run this recipe reports on has finished (was 24/35 epochs
  when #6656 merged); the numbers should reflect the final model.
- #6658 landed after #6656 merged and changes espnet3's inference default
  cluster-wide. Without the `build_output` fix, this recipe would keep
  working but silently lose the speedup — worth catching now rather than
  leaving a stale recipe next to an upstream default it doesn't support.
- Batched decoding is a meaningful, measured win, worth pinning and
  documenting rather than leaving implicit — `dev_4k`, 1 GPU, same
  checkpoint:

  | `batch_size` | infer time | WER | CER |
  | --- | --- | --- | --- |
  | `null` (serial) | 88.75 min | 2.26 | 0.93 |
  | `4` (batched) | 31.3 min | **2.25** | 0.93 |

  **2.8x faster**, WER/CER unchanged within noise. The small WER delta
  comes from `Speech2Text.batch_decode`'s padded batching not being
  perfectly invariant for this config's `Conv2d` subsampling and legacy
  relative-position attention — not a decoding bug.

## Is your PR small enough?

Yes — 5 files changed, 50 insertions(+), 14 deletions(-).

## Additional Context

- Follow-up to #6656.
- Depends on #6658 for the batching behavior being documented/pinned here.
- Pretrained model published:
  https://huggingface.co/espnet/spgispeech_asr_train_asr_conformer6_n_fft512_hop_length256
- `val` (39,341 utterances) is still undecoded, unchanged from #6656.

https://claude.ai/code/session_01SUG8bJYjyQqUquCUk69rRE


egs3/spgispeech/asr/dataset/config.yaml
sw005320, line 1: I think we should avoid hard-coding an absolute path here:

default_source_dir: /work/hdd/bbjs/shared/corpora/spgispeech

This path is specific to one environment and makes the recipe less reproducible on other machines/clusters.

Since the recipe already defines source_env_var: SPGISPEECH, could we simply rely on that instead? Users can then set, for example,

export SPGISPEECH=/path/to/spgispeech

for their local environment.

From what I understand, builder.py already considers the explicit source_dir, the recipe-local path, and the SPGISPEECH environment variable when resolving the source, so I don't think we need a machine-specific default_source_dir here.

I would prefer keeping the recipe itself environment-independent.

sw005320: @Masao-Someki, do we have a mechanism like db.sh in espnet2?
We should do this for this purpose.

chenehk: You are right, I should have left this in my local. The current fall-back is `explicit source_dir -> recipe-local download/spgispeech -> $SPGISPEECH`, after which we prompt the user to set the path, which I think should be enough. 

I've deleted the hard-coded path in https://github.com/espnet/espnet/commit/903e3845348db457b8fdf02ac6d9a9995c341ecb, and I will comply if we make a unified path config in the `TEMPLATE`.

sw005320 (approved):
OK, cool!

egs3/spgispeech/asr/dataset/config.yaml
Masao-Someki: We can specify environment variable! You can check [this file](https://github.com/chenehk/espnet/blob/egs3/spgispeech-asr/egs3/librispeech_100/asr/dataset/config.yaml)

PR #6532, opened 2026-08-26
Add ESPnet3 speaker verification system and VoxCeleb recipe

popcornell:
Add a `spk` system to espnet3 and a VoxCeleb 1+2 recipe on top of it.

espnet3/systems/spk:
- `task.py` follows the existing `systems/asr/task.py` convention of carrying the espnet2 task class, and registers the ESPnet SSL frontend so XEUS-style upstreams can be used alongside the s3prl ones.
- `ESPnetSpeakerVerificationModel` trains as closed-set classification but scores trial pairs when a batch carries `speech2`. Closed-set validation loss is a poor proxy for verification quality, and `SpeakerVerificationScoring`returns `valid/eer` and `valid/mindcf` and checkpoints are selected on EER.
- `Speech2Score` produces `EER` and `MinDCF` and score the resulting SCP files in `measure`.

egs3/voxceleb/spk:
- The builder validates the corpus, derives Kaldi-style manifests, downloads and converts the Vox1-O protocol, and lists MUSAN and RIRS_NOISES for augmentation.
- Configurations for RawNet3 from scratch and for ECAPA-TDNN on a **jointly fine-tuned XEUS encoder**, held frozen for the first updates.
- Validation scores a class-balanced 10k subset of Vox1-O each epoch; `infer` scores the full list.

Stages are create_dataset/train/infer/measure plus publication.

## Summary by CodeRabbit

- **New Features**
  - Added speaker-verification support with trial scoring, speaker embeddings, EER, and minDCF metrics.
  - Added complete template and VoxCeleb recipes for dataset preparation, training, inference, evaluation, and model publication.
  - Added support for AAC audio conversion, optional augmentation data, debug runs, and configurable experiment stages.

- **Bug Fixes**
  - Improved checkpoint averaging and distributed validation scoring.
  - Prevented stage-log rotation conflicts and clarified local parallel execution behavior.

- **Documentation**
  - Added setup, dataset requirements, usage instructions, configuration guidance, and model README templates.

PR #6742, opened 2026-09-20
[espnet3] Add VoxLingua107 LID recipe and LIDSystem for ESPnet3

Shunsuke32:
## What did you change?

Port the ESPnet2 VoxLingua107 language identification recipe to ESPnet3. This adds `LIDSystem`, `egs3/TEMPLATE/esp2_lid`, and `egs3/voxlingua107/esp2_lid`.

The recipe includes downloading and extracting the official ZIP files, collecting statistics, MMS/ECAPA-TDNN training, and inference. It provides accuracy, precision/recall/F1, per-language scores, lists of incorrect predictions, embeddings, embedding summaries by language, and t-SNE plots. Optional speed perturbation, model packing, and Hub upload are also supported.

To run LID using ESPnet3's existing tools for datasets, training, inference, and model sharing. The following changes need review.

### 1. Category sampling in the shared DataLoader builder

`DataLoaderBuilder` now calls the existing ESPnet2 `build_category_batch_sampler` for `catbel`, `catpow`, and `catpow_balance_dataset`. It reuses the ESPnet2 sampling code. Other sampler types still use the same function as before.

### 2. LID statistics and combining datasets

`LIDSystem.collect_stats` writes speech shapes and language/category mappings using the sample IDs in the combined Dataset. It also records which source dataset each sample belongs to, for dataset-balanced sampling. Datasets are combined using the existing `DataOrganizer` and `CombinedDataset`.

Supports `extract_feats_in_collect_stats: false`, matching the ESPnet2 VoxLingua107 recipe.

### 3. Recipe callbacks for checkpoints and FP32 precision

The existing `ModelCheckpoint` selects the best checkpoint using `valid/accuracy` and `mode=max`. A recipe callback updates the `valid.accuracy.best.pth` link after each epoch. This lets inference use the best individual checkpoint before training ends.

The ESPnet2 VoxLingua107 recipe uses `use_tf32: false`, which leaves FP32 matrix multiplication at PyTorch's default `highest` in my environment. This recipe uses a callback to set `highest` to match that behavior, while the shared ESPnet3 training code sets `high`. Which should this recipe match: ESPnet2’s `highest` setting or ESPnet3’s shared `high` setting?

### 4. Shared checkpoint averaging

The shared callback includes the `on_train_end` fix from [#6532](https://github.com/espnet/espnet/pull/6532). It runs averaging again after the final checkpoint is saved. The existing checks for matching model keys are kept. This change overlaps with #6532, so we should avoid merging the same fix twice.

## Is your PR small enough?

No.

## Additional Context

Small-scale training, packing, private Hub upload, and inference from the downloaded model bundle were tested with generated data. Downloading and preparation were tested on small real-data samples from FLEURS, ML-SUPERB dev/dev_dialect, and VoxPopuli. Babel requires data from LDC and has not yet been tested with the actual corpus.

I trained the model on VoxLingua107 and evaluated it on each corpus. The results are below.

| Dataset | Split | Evaluated languages | Utterances | Accuracy (%) |
| VoxLingua107 | dev | 33 | 1,609 | 93.78 |
| FLEURS | test | 83 | 63,885 | 94.49 |
| ML-SUPERB 2.0 | dev | 91 | 17,631 | 88.31 |
| ML-SUPERB 2.0 | dev_dialect | 8 | 7,095 | 75.79 |
| VoxPopuli | test | 16 | 16,991 | 88.42 |


Masao-Someki:
Hi, based on our discussion on system naming rules, could you change the system name to `esp2_lid`?

Shunsuke32:
Sure! I've renamed the system to `esp2_lid`. I've also completed training and evaluation and added the results to the PR description and README.

egs3/TEMPLATE/esp2_lid/conf/inference.yaml
```diff
@@ -0,0 +1,13 @@
+defaults:
+  - ../../asr/conf/inference
+  - _self_
```
Masao-Someki: Could you copy asr's inference code here? We want to keep these configs self-consistent, and relative path to identify config inside library is easily cause a bug in my experience..

egs3/TEMPLATE/esp2_lid/conf/metrics.yaml
```diff
@@ -0,0 +1,9 @@
+defaults:
+  - ../../asr/conf/metrics
+  - _self_
```
Masao-Someki: ditto

egs3/TEMPLATE/esp2_lid/conf/publication.yaml
```diff
@@ -0,0 +1,5 @@
+defaults:
+  - ../../asr/conf/publication
+  - _self_
```
Masao-Someki: ditto

egs3/voxlingua107/esp2_lid/dataset/config.yaml
```diff
@@ -0,0 +1,15 @@
+builder:
+  default_source_dir: ${oc.env:VOXLINGUA107,/path/to/voxlingua107}
```
Masao-Someki: You can just specify the environment variable like this:

https://github.com/espnet/espnet/blob/a4851823e8b54ee8295aabfae614f28eafd66a3e/egs3/librispeech_100/asr/dataset/config.yaml#L3

egs3/voxlingua107/esp2_lid/dataset/__init__.py
```diff
@@ -0,0 +1,6 @@
+"""VoxLingua107 dataset module."""
+
+from .builder import VoxLingua107Builder as DatasetBuilder
+from .dataset import VoxLingua107Dataset as Dataset
```
Masao-Someki: Could you use the full module path?
So it would be sth like `egs3.voxlingua107.esp2_lid.dataset.builder`

egs3/voxlingua107/esp2_lid/conf/inference_embeddings.yaml
```diff
@@ -0,0 +1,14 @@
+defaults:
+  - inference
+  - _self_
```
Masao-Someki: Could we avoid using Hydra defaults here? ESPnet3 recipe configs should remain self-consistent. Some duplication is fine: we want each config to describe one complete training experiment, one inference trial, and so on.

egs3/voxlingua107/esp2_lid/conf/inference_prepared.yaml
```diff
@@ -0,0 +1,16 @@
+defaults:
+  - inference
+  - _self_
```
Masao-Someki: ditto

egs3/voxlingua107/esp2_lid/conf/metrics_embeddings.yaml
```diff
@@ -0,0 +1,16 @@
+defaults:
+  - metrics
+  - _self_
```
Masao-Someki: ditto

egs3/voxlingua107/esp2_lid/dataset/builder.py
```diff
@@ -0,0 +1,355 @@
...
+        if self.is_source_prepared(source_dir=source_root):
+            return
+        source_root.mkdir(parents=True, exist_ok=True)
+        marker = source_root / _PREPARING
+        marker.touch()
+        url_list = source_root / "zip_urls.txt"
+        if not url_list.is_file():
+            temporary = source_root / "zip_urls.txt.part"
+            subprocess.run(
+                ["wget", "-O", str(temporary), zip_urls_url or _CFG["zip_urls_url"]],
```
Masao-Someki: You can consider using `espnet3.utils.download_utils` to download files

egs3/voxlingua107/esp2_lid/dataset/builder.py
```diff
@@ -0,0 +1,355 @@
...
+        zip_urls_url: str | None = None,
+        dev_zip_url: str | None = None,
+        **_kwargs,
+    ) -> None:
+        """Download and unzip the official corpus as in ESPnet2 local/data.sh.
+
+        Keep completed archives and resume partial downloads with wget. A marker
+        prevents interrupted extraction from being treated as a prepared source.
+        Audio is extracted unchanged; no cropping or resampling is performed.
+        """
```
Masao-Someki, line 356: Do we have multiple files to download?
I think we can write DownloadProvider/DownloadRunner with provider/runner framework in `espnet3.parallel` to parallelize download and unzip them.

egs3/voxlingua107/esp2_lid/dataset/builder.py
```diff
@@ -0,0 +1,355 @@
...
+        self,
+        source_dir: str | Path | None = None,
+        recipe_dir: str | Path | None = None,
+        data_dir: str | Path | None = None,
+        **_kwargs,
+    ) -> None:
+        """Create deterministic train/development manifests and mappings."""
+        source_root = resolve_source_root(source_dir)
+        if not self.is_source_prepared(source_dir=source_root):
+            self.prepare_source(source_dir=source_root)
```
Masao-Someki: You don't have to check this, because `build` will run only after source-related functions are completed, and `is_built` is false.

egs3/voxlingua107/esp2_lid/dataset/builder.py
```diff
@@ -0,0 +1,355 @@
...
+        )
+
+    def build(
+        self,
+        source_dir: str | Path | None = None,
+        recipe_dir: str | Path | None = None,
+        data_dir: str | Path | None = None,
+        **_kwargs,
+    ) -> None:
+        """Create deterministic train/development manifests and mappings."""
```
Masao-Someki: Let's add examples showing what each file would look like, such as `utt2lang`, `lang2utt`, and `category2utt`, and explain how they are supposed to be used in this recipe.

egs3/voxlingua107/esp2_lid/dataset/dataset.py
```diff
@@ -0,0 +1,118 @@
...
+    def __init__(
+        self,
+        split: str,
+        source_dir: str | Path | None = None,
+        sample_rate: int = 16000,
+        recipe_dir: str | Path | None = None,
+        data_dir: str | Path | None = None,
+        speed_perturb_factors: Sequence[float] | None = None,
+    ) -> None:
+        """Read only the requested split's manifest and referenced audio.
```
Masao-Someki, line 61: Please add Args.

egs3/voxlingua107/esp2_lid/src/babel_prepare.py
Masao-Someki, line 1: This is for evaluation, right?
I think technically we should create babel recipe and add test dataset in the `inference.yaml`
But since currently we don't have them let's keep this preparation script as-is, and add it when we had Babel recipe (which is required when craeting OWSM recipe for me..)

egs3/voxlingua107/esp2_lid/src/babel_prepare.py
```diff
@@ -0,0 +1,101 @@
...
+# Nonlexical events omitted by Babel ASR's prepare_acoustic_training_data.pl.
+NONLEXICAL = re.compile(
+    r"\(\(\)\)|<(?:sta|male-to-female|female-to-male|lipsmack|breath|cough|"
+    r"laugh|click|ring|dtmf|int|foreign|overlap|prompt|no-speech)>|~"
+)
+TIMESTAMP = re.compile(r"^\[([0-9.]+)\]$")
+
+
+def segments(transcript):
+    """Pair consecutive timestamps, keeping segments containing lexical speech."""
```
Masao-Someki: For these public funtions, please add Args, Returns, and Example snippet as a minimal docstring!

egs3/voxlingua107/esp2_lid/src/fleurs_prepare.py
```diff
@@ -0,0 +1,99 @@
...
+    "xh": "xho",
+    "zu": "zul",
+    "fil": "tgl",
+}
+REPOSITORY = "google/xtreme_s"
+REVISION = "e71d660cb63834e9aec8462f50fd0b00232b5aaf"
+
+
+def language_code(tag):
+    """Map the FLEURS locale to the ISO-639-3 labels used for LID."""
```
Masao-Someki: ditto

egs3/voxlingua107/esp2_lid/src/ml_superb2_prepare.py
```diff
@@ -0,0 +1,76 @@
...
+    write_manifest,
+)
+
+REPOSITORY = "espnet/ml_superb_hf"
+REVISION = "5a0634ecdbffdc99d8b55eb2b2486f3deb89691c"
+LID_MAP = {"org_jpn": "jpn", "lga": "lug", "ory": "ori", "arb": "ara"}
+
+
+def normalize_lid(split, uttid, lid):
+    """Follow egs2/ml_superb2/asr1/local/download.py, including dialect labels."""
```
Masao-Someki: ditto

egs3/voxlingua107/esp2_lid/src/prepare_utils.py
Masao-Someki, line 1: ditto

egs3/voxlingua107/esp2_lid/src/voxpopuli_prepare.py
```diff
@@ -0,0 +1,147 @@
...
+from pathlib import Path
+
+from egs3.voxlingua107.esp2_lid.dataset.builder import _ISO3_CODES
+from egs3.voxlingua107.esp2_lid.src.prepare_utils import parser, write_manifest
+
+BASE_URL = "https://dl.fbaipublicfiles.com/voxpopuli"
+LANGUAGES = "en de fr es pl it ro hu cs nl fi hr sk sl et lt".split()
+
+
+def download(url, path):
```
Masao-Someki: Let's use `espnet3.utils.download_utils`

egs3/voxlingua107/esp2_lid/src/voxpopuli_prepare.py
```diff
@@ -0,0 +1,147 @@
...
+    """Retain completed downloads and replace only after a successful transfer."""
```
Masao-Someki: Please add Args, and Example

egs3/voxlingua107/esp2_lid/src/voxpopuli_prepare.py
```diff
@@ -0,0 +1,147 @@
...
+    path.parent.mkdir(parents=True, exist_ok=True)
+    if not path.exists():
+        temporary = path.with_suffix(path.suffix + ".part")
+        with urllib.request.urlopen(url) as source, temporary.open("wb") as output:
+            shutil.copyfileobj(source, output)
+        temporary.replace(path)
+
+
+def annotations(path, split, limit=None):
+    """Select ASR utterances as in ESPnet2 OWSM's prepare_voxpopuli.py."""
```
Masao-Someki: ditto

egs3/voxlingua107/esp2_lid/src/voxpopuli_prepare.py
```diff
@@ -0,0 +1,147 @@
...
+                break
+    return selected
+
+
+def download_recordings(source_dir, sessions, base_url=BASE_URL):
+    """Fetch required members of official yearly tar files using HTTP ranges.
+
+    Archive seeking skips unrelated recordings. Completed recordings are reused;
+    interrupted members are retried. No archive paths are extracted unchecked.
+    """
```
Masao-Someki: ditto

egs3/voxlingua107/esp2_lid/src/voxpopuli_prepare.py
```diff
@@ -0,0 +1,147 @@
...
+    argparser.add_argument(
+        "--splits", nargs="+", choices=["train", "dev", "test"], default=["test"]
+    )
+    argparser.add_argument(
+        "--skip-download-audio",
+        action="store_true",
+        help="Use existing raw_audios/original/<year>/*.ogg",
+    )
+    args = argparser.parse_args()
+    for split in args.splits:
```
Masao-Someki, line 176: I think you can write runner/provider to parallelize.
Good part of runner/provider is that you can both parallelize in local job or HPC environment, it will automatically submit jobs to parallelize your runner implementation

espnet3/components/callbacks/default_callbacks.py
Masao-Someki, line 1: (Just a comment) Thank you!! I didn't know that behavior and it seems we need this bugfix!!

espnet3/systems/esp2_lid/metrics/accuracy.py
Masao-Someki, line 1: I am not sure if we can just call it `Accuracy` as metrics..
I think we can think of splitting this to multiple metrics

espnet3/systems/esp2_lid/callbacks.py
```diff
@@ -0,0 +1,43 @@
...
+import torch
+from lightning.pytorch.callbacks import Callback
+
+
+class ESPnet2MatmulPrecision(Callback):
+    """Keep FP32 matrix multiplication consistent with ESPnet2 use_tf32=false."""
+
+    def on_fit_start(self, trainer, pl_module):
+        """Override the common ESPnet3 training entrypoint's TF32 setting."""
+        torch.set_float32_matmul_precision("highest")
```
Masao-Someki: I am not sure how this affects to training performance, is this important for Language Identification task?

espnet3/systems/esp2_lid/callbacks.py
```diff
@@ -0,0 +1,43 @@
...
+
+
+class BestCheckpointLink(Callback):
+    """Expose the best individual checkpoint alongside the common top-K average."""
```
Masao-Someki: Let's add more details on what this callback is doing, and what file we're expecting to get after processing this callback class!

espnet3/systems/esp2_lid/collect_stats.py
```diff
@@ -0,0 +1,103 @@
...
+        language = sample["lid_labels"]
+        if not isinstance(language, str) or not language or len(language.split()) != 1:
+            raise ValueError(
+                "LID statistics require a single language string per sample"
+            )
+        categories[language].append(str(index))
+        yield [str(index)], {"speech": sample["speech"].unsqueeze(0)}
+
+
+def collect_speech_shapes(config) -> None:
```
Masao-Someki: How about using `espnet3.parallel` here, you can refer to the espnet3's collect_stats, and refer to CollectStatsRunner/CollectStatsProvider and implement LIDCollectStatsRunner/LIDCollectStatsProvider!

espnet3/systems/esp2_lid/inference.py
```diff
@@ -0,0 +1,116 @@
...
+
+    @staticmethod
+    def _load_languages(lang2utt: str | Path) -> list[str]:
+        return [
+            line.split(maxsplit=1)[0]
+            for line in Path(lang2utt).read_text(encoding="utf-8").splitlines()
+            if line.strip()
+        ]
+
+    def _prepare_speech(self, speech) -> tuple[torch.Tensor, torch.Tensor, bool]:
```
Masao-Someki, line 60: Even for the private function let's add at least one line of comment, so that we can easily understand how this function are supposed to be used in this codebase!

espnet3/systems/esp2_lid/inference.py
```diff
@@ -0,0 +1,116 @@
...
+        Args:
+            speech: One one-dimensional waveform, or a list of waveforms.
+
+        Returns:
+            One language string when ``extract_embd=False`` (default), or a
+            dictionary containing ``hyp`` and a CPU NumPy ``embedding`` when
+            true. Batched input returns a list of the corresponding values.
+
+        Raises:
+            ValueError: If a waveform is empty or not one-dimensional.
```
Masao-Someki, line 100: Let's add an Example block!

test/egs3/voxlingua107/esp2_lid/test_source_download.py
Masao-Someki, line 1: I think you don't have to test if we can download the dataset, if the url is correct (which I checked by accessing the URL) it is too much I think

Masao-Someki (commented):
Thank you @Shunsuke32!!
I leave comment, so please take a look and fix them! Some of them are related to parallelization, and I think there's a lot of part that we can use `espnet3.parallel` package to efficiently parallelize the process!

Masao-Someki:
Also maybe in the next PR, could you create an integration test with `mini_an4` recipe?

egs3/TEMPLATE/esp2_lid/conf/inference.yaml
Shunsuke32: I copied the ASR inference defaults into the LID template and removed the relative `defaults` entry.

egs3/TEMPLATE/esp2_lid/conf/metrics.yaml
Shunsuke32: I copied the metric defaults into the LID template and removed the relative `defaults` entry.

egs3/TEMPLATE/esp2_lid/conf/publication.yaml
Shunsuke32: I copied the publication defaults and model-card template into the LID template and removed the ASR dependency.

egs3/voxlingua107/esp2_lid/dataset/config.yaml
Shunsuke32: I changed this to `source_env_var: VOXLINGUA107`. As in the LibriSpeech builder, the builder now looks for the corpus in `source_dir`, then `VOXLINGUA107`, then `download/voxlingua107`, so the configs no longer read the environment variable.

egs3/voxlingua107/esp2_lid/dataset/__init__.py
Shunsuke32: I changed these to full module imports.

egs3/voxlingua107/esp2_lid/conf/inference_embeddings.yaml
Shunsuke32: I removed `defaults` and put the embedding inference settings in this file.

egs3/voxlingua107/esp2_lid/conf/inference_prepared.yaml
Shunsuke32: I removed `defaults` and put the prepared-data inference settings in this file.

egs3/voxlingua107/esp2_lid/conf/metrics_embeddings.yaml
Shunsuke32: I removed `defaults` and put the embedding metric settings in this file.

egs3/voxlingua107/esp2_lid/dataset/builder.py
Shunsuke32: I now use `espnet3.utils.download_utils.download_url` for fresh downloads. I kept wget for resuming partial files and unzip for ZIP extraction.

egs3/voxlingua107/esp2_lid/dataset/builder.py
Shunsuke32: Yes, the corpus has multiple ZIP files. I added `DownloadProvider` and `DownloadRunner` to download and extract them through `espnet3.parallel`.

egs3/voxlingua107/esp2_lid/dataset/builder.py
Shunsuke32: I removed the source preparation call from `build()`. The common `create_dataset()` stage handles it first.

egs3/voxlingua107/esp2_lid/dataset/builder.py
Shunsuke32: I added examples of `utt2lang`, `lang2utt`, and `category2utt`, with a short explanation of how they are used.

egs3/voxlingua107/esp2_lid/dataset/dataset.py
Shunsuke32: I added an `Args` section to the Dataset constructor.

egs3/voxlingua107/esp2_lid/src/babel_prepare.py
Shunsuke32: Thanks, I kept the preparation script for now. It can move to a separate Babel recipe when one is available.

egs3/voxlingua107/esp2_lid/src/babel_prepare.py
Shunsuke32: I added arguments, yielded values, and an example to the Babel preparation function.

egs3/voxlingua107/esp2_lid/src/fleurs_prepare.py
Shunsuke32: I added arguments, the return value, and an example to the FLEURS language mapping function.

egs3/voxlingua107/esp2_lid/src/ml_superb2_prepare.py
Shunsuke32: I added arguments, the return value, and an example to the ML-SUPERB language mapping function.

egs3/voxlingua107/esp2_lid/src/prepare_utils.py
Shunsuke32: I added arguments, return values, and examples to the public preparation helpers.

egs3/voxlingua107/esp2_lid/src/voxpopuli_prepare.py
Shunsuke32: I removed the custom download function. Annotation downloads now use the common download helper through `DownloadRunner`.

egs3/voxlingua107/esp2_lid/src/voxpopuli_prepare.py
Shunsuke32: I removed this function as part of using the common download helper. The replacement runner has a usage example.

egs3/voxlingua107/esp2_lid/src/voxpopuli_prepare.py
Shunsuke32: I added arguments, the return value, and an example to `annotations()`.

egs3/voxlingua107/esp2_lid/src/voxpopuli_prepare.py
Shunsuke32: I added arguments and an example to `download_recordings()`.

egs3/voxlingua107/esp2_lid/src/voxpopuli_prepare.py
Shunsuke32: I moved annotation downloads, recording downloads, and audio preparation to runners and providers. `--parallel-config` selects the existing ESPnet3 execution backend.

espnet3/components/callbacks/default_callbacks.py
Shunsuke32: Thanks for confirming. I kept the final-checkpoint averaging fix.

espnet3/systems/esp2_lid/metrics/accuracy.py
Shunsuke32: I split this into `Accuracy` and `ClassificationReport`. Accuracy now returns only accuracy metrics; the report handles precision, recall, F1, per-language scores, and errors.

espnet3/systems/esp2_lid/callbacks.py
Shunsuke32: I removed the override so the recipe now uses ESPnet3's shared `high` setting. With the same checkpoint, `high` and `highest` gave identical predictions on all 1,609 VoxLingua107 dev utterances (93.78% accuracy). I have not compared training with the two settings.

espnet3/systems/esp2_lid/callbacks.py
Shunsuke32: I documented the link name, checkpoint target, and update timing. It is updated on rank zero after each epoch and at training end.

espnet3/systems/esp2_lid/collect_stats.py
Shunsuke32: I added `LIDCollectStatsProvider` and `LIDCollectStatsRunner` using the existing parallel framework. They preserve global Dataset IDs when merging shapes and language mappings.

espnet3/systems/esp2_lid/inference.py
Shunsuke32: I added a short docstring to `_prepare_speech()`.

espnet3/systems/esp2_lid/inference.py
Shunsuke32: I added single-utterance and batch examples to `__call__()`.

test/egs3/voxlingua107/esp2_lid/test_source_download.py
Shunsuke32: Agreed. I removed the download test file.

Shunsuke32:
Thank you for the review! I've addressed each of your comments. I'll also add a `mini_an4` integration test in the next PR.

PR #6743, opened 2026-09-20
Ask for what each Space actually needs to start

sw005320:
Two of the four Spaces uploaded tonight came up `RUNTIME_ERROR`. Both are
`requirements.txt` not describing the app above it, and neither is visible until
the Space tries to run — the build succeeds either way.

| Space | stage | why |
|---|---|---|
| `espnet/ljspeech-vits` | RUNNING | |
| `espnet/universal-se` | RUNNING, `zero-a10g` | |
| `espnet/speaker-verification` | **RUNTIME_ERROR** | missing `espnet[spk]` |
| `espnet/owsm-ctc-v4` | **RUNTIME_ERROR** | app needs an unreleased API |

### The speaker Space

#6723 moved the default to RawNet3, whose front-end is `asteroid_frontend`,
which imports `asteroid_filterbanks` — and that comes from `espnet[spk]`. The
file still asked for plain `espnet`.

It worked when I tested it locally only because my environment happened to have
the package. A Space installs from `requirements.txt` and nothing else, which is
the whole point of that file.

The README had **predicted this exact failure** a day earlier:

> A RawNet3 checkpoint reached through `SPK_MODEL_TAG` would want `espnet[spk]`,
> for `asteroid_filterbanks`.

and the sentence was not read when the default changed. It now describes what
the file does rather than what it might one day need.

### The OWSM-CTC Space

#6728 moved the app to `best_path()` and to the `decode_long()` that takes a
path and a batch size. Neither exists before 202612; a Space installs espnet
from PyPI, so it built against 202610 and fell over on the first decode. The
released `Speech2Text` has no `best_path` at all, and its `decode_long` takes
neither `batch_size` nor `context_len_in_secs`.

`espnet2/bin/demo.py` warns about precisely this in its own docstring. I
recommended the re-upload anyway, and it replaced a working Space with a broken
one.

`requirements.txt` now asks for `espnet>=202612`, which pip refuses until that
release exists — so the Space says "not yet" rather than running against
something older. The README's publish section says so too.

### So it does not happen a third time

Three tests, since a comment did not hold the first time:

- every demo names a lowest espnet it works with;
- an app calling a name no release has yet asks for the release that has it;
- the task demos ask for their extra.

Reverting either `requirements.txt` fails them:

```
FAILED test_an_app_using_a_new_api_asks_for_the_release_that_has_it[ctc]
FAILED test_the_task_demos_ask_for_their_extra[spk-spk]
```

### Still to do by hand

`espnet/owsm-ctc-v4` is live and broken right now. It has to be put back to the
version before tonight's upload, and uploaded again after 202612 ships.


sw005320:
Fixed. The comparison was lexical, and the finding is right in two ways:

```
202612rc1     >= 202612          lexical=True   pep440=False
202610.post2  >= 202610.post10   lexical=True   pep440=False
```

So a `requirements.txt` pinned to a release candidate would have satisfied the
check for an API that candidate does not carry — and `.post2` read as newer
than `.post10`, which matters now that the next release is a `.postN` at all.

`packaging.Version` now, which espnet already depends on (`pyproject.toml`
line 44). Checked by pinning the CTC demo to `espnet>=202610.post1rc1`: the
test fails, as it should. 68 pass with the real files.

PR #6744, opened 2026-09-20
A patch release is .postN, which is what PyPI accepts

sw005320:
Cutting 202610.post1 stopped here:

```
$ python doc/make_release.py <token> v.202610.post1
milestone 'v.202610.post1' does not look like a release: expected v.YYYYMM, optionally with -patchN
```

The script takes the milestone title as the version, and then required it to be
`v.YYYYMM` or `v.YYYYMM-patchN`. The second cannot work: `version.txt` is what
the package is published as, and **`202610-patch1` is not a version PyPI will
take** — PEP 440 spells it `202610.post1`. So the documented path for a patch
release rejected the only name that works and accepted one that dies at the
upload.

It has never been walked. The tags before 202610 say `-patchN` while their
`version.txt` says `.postN`, and off by one at that:

| tag | version.txt → PyPI |
|---|---|
| `v.202609` | `202609` |
| `v.202609-patch1` | `202609` ← never bumped |
| `v.202609-patch2` | `202609.post1` |
| `v.202609-patch3` | `202609.post2` |

Someone hand-wrote the PEP 440 form each time, and the first one was missed —
which is exactly the tag-versus-published drift this script was written to stop,
sitting inside the script.

The rule now accepts `.postN`. It also moved out of `main()` into
`release_version()`, with a test, because a rule that decides what gets
published should be checkable: `v.202610.post1` is a release, `v.202610-patch1`
is not, and neither is `v.202610.dev1` or a milestone called `Backlog`.

Both `-patchN` spellings stay in history; nothing renames the old tags.


sw005320:
Fixed, and it is a real one here rather than a theoretical one.

`\d` matches every Unicode decimal digit, so the check accepted titles PyPI
will not:

```python
>>> re.fullmatch(r"\d{6}", "２０２６１０")     # fullwidth
<re.Match ...>
>>> re.fullmatch(r"\d{6}", "٢٠٢٦١٠")     # Arabic-Indic
<re.Match ...>
>>> re.fullmatch(r"[0-9]{6}", "２０２６１０")
None
```

That is exactly the failure this check exists to catch before a tag is cut —
a version that looks right, passes, and dies at the upload — and fullwidth
digits are a keystroke away on the keyboard this project's milestones get typed
on.

`[0-9]` now, in both halves of the pattern, with `v.２０２６１０` and
`v.202610.post１` added to the rejected cases. 14 tests pass.

PR #6745, opened 2026-09-20
Add ESPnet3 AN4 Sinc-BLSTMP ASR recipe

CW-JD:
## What did you change?

Port the AN4 Sinc-BLSTMP ASR recipe to the existing ESPnet3 ASRSystem, public data loader, statistics collection, Lightning training, checkpoint callbacks, inference runner and WER/CER/TER metrics. Dataset preparation and output formatting stay inside the recipe; shared framework code is unchanged.

LM training uses the existing ESPnet2 commands because the current ESPnet3 system has no dedicated LM training stage. LM fusion is retained. Public recipe functions include Args, Returns and Examples docstrings.

---

Follow ESPnet3's existing abstractions and modules, as requested in review. Remove the earlier compatibility adapters and numerical-alignment behavior, and validate the native recipe using GPU CTC and standard pre-training validation.

---

## Is your PR small enough?

No. The current diff contains 29 files, 1606 insertions and 0 deletions, covering the recipe, documentation and mirrored tests. This exceeds the template's size limits. It contains no changes to shared ESPnet3, ESPnet2, TEMPLATE or CI files.

---

## Additional Context

- Recipe tests: 20 passed. Black, isort, pycodestyle and flake8, including Google-style recipe docstrings, passed.
- Both compared ASR runs use native GPU CTC, the same tokenizer and an existing native ESPnet2 LM. LM training was not repeated for these ASR runs.
- The ESPnet3 run used one A10, a 25-epoch maximum and standard early stopping, finishing at 21 epochs. The local native ESPnet2 Sinc-RNN baseline completed 25 epochs. All 100 validation and 130 test recordings were evaluated.
- ESPnet3's public-metric test WER/CER/TER are 8.93/5.03/4.86%. The table below separately rescores its existing predictions with SCTK for a common scoring convention.

These are local native baselines, not published official results. Common ESPnet2 tokenization + SCTK scores (%):

| Split | Framework | WER | CER | TER |
| Validation | ESPnet2 | 14.72 | 9.92 | 9.43 |
| Validation | ESPnet3 (SCTK rescored) | 15.06 | 8.98 | 8.54 |
| Test | ESPnet2 | 7.63 | 4.33 | 4.12 |
| Test | ESPnet3 (SCTK rescored) | 8.93 | 5.11 | 4.86 |

The recipe itself uses the public ESPnet3 metrics. Framework-level differences are documented in the README; no bitwise parity is claimed. Similar error rates alone do not prove every migration detail correct.


CW-JD:
ESPnet2 vs. ESPnet3 comparison: 25 ASR epochs, one A10 per run, seed 0, FP32, and CPU CTCLoss. Both runs used the same prepared audio, tokenizer, normalization statistics, and native LM checkpoint (`lm_weight=0.1`). Inference selected the best validation-accuracy checkpoint.

| Split | Utterances | Metric | ESPnet2 | ESPnet3 |
|---|---:|---|---:|---:|
| Validation | 100 | WER (%) | 14.38 | 14.38 |
| Validation | 100 | CER (%) | 8.72 | 8.72 |
| Validation | 100 | TER (%) | 8.29 | 8.29 |
| Test | 130 | WER (%) | 8.28 | 8.28 |
| Test | 130 | CER (%) | 4.52 | 4.52 |
| Test | 130 | TER (%) | 4.30 | 4.30 |

All 100 final ASR checkpoint state tensors were bitwise identical. SCTK integer scoring counts matched for all three metrics on both splits. Hypotheses matched after whitespace normalization; one test hypothesis differed only by a trailing space.

This validates single-GPU ASR parity under the stated conditions. The LM checkpoint was shared, so this is not an independent comparison of two LM training runs. CPU CTCLoss was an experiment override.

egs3/an4/esp2_asr/src/inference.py
Masao-Someki, line 1: Basically could you add little more details on the docstrings? We at least want to have args, returns, examples on each public functions

espnet3/components/data/dataloader.py
Masao-Someki, line 1: Do you have any reason to change this part?

Masao-Someki (commented):
Sorry, I reviewed the code but stopped partway through.. Could you try to reuse more of the existing ESPnet3 modules? We basically don't allow people to touch `espnet3/components/` when creating recipe.

The current codebase looks largely like a copy of ESPnet2, with some workarounds added to make ESPnet2 components run inside ESPnet3. I think it would be better to rely more directly on the ESPnet3 abstractions and modules where possible.

I think some differences in loss/accuracy or WER/CER compared with ESPnet2 are acceptable, since ESPnet3 uses a different trainer and may also use different library and Python versions. If you want to do the exact reproduction, you can create a separate system and keep the ESPnet2-based implementation there instead.

PR #6746, opened 2026-09-20
Add ESPnet3 VOiCES Conformer ASR recipe

CW-JD:
## What did you change?

Port the VOiCES Conformer ASR recipe to the existing ESPnet3 ASRSystem, public data loader, statistics collection, Lightning training, checkpoint callbacks, inference runner and WER/CER/TER metrics. Dataset preparation and output formatting stay inside the recipe; shared framework code is unchanged.

LM training uses the existing ESPnet2 commands because the current ESPnet3 system has no dedicated LM training stage. LM fusion is retained. Public recipe functions include Args, Returns and Examples docstrings.

---

Follow ESPnet3's existing abstractions and modules, as requested in review. Remove the earlier compatibility adapters and numerical-alignment behavior, and validate the native recipe using GPU CTC and standard pre-training validation.

---

## Is your PR small enough?

No. The current diff contains 33 files, 1913 insertions and 0 deletions, covering the recipe, documentation and mirrored tests. This exceeds the template's size limits. It contains no changes to shared ESPnet3, ESPnet2, TEMPLATE or CI files.

---

## Additional Context

- Recipe tests: 36 passed. Black, isort, pycodestyle and flake8, including Google-style recipe docstrings, passed.
- Both compared ASR runs use native GPU CTC, the same tokenizer and an existing native ESPnet2 LM. LM training was not repeated for these ASR runs.
- Both runs completed 50 ASR epochs on the official single-channel devkit, with 12507 training, 660 validation and 6600 test recordings and a 1000-piece vocabulary. The reused LM is devkit-only; full-corpus results and the external-text LM budget are not claimed.
- ESPnet3 used four A10 GPUs, BF16 and 2000000 frontend-feature elements per GPU. Native ESPnet2 used two A10 GPUs, FP16 AMP and an explicit raw-speech/text batch_bins=17500000 override to fit 24 GB GPUs. Accumulation remains 4; native batch distribution and update counts differ.
- After host-memory exhaustion at final ESPnet2 checkpoint averaging, the same official 10-best averaging function was rerun in a fresh CPU process using the saved reporter. Training checkpoint hashes were unchanged; complete GPU inference and scoring then passed.
- ESPnet3's public-metric test WER/CER/TER are 93.40/71.79/91.79%. The table below separately rescores its existing predictions with SCTK for a common scoring convention.

These are local native baselines, not published official results. Common ESPnet2 tokenization + SCTK scores (%):

| Split | Framework | WER | CER | TER |
| Validation | ESPnet2 | 91.01 | 69.29 | 91.33 |
| Validation | ESPnet3 (SCTK rescored) | 92.90 | 71.16 | 91.83 |
| Test | ESPnet2 | 92.35 | 70.37 | 91.54 |
| Test | ESPnet3 (SCTK rescored) | 93.51 | 72.01 | 91.90 |

The recipe itself uses the public ESPnet3 metrics. Framework-level differences are documented in the README; no bitwise parity is claimed. Similar error rates alone do not prove every migration detail correct.


CW-JD:
ESPnet2 vs. ESPnet3 validation used two separate experiments:

- A: One-epoch VOiCES devkit training comparison on four A10 GPUs, using a 1000-piece tokenizer, CPU CTCLoss, and strict deterministic settings in both frameworks.
- B: A separate inference smoke test on one RTX 4090, using earlier one-epoch checkpoints, four validation and four test recordings, beam size 20, and a shared devkit LM with `lm_weight=0.6`.

| Experiment | Check / metric | ESPnet2 | ESPnet3 |
|---|---|---:|---:|
| A: Strict training | Training batches per rank | 117 | 117 |
| A: Strict training | Optimizer updates per rank | 29 | 29 |
| A: Strict training | Validation batches per rank | 7 | 7 |
| A: Strict training | Validation loss | 726.6682113185074 | 726.6682113185074 |
| A: Strict training | Final checkpoint state tensors compared | 654 | 654 |
| B: Validation smoke test, 4 utterances | WER (%) | 756.04 | 756.04 |
| B: Validation smoke test, 4 utterances | CER (%) | 543.23 | 543.23 |
| B: Validation smoke test, 4 utterances | TER (%) | 483.77 | 483.77 |
| B: Test smoke test, 4 utterances | WER (%) | 986.23 | 986.23 |
| B: Test smoke test, 4 utterances | CER (%) | 744.62 | 744.62 |
| B: Test smoke test, 4 utterances | TER (%) | 498.96 | 498.96 |

In experiment A, inputs, RNG states, losses, gradients, parameter updates, and validation records matched across all four ranks. All 654 final checkpoint state tensors were bitwise identical, and all seven validation aggregates matched exactly.

In experiment B, hypotheses and SCTK integer scoring counts matched. The one-epoch smoke results contain many insertion errors, so error rates exceed 100%; these are pipeline checks, not final recognition-accuracy results.

The tokenizer and determinism settings above are experiment overrides, separate from the delivered source defaults. The repaired 50-epoch run, full-test-set inference, and full-release accuracy have not been revalidated.

Masao-Someki:
I have one question.. the validation loss is exactly the same, and I am not sure how it happens..

ci/test_python_espnet3.sh
Masao-Someki, line 1: (This is just a message) Thanks!! Yes we should have put this from the beginning

Masao-Someki (commented):
Hi thanks @CW-JD , could you also fix this PR regarding the comment on [6745](https://github.com/espnet/espnet/pull/6745)?

CW-JD:
> Hi thanks @CW-JD , could you also fix this PR regarding the comment on [6745](https://github.com/espnet/espnet/pull/6745)?

Thank you, @Masao-Someki. I’ll revise both PRs as soon as possible.

CW-JD:
> I have one question.. the validation loss is exactly the same, and I am not sure how it happens..

Thanks for raising this question. The identical validation loss came from a controlled, one-epoch VOiCES devkit comparison on four A10 GPUs. I aligned the initial model parameters, batch composition and order, tokenizer, normalization statistics, and RNG states. I also used CPU CTCLoss and enabled strict deterministic execution in both runs.

Additionally, I introduced compatibility code to match ESPnet2’s optimization behavior, DDP buffer synchronization, and validation-metric aggregation within the ESPnet3 pipeline. In this experiment, the independently recorded training losses, gradients, and parameter updates matched. All 654 final checkpoint tensors and all seven validation aggregates were identical, which explains the identical validation loss.

I understand your concern about adding compatibility code to shared components for exact reproduction. I’ll revise both recipes to use the existing ESPnet3 abstractions and modules, keep this controlled parity experiment separate, and report new results after revalidation.

Masao-Someki:
Thank you @CW-JD, now I understand why we got the same loss!

> I understand your concern about adding compatibility code to shared components for exact reproduction. I’ll revise both recipes to use the existing ESPnet3 abstractions and modules, keep this controlled parity experiment separate, and report new results after revalidation.

Thanks!

CW-JD:
Hi @Masao-Someki, could you please help check the remaining CI failures in #6745 and #6746?

I reviewed the logs after all checks completed. The failures appear to be outside the recipe changes:

- **Both PRs — ESPnet2 ASR configuration checks:** The existing STOP/WavLM recipes downloaded only 107 bytes for the WavLM checkpoint, then failed with `Unsupported operand 60`. An earlier successful master run downloaded approximately 1.18 GB from the same URL.
  [Example log](https://github.com/espnet/espnet/actions/runs/35691812876/job/106630317043)

- **#6745 — ESPnet2 Python tests:** `test_Speech2Text_hugging_face` exceeded its 30-second timeout while downloading a Hugging Face tokenizer.
  [Log](https://github.com/espnet/espnet/actions/runs/35691812876/job/106630317232)

- **#6746 — ESPnet2 Python tests:** `test_SizedDict_shared` exceeded its 5-second timeout while waiting for a child process at `p.join()`.
  [Log](https://github.com/espnet/espnet/actions/runs/35691820546/job/106630326909)

- **#6746 — Debian tests:** Six existing JETS/VITS tests failed while loading SpeechMOS through `torch.hub`. The traceback shows `HTTP Error 403: rate limit exceeded`, followed by `KeyError: 'Authorization'`.
  [Log](https://github.com/espnet/espnet/actions/runs/35691820614/job/106630246593)

The failed CI gate checks summarize these failures. The dedicated ESPnet3 checks passed, and neither PR modifies the affected ESPnet2 code or tests.

Could you help confirm whether these are CI/dependency issues and rerun the failed jobs? I don’t have permission to rerun checks in the upstream repository. Please let me know if further investigation is needed on my side.

Thank you!

Masao-Someki:
Hi, thanks @CW-JD! I just noticed that ESPnet3’s performance seems to be consistently worse than ESPnet2’s..
I’ll take a closer look on espnet3 implementation, but do you have any idea what might be causing the difference?

CW-JD:
Thanks for looking into this, @Masao-Someki. I don’t yet have a confirmed root cause, but there are several differences in my experiments that could contribute.

Using the same ESPnet2 tokenization and SCTK scoring for both sets of hypotheses, the test WERs are:

| Recipe | ESPnet2 | ESPnet3 |
|---|---:|---:|
| AN4 | 7.63% | 8.93% |
| VOiCES devkit | 92.35% | 93.51% |

So the WER gap remains after using a common scorer. Both current runs use native GPU CTC.

The main differences I would investigate are:

1. **VOiCES batching and optimizer updates.** ESPnet2 used two A10 GPUs with a 17.5M raw-speech/text batch budget, reduced for memory constraints. ESPnet3 used four A10 GPUs with a 2M frontend-feature budget per GPU. These produce different batch compositions and update counts despite both running for 50 epochs. ESPnet3 completed only 2,400 optimizer updates with the original 40,000-step warmup, so the relationship between batch size, update count and learning-rate progression seems particularly important to check.

2. **Precision.** The VOiCES ESPnet2 baseline used native FP16 AMP, while ESPnet3 used BF16 mixed precision. I have not isolated the effect of this difference.

3. **AN4 training duration and batching.** ESPnet2 completed 25 epochs, while ESPnet3 stopped after 21 through its configured early stopping. ESPnet3 also batches by frontend-feature lengths rather than the original speech/text shapes. Validation aggregation and checkpoint selection/averaging would be additional points to inspect.

These are single-seed runs, and VOiCES was evaluated only on the devkit. I haven’t run ablations to establish which differences explain the gap, so I would not yet attribute it to a particular ESPnet3 implementation issue.

I can provide the resolved configs and training logs to help investigate. My first priority would be comparing effective batch sizes, optimizer-step/LR histories and the checkpoints actually used for decoding.

egs3/voices/asr/conf/inference.yaml
Masao-Someki, line 1: Could you add comments on the config file? We want more document-like configs, something like we have in the librispeech-100h recipe!

egs3/voices/asr/conf/language_model.yaml
Masao-Someki, line 1: ditto

egs3/voices/asr/conf/lm_native.yaml
Masao-Someki, line 1: ditto

egs3/voices/asr/conf/metrics.yaml
Masao-Someki, line 1: ditto

egs3/voices/esp2_asr/conf/training_conformer.yaml
Masao-Someki, line 1: ditto

egs3/voices/asr/conf/training_devkit.yaml
```diff
@@ -0,0 +1,6 @@
+# Original model/training parameters with the source recipe's devkit mode.
+defaults:
+  - training_conformer
+  - _self_
```
Masao-Someki: Could you copy asr's inference config here? We want to keep these configs self-consistent, and relative path to identify config inside library is easily cause a bug in my experience..

egs3/voices/esp2_asr/dataset/builder.py
```diff
@@ -0,0 +1,418 @@
...
+
+        Returns:
+            None. Completion is recorded by the files written to disk.
+
+        Examples:
+            After prepare_source has completed:
+
+            >>> builder = VoicesBuilder()
+            >>> builder.build(recipe_dir=".")
+        """
```
Masao-Someki, line 344: Generally nice, but I think it's much better with some comments. Could you add oneline comments on this function?

egs3/voices/asr/dataset/dataset.py
```diff
@@ -0,0 +1,108 @@
+"""Read indexed VOiCES audio lazily without changing source files."""
+
+from pathlib import Path
+
+import soundfile as sf
+from torch.utils.data import Dataset as TorchDataset
+
+from .builder import RECIPE_ROOT, SPLITS, read_manifest
```
Masao-Someki: Please use full module path, so that it would be `egs3.voices.esp2_asr.dataset.builder`

egs3/voices/asr/src/language_model.py
Masao-Someki, line 1: Could you simply create `systems/esp2_asr/lm_system.py` and copy LMTask's task.py, and use it to train the language model?

egs3/voices/esp2_asr/run.py
Masao-Someki, line 1: Could you change the `asr` system into `esp2_asr`?

test/egs3/voices/asr/conf/test_training_conformer.py
```diff
@@ -0,0 +1,76 @@
...
+import pytest
+import torch
+import yaml
+from hydra.utils import instantiate
+from omegaconf import OmegaConf
+
+from espnet3.utils.config_utils import load_and_merge_config
+
+ROOT = Path(__file__).resolve().parents[5]
+RECIPE = ROOT / "egs3/voices/asr"
```
Masao-Someki: This one also please change it to `esp2_asr`

egs3/voices/asr/conf/lm_native.yaml
Masao-Someki: But for this model, since we want to copy lm task to `esp2_asr/` and create `lm_system.py`, let's convert to espnet3 style!

Masao-Someki (commented):
Thank you @CW-JD!
Currently I am quickly preparing a PR to change `asr` system into `esp2_asr`, so could you also rename the system name from `asr` to `esp2_asr`?

Also let's create `systems/esp2_asr/lm_system.py` and copy espnet2's lm task file to run training with the lm system!

CW-JD:
Thanks, @Masao-Someki. I've updated both PRs following your feedback:

- Renamed the recipe/test paths and imports to `esp2_asr`.
- Added configuration explanations, builder comments, and public-function docstrings.
- Made the VOiCES devkit training config self-contained.
- Replaced the ESPnet2 LM training wrapper with `LMSystem`, using the copied `LMTask` and the shared ESPnet3 trainer.

The local regression tests and single-A10 LM training and fusion smoke tests passed. Full-budget and multi-GPU training with the new LMSystem have not yet been validated. The previous ASR result tables are explicitly labeled as historical results using an ESPnet2-trained LM.

Both branches include the upstream rename from #6795. Could you take another look when you have time? Thank you!

Masao-Someki:
Thanks!
We'll quickly merge the PR for `asr->esp2_asr`, and will check in detail after that.. currently this PR has so many file changes and I cannot review rn..

Issue #57, opened 2018-01-05
Automatic doc build & deploy based on Travis-CI and travis-sphinx


ShigekiKarita:
Manual deploy is troublesome. I've already tested [travis-sphinx](https://github.com/Syntaf/travis-sphinx)  that automates sphinx-doc build & gh-pages deploy with Travis-CI.

- [.travis.yml](https://github.com/ShigekiKarita/espnet/blob/master/.travis.yml#L46)
- [gh-pages](https://shigekikarita.github.io/espnet/apis/nets.html#nets.e2e_asr_attctc.Encoder)
  - I confirmed that my fix of typo `ENCODER NE(W)TWORK CLASS` was deployed automatically
- you can build HTML doc locally by `travis-sphinx build --source=doc --nowarn` and see `doc/build/index.html`

This enables us to add and confirm documentation rapidly.
Do you have any problem to automate this?

kan-bayashi:
Thanks @ShigekiKarita !
It is what I wanted :)

r9y9:
Is this still working now?

ShigekiKarita:
Yes

r9y9:
Ah okay, doc building system is working but API docs are just missing in https://github.com/espnet/espnet/blob/master/doc/apis/nets.rst.

ShigekiKarita:
https://espnet.github.io/espnet/apis/nets.html
this page is listed in the left column. maybe it is nice to add index.rst

ShigekiKarita:
this nets page has not been updated so far... we need to add more (ASR/TTSInterface, Transformer, Frontend, etc)

PR #6761, opened 2026-09-20
Follow espnet/notebook's flat Demos layout

sw005320:
**Merge after [espnet/notebook#56](https://github.com/espnet/notebook/pull/56).**
Before it, these paths do not exist and `ci/check_demo_links.py` fails the other
way round.

That PR flattens `Demos/` — one `<task>_demo.ipynb` per task, with the eight
that were there moved to `Demos/unmaintained/` — so every link here moves with
it or the daily link check goes red.

| was | now |
|---|---|
| `Demos/TTS/tts_realtime_demo.ipynb` | `Demos/tts_demo.ipynb` |
| `Demos/ASR/asr_realtime_demo.ipynb` | `Demos/unmaintained/asr_realtime_demo.ipynb` |
| `Demos/ASR/asr_transfer_learning_demo.ipynb` | `Demos/unmaintained/…` |
| `Demos/ASR/streaming_asr_demo.ipynb` | `Demos/unmaintained/…` |

The TTS links go to the **new** demo rather than to the old notebook's new
address: there is a maintained replacement, so that is where the README should
send people. The label loses "Real-time", because the new notebook is not that,
and keeping it would be a second wrong link in place of the first.

The ASR links now point into `unmaintained/`. That is honest rather than good —
those notebooks do not run. The README advertising them is the next thing to
fix, and it gets fixed by writing `asr_demo.ipynb`, not by editing this file.


sw005320:
Replaced rather than repointed, as you said.

The first pass moved each link to wherever its notebook had gone, which left the
README advertising six notebooks in `unmaintained/`. They do not run. Pointing
at them is worse than the dead links this PR was fixing: a dead link says so,
and a broken notebook wastes an afternoon first.

The Demos table is now the seven that work, each run end to end on
`espnet==202610.post1` before it was committed:

| | |
|---|---|
| recognition, and the words appearing as the audio arrives | `asr_demo`, `asr_streaming_demo` |
| translation — the same model, a different task symbol | `st_demo` |
| text-to-speech, one voice and then 128 | `tts_demo` |
| enhancement, and what it did to the signal-to-noise | `enh_demo` |
| speaker verification — two recordings, one score | `spk_demo` |
| neural codecs — a waveform as a few integers a frame | `codec_demo` |

The two Colab **drive** links for enhancement go too. They pointed at someone's
personal Drive, with no way to tell what they contain now.

### Streaming ASR

You wanted it back — it is back, as
[`asr_streaming_demo.ipynb`](https://github.com/espnet/notebook/pull/57), not as
a revert. Same idea and the same 40 ms slices as the 2021 notebook, running on
the release, printing the hypothesis as it grows:

```
  0.0s  the cell of the hotel
  1.0s  the cell of the hotels is part of holiday strategy
final  the cell of the hotels is part of holidays strategy to sell off assets and concentrate on property and management
```

Its checkpoint is outside the organization (`D-Keqi/...`, TED-LIUM 2). The only
streaming model the organization has is Chinese, on AISHELL, and there is no
Chinese sample here to feed it. Worth deciding, like the SE default was.

### One link stays in `unmaintained/`

The transfer-learning toy example, which has no replacement. Naming it here
rather than leaving it quietly.

### Order

espnet/notebook#55 → #57 → this. The failing checks on the last run were a
runner shutdown mid-job (`The runner has received a shutdown signal`) in an
`enh` integration test, which this PR cannot reach — it touches two README
files. Re-run issued.

PR #6764, opened 2026-09-20
[espnet3] Bugfix - freezing model parameters

Masao-Someki:
## What did you change?

Added `model.freeze_param` support to ESPnet3 training.

- Apply ESPnet2-compatible prefix selectors before optimizer construction.
- Exclude frozen parameters from optimizers.
- Keep `freeze_param` out of model constructor and ESPnet2 task arguments.
- Add template configuration and regression tests.

Selectors are relative to the wrapped model, for example:

```yaml
  model:
    freeze_param:
      - frontend.upstream.model.feature_extractor
```

  ---

`freeze_param` could be specified in ESPnet3 configs but was not applied, so parameters intended to be frozen were still optimized.

  ---

  ## Is your PR small enough?

  yes

  ---

## Additional Context

* #6133


sw005320:
Can you add a test to ensure the same behaviors between espnet2 and espent3?
This will assure this implementation

sw005320 (approved):

PR #6765, opened 2026-09-20
Wait a rate limit out rather than call the demo links broken

sw005320:
The daily demo-link scan went red three times this morning with

```
scan failed: https://huggingface.co/api/spaces/espnet/TTS: HTTP 429
```

Every link in the repository was fine. A runner shares its address with
everyone else on it, so the Hub throttles this job for requests it never made,
and `ci/check_demo_links.py` treats any non-404 as "the scan could not run",
exits 2, and fails the workflow.

### What changed

The response says when the window reopens — `Retry-After`, or the `t` field of
the RateLimit header, which is the spelling the Hub uses:

```
ratelimit: "api";r=0;t=231
ratelimit-policy: "fixed window";"api";q=500;w=300
```

So a 429 or a 503 is waited out once, up to one five-minute window, and only
then reported — with a message that says what it was:

```
HTTP 429 - rate limited and asked for 140s, not a broken link
```

A throttle that names no delay still gets the retry, after 30 seconds. Waiting
a window in a job that otherwise takes half a minute is cheaper than a red
daily run that someone has to open to find out it was throttling.

The job also gets `timeout-minutes: 20`. It had none, so the default was six
hours, and a hung API would have held a runner for an afternoon.

### Checked

- `python3 ci/check_demo_links.py --self-check` — the retry is covered by four
  new cases: the two header spellings, a 429 that clears after the wait it
  named, one that does not, a wait longer than a window (not slept through),
  and a 503 that names no delay. Each was confirmed to fail when the parsing
  is broken on purpose.
- The real scan, on the throttled address: three runs failed, the run with
  this change waited ~130 s and reported `checked 11 notebook links and 3
  Space links`, exit 0. So the links are good as of today, which is the other
  thing this run proves.
- `ci/check_ci_image_config.py` clean, black/isort/flake8 clean.


sw005320:
> Parse HTTP-date `Retry-After` values.

Right, and it is this branch's own failure rather than a new one: a date went
to `float()`, returned `None`, and the retry then waited the 30 second guess.
A reset 31–300 s out would still be throttled when that retry arrived, and the
scan would exit 2 with every link fine — exactly what the PR is for.

`email.utils.parsedate_to_datetime` reads it now. A date with no zone is taken
as UTC, which the header is required to carry anyway, and one in the past is
zero rather than negative. The self-check gained a date two minutes out and one
from 2015; both fail when the parsing is removed, which I checked.

The docstring-coverage warning is the three closures inside `self_check`
(`rate_limited`, `silent_503`, `far_off`) — test fixtures four lines long, in a
function whose own docstring says what they are for. I have left them.

PR #6766, opened 2026-09-20
Add phone recognition (pr) task family and PhoneticXeus inference on TIMIT

Shikhar-S:
## What did you change?

- Added `espnet3/systems/pr`, a phone recognition task family: `PRSystem` plus `PER` and `PFER` metrics. ESPnet had no phone recognition metric of any kind before this.
- Added `egs3/TEMPLATE/pr`, the task template. It has no `train_tokenizer` stage, because a phone recognizer uses a fixed IPA inventory rather than a learned BPE model.
- Added `egs3/timit/pr`, an inference-only recipe that scores PhoneticXeus [1] on TIMIT. It reuses ESPnet's E-Branchformer components -- `CNNFrontend`, `LinearProjection`, `EBranchformerEncoder` and `CTC` -- and loads the published weights into them.
- `pyproject.toml`: a `pr` extra (`panphon`, `jiwer`).

| TIMIT (6300 utterances) | PER | PFER |
|---|---|---|
| this recipe | 42.5 | 13.3 |
| reported [1] | -- | 13.3 |

[1] reports PFER only.

---

To make a published phone recognizer runnable from ESPnet3. The metrics are the reusable part: PER and PFER [2, 3] apply to any model that outputs IPA, and ESPnet could not compute either before this.

---

## Is your PR small enough?

No -- 31 files, about 1815 lines.

---

## Additional Context

[1] [An Empirical Recipe for Universal Phone Recognition](https://arxiv.org/abs/2603.29042)
[2] [POWSM: A Phonetic Open Whisper-Style Speech Foundation Model](https://arxiv.org/abs/2510.24992)
[3] [PRiSM: Benchmarking Phone Realization in Speech Models](https://aclanthology.org/2026.acl-long.825/)


egs3/TEMPLATE/pr/run.py
```diff
@@ -0,0 +1,230 @@
...
+)
+
+# Default stage list (can be extended/overridden by callers)
+DEFAULT_STAGES: List[str] = [
+    "create_dataset",
+    "collect_stats",
+    "train",
+    "infer",
+    "measure",
+]
```
Masao-Someki, line 32: Could you add publication/demo stages ?
Also `publication.yaml` and `demo.yaml`

egs3/timit/pr/src/inference.py
```diff
@@ -0,0 +1,243 @@
...
+    """Shape one inference result into the fields written to SCP files.
+
+    Args:
+        data: The dataset sample, or a list of them when batched.
+        model_output: The model's return value, matching ``data``.
+        idx: The dataset index, or a list of them when batched.
+
+    Returns:
+        A mapping with ``utt_id``, ``hyp`` and ``ref``, or a list of them.
+
```
Masao-Someki, line 230: Could you add an example here? like one example of output (with phonetic sequence) would look like.

egs3/timit/pr/src/inference.py
```diff
@@ -0,0 +1,243 @@
...
+    def forward(self, speech: np.ndarray | torch.Tensor) -> Dict[str, str]:
+        """Transcribe one utterance into IPA phones.
+
+        Args:
+            speech: A 16 kHz mono waveform, shape ``(samples,)``.
+
+        Returns:
+            ``{"hyp": <concatenated IPA>, "tokens": <slash-separated phones>}``.
+            ``hyp`` drops the special tokens and is what the metrics score;
+            ``tokens`` keeps the model's own phone boundaries for inspection.
```
Masao-Someki, line 198: Let's add an example here also, simple code snippet to show initialization, run with one utterance and the output preview.

Masao-Someki (commented):
Just to confirm, is this recipe intended only for inference/evaluation?
Also, is there any particular reason why the training code cannot be published?

Shikhar-S:
@Masao-Someki  Thanks for the review, I will update this later.
We train PhoneticXeus on IPAPACK++ dataset (17K hours) which will be compute heavy. I am thinking about adding fine-tuning on TIMIT in this recipe itself. This will add the training stage for Phone Recognition task too.

Masao-Someki:
Hi, based on our discussion on system naming rules, could you change the system name to `phoneticxeus`?
I think we can add pretraining code in the future PR with ipapack recipe!

PR #6767, opened 2026-09-20
AMI SOT multi-talker ASR: Whisper timestamp rules for S2T, and an ESPnet3 recipe

cyhuang-tw:
# AMI SOT multi-talker ASR: Whisper timestamp rules for S2T, and an ESPnet3 recipe

Serialized Output Training (SOT) transcribes an overlapping multi-talker
utterance as one sequence, with the speakers separated by an inline symbol and
the segments carried by Whisper `<|x.xx|>` timestamps. This adds the decoding
rules that make that sequence well formed, and the `egs3` recipe that trains
and scores such a model on AMI SDM.

The two halves ship together because neither is useful alone: the rules have no
in-tree user without the recipe, and the recipe cannot decode without the rules.

## What this adds

**`espnet2/s2t/whisper_timestamp_scorer.py`** (new) reimplements
`whisper.decoding.ApplyTimestampRules` on ESPnet's scorer interfaces and extends
it for SOT:

- `WhisperTimestampFilter` is a hard `0.0` / `-inf` mask over the vocabulary.
  It enforces the pair rule, the order rule, the initial-timestamp rule and the
  nonzero-length rule, so a segment cannot close at the timestamp it opened at.
  With `speaker_change` set it scopes them to the current speaker block, so a
  new speaker may open anywhere in the window rather than after the previous
  speaker's last timestamp, and it permits the separator only right after the
  timestamp that closed a segment.
- `WhisperTimestampDecoder` / `BatchWhisperTimestampDecoder` add Whisper's last
  rule, which compares the summed timestamp probability with the best text
  token and forces a timestamp when the timestamps win. That rule needs the
  model probabilities, so it wraps the decoder scorer. `wrap_timestamp_decoder`
  picks the class matching the wrapped decoder's interface, because
  `Speech2Text` chooses batch beam search by `isinstance` over its scorers.

**`espnet2/bin/s2t_inference.py`** gains two optional inference arguments,
`--speaker_change_symbol` and `--adaptive_timestamp`. Both default to off and
the code path with neither set is unchanged.

**`espnet2/bin/whisper_export_vocabulary.py`** stops appending a
speaker-change symbol the vocabulary already holds. A symbol that is already
one BPE token, such as `????` at id 25629, used to appear twice; both
`token_list.index` and the tokenizer resolve the first, so the appended row was
unreachable and the model was one row wider than the alphabet it could emit.

**`espnet2/text/whisper_token_id_converter.py`**,
`espnet2/text/whisper_tokenizer.py` and
`espnet2/bin/whisper_export_vocabulary.py` now copy the tokenizer before adding
to it. `whisper.tokenizer.get_tokenizer` is `lru_cache`d, so one HuggingFace
tokenizer is shared by everything in the process, and the `added_tokens_txt`
path grew that shared object in place. What the next caller saw then depended
on who had run first: `get_num_vocabulary_size()` returned 51865 or 51867, and
because `whisper_export_vocabulary` pads with timestamps up to a fixed total,
every leaked token cost one timestamp at the end of the exported list, so
`<|30.00|>` could be missing from it.

This also moves the vocabulary-size assertion in
`test/espnet2/text/test_whisper_token_id_converter.py` from 51867 to 51865.
51865 is the size with nothing leaked in: 50257 base, 107 specials and 1501
timestamps. The old value held only while two earlier tests in
`test/espnet2/bin/` had each pushed one token into the shared tokenizer.

**`egs3/ami/s2t`** is the recipe: `create_dataset`, `collect_stats`, `train`,
`infer`, `measure` and `pack_model`, under `espnet2.tasks.s2t.S2TTask` with
Whisper small. `create_dataset` fetches its Lhotse CutSet manifests and the AMI
SDM audio itself, so nothing has to be staged by hand.

## Results

AMI SDM test, 6127 utterance groups, beam size 5, the checkpoint published at
[`espnet/multi-talker-whisper-small-ami`](https://huggingface.co/espnet/multi-talker-whisper-small-ami):

| Metric | Value |
| --- | --- |
| cpWER | 28.28 % |
| DER (0.25 s collar) | 8.54 % |

Reproduced end to end from this branch: the published `config.yaml` and
`model.pth` decoded through the recipe's own `infer`/`measure` stages give
exactly those numbers.

## Verification

- Recipe suite: 114 passed against a corpus this recipe builds, 114 against a
  corpus prepared before it, and 103 passed / 11 skipped with no environment
  variables set, which is what CI sees.
- `espnet2` suite for the touched modules passes. The remaining failures in
  the development environment are `test_Speech2Text_quantized` and
  `test_Speech2TextGreedy_quantized`, which raise
  `RuntimeError: unknown architecure` from torch quantization on aarch64 and
  fail identically at the merge base there.
- `create_dataset` verified from nothing: with no manifest directory staged, the
  builder fetched the cutsets from the hub, downloaded the AMI dev audio and cut
  all 3746 dev segments from it. The resulting `text`, `wav.scp`, `utt2spk`,
  `spk2utt`, `text.prev` and `text.ctc` are byte-identical to the reference
  build, and 200 sampled segment waveforms are sample-for-sample identical to an
  independently prepared corpus.
- Training runs on the full 52972-utterance train split and learns; the recipe
  decodes its own checkpoint.
- `flake8`, `black` and `isort` are clean over every file this branch changes.

Supersedes #6448, which proposed the same capability as an `egs2` recipe with a
standalone `decode.py`. It can be closed when this merges.


Masao-Someki:
Could you first add a `sot_finetune` system and put your metrics, whisper part, sot-specific data preprocessing utils there? Also we can have `egs3/TEMPLATE/sot_finetune`

Masao-Someki:
Hi, based on our discussion on system naming rules, could you change the system name to `esp2_s2t`?

Masao-Someki:
It seems this branch has conflict, please fix it!

egs3/ami/esp2_s2t/conf/training.yaml
```diff
@@ -0,0 +1,190 @@
...
+  preprocessor_conf:
+    notime_symbol: ${dataset.preprocessor.notime_symbol}
+    first_time_symbol: ${dataset.preprocessor.first_time_symbol}
+    last_time_symbol: ${dataset.preprocessor.last_time_symbol}
+    na_symbol: ${dataset.preprocessor.na_symbol}
+    fs: ${dataset.preprocessor.fs}
+    speech_length: ${dataset.preprocessor.speech_length}
+    speech_resolution: ${dataset.preprocessor.speech_resolution}
+    time_apply_prob: ${dataset.preprocessor.time_apply_prob}
+    text_prev_apply_prob: ${dataset.preprocessor.text_prev_apply_prob}
```
Masao-Someki, line 97: Is this not used for model definition and used only for inference model instantiation?

egs3/ami/esp2_s2t/dataset/builder.py
```diff
@@ -0,0 +1,387 @@
...
+if __package__:
+    # Normal case: this module is part of the real egs3.ami.esp2_s2t.dataset
+    # package, so the relative import resolves against it directly. Gating
+    # on __package__ instead of wrapping this in try/except ImportError
+    # means a genuine breakage in dataset.py (for example _CONFIG or
+    # _split_dir renamed) surfaces as its own ImportError here, rather than
+    # being caught and silently rerouted into the fallback below.
+    from . import sot_text
+    from .dataset import _CONFIG, _split_dir
+    from .text_norm import get_text_norm
```
Masao-Someki, line 37: Please use full module path

egs3/ami/esp2_s2t/dataset/builder.py
```diff
@@ -0,0 +1,387 @@
...
+    if _norm_key in sys.modules:
+        _text_norm = sys.modules[_norm_key]
+    else:
+        _norm_spec = importlib.util.spec_from_file_location(
+            _norm_key, Path(__file__).resolve().parent / "text_norm.py"
+        )
+        _text_norm = importlib.util.module_from_spec(_norm_spec)
+        sys.modules[_norm_key] = _text_norm
+        _norm_spec.loader.exec_module(_text_norm)
+    get_text_norm = _text_norm.get_text_norm
```
Masao-Someki, line 80: Could you remove these? I think we required to have this long part because we wrote relative module path above and test environment cannot resolve it.

egs3/ami/esp2_s2t/dataset/builder.py
```diff
@@ -0,0 +1,387 @@
...
+
+    # ------------------------------------------------------------- helpers
+    def _root(self, key: str) -> Path:
+        """Resolve a configured directory against ``data_root``."""
+        path = Path(_CONFIG[key])
+        return path if path.is_absolute() else Path(_CONFIG["data_root"]) / path
+
+    def _cutset_path(self, split: str) -> Path:
+        """Return the local path of ``split``'s CutSet manifest."""
+        return self._root("cutset_dir") / _CONFIG["sot"][split]["cutset"]
```
Masao-Someki, line 125: Could you change it to verb, all function name should be verb

egs3/ami/esp2_s2t/dataset/builder.py
```diff
@@ -0,0 +1,387 @@
...
+            return
+        audio_root.mkdir(parents=True, exist_ok=True)
+        script = (
+            Path(__file__).resolve().parents[3]
+            / "egs2"
+            / "ami"
+            / "asr1"
+            / "local"
+            / "ami_download.sh"
+        )
```
Masao-Someki, line 167: Could you try to implement in the other way? egs2 is not included in the pypi package, and this will fail when we distribute this recipe.
One idea is that you might simply use lhotse, it has manifest preparation script for ami.

egs3/ami/esp2_s2t/dataset/config.yaml
```diff
@@ -0,0 +1,82 @@
+# Builder and dataset settings, loaded at import time by dataset/__init__.py.
+builder:
+  # Root the split directories and the relative wav.scp paths resolve against.
+  data_root: ${oc.env:AMI_SOT_DATA_ROOT,.}
```
Masao-Someki, line 4: Please refer to this file, you can simply set the environment variable.
https://github.com/cyhuang-tw/espnet/blob/6baa37f58ed496fa91a181ec76343c9c7b9eaa23/egs3/librispeech_100/asr/dataset/config.yaml#L3

Masao-Someki: Or if you want to put it under recipe, historycally usually we put it under `download/`

egs3/ami/esp2_s2t/dataset/builder.py
```diff
@@ -0,0 +1,387 @@
...
+        logger.info("Wrote the token list to %s", target)
+
+    def _build_split(self, split: str) -> None:
+        """Serialize one split's cutset into a Kaldi-style directory.
+
+        Args:
+            split: Key shared by ``builder.split_dirs`` and ``builder.sot``.
+        """
+        import soundfile as sf
+        from lhotse import CutSet
```
Masao-Someki, line 309: Oh ok you're using lhotse here? Then I think you don't have to download it with huggingface hub or relative-path espnet2 script, you can just use lhotse there

egs3/ami/esp2_s2t/dataset/builder.py
```diff
@@ -0,0 +1,387 @@
...
+            except ValueError:
+                stored = segment_path
+            entries.append((cut.id, str(stored), text))
+
+        entries.sort(key=lambda entry: entry[0])
+        self._write_kaldi_dir(split_dir, entries)
+        logger.info("Wrote %d entries to %s", len(entries), split_dir)
+
+    @staticmethod
+    def _write_kaldi_dir(split_dir: Path, entries) -> None:
```
Masao-Someki, line 361: If we want to use kaldi files, then could you write more in detail that how each files looks like, and how they're used!

egs3/ami/esp2_s2t/dataset/dataset.py
Masao-Someki, line 1: If you want to use lhotse, please consistently use it. I think it is very in-efficient to use HF, espnet2-shellscript, lhotse, kaldi-style files, soundfile, ... to load data.
You can simply construct the final SOT CutSet in Lhotse by concatenating/composing the corresponding cuts, keeping the individual supervisions, and generating the SOT transcript as text_1 <sc> text_2. Then the resulting CutSet itself can be used as the dataset without materializing segmented WAV files and converting them back to Kaldi-style files.

egs3/ami/esp2_s2t/dataset/text_norm.py
```diff
@@ -0,0 +1,60 @@
...
+``chime8_keep_fillers`` is the CHiME-8 DASR English normalizer with filler
+removal switched off. It comes from chime-utils, which is not on PyPI, so it
+is an optional dependency; ``none`` needs nothing.
+"""
+
+from __future__ import annotations
+
+from typing import Callable, Optional
+
+_INSTALL_HINT = "pip install git+https://github.com/chimechallenge/chime-utils@main"
```
Masao-Someki, line 12: Add this in the pyproject.toml, and add warning (if we detect people doesn't have the package) to run `pip install espnet[<system name>]`

egs3/ami/esp2_s2t/src/inference.py
```diff
@@ -0,0 +1,119 @@
...
+from typing import Any, Dict
+
+try:
+    from egs3.ami.esp2_s2t.dataset.dataset import current_utt_id
+except ImportError:
+    # A packed bundle ships src/ and dataset/ at its root with no egs3 package
+    # above them; InferenceModel puts the bundle root on sys.path, so the
+    # top-level spelling resolves there. current_utt_id's state lives in
+    # os.environ, so which copy answers does not matter.
+    from dataset.dataset import current_utt_id
```
Masao-Someki, line 29: Do we need this? I think we can remove this block

egs3/ami/esp2_s2t/src/inference.py
```diff
@@ -0,0 +1,119 @@
...
+from espnet2.text.whisper_token_id_converter import OpenAIWhisperTokenIDConverter
+from espnet2.text.whisper_tokenizer import OpenAIWhisperTokenizer
+
+MODEL_TYPE = "whisper_multilingual"
```
Masao-Someki, line 33: For these default values please include it in default config located under `egs3/TEMPLATE/`

egs3/ami/esp2_s2t/src/inference.py
```diff
@@ -0,0 +1,119 @@
...
+
+# Everything below is built on first use, not at import time. Importing this
+# module must not need the corpus or the Whisper assets, so that a unit test can
+# import it anywhere.
+_CONVERTER = None
+_TOKENIZER = None
+
+
+def _tokenizers():
```
Masao-Someki, line 42: Could you change the function name to verb

egs3/ami/esp2_s2t/src/inference.py
```diff
@@ -0,0 +1,119 @@
...
+    """Return the (converter, tokenizer) pair, built once on first use.
+
+    ``keep_special_tokens`` is what preserves the ``<|x.xx|>`` timestamps
+    through ``ids2tokens``. The default converter drops them.
+    """
+    global _CONVERTER, _TOKENIZER
```
Masao-Someki, line 48: Do these values have to be the global variables?

egs3/ami/esp2_s2t/src/preprocessor.py
```diff
@@ -0,0 +1,138 @@
...
+        ValueError: When the separator does not tokenize to a single id.
+    """
+
+    def __init__(
+        self,
+        train: bool,
+        token_type: str,
+        token_list,
+        bpemodel,
+        *args,
```
Masao-Someki, line 68: remove this

egs3/ami/esp2_s2t/src/preprocessor.py
```diff
@@ -0,0 +1,138 @@
...
+        speaker_change_symbol: str = "????",
+        whisper_language: str = "en",
+        whisper_task: str = "transcribe",
+        **kwargs,
+    ) -> None:
+        super().__init__(train, token_type, token_list, bpemodel, *args, **kwargs)
+
+        shared = dict(
+            model_type="whisper_multilingual",
```
Masao-Someki, line 77: Let's set this from config?

egs3/ami/esp2_s2t/src/preprocessor.py
```diff
@@ -0,0 +1,138 @@
...
+
+        # The base class drops index 0, expecting a prompt id there. Our text
+        # supplies its own <language><task>, so stand a placeholder in that
+        # slot rather than let a real symbol be eaten. The base reads
+        # notime/first_time/last_time from token2id before this runs, and those
+        # three resolve identically either way, so the swap is safe.
+        placeholder = converter.token2id["<|startoftranscript|>"]
+        convert = converter.tokenizer.tokenizer.convert_tokens_to_ids
+        converter.tokens2ids = lambda tokens: [placeholder] + convert(tokens)
+        self.token_id_converter = converter
```
Masao-Someki, line 94: Create helper function to avoid lambda

egs3/ami/esp2_s2t/src/preprocessor.py
```diff
@@ -0,0 +1,138 @@
...
+
+    def __call__(self, uid: str, data: Dict[str, Any]) -> Dict[str, Any]:
+        """Preprocess one sample, leaving speech one-dimensional.
+
+        Args:
+            uid: Utterance id.
+            data: Sample with ``speech`` and the S2T text fields.
+
+        Returns:
+            The processed sample.
```
Masao-Someki, line 131: Please add an example how output looks like!

egs3/ami/esp2_s2t/dataset/sot_text.py
Masao-Someki, line 1: I think you can remove this file if you use lhotse.

egs3/ami/esp2_s2t/readme.md
Masao-Someki, line 1: Please fix the document based on the reviews

egs3/TEMPLATE/esp2_s2t/run.py
```diff
@@ -0,0 +1,34 @@
...
+The argument parsing and stage dispatch are the generic ones. They live
+under ``egs3/TEMPLATE/asr`` because that is where the first system to need
+them put them; nothing in them is ASR-specific.
+"""
+
+from __future__ import annotations
+
+from typing import List
+
+from egs3.TEMPLATE.asr.run import build_parser, main, parse_cli_and_stage_args
```
Masao-Someki, line 12: You can copy from asr system. Let's avoid to have dependency to other system

espnet2/bin/s2t_inference.py
Masao-Someki, line 1: Since we want to archive espnet2 development in the future let's make this as one of the example how to avoid touching espnet2.
Could you simply copy this s2t_inference to the system directory and edit with your new implementation?

espnet2/bin/whisper_export_vocabulary.py
Masao-Someki, line 1: ditto

espnet2/s2t/whisper_timestamp_scorer.py
Masao-Someki, line 1: ditto

espnet2/text/whisper_token_id_converter.py
Masao-Someki, line 1: ditto

espnet2/text/whisper_tokenizer.py
Masao-Someki, line 1: ditto

espnet3/systems/esp2_s2t/metrics/cpwer.py
```diff
@@ -0,0 +1,248 @@
...
+from scipy.optimize import linear_sum_assignment
+
+from espnet2.text.cleaner import TextCleaner
+from espnet3.components.metrics.base_metric import BaseMetric
+
+# The speaker-change symbol belongs to the checkpoint, so it is a constructor
+# argument rather than a constant: a metric cannot know which symbol a model
+# was trained with. Text written with a different spelling has to be converted
+# before it is scored here.
+_SEP = "▁SPKCHANGE▁"  # internal marker unlikely to occur in text
```
Masao-Someki, line 33: I thought speaker change symbol was `"????"`? Let's use the same one, or is this `_SEP` is a different token?

espnet3/systems/esp2_s2t/metrics/cpwer.py
```diff
@@ -0,0 +1,248 @@
...
+    for chunk in text.split(_SEP):
+        chunk = strip_special_tokens(chunk)
+        if cleaner is not None:
+            chunk = cleaner(chunk).strip()
+        if chunk:
+            blocks.append(chunk)
+    return blocks
+
+
+def edit_counts(ref: List[str], hyp: List[str]) -> Tuple[int, int, int, int]:
```
Masao-Someki, line 59: Is this calculating wer? Please use `jiwer`, this implementation is very slow.

espnet3/systems/esp2_s2t/metrics/cpwer.py
Masao-Someki, line 1: I think you can consider using `meeteval`. You can check this file: https://github.com/cyhuang-tw/espnet/blob/espnet3-ami-sot/egs2/chime7_task1/asr1/local/da_wer_scoring.py

espnet3/systems/esp2_s2t/metrics/der.py
Masao-Someki, line 1: You can check https://github.com/cyhuang-tw/espnet/blob/espnet3-ami-sot/egs2/chime7_task1/asr1/local/da_wer_scoring.py
And please use pyannote, I think it's much better than implementing everything from scratch.

egs3/TEMPLATE/esp2_s2t/__init__.py
Masao-Someki, line 1: Please add config, and other files. You can refer to other recipe like librispeech-100h.

espnet3/systems/esp2_s2t/system.py
```diff
@@ -0,0 +1,32 @@
...
+
+        Raises:
+            NotImplementedError: Always. The vocabulary comes with the
+                pretrained model, so there is nothing to train.
+        """
+        raise NotImplementedError(
+            "S2T uses the vocabulary that ships with the pretrained model, "
+            "so there is no tokenizer to train. Drop train_tokenizer from "
+            "--stages."
+        )
```
Masao-Someki, line 32: In this case I think you can just not implementing this?
Also it is not listed in the run.py in `TEMPLATE/`, so user would never reach to this function.

Masao-Someki (commented):
For recipe name maybe `sot_finetune` be fine? @sw005320 do you have any other good idea?

And sorry I didn't check the unit test side, but since there're quite a lot things we need to fix. I'll check unit test after we resolved the review comments.

Masao-Someki:
Hi sorry I forgot to mention, would you create an integration test?
You can create one super small recipe for `mini_an4` dataset, and run it from `ci/test_integration_espnet3.sh`.

You can check these files and PRs:
- [ci/test_integration_espnet3.sh](https://github.com/espnet/espnet/blob/master/ci/test_integration_espnet3.sh)
- [egs3/mini_an4/](https://github.com/espnet/espnet/tree/master/egs3/mini_an4)
- https://github.com/espnet/espnet/pull/6726/changes#diff-03e588e02f42c91b9edbf2812816ee3d684a38a7064bb9174433c9cddafea54c

PR #6774, opened 2026-09-20
[espnet3-recipe] Add LibriCSS recipe for ESPnet3

akreal:
## What did you change?

Add LibriCSS recipe for ESPnet3

---

LibriCSS is a popular benchmark for continuous speech separation and recognition.

There used to be a recipe for ESPnet1, but now it does not exist. 

---

## Is your PR small enough?

Yes.

---


akreal:
@Masao-Someki this PR should be ready for review.

PR #6775, opened 2026-09-20
Add `espnet phonemize`: the phones in a recording, with POWSM

sw005320:
POWSM is published — [`espnet/powsm`](https://huggingface.co/espnet/powsm) and
[`espnet/powsm_ctc`](https://huggingface.co/espnet/powsm_ctc) — but reading
phones out of a file meant writing the loop yourself or copying the snippet
from the model card. This is that snippet as a subcommand:

```
$ espnet phones audio.wav
ðəseɪlʌvðəhoʊtɛlzɪzpʰɑɹtʌvhɑlədeɪstɹætədʒitʰusɛlɔfæsətsəndkʰɑ̃nsəntɹeɪtɑ̃npʰɹɑpə˞timæ̃nədʒmənt

$ espnet phones audio.wav --spaced
ð ə s e ɪ l ʌ v ð ə h o ʊ t ɛ l z ɪ z pʰ ɑ ɹ t ʌ v h ɑ l ə d eɪ s t ɹ æ t ə dʒ i tʰ u s ɛ l ɔ f …
```

(that is `test_utils/ctc_align_test.wav`, "the sale of the hotels is part of
Holiday's strategy to sell off assets and concentrate on property management")

The default is `espnet/powsm_ctc`: it reads a recording of any length through
`decode_long`, and it is about three times faster than the encoder-decoder one
— 5.1 s against 18 s for that clip on this laptop.

`--spaced` exists because the slashes are load-bearing. POWSM writes each phone
between them precisely so that `/pʰ/` is one phone rather than `p` followed by
`ʰ`; printing IPA throws that away, so the other form keeps it.

### Three things had to be fixed for the other checkpoint to work at all

**`decode_long` ended in `KeyError: '<29.00>'`.** The threshold for "this
utterance was cut off by the window rather than finished" was written out as
OWSM's 30 s window minus a second. POWSM's window is 20 s and has no such
token. It is now a second before the end of whatever window the config
describes: OWSM still gets `<29.00>`, POWSM gets `<19.00>`, and the fixture in
the tests gets `<0.00>`.

**That stopped the crash but not the output.** Asked for phones, POWSM fills a
window that is mostly padding with repetitions of what it already said — 118
seconds of decoding to produce `saɪdɪtsaɪdɪtsaɪt…` from a 6.5 s file. So for an
encoder-decoder checkpoint this command decodes one window at a time and joins
them, which is what the model card does. A 39 s file comes back correct.

**The language symbol was a second `KeyError`.** `espnet/powsm` records no
`nolang_symbol` and has no `<nolang>` in its vocabulary at all, so guessing one
failed several seconds after the model had loaded. The symbol now comes from
the config, is checked against the token list, and falls back to `<unk>`, which
is what POWSM-CTC records; a checkpoint with neither is told to name a language
with `--language`.

`Speech2Text._read_audio` is public as `read_audio`: a caller that cuts a
recording up itself should not repeat the resampling or guess the rate the
checkpoint wants.

### Checked

- `espnet phones` against both checkpoints, on a 6.5 s file and a 39 s one, and
  `--spaced`; the 39 s file is the 6.5 s one six times over and comes back as
  six repetitions, which is the right answer.
- Seven new tests in `test_cli.py` (the slashes, `--spaced`, the language
  symbol and the two ways it can be missing, the window-by-window path, and
  text that has no slashes in it), and one in `test_s2t_inference.py` for the
  long-form threshold following the model's window.
- `test_cli.py` 56 passed; `test_s2t_inference.py` passes except
  `test_Speech2Text_quantized`, which fails the same way on master here
  (`Didn't find engine for operation quantized::linear_prepack`).
- `ci/check_ci_image_config.py`, black, isort and flake8 clean.

Next, on top of this: the same two models in `espnet demo`, and then
`espnet align` — CTC segmentation is in the repository three times over
(`asr_align.py`, `s2t_ctc_align.py`, and `egs2/powsm/s2t1/force_align.py`) and
reachable from neither the command line nor the demo.


sw005320:
> is `last_time_symbol` intended to be the generic contract for `decode_long()`?

No, and you are right that it should not be. Changed in 454cb11: the window is
the source of truth — `speech_length`, which every other part of the class
already reads — and the step between two timestamps is that window divided by
the symbols that span it. Nothing parses a symbol's name or assumes `<NN.NN>`
any more, and the threshold is an id rather than a string, so it is never
looked up by a name that may not exist. That is what `KeyError: '<29.00>'`
was. A checkpoint that spells its timestamps differently needs no format
agreed here, only a config naming its first and last.

**Deriving it also settled a disagreement I would otherwise have had to pick a
side of.** POWSM's config states `speech_resolution: 0.04` while its
vocabulary steps 0.02:

```
espnet/owsm_v4_base_102M   window 30 s   states 0.02   1500 steps -> 0.020 s
espnet/powsm               window 20 s   states 0.04   1000 steps -> 0.020 s
```

Read at twice its value, every timestamp points past the end of the window it
came from, and long-form decoding cuts the recording in the wrong places. The
vocabulary is the one to believe. OWSM is unchanged either way, and still
segments the 39 s file into the same utterances.

Two tests: the fixture's own window, then what OWSM and POWSM each describe
(without needing either config present), and one for the stated-resolution
disagreement.

> the encoder-decoder POWSM case can spend ~118 seconds decoding a 6.5-second
> input

That 118 s was `decode_long` on that checkpoint — the timestamp path filling a
padded window with repetitions — and this PR does not take it: for an
encoder-decoder model `espnet phonemize` decodes window by window. Measured
now, on this laptop's CPU:

```
espnet/powsm_ctc   load  7.7 s   decode   4.3 s   for 6.5 s of audio
espnet/powsm       load  7.0 s   decode  16.1 s
```

Still slow, and for the reason you name: the model always sees a full 20 s
window, so a 6.5 s file pays for 13.5 s of padding. Trimming the window to the
audio would be the fix, and it needs measuring rather than assuming — these
models are trained on fixed-length input, and shorter input may cost accuracy.
I will open an issue with these numbers so it is tracked rather than
remembered. The default (`espnet/powsm_ctc`) does not have the problem: it
reads a recording of any length through `decode_long`.

sw005320:
> Does this remain correct if the model's window duration or time-token
> resolution changes in the future?

For the window and the step themselves, yes: both come from the checkpoint,
and a model that changes either changes `speech_length` and its vocabulary
together. Verified at 30 s/0.02 and 20 s/0.02, which exist, and at 10 s/0.01
and 60 s/3 s, which do not:

```
espnet/owsm_v4_base_102M   window 30 s   step 0.020 s   threshold 29.00 s
espnet/powsm               window 20 s   step 0.020 s   threshold 19.00 s
```

But the question is the right one to press on, because the derivation *does*
assume something, and it was assuming it silently: **that the timestamps run
one id a step, from the start of the window to its end**. A model that broke
that would have got times a constant factor out, quietly — the same failure
POWSM's stated `speech_resolution` would have caused.

So in 44c146c that assumption is checked rather than trusted. When the symbols
are named after the times they mark — both models in the wild are — the names
have to agree with `speech_length`, and a config at odds with itself now says
which two of its own statements disagree:

```
this config's timestamps span 20 s (<0.00> to <20.00>) while its
speech_length says 30 s. One of the two is wrong, and either would put every
timestamp in the wrong place
```

The check runs *because* the names happen to be times; it is a check on the
config, not the contract. A checkpoint that writes its timestamps some other
way skips it and is read from its ids alone — which is the point of not making
the names the contract in the first place.

One more edge the question exposed: a step coarser than a second would have
made `round(1/step)` zero, so "an utterance ending near the window's end was
cut off" would have meant "ending exactly at the end", which is no threshold
at all. It goes back one step in that case, and there is a test at 60 s/3 s
for it.

espnet2/bin/cli.py
```diff
@@ -189,6 +193,78 @@ def decode(chunk):
...
+    found = PHONE.findall(decoded)
+    if not found:
+        return decoded
+    return " ".join(found) if spaced else "".join(found)
+
+
+def _no_language(s2t) -> str:
+    """The symbol this checkpoint uses for "work the language out yourself".
+
+    OWSM spells it `<nolang>`; POWSM-CTC records `<unk>`; POWSM itself records
```
chinjouli: Both POWSM and POWSM-CTC use `<unk>`

espnet2/bin/cli.py
```diff
@@ -189,6 +193,78 @@ def decode(chunk):
...
+    s2t = _build(Speech2Text, args, "phonemize")
+    lang_sym = f"<{args.language}>" if args.language else _no_language(s2t)
+
+    if s2t.ctc_only:
+        decoded = _decode(s2t, args.audio, lang_sym, "<pr>")
+    else:
+        # An encoder-decoder checkpoint segments long audio by its own
+        # timestamps, which is right for a transcript and wrong here: asked
+        # for phones, POWSM fills a window that is mostly padding with
+        # repetitions of what it already said. One window at a time instead,
```
chinjouli: "padded with repetitions of what it already said" -> probably padded with silence?

espnet2/bin/cli.py
```diff
@@ -363,6 +439,19 @@ def add(name, help_text, func, device="cpu"):
         help="OWSM language token, ISO 639-3: eng, jpn … (default: OWSM detects it)",
     )
 
+    p = add("phonemize", "recognise the phones in an audio file", cmd_phonemize)
+    p.add_argument("audio", help="audio file, any format soundfile reads")
+    p.add_argument(
+        "--language",
+        default=None,
```
chinjouli: if it's just for powsms, set default to "unk" and encourage users to specify by their own

chinjouli (commented):
lgtm, I just have some comments on the comments

sw005320:
Thank you — all three taken, in b15b43c.

**Both POWSM checkpoints use `<unk>`.** You are right, and my comment read as
if POWSM used something else. What differs is only that POWSM-CTC's config
records `nolang_symbol` and POWSM's does not, which is the whole reason for
the lookup. Reworded to say that.

**The padding is silence; the repetition is the output.** I had the two the
wrong way round in one sentence. A window shorter than the model's is padded
with silence, and asked for phones over that silence the model says again what
it has already said, for as long as the window lasts.

**`--language` is worth naming**, and the help says so now. On
`test_utils/ctc_align_test.wav`, POWSM-CTC:

```
auto            dəseɪlʌvðəhotɛlsɪzpɑɾtʌvhɑlɪdestrɑtəɡitusɛlɔfɑsətsən…
--language eng  ðəseɪlʌvðəhoʊtɛlzɪzpɑɹtʌvhɑlədeɪstɹætədʒitusɛlɔfæsətsənd…
```

English `ɹ` and diphthongs rather than a tap and plain vowels — a clear enough
difference to push people towards naming the language.

I kept the default as the model's own symbol rather than hardcoding `<unk>`,
for one reason: it resolves to `<unk>` for both POWSM checkpoints anyway, and
since POWSM's config does not record it, reading the vocabulary is what makes
that checkpoint work at all. Hardcoding would be right today and wrong the
first time a checkpoint spells it differently. If you would rather the command
refused to guess — required `--language` unless the model records a symbol —
say so and I will do that instead; it is a small change and your call as much
as anyone's.

chinjouli (approved):

PR #6776, opened 2026-09-20
The demo page is the checkpoint's, not OWSM's

sw005320:
**Stacked on #6775** (`espnet phones`) — the diff to read is the last commit.

`espnet demo --model espnet/powsm_ctc` served OWSM's page: a 30 s window for a
model trained on 20, `<nolang>` for a model whose vocabulary has no such token,
and no way to ask for the one thing POWSM is for.

Four things now come off the loaded model rather than the module:

- **The window it decodes in a pass.** `pad()` keeps OWSM's 30 s as its
  default — which is what the two Space apps pass, and what
  `test_demo_apps.py` compares against — and takes the model's window when
  there is one.
- **The symbol for "no language given"**, through `Speech2Text.no_language()`,
  which this moves out of the command line where #6775 had already needed it
  for the same reason.
- **The task menu**, which gains "Recognise phones" when the token list has
  `<pr>` and is unchanged when it does not.
- **The phone output**, spaced rather than run together: POWSM writes each
  phone between slashes so that `/pʰ/` is one phone, and the page keeps that.

### Checked, by building the app and calling the function behind the button

```
espnet/powsm_ctc        window 20 s, <unk>, menu [Transcribe, Recognise phones]
                        phones: d ə s e ɪ l ɔ v ð ə h o t ɛ l s ɪ z p ɑ r t ʌ v …

espnet/owsm_ctc_v4_1B   window 30 s, <nolang>, 26 menu entries, no phone option
                        asr: The sale of the hotels is part of Holi's strategy to
                             sell off assets and concentrate on property management.
```

So the default demo is untouched, and so are the two Space apps: their drift
tests (`test_demo_apps.py`, 79 of them) still pass unchanged, which is the
point of keeping `pad()`'s default where it was.

Three new unit tests: padding to a given window, the window coming from the
config (and falling back to 30 when a config does not say), and the phone
option appearing only for a checkpoint that has the symbol.

### One thing worth knowing

The Language box shows what the model said, and POWSM says something arbitrary
— "Norwegian Bokmål" for the English clip above. POWSM learns language
identification implicitly and its card points at `Speech2Language` for it
rather than at the decoder's first token. Nothing is wrong with the wiring;
the field is simply not meaningful for that checkpoint, and I left it showing
the model's answer rather than hiding it. Say if you would rather it were
hidden when the model is not OWSM.


sw005320:
Good call — `phonemize` it is, in #6775 where the command is defined, and this
branch is rebased on it. `espnet phonemize audio.wav` now, and the align
command that came next is `espnet align`, which was already a verb.

On the wider suggestion — `asr` → `recognize`, `tts` → `synthesize` — I have
left those alone for now, because unlike `phonemize` they are in a release and
renaming them breaks whatever people have already typed into scripts. If you
want the convention across the board, the way that costs nobody anything is to
add the verbs and keep the old names as aliases, hidden from `--help`, with a
line in the release notes. Say the word and I will do that as its own PR;
it is a decision about the public surface rather than a tidy-up.

sw005320:
Both findings were right, and the second one the page was getting visibly
wrong. Addressed in the last commit.

**`best_path()` on an encoder-decoder checkpoint ignores the task.** Measured
on `espnet/powsm`:

```
best_path <asr>: /ð//ə//s//e//ɪ//l//ʌ//v//ð//ə//h//o//ʊ//t//ɛ//l//z/…
best_path <pr> : /ð//ə//s//e//ɪ//l//ʌ//v//ð//ə//h//o//ʊ//t//ɛ//l//z/…
```

The same phones either way — that model's CTC head answers phones whatever you
ask it, and only the decoder reads the symbol. So "Transcribe" returned phones.
The page now calls the object when there is a decoder and keeps `best_path` for
CTC-only checkpoints, where it is both right and an order of magnitude faster.
Long-form on an encoder-decoder model goes window by window, for the reason the
command line does: `decode_long` segments by timestamps, and a phone model
fills a padded window with repetitions.

**`no_language()` raising.** `espnet/powsm` has no symbol for "work it out
yourself", so the page would have raised on every Run. The language menu now
simply has no Detect entry for such a checkpoint and opens on a language
instead — your suggestion, with a test that builds the page on a model whose
`no_language()` raises.

With both, on `espnet/powsm`:

```
Recognise phones  ->  ð ə s e ɪ l ʌ v ð ə h o ʊ t ɛ l z …
Transcribe        ->  T ⁇ E  ⁇ A ⁇ E OF T ⁇ E …
```

The second is POWSM's own ASR — it comes out the same when called directly,
and its model card says as much and points at a retrained variant — so it is
the checkpoint rather than the page. Worth stating rather than leaving for
someone to read as a bug here.

chinjouli (commented):
Looks okay since most changes are included in #6775. 
Some comments on the latest comment:
- Not sure what `best_path()` is for, but for `espnet/powsm` its encoder-CTC only produces phones. If this function is used, worth noting it somewhere.
- As in #6775, setting `<unk>` as POWSM's language token default could solve the `no_language` issue.
- The default checkpoint's English ASR is broken for LibriSpeech due to text normalization; state it in the docs and point users to the retrained version.

sw005320:
Thank you — that confirmation is exactly what I needed for the first one.

**"for `espnet/powsm` its encoder-CTC only produces phones"** — that matches
what I measured (`<asr>` and `<pr>` giving the same phones), and it is the
reason this PR calls the object on an encoder-decoder checkpoint instead. It
was worth writing down where someone will meet it, so `best_path`'s docstring
now says it:

> A CTC branch answers what that branch was trained on, which is not always
> what `task_sym` asks for. POWSM's answers phones whether it is asked for
> `<pr>` or `<asr>` … only the decoder — `__call__` — reads the task.

**The English ASR and the retrained version** — said on the page now, and only
when the loaded checkpoint has `<pr>`:

> This checkpoint is a phonetic model. **Recognise phones** is what it is for;
> its transcription is weaker than a model trained for text, and on POWSM in
> particular the English ASR suffers from a text normalisation problem — the
> authors have since published a retrained variant in that repository's
> `textnorm_retrained` folder.

Better there than found by someone who reads the mangled output as a bug in
this page. If you would rather the demo loaded `textnorm_retrained` by default
when someone passes `--model espnet/powsm`, that is a one-line change — but it
seemed wrong for the command to silently load a different checkpoint than the
tag it was given, so I left the choice with the user and the note with the
page.

**`<unk>` as the default** — it already resolves to `<unk>` for both POWSM
checkpoints. The lookup is not there to choose between models; it is there
because `espnet/powsm`'s config does not record `nolang_symbol` at all, so
reading the vocabulary is what makes that checkpoint work. Hardcoding would be
right today and wrong the first time a checkpoint spells it otherwise. The
`ValueError` path you saw is for a model that has neither symbol, which does
not exist today — it exists so that such a model gets a sentence rather than a
`KeyError` from inside the decoder.

chinjouli (approved):
looks good

sw005320:
Both taken, in 4b8dcca — and the second one turned up a regression I had put in.

**The sample rate.** You are right that it was the odd one out: the window had
become the checkpoint's and the rate had not. There is a `sample_rate(s2t)`
beside `window_secs(s2t)` now, reading `Speech2Text.sample_rate` (and
`preprocessor_conf["fs"]` behind it), and `pad()` and `read_audio()` take a
rate. The module constant stays as the default, because it is what OWSM uses
and what the two Space apps pass — `test_demo_apps.py` compares them against
this module, so the default is the thing that keeps those honest.

**The decoding capability.** Moved into the inference object as
`Speech2Text.decode_window(speech, lang_sym, task_sym)`: CTC-only is read off
its head, a model with a decoder is called. Both front ends had to know which
kind of checkpoint they were holding, and as you say, it is the checkpoint
that knows.

What is left in `demo.py` is one line, and **checking it against your comment
showed it was wrong**: I had split long-form on `ctc_only`, which took
timestamp segmentation away from encoder-decoder OWSM, where it is the right
thing. The exception is a task, not a kind of model — only a phone model asked
for phones fills the padding with repetitions — so it now reads

```python
phone_pass = task_sym == PHONE_TASK and not s2t.ctc_only
```

and everything else keeps `decode_long`. Thank you for the push; the narrower
rule is both more correct and less model-type logic than what I had.

PR #6777, opened 2026-09-20
One alignment, on torchaudio's forced alignment

sw005320:
**Stacked on #6775 and #6776** — the diff to read is the last two commits.

CTC segmentation is in this repository three times over — `asr_align.py` for an
ASR model, `s2t_ctc_align.py` for OWSM-CTC, and `egs2/powsm/s2t1/force_align.py`,
which uses torchaudio instead — and none of them was reachable from the command
line. This PR began by wrapping the first two. **That was the wrong half to
keep**, and the measurements say so.

### The two algorithms, on the same posteriors

| | `ctc_segmentation` | `torchaudio.forced_align` |
|---|---|---|
| algorithm time (6.5 s audio) | 0.08 s | 0.03 s |
| "the sale of the hotels" | 0.20 – 1.57 | **0.48 – 1.60** |
| text missing the middle utterance | 0.20 – **2.03** (stretched, conf −0.63) | 0.48 – 1.60 (**unchanged**, 0.83) |
| per-token times | ✗ | ✓ |

Speech in that file starts at **0.62 s** by its own energy, so the later start
is the right one: a forced alignment marks where a token *is*, where
segmentation partitions the timeline and hands the leading silence to the first
utterance. And the case CTC segmentation exists for — text that does not cover
the audio — is the one where it does worse. Speed turned out not to be an
argument either way: the model's forward pass is 10–15 s and the alignment is
hundredths of a second.

### What this is now

`espnet2/bin/align.py`, one implementation for every checkpoint with a CTC head:

```
$ espnet align audio.wav --text-file said.txt --tokens
0.64    1.76    0.8269  the sale of the hotels
  0.64  0.72    0.0073  ▁the
  0.88  0.96    0.9743  ▁sale
  1.12  1.20    0.9996  ▁of
  1.20  1.28    0.9948  ▁the
```

It takes a tag and loads whichever inference class the model was published for,
so an ASR checkpoint and an S2T one go the same way. The score is a probability
rather than a log, so a threshold means the same thing across languages and
lengths.

The buffering that reads a long recording into CTC frames is now
`Speech2Text.ctc_log_probs`, and long-form best-path decoding is an argmax over
it — one windowing, read two ways.

### What is not touched

`asr_align.py` and `s2t_ctc_align.py` keep their scripting interfaces, and gain
`from_pretrained` and an explicit `device` (the first two commits, and the
CodeRabbit findings on them). **The OWSM v4 recipe cleans its training data with
the second of them** (`egs2/owsm_v4/s2t1/local/ctc_seg.py`, paper §2.1.1), and
on four kinds of caption the two scores discriminate alike:

| caption | ctc_segmentation | forced alignment |
|---|---|---|
| matches the audio | −0.40 … −0.26 | 0.84 … 0.86 |
| covers a third of it | −0.27 … −0.26 | 0.87 … 0.93 |
| a different text | −8.30 … −6.81 | 0.06 … 0.11 |
| half the words wrong | −3.74 … −3.71 | 0.51 … 0.53 |

Neither notices a caption that covers only part of the recording; both separate
wrong text sharply. So that pipeline has no reason to move, and removing the
`espnet-ctc-segmentation` fork is a later decision than this one — worth
settling on a YODAS sample rather than on one clip.

### Checked

- Both kinds of model end to end: `espnet/owsm_ctc_v4_1B` and
  `espnet/kamo-naoyuki_wsj_transformer2`, three utterances each, with and
  without `--tokens`.
- `espnet asr long.wav` after the refactor: six repetitions transcribed, so
  best-path long-form still reads the same frames.
- `test_align.py` — 6 new tests on emissions written out by hand: which frames
  a token is given, the tokens carried per segment, text that cannot fit, an
  empty utterance, a model with no CTC head.
- `test_cli.py` 62 passed. `test_s2t_inference.py` passes except
  `test_Speech2Text_quantized`, which fails the same way on master here.


chinjouli (approved):
Looks solid

chinjouli (approved):
lgtm!

PR #6778, opened 2026-09-20
Say which tasks run where

sw005320:
The README lists the Hugging Face Spaces in one place, the notebooks in another, and the command line in a third, so a question like "can I do forced alignment without training anything?" means reading all three and hoping you did not miss a fourth. One table answers it: six tasks against the four ways to run a published model.

| Task | Space | Notebook | MCP | CLI |
| :-- | :-- | :-- | :-- | :-- |
| **ASR** | 🟢 | 🟢 | 🟢 `transcribe` | 🟢 `espnet transcribe` |
| **ST** | 🟢 | 🟢 | 🟢 `transcribe(translate_to=…)` | 🟢 `espnet translate` |
| **TTS** | 🟢 | 🟢 | 🟢 `synthesize` | 🟢 `espnet synthesize` |
| **SE** | 🟢 | 🟢 | 🟢 `enhance` | 🟢 `espnet enhance` |
| **PR** | ❌ | ❌ | 🚧 | 🟢 `espnet phonemize` |
| **ALIGN** | ❌ | ❌ | 🚧 | 🟢 `espnet align` |

It also names what is missing, which is the point of writing it down: phone recognition and alignment have no Space and no notebook.

Rewritten against master as of today, after #6775, #6777 and #6779 merged — the CLI column now has all six commands under the names #6779 settled on (`transcribe`, `synthesize`), and the 🚧 in the MCP column is #6780, which gives the server the same six names.

Checked rather than remembered: the CLI commands come from `build_parser()` and the MCP tools from the module (both on master, at the commit this branches from), and `ci/check_demo_links.py` passes on the Space and notebook links — 11 notebooks, 6 Spaces, each Space's `runtime.stage` read from the API.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #6780, opened 2026-09-20
MCP: translate, phonemize and align, under the names the CLI uses

sw005320:
**Stacked on #6775, #6776, #6777 and #6779** — the diff to read is the last
commit. This is the other half of the naming discussion on #6776: the command
line moved to verbs in #6779, and the MCP server gains the three tools it was
missing, under the same names.

```
transcribe  translate  phonemize  align  synthesize  enhance
```

A transcript of an agent's work and a terminal history now describe the same
thing.

- **`translate` is its own tool**, rather than `transcribe(translate_to=…)`.
  An agent picks a tool by its description, and one tool that sometimes
  translates is harder to pick than two that each do one thing. The argument is
  gone — the schema an agent reads is generated from the signature, so there is
  nothing to deprecate.
- **`phonemize`** is POWSM-CTC, with the slashes taken out: `ð ə s eɪ l`.
  `ESPNET_MCP_PR_MODEL` swaps the checkpoint.
- **`align`** is the forced aligner from #6777, on the recognition model this
  server has already loaded — alignment is a Viterbi path through the CTC head
  that `transcribe` reads, so it needs no second tag. One line an utterance:
  start, end, how sure, the text. **The score is a probability**, which matters
  here more than at a terminal: an agent deciding whether a caption belongs to a
  recording gets a number that means the same thing every time, rather than a
  log probability whose scale moves with length and vocabulary.

### Checked

Called over the server the way an agent calls it — `server.call_tool(...)`,
not the functions directly:

```
tools: ['align', 'enhance', 'phonemize', 'synthesize', 'transcribe', 'translate']

phonemize  d ə s e ɪ l ʌ v ð ə h o t ɛ l s ɪ z p ɑ ɾ t ʌ v …
align      0.64  1.76  0.8269  the sale of the hotels
           2.00  3.11  0.8023  is part of holiday's strategy
           3.43  5.91  0.9897  to sell off assets and concentrate on property management
```

`test_mcp_server.py` 13 passed, including seven new ones: the tool list and its
schemas, phones coming back one token each, the model's own "unknown language"
symbol, a language it does not have, alignment's output shape, text that cannot
fit in the recording, and alignment with nothing to align.

That fills the MCP column of the table in #6778 for all six tasks.


sw005320:
Three findings here; one is live and two are about code this branch replaced.

**Positional order** — live, and a fair catch. `device` went in after `ngpu`,
which moved `batch_size` and everything after it along by one; a caller passing
those positionally, as these constructors have allowed for years, would have
bound its batch size to the device. It is the last argument now, after
everything either class already had, and a check on the signature confirms the
old order:

```
asr ['self', 'asr_train_config', 'asr_model_file', 'fs', 'ngpu', 'batch_size', 'dtype', …] … ['device', 'ctc_segmentation_args']
```

Fixed on #6777, which is where that argument is added.

**"Select the alignment implementation from the checkpoint artifacts"** and
**"pass the explicit device to CTCSegmentation"** — both were true of an earlier
version of this branch. The MCP loader no longer touches either
`CTCSegmentation`:

```python
from espnet2.bin.align import ForcedAligner

return ForcedAligner.from_pretrained(ASR_MODEL, device=DEVICE)
```

`ForcedAligner` picks the inference class from the artifacts the tag arrives
with, and takes the device as it is written — which is what both findings ask
for, one layer down.

sw005320:
Thank you — taken in three parts.

**The duplicated `transcribe`.** That is the diff view, not the file. The head has one definition and no `translate_to` anywhere:

```
$ git show fork/mcp/tools:espnet2/bin/mcp_server.py | grep -n '^def \|translate_to'
140:def transcribe(audio_path: str, language: str = "auto") -> str:
157:def translate(audio_path: str, to: str, language: str = "auto") -> str:
173:def phonemize(audio_path: str, language: str = "auto") -> str:
207:def align(audio_path: str, utterances: List[str]) -> str:
256:def synthesize(text: str, output_path: str) -> str:
283:def enhance(audio_path: str, output_path: str) -> str:
```

GitHub is showing the removed lines and their replacement next to each other because the signature and the docstring changed together.

**The README.** Fixed, and a little further than CodeRabbit asked: `espnet models` takes neither `--model` nor `--device`, and `demo` and `models` are not verbs either. It now reads "Every command but `espnet models` takes `--model <tag>` and `--device cuda` … The six that do one task are verbs."

**The boundary.** I agree, and I have done the part of it that is cheap here: three places in the tool docstrings named OWSM's numbers where the behaviour is really the model's — the 30-second window, the 4 GB download, "25 languages". A tool description is the MCP public API, so those are now stated as the model's window, "the default is about 4 GB", and "a code the model was not trained for is reported with the ones it has" — which is what `_symbol` already does, from the checkpoint's own token list.

What stays model-specific is in the implementation, which is where you put the line: `_pr()` reads POWSM's slashes, `_decode` builds an OWSM task symbol. The direction you describe is the one #6776 started — `window_secs(s2t)` and `sample_rate(s2t)` came out of `demo.py` and onto `Speech2Text` — and the same move is what would let `_decode` ask the inference object for a capability instead of naming a task symbol. I would rather do that against a second ASR/ST model than in the abstract, so that the interface is shaped by two implementations and not one; I will open an issue for it so it is not lost.

sw005320:
Issue for the capability boundary: #6788.

espnet2/bin/mcp_server.py
```diff
@@ -107,36 +137,112 @@ def _symbol(model, symbol: str, what: str) -> str:
...
-        the model (about 4 GB) and takes minutes; later calls take 10-30 s per
-        30 s of audio on a laptop CPU, about a second on a GPU.
+        The transcript as plain text. The first call downloads the model -
+        the default is about 4 GB - and takes minutes; later calls take
+        10-30 s per 30 s of audio on a laptop CPU, about a second on a GPU.
+    """
+    return _decode(audio_path, language, "<asr>")
+
+
+def translate(audio_path: str, to: str, language: str = "auto") -> str:
```
chinjouli, line 178: We might want to add a user warning when it's not X-En or En-X translation, since OWSM is only trained on directions with English. Zero-shot ability might exist, but not guaranteed

chinjouli (commented):
👍

espnet2/bin/mcp_server.py
sw005320: Good catch — added.

`translate` now warns when English is on neither side:

```python
if to != "eng" and language not in ("auto", "eng"):
    warnings.warn(
        f"{language} to {to} translation has English on neither side, "
        "which is outside what the model was trained on; the output may "
        "be unreliable",
        UserWarning,
        stacklevel=3,
    )
```

and the docstring — which is the schema an agent reads *before* it calls anything, so it is the part that can stop the call rather than annotate it — now says: "The training data pairs English with the other language, so translation into or out of English is what the model was taught. A direction with English on neither side — Japanese into German, say — is outside that: it decodes, and may be right, but nothing about it was trained."

Two decisions worth naming:

- **The warning does not go into the returned text.** A translation that a pipeline reads should stay a translation. If you would rather an agent always see it, I can prefix a note line to the result instead — say the word.
- **`language="auto"` is documented, not warned about.** The source language is decided inside the model and never comes back out, so the server cannot tell an auto-detected English source from a Japanese one. Warning on every `auto` + non-English target would fire on the common English→X case.

`test_mcp_server.py` asserts both halves: jpn→deu warns, and jpn→eng, eng→deu and auto→deu do not (under `simplefilter("error")`). 14 passed.

PR #6782, opened 2026-09-21
An initialization this version removed must not stop a model loading

sw005320:
`espnet_model_zoo`'s daily run against master has been failing:

```
FAILED ci/test_model.py::test_model - ValueError: Unknown initialization: chainer
```

([the run](https://github.com/espnet/espnet_model_zoo/actions/runs/35569441576/job/106237777710))

`--init` says how to fill the parameters *before* training. Loading a trained
model fills them from the checkpoint instead, so the setting cannot matter
here — but `build_model` applies it anyway and raises on a value it does not
recognise, and `chainer` was one of six documented choices until it was
removed in June 2025 (9e9897a4703). **Every model published with it has been
unloadable since.** `jv_openslr35`, trained on espnet 0.9.7, is the one the
daily run reaches first.

`build_model_from_file` now drops an init this version does not have, with a
line saying so, and the check reads the same list the error message does.

```
WARNING:root:this model was trained with init=chainer, which this version of
espnet does not have. Ignoring it: the weights come from the checkpoint, not
from an initialization.
loaded and decoded: ''
```

(that is `Speech2Text(**ModelDownloader().download_and_unpack("jv_openslr35"))`
loading and decoding, on this branch)

### What this is not

It is not "ignore `init` when loading". A value this version knows is still
applied, because `load_state_dict` here is `strict=False`: a parameter the
checkpoint does not carry keeps what the initialization gave it, and silently
changing that for SSL front-ends and the like would be a real behaviour
change. And a misspelling still fails when building a model to train — three
tests cover exactly those three cases.

### Worth considering separately

The daily zoo run is the only thing that noticed, three months later. It walks
146 published models and stops at the first failure, so there may be more
behind this one; I can make it report every failure rather than the first, if
you want that as its own change.


sw005320:
> Only clear removed initialization values when a checkpoint will load.

Right, and thank you — `build_model_from_file` also builds from a config with
no model file, and there the initialization is all the model will ever have.
Ignoring an unknown one in that case hands back a differently initialized
model with nothing said about it, which is worse than the error.

Fixed in 7cebdc3: the drop happens only when `model_file is not None`.
Building from a config alone still raises `Unknown initialization: chainer`,
with a test for it — on a task that actually applies `args.init`, since the
stand-in task in that file did not, so the test would have passed either way.

`jv_openslr35` still loads and decodes on this branch.

PR #6787, opened 2026-09-21
One CTC segmentation, not two — and one naming for the alignment modules

sw005320:
`espnet2/bin/asr_align.py` and `espnet2/bin/s2t_align.py` were the same algorithm written twice. Measured on master, of their 1801 lines:

| piece | asr | s2t | differing lines |
|---|---|---|---|
| `CTCSegmentationTask` | 87 | 87 | 0 |
| `set_config` | 93 | 93 | 0 |
| `_split_text` | 17 | 17 | 0 |
| `get_segments` | 35 | 35 | 0 |
| `ctc_align` (CLI) | 42 | 42 | 8 — the two argument names |
| `get_parser` | 163 | 163 | 6 — the same two, and the description |
| `main` | 8 | 8 | 0 |
| `__call__` | 29 | 31 | 6 |
| `get_timing_config` | 18 | 14 | 4 |
| `prepare_segmentation_task` | 83 | 95 | 18 |
| `get_lpz` | 25 | 99 | genuinely different |
| `__init__` | 100 | 116 | genuinely different |

After #6777 there were three alignment modules with no explanation of why. This gives the answer the repository can keep: **modules are named after the algorithm, not the task.**

- `espnet2/bin/ctc_segment.py` (new) — the `ctc_segmentation` package's algorithm, once.
- `espnet2/bin/asr_align.py`, `espnet2/bin/s2t_align.py` — how each model produces CTC posteriors, plus its own script interface. 1801 lines become 775 + 203 + 308.
- `espnet2/bin/s2t_ctc_align.py` is renamed to `espnet2/bin/s2t_align.py`, so that the four modules read alike, and each of them opens with the same map of which is which. `from espnet2.bin.s2t_ctc_align import CTCSegmentation` stops working; the script's own arguments are unchanged.
- `espnet2/bin/align.py` — torchaudio forced alignment, what `espnet align` and the MCP server call.

## What stays exactly as it was

Both modules keep the public `CTCSegmentation` name, the same constructor arguments, the same `CTCSegmentationTask` import, and the same command-line arguments — so `egs2/owsm_v4/s2t1/local/ctc_seg.py`, the OWSM-CTC v3.1 README and the four references in the top-level README keep working untouched.

Two model differences are hooks now instead of copies:

- `_lpz_and_length` — OWSM-CTC decodes padded windows, so it reports how many samples its posteriors cover; an ASR model covers its input. `get_lpz` keeps the return type each has had.
- `_tokenize` — the S2T preprocessor expects a language and a task symbol around the text, so it goes through `text_cleaner` / `tokenizer` / `token_id_converter` instead of `preprocess_fn("<dummy>", …)`.

`__init_subclass__` gives each subclass its own `CtcSegmentationParameters`: `set_config` writes into that object, and inheriting one would make configuring either model configure the other. `test_ctc_segment.py` asserts it.

## Verified

Before and after, same inputs, same machine:

- Both models, both text converters, and `estimate_samples_to_frames_ratio` — **byte-identical** segments and confidence scores.
- `asr_align.py --audio test_utils/ctc_align_test.wav` end to end and `s2t_ctc_align.ctc_align(…)` — **byte-identical** output.
- `--help` for both scripts — identical but for each parser's description, which now says what the script does (`ASR Decoding` → `CTC segmentation, with an ASR model`).
- `test_asr_align.py`, `test_s2t_align.py`, `test_from_pretrained_routing.py`, `test_cli.py` and the new `test_ctc_segment.py`: 131 passed, 2 skipped. (`test_s2t_align.py` trips its 5 s `execution_timeout` on this Mac — it does so on master too, at the same place.)


sw005320:
Thank you. One correction to the premise first, because it changes the answer.

There is no `CTCSegmentation` in `espnet2.bin.ctc_segment`. The shared class is `AbsCTCSegmentation`, and it has no model: no `__init__` that loads one, and `get_lpz` raises `NotImplementedError`. So there are **two** public import paths, not three, and they are not two spellings of one class — they take different constructor arguments (`asr_train_config` / `asr_model_file` against `s2t_train_config` / `s2t_model_file`, plus `lang_sym`, `task_sym`, `context_len_in_secs`) and different checkpoints. Which one you import *is* the choice of model kind. That is also why there is no single class here that picks: `from_pretrained` on either one raises a `TypeError` naming the argument when handed the other kind's tag, which is what every other class in `espnet2.bin` does.

The thing worth avoiding is a reader having to work that out from the class hierarchy, so I have said it in the three docstrings and in the README:

- `ctc_segment.py`: "This module is not an entry point. `AbsCTCSegmentation` has no model, so it is never instantiated: a caller imports `CTCSegmentation` from `espnet2.bin.asr_align` for an ASR checkpoint or from `espnet2.bin.s2t_ctc_align` for an OWSM-CTC one."
- `asr_align.py` / `s2t_ctc_align.py`: "This module is where an ASR model is aligned from", with the import line.
- README, in the CTC segmentation section: the same sentence, plus a pointer to `espnet2/bin/align.py` for the forced alignment that takes either kind.

The one entry point that *does* pick for you is `espnet align` / `ForcedAligner` (#6777) — it works from the CTC head alone, so it needs to know nothing about the model beyond that. For `ctc_segmentation` the two paths stay, because the posteriors are produced differently.

Pushed as a separate commit so the refactor and the documentation are reviewable apart.

sw005320:
Renamed, as asked: the four modules now read alike, and each of them opens with the same map.

```
align.py        forced alignment over any CTC head; what `espnet align`
                and the MCP server run
asr_align.py    CTC segmentation with an ASR model, and its script
s2t_align.py    CTC segmentation with an OWSM-CTC model, and its script
ctc_segment.py  the CTC segmentation algorithm those two share; not an
                entry point
```

`espnet2/bin/s2t_ctc_align.py` → `espnet2/bin/s2t_align.py` (as a git rename, so the history follows), and with it `test/espnet2/bin/test_s2t_align.py`, `egs2/owsm_v4/s2t1/local/ctc_seg.py`, the OWSM-CTC v3.1 recipe README, the top-level README and the labeler rule.

Two things worth being explicit about:

- **`from espnet2.bin.s2t_ctc_align import CTCSegmentation` stops working.** I did not leave a shim: an ImportError naming the module is a clearer answer than a deprecation warning, and the in-repo callers are updated here. Say the word if you would rather have a one-release shim and I will add it.
- **The script's arguments do not change.** It is still `--s2t_train_config` and `--s2t_model_file`, so anything calling the script rather than importing it keeps working. Verified against the base: the `--help` listings differ only in the program name and the description line.

Re-ran after the rename: 66 passed, 2 skipped, and the before/after alignment output is still byte-identical to master for both models and both text converters.

sw005320:
Fixed — the old import path works again, so nothing breaks.

`espnet2/bin/s2t_ctc_align.py` stays as a forwarding module, in the same shape `espnet2/bin/s2t_inference_ctc.py` already uses for the class that moved out of it: importing it is harmless, and constructing its `CTCSegmentation` or running the script warns once and names the new module.

```
$ python -c "from espnet2.bin.s2t_ctc_align import CTCSegmentation; print(CTCSegmentation.__mro__[1].__module__)"
espnet2.bin.s2t_align

$ python espnet2/bin/s2t_ctc_align.py --help
espnet2/bin/s2t_ctc_align.py:54: DeprecationWarning: espnet2.bin.s2t_ctc_align is now
espnet2.bin.s2t_align, so that the alignment modules read alike; this one forwards and
will be removed in a future release
```

I was wrong to leave it out: the import line is in the OWSM-CTC v3.1 recipe README, which people have copied into their own scripts, and an ImportError there is a worse answer than a warning.

`test_s2t_align.py` now aligns a recording through the old name and asserts the warning, so the path is kept working rather than merely importable. 66 passed, 2 skipped.

sw005320:
Good catch, and my mistake: `addopts=` was a zero-byte file, not a change.

It is what a pytest invocation of mine left behind in the worktree — this repository's `setup.cfg` sets

```
addopts = --cov-config=.coveragerc --cov=espnet --cov=espnet2 --cov=espnet3
```

so a local run without pytest-cov dies on "unrecognized arguments", and I turned it off with `-o addopts=`. The value got quoted away in a way that made a file of the name instead, and a later `git add -A` picked it up with the docstring commit.

Removed. The PR is the six files it should be — and the lesson is the one already in my notes about `git add -A`: stage by path.

PR #6790, opened 2026-09-21
Two ways an alignment goes wrong quietly

sw005320:
Both found by writing the alignment notebook for [espnet/notebook#61](https://github.com/espnet/notebook/pull/61) and running it against `espnet/owsm_ctc_v4_1B` on `test_utils/ctc_align_test.wav`. Neither is a wrong answer; both are a right answer that says nothing about being useless.

## 1. A text a few tokens too long got torchaudio's words, not ours

```
RuntimeError: forced_align_impl, .../compute.cpp:56, targets length is too long for CTC.
Found log_probs length: 82, targets length: 80, and number of repeats: 3
```

80 tokens against 82 frames passes `len(ids) > len(emissions)`, but CTC needs a blank frame between two of the same token in a row, so those 3 repeats meant 83 frames were needed. The guard counts them now, and says so:

```
80 tokens and 3 blank(s) between repeated ones to align against 82 frames:
this text cannot fit in this recording
```

## 2. A score of 0.0000 with boundaries that look right

The first text anyone tries against that recording is its own reference transcript, which is capitalised:

```
The sale of the hotels                   0.990      (6 tokens)
THE SALE OF THE HOTELS                   0.000     (18 tokens)
```

OWSM's vocabulary has no capitalised words, so each one breaks into letters — `▁T H E ▁ SA L …` — and the mean probability collapses. The times stay roughly right (0.64–1.76 against 0.00–1.44), which is exactly what stops a reader from suspecting the text.

Nothing can be normalised away: the checkpoint's config says `cleaner: null`, OWSM was trained on cased text, and an ASR model trained on WSJ wants the capitals. So this compares instead of assuming — if the same words in another case need less than half as many tokens, that is the spelling the vocabulary has, and the caller is told which:

```
'THE SALE OF THE HOTELS' is 18 tokens for this model, where 'the sale of the
hotels' is 6: the text is spelled a way the vocabulary does not have, and the
score will be near zero however well the audio matches
```

Measured, on that model and that file:

| text | warnings | scores |
|---|---|---|
| natural case | 0 | 0.99, 0.77, 0.99, 0.99 |
| the same words in capitals | 4 | 0.00, 0.00, 0.00, 0.00 |

`Segment.score`'s docstring carries the same example, since the docstring is what promised "1.0 is a perfect match". The check never breaks a run: a tokenizer or converter that cannot encode the variant at all is passed over in silence.

## Tests

`test_align.py`, on the hand-written emissions the file already uses: three frames and `aab`, which needs four; and a vocabulary holding `abc` as one token beside the letters `A`, `B`, `C`, where `ABC` warns and `abc` does not (under `simplefilter("error")`). 11 passed.


sw005320:
Second commit, from the same notebook session and the same recording.

`Segment.score` promises "1.0 is a perfect match", and the first text a reader tries against `test_utils/ctc_align_test.wav` is that file's own reference transcript, which is in capitals:

```
The sale of the hotels                   0.9902
THE SALE OF THE HOTELS                   0.0000
```

OWSM's vocabulary has no capitalised words, so each one breaks into letters — `▁T H E ▁ SA L …`, 79 tokens for the four utterances instead of 24 — and the mean probability collapses. The boundaries stay roughly right (0.00–1.44 against 0.64–1.76), which is what makes it confusing rather than obviously wrong.

Not a code change: the aligner is doing what it says. The docstring now says which text the score is a probability of.

